PR #1931, opened 2023-03-13
fork testing

yorhodes:
- Add CI forking to infra scripts
- Add test-env job for fork testing to CI
- Run checker before running anvil

### Description

_What's included in this PR?_

### Drive-by changes

### Related issues

- Fixes #[issue number here]

### Backward compatibility

Yes
No

None
Yes

### Testing

None
Manual
Unit Tests

### DIFF
--- a/typescript/infra/scripts/core.ts
+++ b/typescript/infra/scripts/core.ts
@@ -1,3 +1,5 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+
 import {
   buildContracts,
   coreFactories,
@@ -6,6 +8,7 @@ import {
 
 import { deployEnvToSdkEnv } from '../src/config/environment';
 import { HyperlaneCoreInfraDeployer } from '../src/core/deploy';
+import { forkAndImpersonateOwner } from '../src/utils/fork';
 import { readJSON, writeJSON } from '../src/utils/utils';
 
 import {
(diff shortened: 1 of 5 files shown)


nambrot:
Taking me off review until it's ready (just assuming its not since you assigned me before all these commits)

typescript/infra/scripts/core.ts
```diff
@@ -40,6 +50,31 @@ async function main() {
     console.info('Could not load partial core addresses, file may not exist');
   }
 
+  if (argv.fork) {
+    await useLocalProvider(multiProvider, argv.fork);
```
nambrot, line 54: Couldn't you just set useLocalProvide before line 33 and then reuse the contract wrapper creation above?

typescript/infra/scripts/govern.ts
```diff
@@ -1,15 +1,33 @@
...
+} from './utils';
 
 async function check() {
-  const environment = await getEnvironment();
+  const argv = await getArgsWithFork().argv;
+  const environment = assertEnvironment(argv.environment);
   const config = getCoreEnvironmentConfig(environment);
 
-  const multiProvider = await config.getMultiProvider();
+  const multiProvider =
```
nambrot, line 22: This feels like common behavior to maybe be extracted?

typescript/infra/scripts/govern.ts
```diff
@@ -21,11 +39,20 @@ async function check() {
...
+    await governor.governChain(argv.fork, true);
+    return;
+  }
+
+  await coreChecker.check();
   await governor.govern();
 }
 
-check().then(console.log).catch(console.error);
+check().then().catch(console.error);
```
nambrot, line 58: Again exit 1 if there are errors

typescript/infra/src/core/govern.ts
```diff
@@ -69,14 +69,27 @@ export class HyperlaneCoreGovernor {
     }
   }
 
-  protected async sendCalls(chain: ChainName) {
+  async governChain(chain: ChainName, confirm = false) {
```
nambrot, line 72: I would suggest to confirm to true by default, and override to false when we are in CI

nambrot (approved):

typescript/infra/scripts/core.ts
yorhodes: yes but I am not passing the previousContracts into the same method

PR #1935, opened 2023-03-14
Graceful handling of errors with key funder's IGP claim or L2 bridging

tkporter:
### Description

Noticed that if an IGP claim fails, it's not gracefully handled

### Drive-by changes

none

### Related issues

n/a

### Backward compatibility

Yes

None

### Testing

Ran locally

### DIFF
--- a/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
+++ b/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
@@ -372,24 +372,6 @@ class ContextFunder {
   async fund(): Promise<boolean> {
     let failureOccurred = false;
 
-    // Returns whether an error occurred
-    const gracefullyHandleError = async (
-      fn: () => Promise<void>,
-      chain: ChainName,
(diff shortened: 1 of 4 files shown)


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -372,16 +372,43 @@ class ContextFunder {
   async fund(): Promise<boolean> {
     let failureOccurred = false;
 
+    // Returns whether an error occurred
+    const gracefullyHandleError = async (
```
mattiecnvr, line 376: I feel like this should be made a function as it is not specific to `fund` and does not capture any local values.

mattiecnvr (approved):

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
tkporter: good call

Issue #1728, opened 2023-02-01
Supporting Arbitrum in GasPaymentPolicyOnChainFeeQuoting


tkporter:
To recap, Arbitrum gas amounts are special. Each opcode that's executed on Arbitrum has the exact same gas cost as any other EVM chain, so the actual amount of gas used by the opcodes that are executed has no difference, but Arbitrum will add an amount of gas to the transaction in addition to the opcode usage (similar to intrinsic gas) that is intended to cover L1 costs. This is because the serialized Arbitrum transactions are posted to L1 Ethereum as calldata

You can think of the gas amount for Arbitrum transactions as the sum of L2 gas and L1 gas. The L2 gas is just "normal" gas that we expect on any other chain -- e.g. this covers gas used by opcodes, L2 calldata, etc. Then the L1 gas covers the cost of posting the transaction to the L1. This is a good resource if you're curious and want a deeper explanation https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9. The L1 gas component changes depending on how big the tx is (more bytes => more calldata => more expensive to post on L1), the L1 gas price (higher L1 gas price means it's more expensive to post transactions to L1), and L2 gas price (a lower L2 gas price means that to cover the L1 costs, more gas needs to be charged)

Anyways, we now have a situation - users on chain are specifying a gas amount that relates to the "L2 gas" (because they specify the amount of gas they expect their logic to consume on the destination), but in the relayer when we make the `eth_estimateGas` call to estimate gas for processing a message, we're getting the full L1 gas + L2 gas estimation. So we need to do a couple things...
1. On-chain, we need to make sure that the oraclized gas price that we use when charging ppl for txs sent to Arbitrum considers the L1 gas costs. This is out of scope for this issue
2. In the relayer, we need to change the GasPaymentPolicyOnChainFeeQuoting to compare the L2 gas amount used by a message to how much has been paid for via interchain gas payments. We can't just use the value we get from `eth_estimateGas` (which is the `tx_cost_estimate.gas_limit`), because this is the sum of L2 gas + L1 gas. This is what this issue seeks to address

[...]

PR #1940, opened 2023-03-15
Use gasEstimateComponents instead of gasEstimateL1Component for arbitrum gas

tkporter:
### Description

DRAFT - opening to get an image I can deploy to testnet3 rc and play around with

See #1728 for full context

In playing around with the existing version where we:
1. Do a normal eth_estimateGas to get the "total" L1 + L2 gas limit
2. Then call gasEstimateL1Component to get the L1 gas limit so we can subtract and get the L2 gas limit

I noticed that it took a few minutes for a message to arbitrumgoerli to eventually observe an L2 gas limit that made sense. For a basic helloworld tx that should only use ~150k-200k L2 gas, sometimes logs were showing l2_gas_limit as 800k, 1M+, 400k, etc. My best guess is that because we were doing the eth_estimateGas in one step and then the L1 component in a separate step, there's a mismatch. The L1 amount seems to vary more over time than I expected

So I'm gonna try using `gasEstimateComponents` instead -- this returns the total L1+L2 gas as well as the L1 gas all in a single atomic step. The downside here is that the `gasEstimateComponents` call will revert if the call we're estimating reverts -- and because of the way we use retrying provider, this can have an unintentional effect of retrying the call that we really shouldn't be. Now that I'm thinking of things, I think there's some room for improvement wrt failed calls and retrying...

Anyways, not fully intending on merging this yet, this is more to help prove my theory and then I'll consider how we can safely use `gasEstimateComponents`

### Drive-by changes

### Related issues

- Fixes #[issue number here]

### Backward compatibility

Yes
No

None
Yes

### Testing

None
Manual
Unit Tests

### DIFF
--- a/rust/agents/relayer/src/msg/gas_payment/mod.rs
+++ b/rust/agents/relayer/src/msg/gas_payment/mod.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 
 use async_trait::async_trait;
 use eyre::Result;
-use tracing::{error, trace};
+use tracing::{debug, error, trace};
 
 use hyperlane_core::{
     db::HyperlaneDB, HyperlaneMessage, InterchainGasExpenditure, InterchainGasPayment,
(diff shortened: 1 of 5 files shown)


tkporter:
replaced by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1949

PR #1952, opened 2023-03-16
Require >30 gwei for Polygon

### LINKED ISSUE #1933: Relayer should always use max priority fee of >= 30 gwei for Polygon txs
Okay so I found this post which is how I figured it out https://www.curvegrid.com/blog/2022-11-21-polygon-transaction-underpriced

The polygon team suggests node operators to reject transactions whose max priority fee (i.e. tip to the validator) is < 30 gwei
The way they suggest doing this is by suggesting to set `--miner.gasprice`  (minimum gas price for a tx to be included in a block) and the `--txpool.pricelimit` (minimum gas price to allow a transaction into the mempool) both to 30 gwei

BUT if you look at the logic that ethers-rs uses for estimating EIP-1559 fees https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/rpc/provider.rs#L417-L444, it figures out what the `max_fee_per_gas` and `max_priority_fee_per_gas` should be by using the `eth_feeHistory` RPC (which returns some info about the past N blocks), and running that data through an "estimator"

That `eth_feeHistory` RPC looks like this, where N=10 which is how many blocks you want info about (this is `utils::EIP1559_FEE_ESTIMATION_PAST_BLOCKS`, and [5.0] is `&[utils::EIP1559_FEE_ESTIMATION_REWARD_PERCENTILE]`:
```
$ cast rpc eth_feeHistory --rpc-url https://falling-crimson-waterfall.matic.quiknode.pro/7f7ee507021fea5aeab05afbd7554d5b0bf632b4/ 10 latest "[5.0]"
{"oldestBlock":"0x2679104","reward":[["0x7580e728c"],["0x7580e728c"],["0x7580e728c"],["0x7137ae102"],["0x7580e728c"],["0x7a4f7187c"],["0x773594000"],["0x71ff1a746"],["0x7a55cfce3"],["0x81347790c"]],"baseFeePerGas":["0x20a3b60ed5","0x20c8193738","0x20466142d0","0x1fb8a1a3b1","0x1f4c347bfa","0x1e51f684bd","0x1e57f010a8","0x1de8b5347d","0x1e8cda3200","0x1e23846282","0x1df0a24eef"],"gasUsedRatio":[0.5348382708079717,0.3763421696435745,0.36275282830371325,0.39318426558205527,0.25013913209513483,0.5061580036096702,0.3854471541209541,0.6715039674101666,0.3922519510699354,0.4472407355786751]}
```

So the issue is that ethers-rs will suggest a max priority fee *based on the contents of the last 10 blocks*, and because the 30 gwei suggestion by Polygon is just a suggestion (i.e. not enforced at the protocol lvl and is entirely up to the nodes to opt in), it's very possible for there to be recent blocks with priority fees < 30 gwei. So ethers-rs suggests a priority fee lower than 30 gwei, but our node providers have set their `--txpool.pricelimit` to 30 gwei!

To confirm this was the case, I looked at one of the logs from the relayer where we got `"transaction underpriced"`-- this pinned one https://cloudlogging.app.goo.gl/Ykh4AZFVz7bXqXeZ9

If you get the raw tx from `spans[3]` and pass it into `ethers.utils.parseTransaction('0x....')` you can decode the raw transaction hex

And lo and behold:
```
{
  type: 2,
  chainId: 137,
  nonce: 322,
  maxPriorityFeePerGas: BigNumber { _hex: '0x05dc55560c', _isBigNumber: true },
  maxFeePerGas: BigNumber { _hex: '0x36add4b7f8', _isBigNumber: true },
  gasPrice: null,
  gasLimit: BigNumber { _hex: '0x042b98', _isBigNumber: true },
```
the max priority fee is less than 30 gwei!
```
$ cast td 0x05dc55560c
25171416588
$ cast --to-unit 25171416588 gwei
25.171416588
```

And to replicate this I tried running that `estimate_1559_fees` function from the problematic block when that log occurred:
https://github.com/tkporter/polygon-priority-fee-debugging/blob/main/src/main.rs

```
$ cargo run -- "https://falling-crimson-waterfall.matic.quiknode.pro/<omitted>"

Latest block's suggestion: max_fee_per_gas "181.226715279" max_priority_fee_per_gas "32.758923569"
Problematic block 40340495's suggestion: max_fee_per_gas "221.061717847" max_priority_fee_per_gas "25.171416588"
```

You can see the `25.171416588` is the exact same as what was in that tx

Therefore I suggest that we always use a min priority fee of 30 gwei on Polygon

nambrot:
### Description

Fixes #1933. It's not pretty but I would like to sleep tonight


rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -225,10 +225,20 @@ where
                 .await?
                 .saturating_add(U256::from(GAS_ESTIMATE_BUFFER))
         };
+        let chain_id = match self.provider.get_chainid().await {
```
mattiecnvr, line 228: Is this needed? Seems like it is not polygon specific but handling something that should have already been?

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -225,10 +225,20 @@ where
...
+            Ok(chainId) => chainId.as_u32(),
+            // Couldn't get chainId, assume not 1559
+            Err(_) => return Ok(tx.gas(gas_limit))
+        };
         let Ok((max_fee, max_priority_fee)) = self.provider.estimate_eip1559_fees(None).await else {
             // Is not EIP 1559 chain
             return Ok(tx.gas(gas_limit))
         };
+        let max_priority_fee = if KnownHyperlaneDomain::try_from(chain_id)? == KnownHyperlaneDomain::Polygon {
```
mattiecnvr, line 237: Maybe use
```rust
if matches!(KnownHyperlaneDomain::try_from(chain_id), Ok(KnownHyperlaneDomain::Polygon)) {..} else {..}
```

mattiecnvr (approved):

rust/chains/hyperlane-ethereum/src/mailbox.rs
nambrot: 1933 is pretty explicit about it that it is polygon specific no?

rust/chains/hyperlane-ethereum/src/mailbox.rs
nambrot: Would you mind sharing what the benefit of that is? It seems less readable to me?

tkporter: We shouldn't use `?` because then an unknown chain (i.e. a PI chain) that supports the `eth_chainId` RPC will always have errors

My suggestion is to get rid of the `let chain_id` above altogether, which imo isn't needed and is potentially inaccurate because we have access to `message.destination`. The EVM chain_id isn't always going to be the Hyperlane domain ID-- it often is, but it's not guaranteed to be, especially not in PI chains

And instead I think we should do exactly what mattie suggests, which involves the `else` block being used if the message.destination isn't a "known" domain. But instead of chain_id use message.destination:
```
if matches!(KnownHyperlaneDomain::try_from(message.domain), Ok(KnownHyperlaneDomain::Polygon)) {..} else {..}
```

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -225,10 +225,23 @@ where
...
+        let chain_id = match self.provider.get_chainid().await {
+            Ok(chainId) => chainId.as_u32(),
+            // Couldn't get chainId, assume not 1559
+            Err(_) => return Ok(tx.gas(gas_limit)),
+        };
         let Ok((max_fee, max_priority_fee)) = self.provider.estimate_eip1559_fees(None).await else {
             // Is not EIP 1559 chain
             return Ok(tx.gas(gas_limit))
         };
+        let max_priority_fee =
```
tkporter, line 237: flagging my comment here bc it's outdated https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1952#discussion_r1140154886

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -225,10 +225,23 @@ where
...
+            if KnownHyperlaneDomain::try_from(chain_id)? == KnownHyperlaneDomain::Polygon {
+                let min_polygon_fee = ethers::utils::parse_units("31", "gwei").unwrap().into();
```
tkporter, line 239: fwiw the condition is >= 30 gwei and not > 30 gwei

Also do we need to unwrap here?

This would probably be most preferable imo:
```
// 30 gwei
let min_polygon_fee = U256::from(30000000000);
```

or if you don't like that maybe
```
let min_polygon_fee = ethers::utils::parse_units("30", "gwei")?.into();
```

rust/chains/hyperlane-ethereum/src/mailbox.rs
mattiecnvr: No, I mean, if you move this match statement to after the `estimate_eip_1559_fees` call you could just `.expect("all 1559 chains should have a chain id")` rather than using a whole match statement.

rust/chains/hyperlane-ethereum/src/mailbox.rs
nambrot: Ahhhh now I get it

rust/chains/hyperlane-ethereum/src/mailbox.rs
nambrot: I actually had your suggestion first (from codepilot!) and thought people liked the use of the utility.

tkporter (approved):
Rip it

asaj (approved):
😴

Issue #1956, opened 2023-03-17
Deploy new testrecipient & unclog message queues


tkporter:
Probably only relevant on testnet, but should confirm we don't have the huge message queue issue on mainnet too

PR #1957, opened 2023-03-17
Deploy new TestRecipients & blacklist old one on testnet

tkporter:
### Description

Context https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1956

### Drive-by changes

New TestRecipient salts, made TestRecipient ownable so that there's some access control on setting the ISM

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1956

### Backward compatibility

Yes

None

### Testing

Deployed

### DIFF
--- a/solidity/contracts/test/TestRecipient.sol
+++ b/solidity/contracts/test/TestRecipient.sol
@@ -1,10 +1,13 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity >=0.8.0;
 
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+
 import {IMessageRecipient} from "../../interfaces/IMessageRecipient.sol";
 import {IInterchainSecurityModule, ISpecifiesInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
 
 contract TestRecipient is
+    Ownable,
     IMessageRecipient,
     ISpecifiesInterchainSecurityModule
 {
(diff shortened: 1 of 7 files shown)


typescript/infra/src/testcontracts/testrecipient.ts
```diff
@@ -33,17 +33,24 @@ export class TestRecipientDeployer extends HyperlaneDeployer<
...
+          [deployer],
+        ),
+      },
     );
     const TestTokenRecipient = await this.deployContract(
       chain,
       'TestTokenRecipient',
       [],
-      { create2Salt: 'TestTokenRecipient' },
+      { create2Salt: 'TestRecipient-March-17-2023' },
```
tkporter, line 53: Probably should've done `TestTokenRecipient-March-17-2023`, but oh well

solidity/contracts/test/TestRecipient.sol
```diff
@@ -42,4 +41,8 @@ contract TestRecipient is
         lastCaller = msg.sender;
         lastCallMessage = message;
     }
+
+    function setInterchainSecurityModule(address _ism) external onlyOwner {
```
nambrot, line 45: Do you need to initialize or does it happen with the constructor?

nambrot (approved):

solidity/contracts/test/TestRecipient.sol
tkporter: Just defaults to `address(0)` which means the default ISM is used

PR #1675, opened 2023-01-26
Deploy to Gnosis

nambrot:
### Description

This PR includes the configuration and deployment artifacts for the Gnosis deployment. Did not deploy middlewares as that has to be re-done anyways

### Drive-by changes

- Changes the Create2Deployer to hardcode the bytecode to be independent of compiler changes
- Dont submit `transferOwnership` txs in parallel in `HyperlaneCoreDeployer` to avoid nonce collision

### Related issues

- Fixes #1551

PR #1959, opened 2023-03-17
Change ethereum finality in config 14 -> 20

tkporter:
### Description

See https://github.com/hyperlane-xyz/issues/issues/415 for context

Essentially all validators are on 20 block finality but we changed the relayer to 14 [recently](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1675/files#diff-67e27b51135e9801b172be277ee0f667ed7b5ea5cb600f987bceb03b75eb3d65R33). It results in a weird scenario where the relayer will retry a message a few times before there's a quorum from validators. Moving back to 20 in the relayer

### Drive-by changes

* Added names to validator addresses which I found useful to know at a glance

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/415

### Backward compatibility

yes

Sorta. We should re-deploy agents from this commit

### Testing

None

### DIFF
--- a/rust/config/mainnet_config.json
+++ b/rust/config/mainnet_config.json
@@ -29,7 +29,7 @@
       },
       "signer": null,
       "protocol": "ethereum",
-      "finalityBlocks": 14,
+      "finalityBlocks": 20,
       "connection": {
         "type": "http",
         "url": ""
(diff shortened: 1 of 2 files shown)


asaj (approved):
Oops I just labeled these validators in my SDK PR :)

asaj (approved):
Dang you beat me to labeling the validators, I have an open PR with that :)

tkporter:
lollll oops

PR #1962, opened 2023-03-18
Fix explorerAddressURL when no signer

nambrot:
### Description

Currently, if the MP doesn't have a signer, the function will always return null


jmrossy (approved):
Good catch, thanks!

PR #1966, opened 2023-03-20
Update submodules

tkporter:
### Description

In `fuel-contracts`, we use the `hyperlane-core` rust package and point to it via git. When cargo tries to install it, it'll look at the entire `hyperlane-monorepo` and install any git submodules. Apparently it breaks if the submodule points to a commit that doesn't belong to a branch in the repo (e.g. look at the warning of https://github.com/hyperlane-xyz/hyperlane-token/commit/f40093e or https://github.com/hyperlane-xyz/hyperlane-app-template/commit/4d7159fe92dbfe814b6de1668f2d91c9d2a2b175). We should really be using commits from main anyways

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

Yes

None

### Testing

unit tests


jmrossy:
+1 for commits from main

jmrossy (approved):

PR #1970, opened 2023-03-20
Portal 

nambrot:
### Description

Log when wrapped token has not been created and catch exceptions


asaj (approved):

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
```diff
@@ -208,12 +208,27 @@ export class LiquidityLayerApp extends HyperlaneApp<LiquidityLayerContracts> {
...
+      await this.multiProvider.handleTx(
+        message.destination,
+        destinationPortalAdapter.completeTransfer(
+          utils.ensure0x(Buffer.from(vaa.vaaBytes, 'base64').toString('hex')),
+        ),
+      );
+    } catch (error: any) {
+      if (
+        error?.error?.reason &&
+        error.error.reason.includes('no wrapper for this token')
```
jmrossy, line 222:
```suggestion
        error?.error?.reason?.includes('no wrapper for this token')
```

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
```diff
@@ -208,12 +208,27 @@ export class LiquidityLayerApp extends HyperlaneApp<LiquidityLayerContracts> {
...
+      ) {
+        console.log(
```
jmrossy, line 224: Would be good to move this file off console to debug like the rest of the SDK

jmrossy (approved):

typescript/infra/scripts/middleware/portal-relayer.ts
```diff
@@ -39,16 +40,27 @@ async function relayPortalTransfers() {
...
+      console.log('chain', chain);
+
       const txHashes = await app.fetchPortalBridgeTransactions(chain);
       const portalMessages = (
         await Promise.all(
           txHashes.map((txHash) => app.parsePortalMessages(chain, txHash)),
         )
       ).flat();
 
+      console.log('Portal messages', portalMessages);
```
tkporter, line 52: I know I added this but consider using the `log` from `@hyperlane-xyz/utils` so that this doesn't take up tons of lines

typescript/infra/scripts/middleware/portal-relayer.ts
```diff
@@ -39,16 +40,27 @@ async function relayPortalTransfers() {
 
   while (true) {
     for (const chain of Object.keys(bridgeAdapterConfigs)) {
+      console.log('chain', chain);
```
tkporter, line 43: same here, maybe we should make the message more significant, like `'Assessing chain'` or something

tkporter (approved):

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
nambrot: Does this actually work? I thought otherwise this ends up as `undefined.includes()` which throws an error?

PR #1973, opened 2023-03-20
Change relayer backoff scaling

### LINKED ISSUE #1896: Relayer backoff results in 5 minute delay
From: https://discord.com/channels/935678348330434570/935678739663192184/1081348632101519420

Just did the Circle Demo and one thing I found was that the message took a surprisingly long time to process. Logs are here: https://cloudlogging.app.goo.gl/VoYwPgRLxpU5grZ59, explorer link https://explorer.hyperlane.xyz/message/ed42f4dfd5ce96aa6e027a603d4b55bd33b8d6bb2f4cb5d107112861af47f127

I first suspected that it's the underlying Circle transfer that took so long, but that wasn't the case. Instead I think its our backoff logic: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/03d32f4009611ae4afd044c1e95a89d2676b7df0/rust/agents/relayer/src/msg/serial_submitter.rs#L196

IIUC, what happened is that we retried to message in quick succession 16 times in the span of 20 seconds, and after 16 times, we basically enforce a minimum 5 minute wait. I'm sure there is a meme somewhere about how hard it is to specify good backoff functions, but IMO we should something like "step-function-y"? I.e. something along the lines of 1,2,4,8,16,32,45,60,75,90 seconds etc. until 5 minutes, and then retry every 5 minutes?

Separately, it looks like "rusoto_core::request" log lines are a bit verbose, even at DEBUG, idk that individual HTTP headers need to be logged? cc @mattiecnvr

mattiecnvr:
### Description

This adds more backoff to earlier requests so that a request which is sent when there are no others won't be retired constantly until it ends up in the 5min backoff phase.

### Drive-by changes

None

### Related issues

- Fixes #1896 

### Backward compatibility

Yes

None

### Testing

Unit Tests


asaj (approved):

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -193,11 +193,11 @@ impl SerialSubmitter {
...
 
-        if msg.num_retries >= 16 {
+        if msg.num_retries >= 1 {
             let required_duration = Duration::from_secs(match msg.num_retries {
-                i if i < 16 => unreachable!(),
-                // wait 5 min
-                i if (16..24).contains(&i) => 60 * 5,
+                i if i < 1 => unreachable!(),
+                // wait 20s to 7.7min with a linear increase
+                i if (1..24).contains(&i) => i as u64 * 20,
```
nambrot, line 200:
```suggestion
                // wait 10s
                i if (1..12).contains(&i) => 10,
                // wait 20s to 7.7min with a linear increase
                i if (12..24).contains(&i) => i as u64 * 20,
```

Obviously this is more of an art than a science, but what do you think about this maybe? I would still love to maintain responsiveness for the fast finality chains as much as we can and not drop down to 20/40/60s if validators are just slightly slow

nambrot (approved):

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: Went ahead and took the flat 10 for the first few and then altered the linear scaling to make more sense between it and the exponential backoff.

PR #1975, opened 2023-03-20
Better rust package management

mattiecnvr:
### Description

At the suggestion of @tkporter this moves common dependencies and configuration to use the new(ish) cargo workspace configuration feature. This was added in July 2022 and allows us to ensure all versions and configurations are the same across different crates. It does not decrease the typing, but it does reduce the updates required when we go to update a package and keeps duplicates next to each other.

This PR does not change any package versions or add/remove any packages. It only moves the configuration.

### Drive-by changes

Also sorted packages by name.

### Backward compatibility

Yes

None

### Testing

Manual


rust/Cargo.toml
```diff
@@ -17,3 +17,36 @@ members = [
...
+license-file = "../LICENSE.md"
+publish = false
+version = "0.1.0"
+
+[workspace.dependencies]
+async-trait = { version = "0.1" }
+color-eyre = { version = "0.6" }
+config = "~0.13.3"
+derive-new = "0.5"
+ethers = { git = "https://github.com/hyperlane-xyz/ethers-rs", tag = "2023-02-10-01", features = ["abigen"] }
```
tkporter, line 34: do you think it's worth having these features at the sub-package level or workspace level? Or is the thinking for features that are used by many packages, those are specified at the workspace level

tkporter (approved):

PR #2003, opened 2023-03-23
Agent release process

nambrot:
### Description

Adds instructions to the readme about a agent release process

Here is an example https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/tag/agents-1.0.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ messages will not be relayed between chains).
 Off-chain agents are **not** automatically re-deployed when new contract deploys
 are merged. Auto-redeploys will be implemented at some future date.
 
-### Publishing Packages
+### Publishing JS/TS Packages
 
 Packages can be versioned and published all at once with commands from the root.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,4 +103,11 @@ For the git submodules, you will have to undo the removal of the `yarn.lock` fil
...
 
-Make a Github release
+Tag the commit with the appropriate version, and then create a github release with a changelog against the previous version https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/new
+
+### Release Agents
+
+- Tag the commit with the appropriate version
+- [Create a Github Release](https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/new) with a changelog against the previous version
+- include the docker image tag in the description of the release
+- make sure that breaking changes are handled and documented
```
mattiecnvr, line 113: Also
`- add human readable list/summary of changes`

See the one I added to the 1.0.1 release

README.md
```diff
@@ -103,4 +103,11 @@ For the git submodules, you will have to undo the removal of the `yarn.lock` fil
 
 Make PRs for the monorepo and the submodules.
 
-Make a Github release
+Tag the commit with the appropriate version, and then create a github release with a changelog against the previous version https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/new
+
+### Release Agents
+
+- Tag the commit with the appropriate version
```
mattiecnvr, line 110: I think we should switch to dates instead of versions. Since we are packaging 3 binaries into one release, defining "breaking changes" by semver is a bit clunky. For example in the 1.0.1 release, the scraper deserves a `MAJOR` version bump as it changed the database format and requires re-indexing. The validator deserves a `PATCH` as it only had minor improvements. The relayer deserves a `MINOR` as new features were added such as `defaultSigner` and sepolia but it should not change how it is is run.

nambrot:
Would you or @tkporter shepharding this through, i just wanted to be a good Samaritan, but this should be owned by y'all imo

tkporter (approved):

PR #2006, opened 2023-03-23
Build only the agent bins in docker

mattiecnvr:
### Description

Minor change to slightly speed up agent image build times. We only need to build the agent binaries in the docker pipeline.

### Drive-by changes

None

### Backward compatibility

Yes

None

### Testing

Unit Tests


tkporter (approved):
what other stuff was there? like run-locally?

mattiecnvr:
> what other stuff was there? like run-locally?

Yes, also a couple of binaries for the scraper. It won't save a lot of time but it is at least a few less binaries to link.

PR #2014, opened 2023-03-24
Fix core fork tests

yorhodes:
Changed in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1907

### Testing

Fork tests


typescript/infra/fork.sh
```diff
@@ -28,7 +28,7 @@ echo "=== Run checker against forked $ENVIRONMENT ==="
 DEBUG=hyperlane:* yarn ts-node ./scripts/check-deploy.ts -e $ENVIRONMENT -f $FORK_CHAIN -m $MODULE
 
 echo "=== Run core deployer against forked $ENVIRONMENT ==="
-DEBUG=hyperlane:* yarn ts-node ./scripts/core.ts -e $ENVIRONMENT -f $FORK_CHAIN -m $MODULE
+DEBUG=hyperlane:* yarn ts-node ./scripts/deploy.ts -e $ENVIRONMENT -f $FORK_CHAIN -m $MODULE
```
asaj, line 31: Why was this failing silently, and how can we make it fail loudly?

asaj (changes requested):
Please do not merge until failures cause CI to fail

typescript/infra/fork.sh
```diff
@@ -24,11 +24,14 @@ while ! cast bn; do
   sleep 1
 done
 
+# exit 1 on any subsequent failures
```
asaj, line 27: Can/should we drop lines 42 and 46?

asaj (approved):

PR #2018, opened 2023-03-24
Script to fix submodules

mattiecnvr:
### Description

I keep finding myself running a few commands in sequence because of the submodules so I figured it was about time to turn it into a script to make things easier for us.

### Drive-by changes

None

### Backward compatibility

Yes

None

### Testing

Manual

### DIFF
--- a/unfuck-submodules.sh
+++ b/unfuck-submodules.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+DIRS=$(git submodule status | grep -oE "typescript/[a-zA-Z0-9_-]+")
+for DIR in $DIRS ; do
+    echo "Removing '$DIR'"
+    rm -rf "$DIR"
+done
+
+git submodule init
+git submodule update
+
+for DIR in $DIRS ; do
+    echo "Cleaning '$DIR'"
+    pushd "$DIR" || continue
+    rm -rf yarn.lock node_modules
+    popd || exit 1
+done
+
+yarn clean
+yarn install
+yarn build


unfuck-submodules.sh
```diff
@@ -0,0 +1,29 @@
...
+    rm -rf "$DIR"
+done
+
+git submodule init
+git submodule update
+
+for DIR in $DIRS ; do
+    echo "Cleaning '$DIR'"
+    pushd "$DIR" || continue
+    rm -rf yarn.lock node_modules
```
jmrossy, line 23: rm the module's `node_modules` dir shouldn't be necessary I think. Have they created problems for you?

unfuck-submodules.sh
```diff
@@ -0,0 +1,29 @@
...
+if [[ "$DIRNAME" != "hyperlane-monorepo" ]]; then
+  echo "Must be run from the root of the monorepo"
+  exit 1
+fi
+
+DIRS=$(git submodule status | grep -oE "typescript/[a-zA-Z0-9_-]+")
+for DIR in $DIRS ; do
+    echo "Removing '$DIR'"
+    rm -rf "$DIR"
+done
```
jmrossy, line 15: So this removes any dirs that have a TS change?

mattiecnvr: It removes any submodules that are in the typescript directory.

unfuck-submodules.sh
mattiecnvr: 🤷‍♀️ 
https://discord.com/channels/935678348330434570/1068617127268990996/1068618656822595675

unfuck-submodules.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
jmrossy, line 1: Suggest moving to a `/scripts` folder to avoid clutter at monorepo root

jmrossy (approved):
See one comment

unfuck-submodules.sh
nambrot, line 1: I'd probably suggest to change to a more reasonable script name for this

nambrot (changes requested):

unfuck-submodules.sh
nambrot, line 1: Can we add this as a yarn script in package.json?

nambrot (approved):

PR #2028, opened 2023-03-28
Scaffolding for multiple ISM types in the relayer

asaj:
### Description

This PR adds scaffolding in the relayer to support building metadata for multiple ISM types.

To add a new ISM type:
- Add a new implementation of MetadataBuilder for that type
- Add the type to the SupportedIsmTypes enum
- Add the type to the match statement in BaseMetadataBuilder

### Drive-by changes

None

### Related issues

- #1757 

### Backward compatibility

Yes

None

### Testing

E2E tests


rust/hyperlane-core/src/traits/ism.rs
```diff
@@ -0,0 +1,15 @@
...
+use async_trait::async_trait;
+use auto_impl::auto_impl;
+
+use crate::{ChainResult, HyperlaneContract};
+
+/// Interface for the Ism chain contract. Allows abstraction over
+/// different chains
+#[async_trait]
+#[auto_impl(&, Box, Arc)]
+pub trait Ism: HyperlaneContract + Send + Sync + Debug {
```
mattiecnvr, line 12: To match how we are doing things already, this should be `InterchainSecurityModule`. The file/module should also be renamed. I think it is fine to keep using things like `MultisigIsm` but the main trait should be spelled out by the conventions we have so far.

rust/agents/relayer/Cargo.toml
```diff
@@ -29,6 +29,8 @@ tracing.workspace = true
 hyperlane-core = { path = "../../hyperlane-core" }
 hyperlane-base = { path = "../../hyperlane-base" }
 hyperlane-ethereum = { path = "../../chains/hyperlane-ethereum" }
+num-derive = "0.3"
```
mattiecnvr, line 32: Since we also use `num-traits` and `num-derive` in `hyperlane-core`, we should probably make this a workspace dependency and update `hyperlane-core` to use it as well.

rust/agents/relayer/src/msg/metadata_builder.rs
```diff
@@ -37,49 +61,60 @@ impl Debug for MetadataBuilder {
     }
 }
 
-impl MetadataBuilder {
-    #[instrument(err, skip(mailbox))]
```
mattiecnvr, line 41: Might want to keep this `instrument`

rust/agents/relayer/src/msg/metadata_builder.rs
```diff
@@ -37,49 +61,60 @@ impl Debug for MetadataBuilder {
...
+        metadata_builder
+            .build(ism_address, message)
             .await
-            .context(CTX)?;
+            .context(CTX)
+    }
+}
+
+#[derive(Clone, new)]
+pub struct LegacyMultisigIsmMetadataBuilder {
```
mattiecnvr, line 94: nit: impl `Deref` for this so you don't have to type `.base` every time

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -279,6 +279,27 @@ impl ChainSetup {
         .context("Building ValidatorAnnounce")
     }
 
+    /// Try to convert the chain setting into a Multisig Ism contract
+    pub async fn build_ism(&self, address: H256, metrics: &CoreMetrics) -> Result<Box<dyn Ism>> {
```
mattiecnvr, line 283: Does the config not already have the ISM address? Feels odd since the others are all using `self.locator(&self.addresses.<name_of_contract>)`

rust/chains/hyperlane-ethereum/src/ism.rs
```diff
@@ -0,0 +1,115 @@
...
+    domain: HyperlaneDomain,
+}
+
+impl<M> EthereumIsm<M>
+where
+    M: Middleware + 'static,
+{
+    /// Create a reference to a mailbox at a specific Ethereum address on some
+    /// chain
+    pub fn new(provider: Arc<M>, locator: &ContractLocator) -> Self {
```
mattiecnvr, line 62: If you make this new fn `async` you can set the ISM type at init.

rust/chains/hyperlane-ethereum/src/ism.rs
```diff
@@ -0,0 +1,115 @@
...
+    HyperlaneProvider, Ism, H256,
+};
+
+use crate::contracts::i_interchain_security_module::{
+    IInterchainSecurityModule as EthereumIsmInternal, IINTERCHAINSECURITYMODULE_ABI,
+};
+use crate::trait_builder::BuildableWithProvider;
+use crate::EthereumProvider;
+
+impl<M> std::fmt::Display for EthereumIsmInternal<M>
```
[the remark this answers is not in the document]
asaj: Copied and pasted from another file, will remove

rust/chains/hyperlane-ethereum/src/ism.rs
asaj: Why is this better than calling it in metadata builder? IIUC if we put it here we need to duplicate the u8->enum logic for each implementation of the trait

rust/hyperlane-base/src/settings/chains.rs
asaj: There are no enshrined ISM addresses, recipients specify whatever ISM address they like.

rust/hyperlane-core/src/traits/ism.rs
```diff
@@ -0,0 +1,15 @@
...
+    /// Returns the validator and threshold needed to verify message
+    async fn module_type(&self) -> ChainResult<u8>;
```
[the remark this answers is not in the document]
asaj: Technically there's no requirement that it's static, though I expect them all to be.

rust/chains/hyperlane-ethereum/src/ism.rs
asaj: I guess the benefit is that if you query it in the constructor you can enforce that the async call is only made once.

rust/hyperlane-base/src/settings/chains.rs
mattiecnvr: Yes, but a given instance will still be attached to a single one, right?

rust/chains/hyperlane-ethereum/src/ism.rs
mattiecnvr: Yes, that

mattiecnvr (approved):

rust/chains/hyperlane-ethereum/src/ism.rs
asaj: The way things are currently structured there's no reason it would ever be called twice, so I'm not sure this is necessary. IMO there's something nice about all the constructors acting the same way (which does not involve contract calls).

rust/hyperlane-base/src/settings/chains.rs
asaj: No, a given instance will not be attached to a single one, the relayer works with whatever ISM is specified by the recipient of a given message.

PR #2029, opened 2023-03-28
Rust Fuel Updates

mattiecnvr:
### Description

This brings fuel up to the latest version. The main change required was to to abigen process. After some digging I figured out they just moved it to a new crate.

### Drive-by changes

- Updated an error message I ran into when trying to run the validator to be more descriptive
- Found an unlinked file which I was able to delete
- Add `*.ignore` files to gitignore. This allows me to block out some configs I was using for testing.

### Related issues

- Fixes hyperlane-xyz/issues#410

### Backward compatibility

Yes

None

### Testing

Manual - Ran the validator against a local node


rust/chains/hyperlane-fuel/src/validator_announce.rs
```diff
@@ -1,31 +0,0 @@
...
-}
-
-impl HyperlaneChain for FuelValidatorAnnounce {
-    fn domain(&self) -> &HyperlaneDomain {
-        todo!()
-    }
-}
-
-#[async_trait]
-impl ValidatorAnnounce for FuelValidatorAnnounce {
```
tkporter, line 24: I think we'll need this file when implementing https://github.com/hyperlane-xyz/fuel-contracts/issues/74, we should probably keep it even if it's not used atm?

PR #2036, opened 2023-03-29
Better Indexing

### LINKED ISSUE #1851: Audit `from` agent settings and increase sync speed when far from tip
Defaults to 2000 right now if not specified https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/a836ce21d39ea916257fa92a18bb44ff6d1e0194/rust/hyperlane-base/src/settings/chains.rs#L107

We should revisit if that is a reasonable chunk size to default on and for popular chains (or chains with high block production) that folks are likely going to index, we should consider increasing the chunk size to allow catching up to tip to be much faster (i.e. goerli, the rollups, binance and polygon maybe)

Otherwise running a relayer from one of those chains might be annoying

### LINKED ISSUE #1872: Indexers should log how far behind tip they are
When looking at relayer looks, you can see the block range it most recently indexed, but it would be nice if it told you how many blocks behind tip it is, and given the polling interval, an ETA of being up to tip. Otherwise it can be hard to know whether the indexing is working as expected

While you are at it, it might be nice to tackle this related problen from #1851 

> Another thing is that the logs get polluted imo when syncing up, i wonder if we should have individual "storing new message/gas payment" events as INFO and instead of just have the group of messages/payments as an info log.

mattiecnvr:
### Description

* Logs the distance from the tip
* Calculates and logs the time to fully sync
* Changes indexing timeout when catching up from 1s to 100ms

### Drive-by changes

* Cleans up some log messages at the `info` level.
* Refactors the indexing logic to be a bit cleaner/simpler
* Updated all indexers to use the eta calculations

### Related issues

- Fixes #1872
- Fixes #1851 

### Backward compatibility

Yes

None

### Testing

Manual


rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -30,43 +36,40 @@ where
...
         })
     }
 
     /// Wait based on how close we are to the tip and update the tip,
     /// i.e. the highest block we may scrape.
     async fn rate_limit(&mut self) -> ChainResult<()> {
+        let update_tip = self.last_tip_update.elapsed() >= Duration::from_secs(30);
         if self.from + self.chunk_size < self.tip {
             // If doing the full chunk wouldn't exceed the already known tip,
             // we don't necessarily need to fetch the new tip. Sleep a tiny bit
```
tkporter, line 49: nit: now that we don't return here, this comment should be changed to not mention anything about fetching the new tip

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -30,43 +36,40 @@ where
...
+        }
+        match self.indexer.get_finalized_block_number().await {
+            Ok(tip) => {
+                // we retrieved a new tip value, go ahead and update.
+                self.last_tip_update = Instant::now();
+                self.tip = tip;
+                Ok(())
+            }
+            Err(e) => {
+                warn!(error = %e, "Failed to get next block range");
```
tkporter, line 69: don't feel strongly but wondering if we should also mention that it's because we were unable to get the finalized block number?

rust/hyperlane-base/src/contract_sync/eta_calculator.rs
```diff
@@ -0,0 +1,65 @@
...
+        let new_rate = (blocks_processed - tip_progression) / elapsed;
+
+        // Calculate the effective rate using a moving average. Only set the past
+        // effective rate once we have seen a move to prevent it taking a long
+        // time to normalize.
+        let effective_rate = if let Some(old_rate) = self.effective_rate {
+            let new_coeff = (elapsed / self.time_window).min(0.9);
+            let old_coeff = 1. - new_coeff;
+
+            let er = new_rate * new_coeff + old_rate * old_coeff;
```
tkporter, line 46: nit: sorry that I always ask for this but I just really dislike implied order of operations and would prefer explicit parentheses

rust/hyperlane-base/src/contract_sync/eta_calculator.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        self.last_block = current_block;
+        self.last_tip = current_tip;
+        let new_rate = (blocks_processed - tip_progression) / elapsed;
+
+        // Calculate the effective rate using a moving average. Only set the past
+        // effective rate once we have seen a move to prevent it taking a long
+        // time to normalize.
+        let effective_rate = if let Some(old_rate) = self.effective_rate {
+            let new_coeff = (elapsed / self.time_window).min(0.9);
```
tkporter, line 43: while unlikely, it's possible for this to be > 1 right? which iiuc would mess up the calculation? if so, maybe worth setting a max too

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -3,9 +3,14 @@ use std::time::{Duration, Instant};
 use async_trait::async_trait;
 use eyre::Result;
 use tokio::time::sleep;
+use tracing::warn;
 
+use crate::contract_sync::eta_calculator::SyncerEtaCalculator;
 use hyperlane_core::{ChainResult, Indexer, SyncBlockRangeCursor};
 
+/// Time window for the moving average used in the eta calculator.
```
tkporter, line 11: maybe worth calling out this is in seconds

rust/hyperlane-base/src/contract_sync/mailbox.rs
```diff
@@ -98,107 +95,119 @@ where
...
+                .await?
+                .into_iter()
+                .map(|(msg, _)| msg)
+                .collect();
+
+            info!(
+                from,
+                to,
+                distance_from_tip = cursor.distance_from_tip(),
+                estimated_time_to_sync = fmt_duration(eta),
```
tkporter, line 123: a little worried that the presence of this in the logs could lead ppl to believe they're not fully synced, even if the time is saying 0 seconds. Blockchain nodes for instance usually only log this when they're behind, so I could see some ppl bringing that knowledge into here and incorrectly believing there's something wrong and their relayer is unable to fully sync or something.

Just calling out in case you think there may be an easy / nice way to avoid the chance of this confusion, otherwise I'm happy to merge with this log and we can always change if ppl seem confused

tkporter (approved):
This is awesome, excited to see it in action

rust/hyperlane-base/src/contract_sync/eta_calculator.rs
mattiecnvr: This already caps it at `0.9`, `(1.5).min(0.9) = 0.9`

mattiecnvr: switching to use `cmp::min` to make it more clear

rust/hyperlane-base/src/contract_sync/mailbox.rs
mattiecnvr: Changed it to say "synced" for the ETA when it is `0`

rust/hyperlane-base/src/contract_sync/eta_calculator.rs
tkporter: oh.. i'm dumb, for some reason I mixed it up with max

PR #2052, opened 2023-03-31
Do not check StorageGasOracle's owner

asaj:
### Description

We recently started checking the owner for every ownable contract.

This broke CI since StorageGasOracle's owner is not expected to match the configured config owner.

### Drive-by changes

None

### DIFF
--- a/typescript/sdk/src/deploy/HyperlaneAppChecker.ts
+++ b/typescript/sdk/src/deploy/HyperlaneAppChecker.ts
@@ -1,13 +1,14 @@
 import { Contract } from 'ethers';
 import { keccak256 } from 'ethers/lib/utils';
 
+import { Ownable } from '@hyperlane-xyz/core';
 import type { types } from '@hyperlane-xyz/utils';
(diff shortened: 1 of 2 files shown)


jmrossy (approved):

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -120,9 +121,11 @@ export abstract class HyperlaneAppChecker<
     }
   }
 
-  // TODO: Require owner in config if ownables is non-empty
-  async checkOwnership(chain: ChainName, owner: types.Address): Promise<void> {
-    const isOwnable = async (contract: Contract): Promise<boolean> => {
+  async ownables(chain: ChainName): Promise<{ [key: string]: Ownable }> {
+    const isOwnable = async (
```
jmrossy, line 125: Nit: not totally convinced an async lambda + three `obj*` functions is simpler here than a good-ol for loop but not a big deal

PR #2055, opened 2023-04-03
Make getDispatchedMessages static

jmrossy:
- Make getDispatchedMessages static (it's useful in cases where a core isn't needed)
- Clarify types and comments for some message utils
- Move required runtime types to SDK deps


typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -83,7 +83,25 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     return;
   }
 
-  getDispatchedMessages(sourceTx: ethers.ContractReceipt): DispatchedMessage[] {
+  waitForMessageProcessing(
```
jmrossy, line 86: No changes, just moved up

yorhodes (approved):

nambrot (approved):

PR #2063, opened 2023-04-04
Fix fork tests 2, fixier

asaj:
### Description

Follow-up to #2062, fixes an issue where `pick` inserts a key into an object that wasn't originally present. 

This was causing HyperlaneApps to be instantiated with undefined addresses, which caused the fork tests to fail.

Also skips fork testing ICAs/IQS due to the IGP mismatch error


jmrossy (approved):
Good catch!

PR #2071, opened 2023-04-06
Fetch with timeout in Liquidity Layer app

tkporter:
### Description

Noticed that the portal relayer had hung on some async task, causing portal messages to stop being relayed. I'm not confident that fetches were to blame, but based off the logs it seems most likely that one of the fetches went wrong and an error was never thrown. So this adds a 10 second timeout for each fetch

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

Yes

None

### Testing

Ran locally, everything still worked


typescript/sdk/src/utils/fetch.ts
```diff
@@ -0,0 +1,14 @@
+import fetch from 'cross-fetch';
+
+// Adapted from https://github.com/node-fetch/node-fetch#request-cancellation-with-abortsignal
+export async function fetchWithTimeout(
```
jmrossy, line 4: A slightly cleaner version that uses async/await correctly (this one need not use `async`)
https://github.com/hyperlane-xyz/hyperlane-explorer/blob/main/src/utils/timeout.ts#L28

yorhodes: also added this recently :) https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/25dfcbaed7f50675b21f65507a6f84602da4af72/typescript/utils/src/utils.ts#L346

tkporter: Ahh nice. Just changed in favor of yours Rossy :)

Didn't use `runWithTimeout` because I have a feeling the signal triggers some cleanup that I'm not 100% sure would occur if I used runWithTimeout

jmrossy (approved):

PR #2073, opened 2023-04-06
Lag count in Validator agent

### LINKED ISSUE #1764: Validator underflow on PI chain
Not sure how this is happening, but it shouldn't:

https://gist.github.com/roger-that-dev/a1c9b243b122f31fad544bf7db1fc810

asaj:
### Description

This PR adds a lag to `mailbox.count()` in the validator agent.

Without this lag, validators can shut down after the first message because the latestCheckpoint call is lagged and will fail due to the lagged count still being zero.

### Drive-by changes

None

### Related issues

- Fixes #1764

### Backward compatibility

Yes

None

### Testing

None


rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -292,8 +292,22 @@ where
...
+                let tip = self
+                    .provider
+                    .get_block_number()
+                    .await
+                    .map_err(ChainCommunicationError::from_other)?
+                    .as_u64();
+                base_call.block(tip.saturating_sub(lag.get()))
+            }
+            None => base_call,
+        };
```
mattiecnvr, line 308: nit: use `if let` because you don't need to unwrap the `None` case
```suggestion
        let call_with_lag = if let Some(lag) = maybe_lag {
            let tip = self
                .provider
                .get_block_number()
                .await
                .map_err(ChainCommunicationError::from_other)?
                .as_u64();
            base_call.block(tip.saturating_sub(lag.get()))
        } else {
            base_call
        };
```

mattiecnvr (approved):

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -292,8 +292,22 @@ where
...
+        let count = call_with_lag.call().await?;
```
tkporter, line 309: could probably just return here but don't feel strongly at all:
```
call_with_lag.call().await
```

tkporter (approved):

tkporter:
also good catch!

PR #2076, opened 2023-04-07
Fix SDK attachContractsMap bug

asaj:
### Description

attachContractsMap should filter out chains for which a complete set of addresses is not present

### Drive-by changes

None

### DIFF
--- a/typescript/sdk/src/contracts.ts
+++ b/typescript/sdk/src/contracts.ts
@@ -4,7 +4,7 @@ import type { types } from '@hyperlane-xyz/utils';
 
 import { MultiProvider } from './providers/MultiProvider';
 import { ChainMap, Connection } from './types';
-import { objMap, pick } from './utils/objects';
+import { objFilter, objMap, pick } from './utils/objects';
 
 export type HyperlaneFactories = {
   [key: string]: ethers.ContractFactory;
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/contracts.ts
```diff
@@ -53,30 +53,42 @@ function getFactory(
...
 
-export function filterAddresses(
-  addresses: HyperlaneAddresses<any>,
+export function filterAddressesMap(
+  addressesMap: HyperlaneAddressesMap<any>,
   factories: HyperlaneFactories,
-): HyperlaneAddresses<any> {
-  return pick(addresses, Object.keys(factories));
+): HyperlaneAddressesMap<typeof factories> {
+  // Filter out addresses that we do not have factories for
```
nambrot, line 60: I don't have full context here, so you might want somebody else to actually review this. IIUC, this is to allow to pass in a superset of addresses and still have attachContracts return the expected contracts shape. Would it be possible to not do this implicitly, but explicitly, i.e. have the caller specify the filtering explicitly?

nambrot (approved):

typescript/sdk/src/contracts.ts
```diff
@@ -53,30 +53,42 @@ function getFactory(
...
+  const pickedAddressesMap = objMap(addressesMap, (_, addresses) =>
+    pick(addresses, Object.keys(factories)),
```
jmrossy, line 62: Nit, hoist Object.keys(factories) to a var as it's used in several places, including in maps

typescript/sdk/src/contracts.ts
```diff
@@ -53,30 +53,42 @@ function getFactory(
...
+  );
+  // Filter out chains for which we do not have a complete set of addresses
+  return objFilter(
+    pickedAddressesMap,
+    (_, addresses): addresses is HyperlaneAddresses<typeof factories> => {
+      return Object.keys(factories)
+        .map((contract) => Object.keys(addresses).includes(contract))
+        .every(Boolean);
```
jmrossy, line 70: `Object.keys(addresses).every(a => factoriesAddrs.includes(a));`

jmrossy (approved):
See optional comments

typescript/sdk/src/contracts.ts
asaj: Doing this in a future PR

PR #2078, opened 2023-04-07
Support restoration of partial deployments in deployer

asaj:
### Description

This PR allows the deployer to be instantiated with addresses even if not all contracts were deployed to a chain.

This allows e.g. reuse of the ProxyAdmin contract when deploying multiple modules

### Drive-by changes

- Minor improvements to objMap and objMapEntries types, so as to cast less in callers

### DIFF
--- a/typescript/infra/src/deploy.ts
+++ b/typescript/infra/src/deploy.ts
@@ -67,14 +67,14 @@ export async function deployWithArtifacts<Config>(
   },
 ) {
   if (cache.read) {
-    let addresses = {};
+    let addressesMap = {};
     try {
-      addresses = readJSONAtPath(cache.addresses);
+      addressesMap = readJSONAtPath(cache.addresses);
     } catch (e) {
       console.error('Failed to load cached addresses');
     }
 
-    deployer.cacheAddresses(addresses);
+    deployer.cacheAddressesMap(addressesMap);
   }
 
   try {
--- a/typescript/sdk/src/contracts.ts
+++ b/typescript/sdk/src/contracts.ts
@@ -4,7 +4,7 @@ import type { types } from '@hyperlane-xyz/utils';
 
 import { MultiProvider } from './providers/MultiProvider';
 import { ChainMap, Connection } from './types';
-import { objMap, pick } from './utils/objects';
+import { objFilter, objMap, pick } from './utils/objects';
 
 export type HyperlaneFactories = {
   [key: string]: ethers.ContractFactory;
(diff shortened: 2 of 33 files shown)


typescript/infra/src/deploy.ts
```diff
@@ -67,14 +67,14 @@ export async function deployWithArtifacts<Config>(
   },
 ) {
   if (cache.read) {
-    let addresses = {};
+    let addressesMap = {};
```
jmrossy, line 70: If we're renaming for clarity, I find `xToY` is a helpful nomenclature for maps

typescript/sdk/src/contracts.ts
```diff
@@ -38,34 +44,31 @@ export function serializeContractsMap<F extends HyperlaneFactories>(
 export function serializeContracts<F extends HyperlaneFactories>(
   contracts: HyperlaneContracts<F>,
 ): HyperlaneAddresses<F> {
-  return objMap(
-    contracts,
-    (_, contract) => contract.address,
-  ) as HyperlaneAddresses<F>;
+  return objMap(contracts, (_, contract) => contract.address);
```
jmrossy, line 47: Does this fix the casting + ts-ignoring we've had to do around this function? If so, happy to see it!

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -341,15 +330,36 @@ export abstract class HyperlaneDeployer<
...
+      .attach(contractAddress)
+      .connect(this.multiProvider.getSignerOrProvider(chain)) as Awaited<
+      ReturnType<F['deploy']>
+    >;
+    if (hit) {
+      this.logger(
+        `Recovered ${contractName.toString()} on ${chain} ${cachedAddress}`,
+      );
+    }
+    return { hit, contract };
```
jmrossy, line 362: Why is a `hit` value useful instead of just checking if a `contract` value does or does not exist in the cache?

jmrossy (approved):
A few small things but nothing blocking

typescript/infra/src/deploy.ts
asaj: Agreed but prefer thsi for now as it's more consistent with how we've named things elsewhere (`xMap` as implicitly `chainToX`)

typescript/sdk/src/contracts.ts
asaj: I think so!

typescript/sdk/src/deploy/HyperlaneDeployer.ts
asaj: It's not, we can just return contract | undefined

PR #2084, opened 2023-04-10
Clear timeout in runWithTimeout

asaj:
### Description

Fixes a bug in which processes (e.g. e2e test) would hang until the timeout completed

### DIFF
--- a/typescript/utils/src/utils.ts
+++ b/typescript/utils/src/utils.ts
@@ -355,7 +355,7 @@ export async function runWithTimeout<T>(
         timeoutMs,
       )),
   );
-  const ret = Promise.race([callback(), timeoutProm]);
+  const ret = await Promise.race([callback(), timeoutProm]);
   // @ts-ignore timeout gets set immediately by the promise constructor
   clearTimeout(timeout);
   return ret;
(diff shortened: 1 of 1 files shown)


typescript/utils/src/utils.ts
```diff
@@ -347,11 +347,16 @@ export async function runWithTimeout<T>(
...
+  let timeout: NodeJS.Timeout;
+  const timeoutProm = new Promise<void>(
+    (_, reject) =>
+      (timeout = setTimeout(
+        () => reject(new Error(`Timed out in ${timeoutMs}ms.`)),
+        timeoutMs,
+      )),
   );
-  return Promise.race([callback(), timeout]);
+  const ret = Promise.race([callback(), timeoutProm]);
```
yorhodes, line 358: hmm I don't think I understand exactly how this works
I saw @jmrossy has a slightly different implementation... https://github.com/hyperlane-xyz/hyperlane-explorer/blob/1fb99eb0bde189bc1be5b909ef7f9a34bb8f0cf6/src/utils/timeout.ts#L49

yorhodes (approved):

typescript/utils/src/utils.ts
asaj: What we need to be able to do is clear the timeout if the other promise resolves first, I'm not sure his implementation does that.

asaj: You were probably confused b/c I meant to await Promise.race and did not?

PR #2091, opened 2023-04-13
Finality blocks config

### LINKED ISSUE #532: Use `finality_blocks` instead of `reorg_period` in validator
In #504, `finality_blocks` was added for each ChainConfig. The `reorg_period` in the Validator agent settings is basically a duplicate of this-- we should remove `reorg_period` in favor of `finality_blocks` and update the `infra` package accordingly

mattiecnvr:
### Description

Remove redundant `reogperiod` and replace with `finalityBlocks`

### Drive-by changes

### Related issues

- Blocked by #2070
- Fixes #532
- Might want to wait for #2098

### Backward compatibility

Yes
No

None
Yes

### Testing

None
Manual
Unit Tests

### DIFF
--- a/rust/agents/validator/src/settings.rs
+++ b/rust/agents/validator/src/settings.rs
@@ -19,8 +19,6 @@ decl_settings!(Validator,
         validator: SignerConf,
         /// The checkpoint syncer configuration
         checkpoint_syncer: CheckpointSyncerConf,
-        /// The reorg_period in blocks
-        reorg_period: u64,
         /// How frequently to check for new checkpoints
         interval: Duration,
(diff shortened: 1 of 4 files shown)


nambrot:
Just calling out that the deployment and backwards compatability path are important here

nambrot:
@mattiecnvr what's the status here?

mattiecnvr:
@nambrot This was set aside since it was not considered important but I was hoping we could merge it with the other settings changes that you proposed a while ago. At this point it feels like it can probably be closed since I think those other changes will already have it included.

PR #2094, opened 2023-04-14
Remove txsubmission type

mattiecnvr:
### Description

Removes a vestigial organ from the gelato days. We no longer need a config to denote txsubmission type.

### Drive-by changes

No

### Related issues

- https://discord.com/channels/935678348330434570/1096400530722537513

### Backward compatibility

Yes

None

### Testing

Unit Tests


asaj (approved):
Out of curiosity, what happens if someone were to still include TxSubmissionType in their config?

mattiecnvr:
> Out of curiosity, what happens if someone were to still include TxSubmissionType in their config?

In theory it should just be ignored in the same way there are lots of env vars that are not used at all.

tkporter (approved):

PR #2099, opened 2023-04-14
Error message to indicate if there are no validators

### LINKED ISSUE #2031: Relayer should indicate if quorum is impossible
A very common error message now is `unable to reach quorum with validators` which is kind of unactionable. A common issue is if the MultisigISM that is used (either on the mailbox or on the recipient) can actually not ever fetch that because there are no validators set for the origin domain. In that case the relayer should indicate that.

mattiecnvr:
### Description

Adds a more descriptive error message for failing to reach validator quorum when there are no validators defined.

### Drive-by changes

No

### Related issues

- Fixes #2031 

### Backward compatibility

Yes

None

### Testing

None


rust/agents/relayer/src/msg/metadata_builder.rs
```diff
@@ -140,10 +142,17 @@ impl MetadataBuilder for LegacyMultisigIsmMetadataBuilder {
...
             .await.context(CTX)?
         else {
-            info!(
-                ?validators, threshold, highest_known_nonce,
-                "Could not fetch metadata: Unable to reach quorum"
-            );
+            if validators.is_empty() {
+                info!(
+                    ism=%multisig_ism.address(),
+                    "Could not fetch metadata: Unable to reach quorum because no validators are defined"
```
asaj, line 148: Rather than defined, perhaps "no validator set for chain X is configured on the recipients ISM"

asaj (approved):

PR #2110, opened 2023-04-17
Hyperlane-Core WASM support

### LINKED ISSUE #1977: Pull core rust types into their own crate with WASM-compatible dependencies
We want to be able to use some core types (like `HyperlaneMessage`, the Encode & Decode traits, etc) in WASM. This is because the Fuel indexer logic is WASM based (see https://github.com/hyperlane-xyz/fuel-contracts/pull/39)

At the moment, `HyperlaneMessage` lives in `hyperlane-core`, which has a few dependencies that are not able to be compiled to WASM. These dependencies include ethers, I think tokio, and I'm sure many others. So for now in the Fuel indexer, we're just copying the Encode & Decode traits (https://github.com/hyperlane-xyz/fuel-contracts/pull/39/files#diff-9d5375e58ae555a71d183e064e25ded61d9e0e6da09c419a5243b312750f9a22), and the HyperlaneMessage type (https://github.com/hyperlane-xyz/fuel-contracts/pull/39/files#diff-c1df921b29777e810d4612bf6e315c71cf5f404d83cf0413cdeea3b4d2134ace). We'll soon be wanting access to the `InterchainGasPayment` struct too.

Ideally we shouldn't be copying code just to get it to work in WASM. Instead we should rip out some of the really basic types from `hyperlane-core` into something like a `hyperlane-types` crate that can compile to WASM, and then use that.

We should make sure that this crate is also able to be used in Sealevel / Solana for this PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1345

mattiecnvr:
### Description

This shuffles some code around and changes dependencies to make `hyperlane-core` WASM compatible and adds a check to the pipeline to ensure it stays that way.

To do this it does refactor hyperlane db to the `base` crate, which makes more sense IMO anyway as that is an implementation detail of the agents and not part of the core protocol.

### Related issues

- Fixes #1977

### Backward compatibility

Yes-ish - It does change where things are but it does not change support. Should only impact people who directly use the libraries.

None

### Testing

Manual
Unit Tests


rust/hyperlane-core/src/error.rs
```diff
@@ -70,9 +71,6 @@ pub enum ChainCommunicationError {
     /// A transaction was dropped from the mempool
     #[error("Transaction dropped from mempool {0:?}")]
     TransactionDropped(H256),
-    /// DB Error
-    #[error(transparent)]
-    DbError(#[from] DbError),
```
tkporter, line 75: weird this was here to begin with, good catch

.github/workflows/rust.yml
```diff
@@ -62,6 +63,8 @@ jobs:
             ./rust
       - name: Check
         run: cargo check --all-features --all-targets
+      - name: Check WASM
```
tkporter, line 66: 🤤

tkporter (approved):
nice, agree db feels more natural in base

PR #2125, opened 2023-04-19
Minor fixes to get LL deploys running smoothly

nambrot:
### Description

This PR makes some minor modifications to the deployment structure for the Liquidity Layer. The most important thing is the embedding of the `proxyAdmin` addresses in the LL addresses file which is not pretty, but I did not want to put it in the SDK right now either, but could be convinced.

In the process of running this, the LL router IGPs have been updated which is non-material, but otherwise the deploy script runs as a no-op now.


typescript/infra/config/environments/testnet3/middleware/liquidity-layer/addresses.json
```diff
@@ -2,23 +2,28 @@
   "goerli": {
     "circleBridgeAdapter": "0xfe9d88aA85c5917822C804b949BcEDE832C02ce2",
     "portalAdapter": "0x68D753982e89CC083917863F6dc9738448B91ef9",
-    "router": "0x2abe0860D81FB4242C748132bD69D125D88eaE26"
+    "proxyAdmin": "0x8f919348F9C4619A196Acb5e377f49E5E2C0B569",
```
jmrossy, line 5: I share your intuition that this is an odd place for this but not strongly opposed to punting on moving it to SDK

typescript/infra/scripts/middleware/circle-relayer.ts
```diff
@@ -27,14 +32,18 @@ async function check() {
...
     multiProvider,
-    bridgeAdapterConfigs,
+    config.liquidityLayerConfig,
   );
 
   while (true) {
-    for (const chain of [Chains.goerli, Chains.fuji]) {
+    for (const chain of Object.keys(
+      // @ts-ignore Can't figure how to make it a type predicate
+      objFilter(config.liquidityLayerConfig, (_, config) => !!config.circle),
```
jmrossy, line 45: What's the error? You may be missing a `: config is LiquidityLayerConfig` type annotation on the filter lambda func

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
```diff
@@ -93,18 +93,26 @@ export class LiquidityLayerDeployer extends ProxiedRouterDeployer<
...
-      objMap(contractsMap, (_, contracts) => ({
-        liquidityLayerRouter: contracts.circleBridgeAdapter,
-      })) as unknown as HyperlaneContractsMap<LiquidityLayerFactories>,
+      objMap(
+        // @ts-ignore Not sure how to specify the type guard
+        objFilter(contractsMap, (_, c) => !!c.circleBridgeAdapter),
+        (_, contracts) => ({
+          liquidityLayerRouter: contracts.circleBridgeAdapter,
+        }),
+      ) as unknown as HyperlaneContractsMap<LiquidityLayerFactories>,
```
jmrossy, line 102: Would be great to improve the types here before merging

typescript/sdk/src/router/ProxiedRouterDeployer.ts
```diff
@@ -50,13 +50,18 @@ export abstract class ProxiedRouterDeployer<
...
-    await super.runIfOwner(chain, proxyAdmin, () =>
-      this.multiProvider.handleTx(
-        chain,
-        proxyAdmin.transferOwnership(config.owner),
-      ),
-    );
+    await super.runIfOwner(chain, proxyAdmin, async () => {
+      this.logger(`Checking ownership of proxy admin to ${config.owner}`);
+
+      if ((await proxyAdmin.owner()) !== config.owner) {
```
jmrossy, line 56: Use `utils.eqAddress` here

typescript/infra/config/environments/testnet3/middleware/liquidity-layer/addresses.json
asaj: If we're shipping the LL, why not put those addresses in the SDK?

typescript/infra/src/config/environment.ts
```diff
@@ -44,6 +45,7 @@ export type EnvironmentConfig = {
   ) => Promise<MultiProvider>;
   helloWorld?: Partial<Record<Contexts, HelloWorldConfig>>;
   keyFunderConfig?: KeyFunderConfig;
+  liquidityLayerConfig?: ChainMap<BridgeAdapterConfig>;
```
asaj, line 48: Consider merging the liquidityRelayerConfig and the liquidityLayerConfig, so that we can enforce that if there is a liquidityLayerConfig there *must* be a liquidityRelayer config.

e.g.

```
export type LiquidityLayerEnvConfig = {
  relayer: LiquidityLayerRelayerConfig,
  bridgeAdapters: ChainMap<BridgeAdapterConfig>
}

export type EnvironmentConfig = {
...
  liquidityLayer?: LiquidityLayerEnvConfig
}
```

typescript/sdk/src/router/ProxiedRouterDeployer.ts
asaj, line 56: In what world isn't this the case? When deployer === config.owner?

asaj (commented):
> The most important thing is the embedding of the proxyAdmin addresses in the LL addresses file which is not pretty, but I did not want to put it in the SDK right now either, but could be convinced.

Wdym? The proxyAdmin addresses are already in the SDK.

typescript/infra/config/environments/testnet3/middleware/liquidity-layer/addresses.json
nambrot: The way the artifacts work would imo require to put in the adapter contract addresses as well at the top level which feels out of place in the SDK artifacts. Definitely not opposed to moving it in, it just felt out of place there for me too.

typescript/infra/scripts/middleware/circle-relayer.ts
nambrot: That did the trick!

typescript/infra/config/environments/testnet3/index.ts
```diff
@@ -39,6 +40,7 @@ export const environment: EnvironmentConfig = {
   helloWorld,
   owners,
   keyFunderConfig,
+  liquidityLayerConfig: bridgeAdapterConfigs,
```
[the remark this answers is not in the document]
nambrot: Incorporated your suggestion below

typescript/infra/src/config/environment.ts
nambrot: Great suggestion

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
```diff
@@ -93,18 +93,26 @@ export class LiquidityLayerDeployer extends ProxiedRouterDeployer<
...
-      objMap(contractsMap, (_, contracts) => ({
-        liquidityLayerRouter: contracts.portalAdapter,
-      })) as unknown as HyperlaneContractsMap<LiquidityLayerFactories>,
+      objMap(
+        // @ts-ignore Not sure how to specify the type guard
+        objFilter(contractsMap, (_, c) => !!c.portalAdapter),
+        (_, contracts) => ({
+          liquidityLayerRouter: contracts.portalAdapter,
+        }),
+      ) as unknown as HyperlaneContractsMap<LiquidityLayerFactories>,
```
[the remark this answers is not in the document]
nambrot: Maybe i'm not understanding the suggestion, but IIUC, these type gymnastics are necessary because of the `enrollRemoteRouters` function signature which expects a `HyperlaneContractsMap` parameterized by the `Factories` type of the deployer, so we have to convince the compiler that the generated map here "matches" that type?

typescript/sdk/src/router/ProxiedRouterDeployer.ts
nambrot: Yes indeed. On testnet, the deployer key is the owner. In those cases, we would just call transferOwnership despite the owner already being said correctly. (i.e. IQS and LL on testnet3 right now)

nambrot:
> > The most important thing is the embedding of the proxyAdmin addresses in the LL addresses file which is not pretty, but I did not want to put it in the SDK right now either, but could be convinced.
> 
> Wdym? The proxyAdmin addresses are already in the SDK.

Probably shouldn't have called it most important, but "controversial" maybe. Basically, the LL addresses are not currently in the SDK, which means infra's deploy logic won't read `proxyAdmin` from the `addresses` file (which prior to this PR contained only the LL router + adapter addresses).

Like mentioned in a comment reply, I was considering putting it in the SDK, but would also need to put the adapter addresses in there which imo would look funny.

I don't feel strongly either way, and imo its easily changeable so would suggest to just keep it this way

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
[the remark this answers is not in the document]
asaj: Yes but objFilter can do that for you

typescript/infra/config/environments/testnet3/index.ts
```diff
@@ -39,6 +40,9 @@ export const environment: EnvironmentConfig = {
   helloWorld,
   owners,
   keyFunderConfig,
-  liquidityLayerRelayerConfig,
+  liquidityLayerConfig: {
+    bridgeAdapter: bridgeAdapterConfigs,
```
asaj, line 44: bridgeAdapters

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
```diff
@@ -93,18 +93,32 @@ export class LiquidityLayerDeployer extends ProxiedRouterDeployer<
...
+      objMap(
+        objFilter(
+          contractsMap,
+          (_, c): c is HyperlaneContracts<LiquidityLayerFactories> =>
+            !!c.circleBridgeAdapter,
+        ),
+        (_, contracts) => ({
+          liquidityLayerRouter: contracts.circleBridgeAdapter,
+        }),
+      ) as unknown as HyperlaneContractsMap<LiquidityLayerFactories>,
```
asaj, line 105: I don't think you need this anymore

asaj (approved):

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
```diff
@@ -115,6 +129,9 @@ export class LiquidityLayerDeployer extends ProxiedRouterDeployer<
     chain: ChainName,
     config: LiquidityLayerConfig,
   ): Promise<HyperlaneContracts<LiquidityLayerFactories>> {
+    // This is just the temp owner for contracts, and HyperlaneRouterDeployer#transferOwnership actually sets the configured owner
+    const tempOwner = await this.multiProvider.getSignerAddress(chain);
```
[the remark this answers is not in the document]
asaj: Test

asaj (approved):

typescript/infra/config/environments/testnet3/middleware/liquidity-layer/addresses.json
nambrot: IIRC, not without changing the infra/scripts/deploy.ts to account for this edge case which did not seem preferable

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerRouterDeployer.ts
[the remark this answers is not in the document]
nambrot: Chatted offline to just punt since this code will get removed soon

PR #2128, opened 2023-04-19
Use proper ISM in ICA contracts

asaj:
### Description

Currently, ICAs are configured without an ISM, which results in them falling back to the mailbox default.

Instead, they should delegate security to whichever ISM was specified in the message.

This PR updates the deployer to deploy the proper ISM by default, and the checker/governor with code that will allow the existing deployments to move to that ISM.

### Drive-by changes

- My formatter seems to be making changes to src/sdk/index.ts

### Backward compatibility

Yes

None

### Testing

None


asaj:
ICA fork testing is expected to fail since the current ICA deployment is now considered invalid

jmrossy (approved):

Issue #2113, opened 2023-04-18
Agent docs/config improvements after EthTokyo


nambrot:
Problems:
            - A lot of configuration fatigue
            - missing field `validator` is too vague
            - No such file or directory (os error 2) is a frustrating error
            - Machine specs section on starting a validator looks out of place
            - `Error: Destination chain base does not have a configured signer` feels not as actionable as we can be
            - `Missing chain configuration for sepolia; this includes protocol type and the connection information` is quite vague
            - in the relayer guide, similar to the validator guide, the specific sequence is a bit unclear. Going page by page gets me to message filtering which i should not need to check out. The Env page does again not distinguish between required and optional vars and does not tell me what the next step is, when i am done
            - Way too easy to miss `HYP_RELAYER_ALLOWLOCALCHECKPOINTSYNCERS`
            - Not mounting the DB_PATH will result in relayers needing to index from genesis every time which is very annoying
            - When running two relayers locally, it is possible to actually have both read/write to the same DB which is very hard to debug and obviously does not work
            -  "no name found" - typo results in weird message
            - When restarting anvil, previous artifacts are no longer valid, but very hard to debug

Possible Solutions (not exhaustive):
            - Collapse running validator/relayer into a single page as much as possible
            - Default validator interval to something like 5s
            - Default HYP_BASE_METRICS (maybe already done)
            - Default HYP_BASE_TRACING_LEVEL and FORMAT (maybe already done)
            - Segment the environment variables/config into required and optional ones (i.e. ones that have reasonable defaults should be further down)
            - “missing field `validator`” should have a more useful message indicate to specify the validator key
            - When using localStorage checkpoint synced, just create the directory if it doesn’t exist
            - When signers are missing, point to default signer as much as possible
            - If an RPC url is missing, call it out as such, don’t say chain configuration is missing with protocol type and connection information
            - HYP_RELAYER_ALLOWLOCALCHECKPOINTSYNCERS needs to be made more obvious. When it’s not enabled but we discover a local storage checkpoint synced, we should log an info log line to indicate that this is the issue why metadata could not be fetched
            - Indicate that relayers when using docker should mount external storage for DB_PATH
            - Detect when the agent runs on a DB for which it was not configured (i.e. its the DB of some other agents)
            - "no name found" - should have a more clear error message that they probably have a typo in the RPC url env var

mattiecnvr:
[...]

PR #2132, opened 2023-04-20
Agent Argument Parsing

### LINKED ISSUE #2034: Allow command-line args for agent configuration
Many folks who have been trying to run the agents find it unintuitive to specify ENV vars. It seems more natural to folks to configure via command-line args. Many node agents like geth do it that way.

While the config library we use does not directly support command-line arguments, the `Source` trait seems flexible enough to allow a thin wrapper over an existing CMD-line argument library to be provided a map of key-value-pairs to loading to the config library. This would seem like an easy enough way to get command line argument support.

Relevant Discord threads:
- https://discord.com/channels/935678348330434570/1077336620941987951/1077529971242840117

mattiecnvr:
### Description

All configs can now be set by arguments as well as env and config files. The arguments will take the highest priority and overwrite the others.

We should also update the docs to include this new support when it gets released, but since it is an entirely new feature it also can be a stealth release which we document after the fact.

### Related issues

- Fixes #2034 

### Backward compatibility

Yes

None

### Testing

Manual
Unit Tests


rust/hyperlane-base/src/settings/loader/mod.rs
```diff
@@ -72,10 +75,11 @@ where
                 .separator("_")
                 .source(Some(filtered_env)),
         )
+        .add_source(CommandLineArguments::default().separator("."))
```
mattiecnvr, line 78: LMK if you think we should use something else such as `-`.

This currently looks like `--chains.ethereum.conenction.url="https://example.com"` but we could also do something like `--chains-ethereum-conenction-url="https://example.com"`.

rust/hyperlane-core/src/config/config_path.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Get the expected command line argument name
+    pub fn arg_name(&self) -> String {
+        let name = self
+            .0
+            .iter()
+            .map(|s| s.as_str().to_case(Case::Camel))
+            .join(".");
+        format!("--{name}")
```
mattiecnvr, line 79: Something to note here is that we might encourage people to use camel cased names for readability but ultimately they won't be displayed here.

rust/utils/run-locally/src/main.rs
```diff
@@ -142,23 +142,29 @@ fn main() -> ExitCode {
     };
 
     let relayer_env = hashmap! {
```
mattiecnvr, line 144: No net changes to the config, just trying to make sure the args correctly are parsed and override the env.

rust/hyperlane-core/src/config/config_path.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn add(self, rhs: ConfigPath) -> Self::Output {
+        self.merge(&rhs)
+    }
+}
+
+impl ConfigPath {
+    /// Add a new part to the path.
+    pub fn join(&self, part: impl Into<String>) -> Self {
+        let part = part.into();
+        debug_assert!(!part.contains('.'));
```
tkporter, line 38: wondering what situation this is supposed to catch and if it should be something we want in the optimized build?

(i know this is just a moved bit of code that already existed)

tkporter (approved):
awesome! lgtm

rust/hyperlane-core/src/config/config_path.rs
mattiecnvr: It catches a case that isn't going to break anything, but that would be nice to catch during debug. Basically we should not have the path include "." but have different segments instead.

nambrot:
@mattiecnvr Do you want me to add another ticket for the docs update or you want to do that in the context of #2113 maybe?

PR #2134, opened 2023-04-20
Better checks on config files

### LINKED ISSUE #2104: Agent errors when config files not provided or path incorrect
Not sure which here, but this error was cryptic:
https://discord.com/channels/935678348330434570/1096444620172431410/1096709297678860298

mattiecnvr:
### Description

This adds some additional validation to the config files that are provided by users to make sure that they are valid before trying to load them.

### Drive-by changes

None

### Related issues

- Fixes #2104

### Backward compatibility

Yes - probably, might be an edge case it drops support for like if someone was using a config file without a `.json` extension

None

### Testing

Manual


tkporter (approved):
nice

PR #2136, opened 2023-04-21
Add static fromAddressesMap constructor to apps

asaj:
### Description

Adds a third constructor. Allows apps to be constructed from:
- Environment name
- ContractsMap
- AddressesMap
 
### Drive-by changes

None


yorhodes (approved):

typescript/sdk/src/HyperlaneApp.ts
```diff
@@ -26,11 +26,22 @@ export class HyperlaneApp<
...
-  static fromAddressesMap<F extends HyperlaneFactories>(
+  getContracts(chain: ChainName): HyperlaneContracts<Factories> {
+    return this.get(chain);
+  }
+
+  getAddresses(chain: ChainName): HyperlaneAddresses<Factories> {
+    return serializeContracts(this.get(chain));
+  }
+
+  static fromAddressesMapHelper<F extends HyperlaneFactories>(
```
jmrossy, line 37: Why not just keep this named `fromAddressesMap` and have subclasses call `super`? The new name seems a bit odd for for a public method

yorhodes: iirc `super.static` is invalid

yorhodes: but agree avoiding rename is nicer
I presume this is not a widely used interface but appreciate the sentiment that we should avoid breaking changtes

jmrossy: Good point but they could access it via the class' name directly

asaj: I don't think that works, as super will return a HyperlaneApp instance (as opposed to the subclass).

Know any ways around this?

jmrossy: Sorry I wasn't clear. I mean access the `HyperlaneApp` static directly like this: 
```
  static fromAddressesMap(
    addressesMap: HyperlaneAddressesMap<any>,
    multiProvider: MultiProvider,
  ): HyperlaneIsmFactory {
    const fromAddressesMap = HyperlaneApp.fromAddressesMap(
      addressesMap,
      ismFactoryFactories,
      multiProvider,
    );
```

asaj: No you were clear, I was replying to the original comment as the others hadn't loaded for me

jmrossy: Okay tbh I should have just looked at this closer before suggesting a solution but on second look here's what I'm really trying to say:
There's no reason for `fromAddressesMapHelper` to be in HyperlaneApp and I recommend just making it a util function somewhere

asaj: Works for me

PR #2139, opened 2023-04-21
Better validation of signers during config parsing

### LINKED ISSUE #2101: Config parsing should handle check for configured signers
We are seeing people get

```
Error: Destination chain mumbai does not have a configured signer
```

which is much more confusing than if it were checked for during config parsing and included the correct config path and env var. We could alternatively add a context now to link to the reference for this error.

https://discord.com/channels/935678348330434570/1096444620172431410/1096823458492600340

mattiecnvr:
### Description

This improves the validation during config parsing to ensure that all the expected signers are configured which means the errors will have the correct config path.

### Drive-by changes

Also updated the checks for missing chain configurations to be more clear and use the chain config path rather than the config that indicated the config was required and removes a now redundant check.

### Related issues

- Fixes #2101

### Backward compatibility

Yes

None

### Testing

Manual


rust/agents/relayer/src/relayer.rs
```diff
@@ -65,14 +65,6 @@ impl BaseAgent for Relayer {
         let core = settings.build_hyperlane_core(metrics.clone());
         let db = DB::from_path(&settings.db)?;
 
-        for dest_chain in &settings.destination_chains {
```
mattiecnvr, line 68: This check is now handled during config parsing and is redundant.

tkporter (approved):

PR #2142, opened 2023-04-24
Do not cancel in-flight CI jobs

asaj:
### Description

Cancelled CI jobs cause ❌ when we want ✅

### DIFF
--- a/.github/workflows/monorepo-docker.yml
+++ b/.github/workflows/monorepo-docker.yml
@@ -1,5 +1,5 @@
 name: Build and Push Monorepo Image to GCR
-on: 
+on:
   push:
     branches: [main]
     tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/monorepo-docker.yml
```diff
@@ -10,7 +10,6 @@ on:
       - 'typescript/infra/**'
 concurrency:
   group: build-push-monorepo-${{ github.ref }}
-  cancel-in-progress: true
```
nambrot, line 13: I thought that was actually desirable in the cases where we keep pushing commits on a PR, we don't really need to comlpete jobs of previous commits or is this a different situation?

nambrot (approved):

Issue #1899, opened 2023-03-06
Omniscient Relayer


nambrot:

nambrot:
A possible solution is what we call "Omniscient Relayer", i.e. a single instance of the relayer can relay from/to all configured chains.

nambrot:
One thing we should be mindful of is that operators may not want to relay between all possible pairs. I think its ok to specify this on-chain in the IGP/oracles, but just worth noting here as well

nambrot:
@mattiecnvr I know we had talked about doing this in a backwards compatible way so that we could roll this out without needing any infra changes, though I do think we should move to this setup for dogfooding. As you are working on this, would you mind scoping the follow up work of what it would take for us to move to a single relayer instance?

PR #2148, opened 2023-04-28
Agent docs and config improvements

### LINKED ISSUE #2113: Agent docs/config improvements after EthTokyo

mattiecnvr:
### Description

Bunch of minor tweaks and updates based on what we learned from EthTokyo.

- Updated Rust readme and the root readme to align with the hyperlane docs
- Defaults to a 5 sec validator interval now
- Makes the metrics port default to `9090` and have a soft failure if it can't start
- Now will automatically create the checkpointsyncer directory if it is missing
- Hyperlane database now officially uses the domain name and id as a prefix internally to support multiple relayers using the same database
- 

### Drive-by changes

Refactored and cleaned up the database code. Removed a lot of duplicate/unnecessary/redundant functions.

### Related issues

- Fixes #2113 

### Backward compatibility

Yes

Yes - This changes the database prefixes internally to use domain name and domain id; also prepares us for #1899 

### Testing

Manual
Unit Tests


rust/README.md
```diff
@@ -5,7 +5,8 @@
 - install `rustup`
   - [link here](https://rustup.rs/)
 
-Note: You should be running >= version `1.68.0` of the rustc compiler, you can see that version with this command and should see similar output:
+Note: You should be running >= version `1.68.0` of the rustc compiler, you can see that version with this command and
```
asaj, line 8: Update the output to match the version requirement

rust/README.md
```diff
@@ -14,6 +15,71 @@ info: This is the version for the rustup toolchain manager, not the rustc compil
...
+production environment.
+
+```bash
+cd rust
+./build.sh <image_tag>
+```
+
+### Deploy Procedure
+
+The contract addresses of each deploy can be found in `rust/config`. The latest
```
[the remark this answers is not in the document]
asaj: Actually this whole paragraph looks wrong

rust/hyperlane-base/src/db/hyperlane_db.rs
```diff
@@ -5,15 +5,18 @@ use tokio::time::sleep;
...
+    HyperlaneDomain, HyperlaneMessage, InterchainGasExpenditure, InterchainGasPayment,
+    InterchainGasPaymentMeta, LogMeta, H256, U256,
 };
 
 use super::{
     storage_types::{InterchainGasExpenditureData, InterchainGasPaymentData},
     DbError, TypedDB, DB,
 };
 
+// these keys MUST not be given multiple uses in case multiple agents are
```
asaj, line 17: I'm not sure I follow this comment

rust/hyperlane-base/src/db/mod.rs
```diff
@@ -95,83 +95,17 @@ impl DB {
...
         Ok(self.0.put(key, value)?)
     }
 
     /// Retrieve a value from the DB
-    fn _retrieve(&self, key: impl AsRef<[u8]>) -> Result<Option<Vec<u8>>> {
+    pub fn retrieve(&self, key: &[u8]) -> Result<Option<Vec<u8>>> {
         Ok(self.0.get(key)?)
     }
 
-    /// Prefix a key and store in the DB
```
asaj, line 107: Unused, I presume?

rust/hyperlane-base/src/db/mod.rs
```diff
@@ -95,83 +95,17 @@ impl DB {
...
-        &self,
-        prefix: impl AsRef<[u8]>,
-        key: &K,
-    ) -> Result<Option<V>> {
-        self.retrieve_decodable(prefix, key.to_vec())
-    }
-
     /// Get prefix db iterator for `prefix`
-    pub fn prefix_iterator(&self, prefix: impl AsRef<[u8]>) -> DBIterator {
+    pub fn prefix_iterator(&self, prefix: &[u8]) -> DBIterator {
```
asaj, line 108: What is this used for?

rust/hyperlane-base/src/db/typed_db.rs
```diff
@@ -1,17 +1,15 @@
...
-use hyperlane_core::{Decode, Encode};
+use hyperlane_core::{Decode, Encode, HyperlaneDomain};
 
 use crate::db::{DbError, DB};
 
 type Result<T> = std::result::Result<T, DbError>;
 
 /// DB handle for storing data tied to a specific type/entity.
 ///
-/// Key structure: ```<type_prefix>_<additional_prefix(es)>_<key>```
```
asaj, line 11: What was the previous "type_prefix", and why is it okay to remove?

rust/hyperlane-base/src/db/typed_db.rs
```diff
@@ -22,12 +20,27 @@ impl AsRef<DB> for TypedDB {
...
+            .iter()
+            .chain(b"_")
+            .chain(&domain.id().to_be_bytes())
+            .chain(b"_")
+            .copied()
+            .collect();
+        Self { domain_prefix, db }
+    }
+
+    fn full_key(&self, prefix: &[u8], key: &[u8]) -> Vec<u8> {
```
asaj, line 37: nit: prefixed_key?

rust/hyperlane-base/src/db/typed_db.rs
```diff
@@ -57,8 +76,7 @@ impl TypedDB {
         key: &K,
         value: &V,
     ) -> Result<()> {
-        self.db
-            .store_keyed_encodable(self.full_prefix(prefix), key, value)
+        self.store_encodable(prefix, key.to_vec(), value)
```
asaj, line 79: Should this function be public? Feels like a potential footgun to have public functions that don't handle the key prefixing

asaj: Especially since full_key is private rn

rust/hyperlane-base/src/settings/base.rs
```diff
@@ -49,7 +49,7 @@ pub struct Settings {
     /// Configuration for contracts on each chain
     pub chains: HashMap<String, ChainConf>,
     /// Port to listen for prometheus scrape requests
-    pub metrics: Option<u16>,
+    pub metrics: u16,
```
asaj, line 52: nit: If it's easy, suggest metrics_port

rust/hyperlane-base/src/db/typed_db.rs
mattiecnvr: It was poorly named, it was only being used as the domain name. This is not a change to logic, just clarity.

rust/hyperlane-base/src/settings/base.rs
mattiecnvr: This one is easy, changing the name in the actual config is not.

rust/README.md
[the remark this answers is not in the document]
mattiecnvr: Fixed.

rust/hyperlane-base/src/db/typed_db.rs
mattiecnvr: Changed.

rust/hyperlane-base/src/db/typed_db.rs
mattiecnvr: This function requires the prefix as an argument. I don't understand your comment.

rust/hyperlane-base/src/db/hyperlane_db.rs
asaj: I still don't follow, can we follow up on discord?

rust/hyperlane-base/src/db/mod.rs
asaj: And what do we use that for?

rust/hyperlane-base/src/db/typed_db.rs
asaj: Tbh coming back to this I don't think I do either. My recollection was that some public functions took a prefixed key, and others did not, but it looks like every public fn takes a prefix and a key

asaj (approved):

PR #2157, opened 2023-05-01
Use SSTORE gas spent for delivered check

yorhodes:
### Description

Probably silly gas optimization h/t @asaj 

### Backward compatibility

No

### Testing

Unit Tests


asaj (commented):
Can't say I condone this :)

What are the gas savings?

PR #2160, opened 2023-05-02
Relayer warns if validator not announced

### LINKED ISSUE #2120: Manual validator announce is cumbersome
Problem:
            - Validator announce is easy to miss
            - Validator announce is annoying to do manual with cast (and the only reason why you still need foundry)
Possible Solution:
            - Hyperlane-deploy should have a command that allows the announcement of a validator by just pointing to the announcement file
            - Relayer should indicate if the reason for quorum is that the validator(s) have not announced themselves

asaj:
### Description

Explicitly call out if an ISM is configured with a validator that has not been announced.

### Related issues

- Fixes #2120


rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -207,11 +207,15 @@ impl BaseMetadataBuilder {
                 }
             }
             if checkpoint_syncers.get(&validator.into()).is_none() {
-                warn!(
-                    ?validator,
-                    ?validator_storage_locations,
-                    "No valid checkpoint syncer configs for validator"
-                );
+                if validator_storage_locations.len() > 0 {
```
mattiecnvr, line 210: nit: Use `is_empty` and flip flop so the negative case is the else

mattiecnvr (approved):

PR #2164, opened 2023-05-03
Verify proxy even if implementation is verified

nambrot:
### Description

`ContractVerifier` would skip the proxy verification if the implementation was already verified. This changes it so that we will still attempt proxy verification.


yorhodes (approved):
nice
were there some unverified proxies?

PR #2166, opened 2023-05-03
Deploy LL + CircleBridgeAdapter to mainnet

nambrot:
### Description

Deploy LL + CircleBridgeAdapter to mainnet with the relevant changes and artifacts


typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
```diff
@@ -141,7 +146,11 @@ export class LiquidityLayerApp extends HyperlaneApp<
           try {
             return [CircleBridgeAdapterInterface.parseLog(_)];
           } catch {
-            return [];
+            try {
+              return [MailboxInterface.parseLog(_)];
```
jmrossy, line 150: Nit: underscore is a bad var name because it's typically used to imply the var isn't used

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
```diff
@@ -253,10 +266,13 @@ export class LiquidityLayerApp extends HyperlaneApp<
       return;
     }
 
+    console.log(`Attempt Circle message delivery`, JSON.stringify(message));
```
jmrossy, line 269: Think I've mentioned it before but it's strange to see console logging here but `debug` lib everywhere else in the SDK

jmrossy (approved):
Approving but disabling auto-merge in case you want to address comments

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
nambrot: I'll do this since this is not a pattern in TS, but _ are often used as the single function arg that takes the least visual space
https://stackoverflow.com/questions/10538924/c-sharp-style-lambdas-or-x
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1717
https://www.baeldung.com/scala/underscore#:~:text=The%20underscore%20can%20be%20used,used%20anywhere%20in%20the%20code.

typescript/sdk/src/middleware/liquidity-layer/LiquidityLayerApp.ts
nambrot: Noted, but this is effectively deprecated so lets keep as is

PR #2173, opened 2023-05-04
Make merkle proofs optional on multisig ISM

yorhodes:
### Description

Validators currently sign `(root, index)` checkpoints and during verification, a `message` is passed as calldata, an `id()` is derived, and a `proof` of `id()` at `index` in `root` is verified

This provides “all or nothing” censorship resistance guarantees because a validator can only sign roots to allow any contained messages to be processed.

We have considered alternatives where validators sign `message` directly and we lose censorship resistance in exchange for eliminating merkle proof verification gas costs.

However, if validators sign `(root, index, message)` tuples, we can skip merkle proof verification on the destination chain while still maintaining censorship resistance by providing two valid metadata formats:

1. existing validator signatures and merkle proof verification of inclusion
2. including merkle proof verification for pathway where validators are censoring `message`  

It’s worth noting the validator is required to index event data to produce this new signature format. However, this does not require historical indexing and new validators being spun up can simply begin indexing from tip.

See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2187 for validator changes
See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2248 for relayer and e2e test changes

### Drive-by changes

Merkle index also optional

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2192

### Backward compatibility

- new ISM deployment is necessary (we could upgrade implementation in theory)
- Validator and relayer upgrades

### Testing

Unit (fuzz) Tests


solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -48,26 +50,43 @@ library MultisigIsmMetadata {
         pure
         returns (bytes32)
     {
-        return bytes32(_metadata[ORIGIN_MAILBOX_OFFSET:MERKLE_PROOF_OFFSET]);
+        return bytes32(_metadata[ORIGIN_MAILBOX_OFFSET:MESSAGE_ID_OFFSET]);
+    }
+
+    function messageId(bytes calldata _metadata)
```
asaj, line 56: Should this go at the end? It's really only needed if we go the merkle proof route

solidity/contracts/libs/CheckpointLib.sol
```diff
@@ -24,7 +25,8 @@ library CheckpointLib {
                     abi.encodePacked(
                         _domainHash,
                         _checkpointRoot,
-                        _checkpointIndex
+                        _checkpointIndex,
```
asaj, line 28: Is it safe to use the same domain separator? I think so, but want to double check...

solidity/contracts/libs/LegacyCheckpointLib.sol
```diff
@@ -0,0 +1,54 @@
...
+     * @param _originMailbox The address of the origin mailbox as bytes32.
+     * @return The digest of the checkpoint.
+     */
+    function digest(
+        uint32 _origin,
+        bytes32 _originMailbox,
+        bytes32 _checkpointRoot,
+        uint32 _checkpointIndex
+    ) internal pure returns (bytes32) {
+        bytes32 _domainHash = domainHash(_origin, _originMailbox);
```
asaj, line 20: Can we use CheckpointLib.domainHash here to dedupe?

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -88,18 +107,16 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+        address[] memory _validators,
+        uint8 _threshold
+    ) internal pure returns (bool) {
         bytes32 _digest = CheckpointLib.digest(
             Message.origin(_message),
             MultisigIsmMetadata.originMailbox(_metadata),
             MultisigIsmMetadata.root(_metadata),
-            MultisigIsmMetadata.index(_metadata)
+            MultisigIsmMetadata.index(_metadata),
+            MultisigIsmMetadata.messageId(_metadata)
```
asaj, line 119: How are we enforcing that the message ID that was signed matches the message that is being verified?

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -6,16 +6,18 @@ pragma solidity >=0.8.0;
...
+ * [  90:????] Validator signatures, 65 bytes each, length == Threshold
+ * [????:????] (Optional) Merkle proof
  */
 library MultisigIsmMetadata {
     uint256 private constant MERKLE_ROOT_OFFSET = 0;
     uint256 private constant MERKLE_INDEX_OFFSET = 32;
     uint256 private constant ORIGIN_MAILBOX_OFFSET = 36;
-    uint256 private constant MERKLE_PROOF_OFFSET = 68;
-    uint256 private constant SIGNATURES_OFFSET = 1092;
+    uint256 private constant MESSAGE_ID_OFFSET = 68;
```
asaj, line 17: Dumb question by why do we have the root in the metadata in the first place? Shouldn't we be deriving it from the proof?

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
yorhodes: its also needed for signature recovery

asaj: Sure but it can (and should) be derived from message in the non-proof case

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
         view
         returns (bool)
     {
-        require(_verifyMerkleProof(_metadata, _message), "!merkle");
-        require(_verifyValidatorSignatures(_metadata, _message), "!sigs");
-        return true;
+        return verifyValidatorSignatures(_metadata, _message);
```
asaj, line 57: Why break this out into a separate function at this point?

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
      * the provided checkpoint.
      * @param _metadata ABI encoded module metadata (see MultisigIsmMetadata.sol)
-     * @param _message Formatted Hyperlane message (see Message.sol).
      */
-    function _verifyValidatorSignatures(
+    function verifyValidatorSignatures(
         bytes calldata _metadata,
         bytes calldata _message
-    ) internal view returns (bool) {
+    ) public view returns (bool) {
```
asaj, line 103: Why public?

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
     {
-        require(_verifyMerkleProof(_metadata, _message), "!merkle");
-        require(_verifyValidatorSignatures(_metadata, _message), "!sigs");
-        return true;
+        return verifyValidatorSignatures(_metadata, _message);
     }
 
-    // ============ Internal Functions ============
+    function computeDigest(bytes calldata _metadata, bytes calldata _message)
+        public
```
asaj, line 61: Why public?

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -3,27 +3,26 @@ pragma solidity >=0.8.0;
...
 library MultisigIsmMetadata {
-    uint256 private constant MERKLE_ROOT_OFFSET = 0;
-    uint256 private constant MERKLE_INDEX_OFFSET = 32;
-    uint256 private constant ORIGIN_MAILBOX_OFFSET = 36;
-    uint256 private constant MERKLE_PROOF_OFFSET = 68;
-    uint256 private constant SIGNATURES_OFFSET = 1092;
-    uint256 private constant SIGNATURE_LENGTH = 65;
+    uint8 private constant MERKLE_INDEX_OFFSET = 0;
+    uint8 private constant ORIGIN_MAILBOX_OFFSET = 4;
+    uint8 private constant MERKLE_TYPE_OFFSET = 36;
```
asaj, line 18: nit: Consider `METADATA_TYPE_OFFSET`

asaj: nit: separately, consider putting this first

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -3,27 +3,26 @@ pragma solidity >=0.8.0;
...
 
-    /**
-     * @notice Returns the merkle root of the signed checkpoint.
-     * @param _metadata ABI encoded Multisig ISM metadata.
-     * @return Merkle root of the signed checkpoint
-     */
-    function root(bytes calldata _metadata) internal pure returns (bytes32) {
-        return bytes32(_metadata[MERKLE_ROOT_OFFSET:MERKLE_INDEX_OFFSET]);
+    enum MerkleType {
+        ROOT,
```
asaj, line 24: nit: Consider `MetadataType`, `SuffixType`, or st

MerkleType implies to me that there are different types of merkle trees

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+        pure
+        returns (bytes32 digest)
```
asaj, line 63: Implicit returns considered harmful

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+        ) {
+            signedMessageId = Message.id(_message);
+            root = MultisigIsmMetadata.root(_metadata);
+        } else {
+            bytes32[32] memory proof;
+            (signedMessageId, proof) = MultisigIsmMetadata.idAndProof(
+                _metadata
+            );
+            root = MerkleLib.branchRoot(
+                signedMessageId,
```
asaj, line 80: This needs to be `Message.id(_message)` or else the ISM will verify arbitrary messages 😬

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -54,53 +54,60 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
-            Message.nonce(_message)
+        if (
+            MultisigIsmMetadata.merkleType(_metadata) ==
+            MultisigIsmMetadata.MerkleType.ROOT
+        ) {
+            signedMessageId = Message.id(_message);
+            root = MultisigIsmMetadata.root(_metadata);
+        } else {
+            bytes32[32] memory proof;
+            (signedMessageId, proof) = MultisigIsmMetadata.idAndProof(
```
asaj, line 76: Consider separate functions for these two things

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
asaj: Please add a test case that would have caught this

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
asaj: And more descriptive names, e.g.`signedMessageId` and `merkleProof`

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -122,4 +82,37 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+            );
+            signedMessageId = MultisigIsmMetadata.signedMessageId(_metadata);
+        }
+
+        return
+            CheckpointLib.digest(
+                Message.origin(_message),
+                MultisigIsmMetadata.originMailbox(_metadata),
+                signedRoot,
+                MultisigIsmMetadata.index(_metadata),
```
asaj, line 114: Ah I realized the index is also optional iff no proof is included, it's just Message.nonce(_message)

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -122,4 +82,41 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
         return true;
     }
+
+    function computeDigest(bytes calldata _metadata, bytes calldata _message)
+        internal
+        pure
+        returns (bytes32)
+    {
+        // signed checkpoint properties
+        bytes32 root;
```
asaj, line 92: nit: Underscores for style consistency

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -122,4 +82,41 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+        uint32 index;
+        bytes32 messageId;
+
+        if (
+            MultisigIsmMetadata.suffixType(_metadata) ==
```
asaj, line 97: nit: On second thought, I wonder if "verificationType" might be more readable?

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -122,4 +82,41 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+            MultisigIsmMetadata.SuffixType.ROOT
+        ) {
+            root = MultisigIsmMetadata.root(_metadata);
```
asaj, line 100: Please add a comment explaining why things are where they are for this verification type (and below for the other)

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -122,4 +82,41 @@ abstract contract AbstractMultisigIsm is IMultisigIsm {
...
+            index = Message.nonce(_message);
+            messageId = Message.id(_message);
+        } else {
+            root = MerkleLib.branchRoot(
+                Message.id(_message),
+                MultisigIsmMetadata.merkleProof(_metadata),
+                index
```
asaj, line 107: Where is index coming from? Should this be `Message.nonce(_message)`?

asaj: Please add a test that would have caught this.

solidity/contracts/libs/LegacyCheckpointLib.sol
asaj: Consider moving domainHash to CheckpointLib

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -3,40 +3,35 @@ pragma solidity >=0.8.0;
...
- * [  32:  36] Root index
- * [  36:  68] Origin mailbox address
- * [  68:1092] Merkle proof
- * [1092:????] Validator signatures, 65 bytes each, length == Threshold
+ * [   0:  32] Origin mailbox address
+ * [  32:  VS] Validator signatures (where VS := 32 + Threshold * 65 bytes)
+ * [  VS:????] Suffix
+ *
+ * Format of suffix:
+ * [  -38:  -1] Merkle root
```
asaj, line 11: I don't understand these negative indices, definitely warrant a comment.

Consider using positive indices and making it clear that they are relative to the suffix start

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -88,4 +67,65 @@ library MultisigIsmMetadata {
         uint256 _end = _start + SIGNATURE_LENGTH;
         return _metadata[_start:_end];
     }
+
+    function suffix(
```
asaj, line 71: How much less gas efficient is it for this to return the actual suffix as a bytes slice, and then to take slices from that? I think it would be much more readable

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -3,40 +3,35 @@ pragma solidity >=0.8.0;
...
+    uint8 private constant SIGNATURE_LENGTH = 65;
 
-    /**
-     * @notice Returns the merkle root of the signed checkpoint.
-     * @param _metadata ABI encoded Multisig ISM metadata.
-     * @return Merkle root of the signed checkpoint
-     */
-    function root(bytes calldata _metadata) internal pure returns (bytes32) {
-        return bytes32(_metadata[MERKLE_ROOT_OFFSET:MERKLE_INDEX_OFFSET]);
+    enum SuffixType {
```
tkporter, line 24: took me a pretty long time to understand why this was called root & why only the root was passed in. It wasn't immediately clear to me that one suffix type corresponded with a message ID being signed directly and another with a merkle proof

Comments explaining would help, or possibly even renaming the types to something like this imo:
```
enum MetadataType {
  SignedMessageId,
  MerkleProof
}
```
But I guess I was thinking of these types as establishing the intent of the metadata and not necessarily the contents, which I don't have a super strong take on

solidity/contracts/libs/isms/MultisigIsmMetadata.sol
```diff
@@ -88,4 +67,65 @@ library MultisigIsmMetadata {
...
+            SuffixType(uint8(bytes1(suffix(_metadata, SUFFIX_TYPE_OFFSET, 1))));
+    }
+
+    /**
+     * @notice Returns the merkle root of the signed checkpoint.
+     * @param _metadata ABI encoded Multisig ISM metadata.
+     * @return Merkle root of the signed checkpoint
+     */
+    function root(bytes calldata _metadata) internal pure returns (bytes32) {
+        assert(suffixType(_metadata) == SuffixType.ROOT);
```
tkporter, line 96: wdyt of not having these asserts and just documenting that certain functions should only be used depending on the type? Just noting that it's using gas sorta unnecessarily

yorhodes: its 1 byte of calldata so imo the assertions are worth it

solidity/contracts/libs/CheckpointLib.sol
yorhodes: discussed offline, this is safe

solidity/contracts/libs/Message.sol
```diff
@@ -10,13 +10,13 @@ import {TypeCasts} from "./TypeCasts.sol";
...
     using TypeCasts for bytes32;
 
-    uint256 private constant VERSION_OFFSET = 0;
-    uint256 private constant NONCE_OFFSET = 1;
-    uint256 private constant ORIGIN_OFFSET = 5;
-    uint256 private constant SENDER_OFFSET = 9;
-    uint256 private constant DESTINATION_OFFSET = 41;
-    uint256 private constant RECIPIENT_OFFSET = 45;
-    uint256 private constant BODY_OFFSET = 77;
+    uint8 private constant VERSION_OFFSET = 0;
```
asaj, line 13: Why change these?

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
```diff
@@ -3,27 +3,35 @@ pragma solidity >=0.8.0;
...
  */
-library MultisigIsmMetadata {
-    uint256 private constant MERKLE_ROOT_OFFSET = 0;
-    uint256 private constant MERKLE_INDEX_OFFSET = 32;
-    uint256 private constant ORIGIN_MAILBOX_OFFSET = 36;
-    uint256 private constant MERKLE_PROOF_OFFSET = 68;
-    uint256 private constant SIGNATURES_OFFSET = 1092;
-    uint256 private constant SIGNATURE_LENGTH = 65;
+library MerkleRootMultisigIsmMetadata {
+    uint8 private constant ORIGIN_MAILBOX_OFFSET = 0;
```
asaj, line 13: Do smaller constants do anythign for us?

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
```diff
@@ -3,27 +3,35 @@ pragma solidity >=0.8.0;
 
 /**
  * Format of metadata:
- * [   0:  32] Merkle root
- * [  32:  36] Root index
- * [  36:  68] Origin mailbox address
+ * [   0:  32] Origin mailbox address
+ * [  32:  36] Signed root index
```
asaj, line 7: signed checkpoint index

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
```diff
@@ -3,27 +3,35 @@ pragma solidity >=0.8.0;
...
+ * [  36:  68] Signed message ID
  * [  68:1092] Merkle proof
- * [1092:????] Validator signatures, 65 bytes each, length == Threshold
+ * [1092:????] Validator signatures (length := threshold)
```
asaj, line 10: threshold * 65?

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
```diff
@@ -3,27 +3,35 @@ pragma solidity >=0.8.0;
...
-    uint256 private constant MERKLE_PROOF_OFFSET = 68;
-    uint256 private constant SIGNATURES_OFFSET = 1092;
-    uint256 private constant SIGNATURE_LENGTH = 65;
+library MerkleRootMultisigIsmMetadata {
+    uint8 private constant ORIGIN_MAILBOX_OFFSET = 0;
+    uint8 private constant MERKLE_INDEX_OFFSET = 32;
+    uint8 private constant SIGNED_MESSAGE_ID_OFFSET = 36;
+    uint8 private constant MERKLE_PROOF_OFFSET = 68;
+    uint16 private constant MERKLE_PROOF_LENGTH = 32 * 32;
+    uint16 private constant SIGNATURES_OFFSET = 1092;
```
asaj, line 18: MERKLE_PROOF_OFFSET + MERKLE_PROOF_LENGTH?

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
```diff
@@ -34,21 +42,20 @@ library MultisigIsmMetadata {
     function index(bytes calldata _metadata) internal pure returns (uint32) {
         return
             uint32(
-                bytes4(_metadata[MERKLE_INDEX_OFFSET:ORIGIN_MAILBOX_OFFSET])
+                bytes4(_metadata[MERKLE_INDEX_OFFSET:MERKLE_INDEX_OFFSET + 4])
```
asaj, line 45: nit: CHECKPOINT_INDEX_OFFSET

solidity/contracts/libs/isms/MessageIdMultisigIsmMetadata.sol
```diff
@@ -0,0 +1,59 @@
...
+            );
+    }
+
+    /**
+     * @notice Returns the merkle root of the signed checkpoint.
+     * @param _metadata ABI encoded Multisig ISM metadata.
+     * @return Merkle root of the signed checkpoint
+     */
+    function root(bytes calldata _metadata) internal pure returns (bytes32) {
+        return bytes32(_metadata[MERKLE_ROOT_OFFSET:MERKLE_ROOT_OFFSET + 32]);
```
asaj, line 38: if you do CHECKPOINT_INDEX, make this CHECKPOINT_ROOT

typescript/sdk/src/ism/contracts.ts
```diff
@@ -1,11 +1,11 @@
 import {
   DomainRoutingIsmFactory__factory,
   StaticAggregationIsmFactory__factory,
-  StaticMultisigIsmFactory__factory,
+  StaticMerkleRootMultisigIsmFactory__factory,
 } from '@hyperlane-xyz/core';
 
 export const ismFactoryFactories = {
-  multisigIsmFactory: new StaticMultisigIsmFactory__factory(),
+  multisigIsmFactory: new StaticMerkleRootMultisigIsmFactory__factory(),
```
asaj, line 8: Should we deploy both?

asaj (commented):
Very nice. Haven't looked at tests yet

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
yorhodes: bytecode size

solidity/contracts/libs/Message.sol
yorhodes: yeah idk

solidity/contracts/libs/isms/MerkleRootMultisigIsmMetadata.sol
yorhodes: we could do this but it feels like a departure from the current pattern?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+    /// syncer, looking for the highest index that >= `threshold` validators
+    /// have returned.
+    ///
+    /// Attempts to find a quorum of signed checkpoints from that index,
+    /// iterating backwards if unsuccessful, until the (optional) index is
+    /// reached.
+    ///
+    /// Note it's possible to not find a quorum.
+    #[instrument(err, skip(self))]
+    pub async fn fetch_checkpoint_in_range(
```
asaj, line 203: Assuming this is the same as the other but with fetch_checkpoint instead?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+        for validator in validators.iter() {
+            let addr = H160::from(*validator);
+            if let Some(checkpoint_syncer) = self.checkpoint_syncers.get(&addr) {
+                // Gracefully ignore an error fetching the checkpoint from a validator's
+                // checkpoint syncer, which can happen if the validator has not
+                // signed the checkpoint at `index`.
+                if let Ok(Some(signed_checkpoint)) = checkpoint_syncer.fetch_checkpoint(index).await
+                {
+                    // If the signed checkpoint is for a different index, ignore it
+                    if signed_checkpoint.value.index != index {
```
asaj, line 284: Should we check the expected message_id here as well?

rust/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -121,7 +121,11 @@ impl S3Storage {
     }
 
     fn checkpoint_key(index: u32) -> String {
-        format!("checkpoint_{}.json", index)
+        format!("checkpoint_{index}.json")
+    }
+
+    fn checkpoint_with_id_key(index: u32) -> String {
```
asaj, line 127: ditto re legacy prefix

rust/hyperlane-core/src/accumulator/incremental.rs
```diff
@@ -11,6 +11,13 @@ pub struct IncrementalMerkle {
     count: usize,
 }
 
+impl IncrementalMerkle {
+    /// Build from components.
+    pub fn new(branch: [H256; TREE_DEPTH], count: usize) -> Self {
```
asaj, line 16: nit: Can you derive this instead?

rust/hyperlane-core/src/traits/indexer.rs
```diff
@@ -1,4 +1,3 @@
-//! An Indexer provides a common interface for bubbling up chain-specific
```
asaj, line 1: Why remove this?

rust/hyperlane-core/src/types/checkpoint.rs
```diff
@@ -20,47 +19,63 @@ pub struct Checkpoint {
...
+    /// hash of message emitted from mailbox checkpoint.index
+    pub message_id: H256,
+}
+
+impl Signable for Checkpoint {
+    /// A hash of the checkpoint contents.
+    /// The EIP-191 compliant version of this hash is signed by validators.
+    fn signing_hash(&self) -> H256 {
+        // sign:
+        // domain_hash(mailbox_address, mailbox_domain) || root || index (as u32) || message_id
```
asaj, line 37: nit: remove message id

rust/utils/run-locally/src/main.rs
```diff
@@ -149,11 +149,14 @@ fn main() -> ExitCode {
         fs::create_dir_all(&log_dir).expect("Failed to make log dir");
     }
     let build_log = concat_path(&log_dir, "build.log");
-    let hardhat_log = concat_path(&log_dir, "hardhat.stdout.log");
+    let node_log = concat_path(&log_dir, "node.stdout.log");
```
asaj, line 152: nit: consider "anvil_log", as "node" is overloaded with the js runtime

rust/utils/run-locally/src/main.rs
```diff
@@ -388,6 +375,24 @@ fn main() -> ExitCode {
     state.watchers.push(kathy_stderr);
     kathy.wait().unwrap();
 
+    for (i, validator_env) in validator_envs.iter().enumerate() {
```
asaj, line 378: nit: update the comment above "Send half the kathy messages before starting the agents"

rust/utils/run-locally/src/main.rs
```diff
@@ -425,24 +431,11 @@ fn main() -> ExitCode {
...
-                    return ExitCode::from(1);
-                }
-                None => {}
-            }
-        }
         if ci_mode {
             // for CI we have to look for the end condition.
             let num_messages_expected = (kathy_messages / 2) as u32 * 2;
-            if kathy_done && termination_invariants_met(num_messages_expected) {
+            if termination_invariants_met(num_messages_expected) {
```
asaj, line 438: nit: Update log messages below to reflect the removal of the kathy finishing requirement

typescript/infra/hardhat.config.ts
```diff
@@ -70,6 +76,7 @@ task('kathy', 'Dispatches random hyperlane messages')
         const remoteId = multiProvider.getDomainId(remote);
         const mailbox = core.getContracts(local).mailbox;
         const igp = igps.getContracts(local).interchainGasPaymaster;
+
```
asaj, line 79: nit: remove

typescript/infra/hardhat.config.ts
```diff
@@ -59,6 +61,10 @@ task('kathy', 'Dispatches random hyperlane messages')
       const recipient = await recipientF.deploy();
       await recipient.deployTransaction.wait();
 
+      const isAutomine: boolean = await hre.network.provider.send(
```
asaj, line 64: What's going on here? Testing if the provider supports this?

typescript/infra/hardhat.config.ts
```diff
@@ -89,6 +97,17 @@ task('kathy', 'Dispatches random hyperlane messages')
           } on ${local} with nonce ${(await mailbox.count()) - 1}`,
         );
         console.log(await chainSummary(core, local));
+        console.log(await chainSummary(core, remote));
+
+        await sleep(timeout);
+      }
+
+      while (taskArgs.mineforever && isAutomine) {
+        // mine confirmation blocks instantly
```
asaj, line 106: I think you can just mine one block at a time

typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -4,7 +4,7 @@ import { ChainMap } from '../types';
 export const defaultMultisigIsmConfigs: ChainMap<MultisigIsmConfig> = {
   // ----------------- Mainnets -----------------
   celo: {
-    type: ModuleType.MULTISIG,
+    type: ModuleType.MERKLE_ROOT_MULTISIG,
```
asaj, line 7: This should be legacy still, right? (and below)

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -116,16 +144,15 @@ export class HyperlaneIsmFactory extends HyperlaneApp<IsmFactoryFactories> {
     );
     await routingIsm.transferOwnership(config.owner);
     const address = dispatchLogs[0].args['module'];
-    return DomainRoutingIsm__factory.connect(address, signer);
+    return IRoutingIsm__factory.connect(address, signer);
```
asaj, line 147: I don't think this change is right, we rely on the specific implementation for the deployer

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -136,7 +163,7 @@ export class HyperlaneIsmFactory extends HyperlaneApp<IsmFactoryFactories> {
       addresses,
       config.threshold,
     );
-    return StaticAggregationIsm__factory.connect(address, signer);
+    return IAggregationIsm__factory.connect(address, signer);
```
asaj, line 166: Ditto

typescript/sdk/src/consts/environments/test.json
```diff
@@ -1,38 +1,41 @@
...
-    "validatorAnnounce": "0xa513E6E4b8f2a923D98304ec87F64353C4D5C853",
-    "proxyAdmin": "0x8A791620dd6260079BF849Dc5567aDC3F2FdC318",
-    "mailbox": "0x0165878A594ca255338adfa4d48449f69242Eb8F",
-    "interchainGasPaymaster": "0x0DCd1Bf9A1b36cE34237eEaFef220932846BCD82",
-    "defaultIsmInterchainGasPaymaster": "0x0B306BF915C4d645ff596e518fAf3F9669b97016",
-    "multisigIsm": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
-    "multisigIsmFactory": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
-    "aggregationIsmFactory": "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
-    "routingIsmFactory": "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"
+    "storageGasOracle": "0x36C02dA8a0983159322a80FFE9F24b1acfF8B570",
```
asaj, line 3: Surprised this file isn't gitignored

typescript/sdk/src/ism/HyperlaneIsmFactoryDeployer.ts
```diff
@@ -33,21 +33,33 @@ export class HyperlaneIsmFactoryDeployer extends HyperlaneDeployer<
   async deployContracts(
     chain: ChainName,
   ): Promise<HyperlaneContracts<IsmFactoryFactories>> {
-    const multisigIsmFactory = await this.deployContract(
+    const merkleRootMultisigIsm = await this.deployContract(
```
asaj, line 36: I'm confused by the name changes here. We're deploying factories, no?

asaj (commented):
Skipped the contracts for expediency

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -32,13 +35,13 @@ impl MultisigCheckpointSyncer {
...
-    pub async fn fetch_checkpoint_in_range(
+    pub async fn legacy_fetch_checkpoint_in_range(
         &self,
-        validators: &Vec<H256>,
+        validators: &[H256],
         threshold: usize,
         minimum_index: u32,
         maximum_index: u32,
-    ) -> Result<Option<MultisigSignedCheckpoint>> {
+    ) -> Result<Option<MultisigSignedCheckpoint<Checkpoint>>> {
```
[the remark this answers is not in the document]
yorhodes: because the `Checkpoint` type is not really legacy

typescript/infra/hardhat.config.ts
yorhodes: makes it agnostic to the node config

typescript/sdk/src/consts/multisigIsm.ts
yorhodes: woops yes

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
yorhodes: uhhh we do? where

typescript/sdk/src/ism/HyperlaneIsmFactoryDeployer.ts
yorhodes: urmmmm

asaj (commented):
Reviewed the diff

rust/agents/validator/src/validator.rs
```diff
@@ -74,17 +103,139 @@ impl BaseAgent for Validator {
...
 
-        run_all(vec![submit.spawn()])
+        let empty_tree = IncrementalMerkle::default();
+        let lag = NonZeroU64::new(self.reorg_period);
+        let tip_tree = self
+            .mailbox
+            .tree(lag)
+            .await
+            .expect("failed to get mailbox tree");
+        let backfill_target = submitter.checkpoint(&tip_tree);
```
asaj, line 155: Should we pull the checkpoint via Mailbox.checkpoint? This way safety relies less on the incremental merkle tree implementation in the agent

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
asaj: Ah okay nvm you're right, we do but we rely on it above, not after the return

rust/hyperlane-base/src/types/multisig.rs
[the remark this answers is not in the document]
asaj: Ah gotcha, sold.

solidity/contracts/isms/multisig/AbstractMerkleRootMultisigIsm.sol
```diff
@@ -0,0 +1,64 @@
...
+import {IInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
+import {AbstractMultisigIsm} from "./AbstractMultisigIsm.sol";
+import {MerkleRootMultisigIsmMetadata} from "../../libs/isms/MerkleRootMultisigIsmMetadata.sol";
+import {Message} from "../../libs/Message.sol";
+import {MerkleLib} from "../../libs/Merkle.sol";
+import {CheckpointLib} from "../../libs/CheckpointLib.sol";
+
+/**
+ * @title MerkleRootMultisigIsm
+ * @notice Manages per-domain m-of-n Validator sets that are used to verify
```
asaj, line 14: nit: This doesn't actually manage a set at all, it provides abstract logic for verifying messages given a `validatorsAndThreshold` implementation

solidity/contracts/isms/multisig/AbstractMessageIdMultisigIsm.sol
```diff
@@ -0,0 +1,55 @@
...
+// ============ Internal Imports ============
+import {IInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
+import {AbstractMultisigIsm} from "./AbstractMultisigIsm.sol";
+import {MessageIdMultisigIsmMetadata} from "../../libs/isms/MessageIdMultisigIsmMetadata.sol";
+import {Message} from "../../libs/Message.sol";
+import {CheckpointLib} from "../../libs/CheckpointLib.sol";
+
+/**
+ * @title MultisigIsm
+ * @notice Manages per-domain m-of-n Validator sets that are used to verify
```
asaj, line 13: ditto

solidity/contracts/isms/multisig/AbstractMessageIdMultisigIsm.sol
```diff
@@ -0,0 +1,55 @@
...
+
+// ============ Internal Imports ============
+import {IInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
+import {AbstractMultisigIsm} from "./AbstractMultisigIsm.sol";
+import {MessageIdMultisigIsmMetadata} from "../../libs/isms/MessageIdMultisigIsmMetadata.sol";
+import {Message} from "../../libs/Message.sol";
+import {CheckpointLib} from "../../libs/CheckpointLib.sol";
+
+/**
+ * @title MultisigIsm
```
asaj, line 12: nit: title

solidity/contracts/isms/multisig/AbstractMultisigIsm.sol
```diff
@@ -8,22 +8,17 @@ import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
...
 import {Message} from "../../libs/Message.sol";
-import {MultisigIsmMetadata} from "../../libs/isms/MultisigIsmMetadata.sol";
-import {CheckpointLib} from "../../libs/CheckpointLib.sol";
 import {MerkleLib} from "../../libs/Merkle.sol";
 
 /**
  * @title MultisigIsm
  * @notice Manages per-domain m-of-n Validator sets that are used to verify
  * interchain messages.
+ * @dev See ./AbstractMerkleRootMultisigIsm.sol and ./AbstractMessageIdMultisigIsm.sol
```
asaj, line 17: Ditto on above comment

solidity/contracts/isms/multisig/StaticMultisigIsmFactory.sol
```diff
@@ -1,16 +1,55 @@
...
+import {MetaProxy} from "../../libs/MetaProxy.sol";
 import {StaticMOfNAddressSetFactory} from "../../libs/StaticMOfNAddressSetFactory.sol";
 
-contract StaticMultisigIsmFactory is StaticMOfNAddressSetFactory {
-    function _deployImplementation()
-        internal
-        virtual
+/**
+ * @title AbstractMetaProxyMultisigIsm
+ * @notice Manages per-domain m-of-n Validator sets that are used
```
asaj, line 12: Sets or set? Modify this comment to be singular

solidity/contracts/isms/multisig/StaticMultisigIsmFactory.sol
```diff
@@ -1,16 +1,55 @@
...
+    function _deployImplementation() internal override returns (address) {
+        return address(new StaticMerkleRootMultisigIsm());
+    }
+}
+
+// solhint-disable-next-line no-empty-blocks
+contract StaticMessageIdMultisigIsm is
+    AbstractMessageIdMultisigIsm,
+    AbstractMetaProxyMultisigIsm
+{
```
asaj, line 47: ditto

solidity/contracts/isms/multisig/StaticMultisigIsmFactory.sol
```diff
@@ -1,16 +1,55 @@
...
         override
-        returns (address)
+        returns (address[] memory, uint8)
     {
-        return address(new StaticMultisigIsm());
+        return abi.decode(MetaProxy.metadata(), (address[], uint8));
+    }
+}
+
+// solhint-disable-next-line no-empty-blocks
```
asaj, line 29: Could probably use documentation for all contracts in this file

solidity/contracts/libs/LegacyCheckpointLib.sol
```diff
@@ -0,0 +1,50 @@
...
+
+// ============ Internal Imports ============
+import {CheckpointLib} from "./CheckpointLib.sol";
+
+library LegacyCheckpointLib {
+    /**
+     * @notice Returns the digest validators are expected to sign when signing checkpoints.
+     * @param _origin The origin domain of the checkpoint.
+     * @param _originMailbox The address of the origin mailbox as bytes32.
+     * @return The digest of the checkpoint.
```
asaj, line 15: ditto

solidity/test/isms/MultisigIsm.t.sol
```diff
@@ -5,28 +5,65 @@ import "forge-std/Test.sol";
...
+        uint32 checkpointIndex = uint32(mailbox.count() - 1);
+        bytes32 messageId = message.id();
+        bytes32 digest = CheckpointLib.digest(
+            domain,
+            mailboxAsBytes32,
+            checkpointRoot,
+            checkpointIndex,
+            messageId
+        );
+        bytes memory metadata = abi.encodePacked(
```
asaj, line 51: nit: Consider putting mailboxAsBytes32 in the prefix, even though it would mean duplication of one line of code

solidity/test/isms/MultisigIsm.t.sol
```diff
@@ -100,7 +121,48 @@ contract MultisigIsmTest is Test {
     ) public {
         vm.assume(0 < m && m <= n && n < 10);
         bytes memory message = getMessage(destination, recipient, body);
-        bytes memory metadata = getMetadata(m, n, seed);
-        assertTrue(ism.verify(metadata, message));
+        bytes memory metadata = getMetadata(m, n, seed, message);
+
+        // changing single bit in metadata should fail signature verification
```
asaj, line 126: Single byte?

asaj (commented):
Re-reviewed the contracts, just nits, we're in the home stretch

rust/agents/validator/src/submit.rs
```diff
@@ -28,89 +33,104 @@ impl ValidatorSubmitter {
...
             sleep(self.interval).await;
         }
-        Ok(())
-    }
 
-    async fn main_task(self) -> Result<()> {
-        self.announce_task().await?;
+        // TODO: remove this once validator is tolerant of tasks exiting
+        loop {
+            sleep(Duration::from_secs(60 * 60)).await;
```
mattiecnvr, line 131: If we keep this, let's make this sleep duration longer, e.g. `u64::MAX`

rust/agents/validator/src/validator.rs
```diff
@@ -74,17 +103,139 @@ impl BaseAgent for Validator {
...
+        for checkpoint_sync_task in self.run_checkpoint_submitters().await {
+            tasks.push(checkpoint_sync_task);
+        }
+
+        run_all(tasks)
+    }
+}
+
+impl Validator {
+    async fn run_message_sync(&self) -> Instrumented<JoinHandle<eyre::Result<()>>> {
```
mattiecnvr, line 120:
```suggestion
    async fn run_message_sync(&self) -> Instrumented<JoinHandle<Result<()>>> {
```

rust/agents/validator/src/validator.rs
```diff
@@ -74,17 +103,139 @@ impl BaseAgent for Validator {
...
+        tokio::spawn(async move {
+            contract_sync
+                .clone()
+                .sync("dispatched_messages", cursor)
+                .await
+        })
+        .instrument(info_span!("MailboxMessageSyncer"))
+    }
+
+    async fn run_checkpoint_submitters(&self) -> Vec<Instrumented<JoinHandle<eyre::Result<()>>>> {
```
mattiecnvr, line 137:
```suggestion
    async fn run_checkpoint_submitters(&self) -> Vec<Instrumented<JoinHandle<Result<()>>>> {
```

rust/agents/validator/src/validator.rs
```diff
@@ -74,17 +103,139 @@ impl BaseAgent for Validator {
...
+                    if !outcome.executed {
+                        error!(
+                            hash=?outcome.txid,
+                            "Transaction attempting to announce validator reverted"
+                        );
+                    }
+                }
+            }
+            sleep(self.interval).await;
+        }
```
mattiecnvr, line 237:
```suggestion
            let Some(locations) = self
                .validator_announce
                .get_announced_storage_locations(&validators)
                .await?
                .first()
            else {
                sleep(self.interval).await;
                continue;
            };
                
            if locations.contains(&self.checkpoint_syncer.announcement_location()) {
                info!("Validator has announced signature storage location");
                break;
            }
            info!("Validator has not announced signature storage location");
            let balance_delta = self
                .validator_announce
                .announce_tokens_needed(signed_announcement.clone())
                .await?;
            if balance_delta > U256::zero() {
                warn!(
                    tokens_needed=%balance_delta,
                    validator_address=?announcement.validator,
                    "Please send tokens to the validator address to announce",
                );
            } else {
                let outcome = self
                    .validator_announce
                    .announce(signed_announcement.clone(), None)
                    .await?;
                if !outcome.executed {
                    error!(
                        hash=?outcome.txid,
                        "Transaction attempting to announce validator reverted"
                    );
                }
            }
            
            sleep(self.interval).await;
        }
```

rust/chains/hyperlane-ethereum/src/interchain_security_module.rs
```diff
@@ -96,9 +97,14 @@ where
...
     #[instrument(err, ret)]
-    async fn module_type(&self) -> ChainResult<u8> {
-        let module_type = self.contract.module_type().call().await?;
-        Ok(module_type)
+    async fn module_type(&self) -> ChainResult<ModuleType> {
+        let module = self.contract.module_type().call().await?;
+        if let Some(module_type) = ModuleType::from_u8(module) {
+            Ok(module_type)
+        } else {
+            warn!("Unknown module type: {}", module);
```
mattiecnvr, line 105: Formatting content into the message string makes it hard to filter in gcp
```suggestion
            warn!(%module, "Unknown module type");
```

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -333,6 +338,67 @@ where
...
+            .call()
+            .await? as usize;
+
+        let tree = IncrementalMerkle::new(branch, count);
+
+        // validate tree built from storage slot lookups matches expected
+        // result from root() view call at consistent block
+        assert_eq!(tree.root(), expected_root);
+
+        return Ok(tree);
```
mattiecnvr, line 399:
```suggestion
        Ok(tree)
```

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+                        trace!(?address, ?index, "Validator returned latest index");
+                        latest_indices.push(index);
+                    }
+                    err => {
+                        debug!(?address, ?err, "Failed to get latest index from validator");
+                    }
+                }
+            }
+        }
+        debug!(latest_indices=?latest_indices, "Fetched latest indices from checkpoint syncers");
```
mattiecnvr, line 227:
```suggestion
        debug!(?latest_indices, "Fetched latest indices from checkpoint syncers");
```

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+    ) -> Result<Option<MultisigSignedCheckpoint<CheckpointWithMessageId>>> {
+        // Keeps track of signed validator checkpoints for a particular root.
+        // In practice, it's likely that validators will all sign the same root for a
+        // particular index, but we'd like to be robust to this not being the case
+        let mut signed_checkpoints_per_root: HashMap<
+            H256,
+            Vec<SignedCheckpointWithSigner<CheckpointWithMessageId>>,
+        > = HashMap::new();
+
+        for validator in validators.iter() {
```
mattiecnvr, line 275: I think this is valid
```suggestion
        for validator in validators {
```

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+            let addr = H160::from(*validator);
+            if let Some(checkpoint_syncer) = self.checkpoint_syncers.get(&addr) {
```
mattiecnvr, line 277: Can we rewrite this with `let-else`?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+                            1 // length of 1
+                        }
+                    };
+                    debug!(
+                        validator = format!("{:#x}", validator),
+                        index = index,
+                        root = format!("{:#x}", root),
+                        signature_count = signature_count,
+                        "Found signed checkpoint"
+                    );
```
mattiecnvr, line 329:
```suggestion
                    debug!(
                        validator=format!("{validator:#x}"),
                        index,
                        root=format!("{root:#x}"),
                        signature_count,
                        "Found signed checkpoint"
                    );
```

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+                                    signed_checkpoints,
+                                )?;
+                            debug!(checkpoint=?checkpoint, "Fetched multisig checkpoint");
+                            return Ok(Some(checkpoint));
+                        }
+                    }
+                } else {
+                    debug!(
+                        validator = format!("{:#x}", validator),
+                        index = index,
```
mattiecnvr, line 344:
```suggestion
                        validator=format!("{validator:#x}"),
                        index,
```

rust/hyperlane-base/src/types/multisig.rs
mattiecnvr: We might consider making the alt-formatting the default since it seems to be used the most

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -179,4 +186,170 @@ impl MultisigCheckpointSyncer {
...
+                        "Found signed checkpoint"
+                    );
+                    // If we've hit a quorum, create a MultisigSignedCheckpoint
+                    if signature_count >= threshold {
+                        if let Some(signed_checkpoints) = signed_checkpoints_per_root.get(&root) {
+                            let checkpoint =
+                                MultisigSignedCheckpoint::<CheckpointWithMessageId>::try_from(
+                                    signed_checkpoints,
+                                )?;
+                            debug!(checkpoint=?checkpoint, "Fetched multisig checkpoint");
```
mattiecnvr, line 337:
```suggestion
                            debug!(?checkpoint, "Fetched multisig checkpoint");
```

rust/hyperlane-core/Cargo.toml
```diff
@@ -33,6 +33,7 @@ thiserror.workspace = true
 # version determined by ethers-rs
 primitive-types = "*"
 lazy_static = "*"
+derive_more = "0.99.17"
```
mattiecnvr, line 36:
```suggestion
derive_more.workspace = true
```

Also please sort alphabetically

mattiecnvr (approved):
I just looked at the rust stuff and mostly focused on rust idoms since I think @asaj already looked more at the logic

mattiecnvr (approved):
I just looked at the rust stuff and mostly focused on rust idioms since I think @asaj already looked more at the logic

rust/agents/validator/src/submit.rs
```diff
@@ -28,89 +33,104 @@ impl ValidatorSubmitter {
...
+            }
+
             sleep(self.interval).await;
         }
-        Ok(())
-    }
 
-    async fn main_task(self) -> Result<()> {
-        self.announce_task().await?;
+        // TODO: remove this once validator is tolerant of tasks exiting
```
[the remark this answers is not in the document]
yorhodes: leaving this for now as we want to capture errors in backfill

PR #2195, opened 2023-05-05
Omniscient Relayer

### LINKED ISSUE #1899: Omniscient Relayer

mattiecnvr:
### Description

Adds support for running a relayer which delivers messages from multiple origin chains instead of just one origin chain.

### Drive-by changes

### Related issues

- Blocked by #2148
- Fixes #1899 

### Backward compatibility

Yes

None

### Testing

Manual
Unit Tests


rust/agents/relayer/src/settings/mod.rs
```diff
@@ -157,7 +158,9 @@ decl_settings!(Relayer,
     Raw {
         /// Database path (path on the fs)
         db: Option<String>,
-        // The name of the origin chain
+        // Comma separated list of origin chains.
+        originchainnames: Option<String>,
```
asaj, line 162: What is the advantage of allowing origin and destination chain names to be specified separately?

asaj: Sorry, how is it backwards compatible? Don't you need to specify origin chain names now?

> the future possibility of having central relayers for certain destination chains
> It also allows people to split up their instances if that is important to them for some reason.

These feel possible via blacklist, no?

IMO separate origin/destination chain config feels more complicated for the typical use case

mattiecnvr: From our discussion this morning at standup:

Instead of `originchainnames` and `destinationchainnames` we should accept just a single list of chains (but maintain support for the old `originchainname` and `destinationchainnames` config for now).

If an operator has a new chain A and wants to connect it with B and C, the most likely case is they want to relay A <-> B and A <-> C, but NOT B <-> C and that can be easily handled with a blacklist. Since this does not change the fact that all of the chains still need to be indexed, it would not need any additional code changes for efficiency.

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,175 @@
...
+#[allow(unused_imports)] // required for enum_dispatch
+use super::pending_message::PendingMessage;
+
+/// A pending operation that will be run by the submitter and cause a
+/// transaction to be sent.
+#[async_trait]
+#[enum_dispatch]
+pub trait PendingOperation {
+    /// Prepare to run this operation. This will be called before every run and
+    /// will usually have a very short gap between it and the run call.
```
asaj, line 21: What is the "run call"?

A sequence diagram might help explain things. IIUC it's validate, prepare, submit?

Suggest putting them in that order in the trait as well

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -129,254 +113,132 @@ impl SerialSubmitter {
...
-            .await
-        {
-            Ok(tx_cost_estimate) => tx_cost_estimate,
-            Err(error) => {
-                info!(?error, "Error estimating process costs");
-                return Ok(false);
+        match op.submit().await {
+            TxRunResult::Success => {
+                self.metrics.txs_submitted.inc();
+                self.validation_queue.push(Reverse(op));
```
asaj, line 157: Why do we push on success?

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,175 @@
...
+    async fn prepare(&mut self) -> TxPrepareResult {
```
asaj, line 22: nit: Consider renaming the return types for consistency. Some suggestions:

PreparationResult/SubmissionResult/ValidationResult
PrepareResult/SubmitResult/ValidateResult

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+            TxValidationResult::Retry
+        }
+    }
+
+    fn next_attempt_after(&self) -> Option<Instant>;
+
+    fn submitted(&self) -> bool;
+
+    fn ready_to_be_processed(&self) -> bool {
```
asaj, line 49: nit: Consider renaming for consistency. Should this be `ready_to_be_submitted`?

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -129,254 +113,132 @@ impl SerialSubmitter {
...
-            Ok(false)
+    /// Validate submitted operations.
+    async fn tick_validate(&mut self) -> Result<()> {
+        while let Some(Reverse(mut op)) = self.validation_queue.pop() {
+            let res: TxValidationResult = op.validate().await;
+            match res {
+                TxValidationResult::Valid => {
+                    self.metrics.txs_validated.inc();
+                }
+                TxValidationResult::NotReady => {
```
asaj, line 180: What is the difference between NotReady and Retry?

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -129,254 +113,132 @@ impl SerialSubmitter {
...
-        //  only these messages, the number of retries could be considered.
-
-        let gas_limit = tx_cost_estimate.gas_limit;
-
-        if let Some(max_limit) = self.transaction_gas_limit {
-            if gas_limit > max_limit {
-                info!("Message delivery estimated gas exceeds max gas limit");
-                return Ok(false);
+            TxRunResult::DoNotRetry => {
+                self.metrics.txs_submitted.inc();
```
asaj, line 160: Why do we increment on DoNotRetry? Presumably we did not submit successfully, right?

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[allow(dead_code)] // Inner types are for present _and_ future use so allow unused variants.
+pub enum TxPrepareResult {
+    /// Txn is ready to be submitted
+    Ready,
+    /// This Txn is not ready to be attempted again yet
+    NotReady,
+    /// Txn preparation failed and we should not try again or it has already
+    /// been processed.
+    DoNotRetry,
```
asaj, line 106: nit: Consider "Drop", ditto below

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,175 @@
...
+            (None, Some(_)) => Less,
+            (None, None) => match (self, other) {
+                (PendingMessage(a), PendingMessage(b)) => a.message.nonce.cmp(&b.message.nonce),
+            },
+        }
+    }
+}
+
+#[allow(dead_code)] // Inner types are for present _and_ future use so allow unused variants.
+pub enum TxPrepareResult {
```
asaj, line 99: nit: If I'm right about validate/prepare/submit being the order, please order these enums accordingly

asaj (commented):
Partial review

rust/agents/relayer/src/msg/pending_operation.rs
asaj: Ah nvm I see now it's prepare, submit, validate, which makes sense given the order in this trait

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -129,254 +113,136 @@ impl SerialSubmitter {
...
-            );
-            Ok(true)
-        } else {
-            info!(
-                hash=?outcome.txid,
-                "Transaction attempting to process transaction reverted"
-            );
-            Ok(false)
+    /// Validate submitted operations.
+    async fn tick_validate(&mut self) -> Result<()> {
```
asaj, line 173: I wonder if we can make things even more generic (and if that's a good idea).

For example, you could imagine three queues of PendingMessages:
1. prepare_queue
2. submit_queue
3. confirm_queue

Each queue implements the PendingMessageQueue trait, which has:
```
    async fn pop(&mut self) -> PendingMessage, PopResult;
```

where:
```
pub enum PopResult {
    /// PendingMessage should be moved to a different queue, e.g...
    /// Successful preparation: prepare -> submit
    /// Successful submission: submit -> confirm
    /// Failed submission: submit -> prepare
    /// Failed confirmation: confirm -> prepare
    Move(PendingMessageQueue),
    
    /// PendingMessage should be dropped entirely, e.g....
    /// prepare queue checks and it's already been delivered
    /// confirmation queue checks and it's been confirmed 
    Drop,
    /// A retry-able error occurred and we should push to the back of
    /// the queue to retry later, e.g....
    /// Prepare queue unable to assemble metadata
    /// Prepare queue failed gas payment enforcement
    Retry (or "Push"),
    /// Pass the error up the chain, this is non-recoverable and indicates a
    /// system failure.
    CriticalFailure(Report),
}

**Review by asaj — COMMENTED** [review]
Big fan of decoupling message delivery into different stages

File: rust/agents/relayer/src/msg/metadata/base.rs [hunk 11]
```diff
@@ -54,6 +54,7 @@ pub struct BaseMetadataBuilder {
     metrics: Arc<CoreMetrics>,
     /// ISMs can be structured recursively. We keep track of the depth
     /// of the recursion to avoid infinit loops.
+    #[new(default)]
```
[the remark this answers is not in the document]
mattiecnvr: It tells the `new` derive-er that this value should be `Default::default()`.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,349 @@
...
+
+use super::{
+    gas_payment::GasPaymentEnforcer,
+    metadata::{BaseMetadataBuilder, MetadataBuilder},
+    pending_operation::*,
+};
+
+/// The message context contains the links needed to submit a message. Each
+/// instance is for a unique origin -> destination pairing.
+pub struct MessageCtx {
```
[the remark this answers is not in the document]
mattiecnvr: On the fence for this one because it is `pub(crate)` and not `pub` (I know it looks like it is `pub` but in the parent module it has a limited export.

rust/agents/relayer/src/msg/metadata/base.rs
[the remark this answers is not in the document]
asaj: Where does the default value come from?

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Mailbox on the destination chain.
+    pub destination_mailbox: CachingMailbox,
+    /// Origin chain database to verify gas payments.
+    pub origin_db: HyperlaneDB,
+    /// Used to construct the ISM metadata needed to verify a message from the
+    /// origin.
+    pub metadata_builder: BaseMetadataBuilder,
```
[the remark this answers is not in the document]
mattiecnvr: Good idea

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,349 @@
...
+    #[new(default)]
+    num_retries: u32,
+    #[new(value = "Instant::now()")]
+    last_attempted_at: Instant,
+    #[new(default)]
+    next_attempt_after: Option<Instant>,
+}
+
+/// State for the next submission attempt generated by a prepare call.
+pub struct PendingMessageState {
```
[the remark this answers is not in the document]
mattiecnvr: I just realized this does not need to be exported. Given that it is only accessible from this file I think that makes the actual name less important. With that in mind I like `SubmissionData`.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,349 @@
...
+                }
+            })
+            .unwrap_or(0);
+        write!(f, "PendingMessage {{ num_retires: {}, since_last_attempt_s: {last_attempt}, next_attempt_after_s: {next_attempt}, message: {:?} }}",
+               self.num_retries, self.message)
+    }
+}
+
+impl PartialEq for PendingMessage {
+    fn eq(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
mattiecnvr: Yes, it should

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,349 @@
...
+        self.num_retries == other.num_retries && self.message.nonce == other.message.nonce
+    }
+}
+
+impl Eq for PendingMessage {}
```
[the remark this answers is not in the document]
mattiecnvr: It informs the Rust compiler that the following is impossible
`a == b && b == c && a != c`

rust/agents/relayer/src/msg/pending_operation.rs
mattiecnvr: Yeah, fixed my doc

rust/agents/relayer/src/msg/pending_operation.rs
mattiecnvr: Yeah, that makes sense

rust/agents/relayer/src/msg/pending_operation.rs
mattiecnvr: It's `Prepare`, `Submit`, `Validate`. Validate makes sure it was not re-org-ed out. LMK if you have a better name for this.

rust/agents/relayer/src/msg/pending_operation.rs
mattiecnvr: I like that

rust/agents/relayer/src/msg/pending_operation.rs
asaj: Yeah I figured this out later. Suggested "Confirm" in a different comment

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
asaj: Maybe say they're equal if the nonce and origin domain are the same? That should be sufficient

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,179 @@
...
+/// transaction to be sent.
+#[async_trait]
+#[enum_dispatch]
+pub trait PendingOperation {
+    /// The domain this operation will take place on.
+    fn domain(&self) -> &HyperlaneDomain;
+
+    /// Prepare to run this operation. This will be called before every run and
+    /// will usually have a very short gap between it and the run call.
+    async fn prepare(&mut self) -> TxPrepareResult {
```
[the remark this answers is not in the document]
mattiecnvr: Vestigial organ, removing.

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,179 @@
...
+        } else {
+            TxValidationResult::Retry
+        }
+    }
+
+    fn next_attempt_after(&self) -> Option<Instant>;
+
+    fn submitted(&self) -> bool;
+
+    fn ready_to_be_processed(&self) -> bool {
```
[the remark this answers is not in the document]
mattiecnvr: Agreed

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -0,0 +1,179 @@
...
+            TxPrepareResult::NotReady
+        }
+    }
+
+    /// Submit this operation to the blockchain and report if it was successful
+    /// or not.
+    async fn submit(&mut self) -> TxRunResult;
+
+    async fn validate(&mut self) -> TxValidationResult {
+        // default implementation is basically a no-op
```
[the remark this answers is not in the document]
mattiecnvr: Done

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -129,254 +113,136 @@ impl SerialSubmitter {
...
-    processed_gauge: IntGauge,
-    messages_processed_count: IntCounter,
-
-    /// Private state used to update actual metrics each tick.
-    max_submitted_nonce: u32,
+pub struct SerialSubmitterMetrics {
+    run_queue_length: IntGauge,
+    validation_queue_length: IntGauge,
+
+    txs_prepared: IntCounter,
```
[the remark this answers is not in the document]
mattiecnvr: Updated

rust/agents/relayer/src/msg/metadata/base.rs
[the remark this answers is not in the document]
mattiecnvr: `impl Default for T`

https://doc.rust-lang.org/stable/book/appendix-03-derivable-traits.html?highlight=default#default-for-default-values

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
mattiecnvr: Went with a different option. We do need to have them only return equal if `Eq` also agrees that they are equal from my understanding of the expected invariants.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+        op_try!(critical: self.ctx.origin_gas_payment_enforcer.record_tx_outcome(&self.message, outcome), "recording tx outcome");
+        if outcome.executed {
+            info!(
+                hash=?outcome.txid,
+                "Message successfully processed by transaction"
+            );
+            self.submitted = true;
+            self.reset_attempts();
+            self.next_attempt_after = Some(Instant::now() + Duration::from_secs(CONFIRM_DELAY_S));
+            op_try!(critical: self.record_message_process_success(), "recording message process success");
```
asaj, line 243: As you mentioned on discord, this should be moved to the confirm step

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
+    domain: HyperlaneDomain,
+    mut rx: mpsc::UnboundedReceiver<Box<DynPendingOperation>>,
+    prepare_queue: OpQueue,
+) -> Result<()> {
+    // Pull any messages sent to this submitter
+    while let Some(op) = rx.recv().await {
+        trace!(?op, "Received new operation");
+        // make sure things are getting wired up correctly; if this works in testing it
+        // should also be valid in production.
+        debug_assert_eq!(*op.domain(), domain);
```
asaj, line 134: Maybe we should use this macro for our PendingMessage submission/prepare data checks?

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
-        loop {
-            self.tick().await?;
-            sleep(Duration::from_millis(200)).await;
+    async fn run(self) -> Result<()> {
+        let Self {
+            domain,
+            metrics,
+            rx: rx_prepare,
+        } = self;
+        let prepare_queue: OpQueue = Default::default();
```
asaj, line 95: nit: Consider comments explaining why prepare/confirm are queues, but submit is a channel

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
+            queue.pop()
         };
-
-        match self.process_message(&msg).await {
-            Ok(true) => {
-                info!(msg=%msg.message, "Message processed");
-                self.record_message_process_success(&msg)?;
-                return Ok(());
+        let Some(Reverse(mut op)) = next else {
+            sleep(Duration::from_millis(200)).await;
```
asaj, line 154: nit: Add a comment e.g. // the queue is empty so wait for a little bit before checking again

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
+            }
+            PendingOperationResult::Reprepare => {
+                metrics.ops_failed.inc();
+                prepare_queue.lock().await.push(Reverse(op));
             }
-            Ok(false) => {
-                info!(msg=%msg.message, "Message not processed");
+            PendingOperationResult::Drop => {
+                // not strictly an error, could have already been processed
+                metrics.ops_prepared.inc();
```
asaj, line 177: It feels a little weird to me that we could count this as prepared

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
+                // this send will pause this task if the submitter is not ready to accept yet
+                tx_submit.send(op).await?;
+            }
+            PendingOperationResult::NotReady => {
+                // none of the operations are ready yet, so wait for a little bit
+                prepare_queue.lock().await.push(Reverse(op));
+                sleep(Duration::from_millis(200)).await;
+            }
+            PendingOperationResult::Reprepare => {
```
asaj, line 171: nit: Wdyt of `Failed` as opposed to `Reprepare`?

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
+            }
+            PendingOperationResult::NotReady => {
+                panic!("Pending operation was prepared and therefore must be ready")
+            }
+            PendingOperationResult::Reprepare => {
+                metrics.ops_failed.inc();
+                prepare_queue.lock().await.push(Reverse(op));
+            }
+            PendingOperationResult::Drop => {
+                metrics.ops_submitted.inc();
```
asaj, line 209: Ditto,

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
-        // If the gas payment requirement hasn't been met, move to the next tick.
-        let Some(gas_limit) = self
-            .gas_payment_enforcer
-            .message_meets_gas_payment_requirement(&msg.message, &tx_cost_estimate)
-            .await?
-        else {
-            info!(?tx_cost_estimate, "Gas payment requirement not met yet");
-            return Ok(false);
+        let Some(Reverse(mut op)) = next else {
+            sleep(Duration::from_secs(60)).await;
```
asaj, line 231: nit: Consider adding a comment as to why sleeping so long here makes sense

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -54,329 +55,243 @@ use super::{
...
-    processed_gauge: IntGauge,
-    messages_processed_count: IntCounter,
+#[derive(Debug, Clone)]
+pub struct SerialSubmitterMetrics {
+    prepare_queue_length: IntGauge,
+    confirm_queue_length: IntGauge,
 
-    /// Private state used to update actual metrics each tick.
-    max_submitted_nonce: u32,
+    ops_prepared: IntCounter,
```
asaj, line 263: Maybe add `dropped`?

rust/utils/run-locally/src/main.rs
```diff
@@ -29,16 +34,27 @@ use nix::{
 };
 use tempfile::tempdir;
 
+const RELAYER_KEYS: &[&str] = &[
```
asaj, line 37: nit: Please add comments saying where these keys come from

rust/utils/run-locally/src/main.rs
```diff
@@ -29,16 +34,27 @@ use nix::{
...
+    "8166f546bab6da521a8369cab06c5d2b9e46670292d85c875ee9ec20e84ffb61",
+    "f214f2b2cd398c806f84e317254e0f0b801d0643303237d97a22a48e01628897",
+    "701b615bbdfb9de65240bc28bd21bbc0d996645a3dd57e7b12bc2bdf6f192c82",
+];
+const VALIDATOR_KEYS: &[&str] = &[
```
asaj, line 42: nit: Please add a comment saying these must be consistent with the ism config for test (and please add a corresponding comment to that config)

asaj (commented):
Really excited about these changes :)

Mostly nits, I think the only material thing left is not marking a message as processed until it's confirmed

rust/agents/relayer/src/msg/mod.rs
```diff
@@ -1,8 +1,16 @@
-//! Processor scans DB for new messages and sends relevant messages
-//! over a channel to a submitter, for delivery.
+//! Processor scans DB for new messages and wraps relevant messages as a
+//! `PendingOperation` and then sends it over a channel to a submitter for
+//! delivery.
 //!
-//! A submitter uses some strategy to try to deliver those messages
-//! to the target blockchain.
+//! A submitter uses some strategy to try to execute the pending operations.
+//! Pending operations have two steps for execution, a `prepare` step and a
```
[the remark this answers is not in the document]
mattiecnvr: Updated.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        if !self.is_ready() {
+            trace!("Message is not ready to be submitted yet");
+            return PendingOperationResult::NotReady;
+        }
+
+        // If the message has already been processed, e.g. due to another relayer having
+        // already processed, then mark it as already-processed, and move on to
+        // the next tick.
+        let is_already_delivered = op_try!(
```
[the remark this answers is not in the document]
mattiecnvr: Both ways made sense to me but went for making a slight change to the serial submitter to check if it has been submitted after calling prepare since that was already kinda supported.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+            self.ctx
+                .destination_mailbox
+                .delivered(self.message.id())
+                .await,
+            "checking message delivery status"
+        );
+        if is_already_delivered {
+            debug!("Message already processed");
```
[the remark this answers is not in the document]
mattiecnvr: Made a change, check now

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+            info!("Message processed");
+            self.submitted = true;
+            op_try!(critical: self.ctx.origin_db.mark_nonce_as_processed(self.message.nonce), "recording message as processed");
+            return PendingOperationResult::Drop;
+        }
+
+        // The Mailbox's `recipientIsm` function will revert if
```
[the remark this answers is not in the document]
mattiecnvr: I am not sure, I did not write that comment, just copied it. I think @tkporter did.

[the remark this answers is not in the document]
mattiecnvr: At this point the retrying provider should not be a problem if it reverts

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+        let is_contract = op_try!(
+            provider.is_contract(&self.message.recipient).await,
+            "checking if message recipient is a contract"
+        );
+        if !is_contract {
+            info!(
+                recipient=?self.message.recipient,
+                "Could not fetch metadata: Recipient is not a contract"
+            );
+            return self.on_retry();
```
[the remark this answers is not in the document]
mattiecnvr: Drop here makes sense based on the comments. Just monkey brained that one.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+        // move onto the next tick.
+        let tx_cost_estimate = op_try!(
+            self.ctx
+                .destination_mailbox
+                .process_estimate_costs(&self.message, &metadata)
+                .await,
+            "estimating costs for process call"
+        );
+
+        // If the gas payment requirement hasn't been met, move to the next tick.
```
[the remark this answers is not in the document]
mattiecnvr: I like it in prepare because it means that the minimum amount of work is done during the "mutation" step (submit).

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub destination_mailbox: CachingMailbox,
+    /// Origin chain database to verify gas payments.
+    pub origin_db: HyperlaneDB,
+    /// Used to construct the ISM metadata needed to verify a message from the
+    /// origin.
+    pub metadata_builder: BaseMetadataBuilder,
+    /// Used to determine if messages from the origin have made sufficient gas
+    /// payments.
+    pub origin_gas_payment_enforcer: Arc<GasPaymentEnforcer>,
+    /// Hard limit on transaction gas when submitting a transaction to the
```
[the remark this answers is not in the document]
mattiecnvr: That is configurable.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        self.submission_data = Some(Box::new(SubmissionData {
+            metadata,
+            gas_limit,
+        }));
+        PendingOperationResult::Success
+    }
+
+    /// Returns the message's status. If the message is processed, either by a
```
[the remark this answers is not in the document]
mattiecnvr: Yeah, actually better to not add it since it is documented in the trait better.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+                hash=?outcome.txid,
+                "Transaction attempting to process transaction reverted"
+            );
+            self.on_retry()
+        }
+    }
+
+    async fn confirm(&mut self) -> PendingOperationResult {
+        if !self.submitted {
+            return PendingOperationResult::Reprepare;
```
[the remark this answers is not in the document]
mattiecnvr: A debug assert should be good

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -0,0 +1,377 @@
...
+                PendingOperationResult::NotReady
+            }
+        }
+    }
+
+    fn next_attempt_after(&self) -> Option<Instant> {
+        self.next_attempt_after
+    }
+
+    fn submitted(&self) -> bool {
```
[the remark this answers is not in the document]
mattiecnvr: I now use this more directly

rust/agents/relayer/src/msg/processor.rs
```diff
@@ -7,29 +8,53 @@ use tokio::{
...
+            self.blacklist,
+            self.prover_sync,
+            self.message_nonce
+        )
+    }
+}
+
 impl MessageProcessor {
+    /// The domain this processor is getting messages from.
+    pub fn domain(&self) -> &HyperlaneDomain {
```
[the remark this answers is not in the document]
mattiecnvr: A log uses it; it does feel reasonable to include conceptually.

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: I do not follow

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: Okay, removing

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: I prefer reprepare since it is a little more clear

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: Better to decrease actually now that I think about it

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
mattiecnvr: Actually there is another reason why calling with lag is not great.

If we call with lag and it says that it is not there but it is at latest, we end up in an edge case where we attempt to submit a transaction that will fail because it is a duplicate.

By not using lag and instead moving it to the confirm queue, we can handle it just as we do any other transaction from then on out.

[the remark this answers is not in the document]
asaj: > If we call with lag and it says that it is not there but it is at latest, we end up in an edge case where we attempt to submit a transaction that will fail because it is a duplicate.

Right, this is what I was referring to when I mentioned the risk of submitting unecessarily

> By not using lag and instead moving it to the confirm queue, we can handle it just as we do any other transaction from then on out.

But we aren't moving it to the confirm queue, we are dropping it

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
asaj: I am sure, please fix as a drive-by

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
asaj: Works for me

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
asaj: Works for me

rust/agents/relayer/src/msg/serial_submitter.rs
asaj: Looks like you agree:
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2195#discussion_r1193916648

rust/agents/relayer/src/msg/serial_submitter.rs
asaj: Works for me

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -17,7 +17,13 @@ use super::{
     pending_operation::*,
 };
 
-const CONFIRM_DELAY_S: u64 = 60 * 10;
+const CONFIRM_DELAY: Duration = if cfg!(any(test, feature = "test-utils")) {
+    // Wait 5 seconds between retries in test mode
```
asaj, line 21: Retries? Shouldn't it be `Wait 5 seconds to confirm txs in test mode`

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -160,8 +167,13 @@ async fn prepare_task(
             PendingOperationResult::Success => {
                 debug!(?op, "Operation prepared");
                 metrics.ops_prepared.inc();
-                // this send will pause this task if the submitter is not ready to accept yet
-                tx_submit.send(op).await?;
+                if op.submitted() {
```
asaj, line 170: Tbh I don't love that this is captured in an struct field as opposed to a PendingOperationResult value.

typescript/infra/config/environments/test/multisigIsm.ts
```diff
@@ -1,5 +1,6 @@
 import { ChainMap, ModuleType, MultisigIsmConfig } from '@hyperlane-xyz/sdk';
 
+// the ISM addresses here must line up with the e2e test's validator addresses
```
asaj, line 3: These are validator addresses not ISM addresses

asaj (approved):

asaj:
Approved pending https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2195#discussion_r1195728223

rust/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
mattiecnvr: Not anymore, current code returns `PendingOperationResult::Success` after setting submitted to true, and then that ends up pushing it through to the confirm queue.

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: Ah, I see, yes, added

rust/agents/relayer/src/msg/pending_message.rs
mattiecnvr: Oops

rust/agents/relayer/src/msg/serial_submitter.rs
mattiecnvr: I ended up moving this logic internally so it does not do anything differently here.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -130,11 +130,7 @@ impl PendingOperation for PendingMessage {
         }
 
         // The Mailbox's `recipientIsm` function will revert if
-        // the recipient is not a contract. This can pose issues with
-        // our use of the RetryingProvider, which will continuously retry
-        // the eth_call to the `recipientIsm` function.
-        // As a workaround, we avoid the call entirely if the recipient is
-        // not a contract.
+        // the recipient is not a contract.
```
asaj, line 133: This still isn't true..

asaj (approved):
🚢

Issue #1867, opened 2023-02-22
Audit config shapes


nambrot:

nambrot:
Questions:

- We generally have moved towards standardizing ChainMetadata (1) to be the "chain connection" config, the only odd ball here is the agent config. Can we have standard chain metadata for the most part?
  - Is it easy enough to specify overriden RPC urls?
  - Should index config 8) go into ChainMetadata as well?
- Should chain metadata have a common way of being extended with contracts for both agents and explorer config to have that standard as well. I.e. ` "anvil" => ({ ...chainMetadata, addresses: { mailbox: "0x" }})`
  - Would it make it easier if `hyperlane-deploy` artifacts are keyed by that key (`addresses`) to copy and paste or even include the artifacts file straight?
- How should owner config 13) be handled? Should it just be assumed to be the deployer and have tooling that checks and governs ownership separately?
- Warp route deployments should use the hyperlane-deploy artificats when necessary instead of letting the user specify manually
- Warp route deployment should probably just spit out a tokenList directly (https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2046)

nambrot:
Here is an attempt to harmonize the agent config and existing SDK configs better so that existing chain metadata + contract artifacts can just be used out of the box. Would then also suggest for the explorer to use this. Looking for your feedback @mattiecnvr @jmrossy 

```diff
   "chains": {
     "celo": {
+      // SDK chain metadata schema
       "name": "celo",
-      "domain": 42220,
-      "addresses": {
-        "mailbox": "0x35231d4c2D8B8ADcB5617A638A0c4548684c7C70",
-        "interchainGasPaymaster": "0x6cA0B6D22da47f091B7613223cD4BB03a2d77918",
-        "validatorAnnounce": "0x9bBdef63594D5FFc2f370Fe52115DdFFe97Bc524"
-      },
+      "displayName": "Celo", // optional
+      "chainId": 42220,
+      "blocks": {
+        "reorgPeriod": 0, // the old finality blocks
+      },
+      "publicRpcUrls": [{
+        "http": "https://forno.celo.org"
+      }],
+
+      // SDK contract addresses schema
+      "mailbox": "0x35231d4c2D8B8ADcB5617A638A0c4548684c7C70",
+      "interchainGasPaymaster": "0x6cA0B6D22da47f091B7613223cD4BB03a2d77918",
[...]

PR #2203, opened 2023-05-08
Spec the ChainMetadata changes for #1867

nambrot:
This is just a PR to have diff to discuss for #1867

### DIFF
--- a/typescript/sdk/src/consts/chainMetadata.ts
+++ b/typescript/sdk/src/consts/chainMetadata.ts
@@ -145,7 +145,7 @@ export const ChainMetadataSchema = z.object({
               .positive()
               .optional()
               .describe(
-                'The maximum number of blocks that this RPC supports getting logs for.',
+                'Maximum difference from the latest block for which the RPC can query data',
               ),
           })
           .optional()
--- a/typescript/sdk/src/consts/chainMetadata.ts
+++ b/typescript/sdk/src/consts/chainMetadata.ts
@@ -98,16 +98,16 @@ export const ChainMetadataSchema = z.object({
   name: z
     .string()
     .describe(
-      `The string identifier of the chain, used as the key in dictornaries.`,
+      'The string identifier of the chain, used as the key in dictionaries.',
     ),
   displayName: z
     .string()
     .optional()
-    .describe(`Human-readable name of the chain.`),
+    .describe('Human-readable name of the chain.'),
   displayNameShort: z
     .string()
     .optional()
-    .describe(`Shorter human-readable name of the chain.`),
+    .describe('Shorter human-readable name of the chain.'),
   nativeToken: z
     .object({
       name: z.string(),
@@ -118,11 +118,14 @@ export const ChainMetadataSchema = z.object({
   rpcUrls: z
     .array(
       z.object({
-        http: z.string().url().describe(`The HTTP RPC endpoint URL.`),
+        http: z
+          .string()
+          .url()
+          .describe('The HTTP RPC endpoint URL. Use https whenever possible.'),
         webSocket: z
           .string()
           .optional()
-          .describe(`The WebSocket RPC endpoint URL.`),
+          .describe('The WebSocket RPC endpoint URL.'),
         pagination: z
           .object({
             maxBlockRange: z
@@ -132,11 +135,21 @@ export const ChainMetadataSchema = z.object({
               .describe(
                 'The maximum number of blocks that this RPC supports getting logs for.',
               ),
-            minBlockNumber: z.number().positive().optional(),
-            maxBlockAge: z.number().positive().optional(),
+            minBlockNumber: z
+              .number()
+              .positive()
+              .optional()
+              .describe('The minimum block number that this RPC supports.'),
+            maxBlockAge: z
+              .number()
+              .positive()
+              .optional()
+              .describe(
+                'The maximum number of blocks that this RPC supports getting logs for.',
+              ),
           })
           .optional()
-          .describe(`Pagination options for the RPC endpoint.`),
+          .describe('Pagination options for the RPC endpoint.'),
         retry: z
           .object({
             maxRequests: z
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -87,36 +87,71 @@ export interface RpcPaginationOptions {
...
+    .number()
+    .positive()
+    .optional()
+    .describe(
+      'The domainId of the chain, should generally default to `chainId`. Consumer of `ChainMetadata` should use this value if present, but otherwise fallback to `chainId`.',
+    ),
+  name: z
+    .string()
+    .describe(
+      `The string identifier of the chain, used as the key in dictornaries.`,
```
mattiecnvr, line 101: nit: these should use `'` instead of back ticks

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -130,20 +165,78 @@ export const ChainMetadataSchema = z.object({
...
-    .optional(),
+    .optional()
+    .describe(`Block settings for the chain/deployment.`),
   transactionOverrides: z.object({}).optional(),
   gasCurrencyCoinGeckoId: z.string().optional(),
   gnosisSafeTransactionServiceUrl: z.string().optional(),
   isTestnet: z.boolean().optional(),
 });
 
+export const HyperlaneDeploymentArtifacts = z.object({
```
mattiecnvr, line 197: Not sure if these deployment artifacts are getting merged into the prior schema but I know we talked about having them flattened into the chain object.

mattiecnvr (commented):
Looks good at a first reading with the caveat that I am unsure if the `HyperlaneDeploymentArtifacts` and `AgentChainMetadataSchema` are getting merged into the chain schema or not.

typescript/sdk/src/consts/chainMetadata.ts
nambrot: On line 219 it gets merged into `AgentChainMetadataSchema`

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -87,36 +87,71 @@ export interface RpcPaginationOptions {
...
       decimals: z.number().positive(),
     })
     .optional(),
-  publicRpcUrls: z
+  rpcUrls: z
     .array(
       z.object({
-        http: z.string().url(),
-        webSocket: z.string().optional(),
+        http: z.string().url().describe(`The HTTP RPC endpoint URL.`),
```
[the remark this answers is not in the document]
nambrot: Hmm the problem is that I do think we need to support http without ssl because often times local nodes don't have it IMO. But we can extend the description to recommend HTTPS as much as possible

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -130,20 +165,78 @@ export const ChainMetadataSchema = z.object({
...
   gnosisSafeTransactionServiceUrl: z.string().optional(),
   isTestnet: z.boolean().optional(),
 });
 
+export const HyperlaneDeploymentArtifacts = z.object({
+  mailbox: z.string().describe(`The mailbox address for the chain.`),
+  interchainGasPaymaster: z
+    .string()
+    .describe(`The interchain gas paymaster address for the chain.`),
+  validatorAnnounce: z
```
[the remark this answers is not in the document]
nambrot: Since every hyperlane deployment has it, imo it should be required. To show things like payments and eventually self-relaying, you do need both

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -130,20 +165,78 @@ export const ChainMetadataSchema = z.object({
...
+    .string()
+    .describe(`The validator announce address for the chain.`),
+
+  index: z.object({
```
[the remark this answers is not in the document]
nambrot: Fair enough, I can put that into the AgentChainMetadata shape

typescript/sdk/src/consts/chainMetadata.ts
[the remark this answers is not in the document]
jmrossy: Okay but note, http won't work in the browser (e.g. in the explorer or warp ui) because non-ssl traffic is forbidden in an https context.

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -132,11 +135,21 @@ export const ChainMetadataSchema = z.object({
...
+              .number()
+              .positive()
+              .optional()
+              .describe('The minimum block number that this RPC supports.'),
+            maxBlockAge: z
+              .number()
+              .positive()
+              .optional()
+              .describe(
+                'The maximum number of blocks that this RPC supports getting logs for.',
```
jmrossy, line 148: No sorry, this is the "Maximum difference from the latest block for which the RPC can query data"

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -132,11 +135,21 @@ export const ChainMetadataSchema = z.object({
               .describe(
                 'The maximum number of blocks that this RPC supports getting logs for.',
```
jmrossy, line 136: The maximum range between block numbers for which the RPC can query data

typescript/sdk/src/consts/chainMetadata.ts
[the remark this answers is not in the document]
nambrot: Yeah noted, we can have stricter validation on the explorer side. I mostly care about the shapes being harmonized

typescript/sdk/src/consts/chainMetadata.ts
nambrot: Is it actually in relation to the latest block (like minBlockNumber) or about the window of getting logs? I would have interpreted it closer to the `chunk` parameter

nambrot: The comment on `RpcPaginationOptions` seems to confirm that?

jmrossy: I think you have it backwards. minBlockNumber is absolute and maxBlockAge is relative to latest

jmrossy: That's what the comments say right? Maybe we're saying the same thing here

nambrot: Sorry yes I looked at the wrong line

jmrossy:
Changes discussed here have been added in a while ago

PR #2246, opened 2023-05-12
Refactor agent event indexing

asaj:
### Description

This PR refactors event indexing in the agents, allowing similar logic to be shared across multiple event types (i.e. messages, deliveries, gas payments) and database types (i.e. the Relayer rocks DB and Scraper SQL DB).

Furthermore, it adds new syncing modes by way of `MessageSyncCursors` that take advantage of the monotonically increasing dispatched message nonce to sync more intelligently.

### Drive-by changes

- Fixes a bug in the existing cursor that caused the same block range to be indexed three times
- Modifies kathy to get rid of the idea of "rounds", just sends messages with a sleep in between
- Minor modifications to the e2e test for performance
- Expand macros in settings
- Add scraper to e2e test

### Opportunities for improvement

- We can further reduce RPC usage (or improve latency) by sharing the view of the latest finalized block number between cursors
- We can speed up the effective time for (a relayer to start deliving messages | the scraper to scraper recent events) by creating forward/backward cursors for gas payments and deliveries where the backwards cursor terminates at index_settings.from
- We can remove the need for index_settings.from by terminating backwards cursors based on the block number that the first message was dispatched at

### Related issues

- Fixes #[issue number here]

### Backward compatibility

Yes

None

### Testing

E2E tests


rust/hyperlane-base/src/contract_sync/metrics.rs
```diff
@@ -1,25 +1,26 @@
 use crate::CoreMetrics;
 use prometheus::{IntCounterVec, IntGaugeVec};
-use std::sync::Arc;
 
 /// Struct encapsulating prometheus metrics used by the ContractSync.
 #[derive(Debug, Clone)]
 pub struct ContractSyncMetrics {
+    // TODO: Does this still apply?
```
asaj: Curious for thoughts on what metrics we should keep here? The forwards/backwards style cursor maybe makes the previous idea of a single "indexed height" and "message nonce" per event type not applicable

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -1,31 +1,143 @@
...
+#[derive(Debug, new, Clone)]
+pub struct ContractSync<T, D: HyperlaneDB<T>, I: Indexer<T>> {
     domain: HyperlaneDomain,
-    db: HyperlaneDB,
+    db: D,
     indexer: I,
-    index_settings: IndexSettings,
     metrics: ContractSyncMetrics,
+    // TODO: Why is this necessary?
+    blah: PhantomData<T>,
```
asaj: @mattiecnvr help please, the compiler is otherwise unhappy with
```
pub struct ContractSync<T, D: HyperlaneDB<T>, I: Indexer<T>>
```

I'm using `D` and `I` so that the impl can be aware of exactly what type of indexer or DB it has, which can be useful for creating special cursors.

There's probably a better way that doesn't involve the cursors being made by the ContractSync, but I'd refactored enough at this point that I just wanted to roll with what I had.

rust/agents/scraper/src/agent.rs
```diff
@@ -106,69 +115,127 @@ impl BaseAgent for Scraper {
...
+        let scraper = self.scrapers.get(domain_id).unwrap();
+        let db = scraper.db.clone();
+        let index_settings = scraper.index_settings.clone();
+        let domain = scraper.domain.clone();
+
+        let mut tasks = Vec::with_capacity(2);
+        let message_sync = self
+            .as_ref()
+            .settings
+            .build_message_sync(
```
mattiecnvr, line 179: That builder macro I had reduces a lot of this boilerplate.

rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -349,38 +245,128 @@ impl SqlChainScraper {
...
 }
 
-impl Delivery {
-    fn as_storable(&self, txn_id: i64) -> StorableDelivery {
-        StorableDelivery {
-            message_id: self.message_id,
-            meta: &self.meta,
-            txn_id,
+#[async_trait]
+impl HyperlaneDB<H256> for HyperlaneSqlDb {
```
mattiecnvr, line 277: Seeing `impl HyperlaneDB<H256>` does not make it immediately clear it is storing message deliveries.

rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -349,38 +245,128 @@ impl SqlChainScraper {
...
+                match block_id {
+                    Some(block_id) => {
+                        let block_number = self.db.retrieve_block_number(block_id).await?;
+                        Ok(block_number)
+                    }
+                    None => Ok(None),
+                }
+            }
+            None => Ok(None),
         }
```
mattiecnvr, line 350:
```suggestion
        let Some(tx_id) = self
            .db
            .retrieve_dispatched_tx_id(self.domain().id(), &self.mailbox_address, nonce)
            .await?
        else {
            return Ok(None);
        };
        
        let Some(block_id) = self.db.retrieve_block_id(tx_id).await? else {
            return Ok(None);
        };
        
        Ok(self.db.retrieve_block_number(block_id).await?)
```

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -98,7 +356,15 @@ impl<I: Indexer> SyncBlockRangeCursor for RateLimitedSyncBlockRangeCursor<I> {
         Ok((from, to, eta))
     }
 
-    fn backtrack(&mut self, start_from: u32) {
-        self.from = u32::min(start_from, self.from);
+    async fn update(&mut self, _: Vec<(T, LogMeta)>) -> eyre::Result<()> {
+        // Store a relatively conservative view of the high watermark, which should allow a single watermark to be
+        // safely shared across multiple cursors, so long as they are running sufficiently in sync
```
mattiecnvr, line 361: What happens if they don't run in sync?

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+                            return Ok(range);
+                        }
+                        None => {
+                            // TODO: Define the sleep time from interval flag
+                            sleep(Duration::from_secs(5)).await;
+                        }
+                    }
+                }
+            }
+        }
```
mattiecnvr, line 261:
```suggestion
        loop {
            if let Some(forward_range) = self.forward.get_next_range().await? {
                self.direction = SyncDirection::Forward;
                return Ok(forward_range);
            }
            
            if let Some(backward_range) = self.backward.get_next_range().await {
                self.direction = SyncDirection::Backward;
                return Ok(range);
            }
            
            // TODO: Define the sleep time from interval flag
            sleep(Duration::from_secs(5)).await;
        }
```

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+                self.cursor.next_block = block_number;
+                // If we found nonce zero, we are done rewinding.
+                if self.cursor.next_nonce == 0 {
+                    self.synced = true;
+                    break;
+                }
+                self.cursor.next_nonce = self.cursor.next_nonce.saturating_sub(1);
+            } else {
+                break;
+            }
```
mattiecnvr, line 165:
```suggestion
            let Some(block_number) = self
                .cursor
                .retrieve_dispatched_block_number(self.cursor.next_nonce)
                .await
            else { break };
            
            self.cursor.next_block = block_number;
            // If we found nonce zero, we are done rewinding.
            if self.cursor.next_nonce == 0 {
                self.synced = true;
                break;
            }
            self.cursor.next_nonce = self.cursor.next_nonce.saturating_sub(1);
```

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+        if self.synced {
+            None
+        } else {
+            // Just keep going backwards.
+            let to = self.cursor.next_block;
+            let from = to.saturating_sub(self.cursor.chunk_size);
+            self.cursor.next_block = from.saturating_sub(1);
+            // TODO: Consider returning a proper ETA for the backwards pass
+            Some((from, to, Duration::from_secs(0)))
+        }
```
mattiecnvr, line 176:
```suggestion
        if self.synced {
            return None;
        }
        
        // Just keep going backwards.
        let to = self.cursor.next_block;
        let from = to.saturating_sub(self.cursor.chunk_size);
        self.cursor.next_block = from.saturating_sub(1);
        // TODO: Consider returning a proper ETA for the backwards pass
        Some((from, to, Duration::from_secs(0)))
```

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+        indexer: Arc<dyn MessageIndexer>,
+        db: Arc<dyn HyperlaneMessageDB>,
+        chunk_size: u32,
+    ) -> Result<Self> {
+        let (count, tip) = indexer.fetch_count_at_tip().await?;
+        let forward_cursor = ForwardMessageSyncCursor::new(MessageSyncCursor::new(
+            indexer.clone(),
+            db.clone(),
+            chunk_size,
+            tip,
```
mattiecnvr, line 212: Tip twice feels odd to me

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+                // We attempt to index a range of blocks that is as large as possible.
+                let from = self.0.next_block;
+                let to = u32::min(tip, from + self.0.chunk_size);
+                self.0.next_block = to + 1;
+                Ok(Some((from, to, Duration::from_secs(0))))
+            }
+            Ordering::Greater => {
+                panic!("Cursor is ahead of mailbox, this should never happen.");
+            }
+        }
```
mattiecnvr, line 109:
```suggestion
        if cursor_count == mailbox_count {
            // We are synced up to the latest nonce so we don't need to index anything.
            // We update our next block number accordingly.
            self.0.next_block = tip;
            Ok(None)
        } else if cursor_count < mailbox_count {
            // The cursor is behind the mailbox, so we need to index some blocks.
            // We attempt to index a range of blocks that is as large as possible.
            let from = self.0.next_block;
            let to = u32::min(tip, from + self.0.chunk_size);
            self.0.next_block = to + 1;
            Ok(Some((from, to, Duration::from_secs(0))))
        } else {
            panic!("Cursor is ahead of mailbox, this should never happen.");
        }
```

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -1,90 +1,348 @@
...
+            let range = self.get_next_range().await?;
+            match range {
+                Some(range) => {
+                    return Ok(range);
+                }
+                None => {
+                    // TODO: Define the sleep time from interval flag
+                    sleep(Duration::from_secs(5)).await;
+                }
+            }
```
mattiecnvr, line 126:
```suggestion
            if let Some(range) = self.get_next_range().await? = {
                return Ok(range)
            }
            
            // TODO: Define the sleep time from interval flag
            sleep(Duration::from_secs(5)).await;
```

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -1,31 +1,140 @@
...
+    pub async fn forward_message_sync_cursor(
+        &self,
+        index_settings: IndexSettings,
+    ) -> Box<dyn ContractSyncCursor<HyperlaneMessage>> {
+        let forward_data = MessageSyncCursor::new(
+            self.indexer.clone(),
+            self.db.clone(),
+            index_settings.chunk_size,
+            index_settings.from,
+            index_settings.from,
```
mattiecnvr, line 119: Same value twice again?

rust/hyperlane-core/src/traits/db.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Gets the block number at which a known message was dispatched.
+    async fn retrieve_dispatched_block_number(&self, nonce: u32) -> Result<Option<u64>>;
+}
+
+/// Extension of HyperlaneDB trait that supports a high watermark for the highest indexed block number.
+#[async_trait]
+#[auto_impl(&, Box, Arc)]
+pub trait HyperlaneHighWatermarkDB<T>: HyperlaneDB<T> + Sync + Send + Debug + 'static {
+    /// Gets the block number high watermark
+    async fn retrieve_high_watermark(&self) -> Result<Option<u32>>;
```
mattiecnvr, line 33: Would it make more sense to return `0` instead of an Option when unknown.

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -125,13 +140,46 @@ where
...
+#[async_trait]
+impl<M> MessageIndexer for EthereumMailboxIndexer<M>
+where
+    M: Middleware + 'static,
+{
+    #[instrument(err, skip(self))]
+    async fn fetch_count_at_tip(&self) -> ChainResult<(u32, u32)> {
+        let tip =
+            <EthereumMailboxIndexer<M> as Indexer<HyperlaneMessage>>::get_finalized_block_number::<
+                '_,
```
mattiecnvr, line 154: Can you not omit the `'_` lifetime?

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -125,13 +140,46 @@ where
...
+                '_,
+            >(self)
+            .await?;
```
mattiecnvr, line 157: Can this be done as
```rs
Indexer<HyperlaneMessage>::get_finalized_block_number(self as _)
```

instead of using the turbofish?

rust/agents/scraper/src/db/block.rs
```diff
@@ -35,6 +35,26 @@ impl FromQueryResult for BasicBlock {
 }
 
 impl ScraperDb {
+    // TODO: Is this correct?
+    pub async fn retrieve_block_number(&self, block_id: i64) -> Result<Option<u64>> {
```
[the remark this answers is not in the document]
mattiecnvr: I think so

rust/agents/scraper/src/db/message.rs
```diff
@@ -61,6 +61,33 @@ impl ScraperDb {
...
+        origin_mailbox: &H256,
+        nonce: u32,
+    ) -> Result<Option<i64>> {
+        #[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
+        enum QueryAs {
+            Nonce,
+        }
+
+        // TODO: Is this correct?
+        let tx_id = message::Entity::find()
```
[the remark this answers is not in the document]
mattiecnvr: Might be able to do a little better by taking the max of the Nonce column instead.

https://www.sea-ql.org/SeaORM/docs/advanced-query/aggregate-function/#group-by

rust/agents/scraper/src/db/txn.rs
```diff
@@ -19,6 +19,22 @@ pub struct StorableTxn {
 }
 
 impl ScraperDb {
+    pub async fn retrieve_block_id(&self, tx_id: i64) -> Result<Option<i64>> {
```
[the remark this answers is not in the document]
mattiecnvr: Should be

mattiecnvr (commented):
There's a lot of changes here; we definitely are going to need to test the scraper before I am comfortable with this PR getting merged.

mattiecnvr (commented):
There's a lot of changes here; we definitely are going to need to test the scraper before I am comfortable with this PR getting merged.

rust/hyperlane-base/src/contract_sync/metrics.rs
mattiecnvr: We could just have the current cursor values

rust/hyperlane-base/src/contract_sync/mod.rs
mattiecnvr: You could move the T type into the function generics instead. The reason you would need `PhatomData` generically speaking is if you have a type that is not actually represented in the trait/struct.

.github/workflows/e2e.yml
```diff
@@ -60,4 +60,4 @@ jobs:
         env:
           E2E_CI_MODE: 'true'
           E2E_CI_TIMEOUT_SEC: '600'
-          E2E_KATHY_ROUNDS: '2'
+          E2E_KATHY_MESSAGES: '10'
```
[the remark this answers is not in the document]
asaj: Well we do it twice, once before we spin up the relayer and once after

rust/hyperlane-core/src/traits/db.rs
```diff
@@ -0,0 +1,36 @@
...
+use async_trait::async_trait;
+use auto_impl::auto_impl;
+use eyre::Result;
+
+use crate::{HyperlaneMessage, LogMeta};
+
+/// Interface for a HyperlaneDB that ingests logs.
+#[async_trait]
+#[auto_impl(&, Box, Arc)]
+pub trait HyperlaneDB<T>: Send + Sync + Debug + 'static {
```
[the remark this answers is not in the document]
asaj: I figured if/when we have two H256s we want to store, we would wrap them in structs

rust/hyperlane-core/src/traits/db.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Store a list of logs and their associated metadata
+    /// Returns the number of elements that were stored.
+    async fn store_logs(&self, logs: &[(T, LogMeta)]) -> Result<u32>;
```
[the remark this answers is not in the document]
asaj: Yeah duplicates. I don't know how to detect this in the sql db

rust/hyperlane-core/src/traits/db.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// Extension of HyperlaneDB trait that supports getting the block number at which a known message was dispatched.
+#[async_trait]
+#[auto_impl(&, Box, Arc)]
+pub trait HyperlaneMessageDB:
```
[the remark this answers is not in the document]
asaj: Yeah, the idea was the minimal set for the abstractions I was looking to share between the two

rust/hyperlane-base/src/db/mod.rs
```diff
@@ -1,106 +1,2 @@
...
-        Ok(self.0.put(key, value)?)
-    }
-
-    /// Retrieve a value from the DB
-    pub fn retrieve(&self, key: &[u8]) -> Result<Option<Vec<u8>>> {
-        Ok(self.0.get(key)?)
-    }
-}
+pub use rocks::*;
+mod rocks;
```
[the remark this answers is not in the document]
asaj: Yeah I was thinking I was going to move the SQL implementation and then I got spooked

rust/hyperlane-core/src/traits/db.rs
asaj: I think an option is a clearer disambiguation of unknown

asaj:
> There's a lot of changes here; we definitely are going to need to test the scraper before I am comfortable with this PR getting merged.

How would you suggest testing? Do we have an RC deployment?

rust/hyperlane-base/src/contract_sync/mod.rs
asaj: How would that work? I assume I needed it in the struct so as to restrict the D/I types.

rust/agents/scraper/src/db/message.rs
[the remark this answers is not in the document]
asaj: I with very low confidence think I did it?

rust/hyperlane-base/src/contract_sync/metrics.rs
asaj: Sure but the ForwardBackward cursor jumps all over the place.

We *could* put the metrics in the cursor, and have it handle it, and the forward/backward cursor could use different labels

rust/utils/run-locally/src/main.rs
```diff
@@ -389,9 +411,7 @@ fn main() -> ExitCode {
         if ci_mode {
             // for CI we have to look for the end condition.
             if kathy_done && retry_queues_empty() {
-                assert_termination_invariants(
-                    kathy_rounds.unwrap() as u32 * kathy_messages_per_round,
-                );
+                assert_termination_invariants(kathy_messages.unwrap() as u32 * 2);
```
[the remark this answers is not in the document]
asaj: What would you suggest? I send one set before and one after.

rust/utils/run-locally/src/main.rs
```diff
@@ -288,6 +281,61 @@ fn main() -> ExitCode {
...
+        }
+    }));
+    state.validator = Some(validator);
+
+    // Run one round of kathy before the relayer comeSigned checkpoints s up
+    println!("Spawning Kathy to send Hyperlane message traffic...");
+    let mut kathy = Command::new("yarn");
+    kathy.arg("kathy");
+    if let Some(r) = kathy_messages {
+        kathy.args(["--messages", &r.to_string(), "--timeout", "1000"]);
```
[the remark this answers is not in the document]
asaj: Yeah I'm gonna axe it

rust/hyperlane-core/src/traits/db.rs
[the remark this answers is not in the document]
asaj: I get:
> the trait cannot be made into an object because method `store_logs` has generic type parameters

rust/agents/relayer/src/relayer.rs
```diff
@@ -33,15 +34,17 @@ use crate::{
...
+    origin_interchain_gas_payment_sync: Arc<WatermarkContractSync<InterchainGasPayment>>,
+    origin_validator_announce: Arc<dyn ValidatorAnnounce>,
+    destination_mailboxes: HashMap<HyperlaneDomain, Arc<dyn Mailbox>>,
     gas_payment_enforcer: Arc<GasPaymentEnforcer>,
     whitelist: Arc<MatchingList>,
     blacklist: Arc<MatchingList>,
     transaction_gas_limit: Option<U256>,
     skip_transaction_gas_limit_for: HashSet<u32>,
     allow_local_checkpoint_syncers: bool,
+    origin_db: Arc<HyperlaneRocksDB>,
```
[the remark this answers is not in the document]
asaj: Indeed it is unecessary

rust/agents/scraper/src/agent.rs
asaj: Created a similar macro

rust/agents/scraper/src/chain_scraper/mod.rs
asaj: What would you recommend? type MessageID = H256?

rust/hyperlane-base/src/contract_sync/mod.rs
asaj: Ditto

rust/hyperlane-base/src/contract_sync/cursor.rs
asaj: If two tasks share a high watermark, then task A may write a watermark that is too high for task B.

The scraper shares a watermark between the delivery and gas payment sync tasks. I didn't want to change the DB schema.

rust/hyperlane-base/src/contract_sync/cursor.rs
asaj: Clippy prefers the match statement

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -63,98 +66,50 @@ impl HyperlaneDB {
...
-        debug!(nonce, ?id, "storing leaf hash keyed by index");
-        self.store_keyed_encodable(MESSAGE_ID, &nonce, &id)?;
-        self.update_latest_nonce(nonce)?;
-        self.update_latest_nonce_for_destination(destination, nonce)
+        // - `id` --> `message`
+        self.store_keyed_encodable(MESSAGE, &id, message)?;
+        // - `nonce` --> `id`
+        self.store_keyed_encodable(MESSAGE_ID, &message.nonce, &id)?;
+        // - `nonce` --> `dispatched block number`
+        self.store_keyed_encodable(
```
tkporter, line 93: Iiuc if people don't wipe the relayer db before running this new version, all messages that have already been indexed won't have an entry for this. Is that ok?

Or is the assumption that this will roll out with the omniscient relayer stuff, which would mean that all messages are re-indexed anyways? (<-- i don't know if this is true, I didn't read the PR)

tkporter (approved):
LGTM

Re: alternate execution environments where we may be polling per message nonce or an IGP payment sequence number instead of looking at block ranges, how were you thinking that could fit into this interface?

rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -349,35 +245,121 @@ impl SqlChainScraper {
...
+        self.db
+            .store_dispatched_messages(&self.mailbox_address, storable)
+            .await?;
+
+        Ok(messages.len() as u32)
+    }
+}
+
+#[async_trait]
+impl HyperlaneLogStore<H256> for HyperlaneSqlDb {
```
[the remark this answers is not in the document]
asaj: Does an alias work for you?

rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -349,35 +245,121 @@ impl SqlChainScraper {
     }
 }
 
-#[derive(Debug, Clone)]
-struct Delivery {
-    message_id: H256,
-    meta: LogMeta,
+#[async_trait]
```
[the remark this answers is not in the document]
asaj: Unfortunately I tried and I think they're different enough that it's probably not worth the effort rn

rust/agents/scraper/src/db/block.rs
```diff
@@ -35,6 +35,24 @@ impl FromQueryResult for BasicBlock {
 }
 
 impl ScraperDb {
+    pub async fn retrieve_block_number(&self, block_id: i64) -> Result<Option<u64>> {
```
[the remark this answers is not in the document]
asaj: Yeah we should be consistent with height and number, but unfortunately we're not, they're used pretty interchangeably throughout the codebase.

Added a comment

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -1,31 +1,151 @@
...
```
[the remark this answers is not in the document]
asaj: We write the block number that each message was dispatched at, and use that to fast forward the cursor.

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
asaj: It's kind of okay? The cursor currently uses this entry to detect the presence of a message, which means it will attempt to reindex all messages, which I suppose isn't great.

I think we can work around this by having it instead get messages by nonce, and if present, attempt to grab the corresponding block number. If that's found, it will use that block number to fast forward, but otherwise it will fast forward the nonce only (and not the block).

asaj: Okay, this is what I did, it does result in an extra DB read but it's technically more correct I guess

Though I do actually assume that the relayers will need to reindex anyways, and the scraper would have been able to fast forward anyways (because it already stores the dispatched block number).

Lmk wyt

rust/agents/scraper/src/chain_scraper/mod.rs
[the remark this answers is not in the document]
tkporter: yeah definitely, that's way better

rust/agents/scraper/src/db/message.rs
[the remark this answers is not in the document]
mattiecnvr: I think more like this:
```rs
let tx_id = message::Entity::find()
    .filter(message::Column::Origin.eq(origin_domain))
    .filter(message::Column::OriginMailbox.eq(address_to_bytes(origin_mailbox)))
    .filter(message::Column::Nonce.eq(nonce))
    .select_only()
    .column_as(message::Column::OriginTxId.max(), QueryAs::Nonce)
    .group_by(message::Column::Origin)
    .into_values::<i64, QueryAs>()
    .one(&self.0)
    .await?
```

I would have to test it to be certain

rust/hyperlane-base/src/contract_sync/metrics.rs
mattiecnvr: That would make sense to me

.github/workflows/e2e.yml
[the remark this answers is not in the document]
mattiecnvr: I would just like the "KATHY_MESSAGES" var to be the number of messages sent in total.

rust/hyperlane-core/src/traits/db.rs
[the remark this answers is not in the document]
mattiecnvr: I suspect there's a way to do it using [exec_with_returning](https://docs.rs/sea-orm/latest/sea_orm/query/struct.Insert.html#method.exec_with_returning). I guess the question is do we rely on this answer? If not I would rather just see a `Result<()>` and we can add it later.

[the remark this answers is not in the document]
mattiecnvr: Oh, yeah, this should work then
```suggestion
    async fn store_logs<I: Iterator<Item=(T, LogMeta)>>(&self, logs: I) -> Result<u32>;
```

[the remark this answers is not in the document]
mattiecnvr: Oh, nvm then. This is fine for now.

rust/hyperlane-base/src/db/mod.rs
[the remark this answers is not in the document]
mattiecnvr: I really really do not want the SQL impl in base unless we really need it to be. That adds a lot of deps because of sea-orm and then all the codegen and crap.

rust/agents/scraper/src/chain_scraper/mod.rs
mattiecnvr: I think a (private?) type of `type DeliveredMessageId = H256` would work.

rust/hyperlane-base/src/contract_sync/cursor.rs
mattiecnvr: I would rather we define `new` by hand because then the knowledge it needs to be duplicated is handled next to the thing that requires knowing it. `new` is just a convenience for certain cases.

rust/hyperlane-base/src/contract_sync/cursor.rs
mattiecnvr: What!? That looks so odd to me.

Funny, seems a few people agree with me on this:
https://github.com/rust-lang/rust-clippy/issues/4531

rust/hyperlane-core/src/traits/db.rs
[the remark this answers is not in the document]
asaj: I did this yesterday

rust/hyperlane-core/src/traits/db.rs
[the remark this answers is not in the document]
asaj: We use it for some metrics

rust/agents/scraper/src/chain_scraper/mod.rs
asaj: I did the former

rust/hyperlane-base/src/contract_sync/cursor.rs
asaj: The amount of code that we'd write vs the amount of knowledge that we'd prevent callers from needing to know doesn't feel like a winning tradeoff to me?

rust/agents/scraper/src/db/message.rs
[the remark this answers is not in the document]
asaj: What's the best way to test it?

rust/hyperlane-core/src/traits/db.rs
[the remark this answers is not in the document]
asaj: I modified the store fns to query the count before and after storing

rust/hyperlane-base/src/contract_sync/metrics.rs
asaj: I updated it to monotonically increase instead. So we will just not track the "backwards" part of the forward/backward sync, for now

PR #2248, opened 2023-05-15
Adapt multisig metadata builder to 3 variants

yorhodes:
### Description

Supports 2 new multisig ISM variants from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2173 in relayer.

### Drive-by changes

- migrates relayer keys to be anvil accounts
- moves from integer to enum module types

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2221

### Backward compatibility

Yes

### Testing

E2e tests


rust/agents/relayer/Cargo.toml
```diff
@@ -33,6 +33,7 @@ hyperlane-base = { path = "../../hyperlane-base" }
 hyperlane-ethereum = { path = "../../chains/hyperlane-ethereum" }
 num-derive.workspace = true
 num-traits.workspace = true
+derive_more = "0.99.17"
```
mattiecnvr, line 36: If we specify the patch version it won't auto update for security/bug fixes.
```suggestion
derive_more = "0.99"
```

rust/chains/hyperlane-ethereum/src/interchain_security_module.rs
```diff
@@ -96,9 +97,9 @@ where
     M: Middleware + 'static,
 {
     #[instrument(err, ret)]
-    async fn module_type(&self) -> ChainResult<u8> {
+    async fn module_type(&self) -> ChainResult<ModuleType> {
         let module_type = self.contract.module_type().call().await?;
-        Ok(module_type)
+        Ok(ModuleType::from_u8(module_type).unwrap_or_default())
```
mattiecnvr, line 102: Do we want to log if we encounter an unknown module type? (Maybe you already do elsewhere?)

mattiecnvr (approved):

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -144,7 +134,29 @@ impl BaseMetadataBuilder {
...
+                validators,
+                threshold,
+                message.nonce,
+                highest_known_nonce,
+            )
+            .await
+            .context(CTX)
+    }
+
+    pub async fn fetch_checkpoint(
```
asaj, line 147: You need two different versions of this function for censorship resistance.

For merkle proof ISMs, you can use *any* checkpoint with nonce >= message nonce. You will need something similar to `fetch_checkpoint_in_range`

For message ID ISMs, you need the specific checkpoint that matches the message nonce, which is what you are doing here.

asaj (changes requested):

asaj:
Also I'd imagine there's a lot of code that can be saved by using generics?

rust/chains/hyperlane-ethereum/src/interchain_security_module.rs
asaj: Yeah I think debug or maybe warn

PR #2249, opened 2023-05-16
Treat chain names as lowercase

### LINKED ISSUE #2103: Agents cannot handle mixed case chain names
Seemingly, at least

https://discord.com/channels/935678348330434570/1096444620172431410/1096800853882572860

mattiecnvr:
### Description

Had to do a little digging but I think this should fix not only the issue that was encountered but fully add support for whatever case they want, it will just treat it as lowercase but still display it correctly for their config paths.

The problem was that `HyperlaneDomain` was already lowercasing the name but we didn't lowercase the key that we used in `from_config_filtered` for the base `RawSettings` to look it up during parsing so we couldn't find the values.

### Drive-by changes

None

### Related issues

- Fixes #2103 

### Backward compatibility

Yes

None

### Testing

None


rust/hyperlane-base/src/settings/base.rs
```diff
@@ -249,7 +249,7 @@ impl Settings {
     /// Try to get the domain for a given chain by name.
     pub fn lookup_domain(&self, chain_name: &str) -> eyre::Result<HyperlaneDomain> {
         self.chains
-            .get(chain_name)
+            .get(&chain_name.to_ascii_lowercase())
```
asaj, line 252: Where is this used? I'm surprised this is necessary

mattiecnvr: Might be easiest to answer with a search result
https://github.com/search?q=repo%3Ahyperlane-xyz%2Fhyperlane-monorepo%20lookup_domain&type=code

This change is necessary to fix the ticket.

nambrot:
@mattiecnvr should we merge this

rust/hyperlane-base/src/settings/base.rs
mattiecnvr: That would make more sense. Changed.

rust/agents/relayer/src/settings/mod.rs
```diff
@@ -246,7 +246,11 @@ impl FromRawConf<'_, RawRelayerSettings> for RelayerSettings {
             #[allow(deprecated)]
             raw.originchainname
         }
-        .map(|s| s.split(',').map(str::to_owned).collect::<Vec<_>>());
+        .map(|s| {
```
asaj, line 249: nit: consider adding a parseChains fn or something to dedupe

asaj (approved):

mattiecnvr:
oops, I did make a commit to dedup but forgot auto merge was on. Will followup with a small PR.

PR #2256, opened 2023-05-17
Support 64-byte txn ids

### LINKED ISSUE #2252: Sealevel agent: TxOutcome should support 64 byte tx ids
On Solana, transaction IDs are 64 bytes (this is because tx IDs are actually the *signature* over a transaction!) instead of 32 bytes. `TxOutcome` at the moment expects the tx id to be 32 bytes. We can get away with hashing the 64 byte Solana tx id to be 32 bytes for the time being, but this sucks because now the original transaction ID can't get recovered. So we should move over to 64 byte transaction IDs (or at least when they're needed)

Update: a version was done, but due to other changes needs to be reworked now

mattiecnvr:
### Description

Adds support to TxOutcome for 512-bit ids. I wanted to just use `.into()` and always store it as a H512 but doing that was not supported and would also create possible issues when displaying it so this is what I came up with.

### Drive-by changes

Explored using a similar solution for Addresses and realized that we had some dead code hanging out so removed that.

### Related issues

- Fixes #2252 

### Backward compatibility

Yes

None

### Testing

None


asaj (changes requested):
LogMeta too, no? Not sure where else we have tx hashes

mattiecnvr:
> LogMeta too, no? Not sure where else we have tx hashes

Before I look at this, I want an answer from @tkporter as to whether he likes this approach at all.

mattiecnvr:
Copying some discussion points:

> We have a number of places where something might be one of a few sizes so it feels weird to me to only do it for this one case which is part of what I don't like about it.
>  But the into/from support for H512 and H256 is not as good as it is between H160 and H256 so the alternative would be to wrap the H512 and H256 types to extend them with more conversions which might also fix some other things but would also possibly just be a pain for everyone.

rust/hyperlane-core/src/types/mod.rs
```diff
@@ -20,6 +22,50 @@ mod message;
...
+}
+
+impl Display for InvalidType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Incorrect internal type")
+    }
+}
+
+/// Quickly create a new primitive type wrapping enum
+macro_rules! define_primitive_hash_enum {
```
tkporter, line 43: Wanna flag that (iiuc) there's some similar logic that's included in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2505 now that we'll be moving to some types defined within hyperlane-core

Because this PR isn't super critical I'd recommend making this PR to whatever branch https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2505 ends up in (i.e. either wait for that PR to merge into `trevor/sealevel-validator-rebase` and make this against `trevor/sealevel-validator-rebase`, or if `trevor/sealevel-validator-rebase` is merged into main by then to put this PR against main)

rust/hyperlane-core/src/traits/mod.rs
```diff
@@ -30,7 +30,7 @@ mod validator_announce;
 #[derive(Debug, Clone, Copy)]
 pub struct TxOutcome {
     /// The txid
-    pub txid: crate::H256,
+    pub txid: crate::TxId,
```
tkporter, line 33: Afaict there are no DB implications to making this change (`record_tx_outcome` only stores a `InterchainGasExpenditure`), just want to double check this is the case

tkporter (commented):
+1 to changing LogMeta's transaction_hash and doing a scan for other places transaction hashes are used

I also wanna flag that a change to LogMeta's transaction_hash to using `TxId` would have DB implications.. Unsure if there's a nice way to preserve backward compatibility / migrate the db / just force reindexing of logs or what

mattiecnvr:
> +1 to changing LogMeta's transaction_hash and doing a scan for other places transaction hashes are used
> 
> I also wanna flag that a change to LogMeta's transaction_hash to using `TxId` would have DB implications.. Unsure if there's a nice way to preserve backward compatibility / migrate the db / just force reindexing of logs or what

We should be able to migrate the DB if needed but it will take some coordination.

> Because this PR isn't super critical I'd recommend making this PR to whatever branch https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2505 ends up in (i.e. either wait for that PR to merge into trevor/sealevel-validator-rebase and make this against trevor/sealevel-validator-rebase, or if trevor/sealevel-validator-rebase is merged into main by then to put this PR against main)

Okay, I can take another stab at this soon to better incorporate the other changes. Curious what your timeline is on needing it?

mattiecnvr:
Replaced by #2612

PR #2268, opened 2023-05-22
Properly store gas payment expenditure data in agent rocks db

asaj:
### Description

This PR fixes a bug that resulted in gas payment expenditures not being properly stored in the rocks DB, making them unretrievable.

### Drive-by changes

- Introduces a macro to HyperlaneRocksDB to prevent similar bugs from happening in the future

### DIFF
--- a/rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
+++ b/rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
@@ -26,7 +26,7 @@ const MESSAGE: &str = "message_";
 const NONCE_PROCESSED: &str = "nonce_processed_";
 const GAS_PAYMENT_FOR_MESSAGE_ID: &str = "gas_payment_for_message_id_v2_";
 const GAS_PAYMENT_META_PROCESSED: &str = "gas_payment_meta_processed_v2_";
-const GAS_EXPENDITURE_FOR_MESSAGE_ID: &str = "gas_expenditure_for_message_id_";
+const GAS_EXPENDITURE_FOR_MESSAGE_ID: &str = "gas_expenditure_for_message_id_v2_";
 const LATEST_INDEXED_GAS_PAYMENT_BLOCK: &str = "latest_indexed_gas_payment_block";
 
 type DbResult<T> = std::result::Result<T, DbError>;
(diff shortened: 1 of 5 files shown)


rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -164,10 +140,10 @@ impl HyperlaneRocksDB {
             return Ok(false);
         }
         // Set the gas payment as processed
-        self.store_gas_payment_meta_processed(&payment_meta)?;
+        self.store_processed_by_gas_payment_meta(&payment_meta, &true)?;
```
mattiecnvr, line 143: Drive by request: fix `store_processed_by_gas_payment_meta` to take `bool` instead of `&bool`

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -320,3 +271,49 @@ where
         Ok(result)
     }
 }
+
+/// Generate a call to ChainSetup for the given builder
+macro_rules! build_store_and_retrieve {
+    ($vis:vis, $setter:ident, $getter:ident, $prefix: ident, $key: ty, $val: ty) => {
```
mattiecnvr, line 277: nit: reformat as:
```rs
($($vis:vis)? $name_base:ident, $key_prefix: ident, $key_ty:ty, $val_ty:ty)
```

Then you can use

```rs
build_store_and_retrieve!(pub message_id_by_nonce, MESSAGE_ID, u32, H256);
```

To do this you will need to use `paste` for the store/retrieve but this will prevent needing to type it twice and also enforce the prefixes.

Further it is better to have the macro write the `impl HyperlaneRocksDB` for each since the macro is scoped outside the impl. That way the restriction to have the macro call is not within the impl block. (Yes, this means there will be multiple impl blocks generated, this is not an issue.)

mattiecnvr (approved):
I like the new macro for this, makes it a lot cleaner.

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -320,3 +271,49 @@ where
         Ok(result)
     }
 }
+
+/// Generate a call to ChainSetup for the given builder
+macro_rules! build_store_and_retrieve {
```
mattiecnvr, line 276: Oh, also instead of `build_store_and_retrieve` I think `make_store_and_retrieve` would be better. Builder implies a runtime builder whereas this is a compile-time thing.

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
asaj: Ah nice, didn't know about paste

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
asaj: I'm not sure there's an easy way to do this?

PR #2269, opened 2023-05-22
Fix CI flakiness

asaj:
### Description

- Don't panic if E2E test fails to fetch a metric
- Run coverage commands sequentially 

### Drive-by changes

- Cleans up some lint errors in forge tests

### DIFF
--- a/rust/utils/run-locally/src/main.rs
+++ b/rust/utils/run-locally/src/main.rs
@@ -26,6 +26,7 @@ use std::{
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
+use eyre::Result;
 use maplit::hashmap;
 use nix::{
     libc::pid_t,
@@ -442,7 +443,7 @@ fn main() -> ExitCode {
         if ci_mode {
             // for CI we have to look for the end condition.
             let num_messages_expected = (kathy_messages / 2) as u32 * 2;
-            if kathy_done && termination_invariants_met(num_messages_expected) {
+            if kathy_done && termination_invariants_met(num_messages_expected).unwrap_or(false) {
                 // end condition reached successfully
                 println!("Kathy completed successfully and agent metrics look healthy");
                 break;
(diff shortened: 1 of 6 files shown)


solidity/coverage.sh
```diff
@@ -1,40 +1,40 @@
 set -e # exit on error
 
 # generates lcov.info
-forge coverage --report lcov
+forge coverage --report lcov && \
```
yorhodes, line 4: I don't really understand why this is necessary and iirc it suppresses logs or something annoying
the `set -e` should exit if any command fails

rust/utils/run-locally/src/main.rs
```diff
@@ -476,38 +476,38 @@ fn fetch_metric(port: &str, metric: &str, labels: &HashMap<&str, &str>) -> Vec<u
...
         .map(|l| l.rsplit_once(' ').unwrap().1.parse::<u32>().unwrap())
-        .collect()
+        .collect())
 }
 
 /// Use the metrics to check if the relayer queues are empty and the expected
 /// number of messages have been sent.
-fn termination_invariants_met(num_expected_messages: u32) -> bool {
-    let lengths = fetch_metric("9092", "hyperlane_submitter_queue_length", &hashmap! {});
+fn termination_invariants_met(num_expected_messages: u32) -> eyre::Result<bool> {
```
mattiecnvr, line 484: nit
```suggestion
fn termination_invariants_met(num_expected_messages: u32) -> Result<bool> {
```

rust/utils/run-locally/src/main.rs
```diff
@@ -476,38 +476,38 @@ fn fetch_metric(port: &str, metric: &str, labels: &HashMap<&str, &str>) -> Vec<u
                 .all(|(k, v)| l.contains(&format!("{}=\"{}\"", k, v)))
         })
         .map(|l| l.rsplit_once(' ').unwrap().1.parse::<u32>().unwrap())
-        .collect()
+        .collect())
```
mattiecnvr, line 479: I think we should change this to something more like:

```rs
resp
    .call()?
    .into_string()?
    .lines()
    .filter(|l| l.starts_with(metric))
    .filter(|l| labels.iter().all(|(k, v)| l.contains(&format!("{k}=\"{v}\""))))
    .map(|l| Ok(
        l.rsplit_once(' ')
            .ok_or(eyre!("Unknown metric format"))?.1
            .parse::<u32>()?
    ))
    .collect()
```

This will catch the errors in parsing and `collect` should auto convert `Iterator<Item=Result<u32>>` into `Result<Vec<u32>>`

mattiecnvr (approved):

PR #2295, opened 2023-05-26
Relayer and Mailbox program supports fetching of recipient isms

### LINKED ISSUE #2209: Sealevel Mailbox - query recipient ISMs
At the moment, the `inbox_process` function always uses the configured default ISM (found in the Inbox account). Ultimately, we want recipients to have the ability to default to whatever the Mailbox's default ISM is, or to have their own "custom" ISM be used. This is done by having some logic like this when `inbox_process` is called:
1. Call the recipient's `interchain_security_module() -> Option<Pubkey>` function
2. If None is given, use the default ISM for security (i.e. call `verify` on the default ISM)
3. If Some is given, use this as the ISM for security (i.e. call `verify` on the program found at this Pubkey)

The one awkward bit here is the ISM that someone specifies will probably require certain accounts to be passed into the invocation. Not 100% sure what the best way of dealing with this is, but I think that can be figured out - there's probably a pattern out there that I'm not aware of?
E.g. the accounts passed into the process function should be something like `[inbox PDA, auth PDA, ISM, ...ISM addresses, recipient, ...recipient addresses]`. So maybe the # of ISM addresses should be passed into the process instruction so that the # of ISM addresses can be figured out when crafting the invoke_signed call?

Anyways, concretely:
1. Change the Mailbox program to query the recipient for what ISM to use
2. Make required changes to agent / tests / other client code to ensure things still work end to end

tkporter:
### Description

* Fixes #2209 
* The relayer doesn't yet get the list of account metas that the ISM.verify instruction uses - this is coming next!
* Removes the ism / ism_accounts from the Inbox data & replaces with just "default_ism"

### Drive-by changes
no
### Related issues

- Fixes #2209

### Backward compatibility

Yes

None

### Testing

Unit Tests & ran end to end locally 😎

### DIFF
--- a/rust/chains/hyperlane-sealevel/src/mailbox.rs
+++ b/rust/chains/hyperlane-sealevel/src/mailbox.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use async_trait::async_trait;
-use borsh::BorshDeserialize;
+use borsh::{BorshDeserialize, BorshSerialize};
 use hyperlane_core::{
     ChainCommunicationError, ChainResult, Checkpoint, ContractLocator, Decode as _, Encode as _,
     HyperlaneAbi, HyperlaneChain, HyperlaneContract, HyperlaneDomain, HyperlaneMessage,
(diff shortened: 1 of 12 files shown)


tkporter:
The followup I mentioned I accidentally just pushed to `trevor/sealevel-validator-rebase` - you can see it here https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/6e1477d3d5428d1370a716cdb944a80d0b3010f3

PR #2299, opened 2023-05-26
Omniscient relayer infra

### LINKED ISSUE #1664: Validator and Relayer deployments should be done separate
Currently, the helm chart deploys both validator and relayer (and scraper). While there are some common config and secrets that can indeed be shared, there are many times were one would want to deploy them independently. Not only to update agent-specific config, but also to allow for separation of the more sensitive validator deployments.

Additionally, completion of this ticket is valuable to make the deployment of validators/relayers easier to use the helm chart for deployment. For that, the README/values.yml file should be made more accessible. Thus, likely secret management should be extracted as well to be agnostic as the helm chart

mattiecnvr:
### Description

- Adds support for omniscient relayers
- Updates config formats to allow better overideability by agent role
- Enables deploying selected agents rather than having to deploy them all at once
- Merges scraper scripts and configuration

### Drive-by changes

Lots of cleanup was made in the infrastructure scripts and source as it was already being touched by the above. While this PR touches a lot of files, many of the changes are just because we have so many scripts and a small change to something they all use instantly can lead to many files changed.

### Related issues

- Fixes hyperlane-xyz/issues#438
- Fixes hyperlane-xyz/issues#447
- Fixes #1664

### Backward compatibility

No

Yes

### Testing

Manual
Unit Tests


mattiecnvr:
At the time of writing this I noticed there are 99 files changed, so to make light of this for any reviewers, I had ChatGPT write up a song for you:

> (Verse 1)
> 99 files changed in the PR, 99 files of code,
> Review one down, test it around,
> 98 files changed in the PR.
> 
> (Verse 2)
> 98 files changed in the PR, 98 files of code,
> Run the linter, see it hinder,
> 97 files changed in the PR.
> 
> (Verse 3)
> 97 files changed in the PR, 97 files of code,
> Merge conflict, as you suspected,
> 96 files changed in the PR.
> 
> (Chorus)
> Oh, the life of a coder is never done,
> Infrastructure cleanup isn't always fun,
> But with each commit, we carry on,
> For the omniscient relayer must run.
> 
> (Verse 4)
> 96 files changed in the PR, 96 files of code,
> Refactor this, hope nothing's amiss,
> 95 files changed in the PR.
> 
> (Verse 5)
> 95 files changed in the PR, 95 files of code,
> Debug that log, lost in the fog,
> 94 files changed in the PR.
> 
> (Chorus)
> In the depths of the code, we add a layer,
> For the glory of the omniscient relayer,
> We'll review, test, and run each prayer,
> For the progress, we are the player.
> 
> (Verse 6)
> 94 files changed in the PR, 94 files of code,
> Fix the syntax, avoid the callback's attacks,
> 93 files changed in the PR.
> 
> (And so on...)
> 
> (Last verse)
> 1 file changed in the PR, 1 file of code,
> Review it now, take your final bow,
> No more files changed in the PR.
> 
> (Outro)
> From 99 down to none, the task is finally done,
> Every bug found, every function sound,
> All hail the omniscient relayer, its support has begun,
> In the world of coding, another victory won.

nambrot (approved):

typescript/infra/scripts/deploy.ts
```diff
@@ -136,6 +156,157 @@ async function main() {
...
+      [],
+    );
+    return {
+      TestRecipient,
+      TestTokenRecipient,
+    };
+  }
+}
+
+async function deployWithArtifacts<Config>(
```
tkporter, line 216: I personally like the pattern on having a script generally be just a `main` fn and then having helper functions live outside the script

also allows for functions like this or `writeAgentConfig` down below to be easily imported / used by other scripts at some point if needed

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -407,25 +393,23 @@ class ContextFunder {
...
-        if (role === KEY_ROLE_ENUM.Relayer) {
-          // Relayer keys should not be funded on the origin chain
-          for (const remote of this.chains) {
-            chainKeys[remote] = chainKeys[remote].concat(
-              keys.filter(
-                (_) => _.chainName !== remote && _.chainName === chain,
-              ),
-            );
+      for (const key of keys) {
+        if (key.role == Role.Relayer || key.role == Role.Kathy) {
```
tkporter, line 403: Do we need this condition at all? I think we cover this in knowing that `rolesToFund` is limited to only relayer or kathy in the `validRoles` check below

tkporter (approved):
Didn't do the deepest of passes, but generally lgtm

I do think we should bring back separate commands to deploy the test contracts though as I've used them before and can see us wanting to use them in the future

mattiecnvr:
I think I resolved your suggestions

PR #2310, opened 2023-05-31
Fix bug in postgres message coercion

yorhodes:
### Description

Goerli had a message dispatched (nonce=1964) with a destination domain that failed to be converted from i32 to u32 when being retrieved from the database. This fixes these bugs for domain ids and nonce and improves logs for this type of failure.

### Backward compatibility

Scraper needs to be redeployed

### Testing

Ran scraper locally against read replica of explorer DB


asaj (approved):
Nice

Issue #2340, opened 2023-06-06
Validator regression: validator panics when starting up without any messages


tkporter:
When a validator starts up, it now spawns a couple tasks:
1. Backfill submitter (if the tree count is > 0): https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/validator/src/validator.rs#L159-L170
2. The new message-id signing checkpoint submitter: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/validator/src/validator.rs#L172-L175
3. The legacy checkpoint submitter: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/validator/src/validator.rs#L176-L179

Before we made these changes, it used to just be a single checkpoint submitter, which is now the legacy checkpoint submitter. There was (& still is) some logic that will wait until the mailbox's count is > 0 to begin calling `latestCheckpoint` (because latestCheckpoint will subtract 1 from the count, and otherwise there will be an underflow in the view call). It now lives within the `legacy_checkpoint_submitter` function: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/validator/src/submit.rs#L135-L146

However the new checkpoint submitter doesn't have this logic - it just dives straight into making `latestCheckpoint` view calls https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/validator/src/submit.rs#L63-L82

We need to make sure that the new checkpoint submitter similarly waits for the first message to be dispatched

PR #2341, opened 2023-06-06
Block validator submitter on nonzero mailbox count

yorhodes:
### Description

_What's included in this PR?_

### Drive-by changes

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2340

### Backward compatibility

Yes
No

None
Yes

### Testing

None
Manual
Unit Tests

### DIFF
--- a/rust/agents/validator/src/validator.rs
+++ b/rust/agents/validator/src/validator.rs
@@ -105,6 +105,18 @@ impl BaseAgent for Validator {
     async fn run(&self) -> Instrumented<JoinHandle<Result<()>>> {
         self.announce().await.expect("Failed to announce validator");
 
+        // Ensure that the mailbox has > 0 messages before we enter the main
+        // validator submit loop. This is to avoid an underflow / reverted
+        // call when we invoke the `mailbox.latest_checkpoint()` method,
+        // which returns the **index** of the last element in the tree
+        // rather than just the size.  See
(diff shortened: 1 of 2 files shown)


rust/agents/validator/src/validator.rs
```diff
@@ -105,6 +105,21 @@ impl BaseAgent for Validator {
     async fn run(&self) -> Instrumented<JoinHandle<Result<()>>> {
         self.announce().await.expect("Failed to announce validator");
 
+        let reorg_period = NonZeroU64::new(self.reorg_period);
```
asaj, line 108: nit: We call this "lag" in run_checkpoint_submitters, let's be consistent (I prefer reorg_period)

rust/agents/validator/src/validator.rs
```diff
@@ -222,6 +236,7 @@ impl Validator {
                         validator_address=?announcement.validator,
                         "Please send tokens to the validator address to announce",
                     );
+                    sleep(self.interval).await;
```
asaj, line 239: Why move this?

asaj (approved):

rust/agents/validator/src/validator.rs
yorhodes: otherwise the validator sleeps after successful announcement which seems silly

PR #2344, opened 2023-06-07
Mattie's single signer

mattiecnvr:
I think this will fix it, but either way it should get you closer.

NOTE: This PR merges into the `single-signer` branch intentionally for @yorhodes 's review and learning.

### DIFF
--- a/rust/agents/scraper/src/agent.rs
+++ b/rust/agents/scraper/src/agent.rs
@@ -152,7 +152,7 @@ impl BaseAgent for Scraper {
     }
 
     #[allow(clippy::async_yields_async)]
-    async fn run(&self) -> Instrumented<JoinHandle<eyre::Result<()>>> {
+    async fn run(self) -> Instrumented<JoinHandle<eyre::Result<()>>> {
         let mut tasks = Vec::with_capacity(self.scrapers.len());
         for domain in self.scrapers.keys() {
             tasks.push(self.scrape(*domain).await);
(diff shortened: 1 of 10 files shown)


rust/agents/scraper/src/agent.rs
```diff
@@ -152,7 +152,7 @@ impl BaseAgent for Scraper {
     }
 
     #[allow(clippy::async_yields_async)]
-    async fn run(&self) -> Instrumented<JoinHandle<eyre::Result<()>>> {
+    async fn run(self) -> Instrumented<JoinHandle<eyre::Result<()>>> {
```
mattiecnvr, line 155: Tbh, I have no idea why I didn't already have it take the `self` receiver. This makes things slightly easier since we can avoid a mutex in the validator.

rust/chains/hyperlane-ethereum/src/singleton_signer.rs
```diff
@@ -1,63 +1,93 @@
...
+#[derive(Clone)]
+pub struct SingletonSignerHandle {
     address: H160,
-    tx: tokio::sync::mpsc::UnboundedSender<SignHashWithCallback>,
+    tx: mpsc::UnboundedSender<SignTask>,
+}
+
+impl fmt::Debug for SingletonSignerHandle {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("SingletonSignerHandle")
```
mattiecnvr, line 40: Intentional impl of debug to avoid `tx` field.

PR #2361, opened 2023-06-08
Update sealevel multisig ism to be compatible with CheckpointWithmessageId

### LINKED ISSUE #2347: Sealevel multisig ISM message ID - ensure consistency with Solidity version & agents
The Sealevel message ID based multisig ISM was originally built while the Solidity impl & agent impls were still in the works. It's possible that there are some inconsistencies in the expected metadata format and/or the checkpoint contents

For example, last I checked the signing digest of the checkpoint used in `rust/sealevel/hyperlane-core` isn't consistent with that of `rust/hyperlane-core` :( This needs to get fixed, and the tests need to be fixed too

tkporter:
### Description

Fixes #2347, changes tests as required

### Drive-by changes

no

### Related issues

- Fixes #2347 

### Backward compatibility

Yes

None

### Testing

Unit Tests


tkporter:
oops, meant to do against feature branch

Issue #2189, opened 2023-05-05
Fallback provider should protect against a stalled node


nambrot:
If the first provider in a set for fallback provider stalls, fallback provider will effectively also stall.

A workaround is to have fallback provider be smarter and change up the order when it detects a stalled node

nambrot:
Stalling = Does not get a new block

PR #2369, opened 2023-06-08
feat(fallback provider): re-prioritize providers on error

daniel-savu:
### Description

Closes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2189

Reprioritize providers in the `FallbackProvider` when they stop receiving new blocks (although a `max_block_time` has elapsed). The `JsonRpcClient::request(...)` implementation now takes a snapshot of the provider priority list and iterates on that. This is to minimize lock contention - no lock is held across an `await` so there is no deadlock risk. If one of the errors above occurs, the priority list is overwritten. 

Note this may cause a prior reprioritization to be lost, if the repriotization occured after the current tokio task's snapshot was taken. I'm not entirely sure whether a specific `FallbackProvider` is referenced by more than one task at a time though.

### Drive-by changes

A `PrometheusJsonRpcClientConfigExt` trait is added, so the provider can be mocked.

### Testing

Unit tests.


rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -98,6 +119,10 @@ impl From<FallbackError> for ProviderError {
     }
 }
 
+fn warn_span(idx: usize, provider: &PrometheusJsonRpcClient<Http>) {
+    let _span = warn_span!("request_with_fallback", provider_index=%idx, ?provider).entered();
```
mattiecnvr, line 123: I don't think this does anything since the span is not returned. Lmk if I am wrong.

Also because of how the macros work you should always invoke them where they are being made since they include line numbers. If you do keep a wrapper fn-like use a macro.

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,19 +143,42 @@ impl JsonRpcClient for FallbackProvider<PrometheusJsonRpcClient<Http>> {
             if !errors.is_empty() {
                 sleep(Duration::from_millis(100)).await;
             }
-            for (idx, provider) in self.0.iter().enumerate() {
+            let prioritized_providers: Vec<&_>;
```
mattiecnvr, line 146: Can we make this a function to highlight the logic on PrioritizedProviders?

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,19 +143,42 @@ impl JsonRpcClient for FallbackProvider<PrometheusJsonRpcClient<Http>> {
...
+                            warn_span(idx, provider);
+                            Ok(serde_json::from_value(v)?)
+                        }
+                    }
+                    RetryableErr(e) | RateLimitErr(e) => {
+                        {
+                            // De-prioritize the current provider by moving it to the end of the queue
+                            // No need to break out of the loop afterwards because we're looping over a snapshot.
+                            let mut priorities = self.0.priorities.write().await;
+                            let p = priorities.remove(idx);
```
mattiecnvr, line 171: Can we also make this a function, something like "deprioritize"

mattiecnvr (approved):
I'm just on my phone so I'll rely mostly on the e2e test to validate things but the code approach looks good to me overall. I do want the span trace addressed before this gets merged though.

nambrot:
So I can't comment super deeply on the change itself per se and I forget what the Retryable vs NotRetryable errors are, but the original intention behind 2189 was simpler I believe. If the first provider never errors, but also not receives new blocks, then the relayer or scraper will effectively be stuck, because no error occurs, and thus Fallback provider does not bother to check if the node is actually up to date?

I think re-prioritization of providers in a FallbackProvider is still useful, but IMO more of an optimization for latency and cost, 2189 is about reliability

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,13 +152,60 @@ impl From<FallbackError> for ProviderError {
...
+}
+
+fn is_stalled(current_block_height: u64, priority: &PrioritizedProviderInner) -> bool {
+    if current_block_height == priority.last_block_height.0 {
+        let now = Instant::now();
+        return now
+            .duration_since(priority.last_block_height.1)
+            .gt(&MAX_BLOCK_TIME);
+    }
+    current_block_height > priority.last_block_height.0
```
tkporter, line 177: think this condition may be inverted

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,13 +152,60 @@ impl From<FallbackError> for ProviderError {
...
 }
 
-fn warn_span(idx: usize, provider: &PrometheusJsonRpcClient<Http>) {
-    let _span = warn_span!("request_with_fallback", provider_index=%idx, ?provider).entered();
+async fn handle_stalled_provider(
+    fallback_provider: &HttpFallbackProvider,
+    priority: &PrioritizedProviderInner,
+) -> Result<(), <HttpFallbackProvider as JsonRpcClient>::Error> {
+    let provider = &fallback_provider.0.providers[priority.index];
+    let current_block: U64 = provider.request("eth_blockNumber", ()).await?;
```
tkporter, line 160: As an optimization to avoid always needing to make an additional eth_blockNumber call whenever any kind of RPC is made, it's probably worth also piggybacking off intentional eth_blockNumber calls that are made, & also avoiding making this eth_blockNumber RPC super frequently

e.g. something like:
* in `request`, check if the RPC being made is an `eth_blockNumber` - if it is, piggyback off this and save it in the `last_block_height`
* if it's not an `eth_blockNumber`, check when the last `eth_blockNumber` for that provider was made. If it's been X mins (maybe 2 mins or something) since we were made aware of the last `eth_blockNumber` for this provider, then make the `eth_blockNumber` request ourselves and check if it's stalled

daniel-savu:
Is suggest turning the TODO I added in the code into a ticket: Make `categorize_client_response` generic over `ProviderError` to allow testing two stalled providers, so that custom errors can be constructed and the for loop in `request` doesn't stop after the first provider

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
daniel-savu: As discussed, I implemented the second bullet point and the first one can be a separate ticket to avoid scope creep

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,15 +265,18 @@ impl JsonRpcClient for FallbackProvider<PrometheusJsonRpcClient<Http>> {
...
+                let provider = &self.inner.providers[priority.index];
                 let fut = match params {
                     Value::Null => provider.request(method, ()),
                     _ => provider.request(method, &params),
                 };
-
                 let resp = fut.await;
+                handle_stalled_provider(self, priority).await?;
                 let _span =
                     warn_span!("request_with_fallback", provider_index=%idx, ?provider).entered();
```
tkporter, line 278: before the `provider_index` meant two things - the first being to indicate the # of providers that had been tried previously when falling back, and the second being which provider in the list of providers was currently being used

now that things have changed a bit, wondering if we should change this to something like `fallback_count=%idx`? and if we keep `provider_index`, to have it be `provider_index=%priority.index` or something of the like

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -98,9 +171,83 @@ impl From<FallbackError> for ProviderError {
     }
 }
 
+async fn handle_stalled_provider<C>(
+    fallback_provider: &FallbackProvider<C>,
```
tkporter, line 175: wdyt of putting all these functions in an `impl<C> FallbackProvider where C: JsonRpcClient` so that we can just use `&self`?

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -98,9 +171,83 @@ impl From<FallbackError> for ProviderError {
...
+        return Ok(());
+    }
+
+    let provider = &fallback_provider.inner.providers[priority.index];
+    let current_block_height: u64 = provider
+        .request(BLOCK_NUMBER_RPC, ())
+        .await
+        .map(|r: U64| r.as_u64())
+        .unwrap_or(priority.last_block_height.0);
+    if current_block_height == priority.last_block_height.0 {
```
tkporter, line 196: I think it's worth expanding this case to `current_block_height <= priority.last_block_height.0`

while unlikely, we've seen some very janky stuff in the past so I could see there being a case where a provider gives a lower block number than before, which we'd also wanna deprioritize on

Providers usually have multiple nodes behind a load balancer and sometimes requests go to different nodes with different views of the world - or a node could even get in a weird state where it has a corrupted datadir and its tip gets reset to a lower block number

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -98,9 +171,83 @@ impl From<FallbackError> for ProviderError {
...
+    let now = Instant::now();
+    if now
+        .duration_since(priority.last_block_height.1)
+        .le(&fallback_provider.max_block_time)
+    {
+        // Do nothing, it's too early to tell if the provider has stalled
+        return Ok(());
+    }
+
+    let provider = &fallback_provider.inner.providers[priority.index];
```
tkporter, line 190: I don't feel strongly but noting that I think we could just make this a param to `handle_stalled_provider` as we did this lookup in the call site already

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -119,15 +265,18 @@ impl JsonRpcClient for FallbackProvider<PrometheusJsonRpcClient<Http>> {
...
+            let priorities_snapshot = take_priorities_snapshot(self).await;
+            for (idx, priority) in priorities_snapshot.iter().enumerate() {
+                let provider = &self.inner.providers[priority.index];
                 let fut = match params {
                     Value::Null => provider.request(method, ()),
                     _ => provider.request(method, &params),
                 };
-
                 let resp = fut.await;
+                handle_stalled_provider(self, priority).await?;
```
tkporter, line 276: Wdyt of doing the stall check before making the actual RPC request, and if the provider in question is deprioritized we fallback to the next provider? Iiuc here we still treat the response from a possibly stalled provider as valid, and only get the "gains" of falling back to another provider in a subsequent RPC. Not a big deal but feels like we can get more for free by a small change

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -139,3 +288,125 @@ impl JsonRpcClient for FallbackProvider<PrometheusJsonRpcClient<Http>> {
         Err(FallbackError::AllProvidersFailed(errors).into())
     }
 }
+
+#[cfg(test)]
+mod tests {
```
tkporter, line 293: love the tests!

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -98,9 +171,83 @@ impl From<FallbackError> for ProviderError {
...
+    if current_block_height == priority.last_block_height.0 {
+        // The `max_block_time` elapsed but the block number returned by the provider is still the same
+        deprioritize_provider(fallback_provider, priority).await;
+    } else {
+        update_last_seen_block(fallback_provider, priority.index, current_block_height).await
+    }
+    Ok(())
+}
+
+async fn deprioritize_provider<C>(
```
tkporter, line 205: A log when this happens would be nice

tkporter (commented):
great job! some small questions / suggestions that I don't feel very strongly about, curious what you think

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
daniel-savu: Spent some time trying to add this but it's taking longer than it's worth it. The biggest problem is that the mocked provider's `request` will never get to sleep so that the deprioritization condition is met. The other issue is that it's taken me a long time to find a type that's `DeserializeOwned` so as to mock a return value that represents a real block and not just an error. 

I also realised that unless I can create a value that implements `DeserializedOwned`, I won't be able to test all the edge cases, because the providers inside `FallbackProvider` are all initialized with `Instant::now()` and tested serially, so if one sleeps for longer than the max block period, all of them will have become potentially stalled - and then `request` will return an error from `serde_json` returning an error in my mock.

tkporter (approved):
LGTM! Nice job

rust/ethers-prometheus/src/json_rpc_client.rs
```diff
@@ -74,7 +74,16 @@ pub struct PrometheusJsonRpcClientConfig {
     pub chain: Option<ChainInfo>,
 }
 
-impl PrometheusJsonRpcClientConfig {
+/// Helper functions for displaying node and chain information
+pub trait PrometheusJsonRpcClientConfigExt {
```
mattiecnvr, line 78: What is the benefit to this being an ext trait vs just added to the `PrometheusJsonRpcClientConfig` impl? We should have full control over the original so I think it makes more sense in this case to update it.

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -50,16 +102,82 @@ impl<T> FallbackProvider<T> {
...
+            .le(&self.max_block_time)
+        {
+            // Do nothing, it's too early to tell if the provider has stalled
+            return Ok(());
+        }
+
+        let current_block_height: u64 = provider
+            .request(BLOCK_NUMBER_RPC, ())
+            .await
+            .map(|r: U64| r.as_u64())
```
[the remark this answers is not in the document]
daniel-savu: Doesn't work sadly, because the `.request(...)` call above is generic and the return type has to be specified somewhere

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -50,16 +102,82 @@ impl<T> FallbackProvider<T> {
...
+            // The `max_block_time` elapsed but the block number returned by the provider has not increased
+            self.deprioritize_provider(priority, provider).await;
+        } else {
+            self.update_last_seen_block(priority.index, current_block_height)
+                .await;
+        }
+        Ok(())
+    }
+
+    async fn deprioritize_provider(&self, priority: &PrioritizedProviderInner, provider: &C) {
```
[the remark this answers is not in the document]
daniel-savu: I moved the log to the call site of `deprioritize_provider` which is fine since it's only one site, otherwise I would've probably kept the `provider` argument to reduce logging duplication

rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -50,16 +102,82 @@ impl<T> FallbackProvider<T> {
...
+        let mut priorities = self.inner.priorities.write().await;
+        priorities.retain(|&p| p.index != priority.index);
+        priorities.push(*priority);
+
+        info!(
+            provider_index=%priority.index,
+            ?provider,
+            "Deprioritizing an inner provider in FallbackProvider",
+        );
+        // Free the write lock
```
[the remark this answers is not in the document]
daniel-savu: Makes sense, thanks for the detailed justification 💯

rust/ethers-prometheus/src/json_rpc_client.rs
daniel-savu: I wasn't able to test the fallback provider logic previously with the concrete `PrometheusJsonRpcClientConfig` type. Now that it uses a generic bound by `PrometheusJsonRpcClientConfigExt`, I could just implement my own `ProviderMock` to test the new fallback provider logic (see the unit tests at the end of `fallback.rs`).

PR #2376, opened 2023-06-12
Deploy mainnet scraper

asaj:
### Description

Updates the deployment to pick up #2375

### Drive-by changes

- Add -c alias for context

### DIFF
--- a/typescript/infra/scripts/utils.ts
+++ b/typescript/infra/scripts/utils.ts
@@ -75,7 +75,8 @@ export function withContext<T>(args: yargs.Argv<T>) {
   return args
     .describe('context', 'deploy context')
     .coerce('context', assertContext)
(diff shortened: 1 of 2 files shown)


typescript/infra/scripts/utils.ts
```diff
@@ -75,7 +75,8 @@ export function withContext<T>(args: yargs.Argv<T>) {
   return args
     .describe('context', 'deploy context')
     .coerce('context', assertContext)
-    .demandOption('context');
+    .demandOption('context')
+    .alias('c', 'context');
```
mattiecnvr, line 79: This conflicts with one of the other options. You'll need to remove the alias from it. (It's chain down below)

nambrot (approved):

Issue #1902, opened 2023-03-07
Persist message retries in run queue


tkporter:
Every time we restart the relayer, the serial submitter starts from absolute scratch.
There are a bunch of messages in the queue that the relayer may have already retried many times & should therefore ideally _not_ be treated as if they've never been retried. So we should persist the number of retries a message has had, so that when we restart the relayer we don't see an insane flurry of RPCs

PR #2382, opened 2023-06-13
Reconstruct relayer `OpQueue` from HyperlaneDB

daniel-savu:
### Description

Stores the `num_retries` field of `PendingMessage` at its own prefix in HyperlaneDB, so that `PendingMessage` can be reconstructed with the same exponential backoff in the `next_attempt_after` field and avoid making unnecesary RPC calls on restart.

To use persisted data, `PendingMessage` is constructed with `from_persisted_retries` instead of the default constructor. If reading from the db fails, the new constructor just defaults to the regular constructor, so it's worth considering replacing the regular constructor with this one entirely.

### Drive-by changes

A new db prefix is added (`PENDING_MESSAGE_RETRY_COUNT_FOR_MESSAGE_ID `), as well as related getter / setter methods. Using the new prefix helps avoid a migration. I do see a `processed_by_nonce` prefix, which could be migrated to store data related to `PendingMessage` in general so let me know what you think. Could be a future improvement to merge its prefix with the prefix added in this PR.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1902

### Backward compatibility

Yes

None

### Testing

A processor is run on some dummy database entries. The `SerialSubmitter`'s end of the channel is used to ensure that the incoming `PendingMessage`s have the correct backoff in `_next_attempt_after`.


rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -300,6 +300,28 @@ impl PendingOperation for PendingMessage {
...
+            .ctx
+            .origin_db
+            .retrieve_pending_message_data_by_message_id(&message.id())
+        {
+            let next_attempt_after =
+                PendingMessage::calculate_msg_backoff(num_retries).map(|dur| Instant::now() + dur);
+            pm.num_retries = num_retries;
+            pm.next_attempt_after = next_attempt_after;
+        } else {
+            info!(
```
tkporter, line 317: I'd suggest making this a trace (or maybe debug) - iiuc this log will be made whenever the relayer tries to process a new message that it didn't know from a previous restart, which is a very common occurrence. People tend to get spooked by messages with the word "failed" in it, so maybe also worth rephrasing as "No persisted retry count for message {}" or something

daniel-savu: Good point, turned that into a trace

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -297,9 +297,43 @@ impl PendingOperation for PendingMessage {
...
 impl PendingMessage {
+    /// Constructor that tries reading the retry count from the HyperlaneDB in order to recompute the `next_attempt_after`.
+    /// In case of failure, behaves like `Self::new(...)`.
+    pub fn from_persisted_retries(message: HyperlaneMessage, ctx: Arc<MessageContext>) -> Self {
+        fn log_db_read_error(maybe_error: Option<DbError>, message_id: &H256) {
+            let mut log_message = format!("Failed to read retry count from HyperlaneDB for message {}. Defaulting to the regular `PendingMessage` constructor.", message_id);
+            if let Some(error) = maybe_error {
+                log_message.push_str(&format!(" Error: {}", error));
+            }
+            info!(log_message);
```
tkporter, line 316: Imo if `maybe_error` is None this should be trace, otherwise something like info or warn makes sense to me. Otherwise I think the default case of a brand new pending message being created will be confusing for ppl

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -329,21 +363,41 @@ impl PendingMessage {
...
+    }
+
+    fn persist_retries(&self) {
+        if let Err(e) = self
+            .ctx
+            .origin_db
+            .store_pending_message_data_by_message_id(&self.message.id(), &self.num_retries)
+        {
+            trace!(
+                "No persisted retry count for message {}. Hyperlane DB error: {}",
```
tkporter, line 390: Think this would be fair as `info` or `warn`, and should say that the persisting of the retry count failed instead of saying that there was no persisted retry count

rust/agents/relayer/src/msg/processor.rs
```diff
@@ -198,3 +201,242 @@ impl MessageProcessorMetrics {
...
+            .await;
+
+            // Expect the HyperlaneDB entry to have been updated, so the `OpQueue` in the submitter
+            // can be accurately reconstructed on restart.
+            // If the retry counts were correctly persisted, the backoffs will have the expected value.
+            pending_messages
+                .iter()
+                .zip(msg_retries_to_set.iter())
+                .for_each(|(pm, expected_retries)| {
+                    // Round up the actuall backoff because it was calculated with an `Instant::now()` that was a fraction of a second ago
```
tkporter, line 431: hope this doesn't prove to add some nondeterminism in the unit tests? Probably is fine but if it ever does provide issues we can consider something like https://docs.rs/mock_instant/latest/mock_instant/

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -312,3 +313,10 @@ make_store_and_retrieve!(pub, processed_by_nonce, NONCE_PROCESSED, u32, bool);
 make_store_and_retrieve!(pub(self), processed_by_gas_payment_meta, GAS_PAYMENT_META_PROCESSED, InterchainGasPaymentMeta, bool);
 make_store_and_retrieve!(pub(self), interchain_gas_expenditure_data_by_message_id, GAS_EXPENDITURE_FOR_MESSAGE_ID, H256, InterchainGasExpenditureData);
 make_store_and_retrieve!(pub(self), interchain_gas_payment_data_by_message_id, GAS_PAYMENT_FOR_MESSAGE_ID, H256, InterchainGasPaymentData);
+make_store_and_retrieve!(
+    pub,
+    pending_message_data_by_message_id,
```
tkporter, line 318: nit: it's not clear that data == retry count unless if you look at how these functions are called, maybe this should be `pending_message_retry_count_by_message_id`?

tkporter (approved):
Looks great! A couple small nits. I think it's important to avoid a high log severity for the expected default case of a retry count not yet being present in the DB

rust/agents/relayer/src/msg/processor.rs
daniel-savu: On my local machine, the difference between instants was less than 0.01s, so I think it's very unlikely that it would ever exceed 0.5s and be rounded the wrong way. Let's see, and I'll keep `mock_instant` in mind 👍

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -329,21 +365,41 @@ impl PendingMessage {
...
+        if let Err(e) = self
+            .ctx
+            .origin_db
+            .store_pending_message_retry_count_by_message_id(&self.message.id(), &self.num_retries)
+        {
+            info!(
+                "Persisting the `num_retries` failed for message {}. Hyperlane DB error: {}",
+                self.message.id(),
+                e
+            );
```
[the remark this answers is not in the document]
mattiecnvr: I think this one should be a warn, also again with formatting log messages.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -359,8 +415,8 @@ impl PendingMessage {
 
 #[derive(Debug)]
 pub struct MessageSubmissionMetrics {
-    last_known_nonce: IntGauge,
-    messages_processed: IntCounter,
+    pub last_known_nonce: IntGauge,
```
mattiecnvr, line 418: This change to `pub` fields feels odd.

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -297,9 +297,45 @@ impl PendingOperation for PendingMessage {
...
+    fn set_retries(&mut self, retries: u32) {
+        self.set_retries(retries);
+    }
 }
 
 impl PendingMessage {
+    /// Constructor that tries reading the retry count from the HyperlaneDB in order to recompute the `next_attempt_after`.
+    /// In case of failure, behaves like `Self::new(...)`.
+    pub fn from_persisted_retries(message: HyperlaneMessage, ctx: Arc<MessageContext>) -> Self {
+        fn log_db_read_error(maybe_error: Option<DbError>, message_id: &H256) {
```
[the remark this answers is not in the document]
daniel-savu: Thanks for this! I noticed that other parts of the code simply log the entire result, I went for that since it's less verbose

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: I needed the fields to be public so I could construct a dummy instance in the unit tests: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/72e13bc8deaec05d218916bbdc91e6db92ce6445/rust/agents/relayer/src/msg/processor.rs#L240

mattiecnvr (approved):

PR #2384, opened 2023-06-13
Reduce hyper client pool idle timeout to 20s

yorhodes:
### Description

- configures hyper http client to close connections after 20s
- makes singleton signer retry (5) times before bubbling error up

### Backward compatibility

Yes

### Testing

Local binary testing against cloud resources

Screen Shot 2023-06-13 at 5 45 47 PM

### DIFF
--- a/rust/chains/hyperlane-ethereum/src/singleton_signer.rs
+++ b/rust/chains/hyperlane-ethereum/src/singleton_signer.rs
@@ -68,7 +68,21 @@ impl SingletonSigner {
     /// Run this signer's event loop.
     pub async fn run(mut self) {
         while let Some((hash, tx)) = self.rx.recv().await {
-            if tx.send(self.inner.sign_hash(&hash).await).is_err() {
+            // retry 5 times
+            let mut retries = 5;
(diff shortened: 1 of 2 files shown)


rust/chains/hyperlane-ethereum/src/singleton_signer.rs
```diff
@@ -62,13 +63,38 @@ impl SingletonSigner {
...
+                inner,
+                rx,
+                retries: 5,
+            },
+            SingletonSignerHandle { address, tx },
+        )
+    }
+
+    /// Change default (5) retries for signing
+    pub fn config_retries(&mut self, retries: usize) {
```
asaj, line 77: Why have an unused setter?

asaj (approved):
Nice detective work

rust/chains/hyperlane-ethereum/src/singleton_signer.rs
yorhodes: wanted to make it configurable while not changing the usage
idk why I chose to do it this way lol

PR #2385, opened 2023-06-13
Increase key funder target balances

asaj:
### Description

This PR multiplies the key funder target balances per chain by ~8.

This is to account for the fact that a single key now delivers messages from *all* origin chains to an individual destination chain, whereas prior to the rollout of the omniscient relayer, each key would deliver messages between a single origin/destination chain pair.

A separate PR will re-deploy the key funder with to pick up these changes.

### Drive-by changes

None


nambrot:
Discussed during standup that we should maybe hold off until https://github.com/hyperlane-xyz/issues/issues/468

yorhodes:
@nambrot what should we do here?

paulbalaji:
replaced by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3284

PR #2387, opened 2023-06-13
Fix latest indexed block gauge

asaj:
### Description

This PR fixes a bug that affected the latest indexed block gauge for message_dispatch.

This allows the gauge to be updated to the "correct" value whether or not a getLogs query actually takes place.

### Drive-by changes

None


rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -377,17 +382,26 @@ impl<T> ContractSyncCursor<T> for RateLimitedContractSyncCursor<T>
...
         } else {
             Duration::from_secs(0)
         };
-        Ok((from, to, eta))
+
+        let rate_limit = self.get_rate_limit().await?;
+        if let Some(rate_limit) = rate_limit {
+            return Ok((CursorAction::Sleep(rate_limit), eta));
+        } else {
+            self.from = to + 1;
```
yorhodes, line 398: im a little confused on the relationship between `self.from` and the `from = to - chunk_size`
seems like we could maybe get away with having only `self.tip`

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -58,29 +59,32 @@ where
             .with_label_values(&[label, chain_name]);
 
         loop {
-            let Ok((from, to, eta)) = cursor.next_range().await else { continue };
-            debug!(from, to, "Looking for for events in block range");
+            indexed_height.set(cursor.latest_block() as i64);
+            let Ok((action, eta)) = cursor.next_action().await else { continue };
```
yorhodes, line 63: would it be useful to warn log the errors?

yorhodes (approved):
not sure I have all the context but approved

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -277,22 +281,25 @@ impl ForwardBackwardMessageSyncCursor {
...
-
-            if let Some(backward_range) = self.backward.get_next_range().await {
-                self.direction = SyncDirection::Backward;
-                return Ok(backward_range);
-            }
+    async fn next_action(&mut self) -> ChainResult<(CursorAction, Duration)> {
+        // TODO: Proper ETA for backwards sync
+        let eta = Duration::from_secs(0);
+        // Prioritize forward syncing over backward syncing.
+        if let Some(forward_range) = self.forward.get_next_range().await? {
```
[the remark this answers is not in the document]
asaj: That's because the backward range does not return a result

rust/hyperlane-base/src/contract_sync/mod.rs
asaj: Yes probably, but I'd prefer not to make more changes outside the scope of this PR, since I was hoping not to do any changes at all

Issue #2366, opened 2023-06-08
Warp route tokens can easily be deployed onto bsc testnet, solana testnet, and nautilus testnet


tkporter:

nambrot:
@tkporter is this closed?

tkporter:
yes

PR #2392, opened 2023-06-15
Existing deployment support in Router tooling, small ISM factory fixes

tkporter:
### Description

To accommodate warp routes that exist on EVM chains *and* other execution environments, we need the ability to enroll remote routers on EVM-based routers that weren't deployed by the router deployer. This adds a new optional `foreignDeployment` property to the `RouterConfig`, which means that deployment will not be attempted to this chain, but it will be enrolled as a remote router on any contracts that are deployed to. The general flow looks something like:
  * Deploy the warp route contracts on all EVM chains, enrolling each other as routers
  * Deploy the warp route contracts on the alt execution environment chains, also configuring the existing EVM contract deployments as `foreignDeployment`s so that they are enrolled as routers on the alt execution environment contracts
  * Now equipped with *all* the router addresses, any new non-EVM router addresses are configured as `foreignDeployment`s, which now enrolls these routers on the EVM chains

Note there's still a lot left to be desired even with this change -- e.g. much of the tooling expects the MultiProvider to include metadata (& even providers) for each and every chain. Because the MultiProvider is of course EVM only, things get a bit weird. For now, I've been able to deploy warp routes with configs relating to a sealevel chain successfully by just running a local Anvil instance and setting the RPC URL of the sealevel config to that anvil instance. We'll want to readdress this at some point, but for now I'd prefer to have a flow that just works and we can be more intentional about bigger deploy / tooling changes to support alt execution environments in the future

This also includes some small changes to the ISM factory - because we werent actually waiting for the `enrollValidators` tx to succeed before trying `setThreshold`, the gas estimation of `setThreshold` would fail because the threshold of 1 exceeded the current on-chain validator set size of 0. It also sometimes resulted in nonce contention

I was able to get everything in hyperlane-deploy working locally by having dependencies be local filepaths

The order of operations for things iiuc looks like:
1. Merge this
2. Ship new SDK version
3. Update hyperlane-token to use this new SDK version
4. Ship new hyperlane-token version
5. Update hyperlane-deploy to use the new SDK version and the new hyperlane-token

### Drive-by changes

none

### Related issues

- partially https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2366

### Backward compatibility

Yes

None

### Testing

unit tests, ran deploy tooling

### DIFF
--- a/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
+++ b/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
@@ -113,8 +113,14 @@ export class HyperlaneIsmFactory extends HyperlaneApp<IsmFactoryFactories> {
         .deploy();
       await this.multiProvider.handleTx(chain, multisig.deployTransaction);
       const originDomain = this.multiProvider.getDomainId(origin!);
-      await multisig.enrollValidators([originDomain], [config.validators]);
-      await multisig.setThreshold(originDomain, config.threshold);
+      await this.multiProvider.handleTx(
+        chain,
+        multisig.enrollValidators([originDomain], [config.validators]),
+      );
+      await this.multiProvider.handleTx(
+        chain,
+        multisig.setThreshold(originDomain, config.threshold),
+      );
       address = multisig.address;
     } else {
       const multisigIsmFactory =
@@ -163,7 +169,10 @@ export class HyperlaneIsmFactory extends HyperlaneApp<IsmFactoryFactories> {
       moduleAddress,
       this.multiProvider.getSigner(chain),
     );
-    await routingIsm.transferOwnership(config.owner);
+    await this.multiProvider.handleTx(
+      chain,
+      await routingIsm.transferOwnership(config.owner),
+    );
     const address = dispatchLogs[0].args['module'];
     return IRoutingIsm__factory.connect(address, signer);
   }
--- a/typescript/sdk/src/router/HyperlaneRouterDeployer.ts
+++ b/typescript/sdk/src/router/HyperlaneRouterDeployer.ts
@@ -3,7 +3,7 @@ import {
   Mailbox__factory,
   Router,
 } from '@hyperlane-xyz/core';
-import { utils } from '@hyperlane-xyz/utils';
+import { types, utils } from '@hyperlane-xyz/utils';
 
 import {
   HyperlaneContracts,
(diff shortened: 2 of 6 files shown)


typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -113,8 +113,14 @@ export class HyperlaneIsmFactory extends HyperlaneApp<IsmFactoryFactories> {
         .deploy();
       await this.multiProvider.handleTx(chain, multisig.deployTransaction);
       const originDomain = this.multiProvider.getDomainId(origin!);
-      await multisig.enrollValidators([originDomain], [config.validators]);
-      await multisig.setThreshold(originDomain, config.threshold);
+      await this.multiProvider.handleTx(
+        chain,
+        multisig.enrollValidators([originDomain], [config.validators]),
+      );
+      await this.multiProvider.handleTx(
```
yorhodes, line 120: ahh good catch

typescript/sdk/src/router/types.ts
```diff
@@ -11,7 +11,13 @@ export type OwnableConfig = {
   owner: types.Address;
 };
 
-export type RouterConfig = ConnectionClientConfig & OwnableConfig;
+export type ExistingDeploymentConfig = {
+  existingDeployment?: types.Address;
+};
```
yorhodes, line 16: I think a better name might be like "foreignDeployment" or something

typescript/sdk/src/router/HyperlaneRouterDeployer.ts
```diff
@@ -151,12 +159,29 @@ export abstract class HyperlaneRouterDeployer<
...
+      configMap,
+      (_chainName, config): config is Config => !config.existingDeployment,
+    );
+    const existingDeployments: ChainMap<types.Address> = objMap(
+      objFilter(
+        configMap,
+        (_chainName, config): config is Config => !!config.existingDeployment,
+      ),
+      (_, config) => config.existingDeployment!,
+    );
```
yorhodes, line 173: this is hard to read

yorhodes (approved):
I think we need to solve this standard serialization format sooner rather than later but approving for now
ideally we remove the conflation of config and deployment artifacts in future

typescript/sdk/src/router/HyperlaneRouterDeployer.ts
tkporter: flipped the map & filter and added some comments, hopefully slightly more clear

PR #2398, opened 2023-06-16
CCIP Read ISM

AlexBHarley:
### Description

CCIP Read ISM implementation,

- [x] Contracts
- [x] Relayer
- [x] Example HTTP backend

### Drive-by changes

### Related issues

- Fixes #[2211]

### Backward compatibility

Yes, relayers will skip messages they can't handle.

Yes, relayers won't be able to submit messages that have a CCIP Read ISM.

### Testing

- [x] Manual
- [x] Unit Tests

### General approach

- [x] Get a basic [CCIP Read implementation working](https://github.com/AlexBHarley/permissionless-chainlink-feeds/tree/main/contracts)
- [x] Build a [mock service that serves CCIP Read compatible data](https://github.com/AlexBHarley/permissionless-chainlink-feeds/tree/main/apps/api)
- [x] Update the Relayer to support CCIP Read ISMs


solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,141 @@
...
+        return keccak256(_message.body());
+    }
+
+    /**
+     * @notice Returns the signature at a given index from the metadata.
+     * @param _metadata ABI encoded module metadata
+     * @param _index The index of the signature to return
+     * @return signature Packed encoding of signature (65 bytes)
+     */
+    function signatureAt(bytes calldata _metadata, uint256 _index)
```
yorhodes, line 132: ideally the abstract CCIP ism doesnt know about signatures or digests
that being said I understand you wanted to just get something working

AlexBHarley:
After chatting with @nambrot last week I've taken a slightly different approach to the original one, interested to hear what you guys think. An implementation of a CcipReadIsm can be found [here](https://github.com/AlexBHarley/permissionless-chainlink-feeds/).

I have a couple open questions I'm not sure about,

- Is there any value in the `AbstractCcipReadIsm` living in the monorepo? I feel it's OK for consumers to implement this logic themselves, but on the other hand it's nice to have the `OffchainLookup` error well-typed for use in the relayer
- Do we actually want to call this `CcipRead`? While it's quite close to matching the spec, it doesn't exactly and that could confuse client side implementations. e.g., relayer queries for information on the destination chain, but the callback function (Mailbox.process) is not only a different contract but a different chain. Simply calling it `OffchainRead` could be better
- If we do keep with CCIP Read, using the custom error approach, I'm unsure how to better parse these. [This section of the code](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2398/files#diff-5d0308f2ded7da5abaff992e4e8eee32845b360c1a9f522ff2f9bdd4c8a14b59R54) is what I'm referring to. Right now I do a Regex match to pull the error out which feels messy

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
...
+        uint8(IInterchainSecurityModule.Types.CCIP_READ);
+
+    // ============ Mutable Storage ============
+
+    /// @notice The URLs to query for offchain data from.
+    string[] public offchainUrls;
+
+    /// @notice Data to pass back to the callback function.
+    // slither-disable-next-line uninitialized-state
+    bytes public offchainExtraData;
```
nambrot, line 46: Does this need to be specified here? IMO there is no need to prescribe that this data is stored in storage right?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
...
+import {IMailbox} from "../../interfaces/IMailbox.sol";
+import {Message} from "../../libs/Message.sol";
+import {AbstractMultisigIsm} from "../multisig/AbstractMultisigIsm.sol";
+
+/// @param sender the address of the contract making the call, usually address(this)
+/// @param urls the URLs to query for offchain data
+/// @param callData context needed for offchain service to service request
+/// @param callbackFunction function selector to call with offchain information
+/// @param extraData additional passthrough information to call callbackFunction with
+error OffchainLookup(
```
nambrot, line 16: Might be useful to comment here the slightly off-spec way of using this event?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
...
+        view
+        returns (bool)
+    {
+        revert OffchainLookup(
+            address(this),
+            offchainUrls,
+            _calldata,
+            // Note: consumers of getOffchainVerifyInfo should call the destination
+            // chain mailbox.process(_metadata, _message)
+            IMailbox.process.selector,
```
nambrot, line 81: Couldn't you get full compliance by calling specifying a function called `process(metadata, message)` which would just call the mailbox (which would call the ISM)?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    // ============ External Functions ============
+
+    /**
+     * @notice Reverts with the data needed to query information offchain before
+     * mailbox submission
+     * @dev See https://eips.ethereum.org/EIPS/eip-3668 for more information
+     * @return bool Ignored
+     */
+    function getOffchainVerifyInfo(bytes calldata _calldata)
```
nambrot, line 70: This should be _message right?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
nambrot: In that case you probably should make the mailbox part of this interface? Though IMO its probably fine for the contracts that are meant for this use case to not be full spec-compliant

solidity/contracts/test/TestCcipReadIsm.sol
```diff
@@ -0,0 +1,42 @@
...
+
+import {IInterchainSecurityModule} from "../interfaces/IInterchainSecurityModule.sol";
+import {IMultisigIsm} from "../interfaces/isms/IMultisigIsm.sol";
+import {AbstractCcipReadIsm} from "../isms/ccip-read/AbstractCcipReadIsm.sol";
+import {Message} from "../libs/Message.sol";
+
+contract TestCcipReadIsm is AbstractCcipReadIsm {
+    using Message for bytes;
+
+    address[] public validators;
```
nambrot, line 12: Is this still intentional here?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
...
+        returns (bool)
+    {
+        revert OffchainLookup(
+            address(this),
+            offchainUrls,
+            _calldata,
+            // Note: consumers of getOffchainVerifyInfo should call the destination
+            // chain mailbox.process(_metadata, _message)
+            IMailbox.process.selector,
+            offchainExtraData
```
nambrot, line 82: IMO, consumers should explicitly have to specify this (and not rely on the abstract contract passing something from storage (which very well may be uninitialized). If you want to, you can  make this into an internal function that a consumer has to override

rust/agents/relayer/src/msg/metadata/ccip_read.rs
```diff
@@ -0,0 +1,100 @@
...
+        const CTX: &str = "When fetching CcipRead metadata";
+        let ism = self.build_ccip_read_ism(ism_address).await.context(CTX)?;
+
+        let response = ism.get_offchain_verify_info(message.body.clone()).await;
+        let info: OffchainLookup = match response {
+            Ok(_) => {
+                info!("incorrectly configured getOffchainVerifyInfo, expected revert");
+                return Ok(None);
+            }
+            Err(raw_error) => {
```
nambrot, line 53: Probably should have send this to you earlier, but have you seen https://github.com/ensdomains/ethers-ccip-read. I wonder if we could reuse logic from there (looks possible?)

I think @yorhodes and I had chatted that the requirement in the spec to require the `data` field in the JSON response is not great, but services like CCTP actually are already compliant https://developers.circle.com/stablecoin/reference/getattestation

nambrot: (Might have to fork it and have it return the calldata of the final call)

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
nambrot, line 1: I assume you were planning on it, but a bit more of a README to show how folks should implement this contract would be nice

solidity/contracts/mock/MockMailbox.sol
```diff
@@ -13,7 +13,7 @@ contract MockMailbox is Versioned {
 
     // ============ Constants ============
     uint32 public immutable localDomain;
-    uint256 public constant MAX_MESSAGE_BODY_BYTES = 2 * 2**10;
+    uint256 public constant MAX_MESSAGE_BODY_BYTES = 3 * 2**10;
```
nambrot, line 16: Was this neccessary?

nambrot:
> * Is there any value in the `AbstractCcipReadIsm` living in the monorepo? I feel it's OK for consumers to implement this logic themselves, but on the other hand it's nice to have the `OffchainLookup` error well-typed for use in the relayer

For sure, IMO its critical to have it in the monorepo as its the reference for the agent implementation as well. Like i commented, it should also include instructions for the spec of this ISM, i.e. how do I override it, what interface does my offchain service have to comply with, etc. Like another comment of mine, the consumer should be forced to implement the extra data of the revert message though.

> * Do we actually want to call this `CcipRead`? While it's quite close to matching the spec, it doesn't exactly and that could confuse client side implementations. e.g., relayer queries for information on the destination chain, but the callback function (Mailbox.process) is not only a different contract but a different chain. Simply calling it `OffchainRead` could be better

Hmm, i'm not sure that I'm following. The relayer would query the ISM on the destination chain for the `OffchainRead` revert message, to then call the Mailbox.process on that very same chain no? (You could even have the ISM call Mailbox.process if you wanted to, to be fully compliant) 

> * If we do keep with CCIP Read, using the custom error approach, I'm unsure how to better parse these. [This section of the code](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2398/files#diff-5d0308f2ded7da5abaff992e4e8eee32845b360c1a9f522ff2f9bdd4c8a14b59R54) is what I'm referring to. Right now I do a Regex match to pull the error out which feels messy

You should be able to look at this https://github.com/ensdomains/ethers-ccip-read/blob/main/src/middleware.rs#L243C1-L244C1

nambrot:
Also putting a review here for the ChainlinkISM since there is no PR to look at:
- It would be nice to see a diff of the original OffchainAggregator and yours. I have terrible memory, but remind me why the Aggregator even has to be modified at all? I thought we were good with making storage writes in the verify function?
- Are you using `HyperlaneConnectionClient` just to have a storage variable for the mailbox? In which case it might be preferable to not have that dependency.
- Why is `verify` permission to `onlyMailbox`?
- Are etherscan keys required? IMO etherscan lets you do 1 request every 5 seconds which should maybe be fine?
- It would be nice to link to a deployment somewhere (for both the service, and the feed on an EVM chain, and show it replicating via etherscan view calls)

solidity/contracts/mock/MockMailbox.sol
AlexBHarley: Yes, the Chainlink messages are too big

AlexBHarley: True... Yes I will revert this

nambrot:
@yorhodes called out that an alternate path for this implementation is to put the CCIP read middleware at the provider level and then have estimate gas to `mailbox.process` just automatically populate the metadata. However, that is probably more involved as there is no out of the box middleware that can be used for this purpose (yet). @AlexBHarley 's approach is sufficient as it enables ISM development without relayer changes, and is technically possible to be spec compliant, and we can always introduce the alternate way in the future

solidity/contracts/Mailbox.sol
```diff
@@ -30,7 +30,7 @@ contract Mailbox is
 
     // ============ Constants ============
 
-    // Maximum bytes per message = 2 KiB (somewhat arbitrarily set to begin)
+    // Maximum bytes per message = 3 KiB (somewhat arbitrarily set to begin)
```
nambrot, line 33: Undo this

solidity/contracts/interfaces/IInterchainSecurityModule.sol
```diff
@@ -9,7 +9,8 @@ interface IInterchainSecurityModule {
         LEGACY_MULTISIG,
         MERKLE_ROOT_MULTISIG,
         MESSAGE_ID_MULTISIG,
-        OPTIMISM
+        OPTIMISM,
```
nambrot, line 12: FYI https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2420/files#diff-7d0323c3aa18eb73dd096474183d65c5d2495f0ea62685db5e6f7117c96e63fcR12

AlexBHarley: Oh yep, which do we think will be merged first? This or that PR

rust/agents/relayer/src/msg/metadata/ccip_read.rs
AlexBHarley: For posterity, it seemed a little complicated to use [ethers-ccip-read](https://github.com/ensdomains/ethers-ccip-read) in the relayer. The types used internally don't line up terribly well, and for a simple GET/POST call I also didn't think it's worth bringing in an extra dep

solidity/contracts/interfaces/IInterchainSecurityModule.sol
nambrot: There is no rush on either of them, asked @yorhodes to take a peek at this one before we merge

rust/agents/relayer/src/msg/metadata/ccip_read.rs
yorhodes: I think it's a bit more complicated than implemented here
eg supporting recursive offchain lookups
happy to punt for now but this feels like a good place to minimize surface area

solidity/contracts/interfaces/isms/ICcipReadIsm.sol
```diff
@@ -0,0 +1,18 @@
...
+
+import {IInterchainSecurityModule} from "../IInterchainSecurityModule.sol";
+
+interface ICcipReadIsm is IInterchainSecurityModule {
+    /**
+     * @notice Reverts with the data needed to query information offchain
+     * and be submitted via the origin mailbox
+     * @dev See https://eips.ethereum.org/EIPS/eip-3668 for more information
+     * @param _message data that will help construct the offchain query
+     * @return bool Ignored
```
yorhodes, line 12: why have a return value at all?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,60 @@
...
+    uint8 public constant moduleType =
+        uint8(IInterchainSecurityModule.Types.CCIP_READ);
+
+    // ============ External Functions ============
+
+    /**
+     * @notice Reverts with the data needed to query information offchain before
+     * mailbox submission
+     * @param _message the Hyperlane encoded message
+     * @return bool Ignored
```
yorhodes, line 53: why?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -0,0 +1,60 @@
...
+ * @dev The AbstractCcipReadIsm provided by Hyperlane is left intentially minimalist as
+ * the range of applications that could be supported by a CcipReadIsm are so broad. However
+ * there are few things to note when building a custom CcipReadIsm.
+ *
+ * 1. `getOffchainVerifyInfo` should revert with a `OffchainLookup` error, which encodes
+ *    the data necessary to query for offchain information
+ * 2. For full CCIP Read specification compatibility, CcipReadIsm's should expose a function
+ *    that in turn calls `process` on the configured Mailbox with the provided metadata and
+ *    message. This functions selector should be provided as the `callbackFunction` payload
+ *    for the OffchainLookup error
```
yorhodes, line 24: IIUC we wouldn't need to do this "pointer indirection" because the `Mailbox.process` would bubble up the revert reason from the `ism.verify` call

solidity/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Core solidity contracts for Hyperlane",
   "version": "1.3.5",
   "dependencies": {
+    "@eth-optimism/contracts": "^0.6.0",
```
yorhodes, line 6: ?

yorhodes (approved):
big unlock for new ISMs 🚀

solidity/contracts/interfaces/isms/ICcipReadIsm.sol
AlexBHarley: Good point, no idea

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
AlexBHarley: Let me be a little pedantic, spec wise, for a second. There's two ways for CCIP Read contracts to work,

1. Call some function, ie. `getOffchainVerifyInfo` which describes how to get and submit the off chain info to a callback
2. Call the function you actually want to call, ie. `Mailbox.process`, which reverts in the absence of off chain data but describes how to get the data. Allowing the caller to then call the function again.

I opted for option 1 because the Mailbox is not CCIP Read aware, so we break a core invariant in CCIP Read that states `If the sender field does not match the address of the contract that was called, return an error to the caller and stop.`. The Mailbox would need to catch the ISM error and revert with its own `OffchainLookup` error in turn.

The Relayer doesn't respect the callback function anyway so it's kind of irrelevant. But IMO option 1 at least has the option of being totally spec compliant if we want

yorhodes: that checks out
I do still prefer option 2 in theory but requiring the Mailbox to be CCIP read aware is quite a departure from the expected scope

Issue #2350, opened 2023-06-07
Sealevel infra plan & tooling to deploy solana-supporting agents


tkporter:
We need the ability to load in Solana keys. These will only be `hexKeys` for the relayer - validator key can & should be configured the same way. Now with omniscient relayer, I believe the scaffolding for per-chain keys isn't there in `typescript/infra` anymore.. Maybe we need to introduce per-protocol keys or something?

We maybe want to do things in a separate context? Testing in RC is probably sufficient for now, but we should decide what our own infrastructure should look like for Solana / Nautilus - are we okay with using the `hyperlane` context?

Desired end state here is:
* the ability to run the `deploy-agents.ts` script and have it work with solana based chains
* Validators can be deployed that validate a sealevel chain
* Able to deploy a relayer that supports both sealevel and ethereum chains - i.e. the Ethereum signer may be an AWS KMS key, but the Sealevel signer would be a hexKey pulled from GCP secrets
* Decision on how we should run infra for Solana & Nautilus

asaj:
Add the notion of chain type to chain metadata
Modify our infra to create/use hex keys for sealevel type chains. Presumably stored in GSM

nambrot:
@tkporter @asaj is this ticket closed with #2402 or is there other work from this?

nambrot:
Closing as we are on RC

PR #2402, opened 2023-06-19
Sealevel chains should always use hex keys for relayers

asaj:
### Description

This PR adds support for protocol-type awareness in infra.

A new field `type` is added to chainMetadata. This field is inspected during key creation and relayer helm chart creation. Relayer keys for sealevel chains are set explicitly to always be hex keys. 

### Drive-by changes

None
### Related issues

- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2350

### Backward compatibility

Yes


typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -28,6 +34,7 @@ export interface ChainMetadata {
   /** Hyperlane domain, only required if differs from id above */
   domainId?: number;
   name: ChainName;
+  type: ProtocolType;
```
nambrot, line 37: Can we rename this to `protocol` or `protocolType`. As per https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1867, would like to harmonize `AgentChainSetupBase` and `ChainMetadata`, or was there a reason to use a different key for the same purpose value? Or @jmrossy did you call this `type` here as well?

typescript/infra/src/config/agent/relayer.ts
```diff
@@ -1,6 +1,7 @@
 import { BigNumberish } from 'ethers';
 
 import { ChainMap, chainMetadata } from '@hyperlane-xyz/sdk';
+import { ProtocolType } from '@hyperlane-xyz/sdk/dist/consts/chainMetadata';
```
jmrossy, line 4: Nit: import path should be @hyperlane-xyz/sdk

typescript/sdk/src/consts/chainMetadata.ts
jmrossy: I called it `protocol`. +1 for renaming to that :)

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -17,6 +17,12 @@ export enum ExplorerFamily {
   Other = 'other',
 }
 
+export enum ProtocolType {
+  Ethereum = 'ethereum',
```
jmrossy, line 21: I also called this 'Ethereum' because that's what the agent did but 'EVM' feels more correct. Now may be a reasonable time to rename it throughout if others agree. @tkporter @nambrot

typescript/infra/src/agents/key-utils.ts
```diff
@@ -25,8 +25,14 @@ export function getCloudAgentKey(
   chainName?: ChainName,
   index?: number,
 ): CloudAgentKey {
-  if (agentConfig.aws && role !== Role.Deployer) {
-    // The deployer is always GCP-based
+  let isAws = !!agentConfig.aws;
+  // The deployer is always GCP-based
+  isAws = isAws && role !== Role.Deployer;
+  let isSealevel =
+    !!chainName && chainMetadata[chainName].type != ProtocolType.Sealevel;
```
jmrossy, line 32: Just double checking, is this right? && type is **not** equal to Sealevel?

typescript/sdk/src/consts/chainMetadata.ts
tkporter: imo the pain of a breaking change like this outweighs the benefit of the rename

typescript/sdk/src/consts/chainMetadata.ts
asaj: Protocol it is!

PR #2403, opened 2023-06-19
simple optimistic ism WIP

daragao:
This PR adds the SImple Optimistic ISM.

Most of the code is a clone from the Routing ISM since what we want to do is similar.


asaj:
Closing per instructions

PR #2414, opened 2023-06-21
Reclaim funds from old relayer keys

mattiecnvr:
### Description

Simple script to migrate funds from the old relayer keys to the new ones. This PR will probably not get merged but act as a historic record since this is a one-off script.

### Drive-by changes

### Related issues

- Fixes hyperlane-xyz/issues#468

### Backward compatibility

No

Yes

### Testing

Manual

### DIFF
--- a/typescript/infra/scripts/migrate-funds-from-aws-keys.ts
+++ b/typescript/infra/scripts/migrate-funds-from-aws-keys.ts
@@ -92,11 +92,11 @@ async function transferForEnv(ctx: Contexts, env: DeployEnvironment) {
       }
     }
     if (!errorOccurred) {
-      // await fromKey.delete();
-      // console.log('Deleted key', {
-      //   from: fromKey.identifier,
-      //   fromKey: fromKey.address,
-      // });
+      await fromKey.delete();
+      console.log('Deleted key', {
+        from: fromKey.identifier,
+        fromKey: fromKey.address,
+      });
     }
   }
 }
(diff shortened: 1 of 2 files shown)


typescript/infra/scripts/migrate-funds-from-aws-keys.ts
```diff
@@ -0,0 +1,164 @@
...
+  const gasToTransfer = await multiProvider.estimateGas(chain, transferTx);
+  const costToTransfer = gasToTransfer.mul(gasPrice);
+  if (costToTransfer.gt(currentBalance)) {
+    console.log('Not enough funds to transfer', {
+      ...logCtx,
+      balance: formatEther(currentBalance),
+    });
+    return;
+  }
+  transferTx.value = currentBalance.sub(costToTransfer);
```
tkporter, line 136: probably wanna set `transferTx.gasPrice = gasPrice` or something similar too, otherwise you can risk the auto-set gas price being higher than `gasPrice` and then the tx will revert bc you don't have enough funds to cover the cost + value

typescript/infra/scripts/migrate-funds-from-aws-keys.ts
```diff
@@ -0,0 +1,164 @@
...
+    finalBalance: formatEther(
+      await multiProvider.getProvider(chain).getBalance(fromKey.address),
+    ),
+    destinationBalance: formatEther(
+      await multiProvider.getProvider(chain).getBalance(toKey.address),
+    ),
+    cost: formatEther(costToTransfer),
+    transferTx,
+  });
+  // await fromKey.delete();
```
tkporter, line 151: I assume you'll send the transferTx here?

May wanna consider running things on a cheap chain with `transferTx.value = 1` first so you can confirm everything looks right. at least my paranoid self would 😄

mattiecnvr:
Closing since we should not actually need this code again. But this is at least a historical record of it.

PR #2434, opened 2023-06-26
Mattie/sealevel dependency conflicts fix

mattiecnvr:
### Description

_What's included in this PR?_

### Drive-by changes

### Related issues

- Fixes #[issue number here]

### Backward compatibility

Yes
No

None
Yes

### Testing

None
Manual
Unit Tests


rust/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -1,62 +1,70 @@
...
     utils::{get_account_metas, simulate_instruction},
     ConnectionConf, SealevelProvider,
 };
 
-use crate::RpcClientWithDebug;
-
-use self::contract::{
-    SerializableAccountMeta, SimulationReturnData, DISPATCHED_MESSAGE_DISCRIMINATOR,
-};
+const SYSTEM_PROGRAM: &str = "11111111111111111111111111111111";
```
tkporter, line 60: I recently learned about this macro https://docs.rs/solana-program/latest/solana_program/pubkey/index.html that lets you do:
```
const SYSTEM_PROGRAM: Pubkey = pubkey!("11111111111111111111111111111111");
```
May as well move over to that here while you're here so we don't have to do the `Pubkey::from_str`s in a few places

rust/hyperlane-core/src/types/mod.rs
```diff
@@ -139,3 +140,13 @@ impl TxCostEstimate {
         self.l2_gas_limit.unwrap_or(self.gas_limit)
     }
 }
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum IsmType {
```
tkporter, line 145: You can rm this type and use `ModuleType` in `rust/hyperlane-core/src/traits/interchain_security_module.rs` instead. But probably need to keep the borsh derivations

rust/patch.rs
```diff
@@ -0,0 +1,332 @@
...
+    version: &'static str,
+    /// Relative path to the monorepo root
+    path: &'static str,
+}
+
+#[derive(Copy, Clone)]
+enum Refspec {
+    Tag(&'static str),
+    Rev(&'static str),
+    // Branch(&'static str),
```
tkporter, line 183: nit: kept intentionally?

tkporter (commented):
Lovely, thank you so much!!!! this is great

Some questions/comments:
1. Looks like e2e tests may need to be updated to run the `./patch.rs` script?
2. Can you update the README to indicate you need `rust-script` and that you need to run `./patch.rs`?
3. Some brief docs on how one would update the patch / add a new patch would be nice too
4. I see there are a few `Cargo.old.*` files kept around that I assume were intended to be removed - e.e.g in `rust/sealevel` and `rust/sealevel/client`
5. What are the implications of this if a 3rd party crate wanted to use one of our crates? E.g. some of our Fuel contract tests that atm live in a separate repo depend on hyperlane-core, will that break now because of the need for patch.rs? https://github.com/hyperlane-xyz/fuel-contracts/blob/main/Cargo.toml#L34

mattiecnvr:
@tkporter Since you are taking this over a bit, I went ahead and removed the vendored directory but kept the patches, you will need those to make the forks. LMK if you have any issues applying them.

mattiecnvr:
This was finished and merged with #2505

PR #2435, opened 2023-06-26
Absorb token and helloworld submodules

yorhodes:
It is very painful to update any dependencies across repos and publish new releases. Fresh clones are also broken.

Removes token and helloworld submodules and merges in subrepos.

Followed https://nuclearsquid.com/writings/subtree-merging-and-you/ for preserving history.


.github/workflows/e2e.yml
```diff
@@ -29,9 +29,6 @@ jobs:
         with:
           submodules: recursive
 
-      - name: remove submodule locks
```
asaj, line 32: Is checkout with submodules, above, still relevant? Can't recall if we have foundry now as a submodule

.github/workflows/node.yml
```diff
@@ -18,9 +18,6 @@ jobs:
         with:
           submodules: recursive
 
-      - name: remove submodule locks
```
asaj, line 21: Ditto

rust/utils/run-locally/src/main.rs
```diff
@@ -297,7 +297,7 @@ fn main() -> ExitCode {
         // don't need to clean in the CI
         build_cmd(&["yarn", "clean"], Some("../"), None);
     }
-    build_cmd(&["yarn", "build:e2e"], Some("../"), None);
+    build_cmd(&["yarn", "build"], Some("../"), None);
```
asaj, line 300: Why not build:e2e anymore?

asaj (approved):
Thanks for doing this

.github/workflows/e2e.yml
yorhodes: yes, still needed for foundry

rust/utils/run-locally/src/main.rs
yorhodes: we don't need separate builds

.github/workflows/e2e.yml
yorhodes: you can see it in `.gitmodules`

PR #2437, opened 2023-06-27
Add markdown docs generation

yorhodes:
### Description

Adds markdown doc generation to solidity and token package

May be preferable to our gitbook contract API docs


typescript/token/package.json
```diff
@@ -35,7 +35,8 @@
   },
   "files": [
     "/dist",
-    "/contracts"
+    "/contracts",
+    "/docs"
```
jmrossy, line 39: This would cause the docs to get included in what gets published to NPM. Is that intentional?

typescript/token/remappings.txt
```diff
@@ -0,0 +1,3 @@
+@openzeppelin=../../node_modules/@openzeppelin
```
jmrossy, line 1: Does forge use this file?

yorhodes: yes

typescript/token/package.json
yorhodes: yeah this is how we are pulling into the docs repo atm 
https://github.com/hyperlane-xyz/docs/blob/main/package.json
probably not the ideal solution

jmrossy (approved):

PR #2446, opened 2023-06-28
Fix incorrect getRemoteInterchainAccount derivation

### LINKED ISSUE #2383: ICA remote ICA derivation seems incorrect
The docs currently recommend the use of `getRemoteInterchainAccount` to calculate the remote ICA address for a given local account, assuming they will call `callRemote` without overriding either the remote router or ISM.

However, it seems like, said derivation is incorrect. For https://explorer.hyperlane.xyz/message/0x10d0295107d4b550e8aa22598fb1df19b545fd35e0cc5e07f21c12081bacef1f, the sender (`0x6feAC9505615345Ed214ab6eEaEAF2f953bfCc01 `) the ICA should be `0x0f38ba177b317869E9E6f310e0c16C335f6d4C5F` (remoteInterchainAccount), in fact it is`0x0e6208650ab3b75e937a8935f7e2553b27693c21`

AlexBHarley:
### Description

A minor change to how we deploy InterchainAccountRouter's and a fix for remote ICA address derivation. 

The fix is outlined clearly in #2383 by Nam but essentially by assigning the `implementation` & `bytecode` variables in the constructor of `InterchainAccountRouter`, we are unable to correctly access them in `getRemoteInterchainAccount` if using a proxy.

By moving initialisation of these variables to `initialize` we correctly store the values in the proxy storage, not the implementation storage.

### Drive-by changes

The `InterchainAccountRouter` should no longer be used unproxied.

### Related issues

- Fixes #2383

### Backward compatibility

Yes

None

### Testing

- [ ] None
- [ ] Manual
- [x] Unit Tests


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -138,6 +127,11 @@ contract InterchainAccountRouter is
             _owner
         );
         require(localDomain == mailbox.localDomain(), "domain mismatch");
+
+        implementation = address(new OwnableMulticall(address(this)));
+        // cannot be stored immutably because it is dynamically sized
```
yorhodes, line 132: this is actually not true and we have tooling for this though might not be worth the hassle (in this PR or generally)
see https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/libs/MetaProxy.sol and usage in https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/fcb1bebb96816126a1f0abe01326edb516efddce/solidity/contracts/libs/StaticMOfNAddressSetFactory.sol#L10

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -67,29 +69,52 @@ contract InterchainAccountRouterTest is Test {
...
+        MockMailbox _mailbox,
+        IInterchainGasPaymaster _igps,
+        IInterchainSecurityModule _ism,
+        address _owner
+    ) public returns (InterchainAccountRouter) {
+        InterchainAccountRouter implementation = new InterchainAccountRouter(
+            _domain
+        );
+
+        TransparentUpgradeableProxy proxy = new TransparentUpgradeableProxy(
```
yorhodes, line 83: this is good for tests but if we want consistent ICA addresses on each chain we have to get a bit clever with CREATE2

yorhodes (commented):
thanks for doing this! I'm still not exactly sure this is the solution we want to ship

solidity/test/InterchainAccountRouter.t.sol
AlexBHarley: I'm not sure I follow, as far as I've seen ICA Router addresses and ICA addresses are not consistent across chains.

yorhodes: nevermind, we actually already made this decision to move away from consistent addresses, I just forgot

yorhodes (approved):

PR #2453, opened 2023-06-30
fix(contract_sync): saturating math in eta calculator

daniel-savu:
### Description

It was observed that this function can be called with a `last_block` that is greater than the `currenct_block`, which results in an underflow. Use saturating math to prevent this.

### Backward compatibility

Yes

None

### Testing

None


tkporter (approved):
LGTM

Issue #2426, opened 2023-06-22
unable to get max cost estimate validator announce tx


nambrot:
https://discord.com/channels/935678348330434570/984123861144600587/1121052248487763979

Link to error https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/1ce08cdbd5173b0a951b082b7b5f6919562a9d12/rust/chains/hyperlane-ethereum/src/validator_announce.rs#L145

It looks like on some chains, there is difficulty to get the max cost estimate for the validator announce tx. This should be handled more gracefully. I suspect this is because either gasAmount or gasPrice are not actually set as expected. If its not possible to estimate, IMO the tx should just be submitted, and if it fails, surface the RPC error (and hope that it shows something a la "not enough balance)

PR #2454, opened 2023-06-30
fix(validator): announce directly if no gas estimate

daniel-savu:
### Description

It has been reported that `contract_call.tx.max_cost()` can fail / not exist on some networks. When that's the case, default to assuming no additional validator gas is required and just send the tx. If the tx fails, report the failure and suggest that gas may be insufficient in the logged error.

### Drive-by changes

None

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2426

### Backward compatibility

Yes

None

### Testing

None

### DIFF
--- a/rust/agents/validator/src/validator.rs
+++ b/rust/agents/validator/src/validator.rs
@@ -213,14 +213,14 @@ impl Validator {
                         hash=?outcome.txid,
                         gas_used=?outcome.gas_used,
                         gas_price=?outcome.gas_price,
-                        "Transaction attempting to announce validator reverted"
+                        "Transaction attempting to announce validator reverted. Make sure you have enough ETH in your account to pay for gas."
                     );
                 }
             }
             Err(e) => {
                 error!(
                     error=?e,
-                    "Failed to announce validator"
+                    "Failed to announce validator. Make sure you have enough ETH in your account to pay for gas."
                 );
             }
         }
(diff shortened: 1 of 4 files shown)


yorhodes (approved):
thank you for doing this!

rust/agents/validator/src/validator.rs
```diff
@@ -204,6 +205,27 @@ impl Validator {
...
+                        hash=?outcome.txid,
+                        gas_used=?outcome.gas_used,
+                        gas_price=?outcome.gas_price,
+                        "Transaction attempting to announce validator reverted. Make sure you have enough ETH in your account to pay for gas."
+                    );
+                }
+            }
+            Err(e) => {
+                error!(
+                    error=?e,
```
mattiecnvr, line 222: Pretty sure we were standardizing on `err`.
```suggestion
            Err(err) => {
                error!(
                    ?err,
```

rust/chains/hyperlane-ethereum/src/validator_announce.rs
```diff
@@ -132,21 +133,24 @@ where
...
-                Ok(cost.saturating_sub(balance))
-            } else {
-                Err(ProviderError::CustomError("Unable to get announce max cost".into()).into())
-            }
-        } else {
-            Err(ProviderError::CustomError("Unable to query balance".into()).into())
-        }
+        let Ok(contract_call) = self
+            .announce_contract_call(announcement, None)
+            .await else {
```
mattiecnvr, line 141: So I know there is no official rustfmt standard on this just yet, but I would prefer if the `else` is on the next line when the first line gets split. At least until rust officially declares something.

Much cleaner than it was though!

mattiecnvr (approved):

PR #2461, opened 2023-07-03
Validators on RC

daniel-savu:
### Description

Allows for deploying validators in the RC context, on both `testnet3` and `mainnet2`

### Drive-by changes

Initially took the approach in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2372 to refactor out the validator keys in a dedicated file so it's easier to modify / add in the future. However, these are not expected to ever change so now the keys are hardcoded in `validators.ts`. There is a cyclic dependency here, because `scripts/get-key-addresses.ts` depends on the hardcoded addresses in `validators.ts` to know how many validator addresses to read from AWS, but to know which addresses to hardcode you need to first run the `get-key-addresses` script. This PR doesn't fix it but I'm mentioning it for future reference.

Also copy-pastes a change from one of @tkporter's open PRs in `typescript/infra/src/agents/aws/validator-user.ts`, to get around an aws bucket creation error.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/485

### Backward compatibility

Yes

None that I know of - but this PR does impact deployments in general. Maybe double check there's enough memory and other hardware resources to support the new validators?

### Testing

Manual - checked the kubernetes pods and S3 checkpoints.


daniel-savu:
Still need to add the mainnet keys and automate the generation of `keys.ts`

typescript/infra/config/environments/mainnet2/validators.ts
```diff
@@ -1,151 +1,157 @@
...
+        'ethereum',
+        count,
+      ),
+    },
+    avalanche: {
+      interval: 5,
+      reorgPeriod: 3,
+      validators: validatorsConfig(
+        {
+          [Contexts.Hyperlane]: ['0x4c327ccb881a7542be77500b2833dc84c839e7b7'],
```
aroralanuk, line 49: Should this hyperlane addresses match the ones you're updating, '0xa7aa52623fe3d78c343008c95894be669e218b8d' here?

typescript/infra/config/environments/mainnet2/validators.ts
```diff
@@ -1,151 +1,157 @@
...
+        'polygon',
+        count,
+      ),
+    },
+    bsc: {
+      interval: 5,
+      reorgPeriod: 15,
+      validators: validatorsConfig(
+        {
+          [Contexts.Hyperlane]: ['0x4c327ccb881a7542be77500b2833dc84c839e7b7'],
```
aroralanuk, line 81: rest of the chains too from here on

asaj (approved):
Nice!

PR #2466, opened 2023-07-03
Cosmos stubs

### LINKED ISSUE #2422: Stub out chains/cosmos in rust agents for integration work

mattiecnvr:
### Description

Adds stubs for Cosmos support.

CC: @yorhodes 

### Drive-by changes

A few minor changes to keep things consistent between the chain crates.

### Related issues

- Fixes #2422 

### Backward compatibility

Yes

None

### Testing

None

### DIFF
--- a/rust/chains/hyperlane-cosmos/build.rs
+++ b/rust/chains/hyperlane-cosmos/build.rs
@@ -1,0 +1,8 @@
+use std::fs;
+
+fn main() {
+    // TODO: build the cosmos contracts/bindings
+    fs::create_dir_all("src/contracts").expect("failed to create contracts dir");
+    fs::write("src/contracts/mod.rs", "// TODO: this should be generated")
+        .expect("failed to write contracts/mod.rs");
+}
--- a/rust/chains/hyperlane-cosmos/src/interchain_gas.rs
+++ b/rust/chains/hyperlane-cosmos/src/interchain_gas.rs
@@ -1,0 +1,44 @@
+use async_trait::async_trait;
+use hyperlane_core::{
+    ChainResult, HyperlaneChain, HyperlaneContract, Indexer, InterchainGasPaymaster,
+};
+use hyperlane_core::{HyperlaneDomain, HyperlaneProvider, InterchainGasPayment, LogMeta, H256};
+
+#[derive(Debug)]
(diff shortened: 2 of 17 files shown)


rust/chains/hyperlane-cosmos/build.rs
```diff
@@ -0,0 +1,8 @@
+use std::fs;
+
+fn main() {
+    // TODO: build the cosmos contracts/bindings
```
byeongsu-hong, line 4: ✅

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,46 @@
...
+    ChainResult, HyperlaneChain, HyperlaneContract, Indexer, InterchainGasPaymaster,
+};
+use hyperlane_core::{HyperlaneDomain, HyperlaneProvider, InterchainGasPayment, LogMeta, H256};
+
+/// A reference to a InterchainGasPaymaster contract on some Cosmos chain
+#[derive(Debug)]
+pub struct CosmosInterchainGasPaymaster {}
+
+impl HyperlaneContract for CosmosInterchainGasPaymaster {
+    fn address(&self) -> H256 {
```
byeongsu-hong, line 12: Cosmos currently has 43 / 63 length of the address, [and will be changed sometimes in future](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-028-public-key-addresses.md#basic-address). So, how can we infer the original length of the address now? Should we just have to decide with whether this address has 12 bytes (24 length of zeros) prefix?

byeongsu-hong: Ah, afaik most CosmWasm contract address has full length of address. So this could not be the critical problem

yorhodes: for ethereum we do a similar left padding of zeroes to convert between the 20 byte and 32 byte types
H256 and H160 have the `from` and `into` traits implemented

mattiecnvr:
My understanding is this PR is here as a reference and probably can be closed either now or in a couple months without any issues.

PR #2467, opened 2023-07-04
Sealevel inclusive ranges

### LINKED ISSUE #2279: Make event indexing sequence range inclusive
To match block range

mattiecnvr:
### Description

Standardized ranges for sealevel to be inclusive.

### Drive-by changes

Minor cleanup and modified some names to be more clear.

### Related issues

- Fixes #2279 

### Backward compatibility

Yes

None

### Testing

None


rust/hyperlane-core/src/traits/indexer.rs
```diff
@@ -20,18 +21,23 @@ pub enum IndexMode {
...
 /// An indexing range.
-#[derive(Copy, Debug, Clone)]
+#[derive(Debug, Clone)]
 pub enum IndexRange {
     /// For block-based indexers
-    Blocks(u32, u32),
+    BlockRange(SequenceRange),
     /// For indexers that look for specific sequences, e.g. message nonces.
-    Sequences(u32, u32),
+    MessageNonceRange(SequenceRange),
```
tkporter, line 36: Because this isn't just for outbound message indexing (i.e. we want to index processed messages based off a counter that's increased for each processed message, & we want something similar for IGP payments), imo SequenceRange makes more sense

tkporter (commented):
nit, I can make this change while you're out

tkporter:
I think this is good to merge but am holding off till the dependency stuff is in, just to prevent unnecessary conflicts

PR #2483, opened 2023-07-06
Fix hardhat test script in typescript/sdk/package.json

jaehunkim:
### Description

The previous `test:hardhat` script was not correctly finding and testing the `.hardhat-test.ts` files in the `src` directory. This commit updates the script to correctly find these files and pass them to the `hardhat test` command. Now, all Hardhat tests in the `src` directory and its subdirectories will be correctly run when the `test:hardhat` script is used.

### Backward compatibility

Yes


yorhodes:
@jaehunkim this might be a platform specific thinng

jmrossy (approved):
Tested locally and in CI, seems to work. Approved

Issue #2450, opened 2023-06-29
Yield-earning Warp Routes with ERC4626


nambrot:
### Context
Tokens that are held in Warp Route's `HypERC20Collateral` back the synthetic `HypERC20` tokens on the remote chains. A warp route deployer may want to create a warp route that puts these tokens to productive use. [ERC4626](https://ethereum.org/en/developers/docs/standards/tokens/erc-4626/) standard was created to allow yield-sources to easily be used without custom integration.

This bounty calls for a `HypERC20Collateral` variant that can deposit the underlying token into a 4626 vault (and withdraws it from the vault on unlock). When withdrawing, the user should get returned the original amount, with the yield accumulating in the warp route (which is tracked and withdrawable by the owner). The change itself should be pretty simple, so to complete this issue, you should also:

- Deploy a Demo of a DAI warp route that deposits the DAI into Savings DAI https://twitter.com/MakerDAO/status/1671220727322906624
- Deploy a demo of an ETH route that deposits the ETH into a 4626 wrapper around Lido https://github.com/superform-xyz/super-vaults/blob/main/src/lido/stETH_swap.sol

This is especially compelling as that allows let's say app-rollup/chain operators to monetize the deposits and capture value from a service they are providing. Obviously this is fundamentally already possible, but having ready to go contracts/products for this use case is key.

vinhyenvodoi98:
i would like to do this issue

vinhyenvodoi98:
so i need to create PR to this [Repo](https://github.com/hyperlane-xyz/hyperlane-token) and try to deploy by use [hyperlane-deploy](https://github.com/hyperlane-xyz/hyperlane-deploy), right ?

nambrot:
We actually moved Warp Routes back to the monorepo, so you should make a PR to the monorepo and then yes use it in hyperlane-deploy (you will need to deploy your own NPM package for that during development)

nambrot:
@vinhyenvodoi98 I would recommend reading up a bit more on the 4626 standard and how it works

vinhyenvodoi98:
i have read it but i am having a bit of trouble demo it on testnet environment.
like I can manually deploy the sDai contract and then use it. However, if they have sDai on testnet environment, that's would be perfect
I'm also having trouble implementing on the hyperlane-monorepo repo. I guess I need more time to understand it

nambrot:
You should be able to use any 4626 compatible token

[...]

PR #2499, opened 2023-07-09
feat: update contract HypERC20CollateralSaving

vinhyenvodoi98:
### Description

I Try to implement HypERC20CollateralSaving that can send token to saving vault (erc ERC4626)

### Drive-by changes

This PR is just PR to make sure I'm thinking in the right direction to solve the problem. 
I will update the final code soon

### Related issues

- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2450

### Backward compatibility

Yes

### Testing

not yet

### DIFF
--- a/typescript/token/contracts/HypERC20CollateralSaving.sol
+++ b/typescript/token/contracts/HypERC20CollateralSaving.sol
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.8.0;
+
+import {TokenRouter} from "./libs/TokenRouter.sol";
+import {Message} from "./libs/Message.sol";
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {ERC4626} from "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
+
+/**
+ * @title Hyperlane ERC20 Token Collateral that wraps an existing ERC20 with remote transfer functionality.
+ * @author Abacus Works
+ */
+contract HypERC20CollateralSaving is TokenRouter {
+    using SafeERC20 for IERC20;
+
+    IERC20 public immutable wrappedToken;
+    ERC4626 public immutable savingToken;
+
+    /**
+     * @notice Constructor
+     * @param erc20 Address of the token to keep as collateral
+     */
+    constructor(address erc20, address erc4626) {
+        uint256 amount = type(uint256).max; // Maximum value of uint256
+        wrappedToken = IERC20(erc20);
+        savingToken = ERC4626(erc4626);
+        savingToken.approve(erc4626, amount);
+    }
+
+    /**
+     * @notice Initializes the Hyperlane router.
+     * @param _mailbox The address of the mailbox contract.
+     * @param _interchainGasPaymaster The address of the interchain gas paymaster contract.
+     */
+    function initialize(address _mailbox, address _interchainGasPaymaster)
+        external
+        initializer
+    {
+        __HyperlaneConnectionClient_initialize(
+            _mailbox,
+            _interchainGasPaymaster
+        );
+    }
+
+    function balanceOf(address _account) external view returns (uint256) {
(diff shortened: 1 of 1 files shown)


vinhyenvodoi98:
@nambrot 
i try to solve issue [2450](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2450) but i don't know I'm really understanding it or not
Please give me some comment

typescript/token/contracts/HypERC20CollateralSaving.sol
```diff
@@ -0,0 +1,80 @@
...
+     * @dev Transfers `_amount` of `wrappedToken` from this contract to `_recipient`.
+     * @inheritdoc TokenRouter
+     */
+    function _transferTo(
+        address _recipient,
+        uint256 _amount,
+        bytes calldata // no metadata
+    ) internal override {
+        // redeem token from vault
+        uint256 amount = savingToken.redeem(_amount, address(this));
```
nambrot, line 76: redeem takes the `shares` quantity which might be different from the token `amount` quantity, so I believe you need to track that separately.

typescript/token/contracts/HypERC20CollateralSaving.sol
```diff
@@ -0,0 +1,80 @@
...
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {ERC4626} from "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
+
+/**
+ * @title Hyperlane ERC20 Token Collateral that wraps an existing ERC20 with remote transfer functionality.
+ * @author Abacus Works
+ */
+contract HypERC20CollateralSaving is TokenRouter {
```
nambrot, line 15: You should probably just extend from HypErc20Collateral

nambrot (commented):
I believe you need a bit more specific accounting/tracking around the shares in the vault

vinhyenvodoi98:
@nambrot 
I have updated contract, please check it out
because erc20 token will not store in Collateral contract so i add `uint256 totalCollateralAssets` to control that
when user transfer back it will calculate base on totalCollateralAssets and total of asset token of erc4626 token

vinhyenvodoi98:
I have deployed it successfully but still need to test it a bit

typescript/token/contracts/HypERC20CollateralSaving.sol
```diff
@@ -0,0 +1,79 @@
...
+     * @inheritdoc HypERC20Collateral
+     */
+    function _transferTo(
+        address _recipient,
+        uint256 _amount,
+        bytes calldata _metadata // no metadata
+    ) internal override {
+        uint256 redeemAmount = _convertToRedeemToken(_amount);
+        // redeem token from vault
+        uint256 amount = erc4626Token.redeem(
```
nambrot, line 55: I believe you are assuming that amount == shares which i don't think has to be true.

typescript/token/contracts/HypERC20CollateralSaving.sol
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @notice Constructor
+     * @param erc20 Address of the token to keep as collateral
+     */
+    constructor(address erc20, address erc4626) HypERC20Collateral(erc20) {
+        erc4626Token = ERC4626(erc4626);
+
+        // approve to erc4626
+        uint256 amount = type(uint256).max; // Maximum value of uint256
+        wrappedToken.approve(erc4626, amount);
```
nambrot, line 25: Is this a good idea?

typescript/token/contracts/HypERC20CollateralSaving.sol
vinhyenvodoi98: @nambrot 
it is quite difficult to implement in reverse
for example i send 10Dai from ethereum to polygon.
in polygon I transfer 3Dai to someone else. so who will get the interrest. 
if I use withdraw instead of redeem, the profit will be hard to calculate

vinhyenvodoi98:
@nambrot 
I think most reasonable way is we will use 2 variables, asset and share. 
Users can only take profit when the number of assets = 0

PR #2516, opened 2023-07-11
Add support for new signed checkpoint types in verify-validators

asaj:
### Description

This makes it easy to see who is running the latest version of the validator.


typescript/utils/src/validator.ts
```diff
@@ -21,64 +21,43 @@ export class BaseValidator {
     return domainHash(this.localDomain, this.mailbox);
   }
 
-  message(root: HexString, index: number) {
-    return ethers.utils.solidityPack(
-      ['bytes32', 'bytes32', 'uint32'],
-      [this.domainHash(), root, index],
-    );
+  message(checkpoint: Checkpoint, messageId?: HexString) {
+    return !!messageId
```
yorhodes, line 25: this is hard to read

yorhodes (approved):

PR #2535, opened 2023-07-18
Add permissionless context, mainnet2 zkevm validator

asaj:


rust/config/mainnet_config.json
```diff
@@ -128,6 +128,20 @@
       "index": {
         "from": 25900000
       }
+    },
+    "zkevm": {
```
nambrot, line 132: Does this set a weird precedent? We don't really want folks to be adding their config here right? I was imagining that we would track this config somewhere in `infra` to signify that

asaj: I don't think we have a good answer, and this was easiest

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -689,6 +689,23 @@ export const zbctestnet: ChainMetadata = {
   isTestnet: true,
 };
 
+export const zkevm: ChainMetadata = {
```
jmrossy, line 692: I agree with Nam's point from his review, I'm not certain that including a PI chain here as a 'core' chain is the right thing. For example, it's inclusion here will fool the Explorer/Warp UI into thinking it's not a PI message, and expect it to have scraped db data. Will that be the case? 

If you're aware of the downside and still prefer to include it, I'm okay with it

rust/config/mainnet_config.json
nambrot: Why can't we just track this config in `infra` and then insert it via helm?

PR #2542, opened 2023-07-20
chore: Use templating for datasource in grafana dashboard file

josedev-union:
### Description

This PR uses templating for datasource. i.e. no need to keep <prometheus> inside the dashboard json file which should be manually replaced by users.


mattiecnvr (approved):

mattiecnvr:
Sorry for the long wait, approved.

PR #2551, opened 2023-07-20
E2E Sealevel Prep

mattiecnvr:
### Description
This PR hardens the e2e test pipeline to failure cases we have seen frequently and also makes the code more generic to better support future environments such as sealevel.

The original ticket describes making one or more traits to help with this but after much experimentation and consideration, I feel that doing so would make the code unnecessarily complex and less flexible to future changes. My solution instead after talking with @tkporter regarding the sealvel e2e test pathway is to:
1) get utility code out of the way to make it more clear what the execution path is
2) simplify the process of running build tasks in the background so that as we add more envs we can have multiple things running concurrently without difficulty
3) standardize our program execution pattern to reduce the understanding overhead of the main execution path

The main thing I kept running into every time I tried to create traits is that we ultimately need a lot of env and context which means either arg passing or even more structures to hold them or we statically code it into that module. The last option was the cleanest but at that point there isn't much difference between creating a trait and just defining a function to call. And even then there is still some live data we define as we go in main that is bulky to pass around but it might become worth it depending.

Also the amount of support code for traits kept trying to ballon because we would need something that constructs the "components", and then runs each of the stages, but it also would need to still know what order to run things in, possibly even within each of those steps.

### Drive-by changes

- Now watches agents to detect early termination
- Standardizes log prefixes to distinguish between e2e logs and sub process logs
- Now logs what commands are run
- Standardized the program args concept to reduce param passing and better support reusable configurations
- Has all build commands run asynchronously by default to prevent `spawn` commands in main
- Fixed an issue where some rust components were being built twice, once as debug and again as release

### Related issues

- Fixes hyperlane-xyz/issues#501
- Should help debug recent flakiness if not resolve it

### Backward compatibility
Yes

### Testing
Manual


rust/utils/run-locally/src/main.rs
```diff
@@ -591,189 +483,31 @@ fn termination_invariants_met(num_expected_messages: u32) -> Result<bool> {
...
-    let mut command = Command::new(format!("target/debug/{}", name));
-    command
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .envs(env)
-        .args(args);
-    spawn_cmd_with_logging(name, command, log_prefix, log_all, log_dir)
+fn filter_anvil_logs(_log: &str) -> bool {
+    // for now discard all anvil logs
+    false
```
tkporter, line 512: should this be true? Interpreted this as whether anvil logs should be silenced

tkporter (approved):
LGTM

For posterity, chatted about the tradeoffs of over-modularizing in the first pass and happy to wait to learn more w/ the sealevel e2e test

rust/utils/run-locally/src/main.rs
mattiecnvr: It is a filter function, so true means keep, false means discard. Follows the pattern of https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter

I will update docs on that though.

PR #2555, opened 2023-07-22
helm: Support different signer key per a chain in relayer

josedev-union:
### Description

This PR supports a different signer key per a chain in relayer.


rust/helm/hyperlane-agent/templates/relayer-external-secret.yaml
```diff
@@ -35,7 +35,7 @@ spec:
   {{- if eq .signer.type "hexKey" }}
   - secretKey: {{ printf "%s_signer_key" .name }}
     remoteRef:
-      key: {{ printf "%s-%s-key-relayer" $.Values.hyperlane.context $.Values.hyperlane.runEnv }}
+      key: {{ printf "%s-%s-key-%s-relayer" $.Values.hyperlane.context $.Values.hyperlane.runEnv .name }}
```
nambrot, line 38: If we merge this change, we would break our infrastructure with it. Can you make this configurable to support both ways?

nambrot (commented):
Thanks for the PR, just one suggestion

rust/helm/hyperlane-agent/templates/relayer-external-secret.yaml
josedev-union: @nambrot we are also gonna migrate from one signer key to multiple keys soon and not using this yet tbh.
But i think this migration just requires external secrets creation and no need any other rollout or changes in k8s applications.

josedev-union: btw, for the backward compatibility, we can introduce a new flag value to switch secret key naming but it seems not a smart solution

josedev-union: got it. I added a new flag value for backward compatibility. Please review again.

rust/helm/hyperlane-agent/templates/relayer-external-secret.yaml
```diff
@@ -35,7 +35,11 @@ spec:
   {{- if eq .signer.type "hexKey" }}
   - secretKey: {{ printf "%s_signer_key" .name }}
     remoteRef:
+  {{- if $.Values.hyperlane.relayer.commonSignerKey }}
```
nambrot, line 38: Maybe rename it to something like `usingDefaultSignerKey`

nambrot (commented):
Basically approved pending consistent naming to the use of defaultSignerKey

nambrot (approved):

PR #2556, opened 2023-07-22
fixes to update new version of sway

chlenc:


avious00:
hey @daniel-savu could you take a look at this?

rust/utils/abigen/src/lib.rs
```diff
@@ -1,12 +1,20 @@
...
 use std::collections::BTreeSet;
 use std::ffi::OsStr;
 use std::fs::{self, File};
 use std::io::Write;
 use std::path::Path;
 
+// use fuels_code_gen::ProgramType;
 use inflector::Inflector;
 
+// to avoid "use of undeclared crate or module `fuels_code_gen`" error
```
daniel-savu, line 10: You should avoid copying in dependency code as much as possible, there's probably a root cause for the error that needs to be fixed instead. Can you share steps to reproduce the problem?

rust/utils/abigen/src/lib.rs
```diff
@@ -1,12 +1,20 @@
-use fuels_code_gen::ProgramType;
 use std::collections::BTreeSet;
 use std::ffi::OsStr;
 use std::fs::{self, File};
 use std::io::Write;
 use std::path::Path;
 
+// use fuels_code_gen::ProgramType;
```
daniel-savu, line 7: Should remove commented out code

daniel-savu (commented):
@chlenc is this PR still relevant or can it be closed?

rust/utils/abigen/src/lib.rs
chlenc: The sway version of this pr can be already not latest, you can reuse this code to update hyperlane to beta-4

Issue #2503, opened 2023-07-10
Sealevel: Kathy messages are sent to/from sealevel chains


tkporter:
Partially blocked by https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2502 - I think we can start this ticket before that's finished though

* We decided in https://discord.com/channels/935678348330434570/935678739663192184/1126547994884460634 to keep this TS native
* We need to decide if we want to do some SDK-level refactoring to accommodate this, or if we should just keep the SDK mostly as-is and shove the sealevel stuff in the Kathy script
* Note key-wise, I don't 100% recall if kathy uses a hexKey or AWS KMS key, but sealevel requires hexKey. Take a look at what was done for the relayer to accommodate using AWS KMS keys for Ethereum & hexKey for sealevel
* Expectation is that Kathy can be ran in testnet3 (or mainnet2 too, if we have Sealevel deployments by the time work is being done here), and messages will be sent to & from sealevel chains

```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2587
- [ ] https://github.com/hyperlane-xyz/issues/issues/558
```

jmrossy:
# Sealevel support proposal

## Overview

We've so far taken a somewhat top-down approach to adding Sealevel support; the warp UI, hyp-deploy have been updated before the SDK and Utils packages. We can continue that trend with HelloWorld/Kathy or we can make a bigger upfront investment and redesign the SDK+Utils now.

## Option 1: Shortest path

Kathy is built around HelloWorld, which is just a thin wrapper around the App/Deployer/Checker abstractions in the SDK. So updating _just_ HelloWorld to be sealevel compatible wouldn't make sense. Sealevel support for Kathy would require a new, parallel HelloWorld version that live alongside the EVM one.

This option is still less work than Option 2 but feels short-sighted.

## Option 2: Multi-env SDK

Updating the common utilities (e.g. address encoder, msg formatter) and SDK primitives (e.g. MultiProvider, App, Checker, Deployer) feels like a daunting task but also seems like the natural next step for Sealevel support.

I won't lay out a complete plan just yet until we have agreement on this direction but structural changes would include:

- Create multi-env compatible address/msg utilities (some of this is done in warp UI already)
- Redesign MultiProvider to handle different provider/connection libs
- Decouple all SDK primitives from Ethers v5 contracts/factories/providers
[...]

PR #2529, opened 2023-07-13
Minimal multi-protocol SDK support

jmrossy:
### Description

- Create MultiProtocolProvider
- Create ProviderType enum and TypedProvider
- Add viem, solana, and ethers6 dev deps to SDK
- Upgrade Typescript and Eslint across monorepo (required for Viem)
- Break out metadata mgmt code into new ChainMetadataManager class
- Implement basic MultiProtocolApp and MultiProtocolRouterApp
- Create Evm and Sealevel router adapters 
- Refactor contract types into separate file to avoid circular dep

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2503

### Backward compatibility

Yes

PR #2559, opened 2023-07-25
Multi-protocol Kathy

jmrossy:
### Description

- Add support for multi-protocol to the HelloWorld app and Kathy service
- Create a new MultiProtocolCore

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2503

### Backward compatibility

Yes

### Testing

- [x]  Test new Kathy abstraction on existing EVM while we wait for Sealevel Kathy to be unblocked
- [x]  Test using local config stubs for both EVM and Sealevel


typescript/sdk/src/app/MultiProtocolApp.ts
```diff
@@ -54,33 +50,32 @@ export class BaseSealevelAdapter<
...
-
   constructor(
     public readonly multiProvider: MultiProtocolProvider<ContractAddrs>,
     public readonly logger = debug('hyperlane:MultiProtocolApp'),
   ) {
     super(multiProvider.metadata);
   }
 
+  // Subclasses should override this with more specific adapters
+  abstract protocolToAdapter(
```
jmrossy, line 68: Changing this from a member var to a method because of tricky errors in subclasses due to generic types

typescript/sdk/src/metadata/ChainMetadataManager.ts
```diff
@@ -286,4 +286,36 @@ export class ChainMetadataManager<MetaExt = {}> {
     }
     return new ChainMetadataManager(newMetadata);
   }
+
+  /**
```
jmrossy, line 290: Pulled this up from MultiProvider to the parent class here so MultiProtocolProvider can use it as well

typescript/sdk/src/metadata/ChainMetadataManager.ts
```diff
@@ -286,4 +286,36 @@ export class ChainMetadataManager<MetaExt = {}> {
...
+   * Create a new instance from the intersection
+   * of current's chains and the provided chain list
+   */
+  intersect(
```
jmrossy, line 294: Pulled this up from MultiProvider to the parent class here so MultiProtocolProvider can use it as well

typescript/sdk/src/core/MultiProtocolCore.ts
```diff
@@ -0,0 +1,71 @@
...
+import { MultiProtocolProvider } from '../providers/MultiProtocolProvider';
+import { TypedTransactionReceipt } from '../providers/ProviderType';
+import { ChainMap, ChainName } from '../types';
+
+import { EvmCoreAdapter } from './adapters/EvmCoreAdapter';
+import { SealevelCoreAdapter } from './adapters/SealevelCoreAdapter';
+import { ICoreAdapter } from './adapters/types';
+import { CoreAddresses } from './contracts';
+
+export class MultiProtocolCore extends MultiProtocolApp<
```
jmrossy, line 19: Ideally this would have been added in #2529 but I hadn't yet realized we'd need it for this

typescript/utils/src/objects.ts
```diff
@@ -27,8 +23,8 @@ export function objMap<
   K extends keyof M,
   O,
   I = ValueOf<M>,
->(obj: M, func: (k: K, v: I) => O): MappedObject<M, O> {
-  return Object.fromEntries<O>(objMapEntries(obj, func)) as MappedObject<M, O>;
+>(obj: M, func: (k: K, v: I) => O): Record<K, O> {
+  return Object.fromEntries<O>(objMapEntries(obj, func)) as Record<K, O>;
```
jmrossy, line 27: I'm not sure why we were using `MappedObject` before, this simpler version returns more accurate types from objMap calls

typescript/sdk/src/core/adapters/SealevelCoreAdapter.ts
```diff
@@ -0,0 +1,149 @@
...
+      );
+    }
+    const receipt = sourceTx.receipt;
+    const logs = receipt.meta?.logMessages;
+    if (!logs)
+      throw new Error('Transaction logs required to check message delivery');
+    const parsedLogs = SealevelCoreAdapter.parseMessageDispatchLogs(logs);
+    if (!parsedLogs.length) throw new Error('Message dispatch log not found');
+    return parsedLogs.map(({ destination, messageId }) => ({
+      messageId: Buffer.from(messageId, 'hex').toString('hex'),
```
tkporter, line 39: curious why the hex string -> buffer -> hex string conversion is needed?

typescript/sdk/src/core/adapters/SealevelCoreAdapter.ts
```diff
@@ -0,0 +1,149 @@
...
+      destinationMailbox,
+      messageId,
+    );
+    const connection = this.multiProvider.getSolanaWeb3Provider(destination);
+
+    await pollAsync(
+      async () => {
+        // If the PDA exists, then the message has been processed
+        // Checking existence by querying for balance
+        await connection.getBalance(pda);
```
tkporter, line 62: does this throw if an account has no balance? Fwiw I tried it locally and it just returned 0 for a random account without throwing

I can't think why anyone would want to do this but because anyone can transfer lamports to whatever account they want, just relying upon balance as a way to see if the PDA "exists" has the edge case of someone sending it some dust

Wdyt of https://solana-labs.github.io/solana-web3.js/classes/Connection.html#getAccountInfo, which returns null if an account has no balance and no data, and having the condition be `accountInfo !== null && accountInfo.data.length > 0`. The real condition of a PDA "existing" is more that there's some data there, which can only be put there by the program when it processes a message

typescript/sdk/src/router/MultiProtocolRouterApps.ts
```diff
@@ -50,10 +49,15 @@ export class MultiProtocolGasRouterApp<
   ContractAddrs extends RouterAddress = RouterAddress,
   IAdapterApi extends IGasRouterAdapter = IGasRouterAdapter,
 > extends MultiProtocolRouterApp<ContractAddrs, IAdapterApi> {
-  public override readonly protocolToAdapter = {
-    [ProtocolType.Ethereum]: EvmGasRouterAdapter,
-    [ProtocolType.Sealevel]: SealevelGasRouterAdapter,
-  };
+  override protocolToAdapter(
```
tkporter, line 52: Noticing this is exactly the same as the super's `protocolToAdapter`, is it necessary?

typescript/sdk/src/providers/MultiProtocolProvider.ts
```diff
@@ -39,10 +40,12 @@ export interface MultiProtocolProviderOptions {
 export class MultiProtocolProvider<
   MetaExt = {},
 > extends ChainMetadataManager<MetaExt> {
-  protected readonly providers: ChainMap<ProviderMap<TypedProvider>> = {};
+  // Chain name -> provider type -> provider
+  protected readonly providers: ChainMap<ProviderMap<TypedProvider>>;
```
tkporter, line 44: Do we expect to have multiple different types of providers for a single chain? e.g. viem and ethers

typescript/infra/scripts/utils.ts
```diff
@@ -196,6 +204,43 @@ export async function getMultiProviderForRole(
...
+  keys: ChainMap<string>,
+  chain: ChainName,
+  manager: ChainMetadataManager<any>,
+) {
+  const protocol = manager.getChainMetadata(chain).protocol;
+  if (protocol === ProtocolType.Ethereum) {
+    return new Wallet(keys[chain]).address;
+  } else if (protocol === ProtocolType.Sealevel) {
+    return Keypair.fromSecretKey(
+      base58ToBuffer(keys[chain]),
```
tkporter, line 237: flagging this should probably be hex, similar to what's done in `typescript/infra/src/agents/gcp.ts`. Probably is an opportunity to dedupe this too given the per-protocol logic will be exactly the same

typescript/infra/scripts/helloworld/kathy.ts
```diff
@@ -370,71 +393,102 @@ async function sendMessage(
...
+      txReceipt = {
+        type: ProviderType.EthersV5,
+        receipt,
+      };
+    } else if (tx.type === ProviderType.SolanaWeb3) {
+      // Utilize the new multi-protocol provider for non-evm chains
+      // This could be done for EVM too but the legacy MP has tx formatting utils
+      // that have not yet been ported over
+      const connection = app.multiProvider.getSolanaWeb3Provider(origin);
+      const payer = Keypair.fromSecretKey(base58ToBuffer(keys[origin]));
```
tkporter, line 447: ditto

typescript/infra/scripts/utils.ts
```diff
@@ -196,6 +204,43 @@ export async function getMultiProviderForRole(
   return multiProvider;
 }
 
+export async function getKeysForRole(
```
tkporter, line 207: May be worth noting this relates to private keys and therefore only works for key types that allow private keys being extracted. Same for `getAddressesForKey`

typescript/infra/src/agents/aws/key.ts
```diff
@@ -55,6 +55,10 @@ export class AgentAwsKey extends CloudAgentKey {
     this.region = agentConfig.aws.region;
   }
 
+  get privateKey(): string {
+    throw new Error('Private key unavailabe for AWS keys');
```
tkporter, line 59: nit typo: unavailabe

typescript/infra/scripts/utils.ts
```diff
@@ -196,6 +204,43 @@ export async function getMultiProviderForRole(
...
+  txConfigs: ChainMap<ChainMetadata>,
+  environment: DeployEnvironment,
+  context: Contexts,
+  role: Role,
+  index?: number,
+): Promise<ChainMap<string>> {
```
tkporter, line 213: Any thoughts on returning `Promise<ChainMap<CloudAgentKey>>` instead? I don't feel strongly, it's just that we already have some existing abstractions for keys that we've used and and here we opt to just handle the privateKey directly, which is a slight departure

tkporter (commented):
Mostly questions and apart from a few small things LGTM! Great job, it's very clean

typescript/infra/scripts/helloworld/kathy.ts
```diff
@@ -146,18 +159,18 @@ async function main(): Promise<boolean> {
...
-    context,
-    Role.Kathy,
-    undefined,
-    connectionType,
-  );
-  const igp = HyperlaneIgp.fromEnvironment(
-    deployEnvToSdkEnv[coreConfig.environment],
-    app.multiProvider,
-  );
+  // const coreConfig = getEnvironmentConfig(environment);
```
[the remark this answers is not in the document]
jmrossy: Thanks for calling this out but I intentionally do for now. I will leave a TODO to clean this up :)

typescript/infra/scripts/helloworld/kathy.ts
jmrossy: Actually this is intended to stay even after getCoreConfigStub is removed. Is there an alternative way you'd prefer we do the sealevel signing?

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -101,12 +95,22 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     return Promise.all(messages.map((msg) => this.waitForProcessReceipt(msg)));
   }
 
+  // TODO consider renaming this, all the waitForMessage* methods are confusing
```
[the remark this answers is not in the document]
jmrossy: I'm up for doing it now but not in this PR which is already quite big

typescript/sdk/src/core/MultiProtocolCore.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import { EvmCoreAdapter } from './adapters/EvmCoreAdapter';
+import { CoreAddresses } from './contracts';
+
+describe('MultiProtocolCore', () => {
+  describe('constructs', () => {
+    it('with constructor', async () => {
+      const multiProvider = new MultiProtocolProvider<CoreAddresses>({
+        ethereum: {
+          ...ethereum,
+          validatorAnnounce: ethers.constants.AddressZero,
```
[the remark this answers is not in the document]
jmrossy: No particular reason, just needed something in there for the test.

typescript/sdk/src/core/adapters/SealevelCoreAdapter.ts
jmrossy: To validate that it is valid hex. I'd rather fail early

typescript/sdk/src/providers/MultiProtocolProvider.ts
jmrossy: I don't necessarily expect it but it is supported. 
One scenario where we'd want that is in a UI that uses both viem and our SDK (b.c. our SDK uses ethers v5)

typescript/sdk/src/router/MultiProtocolRouterApps.ts
jmrossy: Good catch!

typescript/sdk/src/consts/environments/testnet.json
```diff
@@ -150,5 +150,22 @@
     "interchainAccountRouter": "0x9cA4A31af0f3a8fe2806599467912809D3e17ECB",
     "merkleRootMultisigIsmFactory": "0xA9999B4abC373FF2BB95B8725FABC96CA883d811",
     "messageIdMultisigIsmFactory": "0xCCC126d96efcc342BF2781A7d224D3AB1F25B19C"
+  },
+  "solanadevnet": {
```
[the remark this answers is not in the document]
jmrossy: I ended up changing this to fix the SDK tests

typescript/infra/scripts/utils.ts
jmrossy: Will do

tkporter (approved):

Issue #2511, opened 2023-07-11
Sealevel: refactor IndexMode and IndexRange usage, support sequence-based RateLimitedContractSyncCursor indexing for future IGP indexing


tkporter:
For context, before Sealevel we index events exclusively using block ranges. This is because the `eth_getLogs` RPC takes in a block range (`from` and `to`) and what event you're looking for, and it returns all events in that range.

With Sealevel (and Fuel, and probably other future alt exec envs), we don't have this luxury. Solana doesn't offer a great experience for event indexing out of the box. The solution we ended up going with is "sequence-based" indexing, meaning instead of the EVM version asking "give me all message dispatch events that occurred in the block range N and M", we ask "give me the message that was dispatched with nonce N, then give me message with nonce N + 1, etc". We can do this in Sealevel because entire message are actually written to on-chain storage, and are easily found by their nonce.

This change has already been made, but it was kinda shoehorned in in a not super elegant way.

Basically an enum to determine which indexing mode was created that's able to be set for a particular chain:
```
/// Indexing mode.
#[derive(Copy, Debug, Default, Deserialize, Clone)]
#[serde(rename_all = "camelCase")]
pub enum IndexMode {
    /// Block based indexing.
    #[default]
    Block,
    /// Sequence based indexing.
    Sequence,
}
```

And then the `IndexRange` enum is used by the indexing logic:
```
/// An indexing range.
#[derive(Debug, Clone)]
pub enum IndexRange {
    /// For block-based indexers
    BlockRange(RangeInclusive<u32>),
    /// For indexers that look for specific sequences, e.g. message nonces.
    SequenceRange(RangeInclusive<u32>),
}
```

But it's a bit ugly -- if you poke around `rust/hyperlane-base/src/contract_sync/cursor.rs`, you'll see that most of the logic is still very related to the old block-based indexing mode. You can also see in each impl of the Indexer trait (e.g. in `rust/chains/hyperlane-ethereum/src/mailbox.rs` here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/trevor/sealevel-validator-rebase/rust/chains/hyperlane-ethereum/src/mailbox.rs#L134-L141) that there's this sketchy runtime check to do with IndexRanges being of the expected type.

Concretely, we want:
[...]

Issue #2501, opened 2023-07-10
Sealevel: index gas payments in relayer


tkporter:
Blocked by #2217 

* We'll want to start correctly indexing gas payments in the relayer. This will probably be done in a similar manner to message indexing, though this depends heavily upon the approach taken in #2217

PR #2562, opened 2023-07-25
refactor(IndexMode): remove `IndexRange`, use default index mode per env

daniel-savu:
### Description

Removes the `index_mode` config field, default to whichever mode is currently supported by our existing envs. Doing this allows for safely removing runtime checks to see if the correct index mode was configured. Additional work is required to support multiple indexing modes per env. 

Check this discord thread for the full discussion: https://discord.com/channels/935678348330434570/961711527092682783/1133347733411987568

### Drive-by changes

- `IndexRange` is removed entirely and replaced by `RangeInclusive<u32>`
- Moves `is_arbitrum_nitro` into a bigger existing `impl HyperlaneDomain` block for clarity

### Related issues

This is part 1 of https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2511

### Backward compatibility

Yes

### Testing

Relies on existing e2e tests


rust/hyperlane-core/src/traits/indexer.rs
```diff
@@ -54,3 +43,10 @@ pub trait MessageIndexer: Indexer<HyperlaneMessage> + 'static {
     /// Return the latest finalized mailbox count and block number
     async fn fetch_count_at_tip(&self) -> ChainResult<(u32, u32)>;
 }
+
+#[async_trait]
+#[auto_impl(&, Box, Arc)]
+pub trait NonceIndexer<T>: Indexer<T> + 'static {
```
tkporter, line 49: nit: for consistency probably best to keep with "sequence" instead of nonce in situations where it doesn't necessarily only relate to message nonces

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
```diff
@@ -130,6 +125,17 @@ where
     }
 }
 
+#[async_trait]
+impl<M> NonceIndexer<InterchainGasPayment> for EthereumInterchainGasPaymasterIndexer<M>
+where
+    M: Middleware + 'static,
+{
+    async fn nonce_at_tip(&self) -> ChainResult<(u32, u32)> {
+        info!("Gas payment nonce indexing not implemented");
```
tkporter, line 134: panicking here because it should never be used in Ethereum (at least not rn) seems reasonable to me

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -74,3 +76,11 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
         Ok(1)
     }
 }
+
+#[async_trait]
+impl NonceIndexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
+    async fn nonce_at_tip(&self) -> ChainResult<(u32, u32)> {
+        info!("Gas payment indexing not implemented for Sealevel");
+        Ok((1, 1))
```
tkporter, line 84: Returning this for now makes sense to me - https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2501 tracks stitching it together. For now there's no concept of an IGP sequence number / nonce in Ethereum so there's not much else you can do atm

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -36,6 +36,11 @@ pub(crate) struct MessageSyncCursor {
     /// The next block that should be indexed.
     next_block: u32,
     /// The next nonce that the cursor is looking for.
+    /// In the EVM, this is used for optimizing indexing,
+    /// because it's cheaper to make read calls for the nonce than
+    /// to call `eth_getLogs` with a block range.
+    /// In Sealevel, historic queries aren't supported, so the nonce field
+    /// is used to query storage in sequence.
     next_nonce: u32,
```
tkporter, line 44: So i think you may have had this idea and either I'm forgetting what we didn't like about it, I didn't understand it correctly at the time, or I'm imagining things - so atm we have the MessageSyncCursor which already has all the logic to power either block-based indexing or sequence-based indexing, and it just so happens that it's built specifically for getting message nonces / its interface into the db is message specific (i.e. HyperlaneMessageStore)

Is there maybe an opportunity to maybe make this style of cursor less message-specific, so that we can use this type of cursor for IGP payments or message delivery indexing when the chain we're indexing supports the sequence mode? And then for IGP payment or message delivery indexing for chains like Ethereum that *don't* support sequence mode, we can still use the `RateLimitedContractSyncCursor`, but it only works with block ranges

If this is too out there / doesn't make sense I'm also happy with the path you're going down, just wanted to share

daniel-savu: I was considering this (e.g. using a supertrait to share logic between `MessageSyncCursor` and `RateLimitedContractSyncCursor`) as a follow up step once I get an initial working version, but could also try making it work at this stage. 


> for chains like Ethereum that don't support sequence mode, we can still use the RateLimitedContractSyncCursor

Isn't the point of #2511 to add sequence indexing to `RateLimitedContractSyncCursor` for IGP and message delivery, even if that means adding empty stubs to the EVM trait implementations? Is your suggestion to reuse the `MessageSyncCursor` on chains where sequence mode can be used?

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
```diff
@@ -130,6 +124,16 @@ where
     }
 }
 
+#[async_trait]
+impl<M> SequenceIndexer<InterchainGasPayment> for EthereumInterchainGasPaymasterIndexer<M>
+where
+    M: Middleware + 'static,
+{
+    async fn nonce_at_tip(&self) -> ChainResult<(u32, u32)> {
+        panic!("Gas payment nonce indexing not implemented");
```
tkporter, line 133: Maybe worth a quick comment in each of these impls that panic for future readers explaining why this is okay to panic here / why we need to implement this trait even if it's not used

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -74,3 +76,11 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
         Ok(1)
     }
 }
+
+#[async_trait]
+impl SequenceIndexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
+    async fn nonce_at_tip(&self) -> ChainResult<(u32, u32)> {
```
tkporter, line 82: nit: nonce -> sequence for consistency

rust/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -736,6 +733,14 @@ impl Indexer<H256> for SealevelMailboxIndexer {
     }
 }
 
+#[async_trait]
+impl SequenceIndexer<H256> for SealevelMailboxIndexer {
```
tkporter, line 737: This is for indexing deliveries right? Worth changing the log msg on 739, and maybe a comment about us implementing this when we build scraper support for sealevel

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
daniel-savu: have renamed everywhere except for the actual functions that use a nonce, e.g. `retrieve_message_by_nonce`, let me know if that makes sense

daniel-savu:
Still seeing some errors in the e2e test:
```
WARN hyperlane_base::contract_sync::cursor: Target nonce not found, rewinding, next_nonce: 3
<VAL2>     at hyperlane-base/src/contract_sync/cursor.rs:126
<VAL2>     in hyperlane_base::contract_sync::ContractSync with label: "dispatched_messages", domain: "test2"
<VAL2> 


<VAL2>   2023-07-28T19:07:20.764633Z  WARN hyperlane_base::contract_sync::cursor: Rewound to previous known message, block_number: 99
...
```

tkporter (approved):
couple tiny nits, LGTM! nice job

tkporter:
Did this pass the manual sealevel e2e test flow?

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
daniel-savu: Don't know why this change doesn't show up in the PR view, if you check my branch this has already been changed :/

PR #2567, opened 2023-07-26
Set a provider timeout

mattiecnvr:
### Description

After observing the behavior of kathy, it seems like the service hang due to a non-responsive request. This PR is a fairly minor change that specifies a default timeout and allows specifying it in the chain metadata.

This specifically targets the behavior of this line in ethers
https://github.com/ethers-io/ethers.js/blob/bcc4d8c8701cfe980742b7d33dc7b0ee4fc48fec/src.ts/providers/provider.ts#L1487

which seems to allow the application to hang indefinitely under certain cases if no timeout is set.

Word of warning on this PR. I don't personally have a ton of context for working on this and I might be misunderstanding the consequences of this change. I am hoping someone who knows more about the TS lib will be able to validate that this is both 1) a sane default and 2) a good action to take.

### Drive-by changes
None

### Related issues
Related to and documented observations here
https://discord.com/channels/935678348330434570/1133779194711646281

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/typescript/sdk/src/providers/MultiProvider.ts
+++ b/typescript/sdk/src/providers/MultiProvider.ts
@@ -30,7 +30,7 @@ const DEFAULT_RETRY_OPTIONS: RetryProviderOptions = {
   baseRetryMs: 250,
 };
 
-const DEFAULT_PROVIDER_TIMEOUT: number = 1000 * 60 * 5; // 5 min
+export const DEFAULT_PROVIDER_TIMEOUT: number = 1000 * 30; // 30s
 
 export function defaultProviderBuilder(
   rpcUrls: ChainMetadata['rpcUrls'],
--- a/typescript/sdk/src/providers/MultiProvider.ts
+++ b/typescript/sdk/src/providers/MultiProvider.ts
@@ -5,6 +5,7 @@ import {
   ContractTransaction,
   PopulatedTransaction,
   Signer,
+  ethers,
   providers,
 } from 'ethers';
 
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -29,16 +30,22 @@ const DEFAULT_RETRY_OPTIONS: RetryProviderOptions = {
   baseRetryMs: 250,
 };
 
+const DEFAULT_PROVIDER_TIMEOUT: number = 1000 * 60 * 5; // 5 min
```
jmrossy, line 33: Two things here:
1. Please export this from `src/index.ts`
2. 5 minutes seems very high. It's 5min for a single request? I expect if something doesn't return in 30s, it won't every succeed

jmrossy (commented):
Looks good just one comment :)

typescript/sdk/src/providers/MultiProvider.ts
mattiecnvr: My understanding is it is the timeout for waiting for a txn. See the linked code in the description.

jmrossy: I think your link in the description goes to Ethers v6, which is pretty new. We're using V5.

The timeout property starts [here in JSONRpcProvider](https://github.com/ethers-io/ethers.js/blob/v5.7/packages/providers/src.ts/json-rpc-provider.ts#L412), then down to [fetchJson](https://github.com/ethers-io/ethers.js/blob/v5.7/packages/web/src.ts/index.ts#L362), then [fetchData](https://github.com/ethers-io/ethers.js/blob/v5.7/packages/web/src.ts/index.ts#L91), where it uses a [default of 2 minutes](https://github.com/ethers-io/ethers.js/blob/v5.7/packages/web/src.ts/index.ts#L116) if no override is set. 

Is the intention to increase or decrease that default here?

mattiecnvr: I guess I was looking at the wrong code, these changes are not needed then.

mattiecnvr:
See other comments, but closing as not needed.

PR #2568, opened 2023-07-26
Fix scraper deploy failure caused by infra config drift

mattiecnvr:
### Description
The infra config should represent the currently deployed state. This brings it back into alignment. By deploying the version that was in the main branch I saw hard failures in the scraper.

https://cloudlogging.app.goo.gl/7w65L2Sin3cgKtw99

### Drive-by changes

None

### Related issues
https://discord.com/channels/935678348330434570/935679524534911007/1133827653338677310

### Backward compatibility
Yes

### Testing
Manual


mattiecnvr:
Apparently only the scraper is not being deployed with these 🤦‍♀️ 

We should look into how to better handle infra configs so the represent what is deployed.

PR #2569, opened 2023-07-26
Updating the relayer and validator images to 2023-07-25 version

aroralanuk:
### Description

- Updating both rc and hyperlane testnet3 and mainnet2 versions for relayers and validators to `ed7569d-20230725-171222` image which has the necessary changes for `AggregationISM` and `NullMetadataISM`

### Drive-by changes

None

### Related issues

- Release: https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/tag/agents-2023-07-25

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/infra/config/environments/mainnet2/agent.ts
+++ b/typescript/infra/config/environments/mainnet2/agent.ts
@@ -69,7 +69,7 @@ const hyperlane: RootAgentConfig = {
     connectionType: AgentConnectionType.HttpFallback,
     docker: {
       repo,
-      tag: '2deb9b8-20230602-205342',
+      tag: 'ed7569d-20230725-171222',
     },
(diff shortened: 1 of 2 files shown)


mattiecnvr:
I am concerned that we are not updating the scraper as well since we ship the agent images together. I think it sends a bad message if we release the agents but are only using/testing 2/3.

aroralanuk:
> I am concerned that we are not updating the scraper as well since we ship the agent images together. I think it sends a bad message if we release the agents but are only using/testing 2/3.

https://github.com/hyperlane-xyz/issues/issues/493 This is why I didn't update the scraper.

yorhodes (approved):

PR #2585, opened 2023-07-31
Sealevel igp indexing

daniel-savu:
### Description
 
Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2583

Indexes IGP payments related to the relayer's data pda address. Unless this address is specified in the config (`sealevel.relayer_account `), no filtering is applied and all IGP payments are stored in the local database.

### Drive-by changes

- Sets `HYP_BASE_GASPAYMENTENFORCEMENT` in `run-locally` for the relayer, to test that it correctly indexes the IGP payment before submitting the message
- A new config section (`sealevel`) is added to the relayer
- The `MessageIndexer` trait is replaced with `SequenceIndexer<HyperlaneMessage>`, renaming `fetch_count_at_tip` to `sequence_at_tip`. `SequenceIndexer` is now common to both the message and igp indexers.
- The `parse_addr` macro is modified so it can be reused when parsing the sealevel relayer address config too
- `rust/utils/sealevel-test.bash` is included because I was using it to test locally, but I can remove it if the sealevel e2e test already does all the steps there @mattiecnvr  
- Performs a `try_into` conversion that can be removed once https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2610 is done

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2501

### Testing

e2e tests but the pipeline is failing, likely fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2602


rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -1,29 +1,59 @@
...
+    pub fn new(conf: &ConnectionConf, locator: ContractLocator) -> Self {
+        // Set the `processed` commitment at rpc level
+        let rpc_client =
+            RpcClient::new_with_commitment(conf.url.to_string(), CommitmentConfig::processed());
+
         let program_id = Pubkey::from(<[u8; 32]>::from(locator.address));
+        let domain = locator.domain.id();
+        let data_pda = Pubkey::find_program_address(igp_program_data_pda_seeds!(), &program_id);
+
+        debug!(
```
mattiecnvr, line 49: Log data should not be in the log message to make GCP queries/filtering easier.

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
```diff
@@ -136,7 +136,8 @@ where
         // TODO: if `SequenceIndexer` turns out to not depend on `Indexer` at all, then the supertrait
         // dependency could be removed, even if the builder would still need to return a type that is both
         // ``SequenceIndexer` and `Indexer`.
-        panic!("Gas payment nonce indexing not implemented");
+        info!("Gas payment nonce indexing not implemented");
```
mattiecnvr, line 139: What is the case where we might hit this? It feels like returning 0 might lead to issues. Also might want a warn log instead of info if it indicates a misconfig.

mattiecnvr: Random thought... Why not let the chain inform what indexing method to use? At least for now I suspect that every chain will only support one method, why have the config specify it at all.

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -197,7 +196,8 @@ where
     M: Middleware + 'static,
 {
     async fn sequence_at_tip(&self) -> ChainResult<(u32, u32)> {
-        panic!("Message delivery sequence indexing not implemented");
+        info!("Message delivery sequence indexing not implemented");
```
mattiecnvr, line 199: Same thought as before

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -49,12 +79,144 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+            sealevel_config,
+        }
+    }
+
+    async fn get_payment_with_sequence(
+        &self,
+        sequence_number: u64,
+    ) -> ChainResult<SealevelGasPayment> {
+        let payment_bytes = &[
+            &hyperlane_sealevel_igp::accounts::GAS_PAYMENT_DISCRIMINATOR[..],
```
mattiecnvr, line 123: Why the slice operators?

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -49,12 +79,144 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+            bytes: MemcmpEncodedBytes::Base64(payment_bytes),
+            encoding: None,
+        });
+        let config = RpcProgramAccountsConfig {
+            filters: Some(vec![memcmp]),
+            account_config: RpcAccountInfoConfig {
+                encoding: Some(UiAccountEncoding::Base64),
+                // Don't return any data
+                data_slice: Some(UiDataSliceConfig {
+                    offset: 1 + 8 + 8 + 32 + 4 + 32 + 8 + 8, // the offset to get the `unique_gas_payment_pubkey` field
```
mattiecnvr, line 145: Uhhh, I am worried about this breaking. If this temporary and if so please add a `todo` and if not, is this actually standard? I thought borsh was supposed to handle this type of thing.

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -49,12 +79,144 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+            .rpc_client
+            .get_program_accounts_with_config(&self.igp.program_id, config)
+            .await
+            .map_err(ChainCommunicationError::from_other)?;
+
+        // Now loop through matching accounts and find the one with a valid account pubkey
+        // that proves it's an actual message storage PDA.
+        let mut valid_payment_pda_pubkey = Option::<Pubkey>::None;
+
+        for (pubkey, account) in accounts.iter() {
```
mattiecnvr, line 163: Fun fact, instead of `.iter()` you can use `for _ in &accounts`

rust/chains/hyperlane-sealevel/src/utils.rs
```diff
@@ -77,3 +79,14 @@ pub async fn get_account_metas(
...
     Ok(account_metas)
 }
+
+pub async fn get_finalized_block_number(rpc_client: &RpcClientWithDebug) -> ChainResult<u32> {
+    let height = rpc_client
+        .get_block_height()
+        .await
+        .map_err(ChainCommunicationError::from_other)?
+        .try_into()
+        // FIXME solana block height is u64...
```
mattiecnvr, line 89: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2610
Should tag this new issue

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -518,17 +511,19 @@ where
...
+        let cursor_count = self.sync_state.next_sequence;
+        if cursor_count < payment_count {
+            if let Some(range) = self.sync_state.get_next_range(payment_count, tip).await? {
+                return Ok((CursorAction::Query(range), eta));
             }
-        };
-        Ok((action, eta))
+        }
+
+        // TODO: Define the sleep time from interval flag
```
mattiecnvr, line 525: Should we do this in this PR?

rust/utils/run-locally/src/main.rs
```diff
@@ -171,6 +171,22 @@ fn main() -> ExitCode {
...
         .hyp_env("RELAYCHAINS", "invalidchain,otherinvalid")
         .hyp_env("ALLOWLOCALCHECKPOINTSYNCERS", "true")
+        .hyp_env(
+            "GASPAYMENTENFORCEMENT",
+            r#"[{
+                "type": "minimum",
+                "payment": "1",
+                "matchingList": [
+                    {
+                        "originDomain": ["13375","13376"],
```
mattiecnvr, line 181: We might actually want to enforce the gas payment for all chains. We currently do a check for if the gas payment count is as expected, but it would test some extra logic to add this to at least 1 of the test EVM chains.

Also you should be able to write the numbers without quotes

rust/utils/sealevel-test.bash
mattiecnvr, line 1: This file should not be needed anymore

mattiecnvr (approved):
Approved assuming the E2E tests get resolved

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
daniel-savu: @tkporter actually had this idea but it seemed like a substantial task. As you suggested in the planning call today, I wanted to have the indexing just work at first, and only then do major refactoring but it could be a follow up task.

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
daniel-savu: Because `Concat` is implemented for slices but not for arrays, and there is no auto-coercion that can make this work AFAICT

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
daniel-savu: Added a todo comment, @tkporter gave me a reason for it at the time, whose mailbox [implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dce883d3a19b9727196c9d27db06d2e50f60850b/rust/chains/hyperlane-sealevel/src/mailbox.rs#L613) I followed. It may well have been just for the sake of speed and we may be able to determine the offset using borsh instead?

rust/chains/hyperlane-sealevel/src/utils.rs
daniel-savu: added as the last item in the "drive-by changes" section of the description 👍

rust/hyperlane-base/src/contract_sync/cursor.rs
daniel-savu: TBH I copied this comment from [another line](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dce883d3a19b9727196c9d27db06d2e50f60850b/rust/hyperlane-base/src/contract_sync/cursor.rs#L415) in this file. The interval flag is only present in the validator but these cursors are only used by the relayer and scraper I think? So we'd actually need to add the interval to the relayer settings struct

rust/utils/run-locally/src/main.rs
daniel-savu: I'm not sure which domain to set for the EVM ones actually. Is it always `31337` for anvil and could I use this one?

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -208,6 +239,8 @@ pub struct RawChainConf {
     metrics_conf: Option<PrometheusMiddlewareConf>,
     #[serde(default)]
     index: Option<RawIndexSettings>,
+    #[serde(default)]
+    sealevel: Option<RawSealevelConf>,
```
tkporter, line 243: I don't feel *super* strongly but I think leaking protocol-specific things into chain agnostic parts of the code like this is something we should avoid if possible

Originally I was thinking we'd just set `addresses.interchainGasPaymaster` to like `"${IGP program ID}-${relayer IGP account}"` but that's pretty confusing and also conflicts with the way we parse addresses / pubkeys, so I get why you avoided that

I have another idea where we can just specify the "relayer IGP account" and from there derive the IGP program ID. So we'd be able to remove this and just set the value of what's specified here in the existing `addresses.interchainGasPaymaster` place

We can make use of the fact that the IGP program ID is the owner of the "relayer IGP account", which is just a PDA.

So in `interchain_gas.rs` we can have a one-time step to get the IGP program ID by doing something like:
```
let igp_account = self
            .rpc_client
            .get_account_with_commitment(&igp_account_pubkey, CommitmentConfig::finalized())
            .await?
            .value;
let igp_program_id = igp_account.owner;
```

And we can store this in a OnceCell or something, or maybe just an Option and we have a function like `async fn get_igp_program_id() -> &Pubkey` or something like that that interior to it has logic to either use the cached version or actually make the RPC to get it

Curious what you think

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -63,24 +229,54 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
...
+                let sealevel_config = self.sealevel_config.clone().unwrap_or_default();
+                let beneficiary_filter = sealevel_config
+                    .relayer_account
+                    .unwrap_or(sealevel_payment.beneficiary.clone());
+                if beneficiary_filter == sealevel_payment.beneficiary {
+                    payments.push((sealevel_payment.payment, sealevel_payment.log_meta));
+                }
+            }
+        }
+        Ok(payments)
```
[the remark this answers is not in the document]
daniel-savu: Yeah, the way log fetching interacts with the cursor's sync state hasn't been changed. Only if the sequence wasn't found in the range are we [rewinding](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3c5c198f7a389a561f7e2e4fa55d82918aea01ee/rust/hyperlane-base/src/contract_sync/cursor.rs#L123)

rust/hyperlane-base/src/settings/chains.rs
daniel-savu: > leaking protocol-specific things into chain agnostic parts of the code like this is something we should avoid if possible

I agree in general, though if there are environment-specific settings I'd rather have them as separate optional config fields than make counterintuitive changes to the other agnostic fields. 

> Curious what you think

Your suggestion is great since we're reducing the setup complexity, so I'd much rather go for that.

If I have the IGP account as an `H256`, would I build the pubkey with `Pubkey::from(<[u8; 32]>::from(igp_account_h256))`?

rust/config/test_sealevel_config.json
```diff
@@ -5,9 +5,12 @@
       "domain": 13375,
       "addresses": {
         "mailbox": "692KZJaoe2KRcD6uhCQDLLXnLNA5ZLnfvdqjE4aX9iu1",
-        "interchainGasPaymaster": "FixmeFixmeFixmeFixmeFixmeFixmeFixmeFixmeFixm",
+        "interchainGasPaymaster": "GwHaw8ewMyzZn9vvrZEnTEAAYpLdkGYs195XWcLDCN4U",
         "validatorAnnounce": "DH43ae1LwemXAboWwSh8zc9pG8j72gKUEXNi57w8fEnn"
       },
+      "sealevel": {
+        "relayerAccount": "DrFtxirPPsfdY4HQiNZj2A9o4Ux7JaL3gELANgAoihhp"
```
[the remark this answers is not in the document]
daniel-savu: Now that this field is replacing `interchainGasPaymaster`, let me know if the latter should change

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -47,12 +97,137 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+        ]
+        .concat();
+        #[allow(deprecated)]
+        let payment_bytes: String = base64::encode(payment_bytes);
+
+        // First, find all accounts with the matching gas payment data.
+        // To keep responses small in case there is ever more than 1
+        // match, we don't request the full account data, and just request
+        // the `unique_gas_payment_pubkey` field.
+        #[allow(deprecated)]
```
tkporter, line 145: oop is this deprecated?

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -61,16 +236,82 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
...
+            unique_gas_payment_pubkey: expected_unique_gas_payment_pubkey,
+            slot: Default::default(),
+        }
+        .into(),
+    );
+
+    let serialized = gas_payment.into_inner().try_to_vec().unwrap();
+    let sliced_unique_gas_payment_pubkey = Pubkey::new(
+        &serialized
+            [(UNIQUE_GAS_PAYMENT_PUBKEY_OFFSET - 1)..(UNIQUE_GAS_PAYMENT_PUBKEY_OFFSET + 32 - 1)],
```
tkporter, line 311: minus 1 because of the initialized bool being the first byte when `UNIQUE_GAS_PAYMENT_PUBKEY_OFFSET` is used in getProgramAccounts?

if so, a comment for future readers would be nice

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -167,99 +241,115 @@ impl ContractSyncCursor<HyperlaneMessage> for ForwardMessageSyncCursor {
...
-        let range = match self.mode {
-            IndexMode::Block => BlockRange(from..=to),
-            IndexMode::Sequence => {
-                SequenceRange(next_nonce.saturating_sub(MAX_SEQUENCE_RANGE)..=next_nonce)
-            }
-        };
-
-        Some(range)
+        let (count, tip) = self.cursor.indexer.sequence_and_tip().await?;
+        self.cursor.sync_state.get_next_range(count, tip).await
```
tkporter, line 333: Does this return a correct range when sequence indexing? Looking at SyncState's `get_next_range` I don't see anything when the sync direction is backward that results in a backward sequence range being returned.

Wondering if things functionally still work because of the `self.cursor.sync_state.next_sequence = self.cursor.sync_state.next_sequence.saturating_sub(1);` up above, but instead of returning a range like [95, 100], then [90, 95], etc, what's returned is [99, 104], then [98, 103], etc (so the start range does go down, but messages end up being indexed multiple times)

Possible i'm misreading things though

tkporter (approved):
LGTM! Only thing that's important is my question about backward sequence indexing

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
daniel-savu: yep, it's copy-pasted from the mailbox indexing logic but that one deprecates warnings across [the entire file](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f1dabb6498dfed2a88fd743321a5418a095fef86/rust/chains/hyperlane-sealevel/src/mailbox.rs#L1)

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
daniel-savu: I think so, but it's not immediately apparent in the solana docs. Taking a closer look now

rust/sealevel/environments/local-e2e/sealeveltest1/core/program-ids.json
```diff
@@ -3,6 +3,7 @@
   "validator_announce": "DH43ae1LwemXAboWwSh8zc9pG8j72gKUEXNi57w8fEnn",
   "multisig_ism_message_id": "2YjtZDiUoptoSsA5eVrDCcX6wxNK6YoEVW7y82x5Z2fw",
   "igp_program_id": "GwHaw8ewMyzZn9vvrZEnTEAAYpLdkGYs195XWcLDCN4U",
+  "igp_program_data": "HBB1MxTeuRGb3C4bfXLQbu9z2Pn2LTXVQX53Hy7uyeof",
```
tkporter, line 6: were these manually added? if so, they'll be overwritten by the CLI - this file corresponds to https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f1c1910a77c70ec424655f799ea9475610e8432d/rust/sealevel/client/src/core.rs#L360-L373

If this isn't needed I'd recommend just removing it as the igp_program_data can be found like how we do here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2585/files#diff-4da632cca4500e4ca83baf054012f69be1d1298986be83c61211884067a28173R50

rust/utils/run-locally/src/config.rs
```diff
@@ -6,6 +6,7 @@ pub struct Config {
     pub ci_mode: bool,
     pub ci_mode_timeout: u64,
     pub kathy_messages: u64,
+    // TODO: Include count of sealevel messages in a field separate from `kathy_messages`?
```
tkporter, line 9: is this needed?

rust/utils/run-locally/src/invariants.rs
```diff
@@ -87,11 +98,14 @@ pub fn termination_invariants_met(
     .iter()
     .sum::<u32>();
     // The relayer and scraper should have the same number of gas payments.
-    if gas_payments_scraped != gas_payment_events_count {
+    // TODO: Sealevel gas payments are not yet included in the event count.
```
tkporter, line 101: hmm why is this the case?

rust/utils/run-locally/src/main.rs
tkporter: just to keep scope small i'm comfortable leaving this till later

rust/utils/run-locally/src/invariants.rs
```diff
@@ -15,7 +15,7 @@ pub fn termination_invariants_met(
     solana_config_path: &Path,
 ) -> eyre::Result<bool> {
     let eth_messages_expected = (config.kathy_messages / 2) as u32 * 2;
-    let sol_messages_expected = 1;
+    let sol_messages_expected = 15;
```
tkporter, line 18: Maybe worth pulling out into a const that can be used here and in the `for _i in 0..15 {` loop in main.rs?

tkporter (approved):
lgtm! some suggestions and questions

rust/chains/hyperlane-ethereum/src/interchain_gas.rs
```diff
@@ -130,6 +124,23 @@ where
...
+#[async_trait]
+impl<M> SequenceIndexer<InterchainGasPayment> for EthereumInterchainGasPaymasterIndexer<M>
+where
+    M: Middleware + 'static,
+{
+    async fn sequence_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
+        // The InterchainGasPaymasterIndexerBuilder must return a `SequenceIndexer` type.
+        // It's fine if only a blanket implementation is provided for EVM chains, since their
+        // indexing only uses the `Index` trait, which is a supertrait of `SequenceIndexer`.
+        // TODO: if `SequenceIndexer` turns out to not depend on `Indexer` at all, then the supertrait
```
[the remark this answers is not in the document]
daniel-savu: Not addressed yet, this comment is part of the conversation about chain-specific indexer logic

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -47,12 +97,138 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+        .concat();
+        #[allow(deprecated)]
+        let payment_bytes: String = base64::encode(payment_bytes);
+
+        // First, find all accounts with the matching gas payment data.
+        // To keep responses small in case there is ever more than 1
+        // match, we don't request the full account data, and just request
+        // the `unique_gas_payment_pubkey` field.
+        #[allow(deprecated)]
+        let memcmp = RpcFilterType::Memcmp(Memcmp {
```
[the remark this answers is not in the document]
daniel-savu: Yeah this logic is already somewhat duplicated with the one in `mailbox.rs`, but since there's just two uses I also think it's fine for now. Created issue https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2695

rust/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -47,12 +97,138 @@ impl InterchainGasPaymaster for SealevelInterchainGasPaymaster {}
...
+            igp_payment,
+            LogMeta {
+                address: self.igp.program_id.to_bytes().into(),
+                block_number: gas_payment_account.slot,
+                // TODO: get these when building out scraper support.
+                // It's inconvenient to get these :|
+                block_hash: H256::zero(),
+                transaction_id: H512::zero(),
+                transaction_index: 0,
+                log_index: U256::zero(),
```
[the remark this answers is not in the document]
daniel-savu: Much better idea, I didn't spot the opportunity to just add that here 👍

rust/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -682,45 +682,43 @@ impl SealevelMailboxIndexer {
...
-impl MessageIndexer for SealevelMailboxIndexer {
+impl SequenceIndexer<HyperlaneMessage> for SealevelMailboxIndexer {
     #[instrument(err, skip(self))]
-    async fn fetch_count_at_tip(&self) -> ChainResult<(u32, u32)> {
+    async fn sequence_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
         let tip = Indexer::<HyperlaneMessage>::get_finalized_block_number(self as _).await?;
         // TODO: need to make sure the call and tip are at the same height?
-        let count = self.mailbox.count(None).await?;
-        Ok((count, tip))
+        let count = self.mailbox.count(None).await;
```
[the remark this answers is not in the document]
daniel-savu: debugging residue

rust/utils/run-locally/src/config.rs
daniel-savu: Yep, because the sealevel message count is separate from evm message count and set in the test logic here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2585#discussion_r1310024392

rust/utils/run-locally/src/invariants.rs
daniel-savu: I think they're just counted separately in prometheus metrics?

rust/utils/run-locally/src/invariants.rs
daniel-savu: Good point! Since the `MAX_SEQUENCE_RANGE` is 100, we should probably be testing with hundreds rather than tens of messages, or have the sequence range be configurable so we get better cursor logic coverage in the e2e tests.

tkporter: I'm up for that as long as the e2e test run time doesn't change too much

PR #2612, opened 2023-08-05
Support 64-byte txn ids (attempt 2)

### LINKED ISSUE #2252: Sealevel agent: TxOutcome should support 64 byte tx ids

mattiecnvr:
### Description

Adds support to TxOutcome for 512-bit ids and to log meta to support Solana. Now that we have primitive types in our crate we were able to create a better solution than before which just stores the EVM 256bit hashes inside of the 512bit one with leading zeros rather than needing to over engineer an enum container for it.

### Drive-by changes

- Removed some dead code
- Improved primitive type support for 512bit types
- Improved serialization support for 512bit types
- Removed some of the features from the WASM check in hyperlane core. These features were not needed because WASM support is only for fuel at the current moment and it was breaking due to solana. See also #1977

### Related issues

- Fixes #2252 
- Replaces #2256

### Backward compatibility

Yes-ish, changes one of the internal database key strings. At standup we discussed the possibility of migrating the data and concluded that was a can of worms we did not want to open.

None

### Testing

Unit tests


rust/chains/hyperlane-sealevel/src/validator_announce.rs
```diff
@@ -117,10 +118,10 @@ impl ValidatorAnnounce for SealevelValidatorAnnounce {
         _tx_gas_limit: Option<U256>,
     ) -> ChainResult<TxOutcome> {
         warn!(
-            "Announcing validator storage locations within the agents is not supported on Sealevel"
+            "Announcing validator storage locations within the agents is not supported on Sealevel"
```
tkporter, line 121: nit: any idea why the space shows a diff?

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -26,7 +26,7 @@ const MESSAGE_DISPATCHED_BLOCK_NUMBER: &str = "message_dispatched_block_number_"
 const MESSAGE: &str = "message_";
 const NONCE_PROCESSED: &str = "nonce_processed_";
 const GAS_PAYMENT_FOR_MESSAGE_ID: &str = "gas_payment_for_message_id_v2_";
-const GAS_PAYMENT_META_PROCESSED: &str = "gas_payment_meta_processed_v2_";
+const GAS_PAYMENT_META_PROCESSED: &str = "gas_payment_meta_processed_v3_";
```
tkporter, line 29: nice that this was the only required change

tkporter (approved):
🚀 🚀 🚀  awesome

rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -80,7 +80,14 @@ impl HyperlaneSqlDb {
         log_meta: impl Iterator<Item = &LogMeta>,
     ) -> Result<impl Iterator<Item = TxnWithId>> {
         let block_hash_by_txn_hash: HashMap<H256, H256> = log_meta
-            .map(|meta| (meta.transaction_hash, meta.block_hash))
+            .map(|meta| {
+                (
+                    meta.transaction_id
+                        .try_into()
+                        .expect("Solana transaction ids are not supported yet"),
```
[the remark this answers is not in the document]
mattiecnvr: Yes, that is correct.

rust/chains/hyperlane-sealevel/src/validator_announce.rs
mattiecnvr: Yeah, it was a nbsp and I replaced it with a normal ascii space

PR #2617, opened 2023-08-07
Validator announce signers

### LINKED ISSUE #2457: Agents should indicate when private keys are not configured
When running agents that submit txs, if the operators does not specify a private key for the chain, this error message is typical:

```
The method eth_sendTransaction does not exist/is not available
```

That is because ethers-rs will attempt to use the RPC to send the transaction which will be unsuccessful for most RPCs. Instead, agents during startup should validate the configuration and ensure that either a default key is set or a chain-specific key is set.

### LINKED ISSUE #2458: self-Announce assumes that a chain signer is set
Validators specify the validator key via `--validator.key`. However, the auto-announce feature introduced in #2177 uses the standard contract wrappers which are configured via `--chains.[origin chain name].signer`. If the validator does not specify the chain signer, I believe the validator will try to use the RPC's signer to submit a tx, leading to an error like #2457

mattiecnvr:
### Description

This handles a bug where we did not default the chain signer to the validator signature for EVM chains and also handles the case where we might try to announce without setting a chain signer for other chains.

### Drive-by changes

Not really a change but I looked into #2457 and I noticed the validators already check that signers are set for all destination chains so I think the error was just for validators which this PR fixes.

### Related issues

- Fixes #2458
- Fixes  #2457

### Backward compatibility

Yes

### Testing

Unit Tests

### DIFF
--- a/rust/agents/validator/src/settings.rs
+++ b/rust/agents/validator/src/settings.rs
@@ -10,7 +10,7 @@ use hyperlane_base::{
     SignerConf,
 };
 use hyperlane_core::config::*;
-use hyperlane_core::HyperlaneDomain;
+use hyperlane_core::{HyperlaneDomain, HyperlaneDomainProtocol};
 
 decl_settings!(Validator,
     Parsed {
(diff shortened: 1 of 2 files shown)


nambrot:
>  I noticed the validators already check that signers are set for all destination chains so I think the error was just for validators which this PR fixes.

I assume you mean relayer here?

PR #2619, opened 2023-08-07
Rust updates

mattiecnvr:
### Description

Let this one slip a while. Updates our version of rust and as many packages as did not conflict with solana. I think we are going to need to figure something better out going forward with solana since if we have a couple more envs like this, it is going to make updates very hard.

### Drive-by changes

None

### Related issues

General Maintenance

### Backward compatibility

Yes

### Testing

Manual & Unit Tests


tkporter (approved):
lgtm

rust/hyperlane-base/src/metrics/core.rs
```diff
@@ -1,9 +1,8 @@
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
```
daniel-savu, line 3: nitpick: Do we actually need `OnceLock`? The stdlib provides its own `OnceCell` and I was wondering why not switch to that instead

daniel-savu (approved):

rust/hyperlane-base/src/metrics/core.rs
mattiecnvr: This is that. `std::cell::OnceCell` is declared `!Sync` which means we need `std::sync::OnceLock` which is equivlent to the old `once_cell::OnceCell`

PR #2620, opened 2023-08-08
fix: only run cargo fmt hook when rs files changed

daniel-savu:
### Description

The `cargo fmt` pre-commit hook runs regardless of whether any `.rs` files have changed and this is interfering with unrelated workflows. This reduces the scope of the hook to commits with nonempty rust diffs.


tkporter (approved):
nice

jmrossy (approved):

PR #2629, opened 2023-08-09
Fix insertion errors for txn and block

mattiecnvr:
### Description

This resolves the insertion conflict that is being caused by #2246 that introduced multiple concurrent tasks to scrape. This does not address the underlying concurrency issue I discovered, however. Please review the issue comments for more details.

### Drive-by changes

- Cleans up logs a bit (This was useful during testing and I think is useful globally)
- Removed an unused file

### Related issues

Fixes hyperlane-xyz/issues#493
Relates to #2631 

### Backward compatibility

Yes

### Testing

Manual


rust/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -150,20 +149,23 @@ impl HyperlaneSqlDb {
...
-            let mut cur_id = self.db.store_txns(txns_to_insert.drain(..)).await?;
-            for (_hash, (txn_id, _block_id)) in chunk.iter_mut() {
-                debug_assert!(cur_id > 0);
-                let _ = txn_id.insert(cur_id);
-                cur_id += 1;
+            self.db.store_txns(txns_to_insert.drain(..)).await?;
+            let ids = self.db.get_txn_ids(hashes_to_insert.drain(..)).await?;
+
+            for (hash, (txn_id, _block_id)) in chunk.iter_mut() {
+                let _ = txn_id.insert(ids[hash]);
```
daniel-savu, line 168: So we used to make the assumption that `Insert::many` in `sea_orm` works atomically and the inserted txn_ids are consecutive, when in fact they're not because of concurrent updates?

rust/agents/scraper/src/db/block.rs
```diff
@@ -96,11 +94,20 @@ impl ScraperDb {
...
         trace!(?models, "Writing blocks to database");
-        let first_id = Insert::many(models).exec(&self.0).await?.last_insert_id - id_offset;
-        debug_assert!(first_id > 0);
-        Ok(first_id)
+        match Insert::many(models)
+            .on_conflict(
+                OnConflict::column(block::Column::Hash)
+                    .do_nothing()
+                    .to_owned(),
+            )
```
daniel-savu, line 104: Is `on_conflict` needed here to avoid the panic of multiple insertion attempts for the same block?

daniel-savu (approved):

rust/agents/scraper/src/chain_scraper/mod.rs
mattiecnvr: Insert many does work atomically, the issue isn't that we have two simultaneous inserts but that we have two inserts with overlapping data. We used to make sure that we would not query the data if it was in the DB, but now we can't fully do that because

1) A checks database to see if h1, h2, and h3 are present
2) A finds out they are not
3) A begins querying for h1, h2, and h3 from the provider
4) B checks the database to see if h1, h4 are present
5) B finds out they are not
6) B begins querying for h1, h4 from the provider
7) A finishes querying for h1, h2, and h3 and inserts them
8) B finishes querying for h1 and h4 and fails to insert them because h1 is not a hash conflict

The issue is not that 7 and 8 happen together, it is because (1, 2, 3, 7) overlaps with (4, 5, 6, 8).

rust/agents/scraper/src/db/block.rs
mattiecnvr: See other explanation, but basically, we can either insert one at a time or just ignore duplicates. All this is saying is "If you already know this hash, disregard this specific block" It will however still insert all the ones that do not conflict.

This is required if multiple scrapers are running against the same database for the same chain, but this is not actually a supported configuration because there is no reason to do that. The only way to go back to how it was is to implement #2631

PR #2634, opened 2023-08-10
feat: Implement get_finalized_block_number for Cosmos chain

DavidNix:
### Description

Validator: Implements one method for the Indexer trait for Cosmos chains. 

Per the agent guide at https://docs.hyperlane.xyz/docs/protocol/spec#message-indexing

### Drive-by changes

None

### Backward compatibility

Yes

### Testing

Manual test for now which looked good. I'm not familiar with `reqwest` enough yet. Perhaps it allows me to mock out http responses.

### DIFF
--- a/rust/hyperlane-core/src/chain.rs
+++ b/rust/hyperlane-core/src/chain.rs
@@ -9,7 +9,7 @@ use num_traits::FromPrimitive;
 use strum::{EnumIter, EnumString, IntoStaticStr};
 
 use crate::utils::many_to_one;
-use crate::{HyperlaneProtocolError, IndexMode, H160, H256};
+use crate::{HyperlaneProtocolError, H160, H256};
 
 #[derive(Debug, Clone)]
 pub struct Address(pub bytes::Bytes);
@@ -104,6 +104,17 @@ pub enum HyperlaneDomain {
         domain_type: HyperlaneDomainType,
         domain_protocol: HyperlaneDomainProtocol,
     },
+}
+
+impl HyperlaneDomain {
+    pub fn is_arbitrum_nitro(&self) -> bool {
+        matches!(
+            self,
+            HyperlaneDomain::Known(
+                KnownHyperlaneDomain::Arbitrum | KnownHyperlaneDomain::ArbitrumGoerli,
+            )
+        )
+    }
 }
 
 #[cfg(any(test, feature = "test-utils"))]
@@ -347,24 +358,6 @@ impl HyperlaneDomain {
                 domain_protocol, ..
             } => *domain_protocol,
         }
-    }
-
-    pub fn is_arbitrum_nitro(&self) -> bool {
-        matches!(
-            self,
-            HyperlaneDomain::Known(
-                KnownHyperlaneDomain::Arbitrum | KnownHyperlaneDomain::ArbitrumGoerli,
-            )
-        )
-    }
-
-    pub const fn index_mode(&self) -> IndexMode {
-        use HyperlaneDomainProtocol::*;
-        let protocol = self.domain_protocol();
-        many_to_one!(match protocol {
-            IndexMode::Block: [Ethereum],
-            IndexMode::Sequence : [Sealevel, Fuel],
-        })
     }
 }
 
--- a/rust/agents/relayer/src/relayer.rs
+++ b/rust/agents/relayer/src/relayer.rs
@@ -266,11 +266,10 @@ impl Relayer {
         &self,
         origin: &HyperlaneDomain,
     ) -> Instrumented<JoinHandle<eyre::Result<()>>> {
-        let (index_settings, index_mode) =
-            self.as_ref().settings.chains[origin.name()].index_settings_and_mode();
+        let index_settings = self.as_ref().settings.chains[origin.name()].index.clone();
         let contract_sync = self.message_syncs.get(origin).unwrap().clone();
         let cursor = contract_sync
-            .forward_backward_message_sync_cursor(index_settings, index_mode)
+            .forward_backward_message_sync_cursor(index_settings)
             .await;
         tokio::spawn(async move {
             contract_sync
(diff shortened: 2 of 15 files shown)


rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,92 @@
...
+    ) -> ChainResult<Vec<(HyperlaneMessage, LogMeta)>> {
+        todo!()
+    }
+
+    /// Fetches the latest finalized block number from Cosmos API (aka LCD). Cosmos chains have fast finality.
+    async fn get_finalized_block_number(&self) -> ChainResult<u32> {
+        let url = self
+            .base_url
+            .clone()
+            .join("cosmos/base/tendermint/v1beta1/blocks/latest")
```
tkporter, line 70: Is there a Cosmos/Tendermint Rust SDK we could tap into for this & future calls / RPCs? Or is the most common practice in Cosmos-Rust to make http requests & parse them yourself?

Would be nice to not need to declare types like Block & BlockHeader etc if it's possible to use an existing library for it

DavidNix: There is not a Rust one that I would trust. Additionally, the hyperlane-cosmos modules have custom endpoints of which no client lib exists unless we build it. 

We could use GRPC, but:
1. Adds more complexity. 
2. Easier for a validator operator to run their fullnode if they don't have to worry about HTTP2. 
3. If the validator operator trusts a public fullnode, there are more public node operators that expose JSON API endpoints vs. GRPC.

Benefits of using `reqwest`:
1. No extra dependencies.
2. Consuming JSON is simple and straightforward.
3. We should only need a handful of endpoints.

DavidNix: What's the downside to the extra (private) types?

We could also use `serde_json::Value` but I thought the types were cleaner.

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,92 @@
...
+use std::time::Duration;
+use url::Url;
+
+use async_trait::async_trait;
+use eyre::Result;
+use hyperlane_core::{
+    ChainCommunicationError, ChainResult, HyperlaneMessage, Indexer, LogMeta, MessageIndexer,
+};
+
+/// Struct that retrieves event data for a Fuel Mailbox contract
```
mattiecnvr, line 13: Stale doc

rust/chains/hyperlane-cosmos/src/mailbox.rs
mattiecnvr: How commonly is gRPC supported? You say that not all nodes expose it... I personally am a huge fan of gRPC over REST + JSON because it means that we would have much stronger guarantees over data correctness without relying on e2e testing.

I understand that it can be a little bit of setup but, I think it is worth considering if it is widely supported.

Either way we should pull all the network calls into a client wrapper that allows us to add other functions later but also just avoid having to pass around the connection details and that can be responsible for handling the deserialization.

rust/hyperlane-core/src/chain.rs
```diff
@@ -165,6 +167,7 @@ impl HyperlaneDomainProtocol {
             Ethereum => format!("{:?}", H160::from(addr)),
             Fuel => format!("{:?}", addr),
             Sealevel => format!("{:?}", addr),
+            Cosmos => format!("{:?}", addr), // TODO: is this the right format?
```
mattiecnvr, line 170: Actually, since you ask, I am wondering if we meant to use `{:x}` formatting. But at least this would be consistent with what we have.

rust/chains/hyperlane-cosmos/src/mailbox.rs
mattiecnvr: Also if we continue without using gRPC, I agree that using private types for deserialization is nice. I would just like to see them more hidden away into a client module (possibly crate).

DavidNix: > How commonly is gRPC supported?

It depends on the chain. But typically, there are more node operators who offer the JSON endpoints. 

In this context IIUC, the validator operator is expected to run the node, so if we force GRPC, we force them to support HTTP2.

>  I personally am a huge fan of gRPC 

I am too for the same reasons. However, we (at Strangelove) have built several off chain programs that consume the JSON API that have worked without issue in perpetuity. Some background: The Cosmos JSON API uses [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway), so the JSON APIs are autogenerated from the grpc services and types. So GRPC (and all its guarantees) are still powering the JSON API. 

> Either way we should pull all the network calls into a client wrapper 

Happy to do this. However, I feel this is a premature abstraction at this stage. I don't have the full picture of where client calls are needed. I would imagine it would be straightforward to refactor into a client package later. 

Nam has made e2e testing the goal. So my strategy was to achieve that as quickly as possible and refactor to something more elegant later (if necessary).

DavidNix: In Go, I would place these modules under an `internal/exp` package to clearly signal this is experimental/POC work. Then move it out of that package once deemed stable. I'm not sure if there's similar concept in Rust.

nambrot: I like having you implement it in a lower-level first and then abstract later.

You are also right that ultimately, we should use the interface that is most common for chain nodes to expose. In the EVM land (as well as outside), there really is only JSON-RPC interface. We want to maintain the ability for validators to reuse existing nodes (either ones they run or public node operators like Infura or Alchemy)

nambrot: Fwiw Zaki recommended https://github.com/cosmos/cosmos-rust

DavidNix: I guess there's a 4th API that is websocket based where Tendermint can push information over websockets. But it does not work very well and most teams we know abstain from using it.

nambrot: We have similar conclusions for the websocket-based APIs in the EVM world.

Definitely agree on not going with GRPC for now, we could consider it once we are more sure of the right interfaces

Issue #2590, opened 2023-08-02
Sealevel audit: [L-2] - Enable runtime overflow checks


tkporter:
Add the following to all program Cargo.tomls:
```
[profile.release]
overflow-checks = true
```

The Mailbox is specifically called out because of `inbox_process` having `inbox.processed_count += 1`, but we may as well add this in all programs. I assume adding it to a program's Cargo.toml means that all dependencies of it will also have overflow checks, but we should confirm

daniel-savu:
> I assume adding it to a program's Cargo.toml means that all dependencies of it will also have overflow checks, but we should confirm

The profile config of a crate doesn't change based on the profile configs of the crates it imports if that's what you're asking. That's my read of the [docs](https://doc.rust-lang.org/book/ch14-01-release-profiles.html)
> Cargo has default settings for each of the profiles that apply when you haven't explicitly added any [profile.*] sections in the project’s Cargo.toml file. By adding [profile.*] sections for any profile you want to customize, you override any subset of the default settings.

PR #2636, opened 2023-08-10
fix(sealevel): use `overflow-checks` in the release profile

daniel-savu:
### Description

Uses `overflow-checks` in the release profiles of all sealevel programs and libraries, to revert txs when this happens.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2590


tkporter (approved):
tyvm!

Do you happen to know if it's only considered when specified in a program's Cargo.toml where it'd then have overflow checks for the entire program's bytecode (including all libs), or is it a per-dependency / per-lib thing?

daniel-savu:
> tyvm!
> 
> Do you happen to know if it's only considered when specified in a program's Cargo.toml where it'd then have overflow checks for the entire program's bytecode (including all libs), or is it a per-dependency / per-lib thing?

I think it's per-crate, also left a reply here: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2590#issuecomment-1673625297. Crates are compiled separately and then linked, and the cargo configuration of one crate doesn't apply to its upstream / downstream crates.

PR #2646, opened 2023-08-11
Deploy proteustestnet, nautilus, solanadevnet (again), and solana

tkporter:
### Description

just opening for image for now


rust/agents/validator/src/validator.rs
```diff
@@ -268,24 +260,33 @@ impl Validator {
...
-                if balance_delta > U256::zero() {
-                    warn!(
-                        tokens_needed=%balance_delta,
-                        validator_address=?announcement.validator,
-                        "Please send tokens to the validator address to announce",
-                    );
-                } else {
-                    let result = self
+
+                if self.core.settings.chains[self.origin_chain.name()]
```
tkporter, line 264: The commit that's running doesn't actually include this change. However I noticed that the fix that was added recently would just not even create the announcement.json if you don't have a signer set, which we don't want. Instead we want to still create an announcement.json and loudly complain while the validator still isn't announced

rust/sealevel/client/src/warp_route.rs
```diff
@@ -544,15 +544,20 @@ fn init_warp_route(
...
-    let interchain_gas_paymaster = token_config
-        .connection_client
-        .interchain_gas_paymaster
-        .clone()
-        .map(|config| (config.program_id, config.igp_account))
-        .unwrap_or((
-            core_program_ids.igp_program_id,
-            InterchainGasPaymasterType::OverheadIgp(core_program_ids.overhead_igp_account),
-        ));
+    // TODO for now not specifying an IGP for compatibility with the warp route UI.
```
tkporter, line 547: This is intentional - the warp route UI needs to craft specific accounts in order to call into the IGP, which it doesn't support yet. Supporting this will be as simple as:
1. update UI to specify certain accounts when the IGP a warp route is set
2. set the IGP in the warp route

rust/config/mainnet_config.json
```diff
@@ -128,6 +128,39 @@
...
+      "index": {
+        "from": 216377
+      }
+    },
+    "solana": {
+      "name": "solana",
+      "domain": 1399811149,
+      "addresses": {
+        "mailbox": "Ge9atjAc3Ltu91VTbNpJDCjZ9CFxFyck4h3YBcTF9XPq",
+        "interchainGasPaymaster": "FixmeFixmeFixmeFixmeFixmeFixmeFixmeFixmeFixm",
```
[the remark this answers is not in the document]
tkporter: I think 2585 will land before this, if it does I'll update

typescript/infra/scripts/agents/utils.ts
```diff
@@ -32,7 +32,8 @@ export class AgentCli {
     for (const role of this.roles) {
       switch (role) {
         case Role.Validator:
-          for (const chain of this.agentConfig.contextChainNames) {
+          for (const chain of this.agentConfig.contextChainNames[role]) {
+            if (chain !== 'solana' && chain !== 'nautilus') continue;
```
daniel-savu, line 36: If I remember correctly, my PR that was merged into this branch means this is no longer required?

daniel-savu (approved):

typescript/infra/scripts/agents/utils.ts
tkporter: yes ty!

PR #2647, opened 2023-08-11
feat: Add possibility to configure a subfolder inside an S3 bucket

gbouv:
### Description

This change adds the possibility to configure a subfolder inside an S3 bucket, to avoid having to spin up an entire bucket for hyperlane.

### Drive-by changes

Nothing

### Related issues

None, chatted about it on Slack with @nambrot 

### Backward compatibility

Yes

I made the field optional in the `RawCheckpointSyncerConf` so that current config are still valid and will continue to read/write at the root of the bucket. That being said, I'm new to this code so I hope I got it right

### Testing

None


nambrot:
Thanks for the PR @gbouv !

This is a great start, but I believe it's a bit more complicated than this:
- You need to modify the announcement that is generated of this, otherwise the validator will just announce without the folder https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/50f04db1faddb6d471b85386bb977fe9762753df/rust/hyperlane-base/src/types/s3_storage.rs#L212
- You need to modify the construction of the syncer based on the announcement https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/8c5983933b4517334323c38d47dd24236e4c88c4/rust/hyperlane-base/src/types/checkpoint_syncer.rs#L109

And then obviously we need to make sure that previous announcements without a folder continue to work, might be a good test case for y'all

rust/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -220,6 +220,6 @@ impl CheckpointSyncer for S3Storage {
     }
 
     fn announcement_location(&self) -> String {
-        format!("s3://{}/{}", self.bucket, self.region.name())
+        format!("s3://{}/{}/{}", self.bucket, self.region.name(), self.folder)
```
nambrot, line 223: What will this show if `self.folder` is `None`?

gbouv: It cannot be None right? I made the `folder` field in S3Storage a String, so worst case it's an empty string, which should technically work

rust/hyperlane-base/src/types/checkpoint_syncer.rs
```diff
@@ -106,13 +114,21 @@ impl FromStr for CheckpointSyncerConf {
...
         match prefix {
             "s3" => {
-                let [bucket, region]: [&str; 2] = suffix
+                let url_components = suffix
                     .split('/')
-                    .collect::<Vec<_>>()
-                    .try_into()
-                    .map_err(|_| eyre!("Error parsing storage location; could not split bucket and region ({suffix})"))?;
+                    .collect::<Vec<&str>>();
+                let mut folder = String::from("");
```
nambrot, line 120: I think this works, but IMO it would be more idiomatic to have `folder` be `None` instead of the empty string when parsing to avoid empty string equality matching from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2647/files#diff-929c46883a1aa134e41730ee6e9673b502e49d7c4ce6ddfcdbf86b17097f4813R223

nambrot (approved):
Actually nvm, i misinterpreted the code. I do think folder makes more sense as an option, but pending @tkporter or @mattiecnvr 's approval, lgtm

rust/hyperlane-base/src/types/checkpoint_syncer.rs
tkporter: +1 was gonna suggest this as well. And to avoid a situation where someone specifies the folder in the config as `""` where they probably mean they don't want a folder at all, to also default it to `None` around line 92

Think you could also avoid this extra `let mut folder` declaration here and add it into the destructuring below, e.g.:

```
let [bucket, region, folder]: [&str; 3] = match url_components.len() {
                    2 => Ok([url_components[0], url_components[1], None]),
                    3 .. => {
                        Ok([url_components[0], url_components[1], Some(url_components[2..].join("/").as_str())])
                    },
                    // ..
```

tkporter (commented):
Thanks for the PR, agree this is useful! I think it'd be nice to move to `folder: Option<String>`

gbouv:
BTW @nambrot I tested this with Kurtosis and it seems to work for both cases where `folder` is specified or left empty

gbouv:
@tkporter in my last commit I make `folder` and `Option<String>` so that you can see what it looks like. If you prefer it this way, I'll keep it, otherwise I'll revert the last commit. Let me know what you think :)

rust/hyperlane-base/src/types/checkpoint_syncer.rs
gbouv: Okay I just saw that on main the `RawCheckpointSyncerConf ` doesn't even anymore, so I guess your comment makes more sense. I'll merge main back in my branch make `folder` an `Option<String>`

tkporter (approved):
Lgtm, thank you!

gbouv:
@tkporter @nambrot can you help merging this one? I am not authorized to press the merge button

mattiecnvr:
Hi, I just saw this during a merge conflict. I have added changes to the schema in #2687 but this should not have been approved and represents an almost immediate divergence from the defined schema and parser support.

PR #2648, opened 2023-08-11
Fallback routing hook

aroralanuk:
### Description

- Added `FallbackDomainRoutingHook`  which fallbacks to `mailbox.defaultHook`

### Drive-by changes

- None

### Related issues

- v3

### Backward compatibility

Yes

### Testing

- None

### DIFF
--- a/solidity/contracts/hooks/FallbackDomainRoutingHook.sol
+++ b/solidity/contracts/hooks/FallbackDomainRoutingHook.sol
@@ -9,7 +9,11 @@ import {IPostDispatchHook} from "../interfaces/hooks/IPostDispatchHook.sol";
 // ============ External Imports ============
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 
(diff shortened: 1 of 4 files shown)


solidity/contracts/hooks/FallbackDomainRoutingHook.sol
```diff
@@ -0,0 +1,56 @@
...
+        override
+    {
+        try
+            hooks[message.destination()].postDispatch{value: msg.value}(
+                metadata,
+                message
+            )
+        {} catch {
+            mailbox.defaultHook().postDispatch(metadata, message);
+        }
```
yorhodes, line 54: we should catch the unconfigured hook before we do the external call
also why not unify the `postDispatch{value:msg.value}`calls?

aroralanuk: Screenshot 2023-08-14 at 3 34 30 PM

PR #2613, opened 2023-08-07
Sealevel IGP connection client support, warp routes support gas payments

tkporter:
### Description

* Adds IGP configuration to the connection client
* Adds gas router, which are used by warp routes to have configured gas amounts
* Adds resizing of the HyperlaneToken account to warp routes (useful for remote routers & gas routers which can change size)
* Adds IGP configuration to warp routes
* Pays for gas in warp routes is an IGP is configured
* Updates deploy tooling for warp routes to allow the IGP to be set, defaulting to the overhead IGP if one isn't configured

### Drive-by changes

* Removed the CLI command `token init`, which we never really expect to be used directly anymore and is just dead weight. This instead happens in `warp-route deploy` behind the scenes
* Moved IGP tests into their own crate to avoid a cyclical dependency situation between the IGP crate and test-utils. This is why `mailbox-test` exists too

### Related issues

* Fixes #2494 

### Backward compatibility

The storage layout of the warp route programs changes with this. So this is incompatible with existing sealevel testnet warp route deployments

### Testing

e2e test, unit tests

PR #2657, opened 2023-08-15
Add SetInterchainGasPaymaster instruction to warp route programs

tkporter:
### Description

Sadly I missed adding an instruction to allow a warp route owner to set the IGP. This adds that. We're deploying a warp route without an IGP for now because the UI hasn't been updated to work with IGP payments yet (which is required as the UI needs to handcraft specific accounts that are passed in). We'll probably wanna flip the switch to charging gas at some point, so want to make sure we're able to do that

The way the tests are set up involves a lot of code duplications - sorry 🙈 - we should revisit at some point as now isn't the time

### Drive-by changes

n/a

### Related issues

Related #2613 

### Backward compatibility

Fully backward compatible

### Testing

Unit / functional tests

### DIFF
--- a/rust/sealevel/programs/hyperlane-sealevel-token-collateral/tests/functional.rs
+++ b/rust/sealevel/programs/hyperlane-sealevel-token-collateral/tests/functional.rs
@@ -1604,4 +1604,172 @@ async fn test_set_interchain_security_module_errors_if_owner_not_signer() {
         result,
         TransactionError::InstructionError(0, InstructionError::InvalidArgument),
     );
+
+    // Also try using the non_owner as the payer and specifying the correct
+    // owner account, but the owner isn't a signer:
+    let recent_blockhash = banks_client.get_latest_blockhash().await.unwrap();
+    let transaction = Transaction::new_signed_with_payer(
+        &[Instruction::new_with_bytes(
+            program_id,
+            &HyperlaneTokenInstruction::SetInterchainSecurityModule(new_ism)
+                .encode()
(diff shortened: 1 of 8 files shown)


rust/sealevel/programs/hyperlane-sealevel-token-collateral/tests/functional.rs
```diff
@@ -1604,4 +1604,172 @@ async fn test_set_interchain_security_module_errors_if_owner_not_signer() {
...
+        recent_blockhash,
+    );
+    let result = banks_client.process_transaction(transaction).await;
+
+    assert_transaction_error(
+        result,
+        TransactionError::InstructionError(0, InstructionError::InvalidArgument),
+    );
+
+    // Also try using the non_owner as the payer and specifying the correct
```
daniel-savu, line 1752: Initially this looked to me like simply a duplicate of the final steps in `test_set_interchain_security_module_errors_if_owner_not_signer`, but I suppose this one is meant to test whether a non-owner can bypass the owner requirement _after_ an IGP was set, rather than the other test which has the check prior to this?

tkporter: The first one passes in an account that isn't the owner but is a signer, and the second one passes in an account that is the owner but is not a signer. Out of paranoia I want to test both cases to make sure the condition in the programs is that the owner is passed in *and* it's a signer

daniel-savu (approved):

PR #2659, opened 2023-08-15
Streamline Rust Agent Config Shapes

### LINKED ISSUE #2215: Streamline config shapes (rust)
This is the sister ticket to #2214 . It streamlines the existing chain config with the typescript one as they are so close to each other already. In addition to that streamlining benefit, it also captures the following benefits:

- It would specify a default RPC for the agent config (similar to SDK-specified default RPCs in `ChainMetadata`), which would potentially reduce one more argument to need to be specified via ENV vars to get started
- It moves the `rpcConsensusType` apart from `connection` and makes it easier to specify different defaults per agent type (i.e. fallback for relayer and scraper, quorum for validator) and thus obsoletes https://github.com/hyperlane-xyz/issues/issues/438
```[tasklist]
### Subissues
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/532
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2662
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2663
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2702
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2701
```

```[tasklist]
### Tasks
- [x] Update in accordance to spec https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2203/files#diff-0dda34e51b45fcb259676791b216443495e465f6ec7c372fc77ec9abe9b8409aR90
- [x] Split rpc consensus type from connection details
- [x] Update typescript agent definitons
- [x] Update Rust config definitions and parsing for base agent config
- [x] Update Rust config definitions and parsing for each individual agents config
- [x] Write up/propose rollout plan and decide on one
- [x] Update deployment config output
- [x] Update Helm and deploy tooling
- [x] Cleanup old typescript agent config defintions
- [x] Update e2e tests to use new configs
- [x] Verify generated deployment configs are correct
```
```[tasklist]
### PRs
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2658
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2665
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2666
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2687
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2659
```

mattiecnvr:
### Description
Continues the updates to the rust config shapes by updating deployment and runtime expectations. This PR also attempts to rely on the new single source of truth created by the schema in the SDK as much as reasonably possible which helped delete more code but also give some guarantees of consistency.

THIS IS A BREAKING CHANGE! It changes the config shapes the agents want and we should not merge this until we are ready.

Fixes #2215


avious00:
@tkporter for review

rust/hyperlane-base/src/settings/parser/json_value_parser.rs
```diff
@@ -26,7 +27,7 @@ impl<'v> ValueParser<'v> {
 
     /// Get a value at the given key and verify that it is present.
     pub fn get_key(&self, key: &str) -> ConfigResult<ValueParser<'v>> {
-        self.get_opt_key(key)?
+        self.get_opt_key(&key.to_case(Case::Flat))?
```
daniel-savu, line 30: Should we store `Case::Flat` in a global const?

rust/hyperlane-core/src/config/config_path.rs
```diff
@@ -76,10 +76,10 @@ impl ConfigPath {
 
     /// Get the environment variable formatted path.
     pub fn env_name(&self) -> String {
-        ["HYP", "BASE"]
+        ["HYP"]
             .into_iter()
             .chain(self.0.iter().map(|s| s.as_str()))
-            .map(|s| s.to_uppercase())
```
daniel-savu, line 82: It's unclear to me why some parsing logic uses flat case, while in here we use `Case::UpperFlat`

typescript/infra/config/environments/testnet3/agent.ts
```diff
@@ -117,7 +117,7 @@ const releaseCandidate: RootAgentConfig = {
   context: Contexts.ReleaseCandidate,
   rolesWithKeys: [Role.Relayer, Role.Kathy, Role.Validator],
   relayer: {
-    connectionType: AgentConnectionType.HttpFallback,
+    rpcConsensusType: RpcConsensusType.Fallback,
```
daniel-savu, line 120: How about `RpcPolicy` instead of `RpcConsensusType`? Or maybe `RpcConnectionType` to match the helm chart

daniel-savu (approved):

rust/hyperlane-base/src/settings/parser/json_value_parser.rs
mattiecnvr: I don't think that makes sense in this specific case because it would not support any other case.

rust/hyperlane-base/src/settings/parser/json_value_parser.rs
```diff
@@ -67,11 +69,35 @@ impl<'v> ValueParser<'v> {
...
                 cwp: &cwp + i.to_string(),
-            })),
+            }))
+            .map(|itr| Box::new(itr) as Box<dyn Iterator<Item = ValueParser<'v>>>),
+            Value::Object(obj) => obj
+                .iter()
+                .map(|(k, v)| k.parse().map(|k| (k, v)))
+                .collect::<Result<Vec<(usize, &'v Value)>, _>>()
+                .context("Expected array or array-like object where all keys are indexes; some keys are not indexes")
+                .map(|arr| arr.into_iter().sorted_unstable_by_key(|(k, _)| *k))
```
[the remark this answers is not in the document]
mattiecnvr: Added some comments

rust/hyperlane-core/src/config/config_path.rs
mattiecnvr: It is purely styalistic to display env vars as `HYP_KEY_TO_THING` instead of `HYP_key_to_thing`.

typescript/infra/config/environments/testnet3/agent.ts
mattiecnvr: I would rather get away from `rpcConnectionType` since it implies the old choice of `httpFallback, httpQuorum, http, ..`, and I went with `rpcConsensusType` at @tkporter's suggestion.

PR #2658, opened 2023-08-15
Agent Config Typescript Definitions

mattiecnvr:
### Description

Updates the typescript definitions based on the rust code. This is the first in a series of PRs for handling Rust config streamlining.

### Drive-by changes

- A couple of the core types I refined, please validate the definitions.

### Related issues

Progress on #2215 
Relates to #2491

### Backward compatibility

Yes

### Testing

Unit Tests

PR #2665, opened 2023-08-16
Agent Config Base Parsing

mattiecnvr:
### Description

This is a big one for #2215...

1. Splits the parsing of configs from the config definitions to support both the old, new, and future parsers simultaneously. (We might want to rename it from parser to config format or something; not set on this name).
2. Implements the new config format without switching to it or enabling it. This means this PR does introduce dead code that is going to be used in the next PR.

### Drive-by changes

- Deleted some unused (or single use) macros defined in hyperlane base
- Updated the definition of the config parsing trait to simplify implementation
- Added some more ergonomic methods of handling errors in config parsing
- Removed any raw `unwrap()` I noticed since these cause difficult to debug errors in prod.
- Added some additional validation to catch code mistakes in config path definitions

### Related issues

- Progress on #2215
- Comes after #2658

### Backward compatibility

Yes

### Testing
Manual (E2E only for now to make sure I did not break the old config format)


rust/hyperlane-core/src/config/mod.rs
```diff
@@ -122,3 +123,33 @@ impl Display for ConfigParsingError {
...
+/// calling this macro a, b, and c will be unwrapped and assigned to variables of the same name.
+#[macro_export]
+macro_rules! cfg_unwrap_all {
+    ($cwp:ident, $err:ident: $($i:ident),+$(,)?) => {
+        #[allow(unused_parens)]
+        let ($($i),*) = if let ($(Some($i)),*) = ($($i),*) {
+            ($($i),*)
+        } else if $err.is_ok() {
+            $err.push($cwp.clone(), eyre::eyre!(
+                    "Invalid configuration; seeing this error means we forgot to handle a specific error case before unwrapping it. Occurred when accessing ({})",
```
daniel-savu, line 141: Since this `else if` branch only prints an additional message compared to the `else` branch below, I think we could remove the duplication and also possibly reduce confusion about why the first logged parsing error has a slightly different error.

What do you think about making the `ConfigPath` in `ConfigParsingError` optional so that `"Invalid configuration; seeing this error means we forgot to handle a specific error case before unwrapping it."` could be passed in without a path and then just use the logic in the `else branch`?

typescript/sdk/src/metadata/matchingList.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * The types defined here are the source of truth for chain metadata.
+ * ANY CHANGES HERE NEED TO BE REFLECTED IN HYPERLANE-BASE CONFIG PARSING.
+ */
```
daniel-savu, line 4: Good idea to add this

rust/agents/validator/src/settings.rs
```diff
@@ -111,29 +111,31 @@ impl FromRawConf<'_, RawValidatorSettings> for ValidatorSettings {
...
         });
 
-        err.into_result()?;
-
-        let mut base = base.unwrap();
-        let origin_chain = origin_chain.unwrap();
-        let validator = validator.unwrap();
+        cfg_unwrap_all!(
+            cwp,
+            err: base,
```
daniel-savu, line 116: this macro syntax makes it seem like `base` is the only argument that has to do with `err`. How about something like this:
```rust
cfg_unwrap_all!(
            cwp,
            err: [
                base,
                origin_chain,
                validator,
                checkpoint_syncer,
                reorg_period
            ]
        );
        ```

daniel-savu (approved):
LGTM

rust/hyperlane-core/src/config/mod.rs
mattiecnvr: Pushed something along these lines.

PR #2666, opened 2023-08-16
Agent Config Parsing Casing Fixes

### LINKED ISSUE #2662: bug(rust): CLI argument parser doesn't work for some configurations.
~The following configurations are not parsed from either env var or cli args.~
Configurations whose key includes multiple words are not parsed from CLI args correctly. Here are typical ones.(edited)

```sh
chains.<chain_name>.addresses.interchainGasPaymaster
chains.<chain_name>.addresses.validatorAnnounce
```
Btw, when I use a config file, it works.
Here is a discord link for more details https://discord.com/channels/935678348330434570/984123861144600587/1139906822510219284

### LINKED ISSUE #2663: bug(rust): Environment variables are not parsed correctly for some configurations
Configurations whose key includes multiple words are not parsed correctly from environment variables.
For example, 
```text
HYP_BASE_CHAINS_${CHAIN_NAME}_ADDRESSES_INTERCHAIN_GAS_PAYMASTER for chains.<chain_name>.addresses.interchainGasPaymaster
HYP_BASE_CHAINS_${CHAIN_NAME}_ADDRESSES_VALIDATOR_ANNOUNCE for chains.<chain_name>.addresses.validatorAnnounce
```
The reason is because `load_settings_object` just uses `_` as separator.
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d88fd4cd16ebee06356c4b67e6c4ef381f33ff30/rust/hyperlane-base/src/settings/loader/mod.rs#L77-L90

mattiecnvr:
### Description

This is a fix for an issue in the arguments and environment where we are unable to correctly support attribute names due to the casing. The main issue cannot be fully fixed practically by this PR because it would cause some configuration cases to break. For now it adds manual exceptions for the command line argument case and leaves envs as they were. There are then two new parsers that can be used with the new config format.

### Drive-by changes

Also adds this to argument parser since it was trivial to copy/paste it and it will allow us to harden parsing when we switch over.

### Related issues

- Fixes #2662 
- Fixes #2663 
- Progress on #2215 

### Backward compatibility

Yes

### Testing

Unit Tests


rust/hyperlane-base/src/settings/loader/arguments.rs
```diff
@@ -77,12 +86,15 @@ impl Source for CommandLineArguments {
...
-
-            // If separator is given replace with `.`
-            if !separator.is_empty() && separator != "." {
-                key = key.replace(separator, ".");
-            }
+            let key = if let Some(case) = self.casing {
+                // if case is given, replace case of each key component
+                key.split(separator).map(|s| s.to_case(case)).join(".")
+            } else if !separator.is_empty() && separator != "." {
+                // If separator is given replace with `.`
```
daniel-savu, line 93: I think this comment still applies to the entire `let key` block since the first conditional branch also does `.join(".")`

rust/hyperlane-base/src/settings/loader/arguments.rs
```diff
@@ -77,12 +86,15 @@ impl Source for CommandLineArguments {
...
 
-            let mut key = key.to_lowercase();
-
-            // If separator is given replace with `.`
-            if !separator.is_empty() && separator != "." {
-                key = key.replace(separator, ".");
-            }
+            let key = if let Some(case) = self.casing {
+                // if case is given, replace case of each key component
+                key.split(separator).map(|s| s.to_case(case)).join(".")
```
daniel-savu, line 91: Seems like there's two things going on at the same time here:
- split tokens are being converted to `case`
- `separator` occurences are replaced with `"."`. This part is also handled by the `else if` branch

This would be easier to understand to me:
```rust
            // have `key` declared as mutable 
            if let Some(case) = self.casing {
                // if case is given, replace case of each key component
                key = key.split(separator).map(|s| s.to_case(case)).join(separator)
            }
            // If separator is given replace with `.`
            if !separator.is_empty() && separator != "." {
                key = key.replace(separator, ".")
            }
```

rust/hyperlane-base/src/settings/loader/deprecated_arguments.rs
daniel-savu, line 1: Sad to see this much duplication, I guess this is one time when Rust's "composition over inheritance" isn't that great

rust/hyperlane-base/src/settings/loader/deprecated_arguments.rs
```diff
@@ -0,0 +1,346 @@
...
+            // If separator is given replace with `.`
+            if !separator.is_empty() && separator != "." {
+                key = key.replace(separator, ".");
+            }
+
+            if key.ends_with("interchaingaspaymaster") {
+                key = key.replace("interchaingaspaymaster", "interchainGasPaymaster");
+            } else if key.ends_with("validatorannounce") {
+                key = key.replace("validatorannounce", "validatorAnnounce");
+            }
```
daniel-savu, line 93: Will we be able to remove `DeprecatedCommandLineArguments` once your big configuration rework is merged? Kind of unfortunate this is needed

daniel-savu (approved):

rust/hyperlane-base/src/settings/loader/arguments.rs
mattiecnvr: I considered that, but it is better to avoid allocating the extra string so I will instead add some docs to clarify what is happening.

rust/hyperlane-base/src/settings/loader/deprecated_arguments.rs
mattiecnvr: Honestly, I think in this case duplication is for the best. Dry code is good in theory but here we want these to be able to diverge without worrying about one impacting the other.

rust/hyperlane-base/src/settings/loader/environment.rs
```diff
@@ -0,0 +1,182 @@
...
+
+            let key = if let Some(case) = self.casing {
+                // if case is given, replace case of each key component
+                key.split(separator).map(|s| s.to_case(case)).join(".")
+            } else if !separator.is_empty() && separator != "." {
+                // If separator is given replace with `.`
+                key.replace(separator, ".")
+            } else {
+                key
+            };
```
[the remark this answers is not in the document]
mattiecnvr: First gut reaction was not worth, but we do actually use this specific logic 3 times so okay, I can see it.

rust/hyperlane-base/src/settings/loader/deprecated_arguments.rs
mattiecnvr: No, it will have to wait until we actually stop supporting the old config format.

PR #2671, opened 2023-08-17
Add `receive` fn to `HypNative` and `Scaled` variant

### LINKED ISSUE #2622: Allow ETH to be donated to a HypNative contract
The plan for bridging into Nautilus so that bridging gives a user the native Neon EVM token (i.e. ETH on that EVM) involves sending a bunch of ETH tokens to the HypNative contract without actually bridging the tokens. This ETH serves as a liquidity source for bridging *in* to that native warp route.

Atm, there is no fallback function on HypNative and the only payable function is `transferRemote` itself. We should add a function `function donate() external payable;` that anyone can call to donate funds to the HypNative contract

See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2621 for what this looks like in the stopgap version - need to add tests etc

yorhodes:
### Description

Adds an extension to `HypNative` that scales `TokenRouter` message `amount` by configurable `scale`

### Drive-by changes

Adds forge to token package

### Related issues

- Fixes #2622 

### Backward compatibility

Yes

### Testing

Unit Tests


typescript/token/contracts/extensions/HypNativeScaled.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.8.0;
+
+import {HypNative} from "../HypNative.sol";
+import {TokenRouter} from "../libs/TokenRouter.sol";
+
+/**
+ * @title Hyperlane Native Token that scales native value by a fixed factor for consistency with other tokens.
```
nambrot, line 8: Given how complicated it was for us to reason about the direction of the scale, I think a bit more documentation here would be appreciated

typescript/token/test/HypNativeScaled.t.sol
```diff
@@ -0,0 +1,148 @@
...
+        synth = new HypERC20(decimals);
+        synth.initialize(
+            address(environment.mailboxes(synthDomain)),
+            address(environment.igps(synthDomain)),
+            synthSupply,
+            "Zebec BSC Token",
+            "ZBC"
+        );
+
+        native = new HypNativeScaled(scale);
```
nambrot, line 48: This is assuming 18 decimals for the token, and i would write the test cases to be `amount`/user-focused

avious00:
@yorhodes bump to merge this

jmrossy (commented):
Only skimmed the ts changes but lgtm

Issue #2054, opened 2023-04-01
More granular ownership of Mailbox


nambrot:
Right now pausing, default ism setting and proxy admin owning are the same. The former two are the same storage slot. We should allow more granular setting of these permisssioned roles, both on chain and in tooling.

nambrot:
Can I look at the PR for this one?

PR #2675, opened 2023-08-18
Add pausable ism

yorhodes:
### Description

- Add Pausable ISM
- Add tests for pausable ism

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2054

### Backward compatibility

Yes

### Testing

Unit tests


solidity/contracts/isms/PausableIsm.sol
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * @inheritdoc IInterchainSecurityModule
+     * @dev Returns `false` when paused, `true` otherwise.
+     */
+    function verify(bytes calldata, bytes calldata)
+        external
+        view
+        returns (bool)
+    {
+        return !paused();
```
nambrot, line 23: Wdyt about a revert message in here to indicate to users that it's because this ISM has been triggered?

solidity/contracts/isms/PausableIsm.sol
```diff
@@ -0,0 +1,33 @@
...
+pragma solidity >=0.8.0;
+
+// ============ External Imports ============
+import {Pausable} from "@openzeppelin/contracts/security/Pausable.sol";
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+
+// ============ Internal Imports ============
+import {IInterchainSecurityModule} from "../interfaces/IInterchainSecurityModule.sol";
+
+contract PausableIsm is IInterchainSecurityModule, Ownable, Pausable {
```
nambrot, line 11: With v3, do you think PausableHook and PausableISM could be the same contract?

nambrot (approved):
I assume this will even more useful with V3, where folks would use one of the fallback ISMs to be still get the benefit of core dev pausability?

PR #2678, opened 2023-08-18
Auto update pending PRs

mattiecnvr:
### Description

I keep having my PRs that have auto-merge enabled get stuck because someone else merges to main first and then I have to manually click update and wait for it to run CI and then maybe auto-merge if another commit does not slip in first again.

This is tedious and a waste of engineering time. I said something to this effect in earshot of my wife and she mentioned that at google they have an Auto Sync and Auto Merge feature, and it made me realize there was probably an off-the shelf solution. There was. This is it.

This new workflow will watch for any changes to the main branch and then update all PRs that have auto-merge enabled. If there are merge conflicts, this workflow will still fail, but it will at least automate the fairly common, easy case. Also all tests will still be run so no bugs should be introduced by this that would not have been introduced by simply clicking update.

### Drive-by changes

Removed an unused workflow.

### Backward compatibility

Yes

### Testing

~~This PR is my test of it.~~
I actually can't test this until it is in the main branch.


mattiecnvr:
It was a nice thought, apparently this is not supported as well as I thought
image

PR #2685, opened 2023-08-24
Solana and Nautilus RC validators

daniel-savu:
### Description

Adds addresses of the rc validator keys for solana, nautilus and their testnet counterparts


tkporter (approved):
thanks 🙏 

I guess we should encourage this being done when a new chain is added to the `hyperlane` context? To avoid needing to do this as a followup when it's needed later

daniel-savu:
> thanks 🙏
> 
> I guess we should encourage this being done when a new chain is added to the `hyperlane` context? To avoid needing to do this as a followup when it's needed later

Makes sense, but would there be a way to enforce this other than putting together some docs for adding a new chain?

Issue #2690, opened 2023-08-28
Warp Routes should allow LPs to front the liquidity


nambrot:
Warp Routes are secured by ISMs, i.e collateral that is stored on let's say `HypERC20Collateral` is only redeemable if the configured ISM attests to the corresponding instruction from the permissioned contract on the remote chain. When using a relatively slow ISM (like an Optimistic ISM or an underlying canonical optimistic rollup bridge), the latency in transferring via a warp route could be not ideal. This bounty proposes a proof of concept that allow users to specify an optional fee to have LPs provide the underlying token before the ISM "finalizes". To implement this, you would do the following:

- Create variants of the existing Warp route contracts
- On the remote chain, allow warp route transfers to embed "fast metadata" to indicate a user's willingness to pay an LP. For simplicity, start with a single parameter `fastFee` that can be called via a `transferRemote` variant
- On the local chain, allow anybody (LP) to call `fillFastTransfer(recipient, amount, fastFee)` which takes the tokens (amount - fastFee) from `msg.sender` and gives them to the `recipient`. Store evidence of the action in storage.
- Upon `handle` of the message,  check that `fastFee` was set in the metadata, check if somebody has indeed filled the transfer previously, and if so, actually transfer the tokens to the LP. Otherwise, completely the transfer as normally.

To close out this ticket, create unit tests with the mocks that illustrate this behavior

RohanShrothrium:
@nambrot I will work on this!

PR #2714, opened 2023-09-07
rs-2690: Fast warp routes

RohanShrothrium:
### Description

This PR introduces a feature to warp routes such that LPs can fill transactions on local chains and get a fee for doing this.

### Drive-by changes

No.

### Related issues

#2690 

### Backward compatibility

No

### Testing

Added Unit tests.


typescript/token/contracts/FastHypERC20.sol
```diff
@@ -0,0 +1,127 @@
...
+import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+
+/**
+ * @title Hyperlane ERC20 Token Router that extends ERC20 with remote transfer functionality.
+ * @author Abacus Works
+ * @dev Supply on each chain is not constant but the aggregate supply across all chains is.
+ */
+contract FastHypERC20 is ERC20Upgradeable, TokenRouter {
+    uint8 private immutable _decimals;
+    uint256 public fastTransferId;
```
nambrot, line 16: This should work both ways

typescript/token/contracts/FastHypERC20.sol
```diff
@@ -0,0 +1,127 @@
...
+import {Message} from "./libs//Message.sol";
+
+import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+
+/**
+ * @title Hyperlane ERC20 Token Router that extends ERC20 with remote transfer functionality.
+ * @author Abacus Works
+ * @dev Supply on each chain is not constant but the aggregate supply across all chains is.
+ */
+contract FastHypERC20 is ERC20Upgradeable, TokenRouter {
```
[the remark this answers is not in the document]
RohanShrothrium: Right makes sense!

typescript/token/contracts/FastHypERC20.sol
RohanShrothrium: Ohh so on the remote chain someone can provide the HypERC20 as liquidity?

nambrot: Yes

RohanShrothrium:
@nambrot  I have updated the contracts implementing your suggestions.

The tests are not as comprehensive as I would like it to be.
Will make the tests more comprehensive once I have a green flag in regard to the contract implementation.

typescript/token/contracts/FastHypERC20.sol
```diff
@@ -0,0 +1,161 @@
...
+        uint256 _amount,
+        bytes calldata _metadata
+    ) internal returns (bool) {
+        (uint256 _fastFee, uint256 _fastTransferId) = abi.decode(
+            _metadata,
+            (uint256, uint256)
+        );
+
+        FastTranferMetadata memory m_filledMetadata = fastTransfers[
+            _fastTransferId
```
nambrot, line 67: Actually, a warp route can have multiple remotes, so you probably need to key this by the remote domain/router as well

typescript/token/contracts/libs/FastTransfer.sol
```diff
@@ -0,0 +1,175 @@
...
+            ];
+
+            m_filledMetadata.fastFee <= _fastFee &&
+                _recipient == m_filledMetadata.recipient &&
+                _amount == m_filledMetadata.amount
+                ? _fastTransferTo(m_filledMetadata.filler, _amount)
+                : _fastTransferTo(_recipient, _amount);
+        } else {
+            _fastTransferTo(_recipient, _amount);
+        }
```
yorhodes, line 73: ditto for this
would prefer removing the metadata branching
also replace the ternary with a proper if/else

yorhodes (commented):
looking good!

yorhodes (commented):
looking good!

typescript/token/contracts/libs/FastTransfer.sol
```diff
@@ -0,0 +1,175 @@
...
+        internal
+        virtual;
+
+    /**
+     * @dev Should collect `amount` of tokens from `_sender`.
+     * @dev The implementation is delegated.
+     */
+    function _fastRecieveFrom(address _sender, uint256 _amount)
+        internal
+        virtual;
```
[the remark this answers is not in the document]
RohanShrothrium: @yorhodes We are using the same internal functions but outsourcing the transfer logic to the end contracts.
That is, `_transferTo` handles the logic, `_fastTransferTo` and `_fastRecieveFrom` are functions that can be overridden as `_mint`/`erc20.safTransfer` and `_burn`/`erc20.safeTransferFrom` in the Collateral and Synthetic contract implementations.

typescript/token/contracts/libs/FastTransfer.sol
RohanShrothrium: The issue is there is no clean way to handle abi decode errors and we want the fast warp route to not be breaking and still implement the `transferRemote` function that does not encode any metadata.
Without the if statement this is not possible.

RohanShrothrium: @yorhodes Although I tried breaking it down into two functions, let me know what you think!

yorhodes: > The issue is there is no clean way to handle abi decode errors and we want the fast warp route to not be breaking and still implement the `transferRemote` function that does not encode any metadata. Without the if statement this is not possible.

interesting, I see, that makes sense

typescript/token/contracts/libs/FastTokenRouter.sol
```diff
@@ -0,0 +1,195 @@
...
+ * @title Common FastTokenRouter functionality for ERC20 Tokens with remote transfer support.
+ * @author Abacus Works
+ */
+abstract contract FastTokenRouter is TokenRouter {
+    using TypeCasts for bytes32;
+    using Message for bytes;
+    /**
+     * @notice `FastTransferMetadata` is the LP data stored against `fastTransferId`.
+     */
+    struct FastTranferMetadata {
```
yorhodes, line 19: nit: spelling
```suggestion
    struct FastTransferMetadata {
```

typescript/token/contracts/libs/FastTokenRouter.sol
```diff
@@ -0,0 +1,195 @@
...
+        address filler;
+        address recipient;
+        uint256 amount;
+        uint256 fastFee;
+    }
+
+    uint256 public fastTransferId;
+    // maps `fastTransferId` to `FastTranferMetadata`.
+    mapping(bytes32 => FastTranferMetadata) filledFastTransfers;
```
yorhodes, line 28: instead of storing the whole struct can you store a hash and only use a single slot?

typescript/token/contracts/libs/FastTokenRouter.sol
```diff
@@ -0,0 +1,195 @@
...
+     * @return messageId The identifier of the dispatched message.
+     */
+    function fastTransferRemote(
+        uint32 _destination,
+        bytes32 _recipient,
+        uint256 _amountOrId,
+        uint256 _fastFee
+    ) public payable virtual returns (bytes32 messageId) {
+        uint256 _fastTransferId = fastTransferId;
+        fastTransferId = _fastTransferId + 1;
```
yorhodes, line 115: I meant if you have the transfer ID be a commitment to the origin and transfer metadata
```suggestion
        bytes32 fastTransferId = keccak256(fastTransferId, fastTransferMetadata);
```

typescript/token/contracts/libs/FastTokenRouter.sol
```diff
@@ -0,0 +1,195 @@
...
+            keccak256(abi.encodePacked(_origin, _fastTransferId))
+        ];
+
+        if (
+            m_filledMetadata.fastFee <= _fastFee &&
+            _recipient == m_filledMetadata.recipient &&
+            _amount == m_filledMetadata.amount
+        ) {
+            return m_filledMetadata.filler;
+        }
```
yorhodes, line 175: and then something like this that just checks if the commitment to the transfer metadata was filled already
```suggestion
        if (filledFastTransfers[fastTransferId] == fastTransferId) {
            return filler;
        } else {
            return recipient;
        }
```

typescript/token/contracts/libs/FastTokenRouter.sol
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * @dev Transfers `_amount` of token to `_recipient`/`fastFiller` who provided LP.
+     * @dev Called by `handle` after message decoding.
+     */
+    function _transferTo(
+        address _recipient,
+        uint256 _amount,
+        uint32 _origin,
+        bytes calldata _metadata
+    ) internal virtual {
```
yorhodes, line 55: I think its confusing to have the same name but not override the base class

typescript/token/contracts/libs/FastTokenRouter.sol
RohanShrothrium: Because the handle does not know the filler address, that can't be hashed.
But right we can probably hash the rest of the data and store that as a key and ensure it uses a single slot for address.

typescript/token/contracts/libs/FastTokenRouter.sol
RohanShrothrium: But this is a common design choice used right? Even the `GasRouter` implements this type of design right?
A bit reluctant to change the name because `_handle` uses `_transferTo` for the transfer logic and I wouldn't want to change that.
What do you think? @yorhodes

typescript/token/contracts/libs/FastTokenRouter.sol
yorhodes: hmm good point, I like your change

yorhodes (approved):

Issue #2723, opened 2023-09-12
Relayer panic: 'can not convert float seconds to Duration: value is either too big or NaN' in SyncerEtaCalculator


tkporter:
This is a new one :(

On hyperlane context of omniscient relayer on testnet3.

Image: `gcr.io/abacus-labs-dev/hyperlane-agent:892cc5d-20230908-162614`

Logs: https://cloudlogging.app.goo.gl/uR8m3H8dLtC77J7P6

Pasted:
```
{"timestamp":"2023-09-12T13:54:05.436645Z","level":"INFO","fields":{"return":"58"},"target":"hyperlane_sealevel::mailbox","span":{"_maybe_lag":"None","name":"count"},"spans":[{"domain":"solanadevnet","label":"dispatched_messages","name":"ContractSync"},{"name":"sequence_and_tip"},{"_maybe_lag":"None","name":"count"}]}
{"timestamp":"2023-09-12T13:54:08.450546Z","level":"INFO","fields":{"message":"Found log(s) in index range","range":"16414563..=16414648","num_logs":0,"estimated_time_to_sync":"synced"},"target":"hyperlane_base::contract_sync","span":{"domain":"proteustestnet","label":"gas_payments","name":"ContractSync"},"spans":[{"domain":"proteustestnet","label":"gas_payments","name":"ContractSync"}]}
{"timestamp":"2023-09-12T13:54:08.983441Z","level":"INFO","fields":{"message":"Found log(s) in index range","range":"9682143..=9682143","num_logs":0,"estimated_time_to_sync":"synced"},"target":"hyperlane_base::contract_sync","span":{"domain":"goerli","label":"gas_payments","name":"ContractSync"},"spans":[{"domain":"goerli","label":"gas_payments","name":"ContractSync"}]}
thread 'main' panicked at 'can not convert float seconds to Duration: value is either too big or NaN', /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/time.rs:741:23
stack backtrace:
   0:     0x5624797204e1 - std::backtrace_rs::backtrace::libunwind::trace::h782cc21a5acaf6cb
                               at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x5624797204e1 - std::backtrace_rs::backtrace::trace_unsynchronized::hc579eb24ab204515
[...]

PR #2724, opened 2023-09-12
chore(sealevel igp indexing): reduce log verbosity

daniel-savu:
Removes some unnecessary logs from the sealevel igp indexing fn. Also instruments the eta calculator in response to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2723

### DIFF
--- a/rust/hyperlane-base/src/contract_sync/eta_calculator.rs
+++ b/rust/hyperlane-base/src/contract_sync/eta_calculator.rs
@@ -1,6 +1,7 @@
 use std::time::{Duration, Instant};
 
 use derive_new::new;
+use tracing::trace;
 
 /// Calculates the expected time to catch up to the tip of the blockchain.
 #[derive(new)]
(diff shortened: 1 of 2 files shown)


rust/hyperlane-base/src/contract_sync/eta_calculator.rs
```diff
@@ -24,6 +24,7 @@ pub(crate) struct SyncerEtaCalculator {
 
 impl SyncerEtaCalculator {
     /// Calculate the expected time to catch up to the tip of the blockchain.
+    #[instrument(err)]
```
tkporter, line 27: Will this do anything? I thought this just does something when the function returns a Result, but this just returns a Duration

Also because the panic occurs on line 62 or 64 we wouldn't get a log

How about we change this to return a Result and move `Duration::from_secs_f64` to `Duration::try_from_secs_f64` https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.try_from_secs_f64 ?

tkporter (approved):

tkporter (approved):

PR #2725, opened 2023-09-12
Try new foundry toolchain gh action

yorhodes:


yorhodes:
apparently forge team is aware of these performance regressions and working on a fix

PR #2727, opened 2023-09-13
Set igp on sealevel

daniel-savu:
### Description
Getter and setter for the warp route igp

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/571

### Testing

Manual


rust/sealevel/client/src/main.rs
```diff
@@ -345,6 +345,39 @@ struct TransferOwnership {
...
+    igp_account: Pubkey,
+}
+
+#[derive(Args)]
+struct GetIgpArgs {
+    token_type: TokenType,
+}
+
+#[derive(ValueEnum, Clone)]
+enum IgpType {
```
tkporter, line 376: I take it there's no way to use the existing `InterchainGasPaymasterType` enum w/ clap?

rust/sealevel/client/src/main.rs
```diff
@@ -345,6 +345,39 @@ struct TransferOwnership {
...
+#[derive(Args)]
+struct Igp {
+    #[arg(long, short, default_value_t = HYPERLANE_TOKEN_PROG_ID)]
+    program_id: Pubkey,
+    #[command(subcommand)]
+    cmd: GetSetCmd<GetIgpArgs, SetIgpArgs>,
+}
+
+#[derive(Subcommand)]
+enum GetSetCmd<G: Args, S: Args> {
```
tkporter, line 357: nice

rust/sealevel/client/src/main.rs
```diff
@@ -1108,6 +1113,39 @@ fn process_token_cmd(ctx: Context, cmd: TokenCmd) {
...
+                };
+                let instruction = hyperlane_sealevel_token_lib::instruction::set_igp_instruction(
+                    args.program_id,
+                    ctx.payer_pubkey,
+                    set_args.igp_program,
+                    igp_type,
+                )
+                .unwrap();
+
+                ctx.new_txn().add(instruction).send_with_payer();
```
tkporter, line 1132: Could you use `add_with_description`? I recently added this so that multisig txs will have some context with them and was hoping for us to slowly move over to using it

rust/sealevel/libraries/hyperlane-sealevel-token/src/instruction.rs
```diff
@@ -200,3 +200,33 @@ pub fn transfer_ownership_instruction(
 
     Ok(instruction)
 }
+
+/// Transfers ownership.
+pub fn set_igp_instruction(
+    program_id: Pubkey,
+    owner_payer: Pubkey,
+    igp_program: Pubkey,
```
tkporter, line 208: Maybe worth this taking in `Option<(Pubkey, InterchainGasPaymasterType)>` so that we can also use this to unset an IGP if we ever want to in the future? But in the sealevel client, we can always set it to `Some` - just is nice to have maximal flexibility here

tkporter (approved):
lgtm! some small suggestions

rust/sealevel/client/src/main.rs
daniel-savu: None that I could find. It's probably possible to auto-generate this type with a macro but we can't just drop the inner enum types of `InterchainGasPaymasterType` for clap

PR #2729, opened 2023-09-13
Running forge with latest solc

aroralanuk:
### Description

- Removed contracts-bedrock contracts dependency from eth-optimism (which was forcing us to use 0.8.15)
- Added local minimal interfaces for the CrossDomainMessenger contract

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

Unit tests


solidity/foundry.toml
```diff
@@ -3,9 +3,8 @@ src = 'contracts'
 out = 'out'
 libs = ['node_modules', 'lib']
 test = 'test'
-cache_path  = 'forge-cache'
+cache_path = 'forge-cache'
 allow_paths = ["../node_modules"]
-solc = '0.8.15'
```
yorhodes, line 8: we should specify still no?

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -154,17 +149,15 @@ contract OPStackIsmTest is Test {
             (messageId)
         );
 
-        uint40 nonce = ICanonicalTransactionChain(L1_CANNONICAL_CHAIN)
```
yorhodes, line 157: aren't we fork testing?

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -232,15 +225,12 @@ contract OPStackIsmTest is Test {
             (messageId)
         );
 
-        (uint240 nonce, uint16 verison) = Encoding.decodeVersionedNonce(
+        (uint240 nonce, uint16 verison) = decodeVersionedNonce(
```
yorhodes, line 228: nit: spelling here and everywhere

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -232,15 +225,12 @@ contract OPStackIsmTest is Test {
...
             l2Messenger.messageNonce()
         );
-        uint256 versionedNonce = Encoding.encodeVersionedNonce(
-            nonce + 1,
-            verison
-        );
+        uint256 versionedNonce = encodeVersionedNonce(nonce + 1, verison);
 
-        bytes32 versionedHash = Hashing.hashCrossDomainMessageV1(
+        bytes32 versionedHash = hashCrossDomainMessageV1(
```
yorhodes, line 233: imo this is a nice opportunity to do the prank and remove the forking but not gonna force it

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -465,4 +443,102 @@ contract OPStackIsmTest is Test {
                 testMessage
             );
     }
+
+    /// @dev from eth-optimism/contracts-bedrock/contracts/libraries/Hashing.sol
+    /// @notice Hashes a cross domain message based on the V1 (current) encoding.
```
yorhodes, line 448: would be nice not to vendor this ;)

yorhodes (approved):

solidity/foundry.toml
aroralanuk: No, we should let it infer it

solidity/test/isms/OPStackIsm.t.sol
aroralanuk: we're not testing the nonce in the event

PR #2743, opened 2023-09-19
add globalHookMetadata

aroralanuk:
### Description

- Consolidating IGPMetadata and OPStackMetadata to versioned GlobalHookMetadata 
- Add `supportMetadata` to `IPostDispatchHook`

 - Metadata Schema
 [0:1] version
 [1:33] msg.value
 [34:65] Gas limit for message (IGP)
 [66:85] Refund address for message (IGP)
 [86:] Custom metadata

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/610

### Backward compatibility

No

### Testing

Unit tests


solidity/contracts/GasRouter.sol
```diff
@@ -50,9 +50,12 @@ abstract contract GasRouter is Router {
                 _destinationDomain,
                 _mustHaveRemoteRouter(_destinationDomain),
                 "",
-                IGPMetadata.formatMetadata(
+                GlobalHookMetadata.formatMetadata(
+                    1,
```
yorhodes, line 54: I think the idea was that the version is defined by the implementation but the user shouldn't have to know about versions
so `formatMetadata` would prepend whatever version

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+ * [34:65] Gas limit for message (IGP)
+ * [66:85] Refund address for message (IGP)
+ * [86:] Custom metadata
+ */
+library GlobalHookMetadata {
+    uint8 private constant VARIANT_OFFSET = 0;
+    uint8 private constant MSG_VALUE_OFFSET = 1;
+    uint8 private constant GAS_LIMIT_OFFSET = 33;
+    uint8 private constant REFUND_ADDRESS_OFFSET = 65;
+    uint256 private constant MIN_METADATA_LENGTH = 85;
```
yorhodes, line 30: I personally would only advocate for msg value to be global out of these 3

yorhodes: cc @nambrot

solidity/contracts/Router.sol
```diff
@@ -204,9 +204,12 @@ abstract contract Router is HyperlaneConnectionClient, IMessageRecipient {
     ) internal returns (bytes32 _messageId) {
         // Ensure that destination chain has an enrolled router.
         bytes32 _router = _mustHaveRemoteRouter(_destinationDomain);
-        bytes memory metadata = IGPMetadata.formatMetadata(
+        bytes memory metadata = GlobalHookMetadata.formatMetadata(
+            1,
```
yorhodes, line 208: ditto

solidity/contracts/hooks/AbstractMessageIdAuthHook.sol
```diff
@@ -31,10 +31,13 @@ abstract contract AbstractMessageIdAuthHook is
     IPostDispatchHook,
     MailboxClient
 {
+    using GlobalHookMetadata for bytes;
     using Message for bytes;
 
     // ============ Constants ============
 
+    // The variant of the metadata used in the hook
+    uint8 public constant METADATA_VARIANT = 1;
```
yorhodes, line 40: maybe have a constant in the metadata library and make this = library.version

solidity/contracts/hooks/OPStackHook.sol
```diff
@@ -76,14 +76,11 @@ contract OPStackHook is AbstractMessageIdAuthHook {
         internal
         override
     {
+        uint256 _msgValue = metadata.msgValue(0);
```
yorhodes, line 79: what is the 0 here?

solidity/contracts/hooks/OPStackHook.sol
```diff
@@ -76,14 +76,11 @@ contract OPStackHook is AbstractMessageIdAuthHook {
...
-            metadata.msgValue() < 2**255,
+            _msgValue < 2**255,
             "OPStackHook: msgValue must less than 2 ** 255"
         );
-        l1Messenger.sendMessage{value: metadata.msgValue()}(
-            ism,
-            payload,
-            GAS_LIMIT
-        );
+        l1Messenger.sendMessage{value: _msgValue}(ism, payload, GAS_LIMIT);
```
yorhodes, line 84: I think actually keeping it as calldata slicing instead of loading it into memory should be slightly cheaper

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @notice Returns the variant of the metadata.
+     * @param _metadata ABI encoded global hook metadata.
+     * @return variant of the metadata as uint8.
+     */
+    function variant(bytes calldata _metadata) internal pure returns (uint8) {
+        if (_metadata.length < VARIANT_OFFSET + 1) return 0;
```
yorhodes, line 38: this "null value" is a good reason for it to not be 0 indexed

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @notice Returns the specified refund address for the message.
+     * @param _metadata ABI encoded global hook metadata.
+     * @return Refund address for the message as address.
+     */
+    function getCustomMetadata(bytes calldata _metadata)
+        internal
+        pure
+        returns (bytes memory)
```
yorhodes, line 99:
```suggestion
        returns (bytes calldata)
```

yorhodes: what is this used for?

yorhodes: I dont understand

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -169,7 +171,7 @@ contract OPStackIsmTest is Test {
         vm.selectFork(mainnetFork);
 
         bytes memory message = MessageUtils.formatMessage(
-            VERSION,
+            OPTIMISM_VERSION,
```
yorhodes, line 174: this is a different version

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -191,8 +193,12 @@ contract OPStackIsmTest is Test {
         vm.selectFork(mainnetFork);
 
         vm.deal(address(this), uint256(2**255 + 1));
-        bytes memory excessValueMetadata = abi.encodePacked(
-            uint256(2**255 + 1)
+        bytes memory excessValueMetadata = GlobalHookMetadata.formatMetadata(
+            HYPERLANE_VERSION,
```
yorhodes, line 197: ditto, library should include these by itself

solidity/contracts/hooks/ConfigFallbackDomainRoutingHook.sol
```diff
@@ -30,22 +40,44 @@ contract ConfigFallbackDomainRoutingHook is IPostDispatchHook {
         mailbox = IMailbox(_mailbox);
     }
 
+    // ============ External Functions ============
+
+    // @inheritdoc IPostDispatchHook
+    function supportsMetadata(bytes calldata metadata)
```
yorhodes, line 46: can we do something so this isn't duplicated 8 times 🙏

yorhodes (changes requested):

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
nambrot: I would make all three global as well just because these are the 3 parameters we do expect people to use (as well as custom metadata to be appended after).

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+ * Format of metadata:
+ *
+ * [0:1] variant
+ * [1:33] msg.value
+ * [34:65] Gas limit for message (IGP)
+ * [66:85] Refund address for message (IGP)
+ * [86:] Custom metadata
+ */
+library GlobalHookMetadata {
+    uint8 private constant VARIANT_OFFSET = 0;
```
nambrot, line 26: IMO, I would 1 or 2 more bits to allow for more prefixes

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @notice Formats the specified gas limit and refund address into global hook metadata.
+     * @param _variant Variant of the metadata.
+     * @param _msgValue msg.value for the message.
+     * @param _gasLimit Gas limit for the message.
+     * @param _refundAddress Refund address for the message.
+     * @param _customMetadata Additional metadata to include in the global hook metadata.
+     * @return ABI encoded global hook metadata.
+     */
+    function formatMetadata(
```
nambrot, line 114: Does the caller need to specify the variant? I.e. the variant/prefix is tight to the schema no? Also, IMO we should have short-hand functions if you want to just provide msgValue or gasLimit

solidity/contracts/hooks/StaticProtocolFee.sol
```diff
@@ -109,6 +84,30 @@ contract StaticProtocolFee is IPostDispatchHook, Ownable {
...
+        internal
+        override
+    {
+        require(
+            msg.value >= protocolFee,
+            "StaticProtocolFee: insufficient protocol fee"
+        );
+
+        uint256 refund = msg.value - protocolFee;
+        if (refund > 0) payable(message.senderAddress()).sendValue(refund);
```
yorhodes, line 98: should we use metadata refund address here and fallback to message.senderAddress?

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
yorhodes: 2^8 bits isn't enough?

solidity/contracts/libs/hooks/GlobalHookMetadata.sol
```diff
@@ -0,0 +1,130 @@
...
+@@@@@@@@@       @@@@@@@@*/
+
+/**
+ * Format of metadata:
+ *
+ * [0:2] variant
+ * [3:34] msg.value
+ * [35:66] Gas limit for message (IGP)
+ * [67:86] Refund address for message (IGP)
+ * [87:] Custom metadata
```
yorhodes, line 23: these numbers seem off?

yorhodes (approved):

Issue #2502, opened 2023-07-10
Sealevel: create helloworld program and deploy tooling for use by Kathy


tkporter:
* Create a helloworld program (maybe we can choose to piggyback off `test-send-receiver`? though probably makes sense to have it be a distinct program)
* Write deploy tooling for the helloworld program. I think it probably makes sense to piggyback off the existing router logic that the warp route deployment logic uses -- it's not anything generic yet, but we can make it generic

avious00:
PR Review by Rossy and Yorke here; https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2744

PR #2744, opened 2023-09-20
Helloworld program and tooling

tkporter:
### Description

* HelloWorld program added to `rust/sealevel/programs/helloworld`
* Some changes to the Sealevel tooling
  * Refactored the old warp-route-specific deploy management into a more generic framework for idempotently deploying and managing router apps in Sealevel
    * Added helloworld deploy tooling
    * Added ISM & owner checking to this router deployment tooling
  * Added `--require-tx-approval` to prevent txs from being called without first prompting
  * Added a bunch of new commands for creating txs with certain instructions that were needed along the way -- e.g. setting an ISM, deploying a specific multisig ISM (previously only one would be deployed as a part of `core deploy`
  * Added a command for configuring multisig ISMs, `multisig-ism-message-id configure`, that takes in a JSON file of multisig ISM configs and applies them onchain
  * A bit of cleanup / refactor - e.g. removed some old commands like `mailbox receive`
* Added foreignDeployments into `RouterApp`
  * Because RouterApp takes in `contractsMap: HyperlaneContractsMap<Factories>,`, which require attached contracts, a new `readonly foreignDeployments: ChainMap<Address> = {},` is added to the constructor
  * These foreignDeployments are considered in the return value of `remoteChains(chainName: string)`, but not in `chains()` -- this means that `chains()` now concretely means "chains that can be deployed to / interacted with and that there is an entry in `contractsMap` for, and `remoteChains(chainName: string)` returns any and all remote chains, regardless of whether they can be deployed to / interacted with
* Added complete ISM support to the HyperlaneRouterChecker
  * when checking the ISM, if there's not a match and the ISM is a config, then the ISM will be deployed
  * Also added RouterViolation, before it'd just throw if there was a violation
* Updated the Helloworld, IGP, and core tooling to work when AEE deployments are also configured
* Moved to Routing ISM -> Aggregation ISM -> Merkle / Message ID multisig setup

Some things to note:
* atm there are a few places that have a TODO saying to remove something after some multisig txs are executed, I plan to revisit these after we get some sigs
* I've deployed the mainnet sealevel version of helloworld, but haven't been able to enroll it in the EVM chains yet. Waiting for some multisig activity here

### Related issues

#2502

### Backward compatibility

I believe it should all be backward compatible

### Testing

Deployed, ran checkers, etc


typescript/infra/config/environments/mainnet2/chains.ts
```diff
@@ -40,12 +40,26 @@ export const mainnetConfigs: ChainMap<ChainMetadata> = {
   gnosis: chainMetadata.gnosis,
 };
 
+// Blessed non-Ethereum chains.
+export const nonEthereumMainnetConfigs: ChainMap<ChainMetadata> = {
```
jmrossy, line 44: Should this be `sealevelMainnetConfigs` or could other non-evm chains possibly go here? Regardless, I guess a lot of this is temporary anyway

typescript/infra/config/environments/testnet3/gas-oracle.ts
```diff
@@ -55,6 +61,7 @@ const chainTokenRarity: ChainMap<Rarity> = {
   moonbasealpha: Rarity.Common,
   optimismgoerli: Rarity.Mythic,
   arbitrumgoerli: Rarity.Mythic,
+  solanadevnet: Rarity.Common,
```
jmrossy, line 64: Haha I haven't seen this file before.

typescript/infra/config/routingIsm.ts
```diff
@@ -34,81 +39,102 @@ const owners = {
...
+  };
+};
+
+export const multisigIsms = (
+  env: DeployEnvironment,
+  local: ChainName,
+  type: MultisigIsmConfig['type'],
+  context: Contexts,
+): ChainMap<MultisigIsmConfig> =>
+  objMap(
```
jmrossy, line 124: When I think about the code in infra that may have trouble finding a home in the CLI, all this utils-like stuff for creating/transforming configs comes to mind. 
Not a criticism of the code here, just thinking out loud

typescript/infra/scripts/module-can-verify.ts
```diff
@@ -0,0 +1,78 @@
...
+  moduleCanCertainlyVerify,
+} from '@hyperlane-xyz/sdk';
+import { Address, ProtocolType } from '@hyperlane-xyz/utils';
+
+import { deployEnvToSdkEnv } from '../src/config/environment';
+
+import { getArgs, getEnvironmentConfig } from './utils';
+
+const mainnetHyperlaneDefaultIsmCache: ChainMap<Address> = {
+  ethereum: '0x3Ef03aEf1392E5e0C16fd4D22C3c3b4f81C8AF0C',
```
jmrossy, line 13: Would prefer we not keep consts like this in a script file. Can we find it a home in `/config/` somewhere?

typescript/infra/scripts/module-can-verify.ts
```diff
@@ -0,0 +1,78 @@
...
+  const multiProvider = await config.getMultiProvider();
+
+  const core = HyperlaneCore.fromEnvironment(
+    deployEnvToSdkEnv[environment],
+    multiProvider,
+  );
+
+  for (const local of core.chains()) {
+    if (
+      multiProvider.getChainMetadata(local).protocol !== ProtocolType.Ethereum
```
jmrossy, line 43: I'm gonna add a `getProtocol` method to `ChainMetadataManager`. We'll probably make calls like this often

typescript/infra/scripts/print-chain-metadatas.ts
```diff
@@ -0,0 +1,17 @@
+import { getArgs, getEnvironmentConfig } from './utils';
+
+// This script exists to print the chain metadata configs for a given environment
+// so they can easily be copied into the Sealevel tooling. :'(
```
jmrossy, line 4: 😢 indeed, but I think these will fit nicely into the CLI. There's already commands to view chains and their addresses (essentially a view into the SDK consts). These can go there too.

typescript/infra/src/config/helloworld-getters.ts
```diff
@@ -0,0 +1,37 @@
...
+import { objMap } from '@hyperlane-xyz/utils';
+
+import { Contexts } from '../../config/contexts';
+import { routingIsm } from '../../config/routingIsm';
+import { getRouterConfig } from '../../scripts/utils';
+
+import { DeployEnvironment } from './environment';
+import { HelloWorldConfig } from './helloworld';
+
+export const helloWorldConfig = (
```
jmrossy, line 11: Do these belong with https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/testnet3/helloworld.ts ?

typescript/infra/src/govern/ProxiedRouterGovernor.ts
```diff
@@ -23,13 +27,20 @@ export class ProxiedRouterGovernor<
...
-        ConnectionClientViolationType.InterchainSecurityModule
-      ) {
-        this.handleIsmViolation(violation as ConnectionClientViolation);
-      } else {
-        throw new Error(`Unsupported violation type ${violation.type}`);
+      switch (violation.type) {
+        case ConnectionClientViolationType.InterchainSecurityModule:
+          // TODO remove upon transferring ownership of HelloWorld in multisig txs
+          console.log('Skipping ISM violation');
+          // this.handleIsmViolation(violation as ConnectionClientViolation);
```
jmrossy, line 34: Are we fixing this before or after merging?

typescript/infra/src/govern/multisend.ts
```diff
@@ -60,7 +60,7 @@ export class SafeMultiSend extends MultiSend {
     const safeTransactionData = calls.map((call) => {
       return { to: call.to, data: call.data.toString(), value: '0' };
     });
-    const nextNonce = await safeService.getNextNonce(this.safeAddress);
+    let nextNonce = await safeService.getNextNonce(this.safeAddress);
```
jmrossy, line 63: I'm guessing this slipped in unintentionally?

typescript/infra/src/utils/utils.ts
```diff
@@ -260,3 +265,11 @@ export function diagonalize<T>(array: Array<Array<T>>): Array<T> {
   }
   return diagonalized;
 }
+
+export function mustGetChainNativeTokenDecimals(chain: ChainName): number {
```
jmrossy, line 269: Will we need this often? We could add something to `ChainMetadataManager` for it if you like

typescript/sdk/src/consts/environments/mainnet.json
```diff
@@ -153,5 +153,22 @@
     "interchainAccountRouter": "0xe9E3444DDD80c50276c0Fcf316026f6d7fEc2c47",
     "merkleRootMultisigIsmFactory": "0x784b9D0f4eF9fb8444DfB5d24AB221C9D1A85395",
     "messageIdMultisigIsmFactory": "0xC4275763D7b621eb732847957012F1fb35C90BB8"
+  },
+  "solana": {
+    "storageGasOracle": "0x0000000000000000000000000000000000000000",
```
jmrossy, line 158: Would base58 based 0 addrs be better here?

typescript/sdk/src/consts/environments/mainnet.json
```diff
@@ -153,5 +153,22 @@
     "interchainAccountRouter": "0xe9E3444DDD80c50276c0Fcf316026f6d7fEc2c47",
     "merkleRootMultisigIsmFactory": "0x784b9D0f4eF9fb8444DfB5d24AB221C9D1A85395",
     "messageIdMultisigIsmFactory": "0xC4275763D7b621eb732847957012F1fb35C90BB8"
+  },
+  "solana": {
```
jmrossy, line 157: I'm surprised that adding this doesn't cause a bunch of tests to fail. There's are a few places where functions in the SDK enumerate over the chains + contracts and then try to instantiate factories for everything. None of the new multi-protocol stuff has that problem but some old classes do.

That's why I did this for Kathy: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/scripts/helloworld/utils.ts#L74

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -143,9 +187,31 @@ export function appFromAddressesMapHelper<F extends HyperlaneFactories>(
   contractsMap: HyperlaneContractsMap<F>;
   multiProvider: MultiProvider;
 } {
+  // Hack to accommodate non-Ethereum artifacts, while still retaining their
```
jmrossy, line 190: I guess this answers my question above

typescript/sdk/src/gas/HyperlaneIgpChecker.ts
```diff
@@ -141,13 +147,18 @@ export class HyperlaneIgpChecker extends HyperlaneAppChecker<
...
+      ...this.app.remoteChains(local),
+      ...Object.keys(this.configMap[local].gasOracleType),
+    ]);
     for (const remote of remotes) {
       const remoteId = this.multiProvider.getDomainId(remote);
       const actualGasOracle = await igp.gasOracles(remoteId);
       const expectedGasOracle = this.getGasOracleAddress(local, remote);
 
-      if (eqAddress(actualGasOracle, expectedGasOracle)) {
+      if (!eqAddress(actualGasOracle, expectedGasOracle)) {
```
jmrossy, line 161: Nice catch!

typescript/sdk/src/router/HyperlaneRouterChecker.ts
```diff
@@ -13,13 +18,24 @@ import {
...
 export class HyperlaneRouterChecker<
   Factories extends HyperlaneFactories,
   App extends RouterApp<Factories>,
   Config extends RouterConfig,
 > extends HyperlaneAppChecker<App, Config> {
+  constructor(
+    multiProvider: MultiProvider,
+    app: App,
+    configMap: ChainMap<Config>,
+    readonly ismFactory?: HyperlaneIsmFactory,
```
jmrossy, line 34: Why is only the `ismFactory`readonly?

typescript/utils/src/amount.ts
```diff
@@ -117,3 +118,28 @@ export function convertDecimals(
...
+
+/**
+ * Converts a value with `fromDecimals` decimals to a value with `toDecimals` decimals.
+ * Incurs a loss of precision when `fromDecimals` > `toDecimals`.
+ * @param fromDecimals The number of decimals `value` has.
+ * @param toDecimals The number of decimals to convert `value` to.
+ * @param value The value to convert.
+ * @returns `value` represented with `toDecimals` decimals.
+ */
+export function convertDecimalsEthersBigNumber(
```
jmrossy, line 130: Yeah, the use of two BigNumber libs is unfortunate but somewhat necessary because Ethers' BigNumber can't handle decimals, making it poorly suited to handling user input. I'm in favor of avoiding Ethers' BN entirely and normalizing to either BigNumber.js or browser native BigInt. The Adapter classes would be a good place to handle that normalization, eventually

jmrossy (commented):
Reviewed the changes excluded the `/rust` folder. Mostly looks good! Just some small issues/questions

typescript/helloworld/src/app/app.ts
```diff
@@ -21,8 +21,9 @@ export class HelloWorldApp extends RouterApp<HelloWorldFactories> {
     public readonly core: HyperlaneCore,
     contractsMap: HyperlaneContractsMap<HelloWorldFactories>,
     multiProvider: MultiProvider,
+    foreignDeployments: ChainMap<Address> = {},
   ) {
-    super(contractsMap, multiProvider);
+    super(contractsMap, multiProvider, undefined, foreignDeployments);
```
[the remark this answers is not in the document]
tkporter: I think you don't need any changes -- the reason we still need this HelloWorldApp is because it's used for the deployer / checker. The foreign deployment stuff is just so that we only try to interact with the non-foreign EVM deployments, and that the foreign deployments are still enrolled as remote routers on the EVM contracts

typescript/infra/scripts/module-can-verify.ts
tkporter: great idea

typescript/infra/scripts/print-chain-metadatas.ts
tkporter: Ah nice. Think it's worth me keeping this till the CLI gets in?

typescript/infra/config/routingIsm.ts
tkporter: Yeah this is a great callout imo

I think we'll probably want to maintain the ability to programmatically construct these things if possible. I have some ideas but we should chat

typescript/infra/src/config/helloworld-getters.ts
tkporter: I actually had meant to rm this function - the other one is useful for mainnet2 as well as testnet3. But i moved `helloWorldRouterConfig` into `helloworld.ts` instead of this file

typescript/infra/src/govern/ProxiedRouterGovernor.ts
tkporter: Yeah will fix this before merging, now that I have multisig txs signed I can go ahead w this

typescript/infra/src/utils/utils.ts
tkporter: It's only in one place so I'm ok keeping it out, if you want to bring it in though lmk

typescript/sdk/src/consts/environments/mainnet.json
tkporter: Funnily enough not having this caused more things to fail and adding it but accommodating its presence throughout our tooling was easier. Most of the usage of `filterAddressesExcludeProtocol` / `filterAddressesToProtocol` is to accommodate it.

The reason why adding it was necessary was that there are a number of places where we use the `multiProvider.intersect` to get rid of chains that don't have artifacts, but then it creates errors if we e.e.g want to do things like enroll a Solana foreign deployment remote router, which requires the ability to look up the domain ID from the chain name

typescript/sdk/src/contracts/contracts.ts
tkporter: ah yes

typescript/sdk/src/router/HyperlaneRouterChecker.ts
tkporter: mostly wanted the ability to do `this.ismFactory`, and I don't see it as ever needing to be reassigned. Would you prefer to change?

typescript/utils/src/amount.ts
tkporter: would you prefer me to move over to BigNumber.js or browser native BigInt now to avoid this fn or plan that for the future?

jmrossy: No I don't think that's something you should handle in this PR. 
Create an issue to track: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2751

daniel-savu (approved):
Submitting first round of comments as I need to refresh the page due to the new commit :/

rust/sealevel/client/src/router.rs
```diff
@@ -0,0 +1,591 @@
...
+        .iter()
+        .map(|(chain_name, app_config)| (chain_name, app_config.router_config()))
+        .filter(|(_, router_config)| router_config.foreign_deployment.is_some())
+        .map(|(chain_name, router_config)| {
+            let chain_config = chain_configs.get(chain_name).unwrap();
+            (
+                chain_config.domain_id(),
+                hex_or_base58_to_h256(router_config.foreign_deployment.as_ref().unwrap()).unwrap(),
+            )
+        })
```
daniel-savu, line 322: nit: can use `filter_map` here and do `router_config.foreign_deployment.and_then(** closure passed to second "map" ** )`

daniel-savu (approved):
LGTM, mostly reviewed the client changes and since it's just scripting I avoided being very nitpicky. I also didn't focus too much on the router / gas filtering logic, let me know if I should ✌️

jmrossy (approved):

rust/sealevel/client/src/router.rs
```diff
@@ -0,0 +1,591 @@
...
+    pub interchain_gas_paymaster: Option<Pubkey>,
+    #[serde(default)]
+    #[serde(with = "crate::serde::serde_option_pubkey")]
+    pub interchain_security_module: Option<Pubkey>,
+}
+
+impl OptionalConnectionClientConfig {
+    pub fn mailbox(&self, default: Pubkey) -> Pubkey {
+        self.mailbox.unwrap_or(default)
+    }
```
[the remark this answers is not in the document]
tkporter: ah yeah I had meant to make it private

PR #2742, opened 2023-09-19
V3 agent changes 

daniel-savu:
### Description

Closes https://github.com/hyperlane-xyz/issues/issues/561

Builds on top of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2740

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2681 for e2e testing

This PR includes:
- [x] Merkle tree hook indexer
- [x] Merkle tree builder task
- [x] Update submitter to trigger retries if no proof is available yet

Slightly more detailed overview of the work here: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2720#issuecomment-1724038643

Issue #2720, opened 2023-09-08
Agents are resilient to merkle tree double insertion


yorhodes:
Related to https://github.com/hyperlane-xyz/issues/issues/561

relayer merkle builder assumes message nonce implies tree index
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/16146f0f03d0b0be67583cf16b22a3c50abdd977/rust/agents/relayer/src/merkle_tree_builder.rs#L118-L119

validator checkpoint submitter assumes message nonce implies tree index
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/16146f0f03d0b0be67583cf16b22a3c50abdd977/rust/agents/validator/src/submit.rs#L85-L89
```[tasklist]
### Tasks
- [x] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2739
- [x] create a new agent indexer for this event - Kunal (pr incoming today 9/18)
- [x] modify agent merkle tree builder to be independent of message indexing - Daniel & Trevor
- [x] modify relayer to spawn new indexer and remove merkle tree logic from main processor/submitter loops  - Daniel & Trevor
```

daniel-savu:
Relatively quick braindump based on the whiteboarding session below.

**DTM indexing**
- completion criteria:
	- use merkle tree hook as both the required and custom hook (2 insertions)
	- do not use the merkle tree hook at all (0 insertions)
- spawn new merkle builder task that iterates hyperlane DB
	- can do forward cursor indexing for relayer
	- for validator, might need different logic depending on the tip at startup (forward - forward indexing?). Look into validator indexing logic
- remove merkle builder Arc from Processor task, pass it to the merkle tree builder task instead
- Move merkle tree-building logic from message processor task, to merkle tree builder task
- add retrying logic for fetching insertions
- return Ok(None) to trigger retries in the builder, if the merkle tree is not yet available
- mapping between message id and nonce; should drop all usage of message nonce and use the leaf count instead; could store in memory or in storage
	- replace `retrieve_message_by_nonce` with `retrieve_leaf_by_index`, where index a u32

daniel-savu:
finished in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2815

PR #2681, opened 2023-08-22
V3 core deployment tooling (with static hook config assumptions)

yorhodes:
### Description

Updates core deployer to deploy V3 contracts with required hook as merkle tree and default hook as test IGP.

### Related issues

Fixes https://github.com/hyperlane-xyz/issues/issues/559

### Backward compatibility

Yes

### Testing

Unit tests

Issue #2313, opened 2023-06-01
Agent Indexing takes time and can be costy


nambrot:
Problem:
- All agents need to index messages (relayer also indexes IGP payments)
- #2192 allowed relayer indexing to no longer message processing
- Indexing inherently costs RPC requests. Polling for messages costs RPC requests.

Possible solution:
- Modularize the fetching of logs so that other sources of logs can be used (in addition to the default RPC-based method). Note that the current indexing relies on a configurable max window which is relatively small (1999 by default). Other sources presumably can support much larger windows.
- Potential sources:
  - Block explorers like etherscan (can use the block explorer struct from #1867)
  - Scraper Postgres Database
- This could allow the more frequent polling of cheaper sources (while always being independent by still polling the RPC, maybe with less frequency)

nambrot:
@asaj @yorhodes As far as I know, log fetching can be untrusted since agents can build the merkle tree and compare the roots, before using the indexed data?

asaj:
Indeed, it's a liveness but not safety thing

asaj:
@nambrot can't we just upload a cache of the rocks DB?

tkporter:
Fetching the DB ^ is interesting

I'm skeptical of the etherscan / postgres DB indexing. Feels like a lot of effort for a questionable improvement in a post-DTM world

We've spent a lot of time working on the quirks of indexing and I'm wary that interacting with Etherscan APIs will increase the surface area we need to maintain & be more confusing to configure

& for postgres DB, now we would need to take our scraper / DB SLA much more seriously. Would we lose the flexibility to mess with the DB / scraper and just point the explorer to a new DB without affecting anyone else?

nambrot:
Another source that we imo have not considered enough is the graph protocol. It provides very solid indexed data with great performance and is made for querying from agents like ours. Default subgraphs can be deployed and ran by anybody, and then queried by anybody. In fact, that is what the Khalani folks did (@serejke or @samtvlabs if you want to chime in)

HariSeldon23:
We have a bunch of Graph Subgraphs running in our Kubernetes backend. It works fairly well, although some of the debugging tools are quite poor and fixing things has caused us a lot of issues. I believe we have a bunch of updated Helm charts as we needed to make quite a few changes for some of the chains we're live on. 

Happy to help if there's any Graph queries as we've been using them for over 2 years.

[...]

PR #2746, opened 2023-09-21
V3 agents rebase

daniel-savu:
### Description

It's your favourite PR coming right back... V3 agents!

Closes https://github.com/hyperlane-xyz/issues/issues/561

Builds on top of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2742

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2681 for e2e testing

This PR includes:
- [x] Merkle tree hook indexer
- [x] Merkle tree builder task
- [x] Update submitter to trigger retries if no proof is available yet

Slightly more detailed overview of the work here: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2720#issuecomment-1724038643


daniel-savu:
Doesn't compile yet though

typescript/sdk/src/core/HyperlaneCoreDeployer.ts
```diff
@@ -81,8 +81,9 @@ export class HyperlaneCoreDeployer extends HyperlaneDeployer<
       mailbox.initialize(
         owner,
         defaultIsm,
-        merkleTreeHook.address,
-        merkleTreeHook.address,
+        defaultHook,
+        defaultHook,
+        // merkleTreeHook.address,
```
daniel-savu, line 86: Not sure whether we want to commit this. Probably easiest to not change the mailbox setup but know that it was tested "manually" for both double insertion and zero insertion?

.gitignore
```diff
@@ -28,3 +28,4 @@ yarn-error.log
 **/*.ignore
 .vscode
 
+tsconfig.editor.json
```
yorhodes, line 31: whats this?

rust/agents/relayer/src/merkle_tree/builder.rs
```diff
@@ -50,12 +50,6 @@ pub enum MerkleTreeBuilderError {
         /// Root of the incremental merkle tree
         incremental_root: H256,
     },
-    /// Nonce was not found in DB, despite batch providing messages after
-    #[error("Nonce was not found {nonce:?}")]
-    UnavailableNonce {
-        /// Root of prover's local merkle tree
-        nonce: u32,
-    },
```
yorhodes, line 58: we dont need this anymore?

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -307,111 +301,27 @@ where
...
-                .await
-                .map(Into::into)
-                .map_err(ChainCommunicationError::from_other)?;
-        }
-
-        let count = self
+        let nonce = self
             .contract
-            .count()
+            .nonce()
```
yorhodes, line 315: fwiw this is off by one from what it used to be

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -307,111 +301,27 @@ where
...
-                .provider
-                .get_storage_at(
-                    self.contract.address(),
-                    location.into(),
-                    Some(fixed_block_number),
-                )
-                .await
-                .map(Into::into)
-                .map_err(ChainCommunicationError::from_other)?;
-        }
```
yorhodes, line 399: glad to remove this

rust/chains/hyperlane-ethereum/tests/signer_output.rs
```diff
@@ -138,7 +138,7 @@ pub fn output_signed_checkpoints() {
         for i in 1..=3 {
             let signed_checkpoint = signer
                 .sign(Checkpoint {
-                    mailbox_address: mailbox,
+                    merkle_tree_hook_address: mailbox,
```
yorhodes, line 141: nit: rename maibox here?

rust/chains/hyperlane-ethereum/tests/signer_output.rs
```diff
@@ -147,7 +147,7 @@ pub fn output_signed_checkpoints() {
                 .expect("!sign_with");
 
             test_cases.push(json!({
-                "mailbox": signed_checkpoint.value.mailbox_address,
+                "mailbox": signed_checkpoint.value.merkle_tree_hook_address,
```
yorhodes, line 150: ditto, rename here?

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+where
+    M: Middleware + 'static,
+{
+    async fn sequence_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
+        // The InterchainGasPaymasterIndexerBuilder must return a `SequenceIndexer` type.
+        // It's fine if only a blanket implementation is provided for EVM chains, since their
+        // indexing only uses the `Index` trait, which is a supertrait of `SequenceIndexer`.
+        // TODO: if `SequenceIndexer` turns out to not depend on `Indexer` at all, then the supertrait
+        // dependency could be removed, even if the builder would still need to return a type that is both
+        // ``SequenceIndexer` and `Indexer`.
```
yorhodes, line 139: should we be tracking this somewhere? I'm not sure I understand all the implications @tkporter

yorhodes (approved):

.gitignore
daniel-savu: Added by @mattiecnvr afaik

rust/agents/relayer/src/error.rs
```diff
@@ -0,0 +1,7 @@
+// use std::fmt::Display;
+
+#[derive(Debug, thiserror::Error)]
+pub enum RelayerError {
+    #[error("Failed to fetch metadata")]
+    MetadataFetchingError,
+}
```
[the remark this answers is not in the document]
daniel-savu: you're right, just noticed we have `MetadataBuilderError`. But I think it's better to keep this new enum and nest `MetadataBuilderError` in it

rust/agents/relayer/src/merkle_tree/builder.rs
daniel-savu: didn't find any uses of it

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -152,6 +154,22 @@ impl HyperlaneRocksDB {
...
     }
 
+    /// Store the merkle tree insertion event, and also store a mapping from message_id to leaf_index
+    pub fn process_tree_insertion(&self, insertion: &MerkleTreeInsertion) -> DbResult<bool> {
+        if let Ok(Some(_)) = self.retrieve_merkle_tree_insertion_by_leaf_index(&insertion.index()) {
+            debug!(insertion=?insertion, "Tree insertion already stored in db");
+            return Ok(false);
+        }
+        // even if double insertions are ok, store the leaf by `leaf_index` (guaranteed to be unique)
+        // rather than by `message_id` (not guaranteed to be unique), so that leaves can be retrieved
```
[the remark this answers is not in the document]
daniel-savu: ah yes, I'll reword 👍

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -76,6 +77,8 @@ pub struct CoreContractAddresses {
     pub interchain_gas_paymaster: H256,
     /// Address of the ValidatorAnnounce contract
     pub validator_announce: H256,
+    /// Address of the MerkleTreeHook contract
+    pub merkle_tree_hook: Option<H256>,
```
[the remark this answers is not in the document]
daniel-savu: It could be none if there's no deployment. The agents default to querying the mailbox for the merkle tree count in this case (e.g. in sealevel)

rust/hyperlane-base/src/settings/loader/case_adapter.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn collect(&self) -> Result<Map<String, Value>, ConfigError> {
+        self.inner.collect().map(|m| {
+            m.into_iter()
+                .map(|(k, v)| recase_pair(k, v, self.casing))
+                .collect()
+        })
+    }
+}
+
+fn recase_pair(key: String, mut val: Value, case: Case) -> (String, Value) {
```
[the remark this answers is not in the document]
daniel-savu: this is config stuff

rust/chains/hyperlane-ethereum/src/mailbox.rs
daniel-savu: hm, I think this could confuse people comparing sealevel and evm metrics, but otherwise shouldn't functionally change anything?

rust/agents/relayer/src/merkle_tree/builder.rs
```diff
@@ -81,54 +78,43 @@ impl MerkleTreeBuilder {
     #[instrument(err, skip(self), level="debug", fields(prover_latest_index=self.count()-1))]
     pub fn get_proof(
         &self,
-        leaf_index: u32,
+        message_nonce: u32,
         root_index: u32,
-    ) -> Result<Proof, MerkleTreeBuilderError> {
+    ) -> Result<Option<Proof>, MerkleTreeBuilderError> {
+        let Some(message_id) = self
```
tkporter, line 84: could probably do some maps instead of these but not a big deal

rust/agents/relayer/src/merkle_tree/builder.rs
```diff
@@ -81,54 +78,43 @@ impl MerkleTreeBuilder {
...
-            if prover_root != incremental_root {
-                return Err(MerkleTreeBuilderError::MismatchedRoots {
-                    prover_root,
-                    incremental_root,
-                });
-            }
+    pub async fn ingest_message_id(&mut self, message_id: H256) -> Result<()> {
+        const CTX: &str = "When ingesting message id";
+        debug!(?message_id, "Ingesting leaf");
+        self.prover.ingest(message_id).expect("!tree full");
```
tkporter, line 109: nit: would probably remove the `!`, if this panics then I'd read this as meaning "tree not full" when it is in fact full

rust/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use super::builder::MerkleTreeBuilder;
+
+/// Finds unprocessed merkle tree insertions and adds them to the prover sync
+#[derive(new)]
+pub struct MerkleTreeProcessor {
+    db: HyperlaneRocksDB,
+    metrics: MerkleTreeProcessorMetrics,
+    prover_sync: Arc<RwLock<MerkleTreeBuilder>>,
+    #[new(default)]
```
tkporter, line 26: neat

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
daniel-savu: if we do get around to revamping indexing, this will certainly be addressed one way or another. But I think it's low prio for now

rust/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -0,0 +1,99 @@
...
+            // Feed the message to the prover sync
+            self.prover_sync
+                .write()
+                .await
+                .ingest_message_id(insertion.message_id())
+                .await?;
+
+            // Increase the leaf index to move on to the next leaf
+            self.leaf_index += 1;
+            // No need to explicitly send the merkle tree to the submitter, since it's
```
tkporter, line 60: not sure i follow this

rust/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[derive(Debug)]
+pub struct MerkleTreeProcessorMetrics {
+    max_leaf_index_gauge: IntGauge,
+}
+
+impl MerkleTreeProcessorMetrics {
+    pub fn new() -> Self {
+        Self {
+            max_leaf_index_gauge: IntGauge::new("max_leaf_index_gauge", "help string").unwrap(),
```
tkporter, line 96: nit: help string

rust/agents/relayer/src/msg/gas_payment/mod.rs
```diff
@@ -195,7 +194,7 @@ mod test {
         test_utils::run_test_db(|db| async move {
             let hyperlane_db =
                 HyperlaneRocksDB::new(&HyperlaneDomain::new_test_domain("test_no_match"), db);
-            let matching_list = serde_json::from_str(r#"[{"originDomain": 234}]"#).unwrap();
+            let matching_list = serde_json::from_str(r#"[{"origindomain": 234}]"#).unwrap();
```
tkporter, line 197: is this necessary? assuming this is from Mattie's config change but curious if this implies that `originDomain` doesn't work anymore

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -115,13 +128,21 @@ impl BaseMetadataBuilder {
...
     pub async fn get_proof(&self, nonce: u32, checkpoint: Checkpoint) -> Result<Option<Proof>> {
         const CTX: &str = "When fetching message proof";
-        let proof = self
-            .origin_prover_sync
-            .read()
-            .await
-            .get_proof(nonce, checkpoint.index)
-            .context(CTX)?;
-
+        let proof = retry(constant_backoff(), || async {
```
tkporter, line 124: I don't think we should need to retry here because the `get_proof` that's called returns an Option that we can propagate just propagate up. Then we can rely on the retry logic we get from our message submitter when metadata building is None. This assumes the aggregation ISM logic is robust enough where if we have a 1/2 setup of the merkle proof & message ID ISMs, if merkle proof metadata building returns None but the message id based metadata building returns Some, then we're still able to process things correctly

I think this is actually a bit dangerous to have a retry here because it's possible for someone to send a message without inserting it into the merkle tree on the origin, but on the destination they can configure an ISM that requires a merkle proof. In that case then the relayer will never be able to construct a merkle proof, so it'd be retrying unnecessarily

(maybe i'm misreading this tho, by my read this is retrying getting the proof if `get_proof` returns none?)

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -136,8 +157,15 @@ impl BaseMetadataBuilder {
         }
     }
 
-    pub async fn highest_known_nonce(&self) -> u32 {
-        self.origin_prover_sync.read().await.count() - 1
+    pub async fn highest_known_nonce(&self) -> Option<u32> {
+        self.origin_prover_sync.read().await.count().checked_sub(1)
+    }
+
+    pub async fn get_merkle_leaf_by_message_id(&self, message_id: H256) -> Result<Option<u32>> {
```
tkporter, line 157: nit: imo we should clarify this is getting the index, not the leaf itself, ie rename to `get_merkle_leaf_index_by_message_id`

rust/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -12,24 +12,27 @@ use hyperlane_core::{Checkpoint, HyperlaneMessage, SignatureWithSigner, H256};
...
 
+use crate::error::RelayerError;
 use crate::msg::metadata::BaseMetadataBuilder;
 use crate::msg::metadata::MetadataBuilder;
 
 #[derive(new)]
 pub struct MultisigMetadata {
     checkpoint: Checkpoint,
     signatures: Vec<SignatureWithSigner>,
+    merkle_leaf_id: Option<u32>,
```
tkporter, line 23: nit: similarly this should clarify this is the index, e.g. `merkle_leaf_index` or `message_leaf_index` or something

rust/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -52,40 +55,53 @@ pub trait MultisigIsmMetadataBuilder: AsRef<BaseMetadataBuilder> + Send + Sync {
...
-                ethers::abi::encode(&[Token::FixedArray(validator_tokens)])
-            }
-            MetadataToken::Signatures => {
-                let ordered_signatures = order_signatures(validators, &metadata.signatures);
-                let threshold_signatures = &ordered_signatures[..threshold as usize];
-                threshold_signatures.concat()
+    ) -> Result<Vec<u8>> {
+        let build_token = |token: &MetadataToken| -> Result<Vec<u8>> {
+            match token {
+                MetadataToken::MerkleRoot => Ok(metadata.checkpoint.root.to_fixed_bytes().into()),
```
tkporter, line 61: are these named for consistency with the solidity bit? tbh find these confusing, i'd probably rename:

MerkleRoot => CheckpointMerkleRoot
MerkleIndex => MessageMerkleLeafIndex
CheckpointMerkleTree => CheckpointMerkleTreeHook

rust/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
```diff
@@ -51,9 +55,15 @@ impl MultisigIsmMetadataBuilder for MerkleRootMultisigMetadataBuilder {
             return Ok(None);
         };
 
+        let merkle_leaf_id = self
```
tkporter, line 58: to be safe I think we should do the None check here and return early if this is None

I think it's unlikely we'd get None here if the above `get_proof` gives Some, but if we ever did, then we would end up hitting the `.ok_or(RelayerError::MetadataFetchingError)?` which will return an Err when we try to format the metadata. Which we really don't want to happen

rust/agents/relayer/src/msg/metadata/multisig/message_id_multisig.rs
```diff
@@ -50,10 +51,15 @@ impl MultisigIsmMetadataBuilder for MessageIdMultisigMetadataBuilder {
             );
             return Ok(None);
         }
+        let merkle_leaf_id = self
```
tkporter, line 54: Ditto this should short circuit if it's None imo

rust/agents/relayer/src/processor.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use async_trait::async_trait;
+use derive_new::new;
+use eyre::Result;
+use hyperlane_core::HyperlaneDomain;
+use tokio::task::JoinHandle;
+use tracing::{info_span, instrument, instrument::Instrumented, Instrument};
+
+#[async_trait]
+pub trait ProcessorExt: Send + Debug {
```
tkporter, line 11: Ext == extension? not familiar with this naming convention

rust/agents/relayer/src/relayer.rs
```diff
@@ -319,21 +349,41 @@ impl Relayer {
...
+    ) -> Instrumented<JoinHandle<Result<()>>> {
+        let metrics = MerkleTreeProcessorMetrics::new();
+        let merkle_tree_processor = MerkleTreeProcessor::new(
+            self.dbs.get(origin).unwrap().clone(),
+            metrics,
+            self.prover_syncs[origin].clone(),
+        );
+
+        let span = info_span!("MerkleTreeProcessor", origin=%merkle_tree_processor.domain());
+        let processor = Processor::new(Box::new(merkle_tree_processor));
```
tkporter, line 378: ah this is cool

rust/agents/relayer/src/merkle_tree/builder.rs
daniel-savu: problem is that if you `map`/`and_then` you end up with nested `Option<Result<Option<...>>>`. Managed to skip one let-else though

rust/agents/validator/src/submit.rs
```diff
@@ -52,12 +53,12 @@ impl ValidatorSubmitter {
         Checkpoint {
             root: tree.root(),
             index: tree.index(),
-            mailbox_address: self.mailbox.address(),
-            mailbox_domain: self.mailbox.domain().id(),
+            merkle_tree_hook_address: self.merkle_tree_hook.address(),
+            mailbox_domain: self.merkle_tree_hook.domain().id(),
```
tkporter, line 57: nit: imo now that we renamed `mailbox_address` we should call this `merkle_tree_hook_domain` or something. Or maybe `local_domain`, or `domain`

rust/agents/validator/src/validator.rs
```diff
@@ -112,22 +120,26 @@ impl BaseAgent for Validator {
...
+                Ok(_) => {
+                    tasks.push(self.run_message_sync().await);
+                    for checkpoint_sync_task in self.run_checkpoint_submitters().await {
+                        tasks.push(checkpoint_sync_task);
+                    }
+                    break;
+                }
+                _ => {
+                    // Future that immediately resolves
+                    return tokio::spawn(ready(Ok(()))).instrument(info_span!("Validator"));
```
tkporter, line 140: why this? feels weird, unsure if this is a common practice though

rust/agents/relayer/src/merkle_tree/processor.rs
daniel-savu: this was more like me making sense of the architecture, will remove

rust/agents/relayer/src/msg/gas_payment/mod.rs
daniel-savu: I think what tests is whether the cli is robust against casing inconsistencies and that `originDomain` will still work

rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -307,111 +301,27 @@ where
...
-    #[instrument(skip(self))]
-    #[allow(clippy::needless_range_loop)]
-    async fn tree(&self, lag: Option<NonZeroU64>) -> ChainResult<IncrementalMerkle> {
-        let lag = lag.map(|v| v.get()).unwrap_or(0).into();
-
-        // use consistent block for all storage slot or view calls to prevent
-        // race conditions where tree contents change between calls
-        let fixed_block_number = self
+        let lag = maybe_lag.map(|v| v.get()).unwrap_or(0).into();
+        let fixed_block_number: BlockNumber = self
```
tkporter, line 305: Can we actually move back to the `let base_call = self.contract.count();` pattern? the nice part of it is that if there isn't a lag configured, then we save an RPC, and the eth_call is just made against the latest block. This has proved nice in practice

The reason why we needed to have the `fixed_block_number` logic before always specify a block number is because we were getting the tree by making multiple eth_getStorageAt RPCs at a specific height. Now that we can just call `tree()` with a single RPC, we don't need to worry about making multiple calls accidentally at different heights, so we can go back to the pattent we use elsewhere

rust/chains/hyperlane-ethereum/src/mailbox.rs
tkporter: what does this mean? i'm missing the off by one bit

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
tkporter: +1 this 100% needs refactoring at some point but it is a big undertaking

I'd probably put this in the realm of https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2313

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+#[async_trait]
+impl<M> MerkleTreeHook for EthereumMerkleTreeHook<M>
+where
+    M: Middleware + 'static,
+{
+    #[instrument(skip(self))]
+    async fn latest_checkpoint(&self, maybe_lag: Option<NonZeroU64>) -> ChainResult<Checkpoint> {
+        let lag = maybe_lag.map(|v| v.get()).unwrap_or(0).into();
+
+        let fixed_block_number: BlockNumber = self
```
tkporter, line 208: ditto let's move back to the old pattern where we don't always need to make an explicit get_block_number call if the lag is None

rust/hyperlane-core/src/types/checkpoint.rs
```diff
@@ -10,7 +10,7 @@ use crate::{utils::domain_hash, Signable, Signature, SignedType, H160, H256};
 #[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
 pub struct Checkpoint {
     /// The mailbox address
-    pub mailbox_address: H256,
+    pub merkle_tree_hook_address: H256,
     /// The mailbox chain
     pub mailbox_domain: u32,
```
tkporter, line 15: I think i called this out in a diff comment but we should rename & change the `///` comments

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+            index,
+        })
+    }
+
+    #[instrument(skip(self))]
+    #[allow(clippy::needless_range_loop)]
+    async fn tree(&self, maybe_lag: Option<NonZeroU64>) -> ChainResult<IncrementalMerkle> {
+        let lag = maybe_lag.map(|v| v.get()).unwrap_or(0).into();
+
+        let fixed_block_number: BlockNumber = self
```
tkporter, line 235: samesies

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+            .block(fixed_block_number)
+            .call()
+            .await?;
+        let branch = raw_tree
+            .branch
+            .iter()
+            .map(|v| v.into())
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
```
tkporter, line 256: should we be unwrapping? dunno if this was there before - if it was, worth checking if we can easily remove, but if we can't then nbd

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let fixed_block_number: BlockNumber = self
+            .provider
+            .get_block_number()
+            .await
+            .map_err(ChainCommunicationError::from_other)?
+            .saturating_sub(lag)
+            .into();
+
+        // TODO: implement From<Tree> for IncrementalMerkle
```
tkporter, line 243: may not wanna do this because the `Tree` type is ethereum-specific and we should avoid leaking types like that into vm-agnostic code

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let tree = IncrementalMerkle::new(branch, raw_tree.count.as_usize());
+
+        Ok(tree)
+    }
+
+    #[instrument(skip(self))]
+    async fn count(&self, maybe_lag: Option<NonZeroU64>) -> ChainResult<u32> {
+        let lag = maybe_lag.map(|v| v.get()).unwrap_or(0).into();
+        let fixed_block_number: BlockNumber = self
```
tkporter, line 266: same

rust/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -283,11 +283,7 @@ impl std::fmt::Debug for SealevelMailbox {
 impl Mailbox for SealevelMailbox {
     #[instrument(err, ret, skip(self))]
     async fn count(&self, _maybe_lag: Option<NonZeroU64>) -> ChainResult<u32> {
-        let tree = self.tree(_maybe_lag).await?;
-
-        tree.count()
-            .try_into()
-            .map_err(ChainCommunicationError::from_other)
+        <Self as MerkleTreeHook>::count(self, _maybe_lag).await
```
tkporter, line 286: whoa

rust/hyperlane-base/src/contract_sync/cursor.rs
```diff
@@ -225,11 +225,11 @@ impl ForwardMessageSyncCursor {
                 .retrieve_dispatched_block_number(self.cursor.sync_state.next_sequence)
                 .await
             {
-                debug!(next_block = block_number, "Fast forwarding next block");
+                info!(next_block = block_number, "Fast forwarding next block");
```
tkporter, line 228: wondering if this is intentional here and down below

tkporter (commented):
I'd definitely say all the lag block number stuff is not critical to change immediately btw, just should do it at some point

rust/agents/validator/src/validator.rs
daniel-savu: hmm, this is a leftover from before the merkle tree artifact was added. If querying `count` fails on the merkle tree trait object fails, the validator assumes it can gracefully terminate because it's not required to sign anything

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -152,6 +154,22 @@ impl HyperlaneRocksDB {
         Ok(true)
     }
 
+    /// Store the merkle tree insertion event, and also store a mapping from message_id to leaf_index
+    pub fn process_tree_insertion(&self, insertion: &MerkleTreeInsertion) -> DbResult<bool> {
+        if let Ok(Some(_)) = self.retrieve_merkle_tree_insertion_by_leaf_index(&insertion.index()) {
```
tkporter, line 159: Iiuc this short circuiting means that double insertion won't work correctly? E.g. if index 10 and 11 are a double insertion, then index 10 will get stored, but then you will skip storing index 11, which means that the merkle tree processor logic here will break https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2746/files#diff-88d3caf1bed3eeed9ef0ee7e9ac5743c5e5c401e47c9ba77c6257e5ccb5f4a01R73

rust/agents/relayer/src/processor.rs
daniel-savu: Still in RFC but fairly common
https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
daniel-savu: the key here is the merkle tree leaf index rather than the message nonce, which doesn't skip on index 11. Or am I misunderstanding?

tkporter: Oh oops my bad, you're right

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -140,6 +143,37 @@ impl ChainConf {
...
 
+    /// Try to convert the chain setting into a Merkle Tree Hook contract
+    pub async fn build_merkle_tree_hook(
+        &self,
+        metrics: &CoreMetrics,
+    ) -> Result<Box<dyn MerkleTreeHook>> {
+        let ctx = "Building merkle tree hook";
+        // TODO: if the merkle tree hook is set for sealevel, it's still a mailbox program
+        // that the connection is made to using the pda seeds, which will not be usable.
+        let address = self
```
tkporter, line 154: I don't like the silently different behavior tbh

I think we should make merkle_tree_hook mandatory for now, and just change Sealevel in the JSON configs to specify the Mailbox address as the merkleTreeHook

Eventually I could see us wanting to make hooks in general optional, but that feels out of the scope of what this PR should do

rust/sealevel/programs/ism/multisig-ism-message-id/src/processor.rs
```diff
@@ -599,7 +599,7 @@ pub mod test {
             // is handled in compliance with what the Mailbox expects
             InterchainSecurityModuleInstruction::Verify(VerifyInstruction {
                 metadata: MultisigIsmMessageIdMetadata {
-                    origin_mailbox: checkpoint.mailbox_address,
+                    origin_mailbox: checkpoint.merkle_tree_hook_address,
```
tkporter, line 602: nit: we should change `origin_mailbox` too

rust/utils/run-locally/src/invariants.rs
```diff
@@ -63,19 +63,20 @@ pub fn termination_invariants_met(
...
-    if gas_payment_events_count < total_messages_expected {
-        log!(
-            "Relayer has {} gas payment events, expected at least {}",
-            gas_payment_events_count,
-            total_messages_expected
-        );
-        return Ok(false);
-    }
+    // TODO: fix this once eth gas payments are introduced
+    // if gas_payment_events_count < total_messages_expected {
```
tkporter, line 67: I assume we'll just revisit this later?

rust/agents/relayer/src/msg/metadata/base.rs
daniel-savu: That makes sense. So we should be able to remove the `backoff` dep entirely

rust/agents/relayer/src/processor.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#[derive(new)]
+pub struct Processor {
+    ticker: Box<dyn ProcessorExt>,
+}
+
+impl Processor {
+    pub fn spawn(self) -> Instrumented<JoinHandle<Result<()>>> {
+        let span = info_span!("MessageProcessor");
```
daniel-savu, line 27: Flagging to myself that the info span here should not be hardcoded

rust/chains/hyperlane-ethereum/src/mailbox.rs
yorhodes: like mailbox nonce is count - 1

rust/chains/hyperlane-ethereum/src/mailbox.rs
yorhodes: good callout

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
yorhodes: isnt this the ethereum crate?

rust/utils/run-locally/src/invariants.rs
daniel-savu: I think this was commented out because double insertions wouldn't work without setting the gas to zero, which made this condition fail

rust/chains/hyperlane-ethereum/src/mailbox.rs
daniel-savu: let me know if I should subtract one or not

rust/agents/relayer/src/merkle_tree/builder.rs
daniel-savu: Found a way to neatly chain everything and removed let-else entirely 🙆‍♂️

rust/chains/hyperlane-ethereum/src/mailbox.rs
daniel-savu: resolving this as there's no off-by-one difference

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
tkporter: I was originally saying this bc `IncrementalMerkle` is in hyperlane-core and this would require adding `hyperlane-ethereum` as a dependency to `hyperlane-core` (which imo we can't do), but maybe we could instead just do `Into<IncrementalMerkle> for Tree` inside ethers-ethereum

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
daniel-savu: the unwrap is safe here. It's required because although we're iterating over elements of a fixed-size array (`[u8; 32]`), `collect` can only produce a collection that implements `FromIterator` - which isn't the case for fixed-size arrays because an iterator has no knowledge about how many items it's iterating over. So we need the `try_into()` to convert the `Vec<H256>` into `[H256; 32]`. Alternative would be
```rust
        let mut branch: [H256; 32] = [Default::default(); 32];
        raw_tree
            .branch
            .iter()
            .enumerate()
            .for_each(|(index, v)| branch[index] = v.into());
```
This code can still panic with an out-of-bounds access if `raw_tree.branch` exceeds 32, and would probably be less obvious to debug than with the unwrap version. I'd prefer leaving as-is and adding a comment, what's your take?

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
daniel-savu: implemented inside the merkle tree hook module (this same file) since that's where it's used

daniel-savu: btw that comment was already there. in general I kept the merkle tree code intact but it's good we get to revisit these

rust/hyperlane-base/src/contract_sync/cursor.rs
daniel-savu: was meant only for debugging, thanks for pointing out

rust/hyperlane-base/src/settings/chains.rs
daniel-savu: just flagging that even though we have dummy implementations for the merkle tree indexers on sealevel, we won't be covering sealevel in e2e so panics could still occur

PR #2758, opened 2023-09-28
Sealevel IGP debugging, temporarily change Sealevel gas payment policy

tkporter:
### Description

- Some small changes to help with debugging sealevel IGP
- Turns out the GasPaymentAccount data format onchain is different from the one in `main`. This causes IGP indexing to not work in the wild, but it works in e2e. Because we were planning to change the IGP anyways as we transfer ownership to Zebec, this isn't so bad - so temporarily just not enforcing ZBC warp route Solana -> Nautilus gas payments. In practice, we're still effectively enforcing gas payments because the warp route will pay for gas

### Drive-by changes

- Also added the new BSC POSE addy

### Related issues

n/a

### Backward compatibility

yes

### Testing

Deployed, poked around, used cmds

### DIFF
--- a/rust/sealevel/client/src/main.rs
+++ b/rust/sealevel/client/src/main.rs
@@ -422,9 +422,9 @@ struct PayForGasArgs {
     program_id: Pubkey,
     #[arg(long)]
     message_id: String,
-    #[arg(long, default_value_t = 13376)]
+    #[arg(long)]
     destination_domain: u32,
-    #[arg(long, default_value_t = 10000)]
+    #[arg(long)]
     gas: u64,
 }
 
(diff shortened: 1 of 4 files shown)


rust/sealevel/client/src/main.rs
```diff
@@ -1357,8 +1414,8 @@ fn process_igp_cmd(ctx: Context, cmd: IgpCmd) {
                     Some(overhead_igp_account),
                     unique_gas_payment_keypair.pubkey(),
                     H256::from_str(&payment_details.message_id).unwrap(),
-                    13376,
-                    100000,
+                    payment_details.destination_domain,
+                    payment_details.gas,
```
daniel-savu, line 1418: 🙏

daniel-savu (approved):
LGTM

Issue #2759, opened 2023-09-28
ISM factory implementations are not added to verification.json


tkporter:
Ran into this when dealing with verifying DomainRoutingIsms. I think this is also an issue for the multisig ISM variants. Not sure why we didn't have this issue with aggregation ISMs?

The problem is that etherscan's minimal proxy verification requires the implementation to be verified. Atm, it seems like our tooling only verifies the *factory* contract, but not the implementation that the factory deploys in its constructor.

Specifically, for the domain routing ISM case:
- `DomainRoutingIsmFactory` is populated in verification.json, and is therefore able to be easily verified (https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet2/ism/verification.json#L15-L20)
- However the address of the `DomainRoutingIsm` implementation itself that's created here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/isms/routing/DomainRoutingIsmFactory.sol#L49 is *not* put into the verification.json

As a workaround, I needed to find the implementation address, which is annoyingly internal and immutable so there isn't an easy way to get it via RPCs. My process was to look on etherscan at the tx in which the factory was deployed, and then under `Internal Txns` you can see the implementation contract being created: e.g. https://celoscan.io/tx/0x5bed132edc5db3be667cef1e8a59fcf3c02d1d569db096a5daa4f87164dddd1e#internal

I think we need to make sure that this implementation is added to verification.json by our tooling. I'm not so sure there's an easy way for us to even get the address of the implementation in our deploy tooling because it's internal and immutable. Wondering if we need to maybe make it public or add an event so our tooling can pick this up?

PR #2761, opened 2023-09-28
Make immutable implementation of factories public

yorhodes:
### Description

- Makes `implementation` of proxy factories public for client use
- Adds ISM factory implementations to verification artifacts

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2759

### Backward compatibility

No, needs redeployment

### Testing

None


solidity/contracts/libs/StaticAddressSetFactory.sol
```diff
@@ -86,13 +86,45 @@ abstract contract StaticMOfNAddressSetFactory {
...
     {
         bytes memory _metadata = abi.encode(_values, _threshold);
-        bytes memory _bytecode = MetaProxy.bytecode(_implementation, _metadata);
+        bytes memory _bytecode = MetaProxy.bytecode(implementation, _metadata);
         bytes32 _salt = keccak256(_metadata);
         return (_salt, _bytecode);
     }
 }
+
+abstract contract StaticAddressSetFactory is StaticMOfNAddressSetFactory {
```
tkporter, line 100: I kinda liked `StaticNAddressSetFactory` as it is really clear how it differentiates from `StaticMOfNAddressSetFactory`, but don't feel strongly at all

tkporter (approved):
🔥

PR #2767, opened 2023-09-29
Make recipientIsm more robust

yorhodes:
### Description

- Updates mailbox `recipientIsm` to handle more edge cases

### Related issues

> Someone in our Discord ran into this problem ages ago and Dan just did as well. If a recipient contract doesn't specify their own ISM by not having an interchainSecurityModule() fn but they happen to have a fallback function, then the Mailbox's recipientIsm function reverts. The reason is that the Mailbox will attempt to decode empty bytes after calling the recipient's interchainSecurityModule() function into an address.

Fixes https://github.com/hyperlane-xyz/issues/issues/629

### Backward compatibility

Yes

### Testing

Unit tests


tkporter (approved):
nice 🚀

PR #2780, opened 2023-10-05
V3 agents feedback

daniel-savu:
### Description

Part of the remediations from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2746


rust/hyperlane-core/src/types/checkpoint.rs
```diff
@@ -9,10 +9,10 @@ use crate::{utils::domain_hash, Signable, Signature, SignedType, H160, H256};
 /// An Hyperlane checkpoint
 #[derive(Copy, Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
 pub struct Checkpoint {
-    /// The mailbox address
+    /// The merkle tree hook address
     pub merkle_tree_hook_address: H256,
-    /// The mailbox chain
-    pub mailbox_domain: u32,
+    /// The merkle tree hook chain
+    pub merkle_tree_hook_domain: u32,
```
tkporter, line 15: Sorry I know I was confusing here - but because we've already spun up the testnet4 validators and this changes the serialization format of checkpoints in S3, changing this is a backward incompatible change. It'd require all the validators to have their S3 buckets cleared out, which isn't easy to do atm, so I think we should just stay with `mailbox_domain` to preserve the compatibility with the existing checkpoints signed by validators

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -230,53 +242,35 @@ where
...
-
-        let fixed_block_number: BlockNumber = self
-            .provider
-            .get_block_number()
-            .await
-            .map_err(ChainCommunicationError::from_other)?
-            .saturating_sub(lag)
-            .into();
+        let mut call = self.contract.tree();
+        if let Some(lag) = maybe_lag {
```
tkporter, line 246: wonder if it's possible to easily dedupe this logic? like a function that takes in a generic call and the `maybe_lag` and then applies this logic

rust/hyperlane-core/src/utils.rs
```diff
@@ -225,4 +225,28 @@ macro_rules! many_to_one {
...
+/// // before using the macro:
+/// let Some(idx) = self.index_of_next_key()
+/// else {
+///     return Ok(None);
+/// };
+/// // after:
+/// unwrap_or_none_result!(idx, self.index_of_next_key());
+/// ```
+#[macro_export]
+macro_rules! unwrap_or_none_result {
```
tkporter, line 242: nice

tkporter (approved):

PR #2783, opened 2023-10-05
Stash deploy progress

yorhodes:
### Description

- Deployer changes and artifacts from testnet4 deployment

### Drive-by changes

- Adds 

### Backward compatibility

Yes

### Testing

Manual


rust/config/testnet4_config.json
```diff
@@ -0,0 +1,616 @@
...
+          "family": "blockscout"
+        }
+      ],
+      "blocks": {
+        "confirmations": 1,
+        "reorgPeriod": 0,
+        "estimateBlockTime": 5
+      },
+      "isTestnet": true,
+      "merkleRootMultisigIsmFactory": "0xa9C7e306C0941896CA1fd528aA59089571D8D67E",
```
tkporter, line 39: Not for the scope of this PR but having all these unused addresses in the Rust config feels like something we should change at some point

I never deeply reviewed the config change PR so that's on me for not flagging earlier

nambrot (approved):

PR #2793, opened 2023-10-10
Add chain logos

aroralanuk:
### Description

- Add logos for scroll, base, and polygon zkevm

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

None


jmrossy (commented):
I believe the color base.svg is failing to render here because it still has a ` {...props}` line

solidity/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@hyperlane-xyz/core",
   "description": "Core solidity contracts for Hyperlane",
-  "version": "1.5.4-beta0",
+  "version": "1.5.7-beta1",
```
jmrossy, line 4: These should be 3.x now, suggest excluding the version changes from this PR

jmrossy (approved):

PR #2816, opened 2023-10-16
Support multiple origins in PI deployments

### LINKED ISSUE #2813: Support multiple origins in PI deployments
Currently CLI and hyp-deploy only do one actual origin at a time despite potentially deploying to multiple chains (because that's how the deployers in the SDK work).

I propose we replace the two-step origin + remotes selection in the CLI with a single chain multi-select step and assume they should all be interconnected.

jmrossy:
### Description

- Support multiple origins in PI deployments
- Simplify ci test with single core deploy command
- Improve output format of chains list command

### Related issues

Fixes #2813 

### Testing

Tested locally and with e2e test

### DIFF
--- a/typescript/cli/ci-test.sh
+++ b/typescript/cli/ci-test.sh
@@ -22,25 +22,13 @@ set -e
 
 echo "{}" > /tmp/empty-artifacts.json
 
-echo "Deploying contracts to anvil1"
+echo "Deploying contracts to anvil1 and anvil2"
 yarn workspace @hyperlane-xyz/cli run hyperlane deploy core \
-    --chains ./examples/anvil-chains.yaml \
+    --chain-configs ./examples/anvil-chains.yaml \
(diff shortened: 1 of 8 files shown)


yorhodes (approved):
looks good
I would prefer to reuse the SDK deployer's `deploy(chainMap)` than replicate those loops here as sometimes the logic is more sophisticated than just `(for chain in chains deployContracts(chain))`

jmrossy:
> looks good I would prefer to reuse the SDK deployer's `deploy(chainMap)` than replicate those loops here as sometimes the logic is more sophisticated than just `(for chain in chains deployContracts(chain))`

Thanks! In fact we do let the deployers handle it. The loops/reductions are for creating the configs. The one exception of for the ISM deployments via the ISM Factory, which IIUC will need to be changed soon anyway?

PR #2820, opened 2023-10-18
Verify testnet4 deploy

yorhodes:
### Description

- Adds tooling for flattening source code for use in `solidity-single-file` verification
- Recovers all verification artifacts in HyperlaneDeployer from cached addresses

### Drive-by changes

- Adjusts hardhat compiler settings to be pipe through to typechain
- Removes solana from testnet4
- Migrate to etherscan family explorer for scrollsepolia

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/635

### Backward compatibility

- Yes

### Testing

- Verified testnet4 contracts


solidity/flatten.sh
```diff
@@ -0,0 +1,23 @@
+LICENSE="// SPDX-License-Identifier: MIT OR Apache-2.0"
```
nambrot, line 1: Maybe add instructions in here how to generally verify

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -266,21 +271,58 @@ export abstract class HyperlaneDeployer<
   ): Promise<ReturnType<F['deploy']>> {
     const cachedContract = this.readCache(chain, factory, contractName);
     if (cachedContract) {
+      const provider = this.multiProvider.getProvider(chain);
```
nambrot, line 274: Should we check this in? Might be nice to put this in a separate function and then comment calling it out if we need this in the future? Also, it would just duplicate them now otherwise no?

nambrot (approved):

PR #2833, opened 2023-10-23
Testnet4 kathy fixes

aroralanuk:
### Description

- removing `.relayers` from `Values.hyperlane.chains` in external secrets
- redeploying helloworld on `goerli,sepolia,bsctestnet` and running kathy on k8s

### Drive-by changes

none

### Related issues

none

### Backward compatibility

yes

### Testing

Manual


nambrot (approved):

typescript/sdk/src/consts/environments/testnet.json
```diff
@@ -190,22 +190,5 @@
     "protocolFee": "0x244d1F7e30Be144A87602905baBF86630e8f39DC",
     "mailbox": "0x2d1889fe5B092CD988972261434F7E5f26041115",
     "validatorAnnounce": "0x99303EFF09332cDd93E8BC8b2F07b2416e4501e5"
-  },
```
jmrossy, line 193: @tkporter added this so he should confirm if we're okay removing it. As I recall, here's how it went with these artifacts:
1. First I tried to add it, found it broke a bunch of SDK tooling, so removed it
2. Then I hacked the chain metadata + artifacts into Kathy manually
3. Then @tkporter added these artifacts and updated the SDK to not choke
4. Then @aroralanuk commented out my Kathy hack: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/v3/typescript/infra/scripts/helloworld/utils.ts#L89

Not totally sure where we're at now with V3 + kathy + sealevel, can you clarify?

nambrot: what was the choke? Also found the presence of these addresses to be an issue

nambrot (approved):

typescript/sdk/src/consts/environments/testnet.json
jmrossy: @nambrot It was from places where the tooling iterates over a chain map and does things for each chain assuming they're all EVM, like instantiating ethers factories for example. 
Trevor remedied it with utils like this but it's possible some paths remain unfixed: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/contracts/contracts.ts#L69

nambrot: So wouldn't solanadevnet not having in there be better? That way we are not iterating over a chainmap for which evm chain metadata doesn't exist?

jmrossy: We can make it work without the sol artifacts in there but there are downsides too. That's why I had to weave together metadata + artifacts to get Kathy to work (before Trevor added this). In general, the tooling expects addresses to be there for core chains. 

The better fix is to rethink the artifact management + tooling, which we're kinda doing (artifacts at least).

For now, I'd recommend just doing whatever works. I only commented to point out that removing this may break deploying tool trevor put together for sealevel.

tkporter: +1 to doing whatever works for now and we can revisit when we add sealevel support for v3. My feeling is we're in a weird half state in v3 where solanadevnet is present in some places but not others

PR #2836, opened 2023-10-24
Rm finality blocks and refactor checkpoint_submitter to be resilient to edge cases

tkporter:
### Description

* https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/532 wasn't actually fully closed out - finality_blocks was used still for indexing. Sadly testnet4 infra wasn't configuring finality blocks anymore, so we weren't indexing only finalized blocks. Moved fully to reorg_period
* Refactored `checkpoint_submitter` in light of races revealed by the above ^ problem. It used to assume that message indexing and the `latest_checkpoint()` call would align with one another, and wasn't resilient to the tree already being past the correctness checkpoint. A couple situations were possible before that aren't now:
  a. Indexing is ahead of the latest_checkpoint() call, which will result in tree ingesting the new indexed messages and the tree being ahead of the correctness checkpoint :(
  b. It's possible for the tree() call that constructs the tree initially to be made against a block that's after the next latest_checkpoint() call, which would result in the tree being ahead of the correctness checkpoint from the very beginning :(

### Drive-by changes

removed a function that wasn't being used anymore

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/532

### Backward compatibility

Removes finality blocks entirely

### Testing

Builds, e2e

### DIFF
--- a/rust/agents/validator/src/submit.rs
+++ b/rust/agents/validator/src/submit.rs
@@ -209,6 +209,7 @@ impl ValidatorSubmitter {
         Ok(())
     }
 
+    // Signs and submits any previously unsubmitted checkpoints.
     async fn sign_and_submit_checkpoints(
         &self,
         checkpoints: Vec<CheckpointWithMessageId>,
--- a/rust/agents/validator/src/submit.rs
+++ b/rust/agents/validator/src/submit.rs
@@ -209,7 +209,7 @@ impl ValidatorSubmitter {
         Ok(())
     }
 
-    // Signs and submits any previously unsubmitted checkpoints.
+    /// Signs and submits any previously unsubmitted checkpoints.
     async fn sign_and_submit_checkpoints(
         &self,
         checkpoints: Vec<CheckpointWithMessageId>,
--- a/rust/agents/validator/src/submit.rs
+++ b/rust/agents/validator/src/submit.rs
@@ -98,10 +98,7 @@ impl ValidatorSubmitter {
             //
             // In this case, we just sleep a bit until we fetch a new latest checkpoint
             // that at least meets the tree.
-            //
-            // tree.index() will panic if the tree is empty, so we use tree.count() instead
-            // and convert the latest_checkpoint.index to a count by adding 1.
-            if latest_checkpoint.index + 1 < tree.count() as u32 {
+            if !checkpoint_exceeds_tree(&latest_checkpoint, &tree) {
                 debug!(
                     ?latest_checkpoint,
                     tree_count = tree.count(),
@@ -130,11 +127,8 @@ impl ValidatorSubmitter {
         correctness_checkpoint: &Checkpoint,
     ) -> Result<()> {
         // This should never be called with a tree that is ahead of the correctness checkpoint.
-        //
-        // tree.index() will panic if the tree is empty, so we use tree.count() instead
-        // and convert the correctness_checkpoint.index to a count by adding 1.
         assert!(
-            correctness_checkpoint.index + 1 >= tree.count() as u32,
+            !checkpoint_exceeds_tree(correctness_checkpoint, &tree),
             "tree (count: {}) is ahead of correctness checkpoint {:?}",
             tree.count(),
             correctness_checkpoint,
@@ -321,6 +315,13 @@ impl ValidatorSubmitter {
             sleep(self.interval).await;
         }
     }
+}
+
+/// Returns whether the checkpoint's index exceeds that of the tree.
+fn checkpoint_exceeds_tree(checkpoint: &Checkpoint, tree: &IncrementalMerkle) -> bool {
+    // tree.index() will panic if the tree is empty, so we use tree.count() instead
+    // and convert the correctness_checkpoint.index to a count by adding 1.
+    checkpoint.index + 1 > tree.count() as u32
 }
 
 #[derive(Clone)]
--- a/rust/agents/validator/src/submit.rs
+++ b/rust/agents/validator/src/submit.rs
@@ -58,74 +58,67 @@ impl ValidatorSubmitter {
         }
     }
 
-    /// Submits signed checkpoints.
-    /// If a target checkpoint is provided, it must not be bethind the provided tree, and
-    /// the submitter will only submit checkpoints until it reaches the target checkpoint.
-    /// If a target checkpoint is not provided, the submitter will submit checkpoints indefinitely.
-    #[instrument(err, skip(self, tree), fields(domain=%self.merkle_tree_hook.domain()))]
-    pub(crate) async fn checkpoint_submitter(
+    /// Submits signed checkpoints from index 0 until the target checkpoint (inclusive).
+    /// Runs idly forever once the target checkpoint is reached to avoid exiting the task.
+    #[instrument(err, skip(self), fields(domain=%self.merkle_tree_hook.domain()))]
+    pub(crate) async fn backfill_checkpoint_submitter(
         self,
-        mut tree: IncrementalMerkle,
-        target_checkpoint: Option<Checkpoint>,
+        target_checkpoint: Checkpoint,
     ) -> Result<()> {
-        // If the target checkpoint is provided, the provided tree must not be behind it.
-        if let Some(target) = target_checkpoint {
-            assert!(tree.index() <= target.index);
-        }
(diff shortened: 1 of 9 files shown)


rust/agents/validator/src/submit.rs
```diff
@@ -58,17 +58,22 @@ impl ValidatorSubmitter {
...
     pub(crate) async fn checkpoint_submitter(
         self,
         mut tree: IncrementalMerkle,
         target_checkpoint: Option<Checkpoint>,
     ) -> Result<()> {
-        let mut checkpoint_queue = vec![];
-
-        let mut reached_target = false;
+        // If the target checkpoint is provided, the provided tree must not be behind it.
+        if let Some(target) = target_checkpoint {
```
nambrot, line 72: Isn't this mostly the case for the backfill submitter at startu[? I.e. the tree is empty, but the target is the current checkpoint on-chain?

rust/agents/validator/src/submit.rs
```diff
@@ -83,8 +88,64 @@ impl ValidatorSubmitter {
...
+
+    /// Submits signed checkpoints relating to the given tree until the correctness checkpoint (inclusive).
+    /// Only submits the signed checkpoints once the correctness checkpoint is reached.
+    async fn submit_checkpoints_until_correctness_checkpoint(
+        &self,
+        tree: &mut IncrementalMerkle,
+        correctness_checkpoint: &Checkpoint,
+    ) -> Result<()> {
+        // This should never be called with a tree that is ahead of the correctness checkpoint.
+        assert!(correctness_checkpoint.index >= tree.index());
```
nambrot, line 140: Can we add error messages here?

rust/agents/validator/src/submit.rs
```diff
@@ -83,8 +88,66 @@ impl ValidatorSubmitter {
...
+
+        // All intermediate checkpoints will be stored here and signed once the correctness
+        // checkpoint is reached.
+        let mut checkpoint_queue = vec![];
+
+        // If the correctness checkpoint is ahead of the tree, we need to ingest more messages.
+        //
```
nambrot, line 147: Nit: >= is more intuitive to me

nambrot: Though in the spirit of this PR, I feel like if strictly greater than occurs, then that's maybe worth logging

rust/agents/validator/src/submit.rs
```diff
@@ -96,76 +159,73 @@ impl ValidatorSubmitter {
...
                     checkpoint,
                     message_id,
                 });
+            } else {
+                // If we haven't yet indexed the next merkle tree insertion but know that
+                // it will soon exist (because we know the correctness checkpoint), wait a bit and
+                // try again.
+                sleep(Duration::from_millis(100)).await
+            }
+        }
```
nambrot, line 174: We only know >= no?

nambrot: Might good to do that assertion?

rust/agents/validator/src/submit.rs
```diff
@@ -83,8 +88,66 @@ impl ValidatorSubmitter {
...
-            // ingest available messages from DB
-            while let Some(insertion) = self
+            // This can only happen if target_checkpoint is None and the latest checkpoint
+            // fetched from onchain has an index that is behind the tree's index.
+            // This may occur e.g. if RPC providers are unreliable and make calls against
+            // inconsistent block tips.
+            //
+            // In this case, we just sleep a bit until we fetch a new latest checkpoint
+            // that at least meets the tree.
+            if correctness_checkpoint.index < tree.index() {
```
nambrot, line 98: Would be useful to add a debug line here?

rust/agents/validator/src/submit.rs
```diff
@@ -58,17 +58,22 @@ impl ValidatorSubmitter {
         }
     }
 
+    /// Submits signed checkpoints.
+    /// If a target checkpoint is provided, it must not be bethind the provided tree, and
```
nambrot, line 62: I wonder if we should just have two separate functions for when the target is provided vs. not. That imo would make the code more readable

rust/agents/validator/src/submit.rs
```diff
@@ -96,76 +159,73 @@ impl ValidatorSubmitter {
...
-                    info!(index = checkpoint.index, "Signed all queued checkpoints");
-
-                    self.metrics
-                        .latest_checkpoint_processed
-                        .set(checkpoint.index as i64);
-
-                    // break out of submitter loop if target checkpoint is reached
-                    reached_target = target_checkpoint.is_some();
-                    break;
+            // drain and sign all checkpoints in the queue
```
nambrot, line 183: Nit: Maybe move this logic into its own function

rust/agents/validator/src/submit.rs
```diff
@@ -96,76 +159,73 @@ impl ValidatorSubmitter {
...
+                index = checkpoint.index,
+                "Signed all queued checkpoints until index"
+            );
 
-        // TODO: remove this once validator is tolerant of tasks exiting
-        loop {
-            sleep(Duration::from_secs(u64::MAX)).await;
+            self.metrics
+                .latest_checkpoint_processed
+                .set(checkpoint.index as i64);
```
nambrot, line 217: Nit: I still like this condition ahead of the happy path as a "revert reason" instead of a true if/else branching condition

rust/agents/validator/src/submit.rs
tkporter: Yeah this was just to be safe without assuming how it's called, though now with the fn split this can go away

rust/agents/validator/src/submit.rs
tkporter: nah it only breaks out of the above while loop if the indices are the same

rust/agents/validator/src/submit.rs
tkporter: Personally feel that converting the index to a count with the + 1 makes it easier to understand, I'll make a better comment too

nambrot (approved):

rust/agents/validator/src/submit.rs
```diff
@@ -58,33 +58,92 @@ impl ValidatorSubmitter {
...
+
+        // All intermediate checkpoints will be stored here and signed once the correctness
+        // checkpoint is reached.
+        let mut checkpoint_queue = vec![];
+
+        // If the correctness checkpoint is ahead of the tree, we need to ingest more messages.
+        //
+        // tree.index() will panic if the tree is empty, so we use tree.count() instead
+        // and convert the correctness_checkpoint.index to a count by adding 1.
+        while correctness_checkpoint.index + 1 > tree.count() as u32 {
```
yorhodes, line 145: is there no `tree.index()`?

rust/agents/validator/src/submit.rs
```diff
@@ -96,76 +155,85 @@ impl ValidatorSubmitter {
...
+            debug!(
+                index = queued_checkpoint.index,
+                "Signed and submitted checkpoint"
+            );
 
-        // TODO: remove this once validator is tolerant of tasks exiting
-        loop {
-            sleep(Duration::from_secs(u64::MAX)).await;
+            // small sleep before signing next checkpoint to avoid rate limiting
+            sleep(Duration::from_millis(100)).await;
```
yorhodes, line 234: did you encounter this? assume this is specific to KMS so might be nice not to have it here but not critical

rust/agents/validator/src/submit.rs
yorhodes: can you use `tree_exceeds_checkpoint` here?

tkporter: See the comment above on tree.index(), it panics if the count is 0 so this is the unfortunate workaround

And I don't think so because `tree_exceeds_checkpoint` returns true if tree.index > checkpoint.index (so doing `!tree_exceeds_checkpoint` would be if tree.index <= checkpoint.index) and we want the condition here to be checkpoint.index > tree.index

rust/agents/validator/src/submit.rs
tkporter: No I just copied this from the existing logic, figured this was from that weird KMS thing too so didn't want to change

Issue #2837, opened 2023-10-24
Remove legacy checkpoints from V3 agents


tkporter:
- [x] make sure legacy multisigs aren't used whatsoever in v3
- [x] remove legacy checkpoint metadata building in relayer
- [x] remove legacy checkpoint submission task in validator

PR #1112, opened 2022-09-23
Info log in validator when polling checkpoints

tkporter:
### Description

Following some feedback, it's not super clear if a validator is working correctly because we don't log much

### Drive-by changes

Minor log changes

### Related issues

n/a

### Backward compatibility

_Are these changes backward compatible?_

Yes

_Are there any infrastructure implications, e.g. changes that would prohibit deploying older commits using this infra tooling?_

None

### Testing

_What kind of testing have these changes undergone?_

None

PR #2839, opened 2023-10-24
Remove legacy multisig from agents

yorhodes:
### Description

- Deprecates legacy multisig support in agents

### Drive-by changes

- Moves `order_signatures` from `metadata_builder` to `checkpoint_syncer`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2837

### Backward compatibility

- relayer no longer supports delivering to legacy multisig
- validator no longer supports signing checkpoints without message IDs

### Testing

E2E


rust/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -87,18 +78,12 @@ pub trait MultisigIsmMetadataBuilder: AsRef<BaseMetadataBuilder> + Send + Sync {
...
-                        .map(|x| Token::FixedBytes(x.to_fixed_bytes().into()))
-                        .collect();
-                    Ok(ethers::abi::encode(&[Token::FixedArray(validator_tokens)]))
-                }
-                MetadataToken::Signatures => {
-                    let ordered_signatures = order_signatures(validators, &metadata.signatures);
-                    let threshold_signatures = &ordered_signatures[..threshold as usize];
-                    Ok(threshold_signatures.concat())
-                }
+                MetadataToken::Signatures => Ok(metadata
```
tkporter, line 81: If we're removing `order_signatures` and grabbing only `threshold` signatures, can you make sure we document somewhere where we get that guarantee? E.i. `rust/hyperlane-base/src/types/multisig.rs`'s `fetch_checkpoint` now must return exactly `threshold` signatures and in the same order as `validators`. It does this in practice but is never explicitly documented, hence why we had this logic here

rust/agents/validator/src/submit.rs
```diff
@@ -235,86 +235,6 @@ impl ValidatorSubmitter {
...
-            let latest_checkpoint = self
-                .merkle_tree_hook
-                .latest_checkpoint(self.reorg_period)
-                .await?;
-
-            self.metrics
-                .legacy_latest_checkpoint_observed
-                .set(latest_checkpoint.index as i64);
-
-            // Occasional info to make it clear to a validator operator whether things are
```
tkporter, line 278: Can we keep the occasional logs added in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/1112 ? They're not in the new submitter, but we got a lot of feedback from people that without it they weren't sure if the validator was actually working because of long periods between messages

rust/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -142,12 +127,9 @@ impl<T: MultisigIsmMetadataBuilder> MetadataBuilder for T {
             .await
             .context(CTX)?
         {
+            // TODO: assert ordering?
```
tkporter, line 130: rm? or did you have plans here

tkporter (approved):

Issue #2856, opened 2023-10-26
Validator latest index writing


yorhodes:
#2839 broke the merkle_root_multisig metadata builder by removing updates to the latest_index on the validator's checkpoint syncer

We should change the e2e test to be 2/2 aggregation of message_id and merkle_root to ensure this behavior is being tested

PR #2858, opened 2023-10-26
Fix validator checkpoint syncer latest index

yorhodes:
### Description

- Ensure validator writes latest index to checkpoint syncer

### Drive-by changes

- Adds `write_latest_index` to `checkpoint_syncer` trait

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2856

### Backward compatibility

Yes

### Testing

E2E


rust/agents/validator/src/submit.rs
```diff
@@ -136,6 +136,10 @@ impl ValidatorSubmitter {
             self.submit_checkpoints_until_correctness_checkpoint(&mut tree, &latest_checkpoint)
                 .await?;
 
+            self.checkpoint_syncer
+                .write_latest_index(latest_checkpoint.index)
```
tkporter, line 140: wonder if we should have some logic like this https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2839/files#diff-97b8300d4321afb2d386072aadf926cea1cc87f2e625e68bc427b25a176f1808L102-L109 to prevent a bunch of writes

rust/hyperlane-base/src/types/local_storage.rs
```diff
@@ -62,6 +62,14 @@ impl CheckpointSyncer for LocalStorage {
...
     }
 
+    async fn write_latest_index(&self, index: u32) -> Result<()> {
+        let path = self.latest_index_file_path();
+        tokio::fs::write(&path, index.to_string())
+            .await
+            .with_context(|| format!("Writing index to {path:?}"))?;
+        Ok(())
+    }
+
```
tkporter, line 72: I have no idea how this happened or we didn't see earlier but check out https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/latest-index-test/rust/hyperlane-base/src/types/local_storage.rs#L36-L39, this should be the same as the S3 one I think? https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b141aece73e6ec910c6d2d3294edad008d6db3df/rust/hyperlane-base/src/types/s3_storage.rs#L134-L136

Maybe this is why e2e is failing?

tkporter: Can we actually just put these strings into shared consts

rust/agents/validator/src/submit.rs
nambrot: agreed

rust/hyperlane-base/src/types/local_storage.rs
tkporter: actually hmm I don't think this would cause things to fail anyways

yorhodes: these don't need to be the same for the different modalities right? agree they probably should be shared

tkporter: yeah you're right, they don't need to be the same, my mistake

rust/agents/validator/src/submit.rs
yorhodes: not totally clear to me what we gain from this if we still have to do read IO?

nambrot (approved):

PR #2861, opened 2023-10-26
fix: e2e working with aggregation 2/2 of `message_id` and `merkle_root`

nambrot:
### Description

- Removed the double mapping logic from the the `merkle_tree_builder.get_proof()` for fixing the following issue: 
 > `get_proof` was expecting a nonce which it then uses to retrieve the `leaf_index` i.e. nonce -> message_id -> leaf_index and then `prove_against_previous` but in our case, we already got the leaf_index in merkle_tree_multisig.rs so what we end up doing is trying to look up the leaf_index twice and either not finding the message_id or proving wrong leaf_index which both causes the "cannot fetch metadata" for the merkle tree builder

### Drive-by changes

- additional logging for aggregation to indicate which ism (moduleType, address) we are missing
- trace -> debug logging for `sign_and_submit_checkpoint`

### Related issues

- closes https://github.com/hyperlane-xyz/issues/issues/695

### Backward compatibility

Yes

### Testing

e2e


rust/agents/relayer/src/merkle_tree/builder.rs
```diff
@@ -14,6 +14,7 @@ use crate::prover::{Prover, ProverError};
 /// Struct to sync prover.
 #[derive(Debug)]
 pub struct MerkleTreeBuilder {
+    #[allow(dead_code)]
```
yorhodes, line 17: ?

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -90,15 +90,15 @@ impl AggregationIsmMetadataBuilder {
...
+        invalid_isms: Vec<(ModuleType, H256)>,
     ) -> Option<Vec<SubModuleMetadata>> {
         let gas_cost_results: Vec<_> = join_all(
             sub_modules
                 .iter()
                 .map(|module| module.ism.dry_run_verify(message, &(module.meta.metadata))),
         )
         .await;
-
-        // Filter out the ISMs without a gas cost estimate
```
yorhodes, line 101: ?

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -107,7 +107,14 @@ impl AggregationIsmMetadataBuilder {
 
         let metas_and_gas_count = metas_and_gas.len();
         if metas_and_gas_count < threshold {
-            info!("Could not fetch all metadata: Found {metas_and_gas_count} of the {threshold} required ISM metadata pieces");
+            info!("Could not fetch all metadata: Found {metas_and_gas_count} of the {threshold} required ISM metadata pieces for message_id {message_id}", metas_and_gas_count=metas_and_gas_count, threshold=threshold, message_id=message.id());
+            for (module_type, ism_address) in invalid_isms {
+                info!(
+                    "Invalid ISM: {module_type} at {ism_address}",
```
yorhodes, line 113: Invalid? i don't think this is the only reason why estimate gas would fail

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -140,21 +147,34 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
+        let (valid_sub_modules, invalid_sub_modules): (Vec<_>, Vec<_>) = metas
             .into_iter()
             .enumerate()
             .zip(sub_modules.into_iter())
-            .filter_map(|((index, meta_result), sub_module_result)| {
-                match (meta_result, sub_module_result) {
-                    (Ok(Some(meta)), Ok(ism)) => Some(IsmAndMetadata::new(ism, index, meta)),
-                    _ => None,
-                }
+            .partition(|&((_, ref meta_result), ref sub_module_result)| {
```
yorhodes, line 154: i don't understand what these changes are

yorhodes (approved):
not sure what the unrelated changes are but probably @tkporter or @daniel-savu should approve first if you want to merge those

rust/agents/relayer/src/merkle_tree/builder.rs
aroralanuk: Otherwise the lint-rs workflow was failing

rust/agents/relayer/src/msg/metadata/aggregation.rs
aroralanuk: "// Filter out the ISMs with a gas cost estimate"

rust/agents/relayer/src/msg/metadata/aggregation.rs
aroralanuk: this means missing ism as in which don't satisfy the requirements of aggregation but good point, invalid is maybe not the best term for it.

rust/agents/relayer/src/msg/metadata/aggregation.rs
aroralanuk: right now the aggregation log is "could not find all metadata: found 1 out of 2 required ISMs" which doesn't tell which ism is not working so the partition separates the working and missing isms which helps to log which ISMs aren't working in case metas_and_gas_count < threshold

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -107,7 +107,14 @@ impl AggregationIsmMetadataBuilder {
 
         let metas_and_gas_count = metas_and_gas.len();
         if metas_and_gas_count < threshold {
-            info!("Could not fetch all metadata: Found {metas_and_gas_count} of the {threshold} required ISM metadata pieces");
+            info!("Could not fetch all metadata: Found {metas_and_gas_count} of the {threshold} required ISM metadata pieces for message_id {message_id}", metas_and_gas_count=metas_and_gas_count, threshold=threshold, message_id=message.id());
+            for (module_type, ism_address) in invalid_isms {
```
daniel-savu, line 111: I'd rather print all of these as part of the `info` log on the line above, since all are related to the same error. You can just add `?invalid_isms` to the list of args in `info`

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -140,21 +147,34 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
+                )
             })
             .collect();
 
+        let mut invalid_isms: Vec<(ModuleType, H256)> = Vec::new();
+        // get the address and module_type for each invalid sub module (for logging in cheapest_valid_metas)
+        for entry in &invalid_sub_modules {
+            let (_, ism_result) = entry;
+            if let Ok(ism_box) = ism_result {
+                invalid_isms.push((ism_box.module_type().await.unwrap(), ism_box.address()));
```
daniel-savu, line 173: We'd end up doing a lot of repeated queries with this unless we cache. 

I'd be happy to just print the ism addresses for now, that's still an improvement:
```rust
let invalid_isms: Vec<H256> = invalid_sub_modules.into_iter().map(|(_, addr)| addr).collect();
```

Also since `invalid_isms` isn't relevant to `cheapest_valid_metas`, you could take the `info!("Could not fetch all metadata: ...)` out of there and add it after this call if `maybe_aggregation_metadata.len() < threshold`

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -140,21 +147,34 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
-                    _ => None,
-                }
+            .partition(|&((_, ref meta_result), ref sub_module_result)| {
+                matches!((meta_result, sub_module_result), (Ok(Some(_)), Ok(_)))
+            });
+        let valid_sub_module_metas: Vec<_> = valid_sub_modules
+            .into_iter()
+            .map(|((index, meta_result), sub_module_result)| {
+                IsmAndMetadata::new(
+                    sub_module_result.unwrap(),
```
tkporter, line 161: We shouldn't have any unwraps, even if logically we know that `valid_sub_modules` when it's set above shouldn't panic here. It's a bit harder to understand and more prone to panics if we ever change code around here

If you take up my previous suggestion, we could do something like this. We have access to the itertools crate which gives us some extra power to return different left and right types using [partition_map](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_map)

```
// Partitions things into
// ok_sub_modules: Vec<IsmAndMetadata>, which we'll then use later in a call to cheapest_valid_metas
// err_sub_modules: Vec<(H256 /* ism address */, Option<ModuleType>)>, which keeps the ISM address
//    even if there is an error getting the module type, and optionally has the module type if it was successful

let (ok_sub_modules, err_sub_modules): (Vec<IsmAndMetadata>, Vec<(H256 /* ism address */, Option<ModuleType>)>) = ism_and_metas
  .into_iter()
  .enumerate()
  .partition_map(|(index, (ism_address, ism_and_meta_result)) | {
    match (ism_and_meta_result) {
      // Everything successful
      Ok((ism, _module_type, Some(meta)) => Either::Left(IsmAndMetadata::new(ism, index, meta)),
      // Meta was None, but we still got module_type
      Ok((_ism, module_type, None) => Either::Right((ism_address, Some(module_type)),
      // Unable to get module_type, still keep the ism address
      _ => Either::Right((ism_address, None)),
    }
  });
```

tkporter (commented):
lmk if you wanna chat about my suggestions - motivation is to remove all unwraps and to avoid the extra work just for logging purposes

rust/agents/relayer/src/merkle_tree/builder.rs
```diff
@@ -75,23 +76,16 @@ impl MerkleTreeBuilder {
...
-            })
-            .transpose()
-            .map_err(Into::into)
+        match self
+            .prover
+            .prove_against_previous(leaf_index as usize, root_index as usize)
+        {
+            Ok(proof) => Ok(Some(proof)),
+            Err(prover_err) => Err(MerkleTreeBuilderError::from(prover_err)),
+        }
```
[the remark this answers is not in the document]
aroralanuk: Don't we need to wrap it Option<> still? your suggestions give a 

"mismatched types
expected enum `Result<std::option::Option<Proof>, _>`
   found enum `Result<Proof, _>" error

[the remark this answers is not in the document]
daniel-savu: you can change the return signature to no longer require the Option

rust/agents/relayer/src/msg/metadata/aggregation.rs
aroralanuk: So, we're returning `(Box<dyn InterchainSecurityModule>, Option<Vec<u8>>, Module_Type>)` or `(Box<dyn InterchainSecurityModule>, Option<Vec<u8>>, Option<Module_Type>)`. If it's the latter, we'll be matching against different types in the `build` function instead of `build_ism_and_metadata`?

aroralanuk: `module_type()` revert can occur when we're calling build too?

aroralanuk: if this is the case, we don't need `Some(module_type)` in the partition, bc we already UnsupportedModuleType error?

tkporter: hmm reading this through rn

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -244,4 +223,42 @@ impl BaseMetadataBuilder {
...
+            ModuleType::Routing => Box::new(RoutingIsmMetadataBuilder::new(base)),
+            ModuleType::Aggregation => Box::new(AggregationIsmMetadataBuilder::new(base)),
+            ModuleType::Null => Box::new(NullMetadataBuilder::new()),
+            ModuleType::CcipRead => Box::new(CcipReadIsmMetadataBuilder::new(base)),
+            _ => return Err(MetadataBuilderError::UnsupportedModuleType(module_type).into()),
+        };
+        let meta = metadata_builder
+            .build(ism_address, message)
+            .await
+            .context("When building metadata");
```
tkporter, line 257: To have the same behavior as before we should still propagate the error if we get a Result::Err, so we should put a `?;` instead of the ok().flatten() down below

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -127,34 +128,38 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
-            .filter_map(|((index, meta_result), sub_module_result)| {
-                match (meta_result, sub_module_result) {
-                    (Ok(Some(meta)), Ok(ism)) => Some(IsmAndMetadata::new(ism, index, meta)),
-                    _ => None,
-                }
-            })
+            .filter_map(|sub_module_and_meta| sub_module_and_meta.ok())
             .collect();
-
+        let (ok_sub_modules, err_sub_modules): (Vec<_>, Vec<_>) = no_err_sub_modules_and_metas
```
tkporter, line 145: So my I was thinking was:
1. we want the union of the partitions `ok_sub_modules U err_sub_modules` to give us all the submodules in the aggregation ISM
2. for any err_sub_modules, we want to have as much info as possible about these sub_modules so we can log it. The most useful info is the ISM address and the module type, if we have it
  a. An error could occur for any of the reasons [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/base.rs#L77-L103), which includes being unable to build an ISM (likely an RPC error), unable to get the module_type (ism.module_type() could revert, or RPC error), too deep of a clone of base, invalid ModuleType, or building the metadata failed

Because of (1) I was imagining we wouldn't have the filter_map up above at all, and instead when we do the partition here we'd iter through the `sub_modules_and_meta` directly, and if one was an Err then we wouldn't have access to the module type, which is why I suggested it be optional. Atm I think `err_sub_modules` will just include submodules where we were just unable to build the metadata, but all other fallible steps were successful. Which probably covers most of the cases, but I think it'd be good to cover all the cases so we get as much info as possible when we log `err_isms`

So the cases for `err_sub_modules` would be:
- Result::Err
- Result::Ok, but metadata was None

In which case I wonder if we should just have err_sub_modules / err_isms be typed as something like `Vec<(H256, Result<IsmWithMetadataAndType>)>` so we get the addresses & result for each of the sub modules with errors

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -127,34 +128,38 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
+            .into_iter()
+            .enumerate()
+            .partition_map(
+                |(index, sub_module_and_meta)| match sub_module_and_meta.metadata {
+                    Some(_) => Either::Left(IsmAndMetadata::new(
```
tkporter, line 150: you can do `Some(metadata) => ` and then use metadata instead of `sub_module_and_meta.metadata.unwrap()`

rust/agents/relayer/src/msg/metadata/aggregation.rs
tkporter: See this comment https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2861#discussion_r1384867953

rust/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -127,34 +128,38 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
-                .map(|ism_address| self.base.build(*ism_address, message)),
-        )
-        .await;
-
-        let sub_modules = join_all(
+        let sub_modules_and_metas = join_all(
             ism_addresses
                 .iter()
-                .map(|ism_address| self.base.build_ism(*ism_address)),
+                .map(|ism_address| self.base.build_ism_and_metadata(*ism_address, message)),
```
tkporter, line 134: Was thinking you could add ism_address here, like `map(|ism_address|) (*ism_address, self.base.build_ism_and_meta(/* ... */)))`

Edit: but this is wrong

tkporter (approved):

PR #2865, opened 2023-10-28
Cosmos / CosmWasm agents

tkporter:


rust/agents/validator/src/validator.rs
```diff
@@ -289,4 +310,71 @@ impl Validator {
...
+    use ethers::{
+        signers::Wallet,
+        utils::{self},
+    };
+    use hyperlane_core::{Announcement, HyperlaneSigner, Signable, H256};
+    use hyperlane_ethereum::Signers;
+    use k256::ecdsa::SigningKey;
+
+    #[tokio::test]
+    async fn sign_manual() -> eyre::Result<()> {
```
tkporter, line 325: Seems like this test case maybe doesn't do anything anymore after we moved away from raw signers / bech32 validators and instead kept with the same ethereum-based announcement signatures?

rust/build
```diff
@@ -0,0 +1,20 @@
+TAG=$1
```
tkporter, line 1: afiact this is just a dupe of `build.sh` and can be removed?

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
```diff
@@ -0,0 +1,81 @@
...
+use hyperlane_core::{
+    AggregationIsm, ChainResult, ContractLocator, HyperlaneChain, HyperlaneContract,
+    HyperlaneDomain, HyperlaneMessage, HyperlaneProvider, RawHyperlaneMessage, H256,
+};
+use tracing::instrument;
+
+/// A reference to an AggregationIsm contract on some Cosmos chain
+#[derive(Debug)]
+pub struct CosmosAggregationIsm {
+    _conf: ConnectionConf,
```
tkporter, line 19: seems we can just remove the unused vars here

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
```diff
@@ -0,0 +1,81 @@
...
+            },
+        };
+
+        let data = self.provider.wasm_query(payload, None).await?;
+        let response: ModulesAndThresholdResponse = serde_json::from_slice(&data)?;
+
+        let modules: Vec<H256> = response
+            .modules
+            .iter()
+            .map(|module| bech32_decode(module.clone()))
```
tkporter, line 76: we can do into_iter and avoid the clone

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[async_trait]
+impl AggregationIsm for CosmosAggregationIsm {
+    #[instrument(err)]
+    async fn modules_and_threshold(
+        &self,
+        message: &HyperlaneMessage,
+    ) -> ChainResult<(Vec<H256>, u8)> {
+        let payload = ModulesAndThresholdRequest {
```
tkporter, line 64: wonder if we should just put a constructor on ModulesAndThresholdRequest that takes in a &HyperlaneMessage

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use crate::grpc::WasmGrpcProvider;
+use crate::rpc::{CosmosWasmIndexer, WasmIndexer};
+use crate::signers::Signer;
+use crate::{ConnectionConf, CosmosProvider};
+
+/// A reference to a InterchainGasPaymaster contract on some Cosmos chain
+#[derive(Debug)]
+pub struct CosmosInterchainGasPaymaster {
+    _conf: ConnectionConf,
```
tkporter, line 21: can rm the unused vars

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+        let indexer: CosmosWasmIndexer = CosmosWasmIndexer::new(conf, locator, event_type.clone());
+
+        Self {
+            indexer: Box::new(indexer),
+        }
+    }
+
+    fn get_parser(&self) -> fn(attrs: Vec<EventAttribute>) -> Option<InterchainGasPayment> {
+        |attrs: Vec<EventAttribute>| -> Option<InterchainGasPayment> {
+            let mut res = InterchainGasPayment {
```
tkporter, line 79: let's refactor all this to get rid of the unwraps and the match statement with the hardcoded base64

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+/// A reference to a InterchainGasPaymasterIndexer contract on some Cosmos chain
+#[derive(Debug)]
+pub struct CosmosInterchainGasPaymasterIndexer {
+    indexer: Box<CosmosWasmIndexer>,
+}
+
+impl CosmosInterchainGasPaymasterIndexer {
+    /// create new Cosmos InterchainGasPaymasterIndexer agent
+    pub fn new(conf: ConnectionConf, locator: ContractLocator, event_type: String) -> Self {
+        let indexer: CosmosWasmIndexer = CosmosWasmIndexer::new(conf, locator, event_type.clone());
```
tkporter, line 70: nit: don't need to event_type.clone()

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    async fn get_finalized_block_number(&self) -> ChainResult<u32> {
+        self.indexer.latest_block_height().await
+    }
+}
+
+#[async_trait]
+impl SequenceIndexer<InterchainGasPayment> for CosmosInterchainGasPaymasterIndexer {
+    async fn sequence_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
+        // TODO: implement when sealevel scraper support is implemented
```
tkporter, line 170: nit. can remove this todo

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+        let tip = self.indexer.latest_block_height().await?;
+        Ok((None, tip))
+    }
+}
+
+#[async_trait]
+impl SequenceIndexer<H256> for CosmosInterchainGasPaymasterIndexer {
```
tkporter, line 177: can rm entirely

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn domain(&self) -> &HyperlaneDomain {
+        &self.domain
+    }
+
+    fn provider(&self) -> Box<dyn HyperlaneProvider> {
+        Box::new(CosmosProvider::new(self.domain.clone()))
+    }
+}
+
+fn ism_type_to_module_type(ism_type: hpl_interface::ism::ISMType) -> ModuleType {
```
tkporter, line 60: can maybe do a `From` or `Into` impl instead

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        // Handle both the ISMType response and the ModuleTypeResponse response.
+        let ismtype_response = serde_json::from_slice::<QueryIsmModuleTypeResponse>(&data);
+        let moduletye_response =
+            serde_json::from_slice::<hpl_interface::ism::ModuleTypeResponse>(&data);
+
+        Ok(match (ismtype_response, moduletye_response) {
+            (Ok(v), _) => ism_type_to_module_type(v.typ),
+            (_, Ok(v)) => ism_type_to_module_type(v.typ),
+            _ => ModuleType::Null,
```
tkporter, line 95: we should probably error in this case?

Also I'm not sure what the difference is between `ismtype_response` and `moduletye_response` is

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// Dry runs the `verify()` ISM call and returns `Some(gas_estimate)` if the call
+    /// succeeds.
+    async fn dry_run_verify(
+        &self,
+        message: &HyperlaneMessage,
+        metadata: &[u8],
+    ) -> ChainResult<Option<U256>> {
+        Ok(Some(U256::from(1000))) // TODO
```
tkporter, line 106: let's clarify in the TODO that this is just needed when aggregation ISM support is added

rust/chains/hyperlane-cosmos/src/libs/binary.rs
```diff
@@ -0,0 +1,28 @@
+use hyperlane_core::{H160, H256, H512};
+use std::cmp;
+
+/// Convert H256 to H512
+pub fn h256_to_h512(v: H256) -> H512 {
```
tkporter, line 5: pretty sure we have utils or Into/From impls for all these already in some way and can replace these

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use bech32::{FromBase32, ToBase32};
+use cosmrs::crypto::secp256k1::SigningKey;
+use hyperlane_core::{ChainCommunicationError, ChainResult, H160, H256};
+use ripemd::Ripemd160;
+use sha2::{Digest, Sha256};
+
+/// decode bech32 address to H256
+pub fn bech32_decode(addr: String) -> H256 {
+    let (_hrp, data, _variant) = bech32::decode(addr.as_str()).unwrap();
```
tkporter, line 11: let's rm the unwraps here and change this to return a result

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+    let mut result: [u8; 32] = [0; 32];
+
+    let start_point = cmp::max(0, 32 - value.len());
+    result[start_point..32].copy_from_slice(value.as_slice());
+
+    H256::from(result)
+}
+
+/// encode H256 to bech32 address
+pub fn digest_to_addr(digest: H256, prefix: &str) -> ChainResult<String> {
```
tkporter, line 23: maybe shouldn't be using ChainResult for all these fns

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+        .finalize()
+        .as_slice()
+        .try_into()
+        .map_err(|_| ChainCommunicationError::ParseError {
+            msg: "ripemd160".to_string(),
+        })
+}
+
+/// encode H256 to bech32 address
+pub fn pub_to_addr(pub_key: Vec<u8>, prefix: &str) -> ChainResult<String> {
```
tkporter, line 67: wonder if there are already functions like this that cosmrs gives us. This file could use some love

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+        })?;
+
+    Ok(addr)
+}
+
+/// encode H256 to bech32 address
+pub fn priv_to_binary_addr(priv_key: Vec<u8>) -> ChainResult<H160> {
+    let sha_hash = sha256_digest(
+        SigningKey::from_slice(priv_key.as_slice())
+            .unwrap()
```
tkporter, line 85: more unwraps to rm here and line 98

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+            .public_key()
+            .to_bytes(),
+    )?;
+    let rip_hash = ripemd160_digest(sha_hash)?;
+
+    Ok(H160::from_slice(rip_hash.as_slice()))
+}
+
+/// encode H256 to bech32 address
+pub fn priv_to_addr_string(prefix: String, priv_key: Vec<u8>) -> ChainResult<String> {
```
tkporter, line 95: this could reuse `pub_to_addr`. Let's just look at this file deeply

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+    utils::fmt_bytes, ChainResult, HyperlaneChain, HyperlaneContract, HyperlaneDomain,
+    HyperlaneMessage, HyperlaneProvider, Indexer, LogMeta, Mailbox, TxCostEstimate, TxOutcome,
+    H256, U256,
+};
+use hyperlane_core::{ContractLocator, Decode, RawHyperlaneMessage, SequenceIndexer};
+use tracing::{instrument, warn};
+
+/// A reference to a Mailbox contract on some Cosmos chain
+pub struct CosmosMailbox {
+    _conf: ConnectionConf,
```
tkporter, line 32: can rm

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    fn provider(&self) -> Box<dyn HyperlaneProvider> {
+        Box::new(CosmosProvider::new(self.domain.clone()))
+    }
+}
+
+impl Debug for CosmosMailbox {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        // Debug::fmt(&(self as &dyn HyperlaneContract), f)
```
tkporter, line 74: should do this and remove the todo

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+        let response: mailbox::DefaultIsmResponse = serde_json::from_slice(&data)?;
+
+        // convert Hex to H256
+        let ism = H256::from_slice(&hex::decode(response.default_ism)?);
+        Ok(ism)
+    }
+
+    #[instrument(err, ret, skip(self))]
+    async fn recipient_ism(&self, recipient: H256) -> ChainResult<H256> {
+        let address = verify::digest_to_addr(recipient, &self.signer.prefix)?;
```
tkporter, line 150: we should get this from another source, i think it may be in the connection conf

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// chain
+    pub fn new(conf: ConnectionConf, locator: ContractLocator, signer: Signer) -> Self {
+        let provider: WasmGrpcProvider =
+            WasmGrpcProvider::new(conf.clone(), locator.clone(), signer.clone());
+
+        Self {
+            _conf: conf,
+            domain: locator.domain.clone(),
+            address: locator.address,
+            signer,
```
tkporter, line 50: we should not make signer required

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+            nonce: general::EmptyStruct {},
+        };
+
+        let data = self
+            .provider
+            .wasm_query(GeneralMailboxQuery { mailbox: payload }, lag)
+            .await;
+
+        if let Err(e) = data {
+            warn!("error: {:?}", e);
```
tkporter, line 93: move this to a `?` above and remove the warn!, we instrument anyways so will see a log

rust/chains/hyperlane-cosmos/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+base64 = { workspace = true }
+serde_json = { workspace = true }
+tonic = { workspace = true }
+sha2 = { workspace = true }
+ripemd = { workspace = true }
+hyper = { workspace = true }
+hyper-tls = { workspace = true }
+sha256 = { workspace = true }
+hex = { workspace = true }
+hpl-interface = { version = "0.0.2" }
```
daniel-savu, line 29: why not move this to workspace level?

rust/chains/hyperlane-cosmos/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+hyperlane-core = { path = "../../hyperlane-core" }
+
+# These should only be used if it _must_ be used to interop with the inner library,
+# all errors exported from a chain crate should be using thiserror or handrolled to
+# make error handling easier.
+# eyre = "never"
+# anyhow = never
```
daniel-savu, line 37: looks removable

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
```diff
@@ -0,0 +1,81 @@
...
+    domain: HyperlaneDomain,
+    address: H256,
+    _signer: Signer,
+    provider: Box<WasmGrpcProvider>,
+}
+
+impl CosmosAggregationIsm {
+    /// create new Cosmos AggregationIsm agent
+    pub fn new(conf: ConnectionConf, locator: ContractLocator, signer: Signer) -> Self {
+        let provider = WasmGrpcProvider::new(conf.clone(), locator.clone(), signer.clone());
```
daniel-savu, line 29: @tkporter mentioned this and I double checked, I think the only rpc used from the config is the "grpcUrl", with an unecesarry "rpcUrl" still specified because that's not optional. Don't think it's a good idea to overload the "rpcUrl" with a grpc entry, but maybe the former could be made optional?

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
daniel-savu: which also allows for rewriting this line as `.map(bech32_decode)`

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// create new Cosmos InterchainGasPaymasterIndexer agent
+    pub fn new(conf: ConnectionConf, locator: ContractLocator, event_type: String) -> Self {
+        let indexer: CosmosWasmIndexer = CosmosWasmIndexer::new(conf, locator, event_type.clone());
+
+        Self {
+            indexer: Box::new(indexer),
+        }
+    }
+
+    fn get_parser(&self) -> fn(attrs: Vec<EventAttribute>) -> Option<InterchainGasPayment> {
```
daniel-savu, line 77: I'm not the biggest fan of returning closures in these `get_parser` fns. The name is also a bit misleading since it suggests you're getting a struct / object

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -0,0 +1,183 @@
...
+        |attrs: Vec<EventAttribute>| -> Option<InterchainGasPayment> {
+            let mut res = InterchainGasPayment {
+                message_id: H256::zero(),
+                payment: U256::zero(),
+                gas_amount: U256::zero(),
+                destination: 0,
+            };
+
+            for attr in attrs {
```
daniel-savu, line 86: first thing that stands out is that we could `fold` and iterator using a `Default` for `InterchainGasPayment`, along with a builder pattern for setting individual fields in `InterchainGasPayment`.

Then I think faulty decodings could be filtered out as part of the iteration, instead of unwrapping

daniel-savu (commented):
had a relatively brief look, will continue in the morning

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -0,0 +1,108 @@
...
+        // Handle both the ISMType response and the ModuleTypeResponse response.
+        let ismtype_response = serde_json::from_slice::<QueryIsmModuleTypeResponse>(&data);
+        let moduletye_response =
+            serde_json::from_slice::<hpl_interface::ism::ModuleTypeResponse>(&data);
+
+        Ok(match (ismtype_response, moduletye_response) {
+            (Ok(v), _) => ism_type_to_module_type(v.typ),
+            (_, Ok(v)) => ism_type_to_module_type(v.typ),
+            _ => ModuleType::Null,
+        })
```
daniel-savu, line 96: I would first assign this to a variable, say `module_type`, and then return `Ok(module_type)`. Also, although the `match` here is more concise, to me this logic is easier to understand when written as
```suggestion
                let module_type = if let Ok(ism_type) = ismtype_response {
            ism_type_to_module_type(ism_type.typ)
        } else if let Ok(module_type) = moduletye_response {
            ism_type_to_module_type(module_type.typ)
        } else {
            ModuleType::Null
        };
        Ok(module_type)
```

rust/chains/hyperlane-cosmos/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+mod validator_announce;
+
+pub use self::{
+    aggregation_ism::*, interchain_gas::*, interchain_security_module::*, libs::*, mailbox::*,
+    merkle_tree_hook::*, multisig_ism::*, providers::*, routing_ism::*, signers::*,
+    trait_builder::*, trait_builder::*, validator_announce::*, validator_announce::*,
+};
+
+/// Safe default imports of commonly used traits/types.
+pub mod prelude {}
```
daniel-savu, line 29: do we plan on including anything in the prelude or shall we remove this?

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+use hyperlane_core::{ChainCommunicationError, ChainResult, H160, H256};
+use ripemd::Ripemd160;
+use sha2::{Digest, Sha256};
+
+/// decode bech32 address to H256
+pub fn bech32_decode(addr: String) -> H256 {
+    let (_hrp, data, _variant) = bech32::decode(addr.as_str()).unwrap();
+
+    let value = Vec::<u8>::from_base32(&data).unwrap();
+    let mut result: [u8; 32] = [0; 32];
```
daniel-savu, line 14: would be nice to remove these magic numbers here and below. Can you use a constant imported from one of the crates at the top of the file for the `32`, and also just assign `Default::default()` here?

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let start_point = cmp::max(0, 32 - value.len());
+    result[start_point..32].copy_from_slice(value.as_slice());
```
daniel-savu, line 17: since there's only 32 elements, this can technically be
```suggestion
    result[start_point..].copy_from_slice(value.as_slice());
```

daniel-savu: Since the docs of `copy_from_slice ` say that "This function will panic if the two slices have different lengths.", but that's guaranteed not to happen in this case, I would add a comment to inform future modifications of this fn about this context

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/// encode H256 to bech32 address
+pub fn digest_to_addr(digest: H256, prefix: &str) -> ChainResult<String> {
+    let addr = bech32::encode(
+        prefix,
+        digest.as_bytes().to_base32(),
+        bech32::Variant::Bech32,
+    )
+    .map_err(|_| ChainCommunicationError::InvalidRequest {
```
daniel-savu, line 29: technically this isn't a chain communcation error though

rust/chains/hyperlane-cosmos/src/libs/verify.rs
daniel-savu: agreed, this adds quite a bit of dense logic to maintain

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+        // Debug::fmt(&(self as &dyn HyperlaneContract), f)
+        todo!()
+    }
+}
+
+#[async_trait]
+impl Mailbox for CosmosMailbox {
+    #[instrument(level = "debug", err, ret, skip(self))]
+    async fn count(&self, lag: Option<NonZeroU64>) -> ChainResult<u32> {
+        let payload = mailbox::NonceRequest {
```
daniel-savu, line 83: any chance to use `Default::default()` here?

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+            },
+        };
+
+        let data = self
+            .provider
+            .wasm_query(GeneralMailboxQuery { mailbox: payload }, None)
+            .await?;
+        let response: mailbox::RecipientIsmResponse = serde_json::from_slice(&data)?;
+
+        // convert Hex to H256
```
daniel-savu, line 164: I think this comment is stale

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+        let process_message = ProcessMessageRequest {
+            process: ProcessMessageRequestInner {
+                message: hex::encode(RawHyperlaneMessage::from(message)),
+                metadata: hex::encode(metadata),
+            },
+        };
+
+        let response: SimulateResponse = self.provider.wasm_simulate(process_message).await?;
+        let result = TxCostEstimate {
+            gas_limit: U256::from(response.gas_info.unwrap().gas_used),
```
daniel-savu, line 212: shouldn't unwrap here, thinking we should have a commit that removes all unwraps in the cosmos crates to avoid missing any of them

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -0,0 +1,378 @@
...
+            gas_price: U256::from(2500),
```
daniel-savu, line 213: I'd rather have the `2500` as a constant

daniel-savu (commented):
some more suggestions, hoping to finish reviewing tomorrow

rust/hyperlane-base/src/settings/signers.rs
```diff
@@ -104,7 +133,34 @@ impl BuildableWithSignerConf for Keypair {
...
+impl BuildableWithSignerConf for hyperlane_cosmos::Signer {
+    async fn build(conf: &SignerConf) -> Result<Self, Report> {
+        Ok(match conf {
+            SignerConf::HexKey { .. } => bail!("HexKey signer is not supported by cosmos"),
+            SignerConf::Aws { .. } => bail!("Aws signer is not supported by cosmos"),
+            SignerConf::CosmosKey { key, prefix } => {
+                hyperlane_cosmos::Signer::new(key.as_bytes().to_vec(), prefix.clone())
+            }
+            SignerConf::Node => bail!("Node signer is not supported by cosmos"),
+        })
```
daniel-savu, line 158: could be simplified to 
```suggestion
    async fn build(conf: &SignerConf) -> Result<Self, Report> {
        if let SignerConf::CosmosKey { key, prefix } = conf {
            return Ok(hyperlane_cosmos::Signer::new(
                key.as_bytes().to_vec(),
                prefix.clone(),
            ));
        } else {
            bail!(format!("{conf:?} key is not supported by cosmos"));
        }
```

rust/chains/hyperlane-cosmos/src/payloads/general.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EmptyStruct {}
```
daniel-savu, line 4: I assume serde treats empty struct and unit structs identically, could we use a unit struct here for conciseness?

daniel-savu (commented):
another round of comments

rust/hyperlane-base/src/settings/signers.rs
```diff
@@ -28,7 +28,14 @@ pub enum SignerConf {
         /// The AWS region
         region: Region,
     },
-    /// Assume the local node will sign on RPC calls automatically
+    /// Cosmos Specific key
+    CosmosKey {
+        /// Private key value
+        key: H256,
+        /// Prefix for cosmos address
+        prefix: String,
```
daniel-savu, line 36: can this be an enum or string literals?

rust/hyperlane-core/src/chain.rs
```diff
@@ -367,7 +375,7 @@ impl HyperlaneDomain {
         use HyperlaneDomainProtocol::*;
         let protocol = self.domain_protocol();
         many_to_one!(match protocol {
-            IndexMode::Block: [Ethereum],
+            IndexMode::Block: [Ethereum, Cosmos], // TODO: Is cosmos index-mode is correct?
```
daniel-savu, line 378: I believe the answer is yes and we should remove this comment @tkporter

rust/hyperlane-core/src/error.rs
```diff
@@ -4,6 +4,12 @@ use std::fmt::{Debug, Display, Formatter};
 use std::ops::Deref;
 
 use crate::config::StrOrIntParseError;
+use cosmrs::proto::prost;
+use cosmrs::Error as CosmrsError;
+// use ethers_contract::ContractError;
+// use ethers_core::types::SignatureError;
+// use ethers_providers::{Middleware, ProviderError};
```
daniel-savu, line 11: not needed

rust/hyperlane-core/src/traits/signing.rs
```diff
@@ -110,9 +119,37 @@ impl<T: Signable> SignedType<T> {
     pub fn recover(&self) -> Result<H160, crate::HyperlaneProtocolError> {
         let hash = ethers_core::types::H256::from(self.value.eth_signed_message_hash());
         let sig = ethers_core::types::Signature::from(self.signature);
+
         Ok(sig.recover(hash)?.into())
     }
 
+    /// Recover the public key of the signer
+    #[cfg(feature = "ethers")]
+    pub fn recover_pubkey(&self) -> Result<Vec<u8>, crate::HyperlaneProtocolError> {
```
daniel-savu, line 128: doesn't look like this is used anywhere

rust/neutron_relayer_config.json
daniel-savu, line 1: this file should be removed, was only meant for local use

rust/neutron_validator_config.json
daniel-savu, line 1: let's remove

daniel-savu:
During testing I noticed that the relayer fetches the entire events list on each block, we should look into this since it's quite inefficient

rust/chains/hyperlane-cosmos/src/validator_announce.rs
```diff
@@ -0,0 +1,139 @@
...
+                validator: hex::encode(announcement.value.validator),
+                storage_location: announcement.value.storage_location,
+                signature: hex::encode(announcement.signature.to_vec()),
+            },
+        };
+
+        // Perform gas estimation if a limit wasn't already provided.
+        // TODO: refactor `wasm_send` to do this gas estimation automatically
+        // instead of using a default hardcoded value if the gas limit is `None`.
+        let tx_gas_limit = if let Some(gas_limit) = tx_gas_limit {
```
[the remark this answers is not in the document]
tkporter: unsure, I think the types imply you need to do gas estimation yourself. at the least we should put this in the provider

rust/hyperlane-core/src/error.rs
```diff
@@ -4,6 +4,12 @@ use std::fmt::{Debug, Display, Formatter};
 use std::ops::Deref;
 
 use crate::config::StrOrIntParseError;
+use cosmrs::proto::prost;
+use cosmrs::Error as CosmrsError;
+// use ethers_contract::ContractError;
```
tkporter, line 9: rm

rust/hyperlane-core/src/error.rs
```diff
@@ -78,6 +84,51 @@ pub enum ChainCommunicationError {
     /// Failed to parse strings or integers
     #[error("Data parsing error {0:?}")]
     StrOrIntParseError(#[from] StrOrIntParseError),
+    /// Tendermint RPC Error
```
tkporter, line 87: would prefer to not put execution env specific things in here

rust/hyperlane-core/src/traits/signing.rs
tkporter, line 128: do we need this

rust/neutron_relayer_config.json
```diff
@@ -0,0 +1,30 @@
+{
```
tkporter, line 1: rm

rust/neutron_validator_config.json
```diff
@@ -0,0 +1,12 @@
+{
```
tkporter, line 1: rm

rust/utils/run-locally/src/cosmos/link.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub set: EnrollValidatorSet,
+}
+
+#[cw_serde]
+pub struct EnrollValidatorSet {
+    pub domain: u32,
+    pub validator: String,
+}
+
+fn link_network(
```
daniel-savu, line 86: my understanding is that this function set up and registers a validator, so maybe a better name would be `set_up_validator()`?

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+use crate::AGENT_BIN_PATH;
+use cli::{OsmosisCLI, OsmosisEndpoint};
+
+use self::deploy::deploy_cw_hyperlane;
+use self::source::{CLISource, CodeSource};
+
+// const OSMOSIS_CLI_GIT: &str = "https://github.com/osmosis-labs/osmosis";
+// const OSMOSIS_CLI_VERSION: &str = "19.0.0";
+
+const OSMOSIS_CLI_GIT: &str = "https://github.com/hashableric/osmosis";
```
daniel-savu, line 38: any reason we're not using the official repo?

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+    [
+        KEY_HPL_VALIDATOR,
+        KEY_HPL_RELAYER,
+        KEY_VALIDATOR,
+        KEY_ACCOUNTS1,
+        KEY_ACCOUNTS2,
+        KEY_ACCOUNTS3,
+    ]
+}
+const CW_HYPERLANE_GIT: &str = "https://github.com/many-things/cw-hyperlane";
```
daniel-savu, line 59: we should host this under the hyperlane github organisation

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+#[cw_serde]
+pub struct MockDispatchInner {
+    pub dest_domain: u32,
+    pub recipient_addr: String,
+    pub msg_body: String,
+    pub hook: Option<String>,
+    pub metadata: String,
+}
+
+pub fn install_codes(dir: Option<PathBuf>, local: bool) -> BTreeMap<String, PathBuf> {
```
daniel-savu, line 95: rename to `install_cw_hyperlane_release`?

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+}
+
+pub struct CosmosHyperlaneStack {
+    pub validators: Vec<AgentHandles>,
+    pub relayer: AgentHandles,
+}
+
+impl Drop for CosmosHyperlaneStack {
+    fn drop(&mut self) {
```
daniel-savu, line 205: nice

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+            );
+        }
+    }
+
+    let _stack = CosmosHyperlaneStack {
+        validators: hpl_val.into_iter().map(|v| v.join()).collect(),
+        relayer: hpl_rly.join(),
+    };
+
+    sleep(Duration::from_secs(1000)); // wait for a long time
```
daniel-savu, line 498: unclear to me what this sleep does

rust/utils/run-locally/src/cosmos/source.rs
```diff
@@ -0,0 +1,153 @@
...
+        Self::Remote {
+            url: url.to_string(),
+            version: version.to_string(),
+        }
+    }
+}
+
+impl CodeSource {
+    fn install_local(src: String) -> BTreeMap<String, PathBuf> {
+        // make contract_name => path map
```
daniel-savu, line 44: this comment is present repeat three times across the project, and the code it describes looks identical each time. Let's try reusing this function everywhere instead of duplicating

rust/utils/run-locally/src/cosmos/source.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn remote(url: &str, version: &str) -> Self {
+        Self::Remote {
+            url: url.to_string(),
+            version: version.to_string(),
+        }
+    }
+}
+
+impl CLISource {
+    fn install_remote(dir: Option<PathBuf>, git: String, version: String) -> PathBuf {
```
daniel-savu, line 125: rename to `install_cli`?

solidity/contracts/Mailbox.sol
daniel-savu, line 1: this and all other `.sol` files are unrelated to this PR and should be removed

rust/utils/run-locally/src/cosmos/crypto.rs
```diff
@@ -0,0 +1,43 @@
+use k256::ecdsa::{SigningKey, VerifyingKey};
```
tkporter, line 1: maybe unnencessary here

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
daniel-savu: slack: ismtype_response is legacy

rust/chains/hyperlane-cosmos/src/libs/verify.rs
daniel-savu: shockingly there is no rust implementation for this, but at a glance it matches the one in go: https://github.com/cosmos/cosmos-sdk/blob/177e7f45959215b0b4e85babb7c8264eaceae052/crypto/keys/secp256k1/secp256k1.go#L155

should we add a ticket to contribute this to cosm-rs so it has more eyes on it?

daniel-savu: managed to remove all of the custom logic in here, the hashing is happening in a tendermint crate that's not under the cosmos github org so had missed it initially

rust/chains/hyperlane-cosmos/src/mailbox.rs
daniel-savu: you mean get the prefix from another source? I removed the connection conf because it wasn't used, but why not use the signer for this?

rust/chains/hyperlane-cosmos/src/mailbox.rs
daniel-savu: only in the mailbox, or everywhere?

rust/Cargo.toml
```diff
@@ -145,6 +164,21 @@ url = "2.3"
...
+ripemd = "0.1.3"
+hyper = "0.14"
+hyper-tls = "0.5.0"
+hex = "0.4.3"
+tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"]}
+hpl-interface = "0.0.6-rc5"
+elliptic-curve = "0.12.3"
+
+## TODO: remove this
+cosmwasm-schema = "1.2.7"
```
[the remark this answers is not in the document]
daniel-savu: can't be removed, it's currently in use

rust/chains/hyperlane-cosmos/src/payloads/merkle_tree_hook.rs
```diff
@@ -0,0 +1,46 @@
...
+use super::general::EmptyStruct;
+
+const TREE_DEPTH: usize = 32;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct MerkleTreeGenericRequest<T> {
+    pub merkle_hook: T,
+}
+
+// --------- Requests ---------
```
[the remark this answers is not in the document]
daniel-savu: actually all of these should be importable from `hpl_interface`, looking into it

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -0,0 +1,332 @@
...
+    ChainCommunicationError, ChainResult, ContractLocator, HyperlaneDomain, H256, U256,
+};
+use serde::Serialize;
+use std::num::NonZeroU64;
+
+use crate::verify;
+use crate::{signers::Signer, ConnectionConf};
+
+const DEFAULT_GAS_PRICE: f32 = 0.05;
+const DEFAULT_GAS_ADJUSTMENT: f32 = 1.25;
```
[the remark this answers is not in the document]
daniel-savu: no :(

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub fn get_contract_addr(&self) -> ChainResult<String> {
+        verify::digest_to_addr(self.address, self.conf.get_prefix().as_str())
+    }
+}
+
+#[async_trait]
+impl WasmIndexer for CosmosWasmIndexer {
+    fn get_client(&self) -> ChainResult<HttpClient> {
+        Ok(HttpClient::builder(self.get_conn_url()?.parse()?)
+            // indexing fails unless this is commented out. I assume the decoding in `CompatMode::V0_34`
```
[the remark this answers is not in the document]
daniel-savu: we did find this out the hard way I guess :P

rust/chains/hyperlane-cosmos/src/mailbox.rs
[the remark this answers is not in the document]
tkporter: tracking this here https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2942

daniel-savu (approved):

PR #2868, opened 2023-10-29
Check for IBC denom collateral and cosmos addresess

nambrot:

### DIFF
--- a/typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
+++ b/typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
@@ -319,6 +319,13 @@ export class CwHypNativeAdapter
     );
   }
 
+  async getBalance(address: string): Promise<string> {
+    const provider = await this.getProvider();
+    const denom = await this.denom();
+    const balance = await provider.getBalance(address, denom);
+    return balance.amount;
+  }
+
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
```diff
@@ -398,6 +405,5 @@ export class CwHypCollateralAdapter
     public readonly gasDenom = 'token',
   ) {
     super(chainName, multiProvider, addresses, gasDenom);
-    throw new Error('Not yet implemented');
```
jmrossy, line 401: I added this constructor to throw this error. If `CwHypCollateralAdapter` works the same as `CwHypNativeAdapter` we can just delete this constructor def entirely

typescript/utils/src/addresses.ts
```diff
@@ -77,7 +78,8 @@ export function isValidAddressSealevel(address: Address) {
 // Slower than isAddressCosmos above but actually validates content and checksum
 export function isValidAddressCosmos(address: Address) {
   try {
-    const isValid = address && fromBech32(address);
+    const isValid =
+      IBC_DENOM_REGEX.test(address) || (address && fromBech32(address));
```
jmrossy, line 82: nit `address && ` should go first

PR #2859, opened 2023-10-26
Add Cosmos support to Utils and SDK

yorhodes:
### Description

- Adds Cosmos support to the SDK
- Implements CosmWasm and Cosmos adapters for use in the Nexus UI

### Backward compatibility

Yes

### Testing

- Unit Tests for chainmetadata and provider changes
- Manual for contract interactions

PR #2869, opened 2023-10-29
Infra for neutron & neutrontestnet

nambrot:


typescript/infra/src/agents/gcp.ts
```diff
@@ -101,6 +106,17 @@ export class AgentGCPKey extends CloudAgentKey {
         return Keypair.fromSeed(
           Buffer.from(strip0x(this.privateKey), 'hex'),
         ).publicKey.toBase58();
+      case ProtocolType.CosmWasm:
```
jmrossy, line 109: This doesn't match the Protocol Type Yorke and I added on our branch. We called it `Cosmos`, not `CosmWasm`

jmrossy: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2859/files

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -943,6 +943,67 @@ export const polygonzkevm: ChainMetadata = {
   },
 };
 
+export const neutrontestnet: ChainMetadata = {
+  protocol: ProtocolType.CosmWasm,
+  domainId: 33333,
+  chainId: 33333,
+  name: Chains.neutrontestnet,
+  displayName: 'Neutron Testnet',
+  displayNameShort: 'Neutron Testnet',
```
jmrossy, line 952: This field is optional, no need to include it if it will be the same as displayName

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -943,6 +943,67 @@ export const polygonzkevm: ChainMetadata = {
   },
 };
 
+export const neutrontestnet: ChainMetadata = {
+  protocol: ProtocolType.CosmWasm,
```
jmrossy, line 947: See above about ProtocolType

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -943,6 +943,67 @@ export const polygonzkevm: ChainMetadata = {
...
+    name: 'Neutron',
+    symbol: 'NTRN',
+    decimals: 6,
+  },
+  blocks: {
+    confirmations: 1,
+    reorgPeriod: 0,
+    estimateBlockTime: 3,
+  },
+  rpcUrls: [{ http: 'http://54.149.31.83:26657' }],
```
jmrossy, line 963: To get the UI's keplr integration to work, I needed **both** an RPC url and a REST url (cosmos has 3 types, RPC, gRPC, and REST). We'll need a rest api here too, which I believe is `http://54.149.31.83:1317`. For simplicity, the warp UI assumes the 1st is RPC and the 2nd is REST for cosmos chains

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -943,6 +943,67 @@ export const polygonzkevm: ChainMetadata = {
...
+  isTestnet: true,
+  prefix: 'dual',
```
jmrossy, line 965: again, this doesn't match the names used on our other branch. it's called `bech32prefix`. And we need a `slip44` field too

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -943,6 +943,67 @@ export const polygonzkevm: ChainMetadata = {
...
+};
+
+export const neutron: ChainMetadata = {
```
jmrossy, line 968: ditto

typescript/infra/src/agents/gcp.ts
tkporter: I'll rename - fwiw this is just the enum variant name and the value is "cosmos" for consistency with what the agents want. But regardless I feel like calling cosmwasm "cosmos" is weird because of our work on cosmos sdk, which I guess we can just call "cosmossdk" when we ship that

jmrossy: Hmm sorry I don't quite follow. Do you prefer `Cosmos = 'cosmos'`, `CosmWasm = 'cosmos', or `CosmWasm = 'cosmwasm'` here? I prefer the 1st one because the there are utils that will rely on this protocol type which are not specific to cosmwasm

tkporter: CosmWasm = 'cosmwasm' would've been my preference but I'll just stick with Cosmos = 'cosmos'

typescript/sdk/src/consts/chainMetadata.ts
tkporter: sounds good, I'll rename prefix but will just wait for your other changes for the slip 44 part

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -1005,6 +1005,31 @@ export const polygonzkevm: ChainMetadata = {
   },
 };
 
+export const neutrontestnet: ChainMetadata = {
```
nambrot, line 1008: Should this either also include neutron, or not be in here?

typescript/sdk/src/metadata/agentConfig.ts
```diff
@@ -63,6 +64,13 @@ const AgentSignerAwsKeySchema = z
   .describe(
     'An AWS signer. Note that AWS credentials must be inserted into the env separately.',
   );
+const AgentSignerCosmosKeySchema = z
```
nambrot, line 67: I think @jmrossy has some kind of logic that we can enforce to only show up/require when the protocol type is cosmos

tkporter: added a `refine`

tkporter (approved):

PR #2870, opened 2023-10-30
Add announcement key in config

nambrot:
Adds a key in there that if it has funds, can be used the announce. Otherwise the operator has to specify their own key or somebody else has to announce


tkporter (approved):
just wanna make sure that both these aren't funded yet? afaict the neutron address is `neutron1kknekjxg0ear00dky5ykzs8wwp2gz62z9s6aaj` and the Eth address is `0x8b1eEb59f7aBa48ed2A3E892a5F432584b3b153b`

PR #2872, opened 2023-10-30
Add monitoring script for neutron

aroralanuk:
### Description

- Adding monitoring script for neutron and mantapacific

### Drive-by changes

- filter wagmi chainMetadata

### Backward compatibility

Yes

### Testing

Manual


typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -685,6 +685,35 @@ export const proteustestnet: ChainMetadata = {
...
+  },
+  blocks: {
+    confirmations: 1,
+    reorgPeriod: 0,
+    estimateBlockTime: 3,
+  },
+  blockExplorers: [
+    {
+      name: 'Manta Pacific Explorer',
+      url: 'https://pacific-explorer.manta.network/',
```
jmrossy, line 708: I recommend removing the trailing `/` here in the url. 
The utilities to assemble tx/address urls using this aren't as robust as they should be

PR #2879, opened 2023-10-31
Fix lag issue with Mailbox message sequence_and_tip

tkporter:
### Description

The old implementation of the `sequence_and_tip` fn in mailbox.rs used to look like this, where it'd end up passing the tip as the lag!
```
let sequence = match NonZeroU64::new(tip as u64) {
            None => None,
            Some(n) => Some(self.nonce(Some(n)).await?),
        };
```

Shuffled things around where we correctly just specify the `tip` as the block number instead of a lag. This means that all functions that were calling `wasm_query` with the lag needed to start getting the lagged block num instead

Sometimes I believe this would create a bug where we'd ask for a block height 1 and get back RPC errors. I believe most of the time we'd ask for a block height of 0, in which case it'd just act like we were asking for the tip

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

n/a

### Testing

plan to roll it out once I get an image


rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -133,41 +132,19 @@ impl WasmProvider for WasmGrpcProvider {
...
-    async fn wasm_query<T>(&self, payload: T, maybe_lag: Option<NonZeroU64>) -> ChainResult<Vec<u8>>
+    async fn wasm_query<T>(&self, payload: T, block_height: Option<u64>) -> ChainResult<Vec<u8>>
     where
         T: Serialize + Send + Sync,
     {
-        let mut client = WasmQueryClient::connect(self.get_conn_url()?).await?;
-
-        let mut request = tonic::Request::new(QuerySmartContractStateRequest {
-            address: self.get_contract_addr()?,
-            query_data: serde_json::to_string(&payload)?.as_bytes().to_vec(),
```
daniel-savu, line 144: good spot

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -280,37 +282,6 @@ impl CosmosMailboxIndexer {
             None
         }
     }
-
-    #[instrument(level = "debug", err, ret, skip(self))]
-    async fn count(&self, lag: Option<NonZeroU64>) -> ChainResult<u32> {
```
daniel-savu, line 285: were these not used at all?

daniel-savu (approved):

rust/chains/hyperlane-cosmos/src/mailbox.rs
tkporter: no 😛

PR #2880, opened 2023-11-01
fix: progress on removing unwraps

daniel-savu:


rust/hyperlane-core/src/types/primitive_types.rs
```diff
@@ -321,6 +321,14 @@ impl_fixed_hash_serde!(H160, 20);
 impl_fixed_hash_serde!(H256, 32);
 impl_fixed_hash_serde!(H512, 64);
 
+// impl TryFrom<String> for H256 {
```
tkporter, line 324: rm or keep?

tkporter (approved):

PR #2881, opened 2023-11-01
refactor: finish removing unwraps

daniel-savu:
### Description

- Removes unwraps and expects from everywhere but `run-locally`
- Precomputes the `address` in `Signer` at construction time, to propagate the error early and keep signatures ergonomic by not requiring a `Result`. Couldn't also precompile `SigningKey` (the privkey type) because it's not `Sync` 😢
- Defines a `hyperlane-cosmos`-specific error enum (`HyperlaneCosmosError`), which can be converted to `ChainCommunicationError` with the `?` operator. This is a pattern we'd like to refactor towards in the future, to remove dependencies from `hyperlane-core` as much as possible
  - One inconvenience is that you need to `.map_err()` to `HyperlaneCosmosError` first, to use `?`. I wish `?` had deref coercion semantics where it'd keep covnerting until an error matches, but I assume this isn't possible because while you can only have a single `Deref` impl, you can have multiple `From<Err>` impls. 
  - Writing this I'm realising we could write a small macro to implement `From<Err> for ChainCommunicationError` for all the sub-errors of `HyperlaneCosmosError` et al (cc @tkporter)

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/error.rs
+++ b/rust/chains/hyperlane-cosmos/src/error.rs
@@ -1,0 +1,22 @@
+// use bech32::Error;
+use hyperlane_core::ChainCommunicationError;
+
+/// Errors from the crates specific to the hyperlane-cosmos
+/// implementation.
+/// This error can then be converted into the broader error type
+/// in hyperlane-core using the `From` trait impl
+#[derive(Debug, thiserror::Error)]
+pub enum HyperlaneCosmosError {
+    /// bech32 error
+    #[error("{0}")]
+    Bech32(#[from] bech32::Error),
+    /// gRPC error
+    #[error("{0}")]
+    GrpcError(#[from] tonic::Status),
+}
+
+impl From<HyperlaneCosmosError> for ChainCommunicationError {
+    fn from(value: HyperlaneCosmosError) -> Self {
+        ChainCommunicationError::from_other(value)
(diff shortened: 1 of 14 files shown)


rust/chains/hyperlane-cosmos/src/error.rs
```diff
@@ -0,0 +1,22 @@
+// use bech32::Error;
+use hyperlane_core::ChainCommunicationError;
+
+/// Errors from the crates specific to the hyperlane-cosmos
```
tkporter, line 4: nice

rust/chains/hyperlane-cosmos/src/error.rs
```diff
@@ -0,0 +1,22 @@
+// use bech32::Error;
```
tkporter, line 1: nit: rm

rust/hyperlane-core/src/error.rs
```diff
@@ -6,6 +6,7 @@ use std::ops::Deref;
 use crate::config::StrOrIntParseError;
 use cosmrs::proto::prost;
 use cosmrs::Error as CosmrsError;
+// use fixed_hash::rustc_hex::FromHexError;
```
tkporter, line 9: nit: rm

tkporter (approved):
🔥

PR #2883, opened 2023-11-02
deploy warp routes

yorhodes:
- Add token deployer tests
- Add warp route deploy to infra
- Remove stale factory factories
- Add fallback ISM deployer tooling
- Add neutron ISM config


typescript/sdk/src/consts/chains.ts
```diff
@@ -24,9 +24,9 @@ export enum Chains {
   mantapacific = 'mantapacific',
   moonbasealpha = 'moonbasealpha',
   moonbeam = 'moonbeam',
+  neutron = 'neutron',
```
jmrossy, line 27: These should be alphabetical, why move it up?

jmrossy (approved):

typescript/sdk/src/consts/chains.ts
yorhodes: mistake

PR #2884, opened 2023-11-02
Ignore constructor and immutable OZ plugin errors

0xmichalis:
The OpenZeppelin upgrade plugin will fail to deploy the Router contract because of the following errors:

```
@hyperlane-xyz/core/contracts/client/Router.sol:24: Contract `Router` has a constructor
    Define an initializer instead
    https://zpl.in/upgrades/error-001

@hyperlane-xyz/core/contracts/client/MailboxClient.sol:53: Contract `MailboxClient` has a constructor
    Define an initializer instead
    https://zpl.in/upgrades/error-001

@hyperlane-xyz/core/contracts/client/MailboxClient.sol:17: Variable `mailbox` is immutable
    Use a constant or mutable variable instead
    https://zpl.in/upgrades/error-005

@hyperlane-xyz/core/contracts/client/MailboxClient.sol:19: Variable `localDomain` is immutable
    Use a constant or mutable variable instead
    https://zpl.in/upgrades/error-005
```

All of these errors can be ignored as the constructors are used to set the immutables and the immutables can be read by the proxy w/o an issue.

### DIFF
--- a/solidity/contracts/client/MailboxClient.sol
+++ b/solidity/contracts/client/MailboxClient.sol
@@ -14,8 +14,10 @@ import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Own
 abstract contract MailboxClient is OwnableUpgradeable {
     using Message for bytes;
 
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
     IMailbox public immutable mailbox;
 
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
     uint32 public immutable localDomain;
 
     IPostDispatchHook public hook;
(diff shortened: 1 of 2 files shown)


yorhodes (commented):
I would prefer to avoid polluting the source with these
can you simply change the configuration of the upgrade plugin with `unsafeAllow: ['state-variable-immutable']` https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables?

0xmichalis:
> I would prefer to avoid polluting the source with these can you simply change the configuration of the upgrade plugin with `unsafeAllow: ['state-variable-immutable']` https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#why-cant-i-use-immutable-variables?

This would disable the protection across all contracts that are part of the proxy deployment which is not ideal. It's also easier to get it fixed in the library vs every OZ plugin user having to configure the plugin to workaround.

PR #2885, opened 2023-11-02
Correctly consider reorg period when indexing, set reorg period to 1 on Neutron

tkporter:
### Description

Hopeful quick fix to indexing issues we're seeing. Leading theory is that the tx_search is asking for a block range that the servicing RPC node doesn't actually have state for.

Curious if @nambrot you feel we should go with a more conservative reorg period, maybe 2 blocks?

My personal theory is that this isn't just bad RPC load balancing, but that the tip that's returned is simply committed to but the state transition hasn't yet been executed, see https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#sdk-context

> First, the Tendermint engine will seek 2/3+ consensus on a list of transactions to be included in the next block. This is done without executing them. They are simply subjected to a minimal pre-filter by the Cosmos SDK module, to ensure they are validly formatted transactions, with sufficient gas fees, and signed by an account with sufficient fees to pay it. Notably, this means many transactions that error may be included in a block.
> Once a block is committed (typically every 5s or so), the transactions are then fed to the Cosmos SDK sequentially in order to execute them. Each one returns a result or error along with event logs, which are recorded in the TxResults section of the next block. The AppHash (or merkle proof or blockchain state) after executing the block is also included in the next block.

### Drive-by changes

Fix a typo for `CosmosMerkleTreeHookIndexer`

### Related issues

n/a

### Backward compatibility

non breaking

### Testing

None 😛

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/interchain_gas.rs
+++ b/rust/chains/hyperlane-cosmos/src/interchain_gas.rs
@@ -59,8 +59,14 @@ pub struct CosmosInterchainGasPaymasterIndexer {
 
 impl CosmosInterchainGasPaymasterIndexer {
     /// create new Cosmos InterchainGasPaymasterIndexer agent
-    pub fn new(conf: ConnectionConf, locator: ContractLocator, event_type: String) -> Self {
-        let indexer: CosmosWasmIndexer = CosmosWasmIndexer::new(conf, locator, event_type.clone());
+    pub fn new(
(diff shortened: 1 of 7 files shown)


daniel-savu (approved):

daniel-savu:
So tx inclusion and tx execution are decoupled :o

> a list of transactions to be included in the next block. This is done without executing them

PR #2886, opened 2023-11-02
Index merkle tree insertions just like messages

tkporter:
### Description

Moves merkle tree hook insertion indexing away from the watermarked contract sync to the same method that message indexing uses

* Generalizes all the MessageContractSync / MessageSyncCursor etc to work with any "sequenced" data. Sequenced data means a data type that has a sequence # that increases 1 at a time as it's created. E.g. a message's nonce is its sequence, or a merkle tree insertion's leaf index is its sequence

### Drive-by changes

Minor idiomatic changes

### Related issues

n/a

### Backward compatibility

I believe this is fully backward compatible

### Testing

Tested locally by running a relayer with an empty data dir & with an existing data dir. Also tested a validator locally

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
+++ b/rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
@@ -179,9 +179,18 @@ pub struct CosmosMerkleTreeHookIndexer {
 
 impl CosmosMerkleTreeHookIndexer {
     /// create new Cosmos MerkleTreeHookIndexer agent
-    pub fn new(conf: ConnectionConf, locator: ContractLocator, signer: Signer, reorg_period: u32) -> Self {
-        let indexer: CosmosWasmIndexer =
-            CosmosWasmIndexer::new(conf.clone(), locator.clone(), EVENT_TYPE.to_string(), reorg_period);
+    pub fn new(
+        conf: ConnectionConf,
+        locator: ContractLocator,
+        signer: Signer,
+        reorg_period: u32,
+    ) -> Self {
+        let indexer: CosmosWasmIndexer = CosmosWasmIndexer::new(
+            conf.clone(),
+            locator.clone(),
+            EVENT_TYPE.to_string(),
+            reorg_period,
+        );
 
         Self {
             merkle_tree_hook: CosmosMerkleTreeHook::new(conf, locator, signer),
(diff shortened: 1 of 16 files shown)


rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
```diff
@@ -104,26 +104,14 @@ impl MerkleTreeHook for CosmosMerkleTreeHook {
     }
 
     /// Gets the current leaf count of the merkle tree
-    #[instrument(level = "debug", err, ret, skip(self))]
```
daniel-savu, line 107: logging the return value was a bit much, but it probably makes sense to still log on errors?

rust/chains/hyperlane-ethereum/src/merkle_tree_hook.rs
```diff
@@ -148,14 +148,9 @@ where
     M: Middleware + 'static,
 {
     async fn sequence_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
-        // The InterchainGasPaymasterIndexerBuilder must return a `SequenceIndexer` type.
-        // It's fine if only a blanket implementation is provided for EVM chains, since their
-        // indexing only uses the `Index` trait, which is a supertrait of `SequenceIndexer`.
-        // TODO: if `SequenceIndexer` turns out to not depend on `Indexer` at all, then the supertrait
```
daniel-savu, line 154: this todo should stay afaict

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -270,16 +272,31 @@ impl HyperlaneLogStore<MerkleTreeInsertion> for HyperlaneRocksDB {
...
+impl HyperlaneSequenceIndexerStore<MerkleTreeInsertion> for HyperlaneRocksDB {
+    /// Gets data by its sequence.
+    async fn retrieve_by_sequence(&self, sequence: u32) -> Result<Option<MerkleTreeInsertion>> {
+        let insertion = self.retrieve_merkle_tree_insertion_by_leaf_index(&sequence)?;
+        Ok(insertion)
+    }
+
+    /// Gets the block number at which the log occurred.
+    async fn retrieve_log_block_number(&self, sequence: u32) -> Result<Option<u64>> {
+        let number = self.retrieve_merkle_tree_insertion_block_number_by_leaf_index(&sequence)?;
```
daniel-savu, line 299: I wasn't able to find where the `store_` counterpart of this is called, although `get_next_range` depends on this call. Wondering how it didn't cause any issues on your local, maybe because of the ism setup?

daniel-savu (approved):

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
tkporter: oh good call - the reason things worked is because it's not strictly required

PR #2887, opened 2023-11-02
refactor: delegate keypair encoding to cosmrs

daniel-savu:
### Description

Delegates bech32 encoding, decoding, and cosmos address building from a private key to cosmrs/tendermint upstream crates. The only place where this wasn't doable was the tests, because of a cyclic dependency - left a TODO comment there.

As part of this, renamed `rust/chains/hyperlane-cosmos/src/libs/verify.rs` -> `rust/chains/hyperlane-cosmos/src/libs/address.rs`

~~**Warning**: We should re-enable e2e and make sure everything still works locally, since the actual `encode`/`decode` functions that end up being called _are_ different~~

### Testing
Tested manually EVM <> Duality (both ways)


rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -84,16 +70,9 @@ impl InterchainSecurityModule for CosmosInterchainSecurityModule {
...
-        let ismtype_response = serde_json::from_slice::<QueryIsmModuleTypeResponse>(&data);
-        let moduletye_response =
-            serde_json::from_slice::<hpl_interface::ism::ModuleTypeResponse>(&data);
-
-        Ok(match (ismtype_response, moduletye_response) {
-            (Ok(v), _) => ism_type_to_module_type(v.typ),
-            (_, Ok(v)) => ism_type_to_module_type(v.typ),
-            _ => ModuleType::Null,
-        })
+        let module_type_response =
```
tkporter, line 73: 😎

rust/chains/hyperlane-cosmos/src/libs/address.rs
```diff
@@ -0,0 +1,53 @@
...
+use cosmrs::{crypto::secp256k1::SigningKey, AccountId};
+use derive_new::new;
+use hyperlane_core::{ChainResult, H256};
+
+/// decode bech32 address to H256
+pub fn bech32_decode(addr: String) -> ChainResult<H256> {
+    let account_id = AccountId::from_str(&addr)?;
+
+    // although `from_slice` can panic if the slice is not 32 bytes long,
+    // we know that we're passing in a value that is 32 bytes long because it was decoded from
```
tkporter, line 12: I believe bech32 is named that because it's the encoding is done in base 32 and not because it's 32 bytes long. I think the pattern in Cosmos is that EOA addresses are 20 bytes and smart contracts are 32 bytes. So we should probably be more defensive here

Looks like account IDs can get up to 255 bytes https://docs.rs/cosmrs/latest/cosmrs/struct.AccountId.html#associatedconstant.MAX_LENGTH

rust/chains/hyperlane-cosmos/src/libs/address.rs
```diff
@@ -0,0 +1,53 @@
...
+    // bech32
+    Ok(H256::from_slice(&account_id.to_bytes()))
+}
+
+/// Wrapper around the cosmrs AccountId type that abstracts keypair conversions and
+/// bech32 encoding
+#[derive(new)]
+pub struct CosmosAddress {
```
tkporter, line 20: may be a good place to have some quick unit tests if we're uncertain about the encoding / decoding like iiuc you flag in the PR description

rust/chains/hyperlane-cosmos/src/libs/binary.rs
```diff
@@ -1,28 +0,0 @@
-use hyperlane_core::{H160, H256, H512};
```
tkporter, line 1: 🔥

rust/chains/hyperlane-cosmos/src/libs/verify.rs
```diff
@@ -1,132 +0,0 @@
-use std::cmp;
```
tkporter, line 1: 🪦

rust/chains/hyperlane-cosmos/src/multisig_ism.rs
```diff
@@ -78,7 +77,7 @@ impl MultisigIsm for CosmosMultisigIsm {
         let validators: ChainResult<Vec<H256>> = response
             .validators
             .iter()
-            .map(|v| H160::from_str(v).map(h160_to_h256).map_err(Into::into))
+            .map(|v| H256::from_str(v).map_err(Into::into))
```
tkporter, line 80: I believe H256::from_str panics if the input length isn't 32 bytes, so we'll probably need to keep doing H160::from_str(v) and then .into() to H256

tkporter (approved):
A few spots that I think may cause issues, if we don't have e2e maybe you could stand up a testnet relayer locally and send a message on testnet or something?

rust/chains/hyperlane-cosmos/src/multisig_ism.rs
daniel-savu: great point

rust/chains/hyperlane-cosmos/src/libs/address.rs
daniel-savu: unit tests broke this, good point

daniel-savu:
Fixed the address generation, just need to add a couple more unit tests

rust/utils/run-locally/src/cosmos/crypto.rs
```diff
@@ -1,3 +1,7 @@
+// TODO: this file can be removed if `CosmosAddress` can be imported from `hyperlane-cosmos`.
```
tkporter, line 1: I assume this is what you're currently looking into?

tkporter (approved):
lgtm! nice job

rust/utils/run-locally/src/cosmos/crypto.rs
daniel-savu: actually no but I'll look into this as well

PR #2890, opened 2023-11-03
Optional cosmos signer

daniel-savu:
Removes cosmos signer unwraps by making the signer optional. Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2887


rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -289,8 +292,9 @@ impl WasmProvider for WasmGrpcProvider {
     where
         T: Serialize + Send + Sync,
     {
+        let signer = self.get_signer()?;
```
tkporter, line 295: Flagging that between this & how wasm_simulate -> simulate_raw_tx -> generate_raw_tx which also requires the signer, in practice a signer is still required. I think it's fine to not address this in this PR and for us to just refactor `grpc.rs` a bunch so that simulating calls doesn't require a signer anymore, but wanted to call out

tkporter (approved):

PR #2896, opened 2023-11-07
Remove FS dep from sdk

yorhodes:
### Description

- Removes FS introduced by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2826

### Backward compatibility

- No, breaks Op Ism recovery
- Opens https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2898

### Testing

None

### DIFF
--- a/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
+++ b/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
@@ -1,7 +1,5 @@
 import { debug } from 'debug';
 import { ethers } from 'ethers';
-import fs from 'fs';
-import path from 'path';
 
 import {
   DomainRoutingIsm__factory,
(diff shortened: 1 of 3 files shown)


typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -226,20 +223,11 @@ export class HyperlaneIsmFactory extends HyperlaneApp<FactoryFactories> {
   }
 
   private async deployOpStackIsm(chain: ChainName, config: OpStackIsmConfig) {
-    const recoveredIsm = getDeployedIsms(config.origin, chain, config.type);
```
aroralanuk, line 229: I would prefer to throw an "unimplemented" error rather than just deploy for future error handling to avoid misconfigured opstackhooks <> opstackisms

aroralanuk (changes requested):

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
yorhodes: opened https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2898
generally if you wanted to remediate you had to get on this quickly

aroralanuk (approved):

PR #2904, opened 2023-11-09
Add gap back to MailboxClient

0xmichalis:
The gap should ensure MailboxClient is
upgrade-safe as long as future additions
to the contract reduce the gap properly.


aspiers:
Would be super grateful if y'all can keep these contracts safely upgradeable in future! :pray:

yorhodes:
pulled in to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2984

PR #2905, opened 2023-11-09
GRPC refactor

tkporter:
### Description

* Creates a single channel that's cloned (which can be done [cheaply and is encouraged](https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html)) when creating a new client.
* general cleanup & renaming
* I believe a chain signer will not be required by e.g. the validator, will confirm this locally

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

yee

### Testing

builds, local test


rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -1,123 +1,241 @@
...
+    /// Generates an unsigned SignDoc for a transaction.
+    async fn generate_unsigned_sign_doc(
+        &self,
+        msgs: Vec<cosmrs::Any>,
+        gas_limit: u64,
+    ) -> ChainResult<SignDoc> {
+        // As this function is only used for estimating gas or sending transactions,
+        // we can reasonably expect to have a signer.
+        let signer = self.get_signer()?;
+        let account_info = self.account_query(signer.address.clone()).await?;
```
daniel-savu, line 131: I assume we're not precomputing this in the constructor because we want to keep it synchronous? Don't think it's worth caching otherwise but was just wondering

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -1,123 +1,241 @@
...
+        let current_height = self.latest_block_height().await?;
+        let timeout_height = current_height + TIMEOUT_BLOCKS;
+
+        let tx_body = tx::Body::new(
+            msgs,
+            "",
```
daniel-savu, line 137: cleaner to use `Default::default()` here

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -1,123 +1,241 @@
...
-        Ok(cosmos_address.address())
+    /// Estimates gas for a transaction containing `msgs`.
+    async fn estimate_gas(&self, msgs: Vec<cosmrs::Any>) -> ChainResult<u64> {
+        // Get a sign doc with 0 gas, because we plan to simulate
+        let sign_doc = self.generate_unsigned_sign_doc(msgs, 0).await?;
+
+        let raw_tx = TxRaw {
+            body_bytes: sign_doc.body_bytes,
+            auth_info_bytes: sign_doc.auth_info_bytes,
+            // The poorly documented trick to simuluating a tx without a valid signature is to just pass
```
daniel-savu, line 187: nice

daniel-savu (approved):
master grpc optimizor 🏎️

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
tkporter: I think "account number" can be cached but fetching the sequence number (i.e. the nonce to use for the next tx) should still be done. Just in case other txs from that key happen to have been made elsewhere

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
tkporter: the expected value is `impl Into<String>` so I'll just do String::default()

PR #2906, opened 2023-11-09
hyperlane

byeongsu-hong:

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -174,8 +174,8 @@ ripemd = "0.1.3"
 hyper = "0.14"
 hyper-tls = "0.5.0"
 hex = "0.4.3"
-tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"]}
-hpl-interface = "0.0.6-rc5"
+tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"] }
+hpl-interface = "=0.0.6-rc3"
 elliptic-curve = "0.12.3"
 
 ## TODO: remove this
--- a/rust/utils/run-locally/src/cosmos/deploy.rs
+++ b/rust/utils/run-locally/src/cosmos/deploy.rs
@@ -53,16 +53,26 @@ pub fn deploy_cw_hyperlane(
     );
 
     // deploy igp set
+    #[cw_serde]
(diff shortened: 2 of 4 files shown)


rust/Cargo.toml
```diff
@@ -174,8 +174,8 @@ ripemd = "0.1.3"
 hyper = "0.14"
 hyper-tls = "0.5.0"
 hex = "0.4.3"
-tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"]}
-hpl-interface = "0.0.6-rc5"
+tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"] }
+hpl-interface = "=0.0.6-rc3"
```
daniel-savu, line 178: how come this is using `rc3` but run-locally uses `rc6`?

rust/utils/run-locally/src/cosmos/deploy.rs
```diff
@@ -53,16 +53,26 @@ pub fn deploy_cw_hyperlane(
     );
 
     // deploy igp set
+    #[cw_serde]
+    pub struct Something {
```
daniel-savu, line 57: I assume this should be named `InstantiateMsg`

byeongsu-hong: Yes, exactly. It's quite weird that same payload does not working

rust/Cargo.toml
byeongsu-hong: it's a hacky approach to make runnable with this - the reason of doing like this is i realized payload in rc6 is not up to date with rc6 contract binary, so i published new version in empty slot temporarily.

rust/utils/run-locally/src/cosmos/deploy.rs
daniel-savu: I'd like to avoid generic names like `Something` though, could you use a more descriptive name?

rust/Cargo.toml
daniel-savu: Makes sense, I'll merge it like this but let's open a follow-up PR to address when it's fixed

daniel-savu (approved):
Thank you!

PR #2909, opened 2023-11-10
chore: use larger runner for github actions

daniel-savu:
Size of rust build exceeds the storage space in the `ubuntu-latest` runner. `larger-runner` has more storage, but let's see if it also has all the system dependencies required to build

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -21,7 +21,7 @@ defaults:
 
 jobs:
   e2e:
-    runs-on: ubuntu-latest
+    runs-on: larger-runner
     steps:
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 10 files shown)


.github/workflows/e2e.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: '*'
```
tkporter, line 7: was this just to ensure it was triggered and should be brought back? I'm bad at gh actions so idk what the implications of this are

tkporter (approved):

.github/workflows/e2e.yml
daniel-savu: `branches` was used as a filter. Not sure why "*" wasn't enough, but by removing it we're saying "always run e2e on PRs"

daniel-savu: so this line should stay removed

PR #2888, opened 2023-11-03
Enroll additional validators on v3 mainnet

nambrot:
### Description

- This PR adds the latest validators to the config

### Drive-by changes

- Fixes fork tests

### Testing

- `test-env mainnet3 core`

PR #2911, opened 2023-11-10
Improve ergonomics of fork tests

yorhodes:
### Description

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2888 introduced breaking changes to non-`arbitrum` chains but fork tests would not cover
This PR adds a `fork-all` script for running fork tests on all `protocol == ethereum` chains in an environment
It also improves the interface of the `fork` script to accept a target chain

### Drive-by changes

- Remove noisy `console.log`s
- swap `polygon, ethereum, polygonzkevm` RPCs
- Fixes `ownerOverrides: proxyAdmin` for chains without gnosis safes

### Backward compatibility

Yes


typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -13,10 +13,10 @@ export const safes: ChainMap<Address> = {
...
   // solana: 'EzppBFV2taxWw8kEjxNYvby6q7W1biJEqwP3iC7YgRe3',
   // TODO: create gnosis safes here
-  base: '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba',
-  scroll: '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba',
-  polygonzkevm: '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba',
-  mantapacific: '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba',
+  base: undefined,
+  scroll: undefined,
+  polygonzkevm: undefined,
+  mantapacific: undefined,
```
jmrossy, line 19: Why unset these?

typescript/sdk/src/consts/chainMetadata.ts
```diff
@@ -320,7 +320,7 @@ export const ethereum: ChainMetadata = {
   nativeToken: etherToken,
   protocol: ProtocolType.Ethereum,
   rpcUrls: [
-    { http: 'https://mainnet.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161' },
+    { http: 'https://ethereum.publicnode.com' },
```
jmrossy, line 323: So the key-ed infura endpoint doesn't support fork testing but this public one does? That's surprising

jmrossy (approved):

typescript/infra/config/environments/mainnet3/owners.ts
yorhodes: these were just the deployer address, not actually gnosis safes and it was misleading

typescript/sdk/src/consts/chainMetadata.ts
yorhodes: I can revert this, I thought it was weird to have the keyed endpoint here

PR #2912, opened 2023-11-10
Dan/more cw agent remediations

daniel-savu:

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,43 +1,43 @@
 [workspace]
 members = [
-    "agents/relayer",
-    "agents/scraper",
-    "agents/validator",
-    "chains/hyperlane-cosmos",
-    "chains/hyperlane-ethereum",
-    "chains/hyperlane-fuel",
-    "chains/hyperlane-sealevel",
-    "ethers-prometheus",
-    "hyperlane-base",
-    "hyperlane-core",
-    "hyperlane-test",
-    "sealevel/client",
-    "sealevel/libraries/access-control",
-    "sealevel/libraries/account-utils",
-    "sealevel/libraries/ecdsa-signature",
-    "sealevel/libraries/hyperlane-sealevel-connection-client",
-    "sealevel/libraries/hyperlane-sealevel-token",
-    "sealevel/libraries/interchain-security-module-interface",
-    "sealevel/libraries/message-recipient-interface",
-    "sealevel/libraries/multisig-ism",
-    "sealevel/libraries/serializable-account-meta",
-    "sealevel/libraries/test-transaction-utils",
-    "sealevel/libraries/test-utils",
-    "sealevel/programs/hyperlane-sealevel-token",
-    "sealevel/programs/hyperlane-sealevel-token-collateral",
-    "sealevel/programs/hyperlane-sealevel-token-native",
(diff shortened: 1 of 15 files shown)


rust/Cargo.toml
daniel-savu, line 1: I used `taplo` to reorder dependencies alphabetically, looks like it reordered all the keys alphabetically too

rust/hyperlane-base/src/settings/parser/connection_parser.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use super::ValueParser;
+
+pub fn build_ethereum_connection_conf(
+    rpcs: &Vec<Url>,
+    chain: &ValueParser,
+    err: &mut ConfigParsingError,
+    default_rpc_consensus_type: &str,
+) -> Option<ChainConnectionConf> {
+    if rpcs.len() <= 1 {
```
tkporter, line 17: do we need to handle the case of it being len 0 where we'd I guess return None?

rust/hyperlane-core/src/traits/mod.rs
```diff
@@ -60,3 +62,18 @@ impl From<ethers_core::types::TransactionReceipt> for TxOutcome {
...
+    type Error = ChainCommunicationError;
+
+    fn try_from(
+        response: cosmrs::proto::cosmos::base::abci::v1beta1::TxResponse,
+    ) -> ChainResult<Self> {
+        Ok(Self {
+            transaction_id: H256::from_slice(hex::decode(response.txhash)?.as_slice()).into(),
+            executed: response.code == 0,
+            gas_used: U256::from(response.gas_used),
+            gas_price: U256::from(response.gas_wanted),
```
tkporter, line 76: I know this wasn't your change but the gas wanted is the gas limit for tx and not the gas price

I assume this was originally set like this because it wasn't available in `TxResponse` - imo it's fine to just set it to 1 or something for now (or to get it from the ceil of here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/trevor/new-featv3-cosmos-oct-28/rust/chains/hyperlane-cosmos/src/providers/grpc.rs#L36, which is how we set the gas price atm - unfortunately there doesn't seem to be a reasonable way of getting it via RPC so atm it's just hardcoded :/), but to add a comment about it

tkporter (approved):

rust/hyperlane-base/src/settings/parser/connection_parser.rs
daniel-savu: we are returning `None` on len 0 here, because `next()` will return `None` and `map` doesn't apply the closure to `None`

daniel-savu: however I agree this implementation could be cleaner, there's an `unwrap` in the `else` branch which relies on the assumption that we're checking the length. Trying to make this nicer

rust/hyperlane-core/src/traits/mod.rs
daniel-savu: I see :/ `DEFAULT_GAS_PRICE` isn't available here because it lives in hyperlane-cosmos, and the TryFrom implementation has to live in this crate due to the orphan rule. I'll just set it to `U256::one()`

PR #2922, opened 2023-11-15
fix: reuse default validators for CLI

aroralanuk:
### Description

- if the default multisig config contains the config for the chain we want to deploy on and we've not explicitiy provided a config, the CLI will use the default config

### Drive-by changes

- ignoring the ismType in `multisig-ism.yaml` and deploying messageIdMultisig for all multisig configs provided indiscriminately. Also, added a note regarding this in the `example/multisig-ism.yaml` file. (otherwise we end up using `MultisigIsmConfig` and `MultisigConfig` interchangeably like in the `buildIgpConfig` which leads to confusion)

**NOTE**: this doesn't affect our inability to not deploy the same multisig with different types for different remote chains because the config provided is origin specific, eg, you cannot use goerli multisig as messageId for arbitrumsepolia and for scrollsepolia as merkleRoot.

- moved `buildMultisigIsmConfigs` to sdk as an effort to dedupe the two variations in infra and cli.

### Related issues

- aimed at https://github.com/hyperlane-xyz/issues/issues/530#issuecomment-1802766305

### Backward compatibility

Yes

### Testing

Manual


jmrossy (approved):
Lgtm

PR #2927, opened 2023-11-16
cli gas preflight

yorhodes:
### Description

- Check min gas on each chain in preflight

### DIFF
--- a/typescript/cli/ci-test.sh
+++ b/typescript/cli/ci-test.sh
@@ -18,8 +18,8 @@ do
     chmod -R 777 /tmp/relayer /tmp/$CHAIN
 done
 
-anvil --chain-id 31337 -p 8545 --state /tmp/anvil1/state > /dev/null &
-anvil --chain-id 31338 -p 8555 --state /tmp/anvil2/state > /dev/null &
+anvil --chain-id 31337 -p 8545 --state /tmp/anvil1/state --gas-price 1 > /dev/null &
+anvil --chain-id 31338 -p 8555 --state /tmp/anvil2/state --gas-price 1 > /dev/null &
 sleep 1
 
 set -e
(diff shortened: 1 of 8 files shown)


typescript/cli/ci-test.sh
```diff
@@ -39,6 +42,11 @@ yarn workspace @hyperlane-xyz/cli run hyperlane deploy core \
     --key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
     --yes
 
+AFTER_CORE=$(cast balance $DEPLOYER --rpc-url http://localhost:8545)
+GAS_PRICE=$(cast gas-price --rpc-url http://localhost:8545)
+CORE_MIN_GAS=$(bc <<< "($BEFORE - $AFTER_CORE) / $GAS_PRICE")
```
jmrossy, line 47: bash math 😎

typescript/cli/src/utils/balances.ts
```diff
@@ -18,10 +19,31 @@ export async function assertNativeBalances(
...
-        throw new Error(
-          `${address} has insufficient balance on ${chain}. At least ${minBalance} required but found ${balance.toString()} ETH`,
-        );
+      if (balanceWei.lt(minBalanceWei)) {
+        const symbol =
+          multiProvider.getChainMetadata(chain).nativeToken?.symbol ?? 'ETH';
+        const error = `${address} has insufficient balance on ${chain}. At least ${minBalance} required but found ${balance.toString()} ${symbol}`;
+        const isResume = await confirm({
+          message: `WARNING: ${error} Continue?`,
+        });
```
jmrossy, line 28: My first thought was that like other confirm requests, this should skip if the user has included `--yes` flag. However, if they're doing that they should probably be made aware of balance issues so I guess let's leave this as is.

jmrossy (approved):
Ship it

PR #2928, opened 2023-11-16
pass in ismfactory

aroralanuk:
### Description

- passing in ismFactory to the HypERC20Deployer and HypERC721Deployer

### Drive-by changes

None

### Testing
 
None

PR #2932, opened 2023-11-16
Fix core contract caching and recovery

jmrossy:
- Fix core contract recovery
- Simplify deployment plan text
- Soften balance warning message


typescript/cli/src/deploy/core.ts
```diff
@@ -284,11 +279,6 @@ async function executeDeploy({
   const ismContracts: ChainMap<{ multisigIsm: DeployedIsm }> = {};
   const defaultIsms: ChainMap<Address> = {};
   for (const ismOrigin of chains) {
-    if (artifacts[ismOrigin].multisigIsm) {
```
aroralanuk, line 287: you can pass in IsmConfig and it should be able to recover the multisigIsm based on the validators you give

checkout https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2928

aroralanuk (changes requested):

typescript/cli/src/deploy/core.ts
aroralanuk: actually, not necessary right now. Approved!

aroralanuk (approved):

PR #2949, opened 2023-11-21
Add yarn changesets

yorhodes:
### Description

- monorepo commands will only run for packages which have changed since baseRef: main

### Drive-by changes

- modify github actions to leverage filtered behavior
- upgrade to yarn 4.0.2

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/708
- Fixes https://github.com/hyperlane-xyz/issues/issues/374


tkporter (approved):

aroralanuk:
Where did you put the `test-sol` job, it's missing from `node.yaml`? Also coverage should be triggered on `solidity/* path too?

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Setup Node.js 18.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Create Release Pull Request or Publish to npm
```
jmrossy, line 30: I wonder if the changeset + CI based release will automatically enable the provenance cert or if we need to do something more for that: https://docs.npmjs.com/generating-provenance-statements

.github/workflows/node.yml
```diff
@@ -90,27 +96,25 @@ jobs:
...
-      - name: sdk
-        run: yarn workspace @hyperlane-xyz/sdk run test
-
-      - name: helloworld
-        run: yarn workspace @hyperlane-xyz/helloworld run test
-
-      - name: infra
-        run: yarn workspace @hyperlane-xyz/infra run test
+      - name: Unit Tests
+        run: yarn test
```
jmrossy, line 117: I don't see where the magic happens to prevent test-sol when no changes were made there. Is that part of this?

yorhodes:
> Where did you put the `test-sol` job, it's missing from `node.yaml`? Also coverage should be triggered on `solidity/* path too?

I renamed it because it now includes the sol and ts tests
coverage will be triggered for any workspace that has a diff, including solidity

package.json
```diff
@@ -12,19 +12,23 @@
     "lint-staged": "^12.4.3",
     "prettier": "^2.8.8"
   },
-  "packageManager": "yarn@4.0.1",
+  "dependencies": {
+    "@changesets/cli": "^2.26.2"
```
jmrossy, line 16: It's a dev dep right?

aroralanuk:
> > Where did you put the `test-sol` job, it's missing from `node.yaml`? Also coverage should be triggered on `solidity/* path too?
> 
> I renamed it because it now includes the sol and ts tests coverage will be triggered for any workspace that has a diff, including solidity

isn't this adding time and blocking non-solidity tests by checking solidity tests

.github/workflows/node.yml
yorhodes: the `--since` flag in the root `workspaces foreach` commands does this, which this leverages

jmrossy (approved):

package.json
yorhodes: https://www.npmjs.com/package/@changesets/cli docs for whatever reason say to add as a normal dep in a monorepo

yorhodes:
> > I renamed it because it now includes the sol and ts tests coverage will be triggered for any workspace that has a diff, including solidity
> 
> isn't this adding time and blocking non-solidity tests by checking solidity tests

for some PRs, it may result in a slight increase in runtime yes...
the test-sol dominated runtime anyway so I think this should be fine

.github/workflows/release.yml
yorhodes: added in db58ae13c

Issue #2959, opened 2023-11-22
Investigate ethers-rs middleware for "stuck" txs


tkporter:
### Context
The relayer submits a transaction with a gas price that is no longer high enough to execute, but instead of the relayer submitting on certain networks like polygon the tx is stuck in the mempool. To "unstuck" the relayer needs to submit a higher bid with x% more gas, but our relayer currently lacks that logic.

https://docs.rs/ethers/latest/ethers/middleware/struct.GasEscalatorMiddleware.html is probably what we want

Some things to note:
- See https://cloudlogging.app.goo.gl/jXpgMgTHsWZpUk1s9 for examples of this. Interestingly, this only seems to happen with polygon txs and still occurs pretty frequently.
- This was originally written with Polygon in mind that we fixed in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2965. Addressing https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2924 is now the main focus of this. Seemingly the Polygon fix wasn't totally sufficient 
- Be careful of our usage around fallback provider here (e.g. if the gas escalation only works at the "inner provider" level and not at a higher level, this may not actually be useful)
- Be careful that we construct new providers pretty frequently, e.g. when constructing metadata. I believe we use the same provider for all tx submission, but maybe we don't. I believe we'd need to use the same long-living provider if we wanted to use the gas escalator middleware

yorhodes:
https://docs.rs/ethers/latest/ethers/middleware/struct.GasEscalatorMiddleware.html

tkporter:
nice, we'll have to make sure the footgun / issues aren't a problem for us & also that this covers the case of immediately getting back a "transaction underpriced" error and not just if the tx doesn't get included in a block in time

daniel-savu:
> Be careful of our usage around fallback provider here (e.g. if the gas escalation only works at the "inner provider" level and not at a higher level, this may not actually be useful)

Afaict ([source code](https://docs.rs/ethers-middleware/2.0.14/src/ethers_middleware/gas_escalator/mod.rs.html#158-163)) the logic is compatible with our fallback provider middleware - we just need to make sure that "Gas Escalator" is the outermost middleware.



> Be careful that we construct new providers pretty frequently, e.g. when constructing metadata. I believe we use the same provider for all tx submission, but maybe we don't. I believe we'd need to use the same long-living provider if we wanted to use the gas escalator middleware

[...]

Issue #771, opened 2022-07-15
Fetch proper gas price from Polygon


asaj:
> error: Error: err: max fee per gas less than block base fee: address 0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba, maxFeePerGas: 100000000000 baseFee: 178411458447 (supplied gas 13894823)

It looks like the gas price is being hardcoded

Ethers apparently has an issue around this:
https://github.com/ethers-io/ethers.js/issues/2828

asaj:
And mumbai, for that matter

asaj:
Well this is interesting:
https://github.com/ethers-io/ethers.js/issues/2828#issuecomment-1283014250

Looks like they technically have EIP-1559 but it basically functions as legacy PGA.

avious00:
hey @tkporter @nambrot did the polygon gas oracle PR close this out?

nambrot:
My understanding of this ticket is to do this for the JS side, we only dont it on the rust side

tkporter:
+1 I'd consider this the work to remove the gas price overrides for polygon / mumbai, e.g. https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/chains.ts#L33-L34

PR #2965, opened 2023-11-22
Use Polygon gas oracle

tkporter:
### Description

After investigating #2959, I found the following

this is a known problem with Polygon, explanation here https://github.com/ethers-io/ethers.js/issues/2828#issuecomment-1283014250

Fun fact Asa looked into this once https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/771

Here's a discussion it in Foundry, which I found hoping that ethers-rs folks had ran into this before
https://github.com/foundry-rs/foundry/issues/1703

Foundry fixed this by using the Polygon gas station oracle which seems to be recommended path https://github.com/foundry-rs/foundry/pull/3368/files#diff-c89a4bbf7a90da118dcf00c5fe70eba78f8e5d95662bb5f039a353113e95042bR205

There's actually a polygon ethers middleware for this https://docs.rs/ethers/latest/ethers/middleware/gas_oracle/polygon/struct.Polygon.html

So I (originally) borrowed this code from Foundry https://github.com/foundry-rs/foundry/blob/master/crates/common/src/provider.rs#L254-L290

Changed to use Middlewares

This also means we can remove our existing janky Polygon logic

### DIFF
--- a/rust/chains/hyperlane-ethereum/src/trait_builder.rs
+++ b/rust/chains/hyperlane-ethereum/src/trait_builder.rs
@@ -190,7 +190,7 @@ pub trait BuildableWithProvider {
     where
         P: JsonRpcClient + 'static,
     {
-        let provider = wrap_with_gas_oracle(Arc::new(Provider::new(client)), locator.domain)?;
+        let provider = wrap_with_gas_oracle(Provider::new(client), locator.domain)?;
 
         Ok(if let Some(metrics) = metrics {
             let provider = Arc::new(PrometheusMiddleware::new(provider, metrics.0, metrics.1));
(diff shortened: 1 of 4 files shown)


yorhodes (approved):
nice work
would be good to just import as a middleware and add to the stack

tkporter:
Changed to use middlewares - note this was a bit more painful / ugly than you'd probably expect, which is a result of the `Middleware` trait not being object safe. So conditional usage of a middleware gets really messy (see https://github.com/gakonst/ethers-rs/issues/592)

The GasOracleMiddleware will apply the provided GasOracle regardless of what chain it is, so we'd need to either:
a) conditionally use the GasOracleMiddleware, or
b) because `GasOracle` is object safe, just always use a GasOracleMiddleware but use a different GasOracle depending on the chain

With (a), we end up with a bunch of really ugly conditional paths because we always need a concrete type for the provider/middleware object

And with (b), we need to have a GasOracle that will signal to the `GasOracleMiddleware` to not use a gas oracle at all - they have `ProviderOracle`, which lets you use the provider's oracle instead. This is fine, but the downside is to make this work we have to make the provider `Arc<M>` bc `ProviderOracle` requires ownership of the provider

rust/chains/hyperlane-ethereum/src/trait_builder.rs
```diff
@@ -237,3 +242,39 @@ async fn build_signing_provider<M: Middleware>(
     let signing_provider = SignerMiddleware::new(provider, signer);
     Ok(signing_provider)
 }
+
+/// Wrap the provider with a gas oracle middleware.
+/// Polygon and Mumbai require using the Polygon gas oracle, see discussion here
+/// https://github.com/foundry-rs/foundry/issues/1703.
+/// Defaults to using the provider's gas oracle.
+fn wrap_with_gas_oracle<M>(
+    provider: Arc<M>,
```
daniel-savu, line 251: based on how `wrap_with_gas_oracle` is used, it could just take `M` and wrap it in the `Arc` in the function body

rust/chains/hyperlane-ethereum/src/trait_builder.rs
```diff
@@ -237,3 +242,39 @@ async fn build_signing_provider<M: Middleware>(
...
+                let gas_oracle = Polygon::new(chain)
+                    .map_err(ChainCommunicationError::from_other)?
+                    .category(GasCategory::Standard);
+                Box::new(gas_oracle)
+            }
+            _ => Box::new(ProviderOracle::new(provider.clone())),
+        }
+    };
+    Ok(GasOracleMiddleware::new(provider, gas_oracle))
+}
```
daniel-savu, line 280: nit:
```suggestion
fn build_polygon_gas_oracle(chain: ethers_core::types::Chain) -> ChainResult<Box<dyn GasOracle>> {
    let gas_oracle = Polygon::new(chain)
        .map_err(ChainCommunicationError::from_other)?
        .category(GasCategory::Standard);
    Ok(Box::new(gas_oracle) as Box<dyn GasOracle>)
}

/// Wrap the provider with a gas oracle middleware.
/// Polygon and Mumbai require using the Polygon gas oracle, see discussion here
/// https://github.com/foundry-rs/foundry/issues/1703.
/// Defaults to using the provider's gas oracle.
fn wrap_with_gas_oracle<M>(
    provider: Arc<M>,
    domain: &HyperlaneDomain,
) -> ChainResult<GasOracleMiddleware<Arc<M>, Box<dyn GasOracle>>>
where
    M: Middleware + 'static,
{
    let gas_oracle: Box<dyn GasOracle> = {
        match domain {
            HyperlaneDomain::Known(KnownHyperlaneDomain::Polygon) => {
                build_polygon_gas_oracle(ethers_core::types::Chain::Polygon)?
            }
            HyperlaneDomain::Known(KnownHyperlaneDomain::Mumbai) => {
                build_polygon_gas_oracle(ethers_core::types::Chain::PolygonMumbai)?
            }
            _ => Box::new(ProviderOracle::new(provider.clone())),
        }
    };
    Ok(GasOracleMiddleware::new(provider, gas_oracle))
}

```

daniel-savu (approved):

PR #2974, opened 2023-11-24
Use Polygon gas oracle (#2965)

nambrot:
Cherry-pick from #2965


tkporter (approved):

nambrot:
@tkporter cherry-picked #3000 onto here (not sure why the commit message doesnt reflect that), that should be all right?

PR #2976, opened 2023-11-24
Relayer balance metrics

daniel-savu:
Done:
- scaffolding for fetching custom agent metrics
- abstractions for building a metrics fetcher for a given VM
- querying cosmos balances; e2e tested.
- querying evm balances; e2e tested.
    - **Note that as a result, evm addresses are now no longer zero-padded when printed in the logs. This may break existing log queries**
- fixed a nasty bug on ubuntu where wasmd (osmosisd dependency, part of the grpc query flow) would panic when a block is specified via `x-cosmos-block-height`. The fix was to bump the version of osmosisd from `19.0.0` to `20.5.0`. **Note that when running e2e on Mac OS, the osmosis version in use is still 19.0.0**. That's because we need a fork that publishes a darwin target binary (currently pointing [here](https://github.com/hashableric/osmosis/releases/download/v19.0.0-mnts/osmosisd-19.0.0-mnts-darwin-arm64.tar.gz))

For follow up PR:
- sealevel balance querying

I'm open to all renaming suggestions, I just tried to speed through and didn't ponder names too much

Relates to https://github.com/hyperlane-xyz/issues/issues/701
Closes https://github.com/hyperlane-xyz/issues/issues/702 (because the balance becomes available in the metrics endpoint for polling)


rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -268,7 +300,10 @@ impl WasmProvider for WasmGrpcProvider {
     where
         T: Serialize + Send + Sync,
     {
-        self.wasm_query_to(self.contract_address.address(), payload, block_height)
+        let contract_address = self.contract_address.as_ref().ok_or_else(|| {
```
daniel-savu, line 303: IMO since the grpc client insn't really tied to an address, it makes more sense to me to make the address field optional and unwrap it like so. I find it makes for a more intuitive / readable implementation. Could also do this is an separate PR where we apply this change to the other VM providers, lmk

rust/hyperlane-base/src/metrics/agent_metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Periodically updates the metrics
+    pub async fn start_updating_on_interval(self, period: Duration) {
+        let mut interval = tokio::time::interval(period);
+        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+        loop {
+            self.update_wallet_balances().await;
+            interval.tick().await;
+        }
+    }
```
daniel-savu, line 122: I think there should be a way to refactor this, since this is mostly copy-pasta from the other `start_updating_on_interval` in the codebase. The only different bit is the call to `self.update_wallet_balances().await;`. However I wasn't able to come up with an async closure type that accepts both cases, since one calls from an Arc and the other one from a ref. Spent 2+ hours on this and feels bad to give up, but hopefully future Rust versions will simplify this.

rust/hyperlane-core/src/metrics/agent.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::{ChainResult, U256};
+
+/// Interval for querying the prometheus metrics endpoint.
+/// This should be whatever the prometheus scrape interval is
+pub const METRICS_SCRAPE_INTERVAL: Duration = Duration::from_secs(60);
+
+/// Trait to be implemented by all chain-specific agent implementations,
+/// to support gathering agent metrics.
+#[async_trait]
+pub trait AgentMetricsFetcher: Send + Sync {
```
tkporter, line 14: Curious why a new trait is preferable instead of just putting it in `HyperlaneProvider` and using that? I see `get_balance` as a general piece of functionality that may be useful outside the context of agent metrics and I could see us expanding generic agent metrics to include other things that the provider can expose, like "latest block height for chain" etc.

I also tend to think that the abstractions for the chains should be closer to "this is a slim interface to interact with this chain that's mostly agnostic to _how_ it's used" when possible rather than "these abstractions are aware of how they're being used by agents". Similar deal for hyperlane-core, I think keeping it more agnostic to the details of what agents are is probably healthy

I see that this is a supertrait of HyperlaneProvider tho which does make sense

rust/hyperlane-core/src/metrics/agent.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// Fetch the balance of the wallet address associated with the chain provider.
+    async fn get_balance(&self, address: String) -> ChainResult<U256>;
```
tkporter, line 16: Curious why String vs H256 was preferable. Strings feel unnecessarily more prone to errors and it's harder to understand what format they'll be in. I think needing to change the formatting of the ethereum signer_address to accommodate this is a good example of the fragility of using strings. Elsewhere in the codebase we use H256 as the chain-agnostic representation for addresses when the value isn't intended for human reading & needs to be formatted a certain way

Iiuc using a String results in needing to parse from String -> H160 for Ethereum which feels less nice than H256 -> H160. And for cosmos we can convert from H256 -> the address easily with `CosmosAddress::from_h256(address, &conf.get_prefix()`

rust/agents/relayer/src/relayer.rs
```diff
@@ -251,12 +264,35 @@ impl BaseAgent for Relayer {
...
+                agent_metrics_fetcher,
+            );
+
+            let fetcher_task = tokio::spawn(async move {
+                agent_metrics
+                    .start_updating_on_interval(METRICS_SCRAPE_INTERVAL)
+                    .await;
+                Ok(())
+            })
+            .instrument(info_span!("AgentMetricsFetcher"));
```
tkporter, line 294: nit: maybe `"AgentMetrics"` as fetcher seems to be used to refer to the chain trait

rust/agents/scraper/migration/bin/common.rs
```diff
@@ -16,6 +18,8 @@ pub async fn init() -> Result<DatabaseConnection, DbErr> {
         .init();
 
     let url = url();
+    let mut options: ConnectOptions = url.clone().into();
+    options.connect_timeout(Duration::from_secs(CONNECT_TIMEOUT));
```
tkporter, line 22: I don't disagree w/ this but curious what motivated this

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -1,23 +1,63 @@
...
+            domain,
+            rpc_client,
+            grpc_client,
+            canonical_asset: conf.get_canonical_asset(),
+        })
+    }
+
+    /// Get a grpc client
+    pub fn grpc(&self) -> WasmGrpcProvider {
+        self.grpc_client.clone()
```
tkporter, line 55: could this return `&WasmGrpcProvider` so we don't have to clone all the time here?

rust/ethers-prometheus/src/middleware/mod.rs
```diff
@@ -627,71 +539,14 @@ impl<M: Middleware + Send + Sync> PrometheusMiddleware<M> {
         }
         if let Some(gas_price_gwei) = gas_price_gwei {
             if let Some(london_fee) = current_block.base_fee_per_gas {
-                let gas = u256_as_scaled_f64(london_fee, 18) * 1e9;
+                let gas = u256_as_scaled_f64(london_fee.into(), 18) * 1e9;
```
tkporter, line 542: curious why the `into()` is now needed, iiuc no types have changed?

rust/hyperlane-base/src/metrics/agent_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Information about the chain this metric is for
+    pub domain: HyperlaneDomain,
+
+    /// Name of the agent the metrics are about
+    pub name: String,
+}
+
+/// Utility struct to update agent metrics
+#[derive(new)]
+pub struct AgentMetrics {
```
tkporter, line 72: The name implies to me imo that there's just a single one of these for each agent, but really it seems like this is tied to a chain that the agent is using. Wondering if a name like `AgentChainMetrics` or something would be more appropriate? open to other ideas

actually looking a bit deeper I think the naming of `Metrics` (above) / `AgentMetrics` and how they're used is confusing

e.g. `create_agent_metrics` above creates `Metrics`, but in the `from_settings` function when creating an agent, `Metrics` is often imported as `AgentMetrics` and called `agent_metrics`, but this is different from `AgentMetrics` here. i think consistent naming would help

rust/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -542,44 +551,62 @@ fn run_locally() {
...
+        );
+        return Ok(false);
+    }
+
+    let ending_relayer_balance: f64 = agent_balance_sum(relayer_metrics_port).unwrap();
+    // Ideally, make sure that the difference is >= gas_per_tx * gas_cost, set here:
+    // https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/c2288eb31734ba1f2f997e2c6ecb30176427bc2c/rust/utils/run-locally/src/cosmos/cli.rs#L55
+    // What's stopping this is that the format returned by the `uosmo` balance query is a surprisingly low number (0.000003999999995184)
+    // but then maybe the gas_per_tx is just very low - how can we check that? (maybe by simulating said tx)
+    if starting_relayer_balance <= ending_relayer_balance {
```
tkporter, line 597: would we ever expect this to not be the case if the messages have been delivered?

ahh or is the intent just to provide some kind of testing that the metrics are working, and less that the relayer was actually delivering the messages?

tkporter (commented):
Looks good - some questions about how the abstractions are set up

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
daniel-savu: I do agree removing `wasm_query` would be a good idea, but there'd be quite a few code shifts so I don't think it's worth it for now

rust/agents/scraper/migration/bin/common.rs
daniel-savu: Random failures in the pipeline, and this was a quickfix I found in https://github.com/SeaQL/sea-orm/issues/742#issuecomment-1129875130

rust/ethers-prometheus/src/middleware/mod.rs
daniel-savu: oops, leftover from me playing around

rust/utils/run-locally/src/cosmos/mod.rs
daniel-savu: yeah it's to test the metrics themselves, will clarify in a comment

rust/hyperlane-core/src/metrics/agent.rs
daniel-savu: As discussed, will stick with `String` because when the Cosmos H256 digest is reinterpreted as an address, all 32 bytes are considered (instead of just the last 20 bytes), because contract addresses are 32 bytes long. Using strings leaves it up to the bech32 encoding function to figure out whether the string is an EOA or contract address

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -220,6 +223,24 @@ impl WasmGrpcProvider {
         Ok(gas_estimate)
     }
 
+    /// Fetches balance for a given `address` and `denom`
+    pub async fn get_balance(&self, address: String, denom: String) -> ChainResult<u64> {
```
tkporter, line 227: I think we may need to make this return a U256 - it's really not clear but it seems like the `Coin.amount` spec allows it to be up to 256 bits

e.g. the Coin struct which mentions `Int`:
```
// @generated
/// Coin defines a token with a denomination and an amount.
///
/// NOTE: The amount field is an Int which implements the custom method
/// signatures required by gogoproto.
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Coin {
    #[prost(string, tag = "1")]
    pub denom: ::prost::alloc::string::String,
    #[prost(string, tag = "2")]
    pub amount: ::prost::alloc::string::String,
}
```

and if you look in the SDK definition for Coin https://github.com/cosmos/cosmos-sdk/blob/main/types/coin.go#L19 it refers to sdk.Int

Which can be found here and has 256 bits
https://github.com/cosmos/cosmos-sdk/blob/main/math/int.go#L15

rust/hyperlane-base/src/metrics/agent_metrics.rs
```diff
@@ -0,0 +1,120 @@
...
+            return;
+        };
+        let chain = self.conf.domain.name();
+
+        match self.provider.get_balance(wallet_addr.clone()).await {
+            Ok(balance) => {
+                // Okay, so the native type is not a token, but whatever, close enough.
+                // Note: This is ETH for many chains, but not all so that is why we use `N` and `Native`
+                // TODO: can we get away with scaling as 18 in all cases here? I am guessing not.
+                let balance = u256_as_scaled_f64(balance, 18);
```
tkporter, line 95: for posterity flagging this

tkporter (approved):

rust/hyperlane-base/src/metrics/agent_metrics.rs
daniel-savu: nice one, fixed by hardcoding decimals by protocol type for now

PR #2983, opened 2023-11-27
fix:CLI won't recover IGP from SDK

aroralanuk:
### Description

- fix to filter out igp in merging the given and SDK artifacts

### Drive-by changes

- None

### Related issues

- None

### Backward compatibility

Yes

### Testing

Manual


jmrossy:
@aroralanuk lmk when you've re-based this and I'll have a look :)

typescript/cli/src/context.ts
```diff
@@ -22,7 +22,16 @@ export function getMergedContractAddresses(
   artifacts?: HyperlaneContractsMap<any>,
 ) {
   return objMerge(
-    sdkContractAddressesMap,
+    // filter out interchainGasPaymaster since we don't want to recover it from SDK artifacts
```
jmrossy, line 25: This was probably addressed somewhere in the flurry of discord messages but why do we do this?

aroralanuk: Say you're deploying between rossychain and arbitrumgoerli. We recover the igp here on arbitrumgoerli and forgo deployment of igp and subsequently the aggregationHook whereas we should deploy a new igp in the case where at least one of the chains is non-core.

Note: we should probably use a FallbackIgp here instead of the regular IGP.

typescript/cli/src/context.ts
```diff
@@ -3,26 +3,38 @@ import { ethers } from 'ethers';
...
   ...hyperlaneEnvironments.mainnet,
 };
 
 export function getMergedContractAddresses(
   artifacts?: HyperlaneContractsMap<any>,
+  chains?: ChainName[],
 ) {
+  // if chains include non sdkContractAddressesMap chains, don't recover interchainGasPaymaster
+  let sdkContractsAddressesToRecover = sdkContractAddressesMap;
+  if (chains?.some((chain) => !(chain in sdkContractAddressesMap))) {
```
jmrossy, line 28: Nit: the `in` keyword has some undesirable pitfalls. It's recommended to use `hasOwnProperty` or `Object.keys` instead: https://www.quora.com/Why-is-extending-built-in-JavaScript-types-evil

jmrossy (approved):

Issue #1742, opened 2023-02-02
CI should show storage layout changes in PRs


nambrot:
looks like foundry supports showing diffs against snapshots for the storage layout of the contracts to force developers to explicitly acknowledge

vinhyenvodoi98:
the target of bounty is this PR right ?

nambrot:
Not quite, this bounty talks about snapshotting and showing storage layout changes

yorhodes:
example https://github.com/morpho-org/morpho-optimizers/pull/1150/

PR #2991, opened 2023-11-28
Add storage layout diff checks

yorhodes:
### Description

- Add `yarn workspace @hyperlane-xyz/core storage` script
- Add `storage-analysis` workflow for comparing diff on PRs between base and HEAD

### Drive-By

- Rename `StaticProtocolFee` to `ProtocolFee` to reflect actual behavior and include in storage layout checks

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/1742

### Backward compatibility

Yes

### Testing

CI


.github/workflows/storage-analysis.yml
```diff
@@ -0,0 +1,66 @@
...
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha }}
+        run: |
+          git fetch origin $BASE_REF
+          git checkout $BASE_REF -- solidity/contracts
+
+      # Run the command on the target branch
+      - name: Run command on target branch
+        run: yarn workspace @hyperlane-xyz/core storage base-storage
+
```
aroralanuk, line 55: It comments an empty file even if we don't change anything.
I'm not sure if the env variables are preserved during separate steps but something like this will be preferred:

```suggestion
- name: Check if diff is non-empty
  env:
    DIFF_EXISTS: false
  id: check_diff
  run: |
    if [ -s layout.diff ]
    then
      export DIFF_EXISTS=true
      
 - name: Comment PR with layout diff
  if: DIFF_EXISTS == 'true'
  uses: yorhodes/actions-comment-pull-request@v2.4.5
  with:
    header: Storage Layout Diff
    filePath: layout.diff
    mdLanguage: diff
    comment_tag: storagelayoutdiff
```

solidity/storage.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+OUTPUT_PATH=${1:-storage}
+EXCLUDE="test|mock|interfaces|libs|upgrade|README|Abstract|Static"
```
aroralanuk, line 3: we don't want StaticMultisigIsm, aggregationIsm, etc - they have a moduleType in storage?

Plus, I named it StaticProtocolFee which is needs the storage layout.

.github/workflows/storage-analysis.yml
yorhodes: https://github.com/yorhodes/actions-comment-pull-request/compare/v2.4.5...v2.4.6
added support in the action itself to keep usage simple

solidity/storage.sh
yorhodes: `moduleType` not in storage
`Static` is the name we have used for constants in bytecode
will rename `StaticProtocolFee` so we capture the storage there

.github/workflows/storage-analysis.yml
yorhodes: it also deleted the existing comment automatically which was unexpected but desirable :)

solidity/storage.sh
yorhodes: done in 1141ce513d96e4847d5c7f4b65040da0c59d40e6

solidity/test/hooks/StaticProtocolFee.t.sol
aroralanuk, line 1: can you rename this to ProtocolFee too?

aroralanuk (approved):

PR #2994, opened 2023-11-28
Improve warp and kurtosis deploy command dx

jmrossy:
### Description

Add artifact selection step to warp deploy
Improve chain selection for kurtosis agent deploy

### Drive-by changes

Fix chain metadata error and add unit test

### Related issues

Fixes https://github.com/hyperlane-xyz/issues/issues/774

### Backward compatibility

Small break in kurtosis deploy args, likely won't be noticed.

### Testing

Manual

### DIFF
--- a/typescript/cli/src/commands/deploy.ts
+++ b/typescript/cli/src/commands/deploy.ts
@@ -38,26 +38,29 @@ const agentCommand: CommandModule = {
   describe: 'Deploy Hyperlane agents with Kurtosis',
   builder: (yargs) =>
     yargs.options({
-      originChain: {
+      origin: {
         type: 'string',
         description: 'The name of the origin chain to deploy to',
       },
-      agentConfiguration: agentConfigurationOption,
-      relayChains: {
+      targets: {
         type: 'string',
         description: 'Comma separated list of chains to relay between',
       },
+      chains: chainsCommandOption,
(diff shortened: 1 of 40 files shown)


typescript/cli/src/commands/options.ts
```diff
@@ -25,13 +25,12 @@ export const outDirCommandOption: Options = {
 export const coreArtifactsOption: Options = {
   type: 'string',
   description: 'File path to core deployment output artifacts',
-  alias: 'ca',
+  alias: 'a',
```
jmrossy, line 28: The single letter is better b.c. it prints nicer help text (and is a more common convention)

typescript/cli/src/commands/deploy.ts
```diff
@@ -38,26 +38,29 @@ const agentCommand: CommandModule = {
   describe: 'Deploy Hyperlane agents with Kurtosis',
   builder: (yargs) =>
     yargs.options({
-      originChain: {
+      origin: {
         type: 'string',
         description: 'The name of the origin chain to deploy to',
       },
-      agentConfiguration: agentConfigurationOption,
```
jmrossy, line 45: Kurtosis should not have used camel case names for the cli args here

nambrot (approved):

PR #2996, opened 2023-11-28
Update gas oracle values

nambrot:
### Description

Updates the values given the current market conditions

### Drive-by changes

Add a sleep for avoid getting rate limited


jmrossy (approved):

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -136,6 +136,8 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
     const ids = chains.map(
       (chain) => this.metadata[chain].gasCurrencyCoinGeckoId || chain,
     );
+    // Coingecko rate limits, so we are adding this sleep
+    await sleep(5000);
```
jmrossy, line 140: I believe this is causing the test to break due to timeout

nambrot: Fixed it by allowing configuration

PR #3001, opened 2023-11-29
Fix Kathy sending from PolygonZkEvm & require MultiProtocolProvider chains to match addresses in MultiProtocolApp

tkporter:
### Description

Fixes 2 issues:
1. Estimates gas in Kathy by explicitly specifying the `from` address due to this bug with PolygonZkEvm when using a non-zero value https://github.com/0xPolygonHermez/zkevm-node/issues/2869
2. Now that we've added neutron & mantapacific as mainnet chains but we don't have helloworld deployments on these chains, Kathy was trying to send to & from these chains. To fix, I changed the constructor of `MultiProtocolApp` to intersect the multiProvider to only work with the chains specified in `addresses`

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

ye

### Testing

Ran kathy locally successfully & sent from polygonzkevm

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "clean": "yarn workspaces foreach --all --parallel run clean",
     "prettier": "yarn workspaces foreach --since --parallel run prettier",
     "lint": "yarn workspaces foreach --since --parallel run lint",
-    "test": "yarn workspaces foreach --since --parallel run test",
+    "test": "yarn workspaces foreach --all --parallel run test",
     "coverage": "yarn workspaces foreach --since --parallel run coverage",
     "version:prepare": "yarn changeset version && yarn workspaces foreach --all --parallel run version:update && yarn install --no-immutable",
     "version:check": "yarn changeset status",
--- a/typescript/infra/config/environments/mainnet3/agent.ts
+++ b/typescript/infra/config/environments/mainnet3/agent.ts
@@ -43,14 +43,14 @@ const hyperlane: RootAgentConfig = {
     rpcConsensusType: RpcConsensusType.Fallback,
     docker: {
       repo,
-      tag: 'c2288eb-20231129-115257',
+      tag: 'f44589e-20231130-114734',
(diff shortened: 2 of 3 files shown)


typescript/helloworld/src/multiProtocolApp/evmAdapter.ts
```diff
@@ -29,6 +29,7 @@ export class EvmHelloWorldAdapter
     destination: ChainName,
     message: string,
     value: string,
+    sender: Address,
```
tkporter, line 32: I'm a bit surprised typescript didn't complain about this not being specified before.. but this is from IHelloWorldAdapter

jmrossy: Doesn't ethers.js automatically add this field just before it signs the tx?

typescript/sdk/src/app/MultiProtocolApp.ts
```diff
@@ -120,6 +120,7 @@ export abstract class MultiProtocolApp<
     public readonly addresses: ChainMap<ContractAddrs>,
     public readonly logger = debug('hyperlane:MultiProtocolApp'),
   ) {
+    multiProvider = multiProvider.intersect(Object.keys(addresses)).result;
```
jmrossy, line 123: I'm not strongly opposed but I don't think will match users intuition. Like say if they grab the MPP from an MPA later, I think they'd expect it to be the same one they initially gave. 

An alternative could be to add validation here to throw if the addresses map chains doesn't match the MPP chains. It would require us to intersect the MPP in Kathy here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/rossy/cli-warp-deploy-artifact-selection/typescript/infra/scripts/helloworld/utils.ts#L131

If you prefer this approach, I'm okay with it

jmrossy (approved):

PR #3005, opened 2023-11-30
fix:remove `agentStartBlock` and use `mailbox.deployedBlock()` instead

aroralanuk:
### Description

- CLI was using a static start block numbers for core chains or getting the latest block number from a PI chain for agent config which is redundant and dangerous. Instead, I updated to using the `mailbox.deployedBlock()` which should precede all other indexable contract deployment and is hence safer.

### Drive-by changes

- filtering the agent configs in CLI is redundant since the `HyperlaneDeploymentArtifactsSchema` requires all the specified entries and if `writeAgentConfig` gets an artifacts which doesn't contain these, we should throw an error and not filter them.

### Related issues

- related to https://github.com/hyperlane-xyz/issues/issues/736

### Backward compatibility

Yes

### Testing

Manual b/w anvil1 and anvil2


typescript/infra/src/deployment/deploy.ts
```diff
@@ -126,14 +127,19 @@ export async function writeAgentConfig(
   } catch (e) {
     console.error('Failed to load cached addresses');
   }
-  // Write agent config indexing from the deployed or latest block numbers.
-  // For non-net-new deployments, these changes will need to be
-  // reverted manually.
```
yorhodes, line 131: nice

yorhodes (commented):
nice cleanup :)

PR #3008, opened 2023-12-01
Add figlet to CLI

yorhodes:
### Description

Adds fun figlet command to CLI :)


typescript/cli/cli.ts
```diff
@@ -13,7 +13,11 @@ import { VERSION } from './src/version.js';
 // From yargs code:
 const MISSING_PARAMS_ERROR = 'Not enough non-option arguments';
 
-console.log(chalk.blue('Hyperlane'), chalk.magentaBright('CLI'));
+console.log(
+  figlet.textSync('Hyperlane', {
+    font: 'ANSI Shadow',
```
aroralanuk, line 18: 🚀🚀🚀

aroralanuk (approved):

jmrossy (approved):
Given that the CLI handles keys, we should probably keep the dep tree minimal. That said, the figlet code looks legit and the solana + cosmos libs we require are almost certainly an insecure mess anyway... so 🤷

PR #3012, opened 2023-12-01
Fix base bridging and a race condition in key funder

tkporter:
### Description

Turns out bridging from L1 to Base has been broken the whole time :( it's because the version of @eth-optimism/sdk we were using was 1.8.0 (according to yarn.lock) which was released in late 2022 https://github.com/ethereum-optimism/optimism/releases/tag/%40eth-optimism%2Fsdk%401.8.0, far before Base was launched. So we'd always get this error in key funder because we rely on the optimism SDK knowing about Base's chain ID:
```
{"chain":"base","error":"Error: cannot get contract AddressManager for unknown L2 chain ID 8453, you must provide an address\n    at getOEContract (/hyperlane-monorepo/node_modules/@eth-optimism/sdk/src/utils/contracts.ts:58:11)\n    at getAllOEContracts (/hyperlane-monorepo/node_modules/@eth-optimism/sdk/src/utils/contracts.ts:121:46)\n    at new CrossChainMessenger (/hyperlane-monorepo/node_modules/@eth-optimism/sdk/src/cross-chain-messenger.ts:170:39)\n    at ContextFunder.bridgeToOptimism (/hyperlane-monorepo/typescript/infra/scripts/funding/fund-keys-from-deployer.ts:652:33)\n    at ContextFunder.bridgeToL2 (/hyperlane-monorepo/typescript/infra/scripts/funding/fund-keys-from-deployer.ts:637:23)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async ContextFunder.bridgeIfL2 (/hyperlane-monorepo/typescript/infra/scripts/funding/fund-keys-from-deployer.ts:492:9)\n    at async gracefullyHandleError (/hyperlane-monorepo/typescript/infra/scripts/funding/fund-keys-from-deployer.ts:774:5)\n    at async /hyperlane-monorepo/typescript/infra/scripts/funding/fund-keys-from-deployer.ts:394:29\n    at async Promise.all (index 9)","level":"error","message":"Error bridging to L2"}
```

A mystery to me is that we'd get that log ^ but sometimes the key funder pod would show as having ran successfully. My best guess here is there was a race condition in `fund` when we had a variable local to the function called `failureOccurred` that many promises which were `Promise.all`'d would read and write to it via `failureOccurred ||= await someFallibleFn()`. I changed the logic here to not have multiple concurrent promises contend for the variable

Also made a small change to not include mantapacific in the list of relayer keys for the Hyperlane context. It's not ever used, and had a downstream effect of us trying to fund the Kathy key on mantapacific, which we don't want to actually do

### Drive-by changes

n/a

### Related issues

n/a

### Backward compatibility

ye

### Testing

ran locally

### DIFF
--- a/typescript/infra/config/environments/mainnet3/chains.ts
+++ b/typescript/infra/config/environments/mainnet3/chains.ts
@@ -65,14 +65,17 @@ export const ethereumChainNames = Object.keys(
   ethereumMainnetConfigs,
 ) as MainnetChains[];
 
+// Remove mantapacific, as it's not considered a "blessed"
+// chain. It's not included in the scraper domains table,
+// and we don't relay to mantapacific on the Hyperlane or RC contexts.
+const hyperlaneContextRelayChains = ethereumChainNames.filter(
+  (chainName) => chainName !== chainMetadata.mantapacific.name,
+);
+
 // Hyperlane & RC context agent chain names.
 export const agentChainNames: AgentChainNames = {
   // Run validators for all chains.
   [Role.Validator]: supportedChainNames,
-  // Only run relayers for Ethereum chains at the moment.
-  [Role.Relayer]: ethereumChainNames,
-  // Remove mantapacific for now, as it's not included in the scraper domains table
-  [Role.Scraper]: ethereumChainNames.filter(
-    (chainName) => chainName !== chainMetadata.mantapacific.name,
-  ),
+  [Role.Relayer]: hyperlaneContextRelayChains,
+  [Role.Scraper]: hyperlaneContextRelayChains,
 };
--- a/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
+++ b/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
@@ -395,10 +395,9 @@ class ContextFunder {
   // Funds all the roles in this.rolesToFund
   // Returns whether a failure occurred.
   async fund(): Promise<boolean> {
-    let failureOccurred = false;
-
     const chainKeys = this.getChainKeys();
     const promises = Object.entries(chainKeys).map(async ([chain, keys]) => {
+      let failureOccurred = false;
       if (keys.length > 0) {
         if (!this.skipIgpClaim) {
           failureOccurred ||= await gracefullyHandleError(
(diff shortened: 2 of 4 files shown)


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -418,10 +417,14 @@ class ContextFunder {
...
       }
+      return failureOccurred;
     });
 
+    let failureOccurred = false;
     try {
-      await Promise.all(promises);
+      failureOccurred = (await Promise.all(promises)).some(
+        (promiseFailure) => promiseFailure,
+      );
```
yorhodes, line 427: isnt this inverted?

yorhodes (approved):

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -65,14 +65,17 @@ export const ethereumChainNames = Object.keys(
   ethereumMainnetConfigs,
 ) as MainnetChains[];
 
+// Remove mantapacific, as it's not considered a "blessed"
+// chain. It's not included in the scraper domains table,
+// and we don't relay to mantapacific on the Hyperlane or RC contexts.
+const hyperlaneContextRelayChains = ethereumChainNames.filter(
```
aroralanuk, line 71: should isRelayChain be a property in chainMetadata which you use to filter by? We might have more of these non-blessed chains in the future and it will quickly become cumbersome to maintain?

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
aroralanuk: I don't think so, this just checks for any failures in the promises

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -418,10 +417,14 @@ class ContextFunder {
...
     } catch (e) {
```
aroralanuk, line 428: this catch doesn't work for all the failureOccurred, only the ones caught in "Promise.all(promises)).some". Failures on line 403 or 418 just get returned as boolean. Can we move all of this into a fallible functions into try catch so we can catch there here?
Otherwise, the "'At least one failure occurred when funding'" message on line 287 isn't super helpful esp with multiple contexts.

aroralanuk:
Thanks for spotting the base funder issue!

typescript/infra/config/environments/mainnet3/chains.ts
tkporter: Good point, I agree that this distinction is super confusing. I don't putting it in the SDK is the right move though as this is is specific to our infrastructure setup. I think the best solution here is to just put some love into supporting HaaS more appropriately in infra, atm it's a bit of a mess

aroralanuk: fair point, pointing it out as I have ran into similar issues bc our adhoc inclusion-exclusion of different chains in different parts of the codebase

PR #3021, opened 2023-12-05
fix:don't prompt user for multisig config if chain is core

aroralanuk:
### Description

- don't restrict user to having two chains for ism config
- if the user accidentally picks two chains, we prompt them again to confirm if they don't want to use the hyperlane validators for their multisigconfig

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/811

### Backward compatibility

Yes

### Testing

Manual between arbitrumgoerli and anvil1

### DIFF
--- a/.changeset/ninety-seas-chew.md
+++ b/.changeset/ninety-seas-chew.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
+---
+
+Allow users to config their own PI chain
--- a/.changeset/ninety-seas-chew.md
+++ b/.changeset/ninety-seas-chew.md
@@ -2,4 +2,7 @@
 '@hyperlane-xyz/cli': patch
 ---
 
-Allow users to config their own PI chain
+Allow users to only configure validators for their chain
+
+- Don't restrict user to having two chains for ism config
+- If the user accidentally picks two chains, we prompt them again to confirm if they don't want to use the hyperlane validators for their multisigConfig
--- a/typescript/cli/src/config/multisig.ts
+++ b/typescript/cli/src/config/multisig.ts
@@ -1,11 +1,7 @@
 import { confirm, input } from '@inquirer/prompts';
 import { z } from 'zod';
 
-import {
-  ChainMap,
-  MultisigConfig,
-  defaultMultisigConfigs,
-} from '@hyperlane-xyz/sdk';
+import { ChainMap, MultisigConfig } from '@hyperlane-xyz/sdk';
 import {
   Address,
   isValidAddress,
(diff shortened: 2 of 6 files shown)


typescript/cli/src/config/multisig.ts
```diff
@@ -84,6 +92,19 @@ export async function createMultisigConfig({
       result[chain] = lastConfig;
       continue;
     }
+    if (Object.keys(sdkContractAddressesMap).includes(chain)) {
+      const reuseCoreConfig = await confirm({
+        message: 'Use existing Hyperlane validators for this chain?',
+      });
+      if (reuseCoreConfig) {
+        result[chain] = {
```
nambrot, line 100: Should we just leave it out of `result` in this case? We can still merge it during the deployment step, but allows the SDK to be the source of truth

typescript/cli/src/config/multisig.ts
```diff
@@ -84,6 +92,19 @@ export async function createMultisigConfig({
...
+          threshold: defaultMultisigConfigs[chain].threshold,
+          validators: defaultMultisigConfigs[chain].validators,
+        };
+        console.log(result[chain]);
```
nambrot, line 104: Remove

typescript/cli/src/utils/chains.ts
```diff
@@ -33,9 +32,9 @@ export async function runSingleChainSelectionStep(
 export async function runMultiChainSelectionStep(
   customChains: ChainMap<ChainMetadata>,
   message = 'Select chains',
-  chainsToFilterOut: ChainName[] = [],
+  chainSelection = false,
```
nambrot, line 35: chainSelection is not descriptive really. How about `requireMultiple` or something like that? Also, wha

typescript/cli/src/utils/chains.ts
```diff
@@ -44,19 +43,14 @@ export async function runMultiChainSelectionStep(
       pageSize: 20,
     })) as string[];
     handleNewChain(chains);
-    if (chains?.length >= 2) return chains;
+    if (chains?.length >= 2 || !chainSelection) return chains;
```
nambrot, line 46: Nit: I personally like reading these conditions in the reverse, basically `if(badCondition) error`

and then just return chains if all none of the bad conditions triggered

.changeset/ninety-seas-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
+---
+
+Allow users to config their own PI chain
```
nambrot, line 5: More descriptive message here please. Something like `Allow users to only configure validators for their chain`

typescript/cli/src/config/multisig.ts
aroralanuk: Makes sense. I was thinking if the user selected the chain, they'll expect to see the config in their yaml file. But. I'm don't feel strongly about either approach.

typescript/cli/src/utils/chains.ts
aroralanuk: 1) What

nambrot: sorry mid-sentence i just asked you in person what `chainsToFilterOut` was used for

nambrot (approved):

PR #3022, opened 2023-12-05
Add ASCII art to README

aroralanuk:


tkporter:
I think it renders as markdown in a weird way, at least that's what the rich diff shows https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3022/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

maybe a code block is needed or something?

PR #3023, opened 2023-12-05
Refactor key management in infra and key funder

tkporter:
### Description

* Remove the `bank` role, which we haven't used since the inception of abacus / hyperlane
* Big changes to `key-utils.ts` so that there's a single source of truth on what kind of keys are used depending on the role & chain. Before this was sprinkled in a few different places
* You can now get an object of `{ [chain]: { [role]: keys[] } }`, so it's super clear what kind of key relates to which chain. For example, before we would use the AWS-based relayer key for EVM chains, and then a GCP-based relayer key for non-EVM chains. But this wasn't really honored by key funder - it had no way of knowing to only fund the AWS relayer on EVM chains, and only fund the GCP relayer on non-EVM chains. Same situation for Kathy, where we want to use AWS keys for EVM chains but the GCP key for non-EVM chains
* On v2 and prior to that we were using the AWS-based key for Kathy. Originally, we also launched v3 this way. However it was changed on v3 to use the GCP key for Kathy, causing us to fund both types of addresses on v3. This makes it more clear that we should be using the AWS-based key for Kathy on EVM chains


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -305,66 +313,86 @@ class ContextFunder {
...
+        agg[idAndAddress.identifier] = idAndAddress;
+        return agg;
+      }, {} as Record<string, KeyAsAddress>);
+
+    const agentConfig = getAgentConfig(context, environment);
+    // Unfetched keys per chain and role, so we know which keys
+    // we need. We'll use this to create a corresponding object
+    // of ReadOnlyCloudAgentKeys using addresses found in the
+    // serialized address file.
+    const roleKeysPerChain = getRoleKeysPerChain(agentConfig);
```
nambrot, line 362: It's probably been a while since I looked at this logic, but if we are just fetching the keys via context and environment, then why are we reading from the filePath again?

tkporter: `getRoleKeysPerChain` doesn't actually fetch any secrets from GCP, so we don't actually know what the addresses are of those keys (which you need the private key to know for a GCP key, or AWS credentials for the AWS keys). The reading from the filePath is so that we can just load in all the addresses without knowing any private keys or anything

nambrot: ahhh right

nambrot (approved):

PR #3025, opened 2023-12-05
Sealevel balance metrics

daniel-savu:
Publishes sealevel balance as a relayer metric

**Drive-by Changes**
- Creates `HyperlaneSealevelError`, a `hyperlane-sealevel` specific type to avoid including sealevel errors in `hyperlane-core`'s `ChainCommunicationError`. It's pretty empty now, but we have a starting point now. This follows the approach used in `hyperlane-cosmos` too.
- Includes an rpc provider instance in `SealevelProvider`, and uses it to add `get_balance` on `SealevelProvider`.

**Testing**
None, as sealevel e2e is currently disabled

### DIFF
--- a/rust/chains/hyperlane-sealevel/src/provider.rs
+++ b/rust/chains/hyperlane-sealevel/src/provider.rs
@@ -38,10 +38,10 @@ impl SealevelProvider {
         let pubkey = Pubkey::from_str(&address).map_err(Into::<HyperlaneSealevelError>::into)?;
         let balance = self
             .rpc_client
-            .get_account(&pubkey)
+            .get_balance(&pubkey)
             .await
             .map_err(Into::<HyperlaneSealevelError>::into)?;
-        Ok(balance.lamports.into())
+        Ok(balance.into())
     }
 }
 
@@ -73,7 +73,7 @@ impl HyperlaneProvider for SealevelProvider {
         Ok(true)
     }
 
-    async fn get_balance(&self, _address: String) -> ChainResult<U256> {
-        todo!() // FIXME
+    async fn get_balance(&self, address: String) -> ChainResult<U256> {
+        self.get_balance(address).await
     }
 }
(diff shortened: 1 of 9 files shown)


rust/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -1,19 +1,47 @@
...
+    pub fn rpc(&self) -> &RpcClientWithDebug {
+        &self.rpc_client
+    }
+
+    /// Get the balance of an address
+    pub async fn get_balance(&self, address: String) -> ChainResult<U256> {
+        let pubkey = Pubkey::from_str(&address).map_err(Into::<HyperlaneSealevelError>::into)?;
+        let balance = self
+            .rpc_client
+            .get_account(&pubkey)
```
tkporter, line 41: May be better to use `get_balance`, `get_account` has two small downsides:
1. it returns any account data, so if this is a program we're also fetching all the code which may be pretty large
2. very annoyingly if the account doesn't exist (defined by there being no data or a balance of 0 in for the account iirc), the RPC client returns an `Err` :|. So if someone starts up their relayer or something without having funded their address yet, I think we'd get some errors when i think we'd prefer this function returning `Ok(U256::zero())`

rust/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -1,19 +1,47 @@
...
+        SealevelProvider { domain, rpc_client }
+    }
+
+    /// Get an rpc client
+    pub fn rpc(&self) -> &RpcClientWithDebug {
+        &self.rpc_client
+    }
+
+    /// Get the balance of an address
+    pub async fn get_balance(&self, address: String) -> ChainResult<U256> {
```
tkporter, line 37: I think you need to update the HyperlaneProvider::get_balance function on line 77 to call this, or to just move this fn impl to line 77

tkporter (approved):

PR #3026, opened 2023-12-05
chore: reduce agents log verbosity, reword

daniel-savu:
There are way too many `No message found in DB for leaf index` logs to have them at debug level. Also the wording makes it seem like an error when it's not.

### DIFF
--- a/rust/agents/relayer/src/merkle_tree/processor.rs
+++ b/rust/agents/relayer/src/merkle_tree/processor.rs
@@ -76,7 +76,7 @@ impl MerkleTreeProcessor {
                 .set(insertion.index() as i64);
             Some(insertion)
         } else {
-            trace!(leaf_index=?self.leaf_index, "No more messages found in DB for leaf index. Done processing.");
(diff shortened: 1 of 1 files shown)


rust/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -76,7 +76,7 @@ impl MerkleTreeProcessor {
                 .set(insertion.index() as i64);
             Some(insertion)
         } else {
-            debug!(leaf_index=?self.leaf_index, "No message found in DB for leaf index");
+            trace!(leaf_index=?self.leaf_index, "No more messages found in DB for leaf index. Done processing.");
```
tkporter, line 79: nit: this kinda reads to me like there are multiple messages for a single leaf index and that the task is done, maybe something like "No merkle tree insertion found in DB for leaf index, waiting for it to be indexed" or something?

tkporter (approved):
Love this change, this log is so annoying

rust/agents/relayer/src/merkle_tree/processor.rs
daniel-savu: ah yes good point

PR #3029, opened 2023-12-05
Manage core governance

nambrot:
### Description

Changes that were required to run the deployer/check to enroll more validators. Some of the changes like removing artifacts from the SDK or changing the chains in infra probably shouldn't be checked in.

### Drive-by changes

- Added another error message for mailbox initialization check (since there is no view call to make)
- Ignoring validator announce violations in the governor

### Related issues

Fixes https://github.com/hyperlane-xyz/issues/issues/673


typescript/sdk/src/consts/environments/mainnet.json
```diff
@@ -178,21 +178,5 @@
     "protocolFee": "0xEc4AdA26E51f2685279F37C8aE62BeAd8212D597",
     "mailbox": "0xFf06aFcaABaDDd1fb08371f9ccA15D73D51FeBD6",
     "validatorAnnounce": "0x9Cad0eC82328CEE2386Ec14a12E81d070a27712f"
-  },
```
nambrot, line 181: Probably shouldn't remove this here?

yorhodes: i don't feel strongly
I doubt others are consuming this now

typescript/sdk/src/core/HyperlaneCoreDeployer.ts
```diff
@@ -106,7 +106,10 @@ export class HyperlaneCoreDeployer extends HyperlaneDeployer<
         ),
       );
     } catch (e: any) {
-      if (!e.message.includes('already initialized')) {
+      if (
+        !e.message.includes('already initialized') &&
+        !e.message.includes('Reverted 0x08c379a')
```
yorhodes, line 111: what is this? add a comment

yorhodes (approved):

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -59,6 +59,12 @@ export type MainnetChains = keyof typeof mainnetConfigs;
 export const supportedChainNames = Object.keys(
   mainnetConfigs,
 ) as MainnetChains[];
+
+export const supportAgentChainNames = Object.keys({
+  ...mainnetConfigs,
+  neutron: chainMetadata.neutron,
+  mantaPacific: chainMetadata.mantapacific,
```
tkporter, line 66: nit: this should be all lowercase to match the name of the chain metadata

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -59,6 +59,12 @@ export type MainnetChains = keyof typeof mainnetConfigs;
 export const supportedChainNames = Object.keys(
   mainnetConfigs,
 ) as MainnetChains[];
+
+export const supportAgentChainNames = Object.keys({
```
tkporter, line 63: nit: `supportAgentChainNames`

tkporter, line 63: should this be used down below in `agentChainNames`? I don't see this used at all

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -13,10 +13,9 @@ export const safes: ChainMap<Address> = {
   gnosis: '0x36b0AA0e7d04e7b825D7E409FEa3c9A3d57E4C22',
   // solana: 'EzppBFV2taxWw8kEjxNYvby6q7W1biJEqwP3iC7YgRe3',
   // TODO: create gnosis safes here
-  base: '',
-  scroll: '',
-  polygonzkevm: '',
-  mantapacific: '',
+  base: '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba',
```
tkporter, line 16: wen safes

tkporter (approved):
btw I wouldn't be surprised if this has an impact on things like deploying agents, key funder, or kathy - could you deploy to those with this change to confirm all is fine

typescript/infra/config/environments/mainnet3/chains.ts
nambrot: Yeah so it kinda isn't used, because I think agentChainNames IMO is only for hyperlane and rc contexts, and the neutron context manualy specifies the chains? I think this is another example where we have to do the improving by removing thing, since its very hard to know what impact removing neutron from nonEthereumMainnetConfigs has

typescript/sdk/src/consts/environments/mainnet.json
nambrot: Fine to do so, i do think to use our tooling like CoreDeployer or CoreChecker (that we maybe want to use for ISM rotation) would rely on this right now)

nambrot: Also Kathy

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -59,6 +59,12 @@ export type MainnetChains = keyof typeof mainnetConfigs;
 export const supportedChainNames = Object.keys(
   mainnetConfigs,
 ) as MainnetChains[];
+
```
jmrossy, line 62: `export const supportedAgentChainNames = [...supportedChainNames, Chains.Neutron, Chains.mantapacific]`

typescript/sdk/src/core/HyperlaneCoreDeployer.ts
```diff
@@ -106,7 +106,11 @@ export class HyperlaneCoreDeployer extends HyperlaneDeployer<
         ),
       );
     } catch (e: any) {
-      if (!e.message.includes('already initialized')) {
+      if (
+        !e.message.includes('already initialized') &&
+        // Some RPC providers dont return the revert reason (nor allow ethers to parse it), so we have to check the message
+        !e.message.includes('Reverted 0x08c379a')
```
jmrossy, line 112: Is this an error message specific to core deployments or something we would want to catch for all providers?

typescript/sdk/src/consts/environments/mainnet.json
jmrossy: It's helpful for the explorer to include the artifacts in the SDK.

typescript/sdk/src/core/HyperlaneCoreDeployer.ts
nambrot: Just core deployments

typescript/infra/config/environments/mainnet3/chains.ts
nambrot: Undid the change

typescript/sdk/src/consts/environments/mainnet.json
nambrot: Undid the removal

PR #3042, opened 2023-12-08
Cosmos gas price config

daniel-savu:
### Description

Adds a cosmos-specific config item for setting the minimum gas price, in the format returned by the `cosmos.base.node.v1beta1.Service/Config` cosmos-sdk grpc endpoint.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/810

### Backward compatibility

No. Will break existing cosmos configs.

### Testing

None yet, will test in e2e


rust/chains/hyperlane-cosmos/src/trait_builder.rs
```diff
@@ -11,6 +16,46 @@ pub struct ConnectionConf {
...
+    pub amount: U256,
+}
+
+impl TryFrom<RawCosmosAmount> for CosmosAmount {
+    type Error = ChainCommunicationError;
+    fn try_from(raw: RawCosmosAmount) -> Result<Self, ChainCommunicationError> {
+        // Converts to U256 by always rounding up.
+
+        // Remove the decimal part
+        let integer = raw
```
tkporter, line 47: Ah I realize now why you mentioned the rounding by removing the decimal part and adding 1, for some reason I didn't connect the dots on implications

It looks like the suggested Osmosis gas price is [`0.0025`](https://github.com/cosmos/chain-registry/blob/master/osmosis/chain.json#L22), which iiuc as written we would result in using a gas price of 1, so we'd be paying 400x for each tx than we need to. Similar for Neutron but not as bad as it's [`0.5`](https://github.com/cosmos/chain-registry/blob/master/neutron/chain.json#L20), so we'd be paying 2x

So I think we'll need to have the amount be able to be less than 1. Either as an f64 or maybe by specifying a numerator and denominator like how we do for `GasPaymentEnforcementPolicy::OnChainFeeQuoting`

I know for TxOutcome / TxCostEstimate we use `gas_price: U256` - I see our options here as:
1. just keep it as U256 and do something similar to what we do here where we ceil the f64 gas price for cosmos chains
2. we change TxOutcome to support < 1 gas prices, either via floats or int numerator / denominator, and then change the storage key for TxOutcomes in hyperlane db for these because this'll change the encoding of existing data. We could maybe even have an enum with a U256 variant and a f64 variant or something, but not sure what that'd give us tbh

I think 2 is probably the nicer solution longer

rust/chains/hyperlane-cosmos/src/trait_builder.rs
```diff
@@ -11,6 +16,46 @@ pub struct ConnectionConf {
     prefix: String,
     /// Canoncial Assets Denom
     canonical_asset: String,
+    /// The minimum gas price set by the cosmos-sdk validator
+    minimum_gas_price: RawCosmosAmount,
```
tkporter, line 20: I think saying "minimum" here may imply that the agent would maybe use a higher gas price and that this value isn't exactly what will be used for txs - I wonder if we should call this `gas_price` but still call out in the documentation that this is likely to just be the minimum gas price set by the node we're interactive with

rust/chains/hyperlane-cosmos/src/trait_builder.rs
daniel-savu: TIL gas price is not something you're reimbursed for :/

Rather than have custom logic for the decimal part, I'm inclined to go for a crate that stores floats on the heap but abstracts the complexity away and can be arbitrarily large
https://crates.io/crates/bigdecimal

rust/chains/hyperlane-cosmos/src/error.rs
```diff
@@ -34,6 +34,9 @@ pub enum HyperlaneCosmosError {
     /// protobuf error
     #[error("{0}")]
     Protobuf(#[from] prost::DecodeError),
+    /// The string is not a valid number
+    #[error("Failed to parse number from string")]
+    NumStrParse,
```
tkporter, line 39: seems like this is maybe unused?

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -147,7 +150,8 @@ impl WasmGrpcProvider {
 
         let auth_info = signer_info.auth_info(Fee::from_amount_and_gas(
             Coin::new(
-                Amount::from((gas_limit as f64 * DEFAULT_GAS_PRICE) as u64),
+                // The fee to pay is the gas limit * the gas price
+                (BigFloat::from(gas_limit) * self.gas_price()).try_into()?,
```
tkporter, line 154: Do you think this needs to be `(BigFloat::from(gas_limit) * self.gas_price()).ceil().try_into()?` ?

Iiuc it may be possible for
- gas_price to be the very minimum that the node accepts
- `BigFloat::from(gas_limit) * self.gas_price()` results in decimals, e.g. 100.1
- the try_into truncates this to 100, causing the amount to not be enough to cover the requested gas_limit at the required minimum gas price

rust/chains/hyperlane-cosmos/src/trait_builder.rs
```diff
@@ -11,6 +16,38 @@ pub struct ConnectionConf {
     prefix: String,
     /// Canoncial Assets Denom
     canonical_asset: String,
+    /// The gas price set by the cosmos-sdk validator. Not that this represents the
```
tkporter, line 19: nit: Not -> Note

rust/hyperlane-base/src/settings/parser/connection_parser.rs
```diff
@@ -94,6 +94,12 @@ pub fn build_cosmos_connection_conf(
         None
     };
 
+    let minimum_gas_price = chain
+        .chain(err)
+        .get_opt_key("minimumGasPrice")
```
tkporter, line 99: iiuc you renamed to gas_price on the cosmos ConnectionConf - looks like here & in a few other places "minimum" is still used

rust/config/mainnet3_config.json
```diff
@@ -432,6 +432,10 @@
       "grpcUrl": "https://grpc-kralum.neutron-1.neutron.org:80",
       "canonicalAsset": "untrn",
       "prefix": "neutron",
+      "minimumGasPrice": {
+        "amount": "1",
```
tkporter, line 436: Should probably set this to 0.5 https://github.com/cosmos/chain-registry/blob/master/neutron/chain.json#L20

tkporter (approved):
Realizing that we'll need to change the typescript SDK too to keep things consistent

I think we need:
- a gas price property like we've done for other cosmos chain metadata properties here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/metadata/chainMetadataTypes.ts#L168
- add the property to the refine here to make sure it's specified for cosmos chains https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/metadata/chainMetadataTypes.ts#L206
- add the gas prices for the cosmos chains here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/consts/chainMetadata.ts

rust/hyperlane-core/src/types/primitive_types.rs
```diff
@@ -337,3 +341,77 @@ impl From<solana_sdk::signature::Signature> for H512 {
         H512(sig.into())
     }
 }
+
+/// Wrapper type around `BigDecimal` to implement various traits on it
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct BigFloat(BigDecimal);
```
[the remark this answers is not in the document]
daniel-savu: You're right, I'll use `FixedPointNumber` instead

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
daniel-savu: great point, I'm rounding up to the nearest integer now

PR #3043, opened 2023-12-11
StandardHookMetadata.sol small doc fix

NicholasDotSol:
### Description
small fix to docs


nambrot:
Thank you for the fix!

Issue #3045, opened 2023-12-11
Sealevel V3: change ISM metadata format to match v3 metadata


tkporter:
Change the message-id-multisig-ism program's metadata format to work with the v3 metadata format

The v2 version of metadata looked like this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/sealevel/programs/ism/multisig-ism-message-id/src/metadata.rs#L18-L21, but now it should look like this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/isms/libs/MessageIdMultisigIsmMetadata.sol#L5-L9

This'll involve changing the test conditions as well, you can possibly get this by looking at test messages from the explorer to get the metadata

Issue #3044, opened 2023-12-11
Sealevel V3: change message version to match V3


tkporter:
- Change the message version that's encoded in the message to match what V3 expects. Iirc it used to be version `0` on v2, but was changed to `3` on v3

PR #3048, opened 2023-12-11
Sealevel v3 contracts

daniel-savu:
### Description

MVP contracts for sealevel v3 - note this doesn't include hooks so IGP payments still have to be made separately. However contracts will now be compatible with the metadata format used by the other V3 deployments.

### Drive-by changes

The unit tests in `rust/sealevel/programs/ism/multisig-ism-message-id/src/metadata.rs` are refactored to reuse struct creation logic.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3044
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3045

### Backward compatibility

No, since this upgrades sealevel v2 to v3.

### Testing

Unit tests


rust/sealevel/programs/ism/multisig-ism-message-id/src/metadata.rs
```diff
@@ -32,7 +34,12 @@ impl TryFrom<Vec<u8>> for MultisigIsmMessageIdMetadata {
         }
 
         let origin_mailbox = H256::from_slice(&bytes[ORIGIN_MAILBOX_OFFSET..MERKLE_ROOT_OFFSET]);
-        let merkle_root = H256::from_slice(&bytes[MERKLE_ROOT_OFFSET..SIGNATURES_OFFSET]);
+        let merkle_root = H256::from_slice(&bytes[MERKLE_ROOT_OFFSET..MERKLE_INDEX_OFFSET]);
+        // This cannot panic since SIGNATURES_OFFSET - MERKLE_INDEX_OFFSET is 4.
+        let merkle_index_bytes: [u8; 4] = bytes[MERKLE_INDEX_OFFSET..SIGNATURES_OFFSET]
+            .try_into()
+            .unwrap();
```
tkporter, line 41: I don't feel strongly but noting that we return a Result here anyways so we could maybe do that instead of unwrapping? will defer to you though

tkporter (approved):
Unsure why tests are failing tbh, but code changes do lgtm

daniel-savu:
> Unsure why tests are failing tbh, but code changes do lgtm

Issue was that since message `version` is now `3` instead of `0`, I needed to recompute validator signatures when testing the multisig ISM

PR #3054, opened 2023-12-13
Fix slither CI action

yorhodes:
### Description

- Fixes slither github action and makes required

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -21,10 +21,14 @@ jobs:
         with:
           path: |
             **/node_modules
-            .yarn/cache
+            .yarn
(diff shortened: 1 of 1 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -21,10 +22,17 @@ jobs:
...
-            .yarn/cache
+            .yarn
           key: ${{ runner.os }}-yarn-cache-${{ hashFiles('./yarn.lock') }}
 
-      - run: yarn workspaces focus @hyperlane-xyz/core
+      - name: yarn-install
+        run: yarn install
+
+      - name: foundry-install
+        uses: onbjerg/foundry-toolchain@v1
```
aroralanuk, line 32: nit: can minmize the number of foundry installs like have it under the same jobs? We do it 8 times in CI now

aroralanuk (approved):

PR #3055, opened 2023-12-13
fix:add forge tests for `TestSendReceiver`

aroralanuk:
### Description

- Added `TestSendReceiverTest` after having removed `testSendReceiver.hardhat-test.ts`
(latter uses blockhash randomness and can fail in CI infrequently but very frustratingly whereas the former is fuzzed and conditionally checked)

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

Fuzz tests


yorhodes (approved):
nice
cc @tkporter

PR #3071, opened 2023-12-18
chore: remove neutron-related deployments from testnet4 context

daniel-savu:
### Description

Removes most occurences of lines that contain the `neutron` string from the testnet4 context. Where validators are set (e.g. [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3e685f9fb04fd1dd10b05250ade5b65150b75086/typescript/infra/config/environments/testnet4/validators.ts#L29)), the neutron context is required for type safety but can stay empty

### DIFF
--- a/rust/config/testnet4_config.json
+++ b/rust/config/testnet4_config.json
@@ -1007,32 +1007,6 @@
       "index": {
         "from": 4558491
       }
-    },
-    "neutrontestnet": {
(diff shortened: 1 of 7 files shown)


tkporter (approved):
💀 🪦

jmrossy (approved):

PR #3077, opened 2023-12-19
feat(cosmos): ism dry-run

daniel-savu:
### Description

Implements `dry_run_verify` for the aggregation ISM on cosmwasm. One remaining issue is that the estimated gas is hardcoded to `1`, because we're actually [just querying](https://github.com/many-things/cw-hyperlane/blob/37fea49429108d0cad46c64c3c1ebc467817ff8c/contracts/isms/aggregate/src/lib.rs#L108) via rpc rather than simulating a tx. The `verify` tx isn't marked as a contract [entrypoint](https://book.cosmwasm.com/basics/entry-points.html) so it can't be called from outside iiuc. (here's the [verify](https://github.com/many-things/cw-hyperlane/blob/37fea49429108d0cad46c64c3c1ebc467817ff8c/contracts/isms/aggregate/src/lib.rs#L124) fn for reference).

Worth mentioning that the query interface for the aggregation ISM is named incorrectly - it should return fields called `threshold` and `modules`, but instead copies the response from the multisig ISM and returns `threshold` and `validators`. This can be particularly misleading because validators have 20-bytes long addresses, whereas modules (contracts) have 32-bytes long addresses.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/807

### Backward compatibility

yes

### Testing

E2E. The ISM setup is `routing` -> `aggregation (1/1)` -> `multisig (1/1)`


rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
```diff
@@ -91,6 +93,23 @@ impl InterchainSecurityModule for CosmosInterchainSecurityModule {
...
+            .provider
+            .grpc()
+            .wasm_query(QueryIsmGeneralRequest { ism: payload }, None)
+            .await?;
+        let response: VerifyResponse = serde_json::from_slice(&data)?;
+        // We can't simulate the `verify` call in CosmWasm because
+        // it's not marked as an entrypoint. So we just use the query interface
+        // and hardcode a gas value - this can be inefficient if one ISM is
+        // vastly cheaper than another one.
+        let dummy_gas_value = U256::one();
```
tkporter, line 112: This is unfortunate :( Did you see if Mitosis had any ideas on what to do here? I guess we can leave this as-is and create and issue to track that this is incomplete.

If we really need this to be accurate in the future we can consider creating a contract that calls into the ISMs to verify

Also slight nit for the comment for posterity, `verify` is accessible via the query entrypoint but just not the execute entrypoint - when I first read this I was surprised that verify wasn't accessible via an entrypoint because it is able to be called by other contracts on chain https://book.cosmwasm.com/basics/entry-points.html

tkporter (approved):

rust/chains/hyperlane-cosmos/src/aggregation_ism.rs
```diff
@@ -66,15 +68,27 @@ impl AggregationIsm for CosmosAggregationIsm {
...
+        let response: VerifyInfoResponse = serde_json::from_slice(&data)?;
 
+        // Note that due to a misnomer in the CosmWasm implementation, the `modules` field is called `validators`.
         let modules: ChainResult<Vec<H256>> = response
-            .modules
-            .into_iter()
-            .map(|module| CosmosAddress::from_str(&module).map(|ca| ca.digest()))
+            .validators
+            .iter()
+            // The returned values are Bech32-decoded Cosmos addresses.
```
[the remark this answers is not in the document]
daniel-savu: I assume so? It's the stringified version of the bytes that comprise the H256 digest

rust/chains/hyperlane-cosmos/src/interchain_security_module.rs
daniel-savu: > verify is accessible via the query entrypoint
That's what we're using, I don't think we can use it to simulate txs though?

Will check with Mitosis as well

PR #3084, opened 2023-12-19
Reduce verbosity of smartprovider logging

jmrossy:


typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -289,16 +289,18 @@ export class HyperlaneSmartProvider
...
       const result = await provider.perform(method, params, reqId);
       return { status: ProviderStatus.Success, value: result };
     } catch (error) {
-      this.logger(
-        `Error performing ${method} on provider ${providerUrl} for reqId ${reqId}`,
-        error,
-      );
+      if (this.options?.debug)
+        this.logger(
+          `Error performing ${method} on provider ${providerUrl} for reqId ${reqId}`,
```
yorhodes, line 301: imo we should just omit this entirely, or maybe only log the domain name?

jmrossy: Okay done, replaced it with the provider index

Issue #3070, opened 2023-12-18
Indexed logs in agents should ensure no duplicates


tkporter:
Ran into an issue with the v2 scraper where p2pify was returning the same message dispatch twice in an eth_getLogs. E.g:

```
cast logs --rpc-url https://[redacted-p2pify-url] --from-block 34339856 --to-block 34341855 --address 0x35231d4c2D8B8ADcB5617A638A0c4548684c7C70 0x769f711d20c679153d382254f59892613b58a97cc876b249134ac25c80f9c814
```

This caused the scraper to crash because it'd identify a conflict in the insertion, and it doesn't know how to handle that. We should probably just do our best to make sure that any fetched logs aren't duplicates

PR #3108, opened 2024-01-02
fix: dedup eth logs

daniel-savu:
### Description

Deduplicates ethereum logs by collecting them into a HashSet before returning a Vec. I couldn't find a way to deduplicate using an ethers config, but since `fetch_logs` is only called once in hyperlane-base, I'm doing the deduplication there "globally". I also noted in doc comments that duplicates may be returned.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3070

### Backward compatibility

Yes

### Testing

Will rely on e2e passing but haven't added unit tests for this scenario

### DIFF
--- a/rust/chains/hyperlane-ethereum/src/mailbox.rs
+++ b/rust/chains/hyperlane-ethereum/src/mailbox.rs
@@ -125,6 +125,7 @@ where
         self.get_finalized_block_number().await
     }
 
+    /// Note: This call may return duplicates depending on the provider used
     #[instrument(err, skip(self))]
     async fn fetch_logs(
         &self,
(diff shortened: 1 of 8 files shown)


rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -125,6 +125,7 @@ where
         self.get_finalized_block_number().await
     }
 
+    /// Note: This call may return duplicates depending on the provider used
```
tkporter, line 128: nice

tkporter (approved):

PR #3116, opened 2024-01-03
Validator observability infra

### LINKED ISSUE #3109: Infra support & grafana alerts for validator observability
following the agent change to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3057, we need:
- ability to configure apps in infra (e.g. important warp routes)
- alert conditions if there are issues with validators in these sets

tkporter:
### Description

- Configures the relayers to start tracking the latest checkpoints of the manta TIA WR, arbitrum TIA WR, and helloworld sets (and, implicitly, the default ISM sets too)
- deployed all agent roles & networks with the new image
- created alerts and added to the nexus dashboard

One thing that's a bit of a bummer is the metrics are only updated whenever we try to construct metadata for a message. Naturally the validators will poll & sign checkpoints slightly out of sync, so pretty frequently the relayer attempts to deliver a message where say 4/6 signatures are needed, and 2 of the validators happen to just have not polled & signed the checkpoint yet. The message is then successfully delivered, and the 2 remaining validators probably sign the checkpoint in a matter of seconds, but the metrics aren't updated to reflect this unless there's another message whose metadata is being constructed. This just means that the graph is a bit ugly when things are working. But when things aren't working, we'll still be able to clearly see which validators are behind.

We may want to consider changing the metrics if this proves confusing. Some ideas:
1. Move to a separate task that occasionally polls the latest checkpoints of the configured app contexts
2. Also track the threshold. This way we can construct alerts & dashboards based off the threshold and not be overly concerned if there are some stragglers

### Drive-by changes

added `bytesToBytes32` so that we can construct matching lists, which the agents expect to be 0x-prefixed, using a router address config that may include protocol-specific address formats

### Related issues

Fixes #3109 

### Backward compatibility

yes

### Testing

deployed


yorhodes (approved):

nambrot:
Should this be merged? It's already live right?

PR #3120, opened 2024-01-03
set gasprice on bsctestnet

nambrot:
### Description

Sets a somewhat arbitrary gas price on bsctestnet to get transactions going for the key-funder (and presumably kathy as well). It's a pretty blunt hammer for a not-helpful `transaction underpriced` error, but IMO fine enough

### Drive-by changes

handleTx would not handle confirmations=0, and just set it to 1


typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -304,8 +304,11 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     chainNameOrId: ChainName | number,
     tx: ContractTransaction | Promise<ContractTransaction>,
   ): Promise<ContractReceipt> {
-    const confirmations =
-      this.getChainMetadata(chainNameOrId).blocks?.confirmations || 1;
+    let confirmations =
+      this.getChainMetadata(chainNameOrId).blocks?.confirmations;
+    if (confirmations === undefined) {
+      confirmations = 1;
+    }
```
jmrossy, line 311: `this.getChainMetadata(chainNameOrId).blocks?.confirmations ?? 1`

tkporter (approved):

PR #3122, opened 2024-01-04
fix: support TLS gRPC endpoints

daniel-savu:
### Description

Supporting TLS gRPC endpoints is as simple as enabling the `tonic` crate features in this PR. However to maintain compatiblity with the plaintext endpoints, we need to always use `http` instead of `https`, otherwise `tonic` will try to establish a TLS connection to them and fail. 

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/893

### Backward compatibility

No, in the sense that all prior plaintext urls need to be changed from `https` to `http`

### Testing

Manual


tkporter (approved):
🔥

PR #3126, opened 2024-01-04
Move to reorg period of 1 for mantapacific

tkporter:
### Description

- Moves to a reorg period of 1 for mantapacific instead of 0 - this should avoid the IGP indexing bugs we're seeing where we first get the latest block, then make an eth_getLogs RPC, but the node servicing the latter RPC doesn't yet know of the new latest block

### Drive-by changes

Remove some duplication of the reorg period

### Related issues

n/a

### Backward compatibility

backward compatible

### Testing

none yet, other than deploying


nambrot (approved):

typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -241,3 +241,9 @@ export function getChainIdNumber(chainMetadata: ChainMetadata): number {
   if (typeof chainMetadata.chainId === 'number') return chainMetadata.chainId;
   else throw new Error('ChainId is not a number, chain may be of Cosmos type');
 }
+
+export function mustGetReorgPeriod(chainMetadata: ChainMetadata): number {
```
jmrossy, line 245: A convention we use in a few places across the SDK (like the MultiProvider and the get methods above) is for `get*` to throw and `tryGet` to return nullish. Suggest `getReorgPeriod`

PR #3127, opened 2024-01-04
Fallback Provider Refactor

daniel-savu:
### Description

- Generalizes the fallback provider. Now any provider type can be used in it, as long as it implements `Into<Box<dyn BlockNumberGetter>> + Clone` (aka it can be wrapped into a struct that reads the current block #, and it can be cloned).
- moves the fallback provider implementation to hyperlane-core
	- had serious issues with getting versions to successfully resolve
	- next time I run into issues, strongly inclined to create a new crate with dependencies for all chains, but which are only depended on by the agents codebase (rather than the contract as well). Chatted with @tkporter about this.
- now fallbackprovider needs to be wrapped in a newtype struct for chain-specific trait implementations
	- If `Deref` is implemented for the newtype struct, all the function calls will still work (although using Deref for newtypes is [debatable](https://users.rust-lang.org/t/understanding-the-perils-of-deref/47958/18))
- the `request` function required by the `JsonRpcClient` still has some reusable logic but I left it there until it's clearer what can be reused by the cosmos fallback provider
	- the fallback provider unit tests still live in hyperlane-ethereum because of the logic in `require`

### Drive-by changes

- needed to use `async-rwlock` instead of `tokio` for the `RwLock` structure used by the fallback provider for interior mutability. This is because the solana program requirements had rust pinned too far into the past for `tokio` to resolve to a version.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/892

### Backward compatibility

Yes

### Testing

Unit tests that were already there for the fallback provider


rust/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -70,6 +37,7 @@ where
             .field(
                 "chain_name",
                 &self
+                    .0
```
tkporter, line 40: is this needed w/ deref? ditto line 50

rust/hyperlane-core/src/rpc_clients/fallback.rs
```diff
@@ -0,0 +1,189 @@
...
+};
+use tracing::info;
+
+use crate::ChainCommunicationError;
+
+/// Read the current block number from a chain.
+#[async_trait]
+pub trait BlockNumberGetter: Send + Sync + Debug {
+    /// Latest block number getter
+    async fn get(&self) -> Result<u64, ChainCommunicationError>;
```
tkporter, line 17: nit: don't feel strongly as maybe this is a common Rust pattern to rely more on the trait name than the function name to describe a function, but wondering if this should be `get_block_number` or something w/ more clarity in the function name

rust/ethers-prometheus/src/json_rpc_client.rs
```diff
@@ -172,3 +185,32 @@ where
...
+
+impl<C: JsonRpcClient + 'static> From<PrometheusJsonRpcClient<C>> for Box<dyn BlockNumberGetter> {
+    fn from(val: PrometheusJsonRpcClient<C>) -> Self {
+        Box::new(JsonRpcBlockGetter::new(val))
+    }
+}
+
+/// Utility struct for implementing `BlockNumberGetter`
+#[derive(Debug, new)]
+pub struct JsonRpcBlockGetter<T: JsonRpcClient>(T);
```
tkporter, line 197: I feel like this & the content below it doesn't feel right to put in `ethers-prometheus` as it's not directly related to ethers-prometheus.

Maybe somewhere in `chains/hyperlane-ethereum/src/rpc_clients`? Or was this motivated by avoiding circular imports? Just feels weird to have this here, but I see that `chains/hyperlane-ethereum/src/rpc_clients/fallback.rs` already assumes that the PrometheusJsonRpcClient is being used, so if it's a huge hassle feel free to leave this as-is

rust/hyperlane-core/src/rpc_clients/fallback.rs
daniel-savu: totally had meant to change this, thanks for pointing out

rust/ethers-prometheus/src/json_rpc_client.rs
daniel-savu: The only way I can move it out of here is if we wrapped `PrometheusJsonRpcClient` in a newtype in `hyperlane-ethereum` (and maybe pulled the Deref-on-newtype cheat to minimize changes). That's due to the orphan rule, since this is where `PrometheusJsonRpcClient` is defined and this trait impl will fail: `impl From<PrometheusJsonRpcClient<C>> for Box<dyn BlockNumberGetter>`

PR #3138, opened 2024-01-09
fix: `testRecipient` CLI bug

aroralanuk:
### Description

- we were recovering `TestRecipient` from sdk artifacts which means the deployer won't have ownership. So, I added a optional `shouldRecover` boolean which I turned off for TestRecipient deployer (this is only for our artifacts, if they have their own testRecipient, it will still recover that if provided)
- moved `TestRecipientDeployer` from cli to sdk and removed duplicated code from infra

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/895

### Backward compatibility

Yes

### Testing

Manual (between anvil1 and optimismgoerli)


.changeset/dirty-ears-push.md
```diff
@@ -0,0 +1,7 @@
+---
+'@hyperlane-xyz/infra': patch
+'@hyperlane-xyz/cli': patch
+'@hyperlane-xyz/sdk': patch
+---
+
+Added `shouldRecover` flag to deployContractFromFactory
```
jmrossy, line 7: Nit: Adding a note about how this fixes testRecipient related errors would be a good inclusion here too

typescript/infra/scripts/deploy.ts
```diff
@@ -138,8 +137,7 @@ async function main() {
     );
     deployer = new LiquidityLayerDeployer(multiProvider);
   } else if (module === Modules.TEST_RECIPIENT) {
-    config = objMap(envConfig.core, (_chain) => true);
-    deployer = new TestRecipientDeployer(multiProvider);
+    throw new Error('Test recipient is not supported. Use CLI instead.');
```
jmrossy, line 140: This is the first small step towards deprecating the infra package :)

typescript/sdk/src/core/TestRecipientDeployer.ts
```diff
@@ -39,7 +38,28 @@ export class TestRecipientDeployer extends HyperlaneDeployer<
...
+      'testRecipient',
+    );
+    let usePreviousDeployment = false;
+    if (
+      predeployed &&
+      eqAddress(
+        await predeployed.owner(),
+        await this.multiProvider.getSignerAddress(chain),
+      )
+    ) {
```
jmrossy, line 53: Per our chat I lean _slightly_ towards removing this check but I'm okay either way :)

jmrossy (approved):

PR #3141, opened 2024-01-09
Pausable ISM

yorhodes:
### Description

- Adds pausable hook and ism config support to deployers
- Configure testnet pausable hook and ISM
- Add pausable ism checking

### Drive-by changes

- Refactors ownable config

### Related issues

- https://github.com/hyperlane-xyz/issues/issues/706

### Backward compatibility

- Yes

### Testing

- Unit Tests and fork tests


typescript/sdk/src/hook/HyperlaneHookDeployer.ts
```diff
@@ -81,8 +84,14 @@ export class HyperlaneHookDeployer extends HyperlaneDeployer<
       config.type === HookType.FALLBACK_ROUTING
     ) {
       hook = await this.deployRouting(chain, config, coreAddresses);
+    } else if (config.type === HookType.PAUSABLE) {
+      hook = await this.deployContract(chain, config.type, []);
+      await this.transferOwnershipOfContracts(chain, config.owner, { hook });
```
jmrossy, line 89: Can you confirm that Pausable is the only ISM type for which we want to transfer ownership after deployment?

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -669,6 +682,17 @@ export async function moduleMatchesConfig(
       matches = true;
       break;
     }
+    case IsmType.PAUSABLE: {
+      const pausableIsm = PausableIsm__factory.connect(moduleAddress, provider);
+      const owner = await pausableIsm.owner();
+      matches = matches && eqAddress(owner, config.owner);
```
jmrossy, line 688: Feel 100% free to ignore this terse but cool syntax but:
`matches &&= eqAddress(owner, config.owner)`

jmrossy (approved):

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
yorhodes: its not the *only* one but we already do this in the `deployIgp` and `deployRouting` functions

yorhodes: its actually a good point, it seems like it may be missing from a few places...

PR #3142, opened 2024-01-10
fix:`HyperlaneIsmFactory` doesn't fail with missing chainMetadata errors

aroralanuk:
### Description

HyperlaneIsmFactory is now wary of (try)getDomainId or (try)getChainName calls which may fail and handles them appropriately.
- switched from chainName to domain in `routingModuleDelta`
- replace `getDomainId` with `tryGetDomainId` and `getChainName` with `tryGetChainName` wherever necessary.

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

Hardhat unit tests

### DIFF
--- a/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
+++ b/typescript/sdk/src/ism/HyperlaneIsmFactory.ts
@@ -29,6 +29,7 @@ import {
   normalizeAddress,
   objFilter,
   objMap,
+  warn,
 } from '@hyperlane-xyz/utils';
 
 import { HyperlaneApp } from '../app/HyperlaneApp';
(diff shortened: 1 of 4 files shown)


typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -428,12 +435,17 @@ export async function moduleCanCertainlyVerify(
   origin: ChainName,
   destination: ChainName,
 ): Promise<boolean> {
+  const originDomainId = multiProvider.tryGetDomainId(origin);
+  const destinationDomainId = multiProvider.tryGetDomainId(destination);
+  if (originDomainId === null || destinationDomainId === null) {
```
jmrossy, line 440: `if (!originDomainId || !destinationDomainId)`
Direct `null` comparisons like this won't cover the case where an empty or undefined value is checked.

typescript/sdk/src/ism/types.ts
```diff
@@ -106,8 +106,8 @@ export type DeployedIsm = ValueOf<DeployedIsmType>;
 
 // for finding the difference between the onchain deployment and the config provided
 export type RoutingIsmDelta = {
-  domainsToUnenroll: ChainName[]; // new or updated isms for the domain
-  domainsToEnroll: ChainName[]; // isms to remove
+  domainsToUnenroll: Domain[]; // new or updated isms for the domain
+  domainsToEnroll: Domain[]; // isms to remove
```
jmrossy, line 110: 👍

PR #3144, opened 2024-01-10
Injective-specific account info protobuf

tkporter:
### Description

Only change needed was to correctly handle the account info. Because we may want to handle injective testnet, added `is_injective` fn to HyperlaneDomain instead of just doing the match directly. Happy to import some of the longer fully qualified names instead


rust/Cargo.toml
```diff
@@ -91,8 +91,12 @@ elliptic-curve = "0.12.3"
 getrandom = { version = "0.2", features = ["js"] }
 hex = "0.4.3"
 hpl-interface = "=0.0.6-rc3"
+# Must be the same version used by Tonic
+http = "0.2.9"
```
daniel-savu, line 95: I assume `http = "*"` wasn't working?

rust/Cargo.toml
```diff
@@ -112,6 +116,7 @@ paste = "1.0"
 pretty_env_logger = "0.5.0"
 primitive-types = "=0.12.1"
 prometheus = "0.13"
+protobuf = "2.28.0"
```
daniel-savu, line 119: we already have protobuf in cargo lock so I'd also try resolving it with a wildcard here

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -271,6 +285,59 @@ impl WasmGrpcProvider {
         .map_err(Into::<HyperlaneCosmosError>::into)?;
         Ok(account)
     }
+
+    /// Injective-specific logic for querying an account.
+    async fn account_query_injective(&self, account: String) -> ChainResult<BaseAccount> {
+        let request = tonic::Request::new(
+            injective_std::types::cosmos::auth::v1beta1::QueryAccountRequest { address: account },
```
daniel-savu, line 292: if I were to nitpick I'd say we should feature flag `account_query_injective` and the injective-related dependencies + imports but it's no big deal. Maybe worth adding a comment about this though

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -271,6 +285,59 @@ impl WasmGrpcProvider {
...
+            .insert(GrpcMethod::new("cosmos.auth.v1beta1.Query", "Account"));
+
+        let response: tonic::Response<
+            injective_std::types::cosmos::auth::v1beta1::QueryAccountResponse,
+        > = grpc_client
+            .unary(req, path, codec)
+            .await
+            .map_err(Into::<HyperlaneCosmosError>::into)?;
+
+        let mut eth_account = injective_protobuf::proto::account::EthAccount::parse_from_bytes(
```
daniel-savu, line 318: wondering what other consumers of `injective_protobuf` there are, would've expected there already is logic somewhere to convert `EthAccount` into `BaseAccount` :/

daniel-savu (approved):

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
tkporter: hmm wouldn't this mean we'd just always have that feature flag enabled, and we'd still have a runtime check? because we may wanna relay between Neutron and Injective, so we need both injective and non-injective code present in this crate and a runtime check to use injective when appropriate. And I'm not sure we'd want to ship a different binary depending on whether we're using injective or not

rust/Cargo.toml
tkporter: i had a problem with it before but trying it again, even with a Cargo.lock from origin/main, works! nice

rust/Cargo.toml
tkporter: nice this worked too

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
tkporter: sadly couldn't find any that would help out here

PR #3147, opened 2024-01-11
Support cosmos contract byte lengths other than 32

### LINKED ISSUE #3143: Cosmwasm contract addresses w/ lengths other than 32 bytes should be supported
## Problem

Neutron cosmwasm contract addresses are 32 bytes. Injective ones are 20. The agent implementation uses H256s for contract addresses that are taken in via config, and therefore always creates a 32 byte `CosmosAddress`. When this is serialized in bech32, we get 32 byte addresses prefixed with 12 0 bytes, when we really need them serialized shorter

## Solution

- Should we allow / require protocol-specific addresses in config files? Atm we need to convert bech32 -> hex and then put it into the config JSONs. Same for Solana, we need to convert base58 -> hex. Maybe taking in bech32 in the config will let us support arbitrary length cosmos contract addresses?
- Other ideas could be to read them in as vec<u8> or something but still require hex? Or maybe to configure the length for contract addresses? but these feel a bit worse tbh

tkporter:
### Description

fixes #3143 

Initially went down a path that would use protocol-specific types for addresses as suggested by #3143. I had made a `HyperlaneConnectionConf` enum with protocol-specific variants whose values were `addresses: CoreContractAddresses<ProtocolSpecificAddressType>` and `connection: ProtocolSpecificConnectionConf`. This worked pretty well until I hit the ISM logic.

Because hyperlane-core is where the Mailbox trait is defined, the return type of `recipient_ism` in the trait cannot involve protocol specific types. It can't be moved to hyperlane-base because hyperlane-base imports the chain crates, so we'd have a cyclic dependency. I experimented with moving away from H256 to something like a Vec<u8> or string, but this felt a bit weird.

In the end we decided to keep H256s as the global representation for contract addresses for now, with the intent of eventually changing this, and to support the varying length situation in a cosmos config

### Drive-by changes

- Added some cosmos specific agent configurations into the sdk
- Moved to bech32_prefix in the agents for consistency with what the SDK's chain metadata already does
- I guess no one's ran cargo test in a while so vectors/message.json got a new v3 message lol

### Related issues

Fixes #3143 

### Backward compatibility

Changes prefix to bech32_prefix in the agent config, and now requires `contractAddressBytes`

### Testing

Tested merged with #3144 and all worked


typescript/sdk/src/metadata/agentConfig.ts
```diff
@@ -92,6 +92,30 @@ export type AgentSignerCosmosKey = z.infer<typeof AgentSignerNodeSchema>;
 export type AgentSignerNode = z.infer<typeof AgentSignerNodeSchema>;
 export type AgentSigner = z.infer<typeof AgentSignerSchema>;
 
+// Additional chain metadata for Cosmos chains required by the agents.
+const AgentCosmosChainMetadataSchema = z.object({
```
jmrossy, line 96: Should `buildAgentConfig` be updated to use this?

vectors/message.json
daniel-savu, line 1: what's this file actually doing? I kept excluding it from my commits

daniel-savu (approved):
🙏

vectors/message.json
tkporter: It's used by a solidity test in `solidity/test/message.test.ts`

solidity/test/message.test.ts
```diff
@@ -12,7 +12,7 @@ import { TestMessage, TestMessage__factory } from '../types';
 
 const remoteDomain = 1000;
 const localDomain = 2000;
-const version = 0;
+const version = 3;
```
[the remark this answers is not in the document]
yorhodes: (or use the contract version)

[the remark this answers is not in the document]
tkporter: good call, I ended up going with the contract version here

typescript/sdk/src/metadata/agentConfig.ts
tkporter: Because for now buildAgentConfig is just used for EVM chains atm, just gonna make a comment about `buildAgentConfig` needing some work before being used in non-EVM contexts for now

PR #3148, opened 2024-01-12
fix: failing ci tests for `SmartProvider` and `OpStackIsm.t.sol`

aroralanuk:
### Description

- Problem: OpStackIsm.t.sol was failing because of very stale block number for fork
Fix: Excluding foundry fork testing  from CI and updating block number

- Problem: SmartProvider getLogs tests failing because the the results end up being an empty list and getConfirmations results a smaller number of confirmations than expected (712 vs 1705184)
fix: excluded from ci

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/solidity/package.json
+++ b/solidity/package.json
@@ -53,7 +53,8 @@
     "flatten": "./flatten.sh",
     "storage": "./storage.sh",
     "prettier": "prettier --write ./contracts ./test",
-    "test": "hardhat test && forge test -vvv",
+    "test:local": "hardhat test && forge test -vvv",
+    "test": "hardhat test && forge test --no-match-test testFork -vvv",
     "gas": "forge snapshot",
     "gas-ci": "yarn gas --check --tolerance 2 || (echo 'Manually update gas snapshot' && exit 1)",
     "slither": "slither ."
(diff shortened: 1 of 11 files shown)


solidity/package.json
```diff
@@ -53,7 +53,8 @@
     "flatten": "./flatten.sh",
     "storage": "./storage.sh",
     "prettier": "prettier --write ./contracts ./test",
-    "test": "hardhat test && forge test -vvv",
+    "test:local": "hardhat test && forge test -vvv",
+    "test": "hardhat test && forge test --no-match-test testFork -vvv",
```
yorhodes, line 57: the inverse is less breaking
eg test:ci

typescript/sdk/package.json
```diff
@@ -61,7 +61,7 @@
     "prepublishOnly": "yarn build",
     "prettier": "prettier --write ./src",
     "test": "yarn test:unit && yarn test:hardhat && yarn test:foundry",
-    "test:unit": "mocha --config .mocharc.json './src/**/*.test.ts' --exit",
+    "test:unit": "mocha --config .mocharc.json './src/**/*.test.ts' --exclude './src/**/SmartProvider.test.ts' './src/**/*.test.ts' --exit",
```
yorhodes, line 64: why not do the same separation here?

yorhodes (approved):

PR #3150, opened 2024-01-13
Fix Smart Provider e2e test

jmrossy:
### Description

Remove hard-coding of min block number and topic1 value.

### Drive-by changes

Add convenience `isAddress` util function

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3148

### Backward compatibility

Yes

### DIFF
--- a/typescript/sdk/package.json
+++ b/typescript/sdk/package.json
@@ -61,7 +61,7 @@
     "prepublishOnly": "yarn build",
     "prettier": "prettier --write ./src",
     "test": "yarn test:unit && yarn test:hardhat && yarn test:foundry",
-    "test:ci": "yarn test:unit && yarn test:hardhat && yarn test:foundry",
+    "test:ci": "yarn test",
     "test:unit": "mocha --config .mocharc.json './src/**/*.test.ts' --exit",
     "test:hardhat": "hardhat test $(find ./src -name \"*.hardhat-test.ts\")",
     "test:metadata": "ts-node ./src/test/metadata-check.ts",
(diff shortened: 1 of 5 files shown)


typescript/sdk/package.json
```diff
@@ -61,7 +61,7 @@
     "prepublishOnly": "yarn build",
     "prettier": "prettier --write ./src",
     "test": "yarn test:unit && yarn test:hardhat && yarn test:foundry",
-    "test:ci": "yarn test:unit --exclude './src/**/SmartProvider.test.ts' && yarn test:hardhat && yarn test:foundry",
+    "test:ci": "yarn test:unit && yarn test:hardhat && yarn test:foundry",
```
aroralanuk, line 64: u can leave it as "test:ci":"yarn test" now :)

typescript/sdk/src/providers/SmartProvider/SmartProvider.test.ts
```diff
@@ -181,11 +179,10 @@ describe('SmartProvider', () => {
       });
 
       it('Handles parallel requests', async () => {
-        const result1Promise = provider.getLogs({
-          address: WETH_CONTRACT,
-          topics: [WETH_TRANSFER_TOPIC0],
-          fromBlock: MIN_BLOCK_NUM,
-          toBlock: MIN_BLOCK_NUM + 100,
+        const result1Promise = provider.call({
```
aroralanuk, line 182: why change it from getLogs to call?

aroralanuk (approved):

typescript/sdk/package.json
jmrossy: Good catch, done

typescript/sdk/src/providers/SmartProvider/SmartProvider.test.ts
jmrossy: Oh just to simplify this test since there's not a static min block number anymore. The nature of the call isn't important here, just that 3 run in parallel.

PR #3160, opened 2024-01-17
fix: add warning for advanced ism config usage

aroralanuk:
### Description

Add bold red warning for advanced ism for both configuring and deploying

Note: we don't merge sdk artifacts for advanced ism core deployment

### Drive-by changes

none

### Related issues

none

### Backward compatibility

Yes

### Testing

Manual


typescript/cli/src/config/ism.ts
```diff
@@ -108,7 +115,11 @@ export async function createIsmConfigMap({
   outPath: string;
   chainConfigPath: string;
 }) {
-  logBlue('Creating a new ISM config');
+  logBlue('Creating a new advanced ISM config');
+  logBoldRed('WARNING: USE AT YOUR RISK.');
+  logRed(
+    'Advanced ISM config means you need to be very comfortable with different ISM types and how they work together topologically. If you are not, please use the basic ISM config instead.',
```
jmrossy, line 121: This is good but I suggest a slightly different wording:
`Advanced ISM configs require knowledge of different ISM types and how they work together topologically. If possible, use the basic ISM configs are recommended.`

typescript/cli/src/deploy/core.ts
```diff
@@ -161,6 +168,10 @@ async function runIsmStep(
   const isIsm = isZODISMConfig(ismConfigPath);
   // separate flow for 'ism' and 'ism-advanced' options
   if (isIsm) {
```
jmrossy, line 170: Can you confirm this check will only apply for advanced configs? The name makes me think it's for any ism config

typescript/cli/src/deploy/core.ts
```diff
@@ -161,6 +168,10 @@ async function runIsmStep(
...
+    logBoldRed('WARNING: YOU ARE DEPLOYING WITH AN ADVANCED ISM CONFIG');
+    logRed(
+      'Advanced ISM config means you need to be very comfortable with different ISM types and how they work together topologically. If you are not, please use the basic ISM config instead.',
```
jmrossy, line 173: ditto as above

typescript/cli/logger.ts
```diff
@@ -50,6 +50,8 @@ export const logPink = (...args: any) =>
 export const logGray = (...args: any) => console.log(chalk.gray(...args));
 export const logGreen = (...args: any) => console.log(chalk.green(...args));
 export const logRed = (...args: any) => console.log(chalk.red(...args));
+export const logBoldRed = (...args: any) =>
```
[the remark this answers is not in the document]
aroralanuk: both but I used bold figuratively here for emphasis not just to mean bold lettering. I'm open to alternatives.

typescript/cli/src/deploy/core.ts
aroralanuk: yes, this is just for advancedIsmConfig.

If I recall correctly, a certain two reviewers didn't like the idea of calling it `isAdvancedIsm` and `readAdvancedIsmConfig` so I changed it to isIsm, saying that this will cause confusion further down the road. 👀

typescript/cli/logger.ts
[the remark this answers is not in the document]
paulbalaji:
```suggestion
export const logBoldUnderlinedRed = (...args: any) =>
```

imo this to make it clearer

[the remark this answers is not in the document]
jmrossy: I'd suggest logRedEmphasis but Paul's suggestion is good too

typescript/cli/src/deploy/core.ts
jmrossy: IIRC we didn't like the name `isZODISMConfig` because of the Zod part, zod being an implementation detail and not a property of the config itself. I still think that :)

My confusion here is because of the variable name though (`isISM`)

I'm in favor of renaming `isZODISMConfig` to `isAdvancedIsm` if that's what you prefer!

aroralanuk: made it isAdvancedIsm

jmrossy (approved):

PR #3162, opened 2024-01-17
short term kludge: sleep 5s after getting blockhash on solana

tkporter:
### Description

context: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3167

This is just a temporary measure and not intended to be anything permanent


daniel-savu (approved):
don't know the context but could also try using finalized commitment when reading the blockhash

PR #3163, opened 2024-01-17
Add optional gRpcUrls field to ChainMetadata schema

jmrossy:
### Description

- Add optional gRpcUrls field to ChainMetadata schema
- Enforce inclusion of restUrls and gRpcUrls for cosmos chains
- Fix neutron rest url 

### Related Issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3161

### Backward compatibility

Yes

### Testing

Unit test


typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -111,6 +111,10 @@ export const ChainMetadataSchemaObject = z.object({
     .array(RpcUrlSchema)
     .describe('For cosmos chains only, a list of Rest API URLs')
     .optional(),
+  gRpcUrls: z
```
tkporter, line 114: don't feel strongly but I've tended to see this stylized as `grpc` instead of `gRpc` (e.g. https://github.com/cosmos/chain-registry/blob/master/injective/chain.json#L311)

tkporter (approved):

typescript/sdk/src/metadata/chainMetadataTypes.ts
jmrossy: agreed, will fix

PR #3175, opened 2024-01-23
Dan/injective testnet agents

daniel-savu:


tkporter (approved):

daniel-savu:
Closing as the injective testnet deployment was considered ephemeral, because its chain id clashes with the id of injective mainnet. @tkporter will open a separate PR for the blacklist

PR #3178, opened 2024-01-23
fix:use custom hook for sending test messages in the CLI

aroralanuk:
### Description

- Saving the deployed top-level hook address if the deployer is not the owner of the mailbox
- CLI command for sending messages uses the custom hook if present in the artifacts

### Drive-by changes

None

### Backward compatibility

Yes

### Testing

Manual


typescript/cli/src/send/message.ts
```diff
@@ -96,6 +96,15 @@ async function executeDelivery({
...
   const mailbox = core.getContracts(origin).mailbox;
 
+  let hook: Address;
+  try {
+    hook = mergedContractAddrs[origin].customHook;
+    logBlue(`Using custom hook ${hook} for ${origin} -> ${destination}`);
+  } catch (e) {
+    hook = await mailbox.defaultHook();
+    logBlue(`Using default hook ${hook} for ${origin} -> ${destination}`);
+  }
```
jmrossy, line 106: I recommend avoiding try/catch for control flow like this.
If you're trying to guard against errors from accessing `customHook`, use optional chaining instead:
`mergedContractAddrs[origin]?.customHook`

typescript/cli/src/send/message.ts
```diff
@@ -106,19 +115,29 @@ async function executeDelivery({
...
     log(`Paying for gas with ${value} wei`);
 
     log('Dispatching message');
-    const messageTx = await mailbox['dispatch(uint32,bytes32,bytes)'](
+    const messageTx = await mailbox[
+      'dispatch(uint32,bytes32,bytes,bytes,address)'
+    ](
       destinationDomain,
       formattedRecipient,
       MESSAGE_BODY,
```
jmrossy, line 135: Completely optional but someone recently asked for the ability to set custom content for the MESSAGE_BODY here (like via a CLI arg). If you feel like adding that on go ahead, otherwise np I'll get to it later

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -240,27 +242,32 @@ export abstract class HyperlaneDeployer<
...
-            `Set hook failed on ${chain}, wanted ${targetHook}, got ${actualHook}`,
+            `Set hook failed on ${chain}, wanted ${targetHook.address}, got ${actualHook}`,
           );
         }
+        return true;
       });
+      // if the signer is not the owner, saving the hook address in the artifacts for later use for sending test messages, etc
+      if (!result) {
+        this.addDeployedContracts(chain, { customHook: targetHook });
+      }
```
jmrossy, line 270: Not directly related to this addition but I'm curious to get your take on the maintainability of the deployers atm. To you feel like we have a flexible enough foundation to build on or is it difficult to support new scenarios?

yorhodes (commented):
I would solve this by not using the `CoreDeployer` on the core chains at all instead of inventing a new artifacts format

typescript/cli/src/send/message.ts
yorhodes: seems easy and useful for testing :)

typescript/cli/src/send/message.ts
aroralanuk: how does the optional chaining help here `mergedContractAddrs[origin]?.customHook` will still be undefined because mergedContractAddrs[origin] is ok but there's no customHook

typescript/cli/src/send/message.ts
aroralanuk: Done!

typescript/sdk/src/deploy/HyperlaneDeployer.ts
aroralanuk: Yorke and I were discussing this IRL. Yorke thinks we need to not deploy the core config on the core chains because we would need to replicate maintain predeployed core artifacts. This means we need to run through some of the `runIfOwner` function needlessly for core chains. I don't feel strongly but do think the deployer should be resilient to these changes.

aroralanuk:
> I would solve this by not using the `CoreDeployer` on the core chains at all instead of inventing a new artifacts format

I thought we discussed this was meant as a followup since this means more work.

typescript/cli/src/send/message.ts
```diff
@@ -96,6 +96,15 @@ async function executeDelivery({
...
 
+  let hook: Address;
+  try {
+    hook = mergedContractAddrs[origin].customHook;
+    logBlue(`Using custom hook ${hook} for ${origin} -> ${destination}`);
+  } catch (e) {
+    hook = await mailbox.defaultHook();
+    logBlue(`Using default hook ${hook} for ${origin} -> ${destination}`);
+  }
+
```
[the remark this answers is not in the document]
aroralanuk: The idea was to inform the user which hook they are using. Using zero address is the same as just calling the quoteDispatch(uint32,bytes32,bytes).

typescript/cli/src/send/message.ts
```diff
@@ -106,19 +115,29 @@ async function executeDelivery({
...
 
     log('Getting gas quote');
-    const value = await mailbox['quoteDispatch(uint32,bytes32,bytes)'](
+    const value = await mailbox[
+      'quoteDispatch(uint32,bytes32,bytes,bytes,address)'
+    ](
       destinationDomain,
       formattedRecipient,
       MESSAGE_BODY,
+      ethers.utils.hexlify([]),
```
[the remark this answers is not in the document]
aroralanuk: Agreed, here: https://github.com/hyperlane-xyz/issues/issues/997

However, don't think this is very important right now.

typescript/cli/src/send/message.ts
jmrossy: When else would `  hook = mergedContractAddrs[origin].customHook;` throw and error? Why do you have a try catch around it?

typescript/cli/src/commands/send.ts
```diff
@@ -52,6 +53,11 @@ const messageOptions: { [k: string]: Options } = {
     description: 'Skip wait for message to be delivered',
     default: false,
   },
+  messageBody: {
+    type: 'string',
+    description: 'Optional Message body',
+    default: 'Hello!',
+  },
```
jmrossy, line 60: This should be added to the message command's builder line because the transfer command won't work with a message body. It may confuse people to see that option in there

typescript/cli/src/send/message.ts
aroralanuk: fixed :)

typescript/cli/src/commands/send.ts
aroralanuk: fixed!

jmrossy (approved):

PR #3181, opened 2024-01-24
Fix address util nullish fallback handling

jmrossy:
### Description

Small fix for address utils

### Related issues

https://github.com/hyperlane-xyz/issues/issues/899

### Backward compatibility

Yes

### Testing

In warp UI


typescript/utils/src/typeof.ts
```diff
@@ -1,6 +1,7 @@
-export function isNullish<T>(val: T) {
-  if (val === null || val === undefined) return true;
-  else return false;
+export function isNullish<T>(
+  val: T,
+): val is T extends null | undefined ? T : never {
+  return val === null || val === undefined;
```
yorhodes, line 4: the conditional type here is a bit hard to read

yorhodes (approved):

typescript/utils/src/typeof.ts
jmrossy: Yeah complex types are like regex. Nearly unparseable by humans. I don't know of a simpler way to do this type narrowing though, sorry

PR #3183, opened 2024-01-27
feat:smarter hook artifact caching

aroralanuk:
### Description

- Right now, we have special case logic for igp in cli/contexts.ts to exempt it from being cached because the cached artifact is from core which the PI deployment doesn't have access to. This is fragile and furthermore, as we add more such hooks, we'll need more special cased logic which is undesirable.

- The core issue here is the PI deployment's inability to configure hooks for their needs which stems from the fact that they're not the owned of the cached artifact. 
- The simplest solution here is to not cache contract address which are A. part of hook factories and B. not owned by the deployer. This makes sure we will never run into issue of the hook deployment caching AW addresses and trying to modify them for the PI deployment and return prematurely (because of the runIfOwner flow)
- Note A: we're filtering out artifacts in hookFactories, not their respective dependencies like storageGasOracle for interchainGasPaymaster so this still implies that the storageGasOracle may be wrongly recovered (we can either add it to hookFactories or  have some way of mapping each hook's dependency) cc @yorhodes 
- Note B: this approach may lead to false positives where we unnecessarily filter out non-ownable contracts but this is still a best solution than the status quo

### Drive-by changes

- removing the special case logic for igp recovery in CLI

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual through CLI

### DIFF
--- a/typescript/cli/src/context.ts
+++ b/typescript/cli/src/context.ts
@@ -10,7 +10,7 @@ import {
   chainMetadata,
   hyperlaneEnvironments,
 } from '@hyperlane-xyz/sdk';
-import { objFilter, objMap, objMerge } from '@hyperlane-xyz/utils';
+import { objFilter, objMerge } from '@hyperlane-xyz/utils';
 
 import { runDeploymentArtifactStep } from './config/artifacts.js';
 import { readChainConfigsIfExists } from './config/chain.js';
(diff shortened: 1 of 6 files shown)


jmrossy:
@aroralanuk looks like this is failing the lint and e2e tests. Lmk when you've fixed those and then I'll have another look

aroralanuk:
closing as stale

PR #3155, opened 2024-01-15
Feat/injective

tkporter:

PR #3140, opened 2024-01-09
Deploy injective and inEVM

yorhodes:
### Description

- Deploy core to injective and inEVM networks

### Drive-by changes

- Update cw-hyperlane dependency to https://github.com/hyperlane-xyz/cosmwasm/releases/tag/v0.0.6

### Related issues

- 

### Backward compatibility

Yes

### Testing

- Fork tests
- e2e tests

PR #3185, opened 2024-01-29
Injective / inevm agent code & config changes

tkporter:
### Description

Taken from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3155 but without any of the ancillary things like changes to the SDK or infra. Merging to get an image off `main` that works w/ injective.

Branch plan is:
- after this is merged, an image off main can be used to run an agent against injective/inevm
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3140 will have all the SDK changes from the contract deployments
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3155 will have the infra changes to start deploying injective / inevm validators / relayer


daniel-savu (approved):

rust/config/mainnet3_config.json
```diff
@@ -450,6 +450,37 @@
         "prefix": "neutron"
       }
     },
+    "injective": {
```
daniel-savu, line 453: shouldn't this file include inevm configs too?

PR #3187, opened 2024-01-29
Created interface for Query Router API

armanthepythonguy:
### Description
Created interface file for Interchain Query Router API

### Drive-by changes
IInterchainQueryRouter.sol created in the interfaces folder.

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual


solidity/contracts/interfaces/IInterchainQueryRouter.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+import "../middleware/libs/Call.sol";
+
+interface IInterchainQueryRouter{
```
yorhodes, line 6: if we are going to have this explicit interface, would be nice to enforce this interface on the `InterchainQueryRouter` contract

yorhodes (changes requested):
for what its worth you can use the `InterchainQueryRouter` contract as an interface as well

solidity/contracts/interfaces/IInterchainQueryRouter.sol
armanthepythonguy: Have pushed required changes

yorhodes (approved):

PR #3190, opened 2024-01-30
feat: deploy new RC

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/921

- [x] update testnet4
- [x] update mainnet3
- [x] estimate and add 10% gas limit buffer for contract deployment
- [x] drive-by add chain id to SmartProvider logs


typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
paulbalaji, line 1: TODO: change how the network is logged

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -52,6 +53,8 @@ export class HyperlaneSmartProvider
     super(network);
     const supportedMethods = new Set<ProviderMethod>();
 
+    this.logger = debug(`hyperlane:SmartProvider:${this.network.chainId}`);
```
jmrossy, line 56: good idea!

jmrossy (approved):
LGTM but recommend a 2nd review for the infra artifact changes since I'm not very familiar with that

paulbalaji:
> LGTM but recommend a 2nd review for the infra artifact changes since I'm not very familiar with that

@tkporter @yorhodes probably a good shout here

typescript/helloworld/src/scripts/deploy.ts
```diff
@@ -18,7 +21,11 @@ async function main() {
   multiProvider.setSharedSigner(signer);
 
   const core = HyperlaneCore.fromEnvironment('testnet', multiProvider);
-  const config = core.getRouterConfig(signer.address);
+  const config = objFilter(
+    core.getRouterConfig(signer.address),
+    (chainName, _): _ is RouterConfig =>
+      chainMetadata[chainName].protocol === ProtocolType.Ethereum,
```
yorhodes, line 27: this continues to be necessary in more and more places :(

typescript/infra/config/environments/mainnet3/helloworld/rc/addresses.json
yorhodes, line 1: nice!

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -167,6 +166,7 @@ const desiredKathyBalancePerChain: ChainMap<string> = {
   scroll: '0.05',
   base: '0.05',
   polygonzkevm: '0.05',
+  viction: '0.05',
```
yorhodes, line 169: nice catch
is inevm missing here too?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -289,9 +289,22 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
+    // estimate gas
+    const deployTx = contractFactory.getDeployTransaction(...params, overrides);
+    const gasEstimated = await signer.estimateGas(deployTx);
+
+    // deploy with 10% buffer on gas limit
+    const contract = await contractFactory.deploy(...params, {
+      ...overrides,
+      gasLimit: gasEstimated.add(gasEstimated.div(10)), // 10% buffer
+    });
+
```
yorhodes, line 306: curious when the gas limit of a CREATE tx would change?

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
yorhodes: should we use network.name here?

typescript/infra/scripts/deploy.ts
```diff
@@ -111,7 +113,10 @@ async function main() {
     deployer = new TestQuerySenderDeployer(multiProvider);
   } else if (module === Modules.HELLO_WORLD) {
     const core = HyperlaneCore.fromEnvironment(env, multiProvider);
-    config = core.getRouterConfig(envConfig.owners);
+    config = objFilter(
+      core.getRouterConfig(envConfig.owners),
```
yorhodes, line 117: I wonder if we should actually implement this filtering inside the `core.getRouterConfig`
the core implementation here is EVM specific

yorhodes (approved):

typescript/sdk/src/providers/MultiProvider.ts
paulbalaji: possible from RPC problems leading to misestimation? I've encountered issues on polygon, bsc, goerli and a few others where the tx fails because the gas limit automatically set is _just_ lower than required

typescript/infra/scripts/deploy.ts
paulbalaji: not a bad shout - didn't want to do that initially in case it affected something else but if it's EVM specific anyway...

Issue #2586, opened 2023-08-01
`get_next_range` errors are not gracefully handled and result in a panic


tkporter:
If `get_next_range` gives an error in the validator or relayer, contract sync does not gracefully handle this and the agent will panic. We should gracefully handle this

cc @daniel-savu who also spotted

nambrot:
When will this error?

PR #3191, opened 2024-01-30
fix: potential rate-limiting bugfix

daniel-savu:
### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2586

### DIFF
--- a/rust/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/hyperlane-base/src/contract_sync/mod.rs
@@ -1,4 +1,6 @@
-use std::{collections::HashSet, fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc};
+use std::{
+    collections::HashSet, fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc, time::Duration,
+};
 
 use cursor::*;
 use derive_new::new;
@@ -16,6 +18,8 @@ use crate::settings::IndexSettings;
 mod cursor;
 mod eta_calculator;
 mod metrics;
+
+const SLEEP_MS: u64 = 5_000;
 
 /// Entity that drives the syncing of an agent's db with on-chain data.
 /// Extracts chain-specific data (emitted checkpoints, messages, etc) from an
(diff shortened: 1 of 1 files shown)


rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -60,13 +64,16 @@ where
         loop {
             indexed_height.set(cursor.latest_block() as i64);
             let Ok((action, eta)) = cursor.next_action().await else {
+                sleep(Duration::from_millis(SLEEP_MS)).await;
                 continue;
             };
-            match action {
-                CursorAction::Query(range) => {
+            let sleep_duration = match action {
+                CursorAction::Query(range) => loop {
```
daniel-savu, line 71: using `loop` so I can break with a return value (the sleep duration)

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -77,16 +84,22 @@ where
                         "Found log(s) in index range"
                     );
                     // Store deliveries
-                    let stored = self.db.store_logs(&logs).await?;
+                    let Ok(stored) = self.db.store_logs(&logs).await else {
```
daniel-savu, line 87: unsure if it's worth sleeping for these ones - I'm assuming this one and the one below _should_ be propagated

rust/hyperlane-base/src/contract_sync/mod.rs
tkporter: interesting, can you maybe add this as an in-code comment too

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -17,6 +19,8 @@ mod cursor;
 mod eta_calculator;
 mod metrics;
 
+const SLEEP_MS: u64 = 5_000;
```
tkporter, line 22: Fwiw Duration::from_millis is const and Copy so we could change line 70's `sleep_duration` to be a `Duration` instead of u64, also that way we don't need this logic `duration.as_millis().try_into().unwrap_or(SLEEP_MS)` on line 99

tkporter (approved):

PR #3204, opened 2024-01-31
feat/CLI tool for monitoring warp route balances

aroralanuk:
### Description

Motivation

After Hyperlane launches a warp route, we need to monitor the balances for the grafana dashboard.

Before

1. Ask Kunal to run the script  
2. Kunal will ask Rossy/Yorke for the warp route configs (tokenAddress, hyperc20Address, etc)
3. Kunal will run it locally and then deploy it in on k8s

After

1. Make a config yaml file to test locally and deploy on k8s with new workload name as the file.

PR Details

- Replaced the statically typed warp route configs with a more concise zod config which parses the yaml file.
- Added the yaml files for nautilus, nexus and injective warp routes.
- Note: The ZOD config is compatible with the updated registry schema but incompatible with the WarpRouteConfigSchema mostly due to the different toplevel keys (base and synthetic vs chainNames) and hence the duplication.
- The chainNames is preferred to base and synthetic bc of its uniqueness as a key and consistency with the rest of the tooling. (eg. injective and inevm has native types on both side where it's not obvious which should be base or synthetic)
- Note: The file names need to follow the `${route}-deployments.yaml` still to be properly parsed by the deploy-warp-monitor.ts script (this path should be relative to monorepo root)

### Drive-by changes

- Warp route monitor can now fetch native balances through the CosmNativeTokenAdapter.

### Related issues

None

### Backward compatibility

No, the script needs a new arg

### Testing

Manual


typescript/sdk/src/metadata/warpRouteConfig.ts
```diff
@@ -0,0 +1,27 @@
+import { z } from 'zod';
+
+import { ProtocolType } from '@hyperlane-xyz/utils';
+
+import { TokenType } from '../token/config';
+import { ChainMap } from '../types';
+
+const TokenConfigSchema = z.object({
```
jmrossy, line 8: What are the advantages of this new schema over the existing CLI one? 
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/config/warp.ts

If this is preferable, should the CLI use it?

aroralanuk: I mentioned it in the description:

> The chainNames is preferred to base and synthetic bc of its uniqueness as a key and consistency with the rest of the tooling. (eg. injective and inevm has native types on both side where it's not obvious which should be base or synthetic)

jmrossy (approved):
As I mentioned, I don't love the addition of a new config shape (this makes 4) but I'm open to solving this more holistically later as you suggested. 

I recommend getting approval from @yorhodes who actually uses the warp deployment script.

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -40,20 +39,36 @@ const warpRouteTokenBalance = new Gauge({
...
 });
 
+export function readWarpRouteConfig(filePath: string) {
+  const config = readYaml(filePath);
+  if (!config) throw new Error(`No warp config found at ${filePath}`);
+  const result = WarpRouteConfigSchema.safeParse(config);
+  if (!result.success) {
+    const firstIssue = result.error.issues[0];
+    throw new Error(
+      `Invalid warp config: ${firstIssue.path} => ${firstIssue.message}`,
```
nambrot, line 49: Wdyt about just print out the whole error object?a

nambrot (approved):
I agree that we probably will want to talk about the schema in the context of WarpCore and the registry, but right now this is a strict improvement over the current status quo, so I'm fine with shipping it

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
jmrossy: He got this from a the SDK's chain metadata utils which extract the first one to make it easier to surface to users (e.g. in explorer). But in this context I agree there's no reason to pull just the first.

aroralanuk: fixed!

PR #3206, opened 2024-02-01
feat: build head of PRs instead of merge commit

paulbalaji:
With our current repo settings for PRs, github checks out the merge commit with main instead of the actual HEAD of the PR. It feels counter-intuitive for CI to build the merge commit each time rather than the commit one has pushed, considering we require branches to be up to date with main before merging anyway.

This can also cause problems when deploying RCs, because your RC monorepo builds (e.g. for keyfunder, kathy) would unintentionally include changes that are not in your RC branch. Experienced first hand as part of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3190.

This change ensures that when running in a PR context it uses the intended HEAD of the PR, rather than the unintended merge commit with main.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          # check out full history
           fetch-depth: 0
 
       - name: yarn-cache
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
 
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
           submodules: recursive
 
       - name: yarn-cache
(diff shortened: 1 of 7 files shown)


tkporter (approved):
maybe best to wait for Yorke's sign off as well, he's spent much more time in github actions than i have

.github/workflows/test.yml
```diff
@@ -82,7 +84,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          # check out full history
```
yorhodes, line 85: this comment is for fetch-depth: 0

yorhodes (approved):
I think ideally we aren't deploying from shas not merged/on main but this seems fairly non-breaking

.github/workflows/test.yml
paulbalaji: oh woops, went trigger happy with a find/replace

.github/workflows/test.yml
```diff
@@ -82,7 +84,7 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
paulbalaji, line 87:
```suggestion
          ref: ${{ github.event.pull_request.head.sha || github.sha }}
          # check out full history
```

PR #3207, opened 2024-02-01
feat: improve create-key logging 

paulbalaji:
improve logging through create-key script and related infra logic

also add `[chainName]` prefix to validator announce logs

drive-by from doing https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3190

### DIFF
--- a/typescript/infra/scripts/utils.ts
+++ b/typescript/infra/scripts/utils.ts
@@ -180,7 +180,7 @@ export function getKeyForRole(
   role: Role,
   index?: number,
 ): CloudAgentKey {
-  debugLog(`Get key for ${role} role`);
+  debugLog(`Getting key for ${role} role`);
   const environmentConfig = environments[environment];
   const agentConfig = getAgentConfig(context, environmentConfig);
   return getCloudAgentKey(agentConfig, role, chain, index);
--- a/typescript/infra/scripts/utils.ts
+++ b/typescript/infra/scripts/utils.ts
@@ -195,9 +195,9 @@ export async function getMultiProviderForRole(
   // TODO: rename to consensusType?
   connectionType?: RpcConsensusType,
 ): Promise<MultiProvider> {
-  debugLog(`Get multiprovider for ${role} role`);
+  debugLog(`Getting multiprovider for ${role} role`);
   if (process.env.CI === 'true') {
-    debugLog('Return multiprovider with default RPCs in CI');
+    debugLog('Returning multiprovider with default RPCs in CI');
     return new MultiProvider(); // use default RPCs
   }
   const multiProvider = new MultiProvider(txConfigs);
--- a/typescript/infra/scripts/utils.ts
+++ b/typescript/infra/scripts/utils.ts
@@ -1,3 +1,4 @@
+import debug from 'debug';
 import path from 'path';
 import yargs from 'yargs';
 
@@ -28,6 +29,8 @@ import { fetchProvider } from '../src/config/chain';
 import { EnvironmentNames, deployEnvToSdkEnv } from '../src/config/environment';
 import { Role } from '../src/roles';
 import { assertContext, assertRole, readJSON } from '../src/utils/utils';
+
+const debugLog = debug('infra:scripts:utils');
 
 export enum Modules {
   // TODO: change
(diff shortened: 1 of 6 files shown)


typescript/infra/scripts/utils.ts
```diff
@@ -171,6 +180,7 @@ export function getKeyForRole(
   role: Role,
   index?: number,
 ): CloudAgentKey {
+  debugLog(`Get key for ${role} role`);
```
tkporter, line 183: nit: not important but elsewhere we say stuff like "Getting ..." etc instead of "Get ...", would be good to be consistent throughout here / this PR

typescript/infra/src/agents/aws/key.ts
```diff
@@ -81,7 +89,7 @@ export class AgentAwsKey extends CloudAgentKey {
 
   get keyConfig(): AwsKeyConfig {
     return {
-      type: AgentSignerKeyType.Aws,
+      type: AgentSignerKeyType.Aws, // Ensure this is the specific enum value expected
```
tkporter, line 92: was this intentional

typescript/infra/src/agents/aws/key.ts
```diff
@@ -53,16 +55,22 @@ export class AgentAwsKey extends CloudAgentKey {
       throw new Error('Not configured as AWS');
     }
     this.region = agentConfig.aws.region;
+    this.logger = debug(`infra:agents:key:aws:${this.identifier}`);
```
tkporter, line 58: this is sick

typescript/infra/src/utils/gcloud.ts
```diff
@@ -182,5 +201,7 @@ function iamConditionsEqual(
   if (a === undefined && b === undefined) {
     return true;
   }
-  return a && b && a.title === b.title && a.expression === b.expression;
+  const isEqual =
```
tkporter, line 204: ¿por que?

tkporter (approved):
love it

typescript/infra/src/utils/gcloud.ts
paulbalaji: linting

I have another task in my mental backlog to update repo settings to auto-lint/format on a file save in vscode

paulbalaji: ah wait this one is probably from adding a log and then removing the log because it's too verbose

typescript/infra/scripts/utils.ts
paulbalaji, line 183:
```suggestion
  debugLog(`Getting key for ${role} role`);
```

typescript/infra/scripts/utils.ts
```diff
@@ -185,7 +195,9 @@ export async function getMultiProviderForRole(
   // TODO: rename to consensusType?
   connectionType?: RpcConsensusType,
 ): Promise<MultiProvider> {
+  debugLog(`Get multiprovider for ${role} role`);
   if (process.env.CI === 'true') {
+    debugLog('Return multiprovider with default RPCs in CI');
```
paulbalaji, line 200:
```suggestion
  debugLog(`Getting multiprovider for ${role} role`);
  if (process.env.CI === 'true') {
    debugLog('Returning multiprovider with default RPCs in CI');
```

typescript/infra/src/agents/aws/key.ts
```diff
@@ -53,16 +55,22 @@ export class AgentAwsKey extends CloudAgentKey {
...
   }
 
   get privateKey(): string {
+    this.logger(
+      'Attempt to access private key, which is unavailable for AWS keys',
```
paulbalaji, line 63:
```suggestion
      'Attempting to access private key, which is unavailable for AWS keys',
```

typescript/infra/src/agents/key-utils.ts
```diff
@@ -17,6 +19,8 @@ import { AgentAwsKey } from './aws/key';
 import { AgentGCPKey } from './gcp';
 import { CloudAgentKey } from './keys';
 
+const debugLog = debug('infra:scripts:utils');
```
paulbalaji, line 22:
```suggestion
const debugLog = debug('infra:agents:key:utils');
```

Issue #3047, opened 2023-12-11
Investigate spawned metric tasks in agents


tkporter:
It seems likely that we're spawning a lot of metric tasks in the agents, as it's part of the `build_ethereum` step. We not only do this ~15 times at startup in the relayer, but also when building metadata!

@daniel-savu pointed out tokio metrics we can start using that will also reveal this to us: https://tokio.rs/blog/2022-02-announcing-tokio-metrics

If we're actually continuing to spawn a task / multiple tasks whenever we build metadata, we'll need to refactor this or somehow enforce there's only a single task per chain

avious00:
@aroralanuk @tkporter - thinking of assigning this to Kunal since it's adjacent to the eigenlayer work he's been doing it lately. does that work?

daniel-savu:
It's not that adjacent imo, more about tokio than about prometheus

PR #3214, opened 2024-02-02
feat: fetch chain-specific metrics in standalone task

daniel-savu:
### Description

Spawning a tokio task to fetch chain-specific metrics in provider middleware was wasteful because we ended up with 15 tasks doing the same job. Now there is a single task that is spawned inside the relayer, which I added in the struct that used to only fetch relayer balance.

Some questions:
- I'm realising in the current state I'm probably removing these metrics from validators. Lmk if yes and I'll add back
- the chain-specific metrics were moved out of `MiddlewareMetrics` because they're no longer part of middleware, but I wasn't sure where to place them. Maybe `CoreMetrics` is a better place than the current `custom_metrics.rs` file?

`tokio-metrics` turned out not to be useful because we'd need to instrument every call site of `tokio::spawn` with it. We should still do it but as a separate task imo.

### Drive-by changes

- This PR also makes it very easy to add the metrics tasks for new chains, by extending the `HyperlaneProvider` trait with a `get_chain_metrics` call which seems reasonably general to me to have implemented by all providers. (Bc we currently only track these for evm chains)
  - the description, naming and logic of the `gas_price` metric was also changed to support new chains 

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3047

### Backward compatibility

Yes

### Testing

Still need to manually test


rust/hyperlane-base/src/metrics/custom_metrics.rs
```diff
@@ -108,12 +150,41 @@ impl AgentMetricsUpdater {
...
+            trace!("Block height for chain {chain} is {height}");
+            block_height
+                .with(&hashmap! { "chain" => chain })
+                .set(height);
+        }
+        if let Some(gas_price) = gas_price {
+            let protocol = self.conf.domain.domain_protocol();
+            let decimals_scale = 10f64.powf(decimals_by_protocol(protocol).into());
+            let gas = u256_as_scaled_f64(current_block.gas_price, protocol) * decimals_scale;
+            trace!("Gas price for chain {chain} is {gas:.1} (lowest denomination)");
```
tkporter, line 176: nit: now that we have e.g. Neutron where we can have a gas price of 0.57 neutron, wonder if we should change the 0.1 here to 0.2

rust/hyperlane-base/src/metrics/custom_metrics.rs
```diff
@@ -108,12 +150,41 @@ impl AgentMetricsUpdater {
...
+    async fn update_block_details(&self) {
+        let block_height = self.chain_metrics.block_height.clone();
+        let gas_price = self.chain_metrics.gas_price.clone();
+        if let HyperlaneDomain::Unknown { .. } = self.conf.domain {
+            return;
+        };
+        let chain = self.conf.domain.name();
+        debug!("Updating metrics for chain ({:?})", chain);
+        let Ok(Some(current_block)) = self.provider.get_chain_metrics().await else {
+            return;
```
tkporter, line 162: maybe worth a log on Err

rust/hyperlane-core/src/traits/provider.rs
```diff
@@ -27,6 +27,9 @@ pub trait HyperlaneProvider: HyperlaneChain + Send + Sync + Debug {
 
     /// Fetch the balance of the wallet address associated with the chain provider.
     async fn get_balance(&self, address: String) -> ChainResult<U256>;
+
+    /// Fetch the balance of the wallet address associated with the chain provider.
```
tkporter, line 31: nit: change comment

rust/hyperlane-core/src/types/chain_data.rs
```diff
@@ -11,6 +13,15 @@ pub struct BlockInfo {
     pub number: u64,
 }
 
+/// Metrics about the chain.
+#[derive(Debug, Clone, Default, new)]
+pub struct ChainInfo {
+    /// Information about the latest block
+    pub latest_block: BlockInfo,
+    /// The current gas price, in the lowest denomination (e.g. gwei)
```
tkporter, line 21: nit: gwei -> wei if we're doing lowest denomination now

rust/hyperlane-base/src/metrics/custom_metrics.rs
```diff
@@ -3,12 +3,15 @@ use std::time::Duration;
 use derive_builder::Builder;
```
tkporter, line 3: curious why custom? don't feel strongly but wondering if there's a more accurate name closer to "chain" or "agent" or something

tkporter (commented):
I think you're right that this removes the metrics from the validator and scraper. is there a way to spawn the metric updater task in agent agnostic code?

rust/hyperlane-base/src/metrics/custom_metrics.rs
```diff
@@ -52,6 +66,34 @@ pub(crate) fn create_agent_metrics(metrics: &CoreMetrics) -> Result<AgentMetrics
...
+#[derive(Clone, Builder)]
+pub struct ChainMetrics {
+    /// Tracks the current block height of the chain.
+    /// - `chain`: the chain name (or ID if the name is unknown) of the chain
+    ///   the block number refers to.
+    #[builder(setter(into, strip_option), default)]
+    pub block_height: Option<IntGaugeVec>,
+
+    /// Tracks the current gas price of the chain. Uses the base_fee_per_gas if
+    /// available or else the median of the transactions.
```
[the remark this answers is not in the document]
daniel-savu: will use None and add a TODO comment about the median

rust/hyperlane-base/src/metrics/custom_metrics.rs
daniel-savu: changed to `agent_metrics.rs` and added a file-level doc comment: `//! Metrics either related to the agents, or observed by them`

rust/hyperlane-core/src/types/chain_data.rs
daniel-savu: my assumption is that we are using low denomination, since I'm now scaling with `decimals_by_protocol(...)`, let me know if not

rust/hyperlane-base/src/metrics/custom_metrics.rs
```diff
@@ -52,6 +66,34 @@ pub(crate) fn create_agent_metrics(metrics: &CoreMetrics) -> Result<AgentMetrics
...
 }
 
+/// Chain-specific metrics
+#[derive(Clone, Builder)]
+pub struct ChainMetrics {
+    /// Tracks the current block height of the chain.
+    /// - `chain`: the chain name (or ID if the name is unknown) of the chain
+    ///   the block number refers to.
+    #[builder(setter(into, strip_option), default)]
+    pub block_height: Option<IntGaugeVec>,
```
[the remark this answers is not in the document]
tkporter: bump on this, happy to leave it optional if you feel it's useful though

rust/chains/hyperlane-ethereum/src/provider.rs
```diff
@@ -119,6 +116,37 @@ where
...
+        let chain_metrics = ChainInfo::new(
+            BlockInfo {
+                hash: block.hash.unwrap().into(),
+                timestamp: block.timestamp.as_u64(),
+                number: block.number.unwrap().as_u64(),
+            },
+            Some(
+                block
+                    .base_fee_per_gas
+                    .ok_or(HyperlaneEthereumError::MissingBlockDetails)?
```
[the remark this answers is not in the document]
daniel-savu: oops

rust/chains/hyperlane-ethereum/src/provider.rs
```diff
@@ -119,6 +116,37 @@ where
...
+            .get_block(BlockNumber::Latest)
+            .await
+            .map_err(|e| {
+                ChainCommunicationError::Other(HyperlaneCustomErrorWrapper::new(Box::new(e)))
+            })?
+        else {
+            return Ok(None);
+        };
+
+        // Since `block` is Some at this point, we're guaranteed to have its `hash` and `number` defined,
```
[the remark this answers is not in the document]
daniel-savu: added a note to clarify

rust/hyperlane-base/src/metrics/custom_metrics.rs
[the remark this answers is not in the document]
daniel-savu: ah, misread as only `gas_price`. I removed the option from `block_height`, but since we're not implementing the gas_price median we have to keep that as optional

PR #3219, opened 2024-02-03
reduce max sequence range to 5

nambrot:

### DIFF
--- a/typescript/infra/config/environments/mainnet2/agent.ts
+++ b/typescript/infra/config/environments/mainnet2/agent.ts
@@ -127,7 +127,7 @@ const hyperlane: RootAgentConfig = {
     connectionType: AgentConnectionType.HttpFallback,
     docker: {
       repo,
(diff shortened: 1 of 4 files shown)


typescript/infra/config/environments/mainnet2/agent.ts
```diff
@@ -127,7 +127,7 @@ const hyperlane: RootAgentConfig = {
     connectionType: AgentConnectionType.HttpFallback,
     docker: {
       repo,
-      tag: '125a197-20231204-113659',
+      tag: '49a581b-20240203-151524',
```
tkporter, line 130: can you update the Solana validator too? line 148

tkporter (approved):

typescript/infra/config/environments/mainnet2/agent.ts
nambrot: Updated but not deployed yet

PR #3227, opened 2024-02-05
fix: add fallback to kurtosis terminal link

paulbalaji:
Users can run into an issue where the URL to deploy to kurtosis does not display correctly for unsupported terminals. I have ran into this myself with my mac setup.

It manifests as:
- "click here" text, no URL
- unformatted text with URL containing "no-width space" characters

Related issue: https://github.com/sindresorhus/terminal-link/issues/18#issuecomment-1068020361

As suggested by a comment on the above issue, we can add a fallback to the `terminalLink` call to simply print the URL as a backup.


jmrossy (approved):
Nice

PR #3230, opened 2024-02-05
Add block pinned caching to env tests

yorhodes:
### Description

- caches foundry forked RPCs to avoid flakiness in CI

### Drive-by changes

- requires manual block number changes when we make onchain changes

### Testing

- fork tests


nambrot (approved):

yorhodes:
closing as the historical pin is too flaky for most public RPCs :/

PR #3237, opened 2024-02-07
fix:populate igp in agent config from cli regardless of hook config

aroralanuk:
### Description

- add igp as the zero address at the last step right writing it to file

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1030

### Backward compatibility

Yes

### Testing

e2e


.github/workflows/test.yml
```diff
@@ -236,8 +236,72 @@ jobs:
...
-      - name: cli e2e tests
-        run: ./typescript/cli/ci-test.sh
+  cli-e2e:
+    runs-on: larger-runner
+    needs: [yarn-build]
+    strategy:
+      matrix:
+        include:
+          - hook: false
+          - hook: true
```
yorhodes, line 246: imo it would be nice to configure this with a hook config file name
as it stands its hard to know what the hook: true means

typescript/cli/src/deploy/core.ts
```diff
@@ -484,6 +484,10 @@ async function writeAgentConfig(
     artifacts,
   ) as ChainMap<HyperlaneDeploymentArtifacts>;
 
+  for (const chain of chains) {
+    mergedAddressesMap[chain].interchainGasPaymaster =
+      ethers.constants.AddressZero;
+  }
```
yorhodes, line 490: seems like we only want to do this when they didn't provide any IGP config, or something along those lines

yorhodes (approved):

.github/workflows/test.yml
```diff
@@ -236,8 +236,72 @@ jobs:
           E2E_CI_TIMEOUT_SEC: '600'
           E2E_KATHY_MESSAGES: '20'
 
-      - name: cli e2e tests
-        run: ./typescript/cli/ci-test.sh
+  cli-e2e:
+    runs-on: larger-runner
+    needs: [yarn-build]
+    strategy:
```
jmrossy, line 242: This is a lot of extra CI code just to run test with two values for hook. Can you add this logic into the CI script instead of here? I've found CI configs are brittle and doesn't get maintained very well so I'd rather keep it simple

typescript/cli/src/deploy/core.ts
jmrossy: Agreed, what's the value of always including a zero address?

jmrossy (changes requested):

.github/workflows/test.yml
aroralanuk: for another issue for now

.github/workflows/test.yml
aroralanuk: fixed

jmrossy (approved):

Issue #3234, opened 2024-02-07
Fix `TxSearch` with block-by-block indexing


avious00:
**Fix Neutron Validator Indexing for `TxSearch` with block-by-block indexing**

### Problem
We need to design a sustainable method for Cosmos indexing, `TxSearch` is unreliable.
- it can give back junk responses depending on the timeout of the RPC server
- we require an archive node to index older messages
- sometimes TxSearch isn’t public because it’s such an expensive RPC
- Results in neutron validators getting stuck or not being able to index older messages, so we have gaps in checkpoints

### Solution
- We'll go with block-by-block indexing
- The original [get_event_log](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2865/commits/6770bb7574250c1509c8a86bc4d78a1bda23dedb#diff-6c9ba1f6544e14bac471dbff2c001c80edf0493064a1ab0695d78bbe9c9cd6e5L87-L165) function from Mitosis can serve as an example. Might need some prettier logic
- It doesn't appear a single RPC will give us _both_ tx hashes and the events from those txs 
- Following Mitosis's work, we need to use the `/block` endpoint to get tx hashes, and then the `/block_results` endpoint to get the events for txs. The latter endpoint just returns an array where each element contains the events of the tx in the corresponding element in the array of txs from `/block`, but no tx hash. Also `/block` doesn't actually return the TX hash - it returns the raw tx that you then need to hash to get the tx hash.
  - References to this:
    - proof that the order of txs in `/block` and `/block_results` are the same https://github.com/tendermint/tendermint/issues/3866
    - Getting the hash from the raw tx https://github.com/cosmos/cosmos-sdk/issues/3729#issuecomment-482503141
  - Example RPCs:
    - `/block` https://rpc-magnetix.neutron-1.neutron.org/block?height=6162253
    - `/block_results` https://rpc-magnetix.neutron-1.neutron.org/block_results?height=6162253
    - tx in question in that block (note this is a message getting *processed*, not dispatched) https://www.mintscan.io/neutron/tx/336CE709DA9BEC52C1C9A956267611115B5A80C26334AE154DD85A8EFB0E4E6F?height=6162253
[...]

PR #3245, opened 2024-02-11
Cosmos block-by-block indexing

daniel-savu:
### Description

Implements cosmos block-by-block indexing, parallelizing where possible: the `block` and `block_results` calls, and the individual block queries within a range. Also reuses the existing `handle_txs` logic, to improve readability.

### Drive-by changes

Adds `tokio::task::JoinError` to the `hyperlane-core` error enum

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3234
- Added some hacky retrying logic to `get_logs_in_block`, since we were seeing lots of `503` http errors. In https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3264 we should make sure to remove this and propagate the error instead, to retry via the cursor / indexer logic. I edited the description of 3264 to reflect this

### Backward compatibility

Yes

### Testing

e2e


rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -203,10 +205,33 @@ impl Indexer<InterchainGasPayment> for CosmosInterchainGasPaymasterIndexer {
...
+                    self_clone
+                        .indexer
+                        .get_event_log(block_number, Self::interchain_gas_payment_parser)
+                        .await
+                })
+            })
+            .collect();
+
+        // TODO: this can be refactored when we rework indexing, to be part of the block-by-block indexing
+        let result = future::join_all(logs_futures)
```
daniel-savu, line 221: this fn implementation may be prematurely optimized. Do you think we could get rate limited by querying in parallel @tkporter?

rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -352,10 +354,32 @@ impl Indexer<HyperlaneMessage> for CosmosMailboxIndexer {
...
+
+        // TODO: this can be refactored when we rework indexing, to be part of the block-by-block indexing
+        let result = future::join_all(logs_futures)
+            .await
+            .into_iter()
+            .flatten()
+            .filter_map(|res| match res {
+                Ok(logs) => Some(logs),
+                Err(err) => {
+                    warn!(?err, "Failed to fetch logs");
```
tkporter, line 377: the nice part here is for the message & merkle tree indexing we'll automatically rewind to try these again, so less concerned here

rust/chains/hyperlane-cosmos/src/payloads/general.rs
```diff
@@ -20,15 +25,16 @@ pub struct Event {
 pub struct EventAttribute {
     pub key: String,
     pub value: String,
-    pub index: bool,
```
tkporter, line 23: were you getting errors with this still there?

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -1,29 +1,27 @@
...
-
 #[async_trait]
 /// Trait for wasm indexer. Use rpc provider
 pub trait WasmIndexer: Send + Sync {
     /// Get the finalized block height.
     async fn get_finalized_block_number(&self) -> ChainResult<u32>;
 
     /// Get logs for the given range using the given parser.
-    async fn get_range_event_logs<T>(
+    async fn get_event_log<T>(
```
tkporter, line 22: nit: iirc this was the name from mitosis before but it feels a bit off, maybe `get_logs_in_block` or something? also need to update the comment above

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -128,22 +153,26 @@ impl CosmosWasmIndexer {
     // made by the contract we are indexing.
     fn handle_tx<T>(
         &self,
-        tx: tx::Response,
+        block: BlockResponse,
+        tx_events: Events,
+        tx_hash: H256,
+        block_number: u32,
```
tkporter, line 159: I think block_number is available somewhere in `block`

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -128,22 +153,26 @@ impl CosmosWasmIndexer {
...
+        transaction_index: usize,
         parser: for<'a> fn(&'a Vec<EventAttribute>) -> ChainResult<ParsedEvent<T>>,
     ) -> impl Iterator<Item = (T, LogMeta)> + '_
     where
         T: PartialEq + 'static,
     {
-        tx.tx_result.events.into_iter().enumerate().filter_map(move |(log_idx, event)| {
-            if event.kind.as_str() != self.target_event_kind {
+        tx_events.events.into_iter().enumerate().filter_map(move |(log_idx, event)| {
+            if event.typ.as_str() != self.target_event_kind || !event.typ.as_str().starts_with(Self::WASM_TYPE) {
```
tkporter, line 167: why was the or condition needed?

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -128,22 +153,26 @@ impl CosmosWasmIndexer {
...
                 return None;
             }
 
-            parser(&event.attributes)
+            parser(&event.attributes.clone())
```
tkporter, line 171: is the clone needed?

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -190,47 +217,25 @@ impl WasmIndexer for CosmosWasmIndexer {
...
-            .handle_txs(tx_search_result.txs, parser)?
-            .collect::<Vec<_>>();
+        let client = self.provider.rpc().clone();
+        let block_results_fut =
+            tokio::spawn(async move { client.block_results(block_number).await });
 
-        // If there are any more pages, fetch them and append to the result.
-        for page in 2..=last_page {
-            trace!(page, "Performing tx search");
+        let (block_res, block_results_res) = tokio::join!(block_fut, block_results_fut);
```
tkporter, line 235: for my own learning, is there any difference between this approach of spawning 2 tasks and joining vs this? does the following just spawn tasks under the hood?
```
let client1 = self.provider.rpc().clone();
let client2 = self.provider.rpc().clone();
let (block_res, block_results_res) = tokio::join!(
    client1.block(block_number),
    client2.block_results(block_number),
);
```

rust/hyperlane-core/src/error.rs
```diff
@@ -134,6 +134,10 @@ pub enum ChainCommunicationError {
     /// Rpc client error
     #[error(transparent)]
     RpcClientError(#[from] RpcClientError),
+    /// Tokio join error
+    #[cfg(feature = "fallback-provider")]
```
tkporter, line 138: kinda awkward to have this feature relate to something that's used outside of fallback provider. maybe a case for a rename of the feature? don't feel strongly

tkporter (approved):
We should figure out a plan for releasing this. I think we'll wanna keep it on rc for a bit and see what happens. Also realizing that releasing to validators even with the current willingness to miss events if an error getting a block occurred is probably pretty safe because they don't care about IGP payments, just merkle tree insertions / messages which both benefit from knowing when they skip a message

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -203,10 +205,33 @@ impl Indexer<InterchainGasPayment> for CosmosInterchainGasPaymasterIndexer {
...
+            .await
+            .into_iter()
+            .flatten()
+            .filter_map(|res| match res {
+                Ok(logs) => Some(logs),
+                Err(err) => {
+                    warn!(?err, "Failed to fetch logs");
+                    None
```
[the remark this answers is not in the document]
daniel-savu: > Will this warn tell us which block number failed too?

added

> does make me wonder if we should wait to ship this until we have a small indexing refactor

happy to wait for your indexing refactor PR to go in before merging this one

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -86,40 +89,63 @@ impl CosmosWasmIndexer {
...
-                    trace!(tx_hash=?tx.hash, "Indexed tx has failed, skipping");
+            .enumerate()
+            .filter_map(move |(idx, tx)| {
+                let tx_hash = tx_hashes[idx];
+                if tx.code.is_err() {
+                    debug!(?tx_hash, "tx has failed. skipping");
+                    return None;
                 }
-                !tx_failed
+                let Ok(logs) = serde_json::from_str::<Vec<Events>>(&tx.log) else {
```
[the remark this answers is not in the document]
daniel-savu: Hmm I assume it was a preference for types defined in this crate for trait impl purposes, but I'll switch to avoiding the json parse step

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
daniel-savu: it was there initially and I kept it for completeness

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
daniel-savu: simplified - didn't know it could be done this way. It just [polls](https://docs.rs/tokio/latest/src/tokio/macros/join.rs.html#126) every item in the `join!` as a future, until completion. There is a [wrapper](https://docs.rs/tokio/latest/src/tokio/future/maybe_done.rs.html#13) around the items but it's very light

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
daniel-savu: not anymore it seems

rust/hyperlane-core/src/error.rs
daniel-savu: renamed the feature to `async`

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -86,64 +89,85 @@ impl CosmosWasmIndexer {
...
-                    trace!(tx_hash=?tx.hash, "Indexed tx has failed, skipping");
+            .enumerate()
+            .filter_map(move |(idx, tx)| {
+                let tx_hash = tx_hashes.get(idx);
+                if tx.code.is_err() {
+                    debug!(?tx_hash, "Not indexing failed transaction");
+                    return None;
                 }
-                !tx_failed
+                tx_hash.map(|tx_hash| {
```
tkporter, line 130: maybe worth a log if tx_hash is none?

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
tkporter: I'm not so sure we should keep it, iiuc it results in us trying to parse any wasm event? even though we have later checks to make sure the contract address of the event is correct, I think it's potentially unsafe - e.g. if the contract emits a couple different events with the same structure but different name, I think we may risk parsing undesired events

tkporter (commented):
lgtm so far

rust/chains/hyperlane-cosmos/src/mailbox.rs
daniel-savu: Should I remove the warning?

rust/chains/hyperlane-cosmos/src/payloads/general.rs
daniel-savu: I was yes, but it's no longer an issue now that I'm no longer json-parsing, so this can stay the same

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -86,40 +89,63 @@ impl CosmosWasmIndexer {
...
+
+        let tx_hashes: Vec<H256> = block
+            .clone()
+            .block
+            .data
+            .into_iter()
+            .filter_map(|tx| hex::decode(digest(tx.as_slice())).ok())
+            .map(|hash| {
+                H256::from_slice(
+                    Hash::from_bytes(Algorithm::Sha256, hash.as_slice())
```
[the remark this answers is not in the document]
daniel-savu: :( yeah we'll have to stick with this, replies in this thread compute hashes manually as well: https://github.com/cosmos/cosmos-sdk/issues/3217#issuecomment-451154964

daniel-savu:
Successfully tested the relayer on mainnet3 RC
- neutron message id: `0x7e4c607d8a7d8dff892e71912feed121c5192225d89b7dff6ccc18043534f3d2`
- injective message id: `0xa3105155269af226768818d5626236f080732221373193f3eac9aa2188927047 `

rust/chains/hyperlane-cosmos/src/mailbox.rs
tkporter: No I think the warning is useful, can you also add the block num here for consistency with the IGP indexing?

rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
```diff
@@ -285,10 +286,32 @@ impl Indexer<MerkleTreeInsertion> for CosmosMerkleTreeHookIndexer {
...
+
+        // TODO: this can be refactored when we rework indexing, to be part of the block-by-block indexing
+        let result = future::join_all(logs_futures)
+            .await
+            .into_iter()
+            .flatten()
+            .filter_map(|res| match res {
+                Ok(logs) => Some(logs),
+                Err(err) => {
+                    warn!(?err, "Failed to fetch logs");
```
tkporter, line 309: ditto block num is useful imo

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -86,55 +89,68 @@ impl CosmosWasmIndexer {
...
     where
         T: PartialEq + 'static,
     {
-        let logs_iter = txs
+        let Some(tx_results) = block_results.txs_results else {
+            return Ok(vec![]);
+        };
+
+        let tx_hashes: Vec<H256> = block
+            .clone()
```
tkporter, line 108: wondering if the clone here and line 132 are needed / if BlockResponse needs to be owned, iirc cosmos blocks can be MB in size. Not a big deal tho so feel free to leave as is

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -190,47 +204,31 @@ impl WasmIndexer for CosmosWasmIndexer {
...
-        }
-
-        Ok(logs)
+        let client = self.provider.rpc().clone();
+
+        let (block_res, block_results_res) = tokio::join!(
+            client.block(block_number),
+            client.block_results(block_number)
+        );
+        println!(
```
tkporter, line 221: rm

rust/chains/hyperlane-cosmos/src/interchain_gas.rs
```diff
@@ -203,10 +205,34 @@ impl Indexer<InterchainGasPayment> for CosmosInterchainGasPaymasterIndexer {
...
+
+        // TODO: this can be refactored when we rework indexing, to be part of the block-by-block indexing
+        let result = future::join_all(logs_futures)
+            .await
+            .into_iter()
+            .flatten()
+            .filter_map(|(res, block_no)| match res {
+                Ok(logs) => Some(logs),
+                Err(err) => {
+                    warn!(?err, ?block_no, "Failed to fetch logs for block");
```
tkporter, line 229: noting that we may wanna change this to not allow Errs when we do https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3264 - but for now happy to keep this as-is

typescript/infra/src/agents/key-utils.ts
```diff
@@ -386,8 +386,8 @@ async function persistAddressesLocally(
   let relayer, kathy;
   for (const key of keys) {
     if (key.role === Role.Relayer) {
-      if (relayer)
-        throw new Error('More than one Relayer found in gcpCloudAgentKeys');
+      // if (relayer)
```
tkporter, line 389: for posterity I created https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3328 that I'll flesh out

tkporter (approved):
🚢 🚀

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
daniel-savu: all required sadly

typescript/infra/config/environments/mainnet3/tokenPrices.json
```diff
@@ -14,5 +14,6 @@
   "polygonzkevm": "2914.61",
   "inevm": "32.77",
   "viction": "0.750231",
+  "neutron": "0.750231",
```
tkporter, line 17:
```suggestion
  "neutron": "1.76",
```

tkporter: from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3327 when I ran the print token prices script

tkporter (approved):
🚢

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -83,58 +91,117 @@ impl CosmosWasmIndexer {
...
+        let mut retries: u32 = 5;
+        for _ in 1..retries {
+            match f().await {
+                Ok(res) => return Ok(res),
+                Err(err) => match retries.checked_sub(1) {
+                    Some(remaining) => retries = remaining,
+                    None => break,
+                },
+            }
+            debug!("Retrying call");
```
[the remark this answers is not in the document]
daniel-savu: It's a bit annoying to return the last error, and since we'll eventually refactor this into a retrying provider I'll look into it then (added a todo). Added the suggestions to the `debug` 🫡

PR #3249, opened 2024-02-12
Improve cli send transfer command

yorhodes:
### Description

- supports synthetic warp routes
- fixes bug in collateral balance checking
- deduces token balance checks from provided router address

### Testing

- Manual and CLI CI test

### DIFF
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -76,11 +76,16 @@ export async function sendTestTransfer({
   }
 
   if (tokenType === TokenType.collateral) {
+    const collateralRouter = HypERC20Collateral__factory.connect(
+      routerAddress,
+      signer,
+    );
+    const tokenAddress = await collateralRouter.wrappedToken();
     await assertTokenBalance(
       multiProvider,
       signer,
       origin,
-      routerAddress,
+      tokenAddress,
       wei.toString(),
     );
   } else if (tokenType === TokenType.native) {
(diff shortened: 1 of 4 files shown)


aroralanuk (approved):

typescript/cli/src/send/transfer.ts
```diff
@@ -75,31 +74,37 @@ export async function sendTestTransfer({
...
-    );
-    const tokenAddress = await collateralRouter.wrappedToken();
+  // TODO: move to SDK token router app
+  // deduce TokenType
+  // 1. decimals() call implies synthetic
+  // 2. wrappedToken() call implies collateral
+  // 3. if neither, it's native
+  let tokenAddress: Address | undefined;
+  const provider = multiProvider.getProvider(origin);
+  try {
```
aroralanuk, line 85: nit: not a big fan of the nested try catch logic here but I don't know of any other alternatives

typescript/cli/src/send/transfer.ts
```diff
@@ -75,31 +74,37 @@ export async function sendTestTransfer({
...
+      const tokenRouter = HypERC20Collateral__factory.connect(
+        routerAddress,
+        provider,
+      );
+      tokenAddress = await tokenRouter.wrappedToken();
+    } catch (error) {}
+  }
+
+  if (tokenAddress) {
     await assertTokenBalance(
```
aroralanuk, line 104: add a comment suggesting this covers both the synthetic and collateral cases

aroralanuk (approved):

jmrossy (approved):
Could be a neater but this can probably use the WarpCore soon

typescript/cli/src/send/transfer.ts
yorhodes: me neither tbh

PR #3255, opened 2024-02-13
feat: verify with deployment

paulbalaji:
Do note that a big chunk of the diff is just splitting it out `utils` from the `HyperlaneIsmFactory`

---
**ContractVerifier refactoring**
- refactor ContractVerifier to handle a specific contract + chain
	- update access of functions, simplify required calls
	- strongly type API arguments depending on the action type
	- initialise ContractVerifier by parsing input json to construct a map of contractName -> sourceName
- create `PostDeploymentContractVerifier` to handle iteration over several verification inputs
---
**Adding core package build artifact**
- export build artifact in core package, only containing relevant fields
	- setup default ContractVerifier to use the exported build artifact
---
**Misc improvements**
- use `debug extend` to improve/simplify logging
- upstream args required for verification, re-use in verify/deploy.ts 
---
**First pass of updating IsmFactory to leverage existing deployer functions**
- update part of ismfactory to use the HyperlaneDeployer's deploy flow
	- we now get verification for free for pausable/op/test ism


typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,167 +7,109 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
 
   constructor(
-    verificationInputs: ChainMap<VerificationInput>,
-    protected readonly multiProvider: MultiProvider,
-    protected readonly apiKeys: ChainMap<string>,
-    protected readonly source: string, // solidity standard input json
-    protected readonly compilerOptions: CompilerOptions,
+    private multiProvider: MultiProvider,
+    private readonly apiKeys: ChainMap<string>,
+    private source: string, // solidity standard input json
```
paulbalaji, line 27: jmrossy
> Why make these private? I don't think we need member data protection but it's conceivable that we'd want to extend this class one day

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,167 +7,109 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
-  ): Promise<void> {
-    this.logger(`Verifying ${chain}...`);
-    for (const input of inputs) {
-      await this.verifyContract(chain, input);
-    }
+    this.compilerOptions = {
+      codeformat: 'solidity-standard-json-input',
+      compilerversion:
+        compilerOptions?.compilerversion ?? 'v0.8.19+commit.7dd6d404',
+      licenseType: compilerOptions?.licenseType ?? ExplorerLicenseType.MIT,
```
paulbalaji, line 34: jmrossy
> Nit, is it better to assume MIT or just have this field be optional?

paulbalaji (commented):
copied across rossy's comments

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: I kind of went on a tear trying to lock down the class interface, but you're right that we should keep this protected readonly

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -350,6 +352,12 @@ export abstract class HyperlaneDeployer<
     );
     this.addVerificationArtifacts(chain, [verificationInput]);
 
+    // if configured, verify contract
+    await this.options?.contractVerifier?.verifyContract(
+      chain,
+      verificationInput,
+    );
```
paulbalaji, line 359: once piped through, this is all that's required during deployment!

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: I had it fallback to MIT to minimise config but it's probably better to leave it blank as it's optional on explorer APIs

typescript/helloworld/hardhat.config.ts
paulbalaji, line 1: At this point we should just have the helloworld contract with the rest of the solidity right?

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,226 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
 
-  async verifyChain(
-    chain: ChainName,
-    inputs: VerificationInput,
-  ): Promise<void> {
-    this.logger(`Verifying ${chain}...`);
-    for (const input of inputs) {
-      await this.verifyContract(chain, input);
-    }
+    console.log(this.contractMap);
```
paulbalaji, line 54: TODO: remove

typescript/sdk/src/ism/utils.ts
paulbalaji, line 1: big chunk of the diff is just splitting out this file

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
-            case 'apikey':
-              break;
-            default:
-              formData.append(key, value);
-              break;
-          }
-        }
+    const params = new URLSearchParams();
+    params.set('module', 'contract');
+    params.set('action', action);
```
yorhodes, line 77: this is much better :)

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
-          headers: {
-            'Content-Type': 'application/x-www-form-urlencoded',
-          },
-          body: params,
-        });
+    // no need to provide every argument for every request
+    if (options) {
+      for (const [key, value] of Object.entries(options)) {
+        params.set(key, value);
       }
```
yorhodes, line 83: I like `.forEach` here but nbd

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
-      responseText = await response.text();
-      result = JSON.parse(responseText);
-    } catch (e) {
-      this.logger(
-        `[${chain}] Failed to parse response from ${responseText}`,
-        e,
-      );
+    // only include apikey if provided & not blockscout
+    if (family !== ExplorerFamily.Blockscout && this.apiKeys[chain]) {
+      params.set('apikey', this.apiKeys[chain]);
```
yorhodes, line 88: interesting, does blockscout not have API keys?

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
     }
+
+    const url = new URL(apiUrl);
+    const isGetRequest = [
+      ExplorerApiActions.CHECK_STATUS,
+      ExplorerApiActions.CHECK_PROXY_STATUS,
+      ExplorerApiActions.GETSOURCECODE,
+    ].includes(action);
```
yorhodes, line 96: pull this out into a constant maybe?
`const GET_ACTIONS = [
      ExplorerApiActions.CHECK_STATUS,
      ExplorerApiActions.CHECK_PROXY_STATUS,
      ExplorerApiActions.GETSOURCECODE
];`

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
+    if (isGetRequest) {
+      url.search = params.toString();
+    } else if (family === ExplorerFamily.Blockscout) {
+      // Blockscout requires module and action to be query params
+      url.searchParams.set('module', 'contract');
+      url.searchParams.set('action', action);
+    }
```
yorhodes, line 103: interesting, I'm not sure I follow this

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
+
+    const response = await fetch(url.toString(), {
+      method: isGetRequest ? 'GET' : 'POST',
+      headers: isGetRequest
+        ? {}
+        : { 'Content-Type': 'application/x-www-form-urlencoded' },
+      body: isGetRequest ? null : params,
+    });
```
yorhodes, line 111: maybe we should just separate the control flow for clarity instead of ternary

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
+        throw new Error(`[${chain}] ${errorMessage}`);
       }
     }
 
+    if (result.result === ExplorerApiErrors.UNKNOWN_UID) {
+      await sleep(1000); // wait 1 second
+      return this.submitForm(chain, action, verificationLogger, options);
+    }
+
+    if (result.result === ExplorerApiErrors.UNABLE_TO_VERIFY) {
```
yorhodes, line 151: else if?

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
+      const errorMessage = `Contract '${input.name}' not found in provided build artifact`;
+      verificationLogger(errorMessage);
+      throw new Error(`[${chain}] ${errorMessage}`);
+    }
 
     const data = {
-      sourceCode: this.source,
-      contractname: input.name,
+      sourceCode: this.standardInputJson,
+      contractname: `${sourceName}:${input.name}`,
```
yorhodes, line 240: interesting
do we need the name on the input separately?

typescript/sdk/src/deploy/verify/PostDeploymentContractVerifier.ts
yorhodes, line 1: why pull this out of ContractVerifier?

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
```diff
@@ -42,10 +43,15 @@ export class HyperlaneHookDeployer extends HyperlaneDeployer<
     multiProvider: MultiProvider,
     readonly core: ChainMap<Partial<CoreAddresses>>,
     readonly ismFactory: HyperlaneIsmFactory,
-    readonly igpDeployer = new HyperlaneIgpDeployer(multiProvider),
+    contractVerifier?: ContractVerifier,
+    readonly igpDeployer = new HyperlaneIgpDeployer(
+      multiProvider,
+      contractVerifier,
+    ),
```
yorhodes, line 50: why reorder here?

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -54,20 +46,42 @@ import {
...
 } from './types';
+import { routingModuleDelta } from './utils';
 
 export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
   // The shape of this object is `ChainMap<Address | ChainMap<Address>`,
   // although `any` is use here because that type breaks a lot of signatures.
   // TODO: fix this in the next refactoring
   public deployedIsms: ChainMap<any> = {};
 
+  // upon initialization, HyperlaneDeployer will bind itself to HyperlaneIsmFactory
```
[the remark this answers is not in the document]
yorhodes: could we alternatively have HyperlaneIsmFactory extend HyperlaneDeployer?

typescript/sdk/src/ism/utils.ts
yorhodes, line 1: I assume this is just a plain move
in future would be nice to avoid big diff drivebys like this

typescript/sdk/src/token/deploy.ts
yorhodes, line 1: kind of annoying we have to pipe these options through everywhere

typescript/sdk/src/ism/utils.ts
paulbalaji: original plan was to put the ism stuff into a separate PR but then we decided to cut it and keep low hanging fruit like this

solidity/exportBuildArtifact.sh
```diff
@@ -0,0 +1,27 @@
...
+# Define the artifacts directory
+artifactsDir="./artifacts/build-info"
+# Define the output file
+outputFile="./buildArtifact.json"
+
+# log that we're in the script
+echo 'Finding and processing hardhat build artifact...'
+
+# Find the latest JSON build artifact
+jsonFiles=$(find "$artifactsDir" -type f -name "*.json" | sort | tail -n 1)
```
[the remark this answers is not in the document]
paulbalaji: sorting even by last modification might not be correct here, should probably remove `sort` altogether. if someone's having issues I think the modus operandi should be `yarn clean` before `yarn build`

given the scenario of 3 builds as follows:
build1 - use contracts v1
build2 - use contracts v2
build3 - use contracts v1

build3 does not generate any new artifacts because the hash of contracts would be the same as build1

in ci/npm publishing this all works fine because we do a clean build so there'll only be one

typescript/infra/scripts/agent-utils.ts
```diff
@@ -133,6 +134,13 @@ export function withMissingChains<T>(args: yargs.Argv<T>) {
     .alias('n', 'newChains');
 }
 
+export function withBuildArtifactPath<T>(args: yargs.Argv<T>) {
```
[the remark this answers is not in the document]
paulbalaji: I wans't sure of the best place, so I put it where I saw similar helpers used by the deploy script

typescript/infra/scripts/deploy.ts
```diff
@@ -61,18 +71,40 @@ async function main() {
...
+    // extract build artifact contents
+    const buildArtifact = extractBuildArtifact(buildArtifactPath);
+    // instantiate verifier
+    contractVerifier = new ContractVerifier(
+      multiProvider,
+      apiKeys,
+      buildArtifact,
+      ExplorerLicenseType.MIT,
+    );
+  }
```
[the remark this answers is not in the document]
paulbalaji: this flag is only really used when deploying helloworld because that is a different build artifact to the default core one

if you don't pass the artifact it'll fallback to the default contractverifier in the deployers - albeit without api keys

could refactor the deployer args maybe to set the apikeys on a default deployer?

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -141,22 +160,30 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
           config,
           origin,
           mailbox,
+          logger,
         });
         break;
       case IsmType.OP_STACK:
-        contract = await this.deployOpStackIsm(destination, config);
+        contract = await this.deployContractFromFactory(
```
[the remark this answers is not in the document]
paulbalaji: hence the binding

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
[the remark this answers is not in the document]
paulbalaji: probably could, would need to create a new hyperlaneapp as a field inside this factory class and update refs but doable

decided to go with binding because I wasn't sure of the ramifications of changing this class from an app -> deployer

typescript/sdk/src/deploy/verify/PostDeploymentContractVerifier.ts
paulbalaji: separation of concerns and also so I don't have to have it extend MultiGeneric when most of the logic there doesn't require it

want to keep ContractVerifier simple and use-case agnostic - it doesn't care if you're using it as you deploy or after you've deployed

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: yup the expected input on the explorer end is a fully qualified name, something like  `solidity/tokens/Tokens.sol:HypERC20`

this is very much an etherscan specific thing, blockscout don't even require you to provide a name

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: we're returning in the previous if anyway

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -7,243 +7,237 @@ import { sleep, strip0x } from '@hyperlane-xyz/utils';
...
-  ALREADY_VERIFIED_ALT = 'Already Verified',
-  VERIFICATION_PENDING = 'Pending in queue',
-  PROXY_FAILED = 'A corresponding implementation contract was unfortunately not detected for the proxy address.',
-  BYTECODE_MISMATCH = 'Fail - Unable to verify. Compiled contract deployment bytecode does NOT match the transaction deployment bytecode.',
-}
+  private contractMap: { [contractName: string]: string } = {};
 
-export class ContractVerifier extends MultiGeneric<VerificationInput> {
-  protected logger: Debugger;
+  protected readonly standardInputJson: string;
```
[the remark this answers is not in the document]
paulbalaji: we've already extracted everything we need from the json in the constructor, so I thought there's no need to do the json->string step on every request

typescript/sdk/src/deploy/verify/types.ts
```diff
@@ -7,22 +7,83 @@ export type ContractVerificationInput = {
...
+  sources: {
+    [sourceName: string]: {
+      content: string;
+    };
+  };
+};
+
+export type BuildArtifact = {
+  input: SolidityStandardJsonInput;
+  solcLongVersion: string;
```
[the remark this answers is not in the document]
paulbalaji: keeping the type simple as we've got regex in the constructor

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
paulbalaji: to reference it when creating the default HyperlaneIgpDeployer (line 49)

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: can't find a way to inevm/manta as they don't seem to have accounts functionalitiy, but others we don't use like optimism do

(note: eth blockscout + gnosis testnet blockscout support let you create accounts/api keys)

there's also some further work we need on the keys side to organise by explorer type, as currently we can only store one key per network and we don't know the explorer family this key belongs to

typescript/sdk/src/deploy/verify/ContractVerifier.ts
paulbalaji: divergence in blockscout compared to etherscan

all requests, POST or GET, must have the module/action set in the url

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -75,6 +76,24 @@ export abstract class HyperlaneDeployer<
...
+    if (this.options?.ismFactory) {
+      this.options.ismFactory.deployContractFromFactory =
+        this.deployContractFromFactory.bind(this);
+    }
+
+    // if none provided, instantiate a default verifier with SDK's included build artifact
+    if (!this.options?.contractVerifier) {
+      this.options = {
+        ...this.options,
+        contractVerifier: new ContractVerifier(
```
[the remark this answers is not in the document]
paulbalaji: unfortunately not in this file, you'd have to manually comb through license type comments in the contract code

info like this is in the output json, not the input json. even then, would have some edge cases when it could be X or Y license
image

solidity/exportBuildArtifact.sh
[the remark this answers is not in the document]
paulbalaji: maybe this is a very specific example I'm thinking of - i'll update to find the most recently modified file

typescript/infra/src/deployment/verify.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  // return as BuildArtifact
+  return readJSONAtPath(buildArtifactPath) as BuildArtifact;
+}
+
+// fetch explorer API keys from GCP
+export async function fetchExplorerApiKeys(): Promise<ChainMap<string>> {
+  return (await fetchGCPSecret('explorer-api-keys', true)) as any;
+}
```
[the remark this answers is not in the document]
yorhodes: ah I see, all good

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -70,11 +71,24 @@ export abstract class HyperlaneDeployer<
   constructor(
     protected readonly multiProvider: MultiProvider,
     protected readonly factories: Factories,
-    protected readonly options?: DeployerOptions,
+    protected readonly options: DeployerOptions = {},
```
jmrossy, line 74: Nit, if you're going to use a default value then you can remove the optional chaining operators below `?.`

typescript/sdk/src/deploy/verify/ContractVerifier.ts
yorhodes: I find it easier to read visually to see there isnt fall through here but fair

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
yorhodes: oh duh

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -75,6 +76,24 @@ export abstract class HyperlaneDeployer<
   ) {
     this.logger = options?.logger ?? debug('hyperlane:deployer');
     this.chainTimeoutMs = options?.chainTimeoutMs ?? 5 * 60 * 1000; // 5 minute timeout per chain
+
+    if (this.options?.ismFactory) {
+      this.options.ismFactory.deployContractFromFactory =
+        this.deployContractFromFactory.bind(this);
+    }
```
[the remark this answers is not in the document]
jmrossy: This is the first time I've seen a class method being passed around this way. My concern is that the ismFactory method, which is a separate class from this one, can modify this class' data and have unexpected side-effects. 

I don't understand well enough what problem we're solving here but I guess my first thought would be to use composition instead of this approach. I.e. pass the ism factory a reference to this deployer and have it call it as needed

typescript/sdk/src/deploy/verify/PostDeploymentContractVerifier.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { ExplorerFamily } from '../../metadata/chainMetadataTypes';
+import { MultiProvider } from '../../providers/MultiProvider';
+import { ChainMap } from '../../types';
+import { MultiGeneric } from '../../utils/MultiGeneric';
+
+import { ContractVerifier } from './ContractVerifier';
+import { BuildArtifact, CompilerOptions, VerificationInput } from './types';
+
+export class PostDeploymentContractVerifier extends MultiGeneric<VerificationInput> {
```
jmrossy, line 11: All verifiers are post deployment right? 😉

The name is probably fine, just pointing that out.

yorhodes (approved):

typescript/sdk/src/deploy/verify/PostDeploymentContractVerifier.ts
```diff
@@ -0,0 +1,50 @@
...
+  protected logger = debug('hyperlane:PostDeploymentContractVerifier');
+  private contractVerifier: ContractVerifier;
+
+  constructor(
+    verificationInputs: ChainMap<VerificationInput>,
+    private multiProvider: MultiProvider,
```
jmrossy, line 17: `protected readonly`?

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
jmrossy: We should be using objects for SDK classes to allow us to add params without worrying about order

jmrossy (commented):
Mostly looks good, just a few comments

typescript/sdk/src/deploy/HyperlaneDeployer.ts
paulbalaji: good spot thanks, forgot to remove the chaining

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
paulbalaji: completely agree, I've noticed it used in some places but not others

will try to improve this going forward

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -76,11 +76,7 @@ export abstract class HyperlaneDeployer<
   ) {
     this.logger = options?.logger ?? debug('hyperlane:deployer');
     this.chainTimeoutMs = options?.chainTimeoutMs ?? 5 * 60 * 1000; // 5 minute timeout per chain
-
-    if (this.options?.ismFactory) {
-      this.options.ismFactory.deployContractFromFactory =
-        this.deployContractFromFactory.bind(this);
-    }
+    this.options.ismFactory?.setDeployer(this);
```
paulbalaji, line 79: @jmrossy @yorhodes I originally had the binding logic to avoid exposing `deployContractFromFactory`, however the logic seems to be a lot easier to understand this way

typescript/sdk/src/deploy/HyperlaneDeployer.ts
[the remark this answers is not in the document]
paulbalaji: composing means exposing the deploy function, continued discussion here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3255/commits/75781f5966ce6f94274e9225f2503d24ad699684#r1497906716

Issue #3209, opened 2024-02-01
Individual chain failures shouldn't crash the agents


daniel-savu:
## Problem

**Is your feature request related to a problem? Please describe.**
The current way errors are propagated makes it such that a failing fallback provider (or any other propagated errors) will crash the agent. This affects liveness for other chains when it shouldn't, and agents should be "hardened" against this - i.e. have a guarantee that this can't happen.

## Solution

**Describe the solution you'd like**
We should remove all `Result` return types from the functions spawned as tokio tasks. One example is in the message processor, where the [main_loop](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3ea5dd7b6623ec4f7ca90f90145949fe37b8c09a/rust/agents/relayer/src/processor.rs#L31-L32) could log errors and keep retrying - maybe with a sleep in between.

PR #3257, opened 2024-02-13
Self-healing relayer tasks

daniel-savu:
### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3209


rust/agents/validator/src/validator.rs
```diff
@@ -173,12 +178,15 @@ impl BaseAgent for Validator {
...
                     // Future that immediately resolves
-                    return tokio::spawn(ready(Ok(()))).instrument(info_span!("Validator"));
+                    return;
                 }
             }
         }
 
         run_all(tasks)
+            .await
+            .expect("Failed to run validator tasks")
```
daniel-savu, line 188: this is the bit I'm most unhappy about

rust/agents/relayer/src/relayer.rs
```diff
@@ -380,42 +373,54 @@ impl Relayer {
...
+                    db.clone(),
+                    whitelist.clone(),
+                    blacklist.clone(),
+                    metrics,
+                    send_channels.clone(),
+                    destination_ctxs.clone(),
+                );
+                let processor = Processor::new(Box::new(message_processor));
+                // the submitter runs forever, so we only need to handle error case
+                if let Err(err) = tokio::try_join!(processor.spawn()) {
```
tkporter, line 398: can you expand on what the goal is here? My understanding is that the changes made to processor.rs were to make it so that `spawn()` doesn't return a Result, so we'd only hit this condition here if the `spawn` task panics for some reason.

Do you have an opinion whether it's better to crash if one of these tasks panics or if we should try restarting `spawn`? We chatted a bit about this, I thought we settled on panicking. I'm more on the panicking side but I'm open to the respawning. But maybe I'm misunderstanding the logic here

The tradeoffs I see are:

Panicking pros:
- panics should happen extremely rarely and we probably want to learn about these, this would be more obvious to us if it does happen
- probably simpler to reason about task interactions

Panicking cons:
- liveness impact on other tasks

Re-spawning pros:
- no liveness impact on other tasks

Re-spawning cons:
- starting new tasks imo can lead to hard to reason about / buggy behavior due to the relationship that various tasks have with one another



To expand on the hard to reason about / buggy behavior:

- For the message processor tasks, if we restart it then the MessageProcessor starts from scratch and will send all the unprocessed messages from the db via the `send_channels` to the destination submitters again. So iiuc our destination submitter queues will have dupes of all the messages in them, which feels like it could have weird implications
- For the destination submitter tasks, if we restart one of those because of a panic, then we've lost the state of the queues that are populated by receiving unprocessed messages via the message processor's `send_channels`. So we'd only be able to process new messages that come in from the message processor and wouldn't process any older messages that the message processor has already told the previous destination submitter task about
- For the merkle tree processor tasks, iiuc the merkle tree that it builds is an `Arc<RwLock<MerkleTreeBuilder>>` from `self.prover_syncs[origin]` - if we respawn, then it'll have already built out a merkle tree in the first task, and then when it restarts I think it'll keep building upon that existing merkle tree, but it'll do its loop of inserting each message into the merkle tree starting from index 0 again. So I think it may break the merkle tree that's built by double inserting messages into it

daniel-savu: > I thought we settled on panicking

fully in agreement, I had misinterpreted `try_join`'s signature.

Really appreciate the insights into the subtleties of how tasks tie together, definitely don't want to deal with any of that behavior for now

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -179,7 +180,9 @@ async fn prepare_task(
                 debug!(?op, "Operation prepared");
                 metrics.ops_prepared.inc();
                 // this send will pause this task if the submitter is not ready to accept yet
-                tx_submit.send(op).await?;
+                if let Err(err) = tx_submit.send(op).await {
```
daniel-savu, line 183: this error can only occur if the receiving end has been closed. I'm only logging the error rather than propagating / panicking to avoid taking down the tasks for other chains

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -75,18 +74,18 @@ pub struct SerialSubmitter {
     /// Domain this submitter delivers to.
     domain: HyperlaneDomain,
     /// Receiver for new messages to submit.
-    rx: mpsc::UnboundedReceiver<Box<DynPendingOperation>>,
+    rx: Arc<Mutex<mpsc::UnboundedReceiver<Box<DynPendingOperation>>>>,
```
tkporter, line 77: should be good to remove the Arc & Mutex now

rust/agents/relayer/src/relayer.rs
```diff
@@ -304,15 +308,12 @@ impl BaseAgent for Relayer {
             tasks.push(self.run_merkle_tree_processor(origin));
         }
 
-        run_all(tasks)
+        join_all(tasks).await;
```
tkporter, line 311: what's the difference with this change?

rust/agents/relayer/src/relayer.rs
```diff
@@ -391,18 +393,21 @@ impl Relayer {
...
         let span = info_span!("MessageProcessor", origin=%message_processor.domain());
         let processor = Processor::new(Box::new(message_processor));
+
+        let origin = origin.clone();
         tokio::spawn(async move {
-            let res = tokio::try_join!(processor.spawn())?;
-            info!(?res, "try_join finished for message processor");
-            Ok(())
+            // Propagate task panics
+            processor.spawn().await.unwrap_or_else(|err| {
```
tkporter, line 400: the pattern of spawning a task that spawns a task but awaits on it feels a little awkward, would we get the same behavior if we just returned `processor.spawn()` and instrumented it so we know which task it is that had issues?

don't feel strongly

tkporter (approved):
lgtm!

rust/agents/relayer/src/relayer.rs
daniel-savu: `processor .spawn()` returns a Result though, so we need to do this to restrict other tasks from propagating errors

rust/agents/relayer/src/relayer.rs
daniel-savu: `run_all` is a utility from our codebase to abort other tasks when one fails. It does the same thing as `try_join_all` afaict so I'll actually just remove it from the codebase.

`join_all` let's tasks run even if one panics, which I realised we shouldn't do, so I replaced most call sites of `join_all` with `try_join_all` and a panic.

rust/agents/relayer/src/relayer.rs
tkporter: Hmm I don't get it, I think after your changes spawn returns a JoinHandle and not a Result, then awaiting it converts that JoinHandle to a Result that Errs only if there's a panic in the spawned task. And then `run_message_processor` then returns a JoinHandle that will also Err if there's a panic on line 401

So I think the logic rn is similar to the following in Typescript:

try {
  await processor.spawn()
} catch (err) {
  console.log('there was an error')
  throw err
}

which is kinda superfluous except for the console.log adding more info?

daniel-savu: ah finally got what you meant. Yep so we lose the origin log but it's much cleaner now 👍

PR #3260, opened 2024-02-14
feat:deploy on Plume testnet

aroralanuk:
### Description

Finished core deployment on Plume testnet following the run book

NB: 
- The fork tests doesn't use SmartProvider which gives slight discrepancies between the actual deployment and forked deployment like different revert reasons
- postDeploy doesn't filter out chain if given as a param to the script

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1011

### Backward compatibility

Yes

### Testing

None


typescript/infra/config/environments/testnet4/gas-oracle.ts
```diff
@@ -29,6 +29,7 @@ const gasPrices: ChainMap<BigNumber> = {
   chiado: ethers.utils.parseUnits('2', 'gwei'),
   solanatestnet: ethers.BigNumber.from('28'),
   eclipsetestnet: ethers.BigNumber.from('28'),
+  plumetestnet: ethers.utils.parseUnits('1.5', 'gwei'),
```
yorhodes, line 32: this is quite different from the override you are using?

yorhodes (approved):

jmrossy (commented):
LGTM just some minor metadata comments

typescript/infra/config/environments/testnet4/gas-oracle.ts
aroralanuk: Yaaa, this was a 150000% premium, I've dropped it to 0.01 gwei which is 1000% premium

PR #3262, opened 2024-02-14
Sequence-aware indexing refactor

tkporter:
### Description

Part 1 of an indexing refactor. This mostly focuses on the sequence-aware cursors.

Changes:
- Some renames, e.g. `SequenceIndexer` -> `SequenceAwareIndexer`, `sequence_and_tip` -> `latest_sequence_count_and_tip`
- Moved cursors out of `contract_sync/cursor.rs` into `contract_sync/cursors/*`
- The goal was to make the the Forward sequence-aware cursor, the Backward sequence-aware, and the ForwardBackward sequence aware cursors more resilient, easier to understand, and better tested.
- General strategy for doing this was to:
  - Only move onto a new range to index if the logs passed into the cursor's `update` function justify this
  - If updated with any weird data, like gaps in logs, rewind the cursor and retry previously queried ranges

I'd recommend starting the review in the following order:

1. `rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs` - where `ForwardSequenceAwareSyncCursor` & tests live
2. `rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs` - where `BackwardSequenceAwareSyncCursor` & tests live
3. `rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs` - where `ForwardBackwardSequenceAwareSyncCursor` lives


rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            backward: backward_cursor,
+            last_direction: SyncDirection::Forward,
+        })
+    }
+}
+
+#[async_trait]
+impl<T: Sequenced + Debug> ContractSyncCursor<T> for ForwardBackwardSequenceAwareSyncCursor<T> {
+    async fn next_action(&mut self) -> Result<(CursorAction, Duration)> {
+        // TODO: Proper ETA for backwards sync
```
tkporter, line 105: These TODOs (& the ones in the Forward & Backward cursors) were present before & I'll leave that out of scope for now

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+        let all_log_sequences = logs
+            .iter()
+            .map(|(log, _)| log.sequence())
+            .collect::<HashSet<_>>();
+
+        match &self.index_mode {
+            IndexMode::Sequence => {
+                // We require that we've gotten all sequences in the range.
+                let expected_sequences = range.clone().collect::<HashSet<_>>();
+                if all_log_sequences != expected_sequences {
```
tkporter, line 271: I'm realizing the range checking could be more forgiving - e.g. if the expected range is 100..=104 and we get logs with sequences [100, 101, 103, 104] (missing 102), as it's written now it'll throw out the range and rewind to try looking for 100 again. Instead it could accept 100 and 101 and then start looking for 102.

I'm open to making a change to have this be considered - however because we will have still written all the logs to the log store, and because we have logic to fast-forward, we end up having pretty much that same behavior. On the next call to `next_action`, it'll fast forward past 100 and 101 because it'll see that it's already in the db, and then start looking for 102

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    "Current sequence is greater than the onchain sequence count"
+                );
+                None
+            }
+        };
+
+        Ok(range)
+    }
+
+    async fn fast_forward(&mut self) -> Result<()> {
```
daniel-savu, line 154: I'd add a doc comment to this function to explain that fast-forwarding is possible due to the local db, and how that's done

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+        while self
+            .db
+            .retrieve_by_sequence(self.current_indexing_snapshot.sequence)
+            .await?
+            .is_some()
+        {
+            // Require the block number as well.
+            if let Some(block_number) = self
+                .db
+                .retrieve_log_block_number(self.current_indexing_snapshot.sequence)
```
daniel-savu, line 166: Should we rename this to `retrieve_block_number_by_sequence`?

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+                // We don't necessarily expect to hit this target in the next query (because we
+                // have limits to the range size based off the chunk size), but we will use it
+                // as an eventual target.
+                self.target_snapshot = Some(TargetSnapshot {
+                    sequence: target_sequence,
+                    at_block: tip,
+                });
+
+                match &self.index_mode {
+                    IndexMode::Block => {
```
daniel-savu, line 119: At this point the indentation level is a bit high for a single function, making it hard to reason about. We should move this into another function. The u32::min range logic can be shared between `Block` and `Sequence`, which would make it even easier to understand

daniel-savu (commented):
Took lots of context loading, only a few comments for now but more coming soon

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+            // - A log at sequence 4, which was already indexed and should be ignored
+            // - Two logs of sequence 5, i.e. duplicated
+            // - A log at sequence 6, which is unexpected, but tolerated nonetheless
+            cursor
+                .update(
+                    vec![
+                        (MockSequencedData::new(4), log_meta_with_block(90)),
+                        (MockSequencedData::new(5), log_meta_with_block(95)),
+                        (MockSequencedData::new(5), log_meta_with_block(95)),
+                        (MockSequencedData::new(6), log_meta_with_block(100)),
```
daniel-savu, line 836: If you pass the sequences out of order, e.g. `4`, `5`, `6`, `5`, the test will fail bc of the dedupe behavior I mentioned above

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+            );
+
+            // Getting the range again without updating the cursor should yield the same range.
+            let range = cursor.get_next_range().await.unwrap().unwrap();
+            assert_eq!(range, expected_range);
+
+            // Update the cursor with the found log.
+            cursor
+                .update(
+                    vec![(MockSequencedData::new(5), log_meta_with_block(115))],
```
daniel-savu, line 931: If I set the return value of `latest_sequence_querier` above to `sequence: 5, at_block: 88` (updating the target snapshot assertion as well), this call to `update` still works, which feels odd. Should we also filter out entries with `block > target_snapshot.block` in `update`?

daniel-savu (commented):
Mostly done reviewing the forward cursor so far

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// If the sequence is None, this can be thought of as the starting block
+    /// number to index from.
+    pub at_block: u32,
+}
+
+impl LastIndexedSnapshot {
+    fn next_target(&self) -> TargetSnapshot {
+        TargetSnapshot {
+            // If we haven't indexed anything yet, we start at 0, otherwise we increment.
+            sequence: self.sequence.map(|s| s + 1).unwrap_or(0),
```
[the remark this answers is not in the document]
tkporter: this has different behavior when self.sequence is None, it'll set to 1 instead of 0

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+            target_snapshot: None,
+            index_mode,
+        }
+    }
+
+    /// Gets the next range of logs to index.
+    /// If there are no logs to index, returns `None`.
+    /// If there are logs to index, returns the range of logs, either by sequence or block number
+    /// depending on the mode.
+    pub async fn get_next_range(&mut self) -> Result<Option<RangeInclusive<u32>>> {
```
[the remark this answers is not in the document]
tkporter: Idk if it's worth spending much time on this given plans around https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3266 and how this'll boil down to a runtime check anyways

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -59,9 +61,14 @@ where
 
         loop {
             indexed_height.set(cursor.latest_block() as i64);
-            let Ok((action, eta)) = cursor.next_action().await else {
-                sleep(SLEEP_DURATION).await;
-                continue;
+
+            let (action, eta) = match cursor.next_action().await {
```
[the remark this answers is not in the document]
tkporter: Open to this but I think there's something nice about having cursors being a self contained way to figure out what action should be done and having the ContractSync::sync loop take the action

Can we leave this to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3264 to figure out whether to do? As the watermark cursor more heavily involves the sleep action - I think it'll be more clear when doing that one whether it's more natural to change to next_range

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
tkporter: will change the indentation but the values in the u32::min range are different, maybe i'm misinterpreting what you mean though

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
tkporter: doesn't setting the latest sequence count to 5 result in a None action? or did you mean sequence: 6

imo `update` can assume that any logs passed into the cursor are legitimate, and it should succeed as long as it builds upon the last indexed snapshot and there are no gaps in the logs

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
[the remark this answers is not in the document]
daniel-savu: right, I keep getting confused by the difference between sequence and count

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
daniel-savu: Hmm it's not just about the runtime check, but also making the code more readable. It's pretty confusing IMO to have the returned `RangeInclusive` mean different things when the function is called with a different `self` state. Let's do #3266 first and see afterwards

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+    ///
+    /// If Sequence mode:
+    /// - The sequences of the logs must exactly match the range.
+    /// - If there are any gaps, the cursor rewinds and the range will be retried.
+    ///
+    /// If Block mode:
+    /// - Empty logs are allowed, but no gaps are allowed. The logs must build upon the last indexed snapshot.
+    /// - If there are any gaps, the cursor rewinds to the last indexed snapshot, and ranges will be retried.
+    /// - If the target block is reached and the target sequence hasn't been reached, the cursor rewinds to the last indexed snapshot.
+    async fn update(&mut self, logs: Vec<(T, LogMeta)>, range: RangeInclusive<u32>) -> Result<()> {
```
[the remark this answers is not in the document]
tkporter: mm this is a good point

I think my initial goal was to have internal state only relate to things in service of making sure that we're building upon the last indexed sequence, and to have `update` avoid making assumptions about how the cursor is used and to just ensure that the logs passed in build upon the last indexed sequence, but I think the delineation isn't super clear. I'll come back to this

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
daniel-savu: > values in the u32::min range are different

Fair, I didn't look closely enough

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
daniel-savu: Indeed I meant sequence 6. What I mean is - what if there's a reorg? 

 You'd end up with a `TargetSnapshot { seq: 6, block: 88 }`, and `LastIndexedSnapshot { seq: 5, block: 115 }`, but also with a `get_next_range` = `110..=88`

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
daniel-savu: Ideally the cursor implemented the `Iterator` trait but that's not really possible since it has internal state to mutate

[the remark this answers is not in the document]
tkporter: After mulling it over I think having the range explicitly passed in is probably helpful. Curious what you think

The cursor update is really dependent on the logs + range, not just the logs. I think it also could reduce misuse of cursors in the future - by putting a range in the function signature it makes it clear that the cursor updates rely heavily on the range. I think part of the messiness of the prior implementations was that it wasn't super clear what was happening inside a cursor or how a cursor should be used, I think if we put the range as internal state we now implicitly closely tie the cursor's `update` fn to the `get_action` fn

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+        // Remove any duplicates, filter out any logs preceding our current snapshot, and sort in ascending order.
+        let logs = logs
+            .into_iter()
+            .dedup_by(|(log_a, _), (log_b, _)| log_a.sequence() == log_b.sequence())
```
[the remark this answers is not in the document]
tkporter: moved to unique_by just to avoid the dependence on it being sorted

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
daniel-savu: Hmm I think there's potential for misuse if the range is required too, since it's not entirely clear that the range to pass in should be the one from `get_action`. I'm ok either way, just wanted to point this out

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
tkporter: mm good question

I think it's probably okay to not worry about logic to accommodate reorgs. The `latest_sequence_count_and_tip` is meant to return only the finalized sequence count & the finalized tip. So as long as we've configured the node with a correct reorg period it should never encounter a reorg, so imo we can assume that all logs that make it to the `update` function are finalized. If the agent is misconfigured then I think it's ok to have undefined behavior - but certainly adding logs in such a case to complain loudly makes sense to me, I'll add some

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
tkporter: sounds good, I agree it's confusing, happy to consider completion criteria for 3266 to generally make sequence ranges vs block ranges less confusing

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
tkporter: tracking this here https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3265

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -0,0 +1,1089 @@
...
+                LastIndexedSnapshot {
+                    sequence: Some(4),
+                    at_block: 90,
+                }
+            );
+
+            // Expect the range to be:
+            // (start, tip)
+            let range = cursor.get_next_range().await.unwrap().unwrap();
+            let expected_range = 190..=200;
```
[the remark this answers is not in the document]
tkporter: Good call, I'll make a pass through it

daniel-savu (commented):
Done with the backward cursor as well. The unit tests tend to only cover the happy path, would be a good idea to include more duplicates and out-of-order logs in there (I already did this on my local and all works fine). We can actually check if we missed some obvious spots in the logic by having a look at the test coverage report (following [this guide](https://blog.rng0.io/how-to-do-code-coverage-in-rust/) for instance)

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        if let Some(backward_range) = self.backward.get_next_range().await? {
+            self.last_direction = SyncDirection::Backward;
+            return Ok((CursorAction::Query(backward_range), eta));
+        }
+        // TODO: Define the sleep time from interval flag
+        return Ok((CursorAction::Sleep(Duration::from_secs(5)), eta));
+    }
+
+    fn latest_block(&self) -> u32 {
```
daniel-savu, line 122: the doc comment of this function is `The latest block that has been queried`, rather than the latest block of the chain. I feel like always returning the forward cursor's latest block violates that, and we should either change the doc comment or the name of the method to `latest_queried_block`

rust/hyperlane-base/Cargo.toml
```diff
@@ -39,6 +39,7 @@ tracing-error.workspace = true
 tracing-futures.workspace = true
 tracing-subscriber = { workspace = true, features = ["json", "ansi"] }
 tracing.workspace = true
+tracing-test.workspace = true
```
daniel-savu, line 43: Looks like it's only used in tests, let's move under `dev-dependencies`?

rust/hyperlane-core/src/traits/cursor.rs
```diff
@@ -2,22 +2,29 @@ use std::{fmt, ops::RangeInclusive, time::Duration};
...
     /// The next block range that should be queried.
-    async fn next_action(&mut self) -> ChainResult<(CursorAction, Duration)>;
+    /// This method should be tolerant to being called multiple times in a row
+    /// without any updates in between.
+    async fn next_action(&mut self) -> Result<(CursorAction, Duration)>;
 
-    /// The latest block that has been queried
+    /// The latest block that has been queried, used as a proxy for health.
+    /// TODO: consider a better way to assess health
     fn latest_block(&self) -> u32;
```
daniel-savu, line 20: Should it also include a timestamp along with the block number? =

rust/hyperlane-core/src/traits/db.rs
```diff
@@ -34,7 +34,7 @@ where
     async fn retrieve_by_sequence(&self, sequence: u32) -> Result<Option<T>>;
```
daniel-savu, line 34: Nit: unrelated to this PR, but I think this `HyperlaneSequenceIndexerStore` trait should be renamed to `SequenceIndexerStoreReader`. Initially I thought the cursors had write access to the db due to this trait name, but turns out it's just reading for fast-forwarding purposes

daniel-savu (approved):
LFG

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -0,0 +1,874 @@
...
+                // Require the block number as well.
+                if let Some(block_number) = self
+                    .db
+                    .retrieve_log_block_number_by_sequence(current_indexing_snapshot.sequence)
+                    .await?
+                {
+                    self.last_indexed_snapshot = LastIndexedSnapshot {
+                        sequence: Some(current_indexing_snapshot.sequence),
+                        at_block: block_number.try_into()?,
+                    };
```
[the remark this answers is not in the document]
tkporter: agree there's room for deduping stuff between the two. I actually intentionally avoided trying to share logic because this felt like a source of confusion & bugs in the past implementation

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -0,0 +1,874 @@
...
+
+    /// Reads the DB to check if the current indexing sequence has already been indexed,
+    /// iterating until we find a sequence that hasn't been indexed.
+    async fn fast_forward(&mut self) -> Result<()> {
+        if let Some(current_indexing_snapshot) = self.current_indexing_snapshot.clone() {
+            // Check if any new logs have been inserted into the DB,
+            // and update the cursor accordingly.
+            while self
+                .db
+                .retrieve_by_sequence(current_indexing_snapshot.sequence)
```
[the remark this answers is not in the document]
tkporter: great catch, this was a bug that is now covered by unit tests

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
tkporter: agree this is confusing, changed to `latest_queried_block`. We can also spend more time on this in https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3265 - it's not clear to me that the latest queried block is the right proxy for healthy either

rust/hyperlane-core/src/traits/cursor.rs
tkporter: added to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3265

rust/hyperlane-core/src/traits/db.rs
tkporter: +1, I'll just rename, this is confusing

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
tkporter: I spent a bit of time on this and felt it made the tests harder to understand, happy to change if you feel strongly tho

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -0,0 +1,874 @@
...
+        } else {
+            Some(TargetSnapshot {
+                sequence: current_indexing_snapshot.sequence.saturating_sub(logs_len),
+                at_block: *range.start(),
+            })
+        };
+
+        // This means we indexed at least one log that builds on the last snapshot.
+        // Recall logs is sorted in ascending order, so the last log is the "oldest" / "earliest"
+        // log in the range.
```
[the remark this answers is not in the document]
tkporter: The annoying part is disambiguating between a zero sequence when we do
```
TargetSnapshot {
                sequence: current_indexing_snapshot.sequence.saturating_sub(logs_len),
```
to mean that we've indexed sequence zero already (which would happen if logs_len > current_indexing_snapshot.sequence), or to mean that we now want to look for index 0

But yeah this is a bit confusing, I'll change to something hopefully more idiomatic

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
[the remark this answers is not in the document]
daniel-savu: my ideal unit test looks almost like a parameterized function, so I can play around with its input without changing many lines of code, to see if the tested condition still holds. Ok to leave as-is but that's my preference

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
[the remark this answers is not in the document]
daniel-savu: fair

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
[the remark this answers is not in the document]
daniel-savu: mm right, so we can have `current_indexing_snapshot.sequence` = 0 and still not be done with indexing. Thanks for clarifying with that comment

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -180,17 +181,16 @@ impl<T: Sequenced + Debug> BackwardSequenceAwareSyncCursor<T> {
...
-            None
-        } else {
-            Some(TargetSnapshot {
-                sequence: current_indexing_snapshot.sequence.saturating_sub(logs_len),
+        // If the number of logs, which start at the current sequence and go backwards,
+        // exceeds the current indexing snapshot sequence, we've synced everything including
+        // sequence 0. Otherwise, we're not fully synced yet.
+        self.current_indexing_snapshot = current_indexing_snapshot
+            .sequence
+            .checked_sub(logs_len)
```
daniel-savu, line 189: slick

daniel-savu (approved):

PR #3271, opened 2024-02-15
Add injective to mainnet config

yorhodes:
### Description

Add injective to mainnet3 infra config

### Testing

Env tests

### DIFF
--- a/typescript/infra/config/environments/mainnet3/chains.ts
+++ b/typescript/infra/config/environments/mainnet3/chains.ts
@@ -50,9 +50,9 @@ export const ethereumMainnetConfigs: ChainMap<ChainMetadata> = {
 
 // Blessed non-Ethereum chains.
 export const nonEthereumMainnetConfigs: ChainMap<ChainMetadata> = {
-  // solana: chainMetadata.solana,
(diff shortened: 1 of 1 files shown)


typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -50,9 +50,9 @@ export const ethereumMainnetConfigs: ChainMap<ChainMetadata> = {
 
 // Blessed non-Ethereum chains.
 export const nonEthereumMainnetConfigs: ChainMap<ChainMetadata> = {
-  // solana: chainMetadata.solana,
-  // neutron: chainMetadata.neutron,
-  // injective: chainMetadata.injective,
+  solana: chainMetadata.solana,
+  neutron: chainMetadata.neutron,
```
tkporter, line 54: can you just uncomment injective and leave the others commented? solana we're not on mainnet yet and neutron we only have on the neutron context - the configs here are only used to decide what to put in the hyperlane context

(as i write this, more points toward getting rid of the neutron context, lol)

tkporter (approved):

PR #3275, opened 2024-02-15
Fix no domain id for non-evm in IGP deploy+check

yorhodes:
### Description

MultiProvider does not have domain IDs for non-evm domains. This uses chainMetadata and falls back to multiprovider to allow for configuration of remote chains which are not present on the multiprovider.

### Testing

- IGP fork tests


jmrossy (commented):
The 'right' fix here is to continue the march towards multi-protocol utilities which use MultiProtocolProvider as you mentioned. I think the HyperlaneIgpDeployer change may break the CLI PI deployments. AFAI only we use the checker so that's maybe not as important

yorhodes:
> I think the HyperlaneIgpDeployer change may break the CLI PI deployments

looks like the CLI test caught this! :)

typescript/sdk/src/gas/HyperlaneIgpChecker.ts
```diff
@@ -89,7 +90,9 @@ export class HyperlaneIgpChecker extends HyperlaneAppChecker<
         expectedOverhead = 0;
       }
 
-      const remoteId = this.multiProvider.getDomainId(remote);
+      const remoteId =
```
aroralanuk, line 93: can you leave a comment "in case we don't have the chain in core chain metadata (like for CLI), we need the optional ? part here"

aroralanuk (approved):

PR #3278, opened 2024-02-16
CLI option for adding block and gas-config

armanthepythonguy:
### Description

Added option to configure block and gas settings while creating new chain config using CLI

### Related issues

NA

### Backward compatibility

Yes

### Testing

Manual


typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,52 @@ export async function createChainConfig({
     protocol: ProtocolType.Ethereum,
     rpcUrls: [{ http: rpcUrl }],
   };
+  const wantBlockConfig = await confirm({
```
nambrot, line 93: I would hide these behind an advanced question or something. Let's keep it simple in the default case

typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,52 @@ export async function createChainConfig({
...
+    message: 'Do you want to add block config for this chain?',
+  });
+  if (wantBlockConfig) {
+    const blockConfirmation = await input({
+      message:
+        'Enter no. of blocks to wait before considering a transaction confirmed',
```
nambrot, line 99: Probably useful to have some validation/reasonsable bounds here

typescript/cli/src/config/chain.ts
armanthepythonguy: What I understand is instead of directly asking for these configs first the user will be prompted if they want to add advanced configs for this chain, if yes then all these prompts will show up. Please validate my understanding @nambrot

typescript/cli/src/config/chain.ts
armanthepythonguy: Can we give these ranges :- 
confirmations: 0-10
estimateBlockTime: 0-20
reorgPeriod: 0-50

typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,62 @@ export async function createChainConfig({
...
+  });
+  if (wantAdvancedConfig) {
+    const wantBlockConfig = await confirm({
+      message: 'Do you want to add block config for this chain?',
+    });
+    if (wantBlockConfig) {
+      const blockConfirmation = await input({
+        message:
+          'Enter no. of blocks to wait before considering a transaction confirmed(0-10)',
+        validate: (value) => parseInt(value) >= 0 && parseInt(value) <= 10,
```
jmrossy, line 104: I suggest a higher max here. We use 200 for polygon, so how about 500

typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,62 @@ export async function createChainConfig({
...
+      });
+      const blockReorgPeriod = await input({
+        message:
+          'Enter no. of blocks before a transaction has a near-zero chance of reverting(0-50)',
+        validate: (value) => parseInt(value) >= 0 && parseInt(value) <= 50,
```
jmrossy, line 109: Ditto, polygon uses 256 so how about 500 again

typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,62 @@ export async function createChainConfig({
     protocol: ProtocolType.Ethereum,
     rpcUrls: [{ http: rpcUrl }],
   };
+  const wantAdvancedConfig = await confirm({
+    message: 'Do you want to add advanced config for this chain?',
```
jmrossy, line 94: Let's be more specific about what's being offered here. I suggest:
`Do you want to set block or gas properties for this chain config (optional)`

typescript/cli/src/config/chain.ts
```diff
@@ -90,6 +90,62 @@ export async function createChainConfig({
...
+      if (isEIP1559) {
+        const maxFeePerGas = await input({
+          message: 'Enter the max fee per gas in gwei',
+        });
+        const maxPriorityFeePerGas = await input({
+          message: 'Enter the max priority fee per gas in gwei',
+        });
+        metadata.transactionOverrides = {
+          maxFeePerGas: parseInt(maxFeePerGas, 10) * 10 ** 9,
+          maxPriorityFeePerGas: parseInt(maxPriorityFeePerGas, 10) * 10 ** 9,
```
jmrossy, line 137: Why `** 9` here? Did you mean to use 18 for Ether's 18 decimals?
Either way, there's a possibility of Number overrun so you should use `bigint` instead.

typescript/cli/src/config/chain.ts
armanthepythonguy: Sure i will update it !!!

typescript/cli/src/config/chain.ts
armanthepythonguy: Sure i will update it !!!

typescript/cli/src/config/chain.ts
armanthepythonguy: Sure i will update it !!!

typescript/cli/src/config/chain.ts
armanthepythonguy: That **9 is to refactor it in terms of gwei. I will update it with bigint

jmrossy (approved):

Issue #3274, opened 2024-02-15
Add testrecipient back to infra and CoreDeployer


yorhodes:
## Problem

We removed the [`testrecipient` module](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/66d9b2907dd9308e630292876c1c6ad985cf692c/typescript/infra/scripts/deploy.ts#L125) from infra deploy recently. The TestRecipient deployment is also an unnecessary extra step in the runbook.

## Solution

We should add it back and also have the `CoreDeployer` use the `TestRecipientDeployer` internally for one fewer default entrypoints. 

It is still useful to have separate entrypoints for the `testrecipient` deploy to enable contexts with different ISM configurations.

PR #3279, opened 2024-02-16
feat:add `TestRecipient` deployment to core deployer

aroralanuk:
### Description

- Added `TestRecipientDeployer` to the core deployer because most of the times, the TestRecipient is required right after core deployment and it's very lightweight. The TestRecipient deployment is also an unnecessary extra step in the runbook.
- Put it back in infra/deploy which gives the deployer the option to deploy it separately.

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3274

### Backward compatibility

Yes

### Testing

Fork tests


typescript/infra/scripts/deploy.ts
```diff
@@ -122,7 +124,16 @@ async function main() {
...
     deployer = new LiquidityLayerDeployer(multiProvider);
   } else if (module === Modules.TEST_RECIPIENT) {
-    throw new Error('Test recipient is not supported. Use CLI instead.');
+    const addresses = getAddresses(environment, Modules.CORE);
+
+    for (const chain of Object.keys(addresses)) {
+      config[chain] = {
+        interchainSecurityModule:
+          addresses[chain].interchainSecurityModule ??
+          ethers.constants.AddressZero,
```
jmrossy, line 133: It's not very clear why this is required, can you leave a comment?

jmrossy (approved):
LGTM

typescript/infra/scripts/deploy.ts
aroralanuk: done

PR #3280, opened 2024-02-16
Add env spec to chains list CLI command

nbayindirli:
### Description

* Added environment specification to the `hyperlane chains list` CLI command
* Provides CLI-users with the option to specify `--testnet` | `-t` or `--mainnet` | `-m` to only display chains for that environment

### Drive-by changes

* None

### Related issues

* None

### Backward compatibility

* Yes

### Testing

* Manual
  * `yarn hyperlane chains list --help`
```
Options:
      --help     Show help                                             [boolean]
  -m, --mainnet  Only list mainnet chains
  -t, --testnet  Only list testnet chains
```
  * `yarn hyperlane chains list -m`
```
Hyperlane core mainnet chains:
------------------------------
...
```
  * `yarn hyperlane chains list -t`
```
Hyperlane core testnet chains:
------------------------------
...
```
  * `yarn hyperlane chains list -m -t`
```
Error: Arguments mainnet and testnet are mutually exclusive

hyperlane chains list

List all core chains included in the Hyperlane SDK
```
  * `yarn hyperlane chains list`

```
Hyperlane core testnet chains:
------------------------------
...
Hyperlane core mainnet chains:
------------------------------
...
```

### DIFF
--- a/typescript/cli/src/commands/chains.ts
+++ b/typescript/cli/src/commands/chains.ts
@@ -32,7 +32,21 @@ export const chainsCommand: CommandModule = {
 const listCommand: CommandModule = {
   command: 'list',
   describe: 'List all core chains included in the Hyperlane SDK',
-  handler: () => {
+  builder: (yargs) =>
+    yargs
+      .option('mainnet', {
+        alias: 'm',
+        describe: 'Only list mainnet chains',
+      })
+      .option('testnet', {
+        alias: 't',
+        describe: 'Only list testnet chains',
+      })
+      .conflicts('mainnet', 'testnet'),
+  handler: (args) => {
+    const mainnet = args.mainnet as string | undefined;
+    const testnet = args.testnet as string | undefined;
+
     const serializer = (chains: string[]) =>
       chains.reduce<any>((result, chain) => {
         result[chain] = {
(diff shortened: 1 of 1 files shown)


typescript/cli/src/commands/chains.ts
```diff
@@ -41,13 +55,22 @@ const listCommand: CommandModule = {
...
+      logTable(serializer(Mainnets));
+    };
+    const logTestnet = () => {
+      logBlue('\nHyperlane core testnet chains:');
+      logGray('------------------------------');
+      logTable(serializer(Testnets));
+    };
+
+    if (mainnet) return logMainnet();
+    else if (testnet) return logTestnet();
```
jmrossy, line 70: This is fine, but it would be more direct to simply run the code under the if clause instead of defining an arrow function above and then using it immediately.

jmrossy (approved):
Thanks for the improvement! One optional comment

typescript/cli/src/commands/chains.ts
nbayindirli: Since the flags are mutually exclusive and optional, I have the same functions executed below (see line 72 & 73) in order to print out both envs in the absence of both flags. This code reuse was my reasoning for the lambdas

jmrossy: Ah, I didn't notice that. Okay agreed

PR #3283, opened 2024-02-19
fix: lower rusoto timeout to 15s

daniel-savu:
### Description

Applies the fix in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2384 everywhere an `HttpClient` is constructed via rusoto.

It lowers the S3 timeout to 15s based on tips in [this thread](https://github.com/hyperium/hyper/issues/2136#issuecomment-589488526), to avoid `Error during dispatch: connection closed before message completed` errors. Note that we'll probably still run into these issues, but less frequently ([source](https://github.com/rusoto/rusoto/issues/1766#issuecomment-639299961)).


rust/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -91,9 +94,12 @@ impl S3Storage {
 
     /// Gets an authenticated S3Client, creating it if it doesn't already exist.
     fn authenticated_client(&self) -> &S3Client {
+        let mut config = HttpConfig::new();
+        // see https://github.com/hyperium/hyper/issues/2136#issuecomment-589345238
+        config.pool_idle_timeout(HYPER_POOL_IDLE_TIMEOUT);
         self.authenticated_client.get_or_init(|| {
             S3Client::new_with(
-                HttpClient::new().unwrap(),
+                HttpClient::new_with_config(config).unwrap(),
```
tkporter, line 102: could put this in a shared fn but either way works

tkporter (approved):

PR #3284, opened 2024-02-19
feat: move desired balances into KeyFunderConfig + bump thresholds

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2385

- moves desired balances into KeyFunderConfig
- update balances across the board


typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -24,4 +24,35 @@ export const keyFunderConfig: KeyFunderConfig = {
...
+    bsc: '5',
+    celo: '3',
+    ethereum: '5',
+    gnosis: '5',
+    moonbeam: '5',
+    optimism: '3',
+    polygon: '20',
+    polygonzkevm: '3',
+    scroll: '3',
+  },
```
tkporter, line 41: can you add inevm and viction here? and, if it's possible, use types to make sure that all chains in the environment have an entry here?

tkporter (approved):

typescript/infra/config/environments/mainnet3/funding.ts
paulbalaji: had a look through some of the existing types and we're casting stuff as ChainMaps all the way down, probs needs more TLC down the line to tighten up the typings

PR #3286, opened 2024-02-19
feat: parallelise e2e test workflow

paulbalaji:
- parallelise cosmwasm and non-cosmwasm e2e tests
- no need to have an explicit agent build step
	- cc @daniel-savu 
- `e2e` status should still pass as it collates result from the matrix 

before: 26-30m
after: 15-20m

there's likely further optimisation that can be done by caching the agent build artifacts? but I think this should be good enough for now

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,6 +164,9 @@ jobs:
   e2e:
     runs-on: larger-runner
     needs: [yarn-build]
+    strategy:
+      matrix:
+        e2e-type: [cosmwasm, non-cosmwasm]
     steps:
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -220,22 +223,32 @@ jobs:
...
-        run: RUST_BACKTRACE=1 cargo test --release --package run-locally --bin run-locally --features cosmos -- cosmos::test --nocapture
+        run: cargo test --release --package run-locally --bin run-locally --features cosmos -- cosmos::test --nocapture
+        if: matrix.e2e-type == 'cosmwasm'
         working-directory: ./rust
+        env:
+          RUST_BACKTRACE: 1
 
       - name: agent tests excluding CosmWasm
-        run: ./target/release/run-locally
+        run: cargo run --release --package run-locally
```
daniel-savu, line 234: I think this should be `cargo run --release --bin run-locally`

daniel-savu (approved):
🔥

.github/workflows/test.yml
paulbalaji, line 234:
```suggestion
        run: cargo run --release --bin run-locally
```

PR #3294, opened 2024-02-20
Add compile-time types & names on global env vars

nbayindirli:
### Description

* Uses zod to add global access to named & typed env vars via `ENV*`
* As a result, this change enables intellisense and type-safety on all environment variables

### Drive-by changes

* No

### Related issues

* None

### Backward compatibility

* Yes

### Testing

* Manual– verified intellisense is working locally
* Ran the following script which uses process.env: `./typescript/cli/ci-test.sh 'false'`


yorhodes (commented):
interesting
I dont think I've ever seen this but directionally seems useful
we already use zod in the CLI so I wonder if we can enforce a zod schema against a `process.env`
as it stands this adds compile time schema but doesnt actually enforce it at runtime

yorhodes:
> interesting I dont think I've ever seen this but directionally seems useful we already use zod in the CLI so I wonder if we can enforce a zod schema against a `process.env` as it stands this adds compile time schema but doesnt actually enforce it at runtime

quick google surfaced this https://github.com/trpc/examples-next-prisma-starter/blob/5c8373d773cf1e2bb04d6a7aab3fa2e59185e341/src/server/env.js#L15

yorhodes (approved):
love it

typescript/cli/src/utils/env.ts
```diff
@@ -0,0 +1,7 @@
+import z from 'zod';
+
+const envScheme = z.object({
+  HYP_KEY: z.string(),
```
jmrossy, line 4: string().optional()

typescript/cli/src/utils/env.ts
```diff
@@ -0,0 +1,7 @@
...
+});
+
+export const ENV = envScheme.parse(process.env);
```
jmrossy, line 7: I wonder if we should use `safeParse` here instead with a fallback to `{}`. Just wondering if random crap in the env object could cause an error. Maybe not though, just a passing thought

nbayindirli: Agreed. Added along with optional

jmrossy (approved):

typescript/cli/src/utils/env.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const parsedEnv = envScheme.safeParse(process.env);
+
+if (!parsedEnv.success)
+  throw new Error(
+    `Failed to parse environment variables: ${JSON.stringify(
+      parsedEnv.error.format(),
+      null,
+    )}`,
+  );
```
jmrossy, line 15: Either way is fine but I'd be more inclined to fallback to an empty ENV object instead of throwing here. Otherwise this is equivalent to using `parse()`. The idea is to avoid a fatal error based on weird content in a user's ENV vars

nbayindirli: Ah yep that's fair. FWIW I've got plenty of non-HL env vars that didn't pose an issue when running locally. Regardless will simply have this return `{}` if !success

PR #3295, opened 2024-02-21
v2 liveness probe to cover Solana indexing flakiness

tkporter:
### Description

Will restart the pod if:
- the relayer serial submitter's latest nonce is lower than that of the validators
- the relayer serial submitter's latest nonce hasn't increased in the last 10 mins (to avoid a case where the relayer has just spun up and is trying to catch up to the validators)

Fixes https://github.com/hyperlane-xyz/issues/issues/1102

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -39,7 +39,9 @@ RUN apt-get update && \
     apt-get install -y \
         openssl \
         ca-certificates \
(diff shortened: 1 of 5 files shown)


daniel-savu (approved):
managed to make a succesful query with `wget` in the pod and used that response to mock others, lgtm

PR #3301, opened 2024-02-22
feat: tweak keyfunder params

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/1103

- tune down eth L1 required balance
- tune down L2s required balance
- improve comment r.e. 5x L1->L2 multiplier
- update + tidy up L2Chains types


typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -26,20 +26,21 @@ export const keyFunderConfig: KeyFunderConfig = {
   connectionType: RpcConsensusType.Single,
   // desired balance config
   desiredBalancePerChain: {
-    arbitrum: '3',
     avalanche: '5',
-    base: '3',
     bsc: '5',
     celo: '3',
-    ethereum: '5',
+    ethereum: '0.5',
```
aroralanuk, line 32: there's no boost for l1 iirc?

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -539,19 +542,19 @@ class ContextFunder {
   }
 
   private async bridgeIfL2(chain: ChainName) {
-    if (L2Chains.includes(chain)) {
+    if (L2Chains.includes(chain as L2Chain)) {
```
aroralanuk, line 545: would prefer not using the "as" cast here

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -65,29 +65,29 @@ const nativeBridges = {
...
-  | Chains.optimism
-  | Chains.optimismgoerli
-  | Chains.arbitrum
-  | Chains.arbitrumgoerli
-  | Chains.base;
-
-const L2Chains: ChainName[] = [
-  Chains.optimism,
-  Chains.optimismgoerli,
+type L2Chain = (typeof L2Chains)[number];
```
aroralanuk, line 68: it's better to divide the l2 as OpStack, ArbNitro, etc which all have the same funding mechanism (contract interaction) and then union them to get L2Chains

typescript/infra/config/environments/mainnet3/funding.ts
paulbalaji: yup, but lowering this back down due to levels of current use + cost of eth

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: good idea

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: alternate would be to do `(L2Chains as string[])` instead

paulbalaji: also why not cast here, as we're casting a few lines later

paulbalaji: rejigged stuff but it means we're doing a nasty `as readonly string[]` when switching depending on the L2 chain bridge we need

aroralanuk: ok, this works then

PR #3309, opened 2024-02-23
Add scripts for fetching prices and populating config

### LINKED ISSUE #3251: Generate IGP exchange rate and gas price config from network
## Problem

There is too much friction in updating the IGP rates. 

**Additional context**
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3244 hooks up this config to the core/IGP deployers which are run in fork tests in CI

## Solution

We should have a generated JSON that replaces this [manually populated config](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/simplify-igp-oracles/typescript/infra/config/environments/mainnet3/gas-oracle.ts). 

This should use coingecko for fetching exchange rates, which the [CoinGeckoTokenPriceGetter](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6bb5be16a99a679e5bfb3338daaa95369e42e901/typescript/sdk/src/gas/token-prices.ts#L70) already does.

This can use simple `eth_gasPrice` RPCs but should probably use [etherscan gastracker API](https://docs.etherscan.io/api-endpoints/gas-tracker#get-gas-oracle) for fetching historically reasonable gas prices.

All of this work should be surfaced in the CLI somehow rather than going to die in the infra package.

Previous internal ticket: https://github.com/hyperlane-xyz/issues/issues/439

yorhodes:
### Related issues

- Fixes #3251


aroralanuk (approved):

aroralanuk (approved):
Slay

PR #3312, opened 2024-02-26
Add technical stack config for Arbitrum Nitro chains

tkporter:
### Drive-by changes

- consistent stringifying of keys
- change plumetestnet reorg period to 0

### DIFF
--- a/typescript/sdk/src/metadata/chainMetadataTypes.ts
+++ b/typescript/sdk/src/metadata/chainMetadataTypes.ts
@@ -15,6 +15,11 @@ export enum ExplorerFamily {
   Other = 'other',
 }
 
+export enum ChainTechnicalStack {
+  ArbitrumNitro = 'arbitrumnitro',
+  Other = 'other',
+}
+
 // A type that also allows for literal values of the enum
 export type ExplorerFamilyValue = `${ExplorerFamily}`;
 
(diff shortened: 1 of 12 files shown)


typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -15,6 +15,11 @@ export enum ExplorerFamily {
   Other = 'other',
 }
 
+export enum ChainTechnicalStack {
```
jmrossy, line 18: You should probably export this from the SDK's index.ts

typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -15,6 +15,11 @@ export enum ExplorerFamily {
...
+  ArbitrumNitro = 'arbitrumnitro',
+  Other = 'other',
```
jmrossy, line 20: Will we ever use Other or would we just leave it undefined?

tkporter: Mm good question.
The main reason I added `Other` was in Rust I feel a bit more strongly that the enum type should cover all cases instead of having it be optional. At the moment the agents will default to Other if the chain metadata passed into the agents doesn't specify a technical stack. What would you prefer?

1. require techincalStack to be specified in the chain metadata
2. remove the Other variant in Typescript but keep it in Rust
3. remove the Other variant in Typescript & Rust and rely on it being undefined or not (this is kinda more awkward in the agent code)
4. keep as is

jmrossy: I'm fine with 2 or 4, I defer to you :)

jmrossy (approved):

Issue #3197, opened 2024-01-31
Update ICA to v3


avious00:
- allow the ICARouter to handle msg.value in callRemote
- add tests for the above
- add tests for sending and handle multidomain messages

aroralanuk:
Testing cases/contentious scenarios
- msg.value dispatch and handle (pass in msgvalue in msg?)
- address aliasing?

PR #3314, opened 2024-02-26
feat: update ICA to v3 

aroralanuk:
### Description

- Update ICA to accept _hookMetadata per call and per batch of calls.
- Allow ICA to accept msg.value

### Drive-by changes

- Moved IGP from MockMailbox to the InterchainAccountRouterTest contract as it's not needed in the former and doesn't break any tests

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3197

### Backward compatibility

Yes with V3

### Testing

Unit tests with focus on igp as the hook as this will be the one used in production


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -1,12 +1,25 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.13;
 
+/*@@@@@@@       @@@@@@@@@
```
yorhodes, line 4: boosting those contribution numbers lol

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -150,14 +163,16 @@ contract InterchainAccountRouter is Router {
...
      */
     function callRemote(
         uint32 _destination,
         address _to,
         uint256 _value,
-        bytes memory _data
-    ) external returns (bytes32) {
+        bytes memory _data,
+        bytes memory _hookMetadata
+    ) external payable returns (bytes32) {
```
yorhodes, line 175: can we keep these functions for backwards compatibility and just add new ones with hook metadata?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -388,20 +444,29 @@ contract InterchainAccountRouter is Router {
...
      */
     function callRemoteWithOverrides(
         uint32 _destination,
         bytes32 _router,
         bytes32 _ism,
-        CallLib.Call[] calldata _calls
-    ) public returns (bytes32) {
+        CallLib.Call[] calldata _calls,
+        bytes memory _hookMetadata
+    ) public payable returns (bytes32) {
```
yorhodes, line 456: ditto here?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -474,7 +539,33 @@ contract InterchainAccountRouter is Router {
...
+     */
+    function _dispatchMessageWithMetadata(
+        uint32 _destination,
+        bytes32 _router,
+        bytes32 _ism,
+        bytes memory _body,
+        bytes memory _hookMetadata
+    ) private returns (bytes32) {
+        require(_router != bytes32(0), "no router specified for destination");
+        emit RemoteCallDispatched(_destination, msg.sender, _router, _ism);
```
yorhodes, line 561: I would prefer to have only one `mailbox.dispatch` and `emit RemoteCallDispatched` in the contract
can we refactor slightly?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -92,6 +92,9 @@ contract InterchainAccountRouterTest is Test {
     function setUp() public {
         environment = new MockHyperlaneEnvironment(origin, destination);
 
+        igp = new TestInterchainGasPaymaster();
+        gasPaymentQuote = igp.getDefaultGasUsage() * igp.gasPrice();
```
yorhodes, line 96: why not just use the quote fn?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -124,80 +129,102 @@ contract InterchainAccountRouterTest is Test {
...
-        assertEq(ica.owner(), address(destinationRouter));
+    function testFuzz_constructor(address _localOwner) public {
+        OwnableMulticall _account = destinationRouter
+            .getDeployedInterchainAccount(
+                origin,
+                _localOwner,
+                address(originRouter),
+                address(environment.isms(destination))
+            );
+        assertEq(_account.owner(), address(destinationRouter));
```
yorhodes, line 140: im surprised this works?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -124,80 +129,102 @@ contract InterchainAccountRouterTest is Test {
...
+        address _account = originRouter.getRemoteInterchainAccount(
+            address(_localOwner),
+            address(destinationRouter),
+            _ism
         );
         originRouter.enrollRemoteRouterAndIsm(
             destination,
             routerOverride,
-            ismOverride
+            TypeCasts.addressToBytes32(_ism)
```
yorhodes, line 155: why is the ISM bytes32?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -268,89 +307,195 @@ contract InterchainAccountRouterTest is Test {
...
+            ismOverride
+        );
+        uint256 balanceBefore = address(this).balance;
+
+        // act
+        originRouter.callRemote{value: gasLimit * igp.gasPrice() + overpayment}(
+            destination,
+            getCalls(data),
+            metadata
+        );
```
yorhodes, line 420: should we test underpayment with hook metadata too? as it stands these arent really unit tests

yorhodes (approved):
thanks for adding fuzzing
some small asks around preserving the simplest interface

solidity/contracts/middleware/InterchainAccountRouter.sol
aroralanuk: I cannot put empty metadata so we need an if else regardless of having just one function call mailbox.dispatch

solidity/test/InterchainAccountRouter.t.sol
aroralanuk: no pointed out yourself, this is remote ism

solidity/test/InterchainAccountRouter.t.sol
aroralanuk: why?

solidity/contracts/middleware/InterchainAccountRouter.sol
aroralanuk: done

solidity/contracts/middleware/InterchainAccountRouter.sol
aroralanuk: done

solidity/test/InterchainAccountRouter.t.sol
aroralanuk: done

Issue #2572, opened 2023-07-26
Investigate better infrastructure management approaches


mattiekat:
We ran into an RPC issue so I went to re-deploy one of the agents but in doing so the infra config in the repo's main branch had drifted from what the correct deployment configs were and caused the agent to crash loop until I could diagnose and resolve what changes it did not like.

In my opinion it is a huge issue if our infrastructure configs do not match what is in production because it means that when issues happen, we now have to play a game of figure out what commit to deploy from or what needs to be changed.

The specific issue I encountered was a chain configuration mismatch with the scraper, the `main` branch included two new chains for _all_ agents `solanadevnet` and `zbctestnet` but the scraper was not capable of handling this.

We already have the notion of updating the deployed image tags in the repo to match what is in prod so I don't think this is a huge stretch. Running point on production issues is turned into a minefield when you can't even be sure what branch, tag, or hash you can safely deploy a given object from.

PR #3327, opened 2024-02-28
DRAFT: Separate agent chain configurations, re-add some chains to infra, fix eclipsetestnet & solanatestnet configs

tkporter:

### DIFF
--- a/typescript/infra/config/environments/mainnet3/agent.ts
+++ b/typescript/infra/config/environments/mainnet3/agent.ts
@@ -32,7 +32,7 @@ import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
 const repo = 'gcr.io/abacus-labs-dev/hyperlane-agent';
 
 // The chains here must be consistent with the environment's supportedChainNames, which is
-// checked & enforced at runtime.
+// checked / enforced at runtime & in the CI pipeline.
 //
 // This is intentionally separate and not derived from the environment's supportedChainNames
 // to allow for more fine-grained control over which chains are enabled for each agent role.
--- a/typescript/infra/config/environments/mainnet3/agent.ts
+++ b/typescript/infra/config/environments/mainnet3/agent.ts
@@ -31,7 +31,12 @@ import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
 
 const repo = 'gcr.io/abacus-labs-dev/hyperlane-agent';
 
-const hyperlaneContextAgentChainConfig: AgentChainConfig = {
+// The chains here must be consistent with the environment's supportedChainNames, which is
+// checked & enforced at runtime.
+//
+// This is intentionally separate and not derived from the environment's supportedChainNames
+// to allow for more fine-grained control over which chains are enabled for each agent role.
+export const hyperlaneContextAgentChainConfig: AgentChainConfig = {
   [Role.Validator]: {
     [Chains.arbitrum]: true,
     [Chains.avalanche]: true,
@@ -159,10 +164,7 @@ const hyperlane: RootAgentConfig = {
 const releaseCandidate: RootAgentConfig = {
   ...contextBase,
   context: Contexts.ReleaseCandidate,
-  contextChainNames: {
-    ...hyperlaneContextAgentChainNames,
-    [Role.Validator]: ethereumChainNames,
-  },
+  contextChainNames: hyperlaneContextAgentChainNames,
   rolesWithKeys: [Role.Relayer, Role.Kathy, Role.Validator],
   relayer: {
     rpcConsensusType: RpcConsensusType.Fallback,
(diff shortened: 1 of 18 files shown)


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -26,11 +31,84 @@ import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
...
 const repo = 'gcr.io/abacus-labs-dev/hyperlane-agent';
 
+// The chains here must be consistent with the environment's supportedChainNames, which is
+// checked / enforced at runtime & in the CI pipeline.
+//
+// This is intentionally separate and not derived from the environment's supportedChainNames
+// to allow for more fine-grained control over which chains are enabled for each agent role.
+export const hyperlaneContextAgentChainConfig: AgentChainConfig = {
+  [Role.Validator]: {
+    [Chains.arbitrum]: true,
```
tkporter, line 41: I expect this is maybe controversial - it's verbose but I think forcing us to be explicit, including when we *don't* want to support a chain, will help

nambrot:
@tkporter this basically fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2572 ?

typescript/infra/src/agents/key-utils.ts
```diff
@@ -399,8 +407,10 @@ async function persistAddressesLocally(
     multisigValidatorKeys[key.chainName] ||= {
       validators: [],
     };
-    if (key.chainName)
+
+    if (key.role === Role.Validator) {
```
daniel-savu, line 411: Can there ever be a case where the `role` is `validator` but `key.chainName` is undefined?

daniel-savu (approved):

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -26,11 +31,84 @@ import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
...
+    [Chains.avalanche]: true,
+    [Chains.bsc]: true,
+    [Chains.celo]: true,
+    [Chains.ethereum]: true,
+    [Chains.neutron]: true,
+    [Chains.mantapacific]: true,
```
daniel-savu, line 47: shouldn't these be false too, to match the relayer?

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -86,10 +165,7 @@ const hyperlane: RootAgentConfig = {
 const releaseCandidate: RootAgentConfig = {
   ...contextBase,
   context: Contexts.ReleaseCandidate,
-  contextChainNames: {
-    ...contextBase.contextChainNames,
-    [Role.Validator]: ethereumChainNames,
-  },
+  contextChainNames: hyperlaneContextAgentChainNames,
```
daniel-savu, line 168: When these need to diverge, will we just destructure `hyperlaneContextAgentChainNames` and overwrite here? So to roll out a new chain we'd add a new `false` entry to the hyperlane context and then overwrite to true in `rc`?

typescript/infra/src/agents/key-utils.ts
```diff
@@ -385,6 +385,14 @@ async function persistAddressesLocally(
   const multisigValidatorKeys: ChainMap<{ validators: Address[] }> = {};
   let relayer, kathy;
   for (const key of keys) {
+    // Some types of keys come in an AWS and a GCP variant. We prefer
```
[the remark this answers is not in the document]
tkporter: correct, also clarified in the comment

typescript/infra/src/agents/key-utils.ts
tkporter: no - clarified as well

typescript/infra/config/environments/mainnet3/agent.ts
tkporter: we generally run all production validators in the hyperlane context and leave the other contexts for just relaying, added a comment

typescript/infra/config/environments/mainnet3/agent.ts
tkporter: That works, I have less of a strong opinion about how non-hyperlane contexts specify things. My main goal was to just make sure that the hyperlane context isn't reliant on configurations done for managing contracts

But my general thoughts are:
- hyperlane & rc should always have the same chains enabled
- if RC deviated from this, either another `AgentChainConfig` written out or the destructure & overwrite may be natural, I'm unsure if I have an opinion
- any other contexts should specify chains explicitly, like how we do for the neutron context

typescript/infra/config/environments/mainnet3/agent.ts
tkporter: Definitely agree that moving from typescript to JSON / YAML would help in the move to make configs more explicit - this is my agenda for when we move more things out into the CLI. I've been having some conversations recently around this and think we'll start making moves toward this soon

Until we build out a better setup for ensuring "correctness" of parsed JSON/YAML configs, I think there are some benefits to keeping it in TS, like being able to use enums & keeping all agent deployment configuration within a single file

PR #3348, opened 2024-03-01
DO NOT MERGE YET: Set the podManagementPolicy for all StatefulSets to Parallel

### LINKED ISSUE #3347: Upgrading a K8s statefulset when pods are in CrashLoopBackoff should upgrade the pods
## Problem

We've been burned a few times where we make a new deploy because a pod in a statefulset is in CrashLoopBackoff. The default rules for statefulsets seem to be to only upgrade a pod if the pod is in a healthy state. This isn't obvious, and it's not clear that we need to manually delete the crashing pod to have the upgrade take hold

## Solution

https://stackoverflow.com/a/66636003 seems promising

tkporter:
### Description

See #3347 for motivation. https://stackoverflow.com/a/66636003 suggests this change

Docs for more info: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies

Imo the docs don't make it super clear what the implications are here, so I played around with it by creating a dummy helm chart with a statefulset that I played with.

From playing around with it, it seems the only actual difference is that now if statefulset pods are in a CrashLoopBackoff, if you perform a new helm upgrade that changes the statefulset, it'll update that pod even if it's in CrashLoopBackoff.

It still will perform rolling upgrades, as the `updateStrategy` is still set to the default of `RollingUpdate`. Meaning if there are 3 replicas in the statefulset, it'll first update pod `xxx-2`, wait for it to be healthy, then `xxx-1`, etc. This is the desired behavior imo!

**Note**

The `podManagementPolicy` is not a property that can be upgraded on an existing statefulset, which is super annoying. This means that in order to roll this out, we'll need to delete each statefulset and re-create the statefulset, instead of just updating the existing statefulset 😬. We'll need to be extra careful that this doesn't delete PVCs as well.

My plan here is to:
- try this on testnet4 RC first to see if I need to do anything special to preserve the PVCs upon deleting a statefulset
- manually roll this out to every statefulset we run

I'll probably get around to being able to deploy on Monday. My ideal timeline would be:
1. Get an approval on this
2. On Monday, I'll take over on call and send a message to have no one else perform any deploys while I'm rolling this out
3. Go through the painstaking process of rolling this out 😢 

### Related issues

Fixes  #3347 

### Backward compatibility

n/a

### Testing

Manually played with a statefulset that I deployed that would crash

### DIFF
--- a/rust/helm/hyperlane-agent/templates/relayer-statefulset.yaml
+++ b/rust/helm/hyperlane-agent/templates/relayer-statefulset.yaml
@@ -37,6 +37,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.podManagementPolicy }}
+      podManagementPolicy: {{ . }}
+      {{- end }}
       terminationGracePeriodSeconds: 10
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
--- a/rust/helm/hyperlane-agent/templates/scraper-statefulset.yaml
+++ b/rust/helm/hyperlane-agent/templates/scraper-statefulset.yaml
@@ -37,6 +37,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.podManagementPolicy }}
+      podManagementPolicy: {{ . }}
+      {{- end }}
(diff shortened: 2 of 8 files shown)


daniel-savu (approved):
looks good, afaict

tkporter:
Finally capitulating and admitting the effort to roll this out exceeds the benefits from it; im gonna close

Issue #3349, opened 2024-03-01
Validator tasks should not "silently" end when an error occurs


tkporter:
## Problem

In the validator, we spawn 2 checkpoint submitter tasks here:

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e737d998ff23b5aa6cf2be9634789ca5dc5a4275/rust/agents/validator/src/validator.rs#L250-L262

Both of these functions return Results and they don't have appropriate error handling. It's possible for us to hit transient errors in here -- here are examples of us hitting errors in the `checkpoint_submitter` function, causing us to return an Err:
- Gnosis validator, it seems it hit the infamous S3 `connection closed before message completed` error:
- Polygonzkevm validator, intermittent RPC issues: https://cloudlogging.app.goo.gl/uRjJACfHfSAtf1HX8

## Solution

We've chatted in the past about us being more willing to crash in the validator if errors occur so we can loudly learn that there are issues occurring. I think in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3257 we must've made a change that doesn't result in crashing anymore, and instead tasks just silently die with an error.

I don't feel super strongly about what we do, but we need to either:
- crash the validator if one of these tasks exists
- similar to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3257, make sure all tasks in the validator are infallible -- i.e. if an error occurs in a checkpoint submitter loop, we should just continue looping instead of returning a Result::Err and ending the task

We should make sure this covers all tasks in the validators, not just the checkpoint submitter ones -- e.g. indexing, or any others that may exist, like a metric task etc

Issue #2878, opened 2023-10-31
Use overarching Error enum in agents


daniel-savu:
I'd like to have one error enum that includes `ChainCommunicationError`, `HyperlaneCustomErrorWrapper`, eyre errors, and so on. Wondering what the reasoning behind not doing this before was? cc @tkporter

nambrot:
@daniel-savu still relevant I assume?

daniel-savu:
It is yes. We've already started consolidating environment-specific errors into `ChainCommunicationError` using the `From` trait impl approach ([example here for cosmos](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/chains/hyperlane-cosmos/src/error.rs)), but it's a work in progress and many errors are still standalone.

PR #3361, opened 2024-03-05
Validator task retries

daniel-savu:
### Description

- Makes validator tasks infallible by adding retries 
- This fixes a bug where certain tasks would return an error an shut down, affecting liveness. This was desired in the relayer since we don't want individual chain failures to affect the liveness of other chains. Now validator tasks either terminate or panic, and panicking will be propagated by `try_join_all`, causing the agent to shut down.
    - A thing to keep in mind in general is that agents will only terminate if a task panics. If it returns an `Err` but doesn't panic, the task won't be respawned. We should consider the implications of this in the scraper too. If this isn't desired, should consider using `select_all!`

### Drive-by changes

- retry logic is moved from `rust/chains/hyperlane-cosmos/src/providers/rpc.rs` into `rust/hyperlane-core/src/rpc_clients/retry.rs`, so we're even closer to turning it into a retrying provider
- changes several fn signatures within the validator to now return `ChainCommunicationError`s instead of `eyre::Report`s, for compatibility with the retry logic. Also makes `DbError` convertible to `ChainCommunicationError`, for the same reason. This achieves some progress on https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2878
- allows the backfill checkpoint submitter task to terminate, since `try_join_all` is tolerant of this (as described above)

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3349


rust/agents/validator/src/submit.rs
```diff
@@ -133,8 +142,20 @@ impl ValidatorSubmitter {
                 continue;
             }
 
-            self.submit_checkpoints_until_correctness_checkpoint(&mut tree, &latest_checkpoint)
-                .await?;
+            tree = call_and_retry_indefinitely(|| {
```
daniel-savu, line 142: Need to explicitly return `tree` and reassign to keep `tree` updated, because the async closure needs to take ownership instead of mutably borrowing

rust/hyperlane-core/src/error.rs
```diff
@@ -138,6 +138,15 @@ pub enum ChainCommunicationError {
     #[cfg(feature = "async")]
     #[error(transparent)]
     TokioJoinError(#[from] tokio::task::JoinError),
+    /// Custom error
+    #[error("{0}")]
+    CustomError(String),
+    /// Custom error
+    #[error("{0}")]
+    EyreError(#[from] eyre::Report),
```
daniel-savu, line 146: Unhappy about the naming of this, but couldn't come up with a better one

tkporter (approved):
🚀

rust/hyperlane-core/src/rpc_clients/retry.rs
```diff
@@ -0,0 +1,51 @@
...
+#[instrument(err, skip(f))]
+pub async fn call_and_retry_n_times<T>(
+    mut f: impl FnMut() -> Pin<Box<dyn Future<Output = ChainResult<T>> + Send>>,
+    n: usize,
+) -> ChainResult<T> {
+    for retry_number in 1..n {
+        match f().await {
+            Ok(res) => return Ok(res),
+            Err(err) => {
+                debug!(retries=retry_number, error=?err, "Retrying call");
```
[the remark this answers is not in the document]
daniel-savu: hmm let's try, if they get very noisy it'll be a strong indication that we need to rotate RPCs

PR #3364, opened 2024-03-05
Tx fee estimatators and WarpCore interface refinements

jmrossy:
### Description

- Add multi-protocol transaction fee estimators
- Improve WarpCore interface 
- Fix ChainMetadata `transactionOverrides` zod type

### Drive-by changes

Add `gasPrice` field to Neutron chain metadata

### Related issues

https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/123

### Backward compatibility

Small breaking change: Token Adapter `quoteGasPayment` method renamed to `quoteTransferRemoteFee` for clarity.

### Testing

Tested in Warp UI with nexus routes.


typescript/sdk/src/warp/types.ts
```diff
@@ -59,4 +38,29 @@ export const WarpCoreConfigSchema = z.object({
...
+
+// Transaction types for warp core remote transfers
+export enum WarpTxCategory {
+  Approval = 'approval',
+  Transfer = 'transfer',
+}
+
+export type WarpTypedTransaction = TypedTransaction & {
+  category: WarpTxCategory;
+};
```
jmrossy, line 58: Not new, just moved down

typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -174,7 +174,7 @@ export const ChainMetadataSchemaObject = z.object({
     .optional()
     .describe('Block settings for the chain/deployment.'),
   transactionOverrides: z
-    .object({})
+    .record(z.any())
```
jmrossy, line 177: preview type was incorrect and wouldn't allow arbitrary fields

typescript/sdk/src/warp/types.ts
```diff
@@ -59,4 +38,29 @@ export const WarpCoreConfigSchema = z.object({
     .optional(),
 });
 
+// List of constant values for local or interchain fees
+export type FeeConstantConfig = z.infer<typeof FeeConstantConfigSchema>;
+
+// List of chain pairs to blacklist for warp routes
+export type RouteBlacklist = Array<{
```
jmrossy, line 45: Not new, just moved down

typescript/sdk/src/token/Token.ts
```diff
@@ -377,18 +378,39 @@ export class Token implements IToken {
...
    */
-  collateralizes(token: Token): boolean {
-    if (token.chainName !== this.chainName) return false;
-    if (!TOKEN_COLLATERALIZED_STANDARDS.includes(this.standard)) return false;
-    const isCollateralWrapper =
-      this.collateralAddressOrDenom &&
-      eqAddress(this.collateralAddressOrDenom, token.addressOrDenom);
-    const isNativeWrapper = !this.collateralAddressOrDenom && token.isNative();
-    return isCollateralWrapper || isNativeWrapper;
+  equalsAsset(token?: IToken): boolean {
```
jmrossy, line 388: Replaced `collateralizes` with `equalsAsset` which (I think) is more clear and also covers the cosmos native vs ibc token case

typescript/sdk/src/token/IToken.ts
```diff
@@ -82,7 +82,6 @@ export interface IToken extends TokenArgs {
   addConnection(connection: TokenConnection): IToken;
   removeConnection(token: IToken): IToken;
 
-  equals(token: IToken): boolean;
-
-  collateralizes(token: IToken): boolean;
+  equals(token?: IToken): boolean;
+  equalsAsset(token?: IToken): boolean;
```
yorhodes, line 86: I personally dont think this makes things more clear
hard to tell what `equals` vs `equalsAsset` does
also why optional now?

yorhodes: `equalsAsset` sounds like it should be `token.asset.equals(other.asset)`

typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
```diff
@@ -296,7 +298,8 @@ export class CwHypSyntheticAdapter
     weiAmountOrId,
     interchainGas,
   }: TransferRemoteParams): Promise<ExecuteInstruction> {
-    if (!interchainGas) interchainGas = await this.quoteGasPayment(destination);
+    if (!interchainGas)
+      interchainGas = await this.quoteTransferRemoteGas(destination);
```
yorhodes, line 302: nice

typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
```diff
@@ -31,6 +31,10 @@ import {
   TransferRemoteParams,
 } from './ITokenAdapter';
 
+// An estimate of the gas amount for a typical EVM token router transferRemote transaction
+// Computed by estimating on a few different chains, taking the max, and then adding ~50% padding
+export const EVM_TRANSFER_REMOTE_GAS_ESTIMATE = 450_000n;
```
yorhodes, line 36: where is this used?

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -132,20 +150,107 @@ export class WarpCore {
...
+        transaction: txs[0],
+        sender,
+        senderPubKey,
+      });
+    } else if (
+      txs.length === 2 &&
+      originToken.protocol === ProtocolType.Ethereum
+    ) {
+      // For ethereum txs that require >1 tx, we assume the first is an approval
+      // We use a hard-coded const as an estimate for the transferRemote gas
```
yorhodes, line 221: why hardcode this? seems fragile
imo the warp route interface should be independent from the implementation details
also different ERC20 collaterals have different gas usage

typescript/sdk/src/warp/types.ts
```diff
@@ -59,4 +38,29 @@ export const WarpCoreConfigSchema = z.object({
...
+  category: WarpTxCategory;
+};
+
 export type WarpCoreConfig = z.infer<typeof WarpCoreConfigSchema>;
+
+export interface WarpCoreFeeEstimate {
+  interchainQuote: TokenAmount;
+  localQuote: TokenAmount;
+  localDetails: TransactionFeeEstimate;
+}
```
yorhodes, line 66: I find this interface slightly confusing
is `localQuote === localDetails.fee`?

typescript/sdk/src/warp/types.ts
yorhodes, line 58: you really don't like the array approach :P

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -197,13 +304,109 @@ export class WarpCore {
...
+  }: {
+    balance: TokenAmount;
+    destination: ChainNameOrId;
+    sender: Address;
+    senderPubKey?: HexString;
+    feeEstimate?: WarpCoreFeeEstimate;
+  }): Promise<TokenAmount> {
+    const originToken = balance.token;
+
+    if (!feeEstimate) {
```
yorhodes, line 377: I would remove this optional param and always do this estimation here

typescript/sdk/src/providers/transactionFeeEstimators.ts
```diff
@@ -0,0 +1,260 @@
...
+  estimatedGasPrice: Numberish;
+  sender: Address;
+  senderPubKey: HexString;
+  memo?: string;
+}): Promise<TransactionFeeEstimate> {
+  const stargateClient = await provider.provider;
+  const message = transaction.transaction;
+  const registry = new Registry([...defaultRegistryTypes, ...wasmTypes]);
+  const encodedMsg = registry.encodeAsAny(message);
+  const encodedPubkey = encodeSecp256k1Pubkey(Buffer.from(senderPubKey, 'hex'));
```
yorhodes, line 140: can the public key be queried using the account lookup RPC? would prefer not to pollute all the interfaces with this senderPubKey

typescript/sdk/src/providers/transactionFeeEstimators.ts
```diff
@@ -0,0 +1,260 @@
...
+    ...transaction.transaction,
+    from: sender,
+  });
+  const feeData = await provider.provider.getFeeData();
+  let gasPrice: BigNumber;
+  if (feeData.maxFeePerGas && feeData.maxPriorityFeePerGas) {
+    gasPrice = feeData.maxFeePerGas.add(feeData.maxPriorityFeePerGas);
+  } else if (feeData.gasPrice) {
+    gasPrice = feeData.gasPrice;
+  } else {
```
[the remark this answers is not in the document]
jmrossy: Ethers does that already

typescript/sdk/src/providers/transactionFeeEstimators.ts
```diff
@@ -0,0 +1,260 @@
...
+  sender,
+  senderPubKey,
+}: {
+  transaction: TypedTransaction;
+  provider: TypedProvider;
+  chainMetadata: ChainMetadata;
+  sender: Address;
+  senderPubKey?: HexString;
+}): Promise<TransactionFeeEstimate> {
+  if (
```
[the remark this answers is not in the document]
jmrossy: yes because we need explicit if statements for the typescript type narrowing to work

typescript/sdk/src/providers/transactionFeeEstimators.ts
```diff
@@ -0,0 +1,260 @@
...
+  sender: Address;
+}): Promise<TransactionFeeEstimate> {
+  const gasUnits = await provider.provider.estimateGas({
+    ...transaction.transaction,
+    from: sender,
+  });
+  const feeData = await provider.provider.getFeeData();
+  let gasPrice: BigNumber;
+  if (feeData.maxFeePerGas && feeData.maxPriorityFeePerGas) {
+    gasPrice = feeData.maxFeePerGas.add(feeData.maxPriorityFeePerGas);
```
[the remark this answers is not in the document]
jmrossy: In practice I've seen this works pretty well in my testing but overestimates the cost slightly, which is preferable IMO

typescript/sdk/src/token/IToken.ts
jmrossy: It's optional because it's convenient in a few places to compare a token to a value that may be undefined (which of course returns false)

typescript/sdk/src/warp/types.ts
jmrossy: Not quite, one is a bigint and the other is a TokenAmount, meaning we've resolved the details amount to it's associated token

typescript/sdk/src/token/IToken.ts
jmrossy: > equalsAsset sounds like it should be token.asset.equals(other.asset)
Yeah it's an interesting idea. I'm open to it. Want to talk more about it

typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
jmrossy: in WarpCore

typescript/sdk/src/warp/WarpCore.ts
jmrossy: How do you suggest we do estimations for transfers that require an approval. Keep in mind we can only rely on the set of operations that free RPCs allow, meaning no fancy multi-tx simulation

typescript/sdk/src/warp/types.ts
jmrossy: Like I said, the UI _needs_ to know this context :)

typescript/sdk/src/warp/WarpCore.ts
jmrossy: I added this specifically to avoid re-querying for the estimate when we just got it. Unless we implement a query caching layer, I'd like to keep this. The Max button feels slow already

typescript/sdk/src/providers/transactionFeeEstimators.ts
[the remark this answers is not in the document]
jmrossy: Alternatively I could have `assert` calls at the top of each estimator fn which narrows there, then this can be a simple map. Lmk if you prefer that

typescript/sdk/src/token/IToken.ts
jmrossy: Renamed to `isFungibleWith` and added more details to the comment

typescript/sdk/src/providers/transactionFeeEstimators.ts
jmrossy:
```
What you can do is ask the chain if it stores the public key for a given address. If the address has signed transactions before, this gives you a public key on Cosmos.
```
https://github.com/cosmos/cosmjs/issues/889

typescript/sdk/src/warp/WarpCore.ts
yorhodes: oh... I did not understand the problem until just now

yorhodes (approved):

typescript/sdk/src/warp/WarpCore.ts
jmrossy: Improving the comments around here to make this more clear

typescript/sdk/src/warp/types.ts
jmrossy: Removing it for a cleaner interface

typescript/sdk/src/providers/transactionFeeEstimators.ts
jmrossy: Leaving a comment about this but not fixing for now

PR #3369, opened 2024-03-06
Add shell utility for adding delegates to safe

yorhodes:


yorhodes:
this was for ledger signing with cast, but prefer to use the `safe-delegate.ts` script with private keys

PR #3374, opened 2024-03-07
fix: log level order, cosmos logging nit

daniel-savu:
### Description

Our logging prio order was off, double checked with [this crate](https://docs.rs/log/latest/log/enum.Level.html). Debug logs would be visible at `Info` level according to the previous rules.

### Drive-by changes

Removed the debug formatting of a cosmsos log field, so it's no longer surrounded in quotes (bc we'd get `cursor_label: ""InterchainGasPaymentCursor""`)

### DIFF
--- a/rust/hyperlane-base/src/settings/trace/mod.rs
+++ b/rust/hyperlane-base/src/settings/trace/mod.rs
@@ -24,7 +24,7 @@ pub enum Level {
     /// Warn
     Warn = 2,
     /// Debug
-    Debug = 3,
+    Debug = 4,
     /// Trace
     Trace = 5,
     /// Trace + Additional logs from dependencies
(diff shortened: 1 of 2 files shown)


rust/hyperlane-base/src/settings/trace/mod.rs
```diff
@@ -24,15 +24,15 @@ pub enum Level {
...
+    Debug = 4,
     /// Trace
     Trace = 5,
     /// Trace + Additional logs from dependencies
     DependencyTrace = 6,
     /// Info
     #[serde(other)]
     #[default]
-    Info = 4,
+    Info = 3,
```
tkporter, line 35: do you know if this needs to be the last one in the enum? feels weird

tkporter (approved):

PR #3378, opened 2024-03-07
chore: deploy block-by-block cosmos indexing everywhere

daniel-savu:
### Description

Updates the docker images and removes docker overrides


tkporter (approved):
🚀

Issue #3296, opened 2024-02-21
Token deployer address serialization


yorhodes:
## Problem

The [Token Deployer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/8fe24830ee461c4d2fe62b13cb5ad5e9f3135138/typescript/sdk/src/token/deploy.ts#L55) inherits router behavior from the [Router Deployer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/93f864f387d2aa36bf7adeaf543a75efeaec5b15/typescript/sdk/src/router/HyperlaneRouterDeployer.ts#L20). When serializing the deployed contracts output, this behavior is not desired. 

## Solution

Instead of using the `router` key in the output, the key matching the factory used in the `deployContractFromFactory` should be used. This will fix the idempotency of the token deployer.

## Nice to Have

- [ ] Add a test to ensure the idempotency of the token deployer

PR #3379, opened 2024-03-07
fix: correctly recover the warp route router from artifacts

aroralanuk:
### Description

- Writing and read to cache as the [token.type]: address.

NOTE: this breaks if the agents index by {router: address} for warp routes but can we easily fixed by checking for any of the tokenTypes
Ideally, we want a multi-key artifact like { sepolia: [{type:synthetic, standard:erc20, name: 'USD Coin', address: 0x1233..}],} but that's more scope than intended for this issue. cc @daniel-savu @tkporter 

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3296

### Backward compatibility

Yes

### Testing

Manual


typescript/sdk/src/token/config.ts
```diff
@@ -12,6 +12,19 @@ export enum TokenType {
   native = 'native',
 }
 
+export enum TokenStandard {
+  ERC20 = 'ERC20',
+  ERC721 = 'ERC721',
+}
```
jmrossy, line 18: We have a big list of `TokenStandard`s in TokenStandard.ts already

typescript/sdk/src/token/deploy.ts
```diff
@@ -219,7 +208,7 @@ export class HypERC20Deployer extends GasRouterDeployer<
       throw new Error('Invalid ERC20 token router config');
     }
     await this.configureClient(chain, router, config);
-    return { router };
+    return { [getTokenType(config.type, TokenStandard.ERC20)]: router } as any;
```
jmrossy, line 211: Can we use a generic type in this deployContracts method to avoid the multiple new casts we have in this file?

typescript/sdk/src/token/deploy.ts
```diff
@@ -330,54 +319,31 @@ export class HypERC721Deployer extends GasRouterDeployer<
...
+      getTokenType(config.type, TokenStandard.ERC721),
+      [config.mailbox],
+    )) as HypERC721;
   }
 
   router(contracts: HyperlaneContracts<HypERC721Factories>) {
-    return contracts.router;
+    for (const key of Object.keys(hypERC721factories)) {
+      if (contracts[key as keyof typeof contracts]) {
+        return contracts[key as keyof typeof contracts];
```
jmrossy, line 343: A few things that may help here:
- Set an explicit type on `hypERC721factories` where it's defined.
- Add `as const` after `hypERC721factories` definition
- Add this method to the utils object.ts file which maintains the key type. I thought we had this already but apparently not:
```ts
// Useful for maintaining type safety when using Object.keys
export function objKeys<T extends string | number>(obj: Record<T, any>): T[] {
  return Object.keys(obj) as T[];
}
```

typescript/sdk/src/token/deploy.ts
```diff
@@ -139,16 +134,11 @@ export class HypERC20Deployer extends GasRouterDeployer<
...
-
-    return this.deployContractFromFactory(chain, factory, name, [
-      config.token,
-      config.mailbox,
-    ]);
+    return (await this.deployContract(
+      chain,
+      getTokenType(config.type, TokenStandard.ERC20),
+      [config.token, config.mailbox],
+    )) as HypERC20Collateral;
```
yorhodes, line 141: we shouldn't have to cast here
also the `await` is unnecessary

typescript/sdk/src/token/deploy.ts
```diff
@@ -157,19 +147,16 @@ export class HypERC20Deployer extends GasRouterDeployer<
...
-      router = await this.deployContractFromFactory(
-        chain,
-        new HypNativeScaled__factory(),
-        'HypNativeScaled',
-        [config.scale, config.mailbox],
-      );
+      router = (await this.deployContract(chain, 'nativescaledERC20', [
+        config.scale,
+        config.mailbox,
+      ])) as HypNative;
```
yorhodes, line 153: I think we can avoid casting here

typescript/sdk/src/token/deploy.ts
yorhodes: ++

typescript/sdk/src/token/deploy.ts
yorhodes: there is a global `PropertyKey` type or something

typescript/sdk/src/token/deploy.ts
aroralanuk: I've tried TokenRouter but it's exactly work without casting

typescript/sdk/src/token/deploy.ts
```diff
@@ -140,54 +136,56 @@ export class HypERC20Deployer extends GasRouterDeployer<
...
+        TokenType.fastCollateral,
+        [config.token, config.mailbox],
+      );
+    } else {
+      return this.deployContract<TokenType.collateral>(
+        chain,
+        TokenType.collateral,
+        [config.token, config.mailbox],
+      );
+    }
```
yorhodes, line 151:
```suggestion
      return this.deployContract(
        chain,
        isFast ? TokenType.fastCollateral : TokenType.collateral,
        [config.token, config.mailbox],
      );
```
does this not work?

typescript/sdk/src/token/deploy.ts
```diff
@@ -140,54 +136,56 @@ export class HypERC20Deployer extends GasRouterDeployer<
...
+        chain,
+        TokenType.fastSynthetic,
+        [config.decimals, config.mailbox],
+      );
+    } else {
+      router = await this.deployContract<TokenType.synthetic>(
+        chain,
+        TokenType.synthetic,
+        [config.decimals, config.mailbox],
+      );
```
yorhodes, line 187: ditto for here with just using isFast for switching on the key to `deployContract`

typescript/sdk/src/token/deploy.ts
```diff
@@ -330,54 +333,47 @@ export class HypERC721Deployer extends GasRouterDeployer<
...
       );
     } else {
-      router = await this.deployContractFromFactory(
+      return this.deployContract<TokenType.synthetic>(
         chain,
-        new HypERC721__factory(),
-        'HypERC721',
+        TokenType.synthetic,
         [config.mailbox],
       );
```
yorhodes, line 366: I think we can merge these too

typescript/sdk/src/token/deploy.ts
```diff
@@ -330,54 +333,47 @@ export class HypERC721Deployer extends GasRouterDeployer<
...
     }
-    await this.multiProvider.handleTx(
-      chain,
-      router.initialize(config.totalSupply, config.name, config.symbol),
-    );
-    return router;
```
yorhodes, line 376: why remove this?

typescript/sdk/src/token/deploy.ts
```diff
@@ -330,54 +333,47 @@ export class HypERC721Deployer extends GasRouterDeployer<
...
         [config.token, config.mailbox],
       );
     } else {
-      router = await this.deployContractFromFactory(
+      return this.deployContract<TokenType.collateral>(
         chain,
-        new HypERC721Collateral__factory(),
-        'HypERC721Collateral',
+        TokenType.collateral,
         [config.token, config.mailbox],
```
yorhodes, line 346: ditto on merging these

typescript/sdk/src/token/deploy.ts
aroralanuk: it does

typescript/sdk/src/token/deploy.ts
aroralanuk: done

typescript/sdk/src/token/deploy.ts
aroralanuk: done

typescript/sdk/src/token/deploy.ts
aroralanuk: readded back

typescript/cli/src/deploy/warp.ts
yorhodes, line 1: personally not a fan of this new format, would prefer to be consistent with `contractName: address` where `contractName` is a `keyof` the factories but dont feel super strongly

yorhodes (approved):

typescript/cli/src/deploy/warp.ts
aroralanuk: Not a fan either. Ideally, we'll have a multi-key artifact like { sepolia: [{type:synthetic, standard:erc20, name: 'USD Coin', address: 0x1233..}],} but that's a bigger scope.

Issue #3240, opened 2024-02-08
Use metricAppContexts in the relayer to give more granularity on the contents of the prepare queue


tkporter:
## Problem

We have metricAppContexts for observing validator set latest indices. We can use these matching lists to also track the prepare queue with messages relating to important deployments. E.g. if we have a warp route that we're relaying on the `hyperlane` context, we want to alert if that queue is nonzero but don't want to alert if any other user messages (that e.g. may not be deliverable) are in the prepare queue.

Atm we've been able to have nice alerting conditions if a warp route isn't working by just running a relayer in a separate context, or if we're confident that the relayer won't be delivering any other messages to a certain chain (like solana for instance), but this isn't sustainable if we wanna run more things in the hyperlane context

PR #3385, opened 2024-03-11
Use app context classifier in relayer submitter queues

daniel-savu:
### Description

Includes the `app_context` classification in `PendingMessage`, and adds trait methods on `PendingOperation` to require always having such a label on `OpQueue` operations. This is done by reusing the matching list logic from the validator checkpoint labels (https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3057).

The nice thing is that this enables later support for retrying a group of `OpQueue` operations just by specifying the `app_context` label, without adding any new logic, since these labels are essentially matching list results. One downside to using `app_context` for retries is that the endpoint caller is constrained to only the matching lists defined by the relayer operator - however imo only the relayer operator that should be able to trigger retries.

### Drive-by changes

The `OpQueue` type alias is converted to an actual struct, that stores the queue label (for metrics purposes), and also the `IntGaugeVec` metric: the generic group of metrics associated with that queue (basically only `submitter_queue_length` currently).

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3240

### Backward compatibility

Yes

### Testing

Manual, by spinning up a relayer for injective and inevm. Sample metrics, from `--metricAppContexts '[{"name": "injectivelabel", "matchingList": [{"destination_domain": 6909546 }] }, {"name": "inevmlabel", "matchingList": [{"destination_domain": 2525 }] }]'`
```
hyperlane_submitter_queue_length{agent="relayer",app_context="inevmlabel",hyperlane_baselib_version="0.1.0",queue_name="confirm_queue",remote="inevm"} 11
hyperlane_submitter_queue_length{agent="relayer",app_context="inevmlabel",hyperlane_baselib_version="0.1.0",queue_name="prepare_queue",remote="inevm"} 0
hyperlane_submitter_queue_length{agent="relayer",app_context="injectivelabel",hyperlane_baselib_version="0.1.0",queue_name="confirm_queue",remote="injective"} 63
hyperlane_submitter_queue_length{agent="relayer",app_context="injectivelabel",hyperlane_baselib_version="0.1.0",queue_name="prepare_queue",remote="injective"} 13281
```


rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -104,6 +105,16 @@ impl PendingOperation for PendingMessage {
...
     }
 
+    /// The label to use for metrics granularity.
+    fn app_context(&self) -> Option<String> {
+        self.app_context.clone()
+    }
+
+    fn destination(&self) -> String {
+        // TODO: check whether this is the correct way of converting the address to string
+        self.ctx.destination_mailbox.address().to_string()
```
daniel-savu, line 115: @tkporter wondering whether this is sure to match the matching list?

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -131,7 +131,7 @@ impl AppContextClassifier {
     pub async fn get_app_context(
         &self,
         message: &HyperlaneMessage,
-        root_ism: H256,
+        root_ism: Option<H256>,
```
daniel-savu, line 134: making this optional, because this is irrelevant to submitter queue messages. Maybe we'll switch to a fancier filter type that can take in `ism`s as well as other things if this feature gets more use

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -20,7 +20,28 @@ use hyperlane_core::HyperlaneDomain;
 
 use super::pending_operation::*;
 
-type OpQueue = Arc<Mutex<BinaryHeap<Reverse<Box<DynPendingOperation>>>>>;
+#[derive(Debug, Clone, new)]
```
daniel-savu, line 23: converted this alias to a newtype struct, since `push` occurs in several places and that would duplicate calls to metrics

rust/agents/relayer/src/msg/pending_operation.rs
```diff
@@ -32,6 +32,19 @@ pub trait PendingOperation {
     /// The domain this operation will take place on.
     fn domain(&self) -> &HyperlaneDomain;
 
+    /// Label to use for metrics granularity.
+    fn app_context(&self) -> Option<String>;
```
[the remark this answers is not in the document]
daniel-savu: It's nice that the OpQueue only lives in memory. The relayer can just be restarted with a new `app_context` matching list, and that will cause the messages that match to be immediately retried (so there'll never be stale labels in this queue)

rust/agents/relayer/src/msg/processor.rs
```diff
@@ -94,10 +97,18 @@ impl ProcessorExt for MessageProcessor {
 
             debug!(%msg, "Sending message to submitter");
 
+            let domain: HyperlaneDomain = KnownHyperlaneDomain::try_from(msg.destination)?.into();
+            let app_context_classifier = AppContextClassifier::new(
+                self.mailboxes.get(&domain).map(Clone::clone),
```
tkporter, line 102: Because the mailbox is only used by AppContextClassifier for getting the default ISM, and we don't care about ISMs, wondering if we should use something else than just use AppContextClassifier? that way we wouldn't need to add `mailboxes: HashMap<HyperlaneDomain, Arc<dyn Mailbox>>,`, as rn it's not actually used for anything

e.g. maybe rename `AppContextClassifier` to `IsmAwareAppContextClassifier` or something, and pull out the bit that we actually use for queue classification (that doesn't care about ISMs) into its own function that we use here

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -20,7 +20,41 @@ use hyperlane_core::HyperlaneDomain;
...
+        self.get_operation_metric(&op).inc();
+
+        self.queue.lock().await.push(Reverse(op));
+    }
+
+    /// Pop an element from the queue and update metrics
+    async fn pop(&self) -> Option<Reverse<Box<DynPendingOperation>>> {
+        let op = self.queue.lock().await.pop();
+        op.map(|op| {
+            self.get_operation_metric(&op.0).dec();
```
tkporter, line 46: one pretty big downside of my suggestion to update on the pop/push "entrypoints" is that this would update the metric as if the operation that's currently being processed isn't part of the queue

wondering if this can get confusing if e.g. we have one message "stuck" in the queue that we're concerned about, but each time we try to process it the metric shows the queue size as 0, so when looking at alerts / dashboards it looks like we're successfully processing the entire queue

wdyt? i guess we could keep as is, or consider just updating the metrics upon `PendingOperationResult::Success` or something

tkporter (approved):

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: right, so the name of the `domain`. Changed this and updated the PR description

rust/agents/relayer/src/msg/serial_submitter.rs
daniel-savu: there's usually several messages that fail to deliver at once, and the queue size would still increase in those cases - the queue may just be off by one sometimes. In those cases the sum of `prepare_queue` and `confirm_queue` will still be nonzero I believe. I think that's good enough, added a comment about this

tkporter: fwiw during the period that a message is being "worked on" (like building metadata, or waiting for a tx receipt, etc), the message won't be a part of the confirm queue yet, so the sum will be 0. You're right that usually there's more than one in the queue at least at our scale now though. So happy to merge and we can always keep an eye if it leads to confusion

PR #3399, opened 2024-03-11
Update kathy image

nambrot:
### Description

There are probably outdated RPC urls in an older image, updating the image seemed to move kathy out of crash loop


typescript/infra/config/environments/mainnet3/helloworld.ts
```diff
@@ -13,7 +13,7 @@ export const hyperlane: HelloWorldConfig = {
   kathy: {
     docker: {
       repo: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
-      tag: '86b7f98-20231207-153806',
+      tag: '2173b8b-20240311-144231',
```
paulbalaji, line 16: probably should pull latest `main` and update tag after a new monorepo image for this branch is pushed

PR #3406, opened 2024-03-12
fix:add a static lag of 10 to block height in cosmos::merkle_root_hook

aroralanuk:

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
+++ b/rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, num::NonZeroU64, ops::RangeInclusive, str::FromStr};
+use std::{cmp::min, fmt::Debug, num::NonZeroU64, ops::RangeInclusive, str::FromStr};
 
 use async_trait::async_trait;
 use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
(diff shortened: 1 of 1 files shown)


rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
```diff
@@ -331,10 +331,16 @@ impl Indexer<MerkleTreeInsertion> for CosmosMerkleTreeHookIndexer {
 #[async_trait]
 impl SequenceAwareIndexer<MerkleTreeInsertion> for CosmosMerkleTreeHookIndexer {
     async fn latest_sequence_count_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
+        let default_lag = NonZeroU64::new(10).unwrap();
         let tip = self.get_finalized_block_number().await?;
```
tkporter, line 335: we actually pretty much already do this - the `get_finalized_block_number` here takes into consideration the configured reorg period (which is [10](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/config/mainnet3_config.json#L539)) here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2173b8b2629755ed59d92c73df9b1ad955a8d057/rust/chains/hyperlane-cosmos/src/providers/rpc.rs#L218-L229

so we're already making a query 10 blocks back from the tip

PR #3407, opened 2024-03-12
Even higher sol fees

tkporter:
### Description

Average tips from https://solanacompass.com/statistics/fees show 0.0001 SOL per TX now. This bumps the prio fee to that


tkporter:
Gonna close - seems that our woes were from poor tx inclusion from quicknode nodes

PR #3420, opened 2024-03-14
feat: parse warp route artifacts with token-specific key names

daniel-savu:
### Description

Looks like all warp route artifacts until now used to have keys named `router`. The artifacts for USDC and USDT inevm <> EVM have their keys named according to the token type (e.g. `HypERC20`, `HypERC20Collateral` etc). This adds a script to parse such files


tkporter (approved):

Eianaftyio:
:Prep app

Issue #3421, opened 2024-03-14
WarpCore: prevent transfer to zero-ish addresses


jmrossy:
- Strengthen common sense tx data checks, including null-related addresses, before we submit the tx to the wallet to prevent unexpected tx encoding results.
- Add special logging & alerting to Sentry related to null address generation

Thread: https://discord.com/channels/935678348330434570/1217889698400440381

avious00:
@yorhodes is it worthwhile for us to do the same for contract side or no?

PR #3426, opened 2024-03-15
Prevent warp transfers to zero-ish addresses

jmrossy:
### Description

- Prevent warp transfers to zero-ish addresses
- Increase address utility test coverage

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3421

### Backward compatibility

Yes

### Testing

Tested in warp UI and added unit tests

### DIFF
--- a/typescript/utils/src/addresses.ts
+++ b/typescript/utils/src/addresses.ts
@@ -4,6 +4,7 @@ import { utils as ethersUtils } from 'ethers';
 
 import { isNullish } from './typeof';
 import { Address, HexString, ProtocolType } from './types';
+import { assert } from './validation';
 
 const EVM_ADDRESS_REGEX = /^0x[a-fA-F0-9]{40}$/;
 const SEALEVEL_ADDRESS_REGEX = /^[a-zA-Z0-9]{36,44}$/;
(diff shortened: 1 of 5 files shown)


typescript/utils/src/addresses.ts
```diff
@@ -274,6 +277,11 @@ export function addressToBytes(
     new Uint8Array(),
     protocol,
   );
+  assert(
```
aroralanuk, line 280: nit: this check shouldn't this deep down, what if I wanted to purposefully have a zero address for burning tokens?

typescript/utils/src/addresses.ts
```diff
@@ -329,6 +337,10 @@ export function bytesToProtocolAddress(
   toProtocol: ProtocolType,
   prefix?: string,
 ) {
+  assert(
```
aroralanuk, line 340: similar

aroralanuk (approved):

typescript/utils/src/addresses.ts
jmrossy: Fair point. I did consider doing this in the adapters when forming txs but it would have required me to modify more than 10 classes. I'll keep it here for now but I'll watch for this in the UI's Sentry error monitoring and if we ever need to support 0 addresses I'm happy to move it.

PR #3429, opened 2024-03-16
Try Jito submission

tkporter:
### Description

Trying out Jito - if this does well, we can bring something cleaner to v3 in the future

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3434


rust/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -426,15 +546,29 @@ impl Mailbox for SealevelMailbox {
             .as_ref()
             .ok_or_else(|| ChainCommunicationError::SignerUnavailable)?;
 
-        let mut instructions = Vec::with_capacity(2);
+        let mut instructions = Vec::with_capacity(4);
```
daniel-savu, line 549: is this change still required? Looks like regardless of the if-else branches below, the same number of instructions gets pushed into `instructions`

tkporter: changed to 3 - the 2 was incorrect originally

PR #3430, opened 2024-03-16
No address filters

tkporter:


tkporter:
Opened for a hackathon participant that was using an fEVM (filecoin) subnet - their eth_getLogs implementation would give an `address` that is different from the contract that actually emitted the event

PR #3447, opened 2024-03-19
feat: add account -> owner mapping in ICA router

aroralanuk:
### Description

- added accountOwner mapping to the ICA router for easier tracking of remote owners for the transaction type inference by the AppGovernor

### Drive-by changes

none

### Related issues

none

### Backward compatibility

yes

### Testing

unit

### DIFF
--- a/solidity/contracts/middleware/InterchainAccountRouter.sol
+++ b/solidity/contracts/middleware/InterchainAccountRouter.sol
@@ -38,6 +38,11 @@ contract InterchainAccountRouter is Router {
     using TypeCasts for address;
     using TypeCasts for bytes32;
 
+    struct AccountOwner {
+        uint32 origin;
+        bytes32 owner; // remote owner
+    }
+
     // ============ Constants ============
 
     address internal implementation;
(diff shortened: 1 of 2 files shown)


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -45,6 +45,8 @@ contract InterchainAccountRouter is Router {
 
     // ============ Public Storage ============
     mapping(uint32 => bytes32) public isms;
+    // reverse lookup from the ICA account to the remote owner
+    mapping(address => bytes32) public accountOwners;
```
yorhodes, line 49: I was expecting this to be on the storage of the OwnableMulticall but this works too

yorhodes (commented):
consider `uint32 origin, address hubAccount` on the OwnableMulticall contract itself

yorhodes (approved):

solidity/contracts/middleware/InterchainAccountRouter.sol
aroralanuk: OwnableMulticall being proxied means we don't know the origin at the time of the contract creation.

PR #3466, opened 2024-03-21
fix: neutron grpc url default; url parsing error message

daniel-savu:


rust/hyperlane-base/src/settings/parser/mod.rs
```diff
@@ -422,7 +422,7 @@ fn parse_custom_urls(
         .end()
         .map(|urls| {
             urls.split(',')
-                .filter_map(|url| url.parse().take_err(err, || &chain.cwp + "customGrpcUrls"))
+                .filter_map(|url| url.parse().take_err(err, || &chain.cwp + key))
```
tkporter, line 425: nice catch

tkporter (approved):

PR #3422, opened 2024-03-15
Add codespell workflow and fix typos

miles-six:
Hey @jmrossy , I re-Pull request #3411  and ignored the ”receipient“

PR #3467, opened 2024-03-21
Add code spellcheck CI job

jmrossy:
### Description

Fix spelling mistakes and create codespell CI job.
An extension of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3422 with the remaining mistakes fixed

### Backward compatibility

Yes


.codespell/.codespellrc
```diff
@@ -0,0 +1,5 @@
+[codespell]
+skip = .git,node_modules,yarn.lock,Cargo.lock,./typescript/helloworld
```
jmrossy, line 2: There's some bug with codespell causing it to trip on `receivedFrom` in helloworld even though it's on the ignore list 🤷 
Other than that one occurrence it works fine.

rust/hyperlane-core/src/traits/validator_announce.rs
```diff
@@ -24,6 +24,6 @@ pub trait ValidatorAnnounce: HyperlaneContract + Send + Sync + Debug {
     ) -> ChainResult<TxOutcome>;
 
     /// Returns the number of additional tokens needed to pay for the announce
-    /// transaction. Return `None` if the needed tokens canno tbe determined.
+    /// transaction. Return `None` if the needed tokens canno the determined.
```
paulbalaji, line 27:
```suggestion
    /// transaction. Return `None` if the needed tokens cannot be determined.
```

typescript/infra/src/utils/metrics.ts
```diff
@@ -47,7 +47,7 @@ export async function submitMetrics(
  * Start a simple HTTP server to host metrics. This just takes the registry and dumps the text
  * string to people who request `GET /metrics`.
  *
- * PROMETHEUS_PORT env var is used to determine what port ot host on, defaults to 9090.
+ * PROMETHEUS_PORT env var is used to determine what host port, defaults to 9090.
```
paulbalaji, line 50:
```suggestion
 * PROMETHEUS_PORT env var is used to determine what port to host on, defaults to 9090.
```

paulbalaji (approved):
couple nits

yorhodes (approved):

Issue #3259, opened 2024-02-13
Use WarpCore in CLI send transfer command


jmrossy:
Once the WarpCore is implemented, tested, and merged into the SDK, the Warp UI will the updated to use it.
The CLI's `send transfer` command should use it as well. Currently, there's a janky, evm-only custom implementation in the CLI. Using WarpCore will DRY up the UI + CLI and allow the CLI to handle all cases.

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/commands/send.ts#L94

PR #3470, opened 2024-03-21
warpcore send transfer v1

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3259

- migrates the CLI warp transfer to WarpCore
- requires user to provide warp config generated by the previous `warp deploy` step

TODO:
- [ ] update IGP quoting log
- [ ] provide router options to user if addr not provided
	- [ ] add back router val as option (e.g. for ci-test.sh)


typescript/cli/ci-test.sh
```diff
@@ -163,6 +163,7 @@ MESSAGE1_ID=`cat /tmp/message1 | grep "Message ID" | grep -E -o '0x[0-9a-f]+'`
 echo "Message 1 ID: $MESSAGE1_ID"
 
 WARP_ARTIFACTS_FILE=`find /tmp/warp-route-deployment* -type f -exec ls -t1 {} + | head -1`
```
jmrossy, line 165: Totally optional but we could probably cut this and just use the WARP_CONFIG_FILE instead. Though it may be more cumbersome to extract a value from here in bash so feel free to ignore.

typescript/cli/src/deploy/warp.ts
```diff
@@ -255,7 +255,7 @@ async function executeDeploy(params: DeployParams) {
 
   const [contractsFilePath, tokenConfigPath] = prepNewArtifactsFiles(outPath, [
     { filename: 'warp-route-deployment', description: 'Contract addresses' },
-    { filename: 'warp-ui-token-config', description: 'Warp UI token config' },
+    { filename: 'warp-config', description: 'Warp config' },
```
jmrossy, line 258: I'm in favor of this rename but we'll need to update the docs too here: https://docs.hyperlane.xyz/docs/guides/deploy-warp-route#2-deployment

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
-    const token = ERC20__factory.connect(wrappedToken, connectedSigner);
-    const approval = await token.allowance(signerAddress, routerAddress);
-    if (approval.lt(wei)) {
-      const approveTx = await token.approve(routerAddress, wei);
-      await approveTx.wait();
+  if (!routerAddress) {
+    const tokensForRoute = warpCore.getTokensForRoute(origin, destination);
+    if (tokensForRoute.length === 0) {
+      logRed(`No Warp Routes found from ${origin} to ${destination}`);
+      return;
```
jmrossy, line 142: Might be better to throw here so the command has a failure exit code

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
     }
+
+    routerAddress = (await select({
+      message: `Select router address`,
+      choices: [
+        ...tokensForRoute.map((t) => ({
+          value: t.addressOrDenom,
```
jmrossy, line 149: It would be good to add a `description` field here with more info about the token, such as symbol and/or name: https://github.com/SBoudrias/Inquirer.js/tree/master/packages/select#usage

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
-  } else {
-    adapter = new EvmHypSyntheticAdapter(origin, multiProtocolProvider, {
-      token: routerAddress,
-    });
+  const token = warpCore.findToken(origin, routerAddress);
+  if (!token) {
+    logRed(
+      `No Warp Routes found from ${origin} to ${destination} with router address ${routerAddress}`,
+    );
+    return;
```
jmrossy, line 161: ditto

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
+  }
+
+  const senderAddress = await signer.getAddress();
+  const errors = await warpCore.validateTransfer({
+    originTokenAmount: new TokenAmount(wei, token),
```
jmrossy, line 166: totally fine but could also be `token.amount(wei)`

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
+    destination,
+    recipient: recipient ?? senderAddress,
+    sender: senderAddress,
+  });
+  if (errors) {
+    logRed('Unable to validate transfer');
```
jmrossy, line 172: let's also log `errors` here, there shouldn't be any sensitive info in it

typescript/cli/src/send/transfer.ts
```diff
@@ -179,74 +130,77 @@ async function executeDelivery({
...
-  const txReceipt = await multiProvider.handleTx(origin, txResponse);
-
-  const message = core.getDispatchedMessages(txReceipt)[0];
+  const transferTxReceipt = txReceipts[txReceipts.length - 1];
+
+  const message = core.getDispatchedMessages(transferTxReceipt)[0];
   logBlue(`Sent message from ${origin} to ${recipient} on ${destination}.`);
   logBlue(`Message ID: ${message.id}`);
+  logBlue(
+    `View on Hyperlane Explorer: https://explorer.hyperlane.xyz/message/${message.id}`,
```
jmrossy, line 198: This unfortunately won't work for PI chains the explorer doesn't know about. Users would need to add their custom chain metadata there first. It's possible to include chainmetadata along with the msg id but don't feel like you have to tackle that right now:
https://github.com/hyperlane-xyz/hyperlane-explorer/blob/main/src/features/chains/useChainConfigs.ts#L34

jmrossy (commented):
Overall looks great! Just a few minor things to fix

typescript/cli/src/send/transfer.ts
paulbalaji: ahh good point, I'll remove this line and leave it for now then

typescript/cli/src/send/transfer.ts
paulbalaji: poggers

typescript/cli/src/send/transfer.ts
paulbalaji: thinking we should throw here as well

typescript/cli/ci-test.sh
paulbalaji: jq to the rescue!

typescript/cli/src/deploy/warp.ts
paulbalaji: https://github.com/hyperlane-xyz/v3-docs/pull/72

jmrossy (approved):

PR #3471, opened 2024-03-21
fix: proxy factory impl names

paulbalaji:
Fixes issue where contract names we've configured in the deploy script do not match what they actually are in the current contracts

This surfaces up as an issue when attempting to verify these factories (cc @tkporter)


typescript/sdk/src/deploy/contracts.ts
```diff
@@ -21,9 +22,9 @@ type ProxyFactoryImplementations = Record<keyof ProxyFactoryFactories, string>;
...
 // must match contract names for verification
 export const proxyFactoryImplementations: ProxyFactoryImplementations = {
-  merkleRootMultisigIsmFactory: 'StaticMerkleRootMultisigIsm',
-  messageIdMultisigIsmFactory: 'StaticMessageIdMultisigIsm',
-  aggregationIsmFactory: 'StaticAggregationIsm',
-  aggregationHookFactory: 'StaticAggregationHook',
-  routingIsmFactory: 'DomaingRoutingIsm',
+  staticMerkleRootMultisigIsmFactory: 'StaticMerkleRootMultisigIsm',
+  staticMessageIdMultisigIsmFactory: 'StaticMessageIdMultisigIsm',
+  staticAggregationIsmFactory: 'StaticAggregationIsm',
```
paulbalaji, line 27: these three aren't in the `buildArtifact.json` for some reason

paulbalaji: fake news - stale artifact

typescript/sdk/src/deploy/contracts.ts
yorhodes, line 1: thanks for doing this

typescript/cli/examples/contract-artifacts.yaml
yorhodes, line 1: how can we make sure this doesnt go stale?

yorhodes (approved):

typescript/cli/examples/contract-artifacts.yaml
paulbalaji: not sure, but I can add a comment to the relevant part to let folks know they should update the example

PR #3472, opened 2024-03-21
Ancient8 deploy

tkporter:
### Description

- Deploys ancient8 core
- Deploys a USDC warp route from Ethereum to ancient8
- Some changes to the key generation step - specifically:
  - `scripts/create-keys.ts` now expects `--newChainValidators chainname=1` instead of the flag `--newChains`
  - fixed the logic around adding new validators, which previously was broken
- Adds a hook module to deploy.ts - this is meant to be used in a similar fashion to the warp module, where configs are built ad-hoc
- changes `aw-multisig.json` to be `aw-multisig-${context}.json`, as previously the entire aw-multisig.json would be overridden no matter what the current context was


typescript/infra/config/environments/mainnet3/core/verification.json
```diff
@@ -1,2144 +1,2830 @@
 {
```
tkporter, line 1: the reason for this big diff is we moved to deterministic JSON stringifying in a previous PR, and this hadn't been applied yet to verification.jsons. I used https://jsonformatter.org/json-compare to make sure that the only value diff here is just the addition of `ancient8`

typescript/infra/config/environments/mainnet3/ism/verification.json
```diff
@@ -1,2310 +1,2372 @@
 {
```
tkporter, line 1: same as core verification:

the reason for this big diff is we moved to deterministic JSON stringifying in a previous PR, and this hadn't been applied yet to verification.jsons. I used https://jsonformatter.org/json-compare to make sure that the only value diff here is just the addition of `ancient8`

rust/config/mainnet3_config.json
```diff
@@ -1,6 +1,61 @@
 {
   "chains": {
+    "ancient8": {
+      "aggregationHookFactory": "0xEb9FcFDC9EfDC17c1EC5E1dc085B98485da213D6",
```
paulbalaji, line 4: pls do the renamings as done in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3471

typescript/infra/config/aw-multisig-rc.json
yorhodes, line 1: surprised we have 3 validators on all these for rc but nbd

typescript/infra/scripts/print-token-prices.ts
```diff
@@ -20,7 +20,19 @@ async function main() {
 
   const idPrices = await resp.json();
 
-  const prices = objMap(ids, (_, id) => idPrices[id][CURRENCY].toString());
+  const prices = objMap(ids, (_, id) => {
+    const idData = idPrices[id];
+    if (!idData) {
+      throw new Error(
+        `No data for ${id}, did you set gasCurrencyCoinGeckoId in the metadata?`,
+      );
```
yorhodes, line 28: nice

typescript/infra/scripts/print-token-prices.ts
```diff
@@ -20,7 +20,19 @@ async function main() {
...
+    }
+    const price = idData[CURRENCY];
+    if (!price) {
+      throw new Error(`No price for ${id}`);
```
yorhodes, line 32: nit
```suggestion
      throw new Error(`No ${CURRENCY} price for ${id}`);
```

typescript/infra/scripts/deploy.ts
```diff
@@ -121,19 +128,46 @@ async function main() {
       multiProvider,
     );
     const routerConfig = core.getRouterConfig(envConfig.owners);
-    const inevm = {
-      ...routerConfig.inevm,
-      type: TokenType.native,
-      interchainSecurityModule: ethers.constants.AddressZero,
-      owner: safes.inevm,
+
+    const buildIsmConfig = (remote: string): IsmConfig => ({
```
yorhodes, line 132: I think we already have a function like this

typescript/infra/scripts/deploy.ts
```diff
@@ -121,19 +128,46 @@ async function main() {
...
+          ...defaultMultisigConfigs[remote],
+        },
+      ],
+    });
+
+    const ethereum = {
+      ...routerConfig.ethereum,
+      type: TokenType.collateral,
+      // USDC
+      token: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
```
yorhodes, line 151: honestly we should have collateral consts for this stuff

typescript/infra/scripts/deploy.ts
```diff
@@ -194,6 +228,24 @@ async function main() {
...
+      ismFactory,
+    );
+    // Config is intended to be changed for ad-hoc use cases:
+    config = {
+      ethereum: {
+        ...(coreConfig.ethereum.defaultHook as FallbackRoutingHookConfig)
+          .domains.ancient8,
+        owner: DEPLOYER,
+      },
+    };
```
yorhodes, line 248: interesting, I kind of expected this to be `core.defaultHook` but thats also a bit awkward

yorhodes (approved):

jmrossy (approved):
SDK changes lgtm

typescript/infra/config/environments/mainnet3/warp/addresses.json
paulbalaji, line 1: is this information in registry?

typescript/infra/config/aw-multisig-rc.json
tkporter: yeah probably worth cutting these down at some point

typescript/infra/scripts/deploy.ts
tkporter: honestly I don't totally remember why we did it like this but I feel like we had a good reason, something to do with contract ownership or keeping the deployed hook as simple as possible

typescript/infra/config/environments/mainnet3/warp/addresses.json
tkporter: no 😛

Issue #3098, opened 2023-12-22
Retry messages in the relayer queue via on demand API


tkporter:
If there's downtime or a chain has gone down etc, we should be able to easily tell the relayer to retry delivering all or some messages. 

Plan:
- Building upon Kunal's work in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3095, we'll create a new API path for the relayer (maybe `/relayer/`?) where we will have multiple endpoints
- one endpoint will allow us to request that the relayer retries a message or messages that match certain criteria
- The endpoint will take in either:
  - a message ID of a specific message (maybe allow for multiple? dunno)
  - a matching list, so we can match multiple messages easily (covering the case of retrying all messages to a chain or recipient etc)
- The endpoint's code will:
  - have a channel that it will use to send an action to serial submitters tasks (note we have one set of submitter tasks per destination chain - we'll probably need a channel that broadcasts from the endpoint logic to all of the destinations, e.g. in the case of a matching list that would match multiple dest chains, or a message ID where it's unknown what the dest chain is, etc)
  - when a serial submitter task receives an action on this channel, it should find any matching messages and place them at the front of the OpQueue. We may need to move away from a BinaryHeap to a BTreeMap for efficient ordering & lookup. We may also want to reset the # of retries for the message, though I don't feel strongly there

[architecture discord discussion](https://discord.com/channels/935678348330434570/1202202064864935957/1203041613157236836)

nambrot:
Wdyt of doing this as part of https://github.com/hyperlane-xyz/issues/issues/868

Make some simple way to query messages and their retry stats but then also reset or adjust them via the relayer API

tkporter:
Yeah it'd make sense to put it in there

tkporter:
We should decide whether this should be an HTTP API (which requires plumbing thoughout the agent codebase), or if this should be a separate command you run against a local hyperlane DB (which would involve some code that is external to the relayer, like a CLI)

nambrot:
Also thinking if explorer support is proximate enough to be an alternative

nambrot:
Happy for @tkporter to decide what the best trade off to go with is

daniel-savu:
> We should decide whether this should be an HTTP API #868 (which requires plumbing thoughout the agent codebase)

Ideally that plumbing is done by Kunal's work in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3095, which I'll review today

[...]

PR #3476, opened 2024-03-22
Relayer retry endpoints

daniel-savu:
### Description

Adds a `/message_retry` relayer endpoint, which supports retrying OpQueue operations either by ID or by destination domain. Such a request will be sent to an `mpmc` channel's receiving end in the OpQueue, which is emptied when `OpQueue::pop` is called.

Example calls:
```
GET http://127.0.0.1:60843/message_retry?destination_domain=42
GET http://127.0.0.1:60843/message_retry?message_id=0x46910b1329ee53c86a023b322e9ca1c17e5f9f0bee789c77b0abced0a173d714
```

If the endpoint is called specifying both filters, **two** requests will be sent across the channel, one for each condition.

Efficiency note: The entire queue is iterated over if there's at least one retry request in the channel. The good thing is that rebuilding the heap is `O(n)` so not too bad.

### Drive-by changes

- Removed the usage of `enum_dispatch`, because it would have made unit tests messy. The drawback is that we have to work around object safety restrictions in `PendingOperation` - for instance it can't be cloned (but can be stored behind an `Arc` if we ever need this). 
- Added new methods to `PendingOperation`, either to match by ID, or to implement heap element prioritization without depending on the concrete `PendingMessage` type (as was done before). Additions: `id()`, `origin_domain()`, `priority()`. Don't think this conflicts with any operations we may add in the future (e.g. gas oracle updates).
- Although `tokio`'s broadcast channel is multi-producer-multi-consumer by default, the consumer end isn't `Clone` - you instead need to have a producer to call `.subscribe()` on to get a new consumer, so I added an `MpmcChannel` struct to encapsulate keeping a producer around to get new consumers.
- OpQueue is moved into its own file and has unit tests added for retries (covering broadcast to multi-queue, and both retry types).
- Did a bit of agent `server `cleanup, including `EigenNodeApi`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3098

### Backward compatibility

Yes

### Testing

Unit tests for the server route (channel transmitter logic) and for the OpQueue retries (channel receiver logic).


rust/agents/relayer/src/server.rs
```diff
@@ -0,0 +1,170 @@
...
+    tx: Sender<MessageRetryRequest>,
+}
+
+#[derive(Deserialize)]
+struct RawMessageRetryRequest {
+    message_id: Option<String>,
+    destination_domain: Option<u32>,
+}
+
+impl From<RawMessageRetryRequest> for Vec<MessageRetryRequest> {
```
tkporter, line 53: wonder if this should be a TryFrom because it's possible to pass in a malformed message_id, it'd be nice to surface this error in the api response

rust/hyperlane-core/src/types/channel.rs
```diff
@@ -0,0 +1,50 @@
...
+    sender: Sender<T>,
+    /// The receiving end of the channel.
+    pub receiver: Receiver<T>,
+}
+
+impl<T> Clone for MpmcReceiver<T> {
+    fn clone(&self) -> Self {
+        Self {
+            sender: self.sender.clone(),
+            receiver: self.sender.subscribe(),
```
tkporter, line 47: Curious if you considered implications of creating a new subscriber each time we clone https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging

I think it's probably fine, in the worst case we'll have some subscribers that never read from the receiver, so at most we'll keep 1000 operations in the channel, and they'll start to be removed from the channel as new ones come in. Does make me wonder if we should have a lower capacity than 1000, but a few hundred Kb stuck around at most isn't a big deal

rust/hyperlane-base/src/server/base_server.rs
```diff
@@ -12,13 +12,21 @@ pub struct Server {
...
+        self.run_with_custom_routes(vec![])
+    }
+
     /// Run an HTTP server serving agent-specific different routes
     ///
     /// routes:
-    ///   - metrics - serving OpenMetrics format reports on `/metrics`
+    ///  - metrics - serving OpenMetrics format reports on `/metrics`
     ///     (this is compatible with Prometheus, which ought to be configured to scrape this endpoint)
     ///  - additional_routes - additional routes to be served by the server as per the specific agent
```
tkporter, line 25: nit: update comment too

tkporter (approved):
awesome, honestly I expected this to be more intrusive! very exciting

tkporter:
Looks like the WASM check in CI is complaining

rust/agents/relayer/src/msg/op_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+            return;
+        }
+        let mut queue = self.queue.lock().await;
+        let mut repriotized_queue: BinaryHeap<_> = queue
+            .drain()
+            .map(|Reverse(mut e)| {
+                // Can check for equality here because of the PartialEq implementation for MessageRetryRequest,
+                // but can't use `contains` because the types are different
+                if message_retry_requests.iter().any(|r| r == e) {
+                    let hex_id = e.id().encode_hex::<String>();
```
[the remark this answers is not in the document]
daniel-savu: I think doing `H256::to_string` will truncate the id so it's harder to debug

rust/agents/relayer/src/msg/op_queue.rs
```diff
@@ -0,0 +1,305 @@
...
+                    info!(
+                        id = hex_id,
+                        destination_domain, "Retrying OpQueue operation"
+                    );
+                    e.reset_attempts()
+                }
+                Reverse(e)
+            })
+            .collect();
+        queue.append(&mut repriotized_queue);
```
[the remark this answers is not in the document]
daniel-savu: we can't assign to queue directly because it's technically wrapped in a mutex. I think the reason we're able to call `mut` methods on it is deref corecion. 

I agree it would've been more efficient to just swap pointers instead of overwriting the heap, but this function is already O(N) so nbd

rust/agents/relayer/src/msg/serial_submitter.rs
```diff
@@ -126,14 +88,17 @@ impl SerialSubmitter {
...
         } = self;
         let prepare_queue = OpQueue::new(
             metrics.submitter_queue_length.clone(),
             "prepare_queue".to_string(),
+            retry_rx.clone(),
         );
         let confirm_queue = OpQueue::new(
             metrics.submitter_queue_length.clone(),
             "confirm_queue".to_string(),
+            retry_rx,
```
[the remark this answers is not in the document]
daniel-savu: discussed irl that retries in the confirm queue would make sense for chains with lots of forks or where finality can take a long time

rust/agents/relayer/src/server.rs
daniel-savu: good point, actually we were only returning status codes previously but I added error messages now

rust/hyperlane-core/src/types/channel.rs
daniel-savu: hadn't considered this exact edge case. I think that if there are any redundant receivers they're only temporary and are dropped eventually. Also don't think it's a big deal, would rather have very readable code for now

rust/agents/relayer/src/msg/op_queue.rs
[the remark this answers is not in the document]
tkporter: ah oops sorry, makes sense, forgot about the mutex

rust/hyperlane-base/src/server/base_server.rs
tkporter: oh just the rename of additional_routes -> custom_routes

rust/agents/relayer/src/msg/op_queue.rs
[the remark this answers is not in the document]
tkporter: I think to_string does truncate but the Debug implementation doesn't truncate:
```
    #[test]
    fn test_hxxx() {
        let h160 = H160::zero();
        println!("h160 {:?}", h160);

        let h256 = H256::zero();
        println!("h256 {:?}", h256);
    }

---- mailbox::test::test_hxxx stdout ----
h160 0x0000000000000000000000000000000000000000
h256 0x0000000000000000000000000000000000000000000000000000000000000000
```
so I was suggesting just doing `id=?hex_id`

PR #3481, opened 2024-03-25
Logging overhaul

jmrossy:
### Description

- Replace use of debug lib with Pino
- Replace `loggerName` params with Logger types
- Update SDK Readme

### Related issues

Fixes https://github.com/hyperlane-xyz/issues/issues/918

### Backward compatibility

No, breaks anyone passing in custom logger to SDK classes, which is likely no one

### Testing

- Tested in CLI
- Tested in Warp UI


typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
+  },
+  hooks: {
+    logMethod(inputArgs, method, level) {
+      // Pino has no simple way of setting custom log shapes and they
+      // recommend against using pino-pretty in production so when
+      // pretty is enabled we circumvent pino and log directly to console
+      if (ENV_LOG_PRETTY && level >= pino.levels.values[ENV_LOG_LEVEL]) {
+        // eslint-disable-next-line no-console
+        console.log(inputArgs[0]);
+        return null;
```
yorhodes, line 31: I dont really understand this

typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
+if (ENV_LOG_LEVEL === 'none' || ENV_LOG_LEVEL === 'off')
+  ENV_LOG_LEVEL = 'silent';
+
+const ENV_LOG_PRETTY = envVarToBoolean(safelyAccessEnvVar('LOG_PRETTY'));
+
+export const rootLogger = pino({
+  level: ENV_LOG_LEVEL,
+  name: 'hyperlane',
+  formatters: {
+    // Remove pino's default bindings of hostname and pid
```
yorhodes, line 20: pid seems kinda useful?

typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
-    level,
-    message,
-  };
-  logFn(JSON.stringify(fullLog));
-}
+import { LevelWithSilent, pino } from 'pino';
+
+import { envVarToBoolean, safelyAccessEnvVar } from './env';
+
+const DEFAULT_LOG_LEVEL = 'warn';
```
yorhodes, line 5: surprised that this is warn and not info

typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
+
+let ENV_LOG_LEVEL = (
+  safelyAccessEnvVar('LOG_LEVEL') ?? DEFAULT_LOG_LEVEL
+).toLowerCase() as LevelWithSilent | 'none' | 'off';
```
yorhodes, line 9: seems like we want the ENV_LOG_LEVEL to be a pino type? can we do this transform from none|off to silent before the cast?

typescript/cli/env.ts
```diff
@@ -0,0 +1,19 @@
+// This file isn't in the src dir so it it's imported before others
+// See https://github.com/trivago/prettier-plugin-sort-imports/issues/112
+
+// Default rootLogger into pretty mode and info level
+process.env.LOG_LEVEL ??= 'info';
+process.env.LOG_PRETTY ??= 'true';
+
+// Workaround for bug in bigint-buffer which solana-web3.js depends on
```
yorhodes, line 8: does this suppress the failed to load bindings log?

typescript/cli/src/logger.ts
yorhodes, line 1: did we need to move this? seems like we could have avoided a lot of diff here

typescript/utils/src/env.ts
```diff
@@ -7,3 +7,10 @@ export function safelyAccessEnvVar(name: string) {
     return undefined;
   }
 }
+
+export function envVarToBoolean(value: any) {
```
yorhodes, line 11: should this be zod?

typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
+  formatters: {
+    // Remove pino's default bindings of hostname and pid
+    bindings: () => ({}),
+  },
+  hooks: {
+    logMethod(inputArgs, method, level) {
+      // Pino has no simple way of setting custom log shapes and they
+      // recommend against using pino-pretty in production so when
+      // pretty is enabled we circumvent pino and log directly to console
+      if (ENV_LOG_PRETTY && level >= pino.levels.values[ENV_LOG_LEVEL]) {
```
yorhodes, line 28: what does the level check do?

typescript/utils/src/logging.ts
```diff
@@ -1,47 +1,36 @@
...
+
+const DEFAULT_LOG_LEVEL = 'warn';
+
+let ENV_LOG_LEVEL = (
+  safelyAccessEnvVar('LOG_LEVEL') ?? DEFAULT_LOG_LEVEL
+).toLowerCase() as LevelWithSilent | 'none' | 'off';
+// For backwards compat and also to match agent level options
+if (ENV_LOG_LEVEL === 'none' || ENV_LOG_LEVEL === 'off')
+  ENV_LOG_LEVEL = 'silent';
+
```
yorhodes, line 13: can we transform from the env var to `LevelWithSilent` before casting?

typescript/utils/src/logging.ts
yorhodes, line 5: surprised that this defaults to warn instead of info

yorhodes (approved):

typescript/cli/src/logger.ts
jmrossy: The need for it to be outside the `src` folder no longer applies because I had to make a new `.env` file which executed before pino. Given that, it seemed right to have it be in src/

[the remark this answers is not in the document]
jmrossy: I considered it but if we use pino doesn't support colors (which `chalk` gives us). 
If we assume the CLI will only ever log in pretty mode, then I can make these work via pino. Otherwise we'd have to have some kind of split (colored vs pino)

typescript/utils/src/env.ts
jmrossy: I'll double check but not sure zod is a good fit for type coercion like this

typescript/utils/src/logging.ts
jmrossy: I was flip-flopping on this. Thanks for the second opinion, I'll make it `info`

typescript/utils/src/logging.ts
jmrossy: Sure np

typescript/utils/src/logging.ts
jmrossy: Cool np I'll keep that then

typescript/utils/src/logging.ts
jmrossy: It avoids console logging lines that are below the set log level threshold (e.g. hide debug logs)

typescript/utils/src/logging.ts
jmrossy: Does the comment above help? We're constrained by what pino lets us do. Short of restarting this work with a different lib, I'm not sure we can do better

Issue #3457, opened 2024-03-20
Warp route deployment verification is broken


yorhodes:
## Problem

The [HyperlaneDeployer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/02d5549608e97401edc5028ba59de0809c5a459e/typescript/sdk/src/deploy/verify/ContractVerifier.ts#L230) assumes that the verification `input.name` matches the contract name. 

The [HypERC20Deployer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dcb67e97da6e0e3d9abd1533dc2a5ca2ff9b4617/typescript/sdk/src/token/deploy.ts#L138) now uses the `TokenType` as the `contractName` but these do not match the actual contract name.

## Solution

Adjust `HypERC20Deployer` to use the correct contract names.

- [ ] Test warp route deployment **and verification** on testnet

paulbalaji:
useful reference https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3379/files#diff-a7e3646e88e11cdab899cb99fa6e62f74893bdb0415e58879668cb79ff0ad3ec

PR #3490, opened 2024-03-26
fix: warp route deployment verification

paulbalaji:
fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3457

should fix the problem whilst also not breaking the format of existing artifact read/writes

drive-by fix to contractverifier - allows propagation of apikeys from custom chain metadata (e.g. chains.yaml)

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -20,6 +20,10 @@ import {
   WarpCoreConfig,
   getTokenConnectionId,
 } from '@hyperlane-xyz/sdk';
+import {
+  hypERC20contracts,
+  hypERC721contracts,
+} from '@hyperlane-xyz/sdk/dist/token/contracts.js';
 import { Address, ProtocolType, objMap } from '@hyperlane-xyz/utils';
 
 import { log, logBlue, logGray, logGreen } from '../../logger.js';
@@ -255,7 +259,7 @@ async function executeDeploy(params: DeployParams) {
 
   const [contractsFilePath, tokenConfigPath] = prepNewArtifactsFiles(outPath, [
     { filename: 'warp-route-deployment', description: 'Contract addresses' },
-    { filename: 'warp-ui-token-config', description: 'Warp UI token config' },
+    { filename: 'warp-config', description: 'Warp config' },
   ]);
 
   const deployer = isNft
@@ -271,7 +275,7 @@ async function executeDeploy(params: DeployParams) {
 
   logBlue('Deployment is complete!');
   logBlue(`Contract address artifacts are in ${contractsFilePath}`);
-  logBlue(`Warp UI token config is in ${tokenConfigPath}`);
+  logBlue(`Warp config is in ${tokenConfigPath}`);
 }
 
 async function fetchBaseTokenMetadata(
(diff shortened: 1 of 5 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -255,7 +257,7 @@ async function executeDeploy(params: DeployParams) {
 
   const [contractsFilePath, tokenConfigPath] = prepNewArtifactsFiles(outPath, [
     { filename: 'warp-route-deployment', description: 'Contract addresses' },
-    { filename: 'warp-ui-token-config', description: 'Warp UI token config' },
+    { filename: 'warp-config', description: 'Warp config' },
```
yorhodes, line 260: why rename this? this shape is different from the warp deploy config

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -73,7 +73,11 @@ export class ContractVerifier {
     verificationLogger: Debugger,
     options?: FormOptions<typeof action>,
   ): Promise<any> {
-    const { apiUrl, family } = this.multiProvider.getExplorerApi(chain);
+    const {
+      apiUrl,
+      apiKey = this.apiKeys[chain],
```
yorhodes, line 78: this is slightly hard to read

typescript/sdk/src/token/contracts.ts
yorhodes, line 1: honestly I would be happy to change the serialization format from
`[TokenType]: address` to `[ContractName]: address`, which is more consistent with the rest of the codebase

yorhodes: (albeit this is a breaking change)

typescript/cli/src/deploy/warp.ts
```diff
@@ -316,40 +318,47 @@ function getTokenName(token: TokenConfig) {
...
   }> = objMap(contracts, (chain, contract) => {
+    const tokenType = configMap[chain].type;
+    const contractName = isNft
+      ? hypERC721contracts[tokenType as keyof typeof hypERC721contracts]
+      : hypERC20contracts[tokenType as keyof typeof hypERC20contracts];
     return {
-      router: contract[configMap[chain].type as keyof TokenFactories].address,
-      tokenType: configMap[chain].type,
+      router: contract[contractName as keyof TokenFactories]['address'],
+      tokenType,
```
yorhodes, line 333: honestly im not a fan of this different artifact shape
its another unnecessary layer of indirection

yorhodes: [ContractSourceName]: address is more desirable
is this format recovered too?

yorhodes (approved):
looks good
please add a changeset

ltyu (approved):

typescript/cli/src/deploy/warp.ts
paulbalaji: yeah it's already `warp-config` on main, unsure why it's showing up as a change here

paulbalaji:
created an alternative fix that doesn't change any of the warp artifact logic, simply creates a `deployContractWithName` that lets you override the contractName used at deploy time without changing the factory key 
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3494

Issue #3264, opened 2024-02-15
Refactor watermark-based / rate limited indexing


tkporter:
Similar refactor to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3262, but now for watermark-based / rate limited indexing

When indexing data types that are not always sequenced (like IGP payments and message deliveries), we use the `WatermarkContractSync` using the `RateLimitedContractSyncCursor`.

`RateLimitedContractSyncCursor` is a forward-only cursor. The name reflects that - the "rate limiting" part only kicks in when we're close to the tip, this way we don't end up making a bunch of expensive eth_getLogs RPCs (or alternatives for other chains) when e.g. only one new block has come in. Much of the motivation for this was to do with RPC providers charging us a flat cost for a single eth_getLogs request regardless of how many blocks we were actually querying.

Requirements:
- re-evaluate the goals of watermark indexing to begin with - does the rationale behind it still even make sense? this feels rather ethereum-specific tbh -- for instance when we're doing block based indexing when indexing cosmos chain IGP payments, does it really make sense for us to rate limit ourselves at all?
- should we really rate limit ourselves up to 30 seconds? Because we use this for indexing IGP payments, this means we're only polling for new IGP payments every 30 seconds, and therefore if gas payments are being enforced, we're increasing the message delivery latency! Should this be configurable or related to the block time? Feel free to create issues out of this, rather than putting them in this refactor. See https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2308 and sleeping done in other cursors, e.g. https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/18d1d0d54eda245c4758b61261a003f7445edb18/rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs#L396
- consider whether the Sleep action makes sense or if the cursor itself should just sleep (see https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3262#discussion_r1493816338)
- simpler logic - it's too confusing
- unit test it rigorously
- if an error occurs getting a range of logs, that range should be retried. We shouldn't silently move on to the next range!
  - `next_action` should not move the cursor forward - only a call to `update`, which indicates a successful fetching of logs, should
  - remove retrying logic from `rust/chains/hyperlane-cosmos/src/providers/rpc.rs`:`get_logs_in_block`. Instead, propagate errors and make sure failed queries are retried.
[...]

PR #3500, opened 2024-03-27
Range retries in watermark cursor

daniel-savu:
### Description

Changes the logic in the watermark cursor, to update the range in the `update` method rather than in `get_next_range`. This means that RPC errors won't result in ranges being skipped, so IGP indexing should be more reliable.

### Related issues

- First in a sequence of many, for https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3264


rust/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -242,27 +261,36 @@ where
...
+
+#[async_trait]
+impl<T> ContractSyncCursor<T> for RateLimitedContractSyncCursor<T>
+where
+    T: Send + Debug + 'static,
+{
+    async fn next_action(&mut self) -> Result<(CursorAction, Duration)> {
+        let eta = self.sync_eta()?;
 
         let rate_limit = self.get_rate_limit().await?;
```
tkporter, line 276: noting that in this call we update the `tip` and `last_tip_update` regardless of us ever successfully updating the cursor. so if there's an error and we're close to the tip we'll rate limit ourselves and only retry 30s from now

I think this is fine for now but as you continue to refactor we should probably consider this case

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -242,27 +261,36 @@ where
...
-        // TODO: Define the sleep time from interval flag
-        Ok((CursorAction::Sleep(Duration::from_secs(5)), eta))
     }
 
     fn latest_queried_block(&self) -> u32 {
         self.sync_state.next_block.saturating_sub(1)
     }
 
-    async fn update(&mut self, _: Vec<(T, LogMeta)>, _range: RangeInclusive<u32>) -> Result<()> {
+    async fn update(&mut self, _: Vec<(T, LogMeta)>, range: RangeInclusive<u32>) -> Result<()> {
```
tkporter, line 293: what should we do for Cosmos IGP indexing? Atm `fetch_logs` doesn't propagate any errors that we may get along the way, so we end up calling update here and move onto the next range even if we skipped some blocks https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dan/watermark-refactoring/rust/chains/hyperlane-cosmos/src/interchain_gas.rs#L231

Open to this being addressed in a different PR but we should try to have a consistent interface between chain log fetching impls when it comes to errors

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -80,13 +80,11 @@ impl SyncState {
                 let from = self.next_block;
                 let mut to = from + self.chunk_size;
                 to = u32::min(to, tip);
-                self.next_block = to + 1;
```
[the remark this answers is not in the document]
daniel-savu: good point, changed

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
daniel-savu: pushed a fix for this, lmk if it lgty

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
daniel-savu: I think this was the case before right? Less critical than other bits but I took a note

PR #3492, opened 2024-03-26
By default, don't overwrite pushgateway metrics

tkporter:
### Description

- v2 backport of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3491

PR #3506, opened 2024-03-28
Different key funder job name

tkporter:
### Description

- My changes in #3491 / #3492 were kind of accurate - according to https://github.com/prometheus/pushgateway/issues/65 even using pushAdd will displace existing metrics. It's all very confusing!
- I tested this with a script that pushed some test metrics to the push gateway - basically by having the key funder use a different job name than the mainnet2 one, the metrics each of them push won't conflict with one another


daniel-savu (approved):
Did a bit of reading, the grouping key must default to the job name or something. It's so poorly documented

PR #3514, opened 2024-03-28
feat: CLI chain list derived from SDK contract addresses artifact

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/1066

- uses the list of deployed addresses from SDK to determine core mainnet/testnet chains
- tidy up `hyperlane chains list` command to accept `env` (either `mainnet` or `testnet`)

will do docs update to reflect updated chains list command

before:
image

after:
image


nbayindirli (approved):
nice

PR #3517, opened 2024-03-28
feat: add HypERC20 checker

aroralanuk:
### Description

- Added HypERC20App (extends GasRouterApp) and HypERC20Checker (extends RouterChecker bc there's no GasRouterChecker - link to issue)
- remoteChains fetches the remote chains from the local router address is it given. This means, the abstract routerAddress in multiGeneric had to be made and this call had to awaited in a few places where it's used. 
**Why?** previously we assumed that the routers will be fully-connected and will match 1:1 with the config. This is too strong of an assumption for ICAs given we inherit the config through `getRouterConfig` and the ICA accounts will only need to exist on chains which don't have safes. Plus, in some cases, the chains implied by the app factories vs checker config don't match  and this leads to further issues down the line while checking for mailbox client properties or in the govern. Checking the onchain enrollments is the most safe and resilient approach we can take especially given that having a checker for every RouterApp isn't feasible.

### Drive-by changes

- adding native keys to warp artifacts for the appHelper to work in App.fromAddressMap. Note: if there's something using the router key, this isn't breaking unless you expect only one key.

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1191

### Backward compatibility

Yes

### Testing

Manual


typescript/infra/scripts/check-deploy.ts
```diff
@@ -111,6 +115,42 @@ async function check() {
...
+      .reduce((obj, key) => {
+        obj[key] = addresses[key];
+        return obj;
+      }, {} as typeof addresses);
+    const app = HypERC20App.fromAddressesMap(filteredAddresses, multiProvider);
+
+    const checker = new HypERC20Checker(
+      multiProvider,
+      app,
+      config as any,
```
yorhodes, line 150: why `as any`?

typescript/sdk/src/router/RouterApps.ts
```diff
@@ -44,11 +44,13 @@ export abstract class RouterApp<
...
-    return [
-      ...super.remoteChains(chainName),
-      ...Object.keys(this.foreignDeployments),
-    ].filter((chain) => chain !== chainName);
+  // check onchain for remote enrollments
+  override async remoteChains(chainName: string): Promise<ChainName[]> {
+    const router = this.router(this.contractsMap[chainName]);
+    const domains = (await router.domains())
+      .map((domain) => this.multiProvider.tryGetChainName(domain))
+      .filter((domain): domain is string => domain !== null);
```
yorhodes, line 52: this will remove domains that we dont have a chainname for which seems dangerous
maybe throw?

typescript/sdk/src/token/app.ts
```diff
@@ -0,0 +1,35 @@
...
+    super(contractsMap, multiProvider);
+  }
+
+  router(contracts: HyperlaneContracts<HypERC20Factories>): any {
+    for (const key of objKeys(hypERC20factories)) {
+      if (contracts[key]) {
+        return contracts[key];
+      }
+    }
+  }
```
yorhodes, line 22: I worry a bit this will break down if eg there is an ISM or hook or something in the artifacts

typescript/sdk/src/token/checker.ts
```diff
@@ -0,0 +1,90 @@
...
+> {
+  async checkChain(chain: ChainName): Promise<void> {
+    await super.checkChain(chain);
+    await this.checkToken(chain);
+  }
+
+  async checkToken(chain: ChainName): Promise<void> {
+    const checkERC20 = async (
+      token: ERC20,
+      config: HypERC20Config | boolean,
```
yorhodes, line 33: why `| boolean` ?

typescript/sdk/src/utils/MultiGeneric.ts
```diff
@@ -54,7 +54,7 @@ export class MultiGeneric<Value> {
     return Object.fromEntries(entries);
   }
 
-  remoteChains(name: ChainName): ChainName[] {
+  async remoteChains(name: ChainName): Promise<ChainName[]> {
     return this.chains().filter((key) => key !== name);
```
yorhodes, line 58: this isnt async, why change here?

typescript/sdk/src/token/checker.ts
yorhodes, line 1: I can see this being useful

typescript/sdk/src/router/RouterApps.ts
aroralanuk: done

typescript/sdk/src/token/app.ts
aroralanuk: it looks for the specific type, how does ISM or hook change it?

typescript/sdk/src/token/checker.ts
aroralanuk: I'm overloading this for the "if the ERC20 exists here" check

typescript/sdk/src/utils/MultiGeneric.ts
aroralanuk: this fetches from onchain, it's async

typescript/infra/config/environments/mainnet3/warp/addresses.json
```diff
@@ -1,9 +1,10 @@
 {
   "injective": {
+    "native": "inj1mv9tjvkaw7x8w8y9vds8pkfq46g2vcfkjehc6k",
     "router": "inj1mv9tjvkaw7x8w8y9vds8pkfq46g2vcfkjehc6k"
```
jmrossy, line 4: Can you clarify why here there are both type + router keys but in the file below there are just type keys?

typescript/infra/scripts/check-deploy.ts
```diff
@@ -111,6 +115,42 @@ async function check() {
...
+    // test config
+    const plumetestnet = {
+      ...routerConfig.plumetestnet,
+      type: TokenType.synthetic,
+      name: 'Wrapped Ether',
+      symbol: 'WETH',
+      decimals: 18,
+      totalSupply: '0',
+      gas: 0,
+    };
```
jmrossy, line 128: nbd but should the script take this info as params or something? Or at least have the consts defined at the top

typescript/sdk/src/router/RouterApps.ts
```diff
@@ -44,11 +44,17 @@ export abstract class RouterApp<
...
 
-  override remoteChains(chainName: string): string[] {
-    return [
-      ...super.remoteChains(chainName),
-      ...Object.keys(this.foreignDeployments),
-    ].filter((chain) => chain !== chainName);
+  // check onchain for remote enrollments
+  override async remoteChains(chainName: string): Promise<ChainName[]> {
+    const router = this.router(this.contractsMap[chainName]);
+    const domains = (await router.domains()).map((domain) => {
```
jmrossy, line 50: Nit: rename this, the name made me thing it was a domainId

typescript/sdk/src/token/app.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { appFromAddressesMapHelper } from '../contracts/contracts';
+import { HyperlaneAddressesMap, HyperlaneContracts } from '../contracts/types';
+import { MultiProvider } from '../providers/MultiProvider';
+import { GasRouterApp } from '../router/RouterApps';
+
+import { HypERC20Factories, hypERC20factories } from './contracts';
+
+export class HypERC20App extends GasRouterApp<HypERC20Factories, TokenRouter> {
+  constructor(contractsMap: any, multiProvider: any) {
```
jmrossy, line 12: any any?

typescript/sdk/src/token/app.ts
```diff
@@ -0,0 +1,35 @@
...
+    super(contractsMap, multiProvider);
+  }
+
+  router(contracts: HyperlaneContracts<HypERC20Factories>): any {
```
jmrossy, line 16: I definitely sympathize that the types in these class hierarchies can be cumbersome but we should remain strict about the param and return types in function definitions. Forcing a cast inside the fn is less of an issue but strongly typed interfaces are worthwhile

typescript/sdk/src/token/checker.ts
```diff
@@ -0,0 +1,90 @@
...
+> {
+  async checkChain(chain: ChainName): Promise<void> {
+    await super.checkChain(chain);
+    await this.checkToken(chain);
+  }
+
+  async checkToken(chain: ChainName): Promise<void> {
+    const checkERC20 = async (
+      token: ERC20,
+      config: HypERC20Config | boolean, // either check for validity if config is present, or check for existence if config is boolean
```
jmrossy, line 33: I recommend against using union types like this. It's fine when the fn's behavior won't vary (e.g. `add(value:string|number)`) but encoding behavioral differences like this feels fragile. WDYT?

typescript/sdk/src/utils/MultiGeneric.ts
jmrossy: Yorke is correct, this isn't async, only the HypErc20App child is. But you may need to make this parent async too for the class types to be valid, I assume that's why you changed it.

I believe the original intention of these MultiGeneric-based classes was to have the `App` children take all the configs up front at construction time so basic methods like this wouldn't need to be async. This new approach deviates from that but it's probably fine.  

This multi-generic concept of shaping a class around a known map of `ChainName` to a single `Value` makes sense in some cases but not in others. That's partly why I didn't make WarpCore a HypErc20App; its configs are more complex.

typescript/infra/config/environments/mainnet3/warp/addresses.json
aroralanuk: Native is needed for artifact recovery for EVM deployment but not needed for cosmwasm. Just put it for symmetry.

typescript/infra/scripts/check-deploy.ts
aroralanuk: Ya, that would be better. I was just mimicking the deployer setup.

typescript/sdk/src/router/RouterApps.ts
aroralanuk: done

typescript/infra/scripts/check-deploy.ts
aroralanuk: fixed

typescript/sdk/src/token/checker.ts
aroralanuk: talked to yorke and simplified it

typescript/sdk/src/token/app.ts
aroralanuk: fixed

typescript/sdk/src/token/app.ts
aroralanuk: this "any" is hard where the the hypERC20Factories value types and the TokenRouter types don't line up.

typescript/sdk/src/token/checker.ts
aroralanuk: fixed

typescript/infra/scripts/check-deploy.ts
aroralanuk: not fixed

jmrossy (approved):
A few more comments and recommend a ✅ from Yorke too but overall lgtm

typescript/sdk/src/utils/MultiGeneric.ts
yorhodes: this is nbd to me

typescript/sdk/src/token/app.ts
aroralanuk: resolved

PR #3519, opened 2024-03-28
Add global log options to CLI

jmrossy:
### Description

- Add global log options to CLI
- Improve consistency with agent log options
- Other minor log-related cleanup

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3499

### Backward compatibility

Yes because the new log options haven't yet published


typescript/utils/src/logging.ts
```diff
@@ -1,38 +1,98 @@
...
+export function getLogFormat() {
+  return logFormat;
+}
+
+// Note, for brevity and convenience, the rootLogger is exported directly
+export let rootLogger = createHyperlanePinoLogger(logLevel, logFormat);
+
+export function getRootLogger() {
+  return rootLogger;
+}
```
yorhodes, line 52: why is this useful?

typescript/utils/src/logging.ts
```diff
@@ -1,38 +1,98 @@
...
+  logFormat = newLogFormat;
+  logLevel = toPinoLevel(newLogLevel) || logLevel;
+  rootLogger = createHyperlanePinoLogger(logLevel, logFormat);
+  return rootLogger;
+}
+
+export function setRootLogger(logger: Logger) {
+  rootLogger = logger;
+  return rootLogger;
+}
```
yorhodes, line 67: ditto on these global exports that mutate

yorhodes (approved):

typescript/utils/src/logging.ts
```diff
@@ -1,38 +1,98 @@
...
+// and because we use 'off' instead of 'silent' to match the agent options
+export enum LogLevel {
+  Debug = 'debug',
+  Info = 'info',
+  Warn = 'warn',
+  Error = 'error',
+  Off = 'off',
+}
+
+let logLevel: LevelWithSilent =
```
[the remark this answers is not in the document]
jmrossy: It's a good question. JS module-level state is safe and practical but does still have the downsides that all global state has. There are alternatives they all put the onus of managing this state on the consumers of these logging utilities. It's not worth it IMHO

typescript/utils/src/logging.ts
jmrossy: Maybe it's not, I'll remove it. It felt odd to have a `set` function without a `get`

typescript/cli/src/logger.ts
```diff
@@ -1,9 +1,24 @@
...
+  safelyAccessEnvVar,
+} from '@hyperlane-xyz/utils';
+
+let logger = rootLogger.child({ module: 'cli' });
+
+export function configureLogger(logFormat: LogFormat, logLevel: LogLevel) {
+  logFormat =
+    logFormat || safelyAccessEnvVar('LOG_FORMAT', true) || LogFormat.Pretty;
+  logLevel = logLevel || safelyAccessEnvVar('LOG_LEVEL', true) || LogLevel.Info;
+  logger = configureRootLogger(logFormat, logLevel).child({ module: 'cli' });
```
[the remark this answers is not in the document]
jmrossy: It was a precaution in case something logs before the yargs middleware runs but it's probably sufficient to just use rootLogger itself for that case. I'll change it

Issue #3129, opened 2024-01-04
Derive `HookConfig` from `IPostDispatchHook` address


yorhodes:
## Problem

From a `IPostDispatchHook` address, we should be able to derive the [`HookConfig`](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0727a6178eaea332788a30cc2871a8865b28c34c/typescript/sdk/src/hook/types.ts#L61) that this contract reflects from  onchain.

## Solution

- Recursive derivation function from a hook address

```typescript
function deriveConfig(address hook): Promise<HookConfig>;
```

- Should include modular derivation functions for children:

```typescript
function deriveIgpConfig(address igpHook): Promise<IgpConfig>;
```

This represents the Read operation of Hook CRUD management

The CLI command for this is tracked separately in https://github.com/hyperlane-xyz/issues/issues/1183

avious00:
@yorhodes @nambrot is this other issue related? https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2797 can we collapse them

avious00:
@yorhodes to refresh this ticket

PR #3543, opened 2024-04-03
feat: derive `HookConfig` given a hook address

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3129

- `EvmHookReader` class to derive `HookConfig` given an EVM address
	- maps are not enumerable, so temporary compromise is to iterate over known domainIds
	- stringify method with custom replacer to support `BigNumber`
	- `disableConcurrency` option to avoid hitting issues where certain RPCs can't handle a reasonable load of requests
- `read-hook.ts` script for reading hook configs before CLI integration
	- doesn't use multiprovider from envConfig, as it needs to know about non-EVM chains
- drive-by fix `getKnownDomainIds()`

TODO:
- [x] include addresses of all children in the returned `HookConfig`
- [x] tests

### DIFF
--- a/typescript/cli/src/config/hooks.ts
+++ b/typescript/cli/src/config/hooks.ts
@@ -33,6 +33,7 @@ const ProtocolFeeSchema = z.object({
 });
 
 const MerkleTreeSchema = z.object({
+  owner: z.string(),
   type: z.literal(HookType.MERKLE_TREE),
 });
 
@@ -90,6 +91,7 @@ export function presetHookConfigs(owner: Address): HooksConfig {
       owner: owner,
     },
     default: {
+      owner,
       type: HookType.MERKLE_TREE,
     },
   };
--- a/typescript/cli/src/config/hooks.ts
+++ b/typescript/cli/src/config/hooks.ts
@@ -195,7 +195,7 @@ export async function createHookConfig(
     pageSize: 10,
   });
   if (hookType === HookType.MERKLE_TREE) {
-    lastConfig = { type: HookType.MERKLE_TREE };
+    lastConfig = createMerkleRootConfig();
   } else if (hookType === HookType.PROTOCOL_FEE) {
     lastConfig = await createProtocolFeeConfig(chain);
   } else if (hookType === HookType.INTERCHAIN_GAS_PAYMASTER) {
@@ -208,6 +208,14 @@ export async function createHookConfig(
     throw new Error(`Invalid hook type: ${hookType}`);
   }
   return lastConfig;
+}
+
+export async function createMerkleRootConfig(): Promise<HookConfig> {
+  const owner = await input({
+    message: 'Enter owner address',
+  });
+  const ownerAddress = normalizeAddressEvm(owner);
+  return { owner: ownerAddress, type: HookType.MERKLE_TREE };
 }
 
 export async function createProtocolFeeConfig(
--- a/typescript/cli/src/config/hooks.ts
+++ b/typescript/cli/src/config/hooks.ts
@@ -33,7 +33,6 @@ const ProtocolFeeSchema = z.object({
 });
 
 const MerkleTreeSchema = z.object({
-  owner: z.string(),
   type: z.literal(HookType.MERKLE_TREE),
 });
 
@@ -91,7 +90,6 @@ export function presetHookConfigs(owner: Address): HooksConfig {
       owner: owner,
     },
     default: {
-      owner,
       type: HookType.MERKLE_TREE,
     },
   };
@@ -195,7 +193,7 @@ export async function createHookConfig(
     pageSize: 10,
   });
   if (hookType === HookType.MERKLE_TREE) {
-    lastConfig = createMerkleRootConfig();
+    lastConfig = { type: HookType.MERKLE_TREE };
   } else if (hookType === HookType.PROTOCOL_FEE) {
     lastConfig = await createProtocolFeeConfig(chain);
   } else if (hookType === HookType.INTERCHAIN_GAS_PAYMASTER) {
@@ -208,14 +206,6 @@ export async function createHookConfig(
     throw new Error(`Invalid hook type: ${hookType}`);
   }
   return lastConfig;
-}
-
-export async function createMerkleRootConfig(): Promise<HookConfig> {
-  const owner = await input({
-    message: 'Enter owner address',
-  });
-  const ownerAddress = normalizeAddressEvm(owner);
-  return { owner: ownerAddress, type: HookType.MERKLE_TREE };
 }
 
 export async function createProtocolFeeConfig(
--- a/typescript/infra/scripts/read-hook.ts
+++ b/typescript/infra/scripts/read-hook.ts
@@ -18,12 +18,15 @@ import {
 //     yarn tsx scripts/read-hook.ts -e mainnet3 --network inevm --hookAddress 0x19dc38aeae620380430C200a6E990D5Af5480117
 
 async function readHook() {
-  const { environment, network, hookAddress, context, _concurrency } =
+  const { environment, network, hookAddress, context, disableConcurrency } =
     await withContext(withNetwork(getArgs()))
       .string('hookAddress')
       .describe('hookAddress', 'hook address')
-      .number('concurrency')
-      .describe('concurrency', 'maximum number of promises to run concurrently')
+      .boolean('disableConcurrency')
+      .describe(
+        'disableConcurrency',
+        'option to disable parallel iteration over hook domains',
+      )
       .demandOption('hookAddress')
       .demandOption('network').argv;
 
@@ -35,7 +38,11 @@ async function readHook() {
     Role.Deployer,
   );
 
-  const hookReader = new EvmHookReader(multiProvider, network);
+  const hookReader = new EvmHookReader(
+    multiProvider,
+    network,
+    disableConcurrency,
+  );
   const config = await hookReader.deriveHookConfig(hookAddress);
   console.log(EvmHookReader.stringifyConfig(config, 2));
 }
--- a/typescript/infra/scripts/read-hook.ts
+++ b/typescript/infra/scripts/read-hook.ts
@@ -13,18 +13,19 @@ import {
 
 // Examples from <monorepo>/typescript/infra:
 // Fallback routing hook on arbitrum (may take 2-3 minutes):
-//     yarn ts-node scripts/read-hook.ts -e mainnet3 --network arbitrum --hookAddress 0x9e8fFb1c26099e75Dd5D794030e2E9AA51471c25
+//     yarn tsx scripts/read-hook.ts -e mainnet3 --network arbitrum --hookAddress 0x9e8fFb1c26099e75Dd5D794030e2E9AA51471c25
 // IGP hook on inevm (may take 15s):
-//     yarn ts-node scripts/read-hook.ts -e mainnet3 --network inevm --hookAddress 0x19dc38aeae620380430C200a6E990D5Af5480117
+//     yarn tsx scripts/read-hook.ts -e mainnet3 --network inevm --hookAddress 0x19dc38aeae620380430C200a6E990D5Af5480117
 
 async function readHook() {
-  const { environment, network, hookAddress, context } = await withContext(
-    withNetwork(getArgs()),
-  )
-    .string('hookAddress')
-    .describe('hookAddress', 'hook address')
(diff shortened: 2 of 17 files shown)


typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+    };
+  }
+
+  async deriveAggregationConfig(
+    address: Address,
+  ): Promise<AggregationHookConfig> {
+    const hook = StaticAggregationHook__factory.connect(address, this.provider);
+    const hooks = await hook.hooks(ethers.constants.AddressZero);
+    const hookConfigs = await Promise.all(
+      hooks.map(this.deriveHookConfig.bind(this)),
```
yorhodes, line 114: im not totally sure why but iirc `this.fn.bind(this)` is an antipattern
cc @jmrossy

yorhodes: can we just `(hook) => this.deriveHookConfig(hook)`

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+    );
+
+    return {
+      type: HookType.AGGREGATION,
+      hooks: hookConfigs,
+    };
+  }
+
+  async deriveIgpConfig(address: Address): Promise<IgpHookConfig> {
```
yorhodes, line 123: ditto, assert types?

yorhodes: here and elsewhere

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+          const { gasOracle } = await hook.destinationGasConfigs(domainId);
+          const oracle = StorageGasOracle__factory.connect(
+            gasOracle,
+            this.provider,
+          );
+          oracleKey = await oracle.owner();
+        }
+      } catch (error) {
+        // do nothing and continue iterating through known domain IDs
+      }
```
yorhodes, line 164: maybe debug log

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+        // we're going to assume that the owner of the first oracle we encounter
+        // is the owner of all the oracles referenced in this IgpHook
+        if (!oracleKey) {
+          const { gasOracle } = await hook.destinationGasConfigs(domainId);
+          const oracle = StorageGasOracle__factory.connect(
+            gasOracle,
+            this.provider,
+          );
+          oracleKey = await oracle.owner();
+        }
```
yorhodes, line 161: can we assert that they are consistent in an `else` block with `eqAddress`?

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    const owner = await hook.owner();
+    const beneficiary = await hook.beneficiary();
+
+    const overhead: IgpHookConfig['overhead'] = {};
+    const oracleConfig: IgpHookConfig['oracleConfig'] = {};
+
+    let oracleKey: string | undefined;
+
+    for (const domainId of this.multiProvider.getKnownDomainIds()) {
```
yorhodes, line 137: can we capture a ticket for migrating this to enumerables?

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+    };
+  }
+
+  async deriveOpStackConfig(address: Address): Promise<OpStackHookConfig> {
+    const hook = OPStackHook__factory.connect(address, this.provider);
+
+    const messengerContract = await hook.l1Messenger();
+    const destinationDomain = await hook.destinationDomain();
+    const destinationChainName =
+      this.multiProvider.getChainName(destinationDomain);
```
yorhodes, line 202: I worry that this will be fragile
maybe `destinationChain` should be `ChainNameOrId`

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+      try {
+        const domainHook = await hook.hooks(domainId);
+        if (domainHook === ethers.constants.AddressZero) {
+          continue;
+        }
+        domainHooks[chainName] = await this.deriveHookConfig(domainHook);
+      } catch (error) {
+        // if it throws, no entry for that domainId
+        // do nothing and continue iterating through known domain IDs
+      }
```
yorhodes, line 230: ditto, debug log?

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+        return this.deriveOpStackConfig(address);
+      case HookType.ROUTING:
+        return this.deriveDomainRoutingConfig(address);
+      case HookType.FALLBACK_ROUTING:
+        return this.deriveFallbackRoutingConfig(address);
+      case HookType.PAUSABLE:
+        return this.derivePausableConfig(address);
+      default:
+        throw new Error(`Unsupported HookType: ${hookType}`);
+    }
```
yorhodes, line 99: why not map to a derive fn?

typescript/sdk/src/hook/types.ts
yorhodes, line 1: maybe we should remove this layer of indirection and make `HookType` 1:1 with the solidity contract?

typescript/sdk/src/hook/read.ts
paulbalaji: I don't mind it but sure thing, I'll update that here and in the chainmetadata to remove the `.bind`

typescript/sdk/src/hook/types.ts
paulbalaji: I was thinking about this but might be tricky because our HookTypes are string here vs onchain being numbers

we'd still need a mapping somewhere of the enum -> contract either way

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+    return {
+      type: HookType.MERKLE_TREE,
+    };
+  }
+
+  async deriveAggregationConfig(
+    address: Address,
+  ): Promise<AggregationHookConfig> {
+    const hook = StaticAggregationHook__factory.connect(address, this.provider);
+    const hooks = await hook.hooks(ethers.constants.AddressZero);
```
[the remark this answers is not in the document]
paulbalaji: it complains about an "invalid arrayify value" unless I provide `AddressZero`

tried "", "0x0", "0" and some variants of that

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+      case HookType.FALLBACK_ROUTING:
+        return this.deriveFallbackRoutingConfig(address);
+      case HookType.PAUSABLE:
+        return this.derivePausableConfig(address);
+      default:
+        throw new Error(`Unsupported HookType: ${hookType}`);
+    }
+  }
+
+  async deriveMerkleTreeConfig(_: Address): Promise<MerkleTreeHookConfig> {
```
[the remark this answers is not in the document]
paulbalaji: yes - I'll add that to all the funcs

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,284 @@
...
+          value.type === 'BigNumber' &&
+          value.hex
+        ) {
+          return BigNumber.from(value.hex).toString();
+        }
+        // Handle bigint values
+        return typeof value === 'bigint' ? value.toString() : value;
+      },
+      space,
+    );
```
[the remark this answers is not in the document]
paulbalaji: this'll be useful in lots of places so I'll stick it in utils

I'll have a look at pino to see what I can do there

typescript/sdk/src/hook/read.ts
paulbalaji: created https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3560, described it best as I understand but feel free to add more details

typescript/cli/src/config/hooks.ts
```diff
@@ -33,6 +33,7 @@ const ProtocolFeeSchema = z.object({
 });
 
 const MerkleTreeSchema = z.object({
+  owner: z.string(),
```
yorhodes, line 36: why add an owner? I know it might be OwnableUpgradeable but iirc its unused

typescript/utils/src/logging.ts
```diff
@@ -96,3 +97,16 @@ export function createHyperlanePinoLogger(
     },
   });
 }
+
+export function ethersBigNumberReducer(key: string, value: any): any {
+  // Check if the value looks like a serialized BigNumber
```
yorhodes, line 102: I thought this would be before stringifying but ig this works

typescript/sdk/src/hook/types.ts
yorhodes, line 1: still think this should be 1:1 with contracts to reduce confusion

typescript/sdk/src/hook/read.test.ts
yorhodes, line 1: lgtm

typescript/infra/scripts/read-hook.ts
```diff
@@ -0,0 +1,47 @@
...
+  getMultiProviderForRole,
+  withContext,
+  withNetwork,
+} from './agent-utils';
+
+// Examples from <monorepo>/typescript/infra:
+// Fallback routing hook on arbitrum (may take 2-3 minutes):
+//     yarn ts-node scripts/read-hook.ts -e mainnet3 --network arbitrum --hookAddress 0x9e8fFb1c26099e75Dd5D794030e2E9AA51471c25
+// IGP hook on inevm (may take 15s):
+//     yarn ts-node scripts/read-hook.ts -e mainnet3 --network inevm --hookAddress 0x19dc38aeae620380430C200a6E990D5Af5480117
```
yorhodes, line 18: can we parallelize somehow?

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,339 @@
...
+      } catch (error) {
+        // if it throws, no entry for that domainId
+        // do nothing and continue iterating through known domain IDs
+        this.logger.debug(
+          'Domain not configured on Domain Routing Hook',
+          domainId,
+          chainName,
+        );
+      }
+    }
```
yorhodes, line 273: can we parallelize here?

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,339 @@
...
+
+    const owner = await hook.owner();
+    const beneficiary = await hook.beneficiary();
+
+    const overhead: IgpHookConfig['overhead'] = {};
+    const oracleConfig: IgpHookConfig['oracleConfig'] = {};
+
+    let oracleKey: string | undefined;
+
+    for (const domainId of this.multiProvider.getKnownDomainIds()) {
```
yorhodes, line 157: can we parallelize here?

yorhodes (approved):

typescript/sdk/src/hook/types.ts
paulbalaji: would agree but it requires some other changes in SDK/deployers because there's no OP_STACK type on chain but there is in TS

not too much effort imo but would rather do it as a separate PR

typescript/infra/scripts/read-hook.ts
```diff
@@ -0,0 +1,55 @@
...
+    multiProvider,
+    network,
+    disableConcurrency,
+  );
+  const config = await hookReader.deriveHookConfig(hookAddress);
+  console.log(EvmHookReader.stringifyConfig(config, 2));
+}
+
+readHook()
+  .then()
```
jmrossy, line 51: Nit: does this have any effect?

jmrossy (approved):
Looks good, just a few nits. Nice test coverage

typescript/sdk/src/hook/read.ts
```diff
@@ -0,0 +1,372 @@
...
+  IgpHookConfig,
+  MerkleTreeHookConfig,
+  OnchainHookType,
+  OpStackHookConfig,
+  PausableHookConfig,
+  ProtocolFeeHookConfig,
+  RoutingHookConfig,
+} from './types.js';
+
+interface HookReader<_ extends ProtocolType> {
```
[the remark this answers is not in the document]
paulbalaji: good point, I originally had the idea of using it somewhere to make it explicit what protocol type each reader was for. Not needed anymore as the class names make it pretty clear

PR #3573, opened 2024-04-09
add node options to hyperlane CLI script to ignore ExperimentalWarning

nbayindirli:
### Description

* Adds node options to hyperlane CLI script to ignore ExperimentalWarning caused by ESM migration

### Drive-by changes

* None

### Related issues

* n/a

### Backward compatibility

* Yes

### Testing

* Manual

### DIFF
--- a/typescript/cli/package.json
+++ b/typescript/cli/package.json
@@ -30,7 +30,7 @@
     "typescript": "^5.1.6"
   },
   "scripts": {
-    "hyperlane": "node ./dist/cli.js",
+    "hyperlane": "NODE_OPTIONS='--experimental-loader ts-node/esm/transpile-only --no-warnings=ExperimentalWarning' node ./dist/cli.js",
     "build": "yarn version:update && tsc",
     "dev": "yarn version:update && tsc --watch",
     "clean": "rm -rf ./dist",


typescript/cli/package.json
```diff
@@ -30,7 +30,7 @@
     "typescript": "^5.1.6"
   },
   "scripts": {
-    "hyperlane": "node ./dist/cli.js",
+    "hyperlane": "NODE_OPTIONS='--experimental-loader ts-node/esm/transpile-only --no-warnings=ExperimentalWarning' node ./dist/cli.js",
```
jmrossy, line 33: I think only the `--no-warnings=ExperimentalWarning` is relevant here but anyway we won't need this once I'm finished with https://github.com/hyperlane-xyz/issues/issues/917

jmrossy (commented):
Thanks for looking into it but I think we won't need this soon!

typescript/cli/package.json
nbayindirli: Gotcha– this makes total sense. Will close out PR in favor of https://github.com/hyperlane-xyz/issues/issues/917 :)

PR #3583, opened 2024-04-10
Update key funder deployments w/ RpcConsensusType.Fallback

tkporter:
### Description

Making testnet4 and mainnet3 key funder use fallback for consistency & robustness

### Drive-by changes

Updated the deployed key funder images


daniel-savu (approved):
lgtm, we should also make sure we do have more than one entry in the RPCURLs secret for the chains we're running this on

PR #3586, opened 2024-04-11
feat: add dry-run support for warp deploy + gas stats

nbayindirli:
### Description

* New feature to allow users to dry-run the warp deploy script against a forked (base/origin) network of their choice
* To run: `yarn build && yarn hyperlane deploy warp --dry-run` || `yarn build && yarn hyperlane deploy warp -d`
* Externally enables: `hyperlane deploy warp --dry-run` || `hyperlane deploy warp -d`
* Also adds gas usage util for both warp & core deployments in all contexts, e.g.
  * When running a vanilla core deploy, for example, between alfajores and fuji, you will now see:
```
⛽️ Gas Usage Statistics
        - Gas required for core deploy on alfajores: 0.0058686745 CELO
        - Gas required for core deploy on fuji: 0.0239308515 AVAX
```

### Drive-by changes

* None

### Related issues

* Fixes https://github.com/hyperlane-xyz/issues/issues/819

### Backward compatibility

* Yes

### Testing

* Note: `hl` == `yarn build && yarn hyperlane`. The below tests are only a sample and are not inclusive.

#### Manual Testing
  * With `anvil` NOT running in separate instance:
    * `hl deploy warp -d`
      * Throws:
```
Error: No active anvil node detected.
        Please run `anvil` in a separate instance.
```
  * With `anvil` running in separate instance:
     * `hl deploy warp -d -k c0052e22df5d1f4ae7c51e254Xx00Xx0eb833453eaed6301xXxxx8a30d92d10a` (any private key)
       * Throws `Error: Invalid address length. Please ensure you are passing an address and not a private key.`
     * `hl deploy warp -d -k 0x16F4898F47c085C41d7Cc6b1dc0xX0xXX017dcBb` (any public address)
       * Output:
```
🔎 Verifying anvil node is running...
✅ Successfully verified anvil node is running
Using warp route deployment config at ./configs/warp-route-deployment.yaml
No chain config file provided
? Do you want to use some core deployment address artifacts? This is required for PI chains (non-core chains). no
Forking alfajores for dry-run...
✅ Successfully forked alfajores for dry-run
Impersonating account (0x16F4898F47c085C41d7Cc6b1dc72B91EA617dcBb)...
✅ Successfully impersonated account (0x16F4898F47c085C41d7Cc6b1dc72B91EA617dcBb)
...
✅ Hyp token deployments complete
Writing deployment artifacts
Deployment is complete!
Contract address artifacts are in artifacts/dry-run_warp-route-deployment-2024-04-16-12-27-02.json
Warp config is in artifacts/dry-run_warp-config-2024-04-16-12-27-02.json
⛽️ Gas Usage Statistics
        - Gas required for warp dry-run on alfajores: 0.013310162514578124 CELO
Resetting forked network...
✅ Successfully reset forked network
✅ Warp dry-run completed successfully
```

#### CI Testing
  * Successful CI-backed integration/regression testing via `ci-test.sh`


typescript/cli/ci-test.sh
```diff
@@ -150,6 +154,32 @@ run_hyperlane_deploy_core_dry_run() {
...
+    GAS_PRICE=$(cast gas-price --rpc-url http://127.0.0.1:${CHAIN1_PORT})
+    CORE_MIN_GAS=$(bc <<< "($BEFORE_CORE_DRY_RUN - $AFTER_CORE_DRY_RUN) / $GAS_PRICE")
+    echo "Gas used: $CORE_MIN_GAS"
+
+    CORE_ARTIFACTS_PATH=`find /tmp/core-deployment* -type f -exec ls -t1 {} + | head -1`
+    echo "Core artifacts:"
+    echo $CORE_ARTIFACTS_PATH
+    cat $CORE_ARTIFACTS_PATH
+
+    AGENT_CONFIG_FILENAME=`ls -t1 /tmp | grep agent-config | head -1`
```
jmrossy, line 180: Did you mean to include this stuff? Seems core specific

typescript/cli/ci-test.sh
```diff
@@ -150,6 +154,32 @@ run_hyperlane_deploy_core_dry_run() {
     AGENT_CONFIG_FILENAME=`ls -t1 /tmp | grep agent-config | head -1`
 }
 
+run_hyperlane_deploy_warp_dry_run() {
+    BEFORE_CORE_DRY_RUN=$(cast balance $DEPLOYER --rpc-url http://127.0.0.1:${CHAIN1_PORT});
```
jmrossy, line 158: `BEFORE_CORE_DRY_RUN` -> `BEFORE_DRY_RUN` or `BEFORE_WARP_DRY_RUN`

typescript/cli/src/deploy/warp.ts
```diff
@@ -80,6 +81,25 @@ export async function runWarpRouteDeploy({
...
+    : await getContext({
+        chainConfigPath,
+        coreConfig: { coreArtifactsPath },
+        keyConfig: { key },
+        skipConfirmation,
+      });
+
+  const multiProvider = context.multiProvider;
+  const signer = context.signer;
+  const coreArtifacts = context.coreArtifacts;
```
jmrossy, line 101: Nit: could be `const {mp, signer, core} = context`

typescript/cli/src/utils/files.ts
```diff
@@ -163,6 +163,23 @@ export function prepNewArtifactsFiles(
...
+/**
+ * Retrieves artifacts file metadata for the current command.
+ * @param dryRun whether or not the current command is being dry-run
+ * @returns the artifacts files
+ */
+export function getArtifactsFiles(
+  deploymentFile: ArtifactsFile,
+  configFile: ArtifactsFile,
+  dryRun: boolean,
+): Array<ArtifactsFile> {
```
jmrossy, line 175: Not a big deal but here's what I'd suggest for this if it's going to be a generalized util:
```ts
export function getArtifactsFiles(defaultFiles: ArtifactsFile[], dryRun=false)
// OR
export function getArtifactsFiles(dryRun: boolean, ...defaultFiles: ArtifactsFile[])
```

jmrossy (commented):
Mostly looks good! Just a few small things to fix

typescript/cli/src/deploy/warp.ts
nbayindirli: Ah yep– I was initially re-assigning to one of these fields but no longer relevant. Reverted.

typescript/cli/src/utils/files.ts
nbayindirli: Really love this– refactored for the first form. Did not feel quite right first go but ended up moving on; very happy to make this change

typescript/cli/ci-test.sh
nbayindirli: Ah yes, updated this since first commits. Was just a copy oversight from my other CI test– also ended up change artifact directories for both

typescript/cli/ci-test.sh
nbayindirli: See below

typescript/cli/src/deploy/dry-run.ts
```diff
@@ -34,7 +42,7 @@ export async function verifyAnvil() {
 \tPlease run \`anvil\` in a separate instance.`);
   }
 
-  logGreen('Successfully verified anvil node is running ✅');
+  logGreen('✅ Successfully verified anvil node is running');
```
jmrossy, line 45: 👍

typescript/cli/src/utils/fork.ts
```diff
@@ -128,3 +129,18 @@ export const getLocalProvider = (
 
   return new providers.JsonRpcProvider(url);
 };
+
+/**
+ * Retrieves the balance of an account.
+ * @param address the address to retrieve the balance of
+ * @returns the balance of the account
+ */
+export const getAccountBalance = async (address: Address) => {
```
jmrossy, line 138: If it's anvil specific let's rename it for clarity please

jmrossy (approved):

typescript/cli/src/utils/fork.ts
nbayindirli: This is a +1 for class encapsulation– what are you thoughts on getting that kicked off by wrapping this fork util in a class?

Something like `class ForkUtil` (may have constructor params– perhaps port #s for example, since we will expand to multi-network forking)

instantiated like `const forkUtil = new ForkUtil();`

used like `const impersonatedAccountBalance = forkUtil.getAccountBalance()`

jmrossy: I'm weakly against Class encapsulation in cases like this because they're so redundant with the encapsulation that JS modules already provide. Every file gets private state (via any top-level vars) and an explicit interface (via the file's imports/exports), so wrapping things in a class Java-style doesn't actually get you much. 

That said, classes makes sense where you intend to actually use OO patterns such as inheritance, which is why we use them in the SDK's deployers/checkers/etc. 

So yeah I wouldn't spend any time wrapping up all our various util fns in classes but if you think we'll want a strict hierarchy then I'm open to it in this case.

nbayindirli: Makes complete sense– generally speaking, I bring up class encapsulation as way for us to run context-driven CLI commands in the future (deduping code between commands, like warp and core), which would step closer to OO like the sdk

I am weakly for that beginning here, and as you mentioned it really is not necessary for context-driven commands at any point, so I will leave fork.ts as is and refactor to getAnvilAccountBalance.

Would love to continue chatting offline about the command flow at a high-level 👍

typescript/cli/src/deploy/dry-run.ts
```diff
@@ -48,3 +56,21 @@ export function evaluateIfDryRunFailure(error: any, dryRun: boolean) {
...
+export async function completeDryRun(
+  command: COMMAND,
+  impersonatedAccount: Address,
+  accountBalanceInitial: number,
+) {
+  const accountBalanceDelta =
+    accountBalanceInitial - (await getAnvilAccountBalance(impersonatedAccount));
+  logPink(
+    `⛽️ Approximate gas usage during ${command} dry-run: ${accountBalanceDelta} wei`,
+  );
```
yorhodes, line 69: can we pull this out of dry run branch?

typescript/cli/src/deploy/warp.ts
```diff
@@ -253,16 +279,29 @@ async function executeDeploy(params: DeployParams) {
...
   const deployer = isNft
     ? new HypERC721Deployer(multiProvider)
     : new HypERC20Deployer(multiProvider);
 
-  const deployedContracts = await deployer.deploy(configMap);
+  const baseOnlyConfigMap = { [params.origin]: configMap[params.origin] };
+
+  const deployedContracts = params.dryRun
+    ? await deployer.deploy(baseOnlyConfigMap)
+    : await deployer.deploy(configMap);
```
yorhodes, line 304:
```suggestion
  const config = params.dryRun ? { [params.origin]: configMap[params.origin] } : configMap;

  const deployedContracts = await deployer.deploy(config);
```

typescript/cli/src/utils/files.ts
```diff
@@ -163,6 +163,22 @@ export function prepNewArtifactsFiles(
...
+ */
+export function getArtifactsFiles(
+  defaultFiles: ArtifactsFile[],
+  dryRun = false,
+): Array<ArtifactsFile> {
+  if (dryRun)
+    defaultFiles.forEach((defaultFile) => {
+      defaultFile.filename = `dry-run_${defaultFile.filename}`;
+    });
+  return defaultFiles;
```
yorhodes, line 179: nit: prefer objMap to mutation

typescript/cli/src/utils/fork.ts
```diff
@@ -16,6 +16,7 @@ export enum ANVIL_RPC_METHODS {
   IMPERSONATE_ACCOUNT = 'anvil_impersonateAccount',
   STOP_IMPERSONATING_ACCOUNT = 'anvil_stopImpersonatingAccount',
   NODE_INFO = 'anvil_nodeInfo',
+  GET_BALANCE = 'eth_getBalance',
```
yorhodes, line 19: would prefer to keep this in vanilla ethers provider usage

yorhodes (commented):
looks good, I think we can generalize the fund usage to non dryrun path

typescript/cli/examples/fork/anvil-chains.yaml
```diff
@@ -13,6 +13,12 @@ anvil:
     name: Ether
     symbol: ETH
     decimals: 18
+alfajores:
```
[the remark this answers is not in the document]
nbayindirli: I want to use alfajores here since we need an account with a balance and 0xfaD1C94469700833717Fa8a3017278BC1cA8031C has plenty. Using an address funded on ethereum mainnet doesn't seem like a great choice. It's forked but the potential to move or use assets on ethL1 is high

typescript/cli/ci-test.sh
```diff
@@ -150,6 +154,29 @@ run_hyperlane_deploy_core_dry_run() {
     AGENT_CONFIG_FILENAME=`ls -t1 /tmp | grep agent-config | head -1`
 }
 
+run_hyperlane_deploy_warp_dry_run() {
+    BEFORE_WARP_DRY_RUN=$(cast balance $DEPLOYER --rpc-url http://127.0.0.1:${CHAIN1_PORT});
+
+    echo -e "\nDry-running warp route deployments to Alfajores"
```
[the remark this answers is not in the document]
nbayindirli: This is explicit since we pass ` --targets alfajores` into the command below below

typescript/cli/src/commands/deploy.ts
```diff
@@ -26,11 +26,18 @@ import {
   warpConfigCommandOption,
 } from './options.js';
 
+export enum COMMAND {
```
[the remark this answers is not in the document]
nbayindirli: Definitely was not intentional. Reverted.

typescript/cli/src/deploy/core.ts
```diff
@@ -146,9 +148,13 @@ export async function runCoreDeploy({
     ...deploymentParams,
     minGas: MINIMUM_CORE_DEPLOY_GAS,
   });
+
+  let accountBalanceInitial: number = 0;
+  if (dryRun) accountBalanceInitial = await getAnvilAccountBalance(key);
```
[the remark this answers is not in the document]
nbayindirli: Extrapolated for multi-network & non-dryrun support

typescript/cli/src/deploy/warp.ts
nbayindirli: Much better- thanks

typescript/cli/src/utils/files.ts
nbayindirli: Now using `map` instead

typescript/cli/src/utils/files.ts
```diff
@@ -163,6 +163,23 @@ export function prepNewArtifactsFiles(
...
+ * @returns the artifacts files
+ */
+export function getArtifactsFiles(
+  deploymentFile: ArtifactsFile,
+  configFile: ArtifactsFile,
+  dryRun: boolean,
+): Array<ArtifactsFile> {
+  if (dryRun) {
+    deploymentFile.filename = `dry-run_${deploymentFile.filename}`;
+    configFile.filename = `dry-run_${configFile.filename}`;
```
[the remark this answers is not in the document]
nbayindirli: If we use the same artifact filenames for dry-runs, developers may think those artifacts exist on testnet or mainnet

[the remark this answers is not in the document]
yorhodes: gotcha
I think I interpreted `configFile` as an input to deploy, rather than as an output

typescript/cli/src/deploy/utils.ts
```diff
@@ -96,3 +100,54 @@ export function isISMConfig(
...
+      : multiProvider.getProvider(chain);
+    const currentBalance = await provider.getBalance(userAddress);
+    const balanceDelta = initialBalances[chain].sub(currentBalance);
+    if (dryRun && balanceDelta.lt(0)) break;
+    logPink(
+      `\t- Gas used during ${command} ${
+        dryRun ? 'dry-run' : 'deploy'
+      } on ${chain}: ${balanceDelta} wei or ${ethers.utils.formatEther(
+        balanceDelta,
+      )} ether`,
```
yorhodes, line 144: nit
```suggestion
      `\t- Balance used during ${command} ${
        dryRun ? 'dry-run' : 'deploy'
      } on ${chain}: ${ethers.utils.formatEther(
        balanceDelta,
      )} ether`,
```
idk if we need both formats
also would be nice to use the `chainMetadata.nativeToken.symbol`/`decimals` instead of `ether`

yorhodes (approved):

typescript/cli/src/deploy/utils.ts
nbayindirli: We're using `ethers` to calculate balance and all EVM-compatible chains have a native decimals of 18 and are denominated in wei, so this doesn't seem pertinent

Re symbol I do agree, so will update this

Issue #3452, opened 2024-03-20
Implement a CLI command to process a message


tkporter:
Following the metadata building work done in https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3448, we'll be able to build metadata to deliver messages. We should then:

- [ ] Add a CLI command to process a message
  - Note we may require the tx of the message to be passed in so that we can get the message contents and the merkle tree insertion index

PR #3604, opened 2024-04-14
Implement NULL metadata relaying in the SDK/CLI

yorhodes:
### Description

- Add NULL self relay to core app
- Add self relay to send message command

### Drive-by changes

- Migrate to assert util in ISM Factory
- Default to `HYP_KEY` in shared key option

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3608
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3609
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3452

### Backward compatibility

Yes

### Testing

Manual

https://github.com/hyperlane-xyz/hyperlane-monorepo/assets/3020995/e44683c3-bca4-41d5-a4c7-45718f05edf4

https://github.com/hyperlane-xyz/hyperlane-monorepo/assets/3020995/50565be1-d266-4d2d-ae27-5fbb5539b93f


typescript/cli/src/send/message.ts
```diff
@@ -146,6 +146,9 @@ async function executeDelivery({
     logBlue(`Sent message from ${origin} to ${recipient} on ${destination}.`);
     logBlue(`Message ID: ${message.id}`);
     log(`Message: ${JSON.stringify(message)}`);
+
+    await core.relayMessage(message);
```
jmrossy, line 150: Should we check if the destination is a core chain and let regular relay handle those?

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -85,6 +89,50 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+    message: DispatchedMessage,
+  ): Promise<ethers.ContractReceipt> {
+    const destinationChain = this.getDestination(message);
+    const mailbox = this.contractsMap[destinationChain].mailbox;
+    const ismConfig = await this.getRecipientIsmConfig(message);
+    if (typeof ismConfig === 'string') {
+      throw new Error(
+        `No ISM config found for recipient ${message.parsed.recipient}`,
+      );
+    }
```
jmrossy, line 117: Does a string config mean it's an address? The error message is confusing me a bit here

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -85,6 +89,50 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+    // TODO: implement metadata builders
+    let metadata: string;
+    const moduleType = ismTypeToModuleType(ismConfig.type);
+    switch (moduleType) {
+      case ModuleType.NULL:
+        metadata = '0x';
+        break;
+      default:
+        throw new Error(`Unsupported module type ${moduleType}`);
+    }
```
jmrossy, line 128: In my ongoing war against switch statements, the worst control flow tool ever invented, I'll complain one more time:
is this really better than
```ts
if ( moduleType === ModuelType.NULL) metadata = '0x';
else throw new Error(`Unsupported module type ${moduleType}`);
```

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -176,6 +177,21 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
...
         });
         break;
+      case IsmType.TRUSTED_RELAYER:
+        if (!this.deployer) {
+          throw new Error(`HyperlaneDeployer must be set to deploy ${ismType}`);
+        } else if (!mailbox) {
+          throw new Error(
+            `Mailbox address is required for deploying ${ismType}`,
+          );
+        }
```
jmrossy, line 187: NIt: Could be `assert(this.deployer, `HyperlaneDeployer must be set to deploy ${ismType})`.
See assert in our utils package

jmrossy (approved):
Nice work here! Impressed with how fast you put this together. 
I will say though, I think our M.O. should be for all new SDK features to be multi-protocol first but I'll leave that decision to you

typescript/sdk/src/core/HyperlaneCore.ts
nbayindirli: +1, so long as `ModuleType.NULL` comes first in the if statement (purely cosmetic)

LOGAF super low on switch statements

typescript/cli/src/config/ism.ts
```diff
@@ -159,36 +165,34 @@ export async function createIsmConfig(
...
       },
       {
         value: IsmType.FALLBACK_ROUTING,
-        name: IsmType.FALLBACK_ROUTING,
         description:
           "You can specify ISM type for specific chains you like and fallback to mailbox's default ISM for other chains via DefaultFallbackRoutingISM",
       },
       {
         value: IsmType.AGGREGATION,
-        name: IsmType.AGGREGATION,
```
paulbalaji, line 185: thanks for the drive-by

typescript/cli/src/send/message.ts
paulbalaji: imo self-relay is analogous to `relay --force` so it makes sense to bypass regular relaying

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -85,6 +89,50 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+    return destinationMailbox.mailbox.recipientIsm(ethAddress);
+  }
+
+  async getRecipientIsmConfig(
+    message: DispatchedMessage,
+  ): Promise<WithAddress<IsmConfig>> {
+    const destinationChain = this.getDestination(message);
+    const ismReader = new EvmIsmReader(this.multiProvider, destinationChain);
+    const address = await this.getRecipientIsmAddress(message);
+    return ismReader.deriveIsmConfig(address);
```
paulbalaji, line 104: image

typescript/sdk/src/core/HyperlaneCore.ts
paulbalaji: does `getRecipientIsmConfig` actually return a type `string`? I would've thought that with the type `WithAddress<IsmConfig>`, we'd get back an object instead of just the string i.e.
```
{
    string: '0xsdsd',
    address: '0xsdsd',
}
```

curious if you hit a problem when running that led to this check

typescript/sdk/src/core/HyperlaneCore.ts
paulbalaji: rossy v switch statements
image

solidity/contracts/isms/TrustedRelayerIsm.sol
```diff
@@ -0,0 +1,27 @@
...
+import {IInterchainSecurityModule} from "../interfaces/IInterchainSecurityModule.sol";
+import {Message} from "../libs/Message.sol";
+import {Mailbox} from "../Mailbox.sol";
+
+contract TrustedRelayerIsm is IInterchainSecurityModule {
+    using Message for bytes;
+
+    uint8 public moduleType = uint8(Types.NULL);
+    Mailbox public immutable mailbox;
+    address public immutable trustedRelayer;
```
paulbalaji, line 14: update `deriveNullConfig` in the ISM reader pls, if contract has `trustedRelayer` property then you can infer it's a `TrustedRelayerIsm`

typescript/sdk/src/core/HyperlaneCore.ts
yorhodes: changed your return type to `Exclude`

typescript/sdk/src/core/HyperlaneCore.ts
yorhodes: I got rid of the mutation but keeping the switch case after chatting offline :P

typescript/cli/src/send/message.ts
yorhodes: added an explicit --selfrelay flag

typescript/sdk/src/ism/read.ts
```diff
@@ -184,9 +192,27 @@ export class EvmIsmReader implements IsmReader {
 
   async deriveNullConfig(
     address: Address,
-  ): Promise<
-    WithAddress<PausableIsmConfig | TestIsmConfig | OpStackIsmConfig>
-  > {
+  ): Promise<WithAddress<NullIsmConfig>> {
+    // if it has trustedRelayer() property --> TrustedRelayer ISM
```
paulbalaji, line 196: thanks for adding! might need to also mock `TrustedRelayerIsm__factory` in the associated test to make CI happy

paulbalaji (approved):
super stoked with how quick this came together

typescript/cli/src/commands/send.ts
```diff
@@ -85,6 +93,7 @@ const messageCommand: CommandModule = {
       messageBody: ethers.utils.hexlify(ethers.utils.toUtf8Bytes(messageBody)),
       timeoutSec,
       skipWaitForDelivery,
+      selfRelay,
```
nbayindirli, line 96: Strongly in favor of this being '`--self-relay`' via CLI

nbayindirli (changes requested):

typescript/cli/src/commands/send.ts
```diff
@@ -32,6 +32,7 @@ export const selfrelay: Options = {
   type: 'boolean',
   description: 'Relay message on destination chain',
   default: false,
+  alias: ['s', 'sr'],
```
nbayindirli, line 35: Thank you!

nbayindirli (approved):
🚢

Issue #3582, opened 2024-04-10
Replace Pino logger with simpler alternative


jmrossy:
## Overview

In https://github.com/hyperlane-xyz/issues/issues/918 the monorepo's logging was overhauled to replace Debug with Pino.
A few alternative libs were considered but Pino was chosen. It turns out that was a bad choice. Pino is difficult to configure, lacks an easy way to format logs in production (e.g. [pretty print not recommended](https://github.com/pinojs/pino-pretty?tab=readme-ov-file#programmatic-integration)), and has an overly complicated [bundling requirements](https://getpino.io/#/docs/bundling) that worsens DX for SDK users. These bundling quirks also prevented us from using [Pino in Docusaurus](https://github.com/hyperlane-xyz/v3-docs/issues/80#issuecomment-2048165091).

We should replace Pino with something simpler.

## Related issues

https://github.com/hyperlane-xyz/issues/issues/918
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3481
https://github.com/hyperlane-xyz/v3-docs/issues/80

jmrossy:
Another one, from @tkporter:

```
$ LOG_LEVEL=DEBUG ./fork.sh mainnet3 warp inevm                               1 ↵
+ DEPLOYER=0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba
++ cast balance 0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba --rpc-url http://localhost:8545
+ BEFORE=5872881681565083840
+ yarn tsx ./scripts/deploy.ts -e mainnet3 -f inevm -m warp
TypeError: Cannot read properties of undefined (reading 'Symbol(pino.msgPrefix)')
    at LOG (/Users/trevor/abacus-monorepo/node_modules/pino/lib/tools.js:68:22)
    at logMethod (/Users/trevor/abacus-monorepo/typescript/utils/src/logging.ts:95:23)
    at hookWrappedLog (/Users/trevor/abacus-monorepo/node_modules/pino/lib/tools.js:37:10)
    at getMultiProviderForRole (/Users/trevor/abacus-monorepo/typescript/infra/scripts/agent-utils.ts:289:3)
    at Object.getMultiProvider (/Users/trevor/abacus-monorepo/typescript/infra/config/environments/mainnet3/index.ts:36:12)
[...]

PR #3606, opened 2024-04-15
Ancient8 new ism, misc infra fixes

tkporter:
### Description

- Deploys ancient8 with a new ISM where we are not quorum
- Pulls out warp config to live in a single file / function so that deploy.ts and check-deploy.ts are in sync. Otherwise fork.sh is broken for warp routes
- Shuffles around the aw multisig serialization stuff - now they're segmented by environment and context and are written to `config/{env}/aw-validators/{context}.json`. This was in response to some confusing merging logic, see https://discord.com/channels/935678348330434570/1228377321082327141/1229464197297868910

### Drive-by changes

Needed to specify pino to work in pretty mode for warp monitoring :/ this is necessary because pino doesn't work otherwise. After #3582 we should change this


typescript/infra/helm/warp-routes/templates/_helpers.tpl
```diff
@@ -58,6 +58,9 @@ The warp-routes container
 - name: warp-routes
   image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
   imagePullPolicy: IfNotPresent
+  env:
```
tkporter, line 61: This is a temporary measure because otherwise pino doesn't work :( once we move away from pino in #3582 we should revisit this and move back to json structured logging

typescript/infra/scripts/warp-routes/helm.ts
```diff
@@ -27,12 +29,16 @@ function getHelmReleaseName(route: string): string {
 }
 
 function getWarpRoutesHelmValues(configFilePath: string) {
+  // The path should be relative to the monorepo root
```
tkporter, line 32: Not a fan of my hack here so if anyone has suggestions on how to convert the `configFilePath` that's passed in as a CLI arg to be a path relative to the monorepo root, pls do share

yorhodes (approved):
thanks for cleaning this up
we should expect warp fork tests to be working now right?

tkporter:
yeah there was a circular dep that was causing them to fail

PR #3610, opened 2024-04-15
Indexing Type Wrapper

daniel-savu:
### Description

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3544.

Wraps all types to be indexed in an `Indexed<T>` struct in the relayer. This allows for using sequece-aware cursors for data that doesn't contain a sequence but can have it inferred from metadata (as is the case for the IGP on sealevel).

`Indexed<T>` stores the sequence as an `Option`, so the db can have additional sequence data stored in it when the sequence is defined - this is how `process_indexed_gas_payment` is implemented now.

In addition, the IGP payments and Deliveries are configured to be indexed with a `SequenceAware` cursor on sealevel (see `cursors/mod.rs`).

### Drive-by changes

Implements the custom `Encode` and `Decode` traits for `IndexingDecorator`, which can be generalized to `Option<T>` where `T: Encode + Decode`. 

### Testing

E2E


rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -107,6 +108,41 @@ impl HyperlaneRocksDB {
         }
     }
```
tkporter, line 110: I think we need to update `impl HyperlaneSequenceAwareIndexerStoreReader<InterchainGasPayment> for HyperlaneRocksDB` to fetch stored sequences from the db https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dan/indexing-decorator/rust/hyperlane-base/src/db/rocks/hyperlane_db.rs#L367-L377

rust/hyperlane-core/src/types/indexing.rs
```diff
@@ -0,0 +1,62 @@
+use derive_new::new;
+
+use crate::{HyperlaneMessage, MerkleTreeInsertion, Sequenced};
+
+#[derive(Debug, Clone, Default, Copy, PartialEq, Eq, Hash, new)]
+/// Additional indexing data associated to a type
+pub struct IndexingDecorator {
+    // this field could be made optional if it doesn't apply to all instances
```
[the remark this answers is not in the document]
daniel-savu: oops I already made it optional so I should remove the comment. It's already optional for the IGP, bc there are chains where it doesn't have a sequence

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -107,6 +108,41 @@ impl HyperlaneRocksDB {
...
+        let payment = *(indexed_payment.inner());
+
+        let gas_processing_successful = self.process_gas_payment(payment, log_meta)?;
+        if indexed_payment.sequence().is_none() {
+            // only store the payment and return early if there's no sequence
+            return Ok(gas_processing_successful);
+        }
+        // otherwise store the indexing decorator as well
+        let gas_payment_key = payment.into();
+        if let Ok(Some(_)) = self.retrieve_indexing_decorator_by_gas_payment_key(&gas_payment_key) {
```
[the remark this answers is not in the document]
daniel-savu: ah is that why the payment is checked for with `payment_meta` rather than the payment key in `process_gas_payment`?

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -107,6 +108,41 @@ impl HyperlaneRocksDB {
...
+            trace!(
+                ?indexed_payment,
+                ?log_meta,
+                "Attempted to process an already-processed indexed gas payment"
+            );
+            // Return false to indicate the gas payment was already processed
+            return Ok(false);
+        }
+
+        self.store_indexing_decorator_by_gas_payment_key(
```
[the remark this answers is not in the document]
daniel-savu: maybe storing by gas payment key isn't the best approach, but the point is to establish a link between the sequence number and the IGP payment, which is required for implementing the `get_by_sequence` you flagged here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3610/files/cf4751210f2fc6871040d46dde0576b735f11509#r1568826216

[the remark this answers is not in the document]
tkporter: ah i see, that makes sense

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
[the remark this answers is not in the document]
tkporter: yeah exactly

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -22,7 +22,7 @@ const MESSAGE_ID: &str = "message_id_";
 const MESSAGE_DISPATCHED_BLOCK_NUMBER: &str = "message_dispatched_block_number_";
 const MESSAGE: &str = "message_";
 const NONCE_PROCESSED: &str = "nonce_processed_";
-const INDEXING_DECORATOR_BY_IGP_KEY: &str = "indexing_decorator_by_igp_key_";
+const GAS_PAYMENT_BY_SEQUENCE: &str = "indexing_decorator_by_igp_key_";
```
tkporter, line 25: wondering if it's worth changing the value too for consistency

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -366,35 +363,61 @@ impl HyperlaneSequenceAwareIndexerStoreReader<MerkleTreeInsertion> for Hyperlane
...
 
     /// Gets the block number at which the log occurred.
-    async fn retrieve_log_block_number_by_sequence(&self, _sequence: u32) -> Result<Option<u64>> {
-        Ok(None)
+    async fn retrieve_log_block_number_by_sequence(&self, sequence: u32) -> Result<Option<u64>> {
+        Ok(self.retrieve_gas_payment_block_by_sequence(&sequence)?)
     }
 }
 
 /// Note that for legacy reasons this watermark may be shared across multiple cursors, some of which may not have anything to do with gas payments
```
tkporter, line 376: I think you may be able to remove this comment now

rust/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -366,35 +363,61 @@ impl HyperlaneSequenceAwareIndexerStoreReader<MerkleTreeInsertion> for Hyperlane
...
     }
+
     /// Stores the block number high watermark
     async fn store_high_watermark(&self, block_number: u32) -> Result<()> {
         let result = self.store_encodable("", LATEST_INDEXED_GAS_PAYMENT_BLOCK, &block_number)?;
         Ok(result)
     }
 }
 
+// Keep this implementation for type compatibility with the `contract_syncs` sync builder
```
tkporter, line 393: this is a bit unfortunate

rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
```diff
@@ -317,6 +317,7 @@ impl Indexer<MerkleTreeInsertion> for CosmosMerkleTreeHookIndexer {
                 }
             })
             .flatten()
+            .map(|(log, meta)| (Indexed::new(log).with_sequence(log.index() as _), meta))
```
tkporter, line 320: nit: iiuc for consistency you can use the `impl From<MerkleTreeInsertion> for Indexed<MerkleTreeInsertion>` like we do elsewhere

rust/hyperlane-base/src/contract_sync/cursors/rate_limited.rs
```diff
@@ -146,10 +78,10 @@ pub enum SyncDirection {
 /// queried is and also handling rate limiting. Rate limiting is automatically
 /// performed by `next_action`.
 pub(crate) struct RateLimitedContractSyncCursor<T> {
+    // TODO: Sequence information is no longer required, so just use Indexer
```
tkporter, line 81: iiuc we can make this move now to `indexer: Arc<dyn Indexer<T>>,`? and to get the tip everywhere we use `let tip = this.indexer.get_finalized_block_number().await?;`

tkporter (approved):
some small suggestions, lgtm!

PR #3612, opened 2024-04-15
fix: CI job using ancient8 instead of inevm

aroralanuk:
### Description

- consistent warp route configs between ancient8 <> ethereum for deploy and check-deploy script which is now used in the CI job.

### Drive-by changes

None

### Related issues

- https://discord.com/channels/935678348330434570/1225043256782622730/1229223336366571612 for fixing https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/8696381654/job/23849604350

### Backward compatibility

Yes

### Testing

CI


yorhodes (commented):
maybe we should wait for https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3606

Issue #2810, opened 2023-10-13
Add remote registry functionality


tkporter:
- [ ] chain metadata can be fetched from remote registry
- [ ] remote registry artifacts like mailbox addresses, validator announce, ISMs, hooks, etc can be referenced by their ID

PR #3615, opened 2024-04-16
Migrate SDK consts to Registry and use new registry utilities

jmrossy:
### Description

- Remove most consts from SDK
- Refactor tests that relied on chainMetadata
- Remove JSON imports/exports in the SDK
- Update the HelloWorld package to read metadata from the registry lib
- Update the Infra package to read/write to a LocalRegistry
- Update the CLI to read and write from registries

### Drive-by changes

- Fix typo in `EvmHypCollateralVault` token standard 
- Use consistent, newest yaml lib version

### Related issues

Fixes https://github.com/hyperlane-xyz/issues/issues/917
Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2810

### Backward compatibility

No: 
- Several SDK exports like `chainMetadata` have been removed
- Common args for the CLI like `--chain` have been replaced with `--registry` and `--overrides`


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           node-version: 18.x
 
       - name: Install Dependencies
-        run: yarn
+        run: yarn install --no-immutable
```
jmrossy, line 34: Just happened to notice this. Yarn does a decent job recognizing it's in CI and then enabling this flag by default but better safe than sorry when it comes to releases

typescript/sdk/src/consts/multisigIsmVerifyCosts.ts
```diff
@@ -0,0 +1,210 @@
+/* eslint-disable sort-keys */
```
jmrossy, line 1: I just converted the file above to ts.
We should avoid imports of files other than js/ts in the SDK because it weakens our compatibility with other JS tooling. For example, by now removing all the json imports, `ts-node`won't complain about the `assert {type json` warning anymore

typescript/sdk/src/warp/test-warp-core-config.yaml
```diff
@@ -0,0 +1,80 @@
+# An example Warp Core config
```
jmrossy, line 1: Renamed the file above and tweaked it to refer only to test chains

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -619,15 +621,17 @@ export class WarpCore {
...
+    // Finally, if the IGP fee token differs from the transfer token,
+    // ensure there's sufficient balance for the IGP fee
     const igpQuote = feeEstimate.interchainQuote;
-    const igpTokenBalance = await igpQuote.token.getBalance(
-      this.multiProvider,
-      sender,
-    );
-    if (igpTokenBalance.amount < igpQuote.amount) {
-      return { amount: `Insufficient ${igpQuote.token.symbol} for gas` };
+    if (!token.isFungibleWith(igpQuote.token)) {
```
jmrossy, line 627: A minor change which 1) avoids an unncessary balance fetch in some cases and 2) helped me make the tests happy again without a bunch of extra stubbing

typescript/sdk/src/consts/multisigIsmVerifyCosts.ts
nbayindirli: Love this tidbit. TFTI(nfo)

typescript/sdk/src/gas/HyperlaneIgpChecker.ts
```diff
@@ -131,9 +128,7 @@ export class HyperlaneIgpChecker extends HyperlaneAppChecker<
       Object.keys(this.configMap[local].oracleConfig ?? {}),
     );
     for (const remote of remotes) {
-      const remoteId =
-        chainMetadata[remote]?.domainId ??
-        this.multiProvider.getDomainId(remote);
+      const remoteId = this.multiProvider.getDomainId(remote);
```
yorhodes, line 131: ditto here and elsewhere

typescript/sdk/src/metadata/ChainMetadataManager.ts
```diff
@@ -61,14 +58,7 @@ export class ChainMetadataManager<MetaExt = {}> {
...
   addChain(metadata: ChainMetadata<MetaExt>): void {
-    const parseResult = safeParseChainMetadata(metadata);
-    if (!parseResult.success) {
-      throw new Error(
-        `Invalid chain metadata for ${
-          metadata.name
-        }: ${parseResult.error.format()}`,
-      );
-    }
+    ChainMetadataSchema.parse(metadata);
```
yorhodes, line 61: was this intentional?

typescript/sdk/src/token/Token.test.ts
```diff
@@ -168,18 +171,32 @@ describe('Token', () => {
...
+      sandbox.stub(multiProvider, 'getCosmJsWasmProvider').returns({
+        getBalance: async () => ({ amount: '100' }),
+        queryContractSmart: async () => ({
+          type: { native: { fungible: { denom: 'denom' } } },
+        }),
+      } as any);
+      sandbox.stub(multiProvider, 'getSolanaWeb3Provider').returns({
+        getBalance: async () => '100',
+        getTokenAccountBalance: async () => ({ value: { amount: '100' } }),
+      } as any);
```
yorhodes, line 191: these are nice
I wonder if we can export this in some test util

yorhodes (approved):
glad we are getting rid of the silly boilerplate fromEnvironment stuff

nbayindirli (commented):
initial pass lgtm– thanks for taking this step to further cleanup the codebase

FYI looks like yarn build in failing in CI

typescript/sdk/src/consts/multisigIsmVerifyCosts.ts
jmrossy: @yorhodes yes that's the tradeoff but the process of converting is just prepending 
`export const multisigIsmVerifyCosts = `. The benefits feel worth it to me. I'll modify the eslintignore config so we don't need that eslint disable comment at the top.

solidity/exportBuildArtifact.sh
```diff
@@ -26,7 +28,10 @@ if [ ! -f "$jsonFiles" ]; then
 fi
 
 # Extract required keys and write to outputFile
-if jq -c '{input, solcLongVersion}' "$jsonFiles" > "$outputFile"; then
+if jq -c '{input, solcLongVersion}' "$jsonFiles" > "$outputFileJson"; then
+  echo "export const buildArtifact = " > "$outputFileJs"
+  cat "$outputFileJson" >> "$outputFileJs"
+  echo "export const buildArtifact: any" > "$outputFileTsd"
```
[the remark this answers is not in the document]
jmrossy: I did considered that but I don't want core and sdk package consumers to need the hardhat package installed for the type checks to pass. Given that it was JSON before and treated as `any`, we aren't regressing at all here.

typescript/sdk/src/consts/environment.ts
```diff
@@ -0,0 +1 @@
+export type HyperlaneEnvironment = 'mainnet' | 'testnet';
```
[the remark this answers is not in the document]
jmrossy: Agreed, I'll move it

typescript/sdk/src/consts/sealevel.ts
```diff
@@ -1,2 +1,10 @@
+import type { ChainMap } from '../types.js';
+
 export const SEALEVEL_SPL_NOOP_ADDRESS =
   'noopb9bkMVfRPU8AsbpTUg8AQkHtKwMYZiFUjNRtMmV';
+
+export const solanaChainToClusterName: ChainMap<string> = {
```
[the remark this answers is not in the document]
jmrossy: It's used only for getting block explorer links. We can eliminate it entirely with an optional `clusterName` field on ChainMetadata for sealevel chains. I can do that now if you're cool with that approach.

typescript/sdk/src/metadata/ChainMetadataManager.ts
jmrossy: It was. Let's let better error msgs surface. IIRC you may have suggested this in the past and I pushed back but after looking closer at it I've come around

typescript/sdk/src/token/Token.test.ts
jmrossy: will do

typescript/sdk/src/metadata/ChainMetadataManager.ts
yorhodes: love it

nambrot: lets goooo

PR #3620, opened 2024-04-18
Deploy RC relayer for extra throughput

tkporter:
### Description

See https://discord.com/channels/935678348330434570/1230452070696484956 for context. General idea as a stopgap is to run a relayer concurrent with the Hyperlane one

### DIFF
--- a/typescript/infra/config/environments/mainnet3/agent.ts
+++ b/typescript/infra/config/environments/mainnet3/agent.ts
@@ -126,8 +126,66 @@ const contextBase = {
 } as const;
 
 const gasPaymentEnforcement: GasPaymentEnforcement[] = [
+  // To cover ourselves against IGP indexing issues and to ensure Nexus
+  // users have the best possible experience, we whitelist messages between
+  // warp routes that we know are certainly paying for gas.
+  {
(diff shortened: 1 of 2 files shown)


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -233,12 +231,11 @@ const releaseCandidate: RootAgentConfig = {
...
-    whitelist: releaseCandidateHelloworldMatchingList,
+    // We're temporarily (ab)using the RC relayer as a way to increase
+    // message throughput.
+    // whitelist: releaseCandidateHelloworldMatchingList,
     gasPaymentEnforcement,
-    transactionGasLimit: 750000,
-    // Skipping arbitrum because the gas price estimates are inclusive of L1
-    // fees which leads to wildly off predictions.
-    skipTransactionGasLimitFor: [chainMetadata.arbitrum.name],
+    metricAppContexts,
```
daniel-savu, line 238: nice, so this wouldn't show up in the dashboards because we're filtering by the `hyperlane` context normally, but if we need to we can check RC thanks to this

daniel-savu (approved):

PR #3622, opened 2024-04-18
Allow gasLimit overrides in the SDK/CLI for deploy txs

nambrot:
### Description

Small PR to actually allow overrides of gasLimit

### DIFF
--- a/typescript/sdk/src/providers/MultiProvider.ts
+++ b/typescript/sdk/src/providers/MultiProvider.ts
@@ -302,7 +302,7 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     const contractFactory = await factory.connect(signer);
 
     // estimate gas
-    const deployTx = contractFactory.getDeployTransaction(...params, overrides);
+    const deployTx = contractFactory.getDeployTransaction(...params);
     const gasEstimated = await signer.estimateGas(deployTx);
 
     // deploy with 10% buffer on gas limit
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -302,15 +302,17 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
     const gasEstimated = await signer.estimateGas(deployTx);
 
     // deploy with 10% buffer on gas limit
     const contract = await contractFactory.deploy(...params, {
-      ...overrides,
       gasLimit: gasEstimated.add(gasEstimated.div(10)), // 10% buffer
+      ...overrides,
     });
 
+    this.logger.debug({ transaction: deployTx }, `Deploying contract`);
```
jmrossy, line 314: A few things:
1) msg should come first ('deploying contract')
2) This will make a lot of noise
3) Why wrap the tx object in another object?

nambrot: Should I move it to trace? I do think its useful to be able to plug in the data in tenderly and debug lets say?

jmrossy: Yeah good idea

nambrot: Done

jmrossy (approved):

PR #3625, opened 2024-04-18
migrate fork util to SDK

nbayindirli:
### Description

* Migrates fork util to SDK

### Drive-by changes

* None

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3624

### Backward compatibility

* Yes

### Testing

* Manual on core & warp


typescript/sdk/src/utils/env.ts
```diff
@@ -0,0 +1,11 @@
+import z from 'zod';
+
+const envScheme = z.object({
+  HYP_KEY: z.string().optional(),
+  ANVIL_IP_ADDR: z.string().optional(),
+  ANVIL_PORT: z.number().optional(),
+});
+
+const parsedEnv = envScheme.safeParse(process.env);
```
jmrossy, line 9: accessing `process.foo` will throw an error when run in the browser, where `process` is undefined.
That's what the `safelyAccessEnvVar` utils is about. See logging.ts in utils for an example

typescript/sdk/src/utils/fork.ts
```diff
@@ -19,10 +14,10 @@ export enum ANVIL_RPC_METHODS {
 }
 
 /**
- * Resets the local node to it's original start (anvil [31337] at block zero).
+ * Resets the local node to it's original state (anvil [31337] at block zero).
  */
 export const resetFork = async () => {
-  logGray(`Resetting forked network...`);
+  rootLogger.info(`Resetting forked network...`);
```
jmrossy, line 20: I recommend making a child logger for this file. `rootLogger.child({module:'fork-utils'})`

typescript/sdk/src/utils/fork.ts
```diff
@@ -1,12 +1,7 @@
 import { providers } from 'ethers';
 
-import { ChainName, MultiProvider } from '@hyperlane-xyz/sdk';
-import { Address, isValidAddressEvm } from '@hyperlane-xyz/utils';
-
-import { logGray, logGreen } from '../logger.js';
-import { warnYellow } from '../logger.js';
-
-import { ENV } from './env.js';
+import { ChainName, ENV, MultiProvider } from '@hyperlane-xyz/sdk';
```
jmrossy, line 3: If this is in the SDK it shouldn't import from the sdk package like this

jmrossy (changes requested):
Needs a few fixes before we can merge :)

typescript/sdk/src/utils/env.ts
nbayindirli: Gotcha– did not have viz into this. Thanks for catching

typescript/sdk/src/utils/env.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { safelyAccessEnvVar } from '@hyperlane-xyz/utils';
+
+const envScheme = z.object({
+  HYP_KEY: z.string().optional(),
+  ANVIL_IP_ADDR: z.string().optional(),
+  ANVIL_PORT: z.number().optional(),
+});
+
+const parsedEnv = envScheme.safeParse(safelyAccessEnvVar('env', true));
```
jmrossy, line 11: Hmm, sorry if I wasn't clear before but I don't think this is quite right either. Wouldn't this try to access an env var called `env`? so `process.env['env']`. Was that the intention or are each of the env vars above meant to be set individually?

yorhodes (approved):

typescript/sdk/src/utils/env.ts
```diff
@@ -0,0 +1,11 @@
+import z from 'zod';
+
+const envScheme = z.object({
+  HYP_KEY: z.string().optional(),
+  ANVIL_IP_ADDR: z.string().optional(),
+  ANVIL_PORT: z.number().optional(),
+});
```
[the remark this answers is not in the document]
nbayindirli: Agreed– I've now removed the HYP_KEY from sdk-side, and similarly the ANVIL vars from the CLI-side.

Are you also against ANVIL vars sdk-side? We can certainly have these be passed into the fork util instead of parsed out of process.env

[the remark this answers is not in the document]
nbayindirli: ^talking through it, I think that's my desired path now as well

typescript/sdk/src/utils/env.ts
nbayindirli: Continuing– it now seems appropriate to encapsulate fork.ts in a class, given that it lives in SDK side + enables IO from CLI, which may need to persist certain values (recently discussed [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3586#discussion_r1564069199). By doing so, we can expose IP/port setters, which can eventually enable multi-chain support (when multiple Anvil instances will need to run)

cc @yorhodes

nbayindirli: Just to keep this PR moving, I've captured work to do this here: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3652

Will hold off on adding this to the current change

jmrossy (approved):

Issue #3547, opened 2024-04-04
[Epic][TxS] Modularize SDK TX Submission (w/ Transformers + Builder)


nambrot:
`Update` actions on modules will return an array of `PopulatedTx` (or equivalent for each VM)
- https://github.com/hyperlane-xyz/issues/issues/1157
- https://github.com/hyperlane-xyz/issues/issues/1156

We should have a tx `Submitter` abstraction that will submit these transactions to the network
This is an attempt to unify the `Governor` and `Deployer` abstractions that have disparate tx submission logic

```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3548
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3550
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3549
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3616
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3635
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3636
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3637
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3663
- [ ] Governor Bravo tx submitter?
- [ ] KMS signer tx submitter
- [ ] Provide fee estimate for proposed txns before submitting
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3676
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3684
```

nbayindirli:
- [ ] Provide fee estimate for proposed txns before submitting

avious00:
@nbayindirli FYI feel free to edit any issue, especially ones you're working on

PR #3627, opened 2024-04-19
feat: Modular Tx Submission – Create Transformer + Submitter + Builder Abstraction

nbayindirli:
## Description

- Adds modular transaction submission support for SDK clients, e.g. CLI.

To-do:
- [ ] Failing CI build due to linting Gnosis Safe package import
- [ ] Export to `sdk/src/index.ts`

Note: 
- Built to eventually expand to [Sealevel/CW support](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dc628476158672d08c766936ef4e2b150f86f566/typescript/sdk/src/providers/ProviderType.ts#L51-L85)

### Transformers
  - Input: `PopulatedTx[]`
  - Output: `HTX[]` (where `HTX extends HyperlaneTx`)
  - Purpose: Given a set of populated transactions, transform those transactions into a set a HyperlaneTxs (for the corresponding submitter), e.g.
```
...
const somePopulatedTxs = ...
const transformer = new InterchainAccountTxTransformer(mp,o,d,p);
const populatedTxs = transformer.transformTxs(somePopulatedTxs);
```

### Submitters
  - Input: `HTX[]` (where `HTX extends HyperlaneTx`)
  - Output: `TxReceipt[] | ResponseData[]`
  - Purpose: Given a set of Hyperlane transactions, execute those transactions for the specified submitter (submitter of type HTX should enforce the transactions being passed are of type HTX), e.g.
```
...
const submitter = new GnosisSafeTxSubmitter(mp,c,p);
const txReceipts =  submitter.submitTxs(populatedTxs);
---
Client-side example: for each gnosisTxReceipt display transactionHash
```

### Builder (Utilizes both Submitters & Transformer)
  - Input: `(TxTransformer<HTX> | TxTransformerType & Chain) & (TxSubmitter<HTX,HTR> | TxSubmitterType & Chain) & HTX[]` (where `HTX extends HyperlaneTx`)
  - Output: `TxReceipt[] | *response data*`
  - Purpose: Given a submitter, an optional transformer, and a set of PopulatedTransactions, transform and submit all transactions, e.g.
```
...
const eV5builder = new TxSubmitterBuilder<EV5Transaction, EV5TransactionReceipt>();
let txReceipts = eV5builder.for(
  new GnosisSafeTxSubmitter(chainA)
).transform(
  InterchainAccountTxTransformer(chainB)
).submit(
  txs
);
txReceipts = eV5builder.for(
  new ImpersonatedAccountTxSubmitter(chainA)
).submit(txs);
txReceipts = eV5builder.for(
  new JsonRpcTxSubmitter(chainC)
).submit(txs);
---
Client-side example: for each txReceipt display transactionHash | response data
```

### Drive-by changes

* None

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3547

### Backward compatibility

- Yes

### Testing

- Testing through CLI unit testing


typescript/sdk/src/providers/transactions/transformer/GnosisSafeTxTransformer.ts
```diff
@@ -0,0 +1,72 @@
...
+    return new GnosisSafeHyperlaneTx(
+      populatedTx,
+      safeAddress,
+      safeTransactionData,
+      safeTxHash,
+      senderAddress,
+      senderSignature,
+      origin,
+    );
+  }
```
yorhodes, line 71: what is the transform to be done here?
we aren't submitting the transactions onchain (only to the offchain API) so outputting another populated transaction I think is less useful for this one

typescript/sdk/src/providers/transactions/transformer/ImpersonatedAccountTxTransformer.ts
```diff
@@ -0,0 +1,38 @@
...
+      txs.push(tx);
+    }
+    return txs;
+  }
+
+  public async transformTx(
+    populatedTx: PopulatedTransaction,
+  ): Promise<ImpersonatedAccountHyperlaneTx> {
+    return new ImpersonatedAccountHyperlaneTx(populatedTx);
+  }
```
yorhodes, line 37: afaict this doesnt transform the tx content either? we just need to call an RPC method before

typescript/sdk/src/providers/transactions/transformer/TxTransformer.ts
```diff
@@ -0,0 +1,63 @@
...
+  public async transformTxs(
+    populatedTxs: PopulatedTransaction[],
+  ): Promise<HyperlaneTx[]> {
+    const txs: HyperlaneTx[] = [];
+    for (const populatedTx of populatedTxs) {
+      const tx = await this.transformTx(populatedTx);
+      txs.push(tx);
+    }
+    return txs;
+  }
```
yorhodes, line 56: this is duped across every implementation?

yorhodes (commented):
I'm a bit confused by the Tx type per submitter
ideally any `TxSubmitter` can take output from any `Module.update(config)`

typescript/sdk/src/batch/.gitkeep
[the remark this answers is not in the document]
nbayindirli: Will likely be removed– this was how I was going to organize the tx logic initially

typescript/sdk/src/providers/transactions/GnosisSafeHyperlaneTx.ts
```diff
@@ -0,0 +1,56 @@
+import { PopulatedTransaction } from 'ethers';
+
+import { HyperlaneTx } from './HyperlaneTx.js';
+
+enum OperationType {
+  Call = 0,
+  DelegateCall = 1,
+}
```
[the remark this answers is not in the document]
nbayindirli: This isn't actually used directly by us– I pulled in the types Gnosis used under the hood for our call to safe.proposeTranaction using `ProposeTransactionProps`. They don't expose via index so we unfortunately cannot reference directly. Other option obviously is to have this untyped

typescript/sdk/src/providers/transactions/GnosisSafeHyperlaneTx.ts
```diff
@@ -0,0 +1,56 @@
...
+  data: string;
+  operation?: OperationType;
+}
+interface SafeTransactionData extends MetaTransactionData {
+  operation: OperationType;
+  safeTxGas: string;
+  baseGas: string;
+  gasPrice: string;
+  gasToken: string;
+  refundReceiver: string;
```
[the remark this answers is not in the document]
nbayindirli: No they are not

typescript/sdk/src/providers/transactions/GnosisSafeHyperlaneTx.ts
```diff
@@ -0,0 +1,56 @@
...
+enum OperationType {
+  Call = 0,
+  DelegateCall = 1,
+}
+interface MetaTransactionData {
+  to: string;
+  value: string;
+  data: string;
+  operation?: OperationType;
+}
```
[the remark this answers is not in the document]
nbayindirli: Might as well maintain there same types if we're going to use any of them–

to reiterate, we could have this be untyped and destruct into proposeTx call

typescript/sdk/src/providers/transactions/HyperlaneTx.ts
```diff
@@ -0,0 +1,12 @@
...
+// TODO: May require additional fields
+type HyperlaneTxProps = {
+  populatedTx: PopulatedTransaction;
+};
+
+export class HyperlaneTx implements HyperlaneTxProps {
+  constructor(public populatedTx: PopulatedTransaction) {
+    this.populatedTx = populatedTx;
+  }
+}
```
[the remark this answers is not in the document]
nbayindirli: This is to enforce the type of tx being passed into our submitters/transformers extends HyperlaneTx– see `HTX` generic is used. Goal is for the builder to then execute txs based on the type, which the transformer would have converted the origin tx to

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
```diff
@@ -0,0 +1,60 @@
...
+    const txReceipts: TransactionReceipt[] = [];
+    for (const hyperlaneTx of hyperlaneTxs) {
+      const receipt = await this.sendTx(hyperlaneTx);
+      txReceipts.push(receipt);
+    }
+    return txReceipts;
+  }
+
+  public async sendTx(hyperlaneTx: HyperlaneTx): Promise<TransactionReceipt> {
+    return await this.multiProvider.sendTransaction(
```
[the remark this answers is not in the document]
nbayindirli: I believe mp.sendTx does return a Promise<ContractReceipt>, which we will need to use immediately by the caller

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Submits the Hyperlane transactions to the TxSubmitter.
+   * @param hyperlaneTxs The Hyperlane transactions to submit
+   * @returns The transaction receipts for the executed transactions
+   */
+  public async submitTxs(hyperlaneTxs: HTX[]): Promise<TransactionReceipt[]> {
+    for (const txSubmitter of this.txSubmitters) {
+      // TODO: Ensure the sendTxs() implementation is called iff that TxSubmitter supports the current HyperlaneTx type
+      return await txSubmitter.sendTxs(hyperlaneTxs);
```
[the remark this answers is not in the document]
nbayindirli: Right– see the TODO. This is a draft PR since there is currently incomplete logic

typescript/sdk/src/index.ts
```diff
@@ -441,6 +441,7 @@ export {
   TestChainNames,
 } from './types.js';
 export { MultiGeneric } from './utils/MultiGeneric.js';
+export { ENV } from './utils/env.js';
```
nbayindirli, line 444: Remove– added post-merge?

.changeset/poor-plums-pump.md
```diff
@@ -0,0 +1,6 @@
+---
+'@hyperlane-xyz/cli': patch
+'@hyperlane-xyz/sdk': patch
+---
+
+Add fork util to SDK
```
nbayindirli, line 6: Remove– added post-merge?

typescript/sdk/src/utils/env.ts
nbayindirli, line 1: Remove– added post-merge?

typescript/sdk/src/providers/transactions/transformer/GnosisSafeTxTransformer.ts
nbayindirli: Makes total sense– for draft PR I just threw the superset of potential data in here; pruning now to be the MVP of fields for the API call

nbayindirli: ^ Applies to other types in this PR as well

typescript/sdk/src/providers/transactions/transformer/TxTransformer.ts
```diff
@@ -0,0 +1,63 @@
...
+import { MultiProvider } from '../../MultiProvider.js';
+import { HyperlaneTx } from '../HyperlaneTx.js';
+
+export enum TxTransformerType {
+  DEFAULT /* Default transformer */,
+  SIGNER /* Private key transformer */,
+  IMPERSONATED_ACCOUNT /* Impersonated account transformer */,
+  GNOSIS_SAFE /* Gnosis Safe transformer */,
+  ICA /* Interchain Account transformer */, // TODO: Add
+}
```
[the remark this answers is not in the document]
nbayindirli: Matter of preference– transformers can be isolated (current impl), or included in the submitter interface.

IMO it seems more extensible to provide clients with the option to use a transformer like
```
const someTxTransformer = new SomeTxTransformer(mp, chain);
const someTxs = someTxTransformer.transformTxs(popTxs);
```

later on passing `someTxs` into the builder or directly to a submitter

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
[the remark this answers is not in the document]
yorhodes: this function also returns a Promise?

typescript/sdk/src/providers/transactions/GnosisSafeHyperlaneTx.ts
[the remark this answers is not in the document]
yorhodes: https://github.com/safe-global/safe-core-sdk/blob/main/packages/safe-core-sdk-types/src/types.ts

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilderTypes.ts
yorhodes, line 1: I find this to be pretty confusing tbh

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,163 @@
...
+export class TxSubmitterBuilder<
+  HTX extends HyperlaneTx,
+  HTR extends HyperlaneTxReceipt,
+> {
+  protected readonly logger: Logger = rootLogger.child({
+    module: 'transactions',
+  });
+
+  private currentSubmitter?: TxSubmitterInterface<HTX, HTR>;
+  private currentTransformer?: TxTransformerInterface<HTX>;
```
yorhodes, line 47: I would expect the builder to be fairly stateless

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,163 @@
...
+  constructor(private readonly multiProvider: MultiProvider) {
+    this.multiProvider = multiProvider;
+  }
+
+  /**
+   * Sets the current submitter for the builder.
+   * @param txSubmitterOrType The submitter to add to the builder
+   */
+  public add(
+    txSubmitterOrType: TxSubmitterXORType<HTX, HTR>,
```
yorhodes, line 58: imo we can just have the consumer pass in their own submitter instances here

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,163 @@
...
+      this.currentSubmitter,
+      'Must specify submitter to submit transactions.',
+    );
+
+    this.logger.info(
+      `Submitting ${hyperlaneTxs.length} transactions to the ${this.currentSubmitter.txSubmitterType} submitter...`,
+    );
+
+    let transformedHyperlaneTxs = hyperlaneTxs;
+    if (this.currentTransformer) {
```
yorhodes, line 141: how do we stack multiple transforms with this implementation?

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    const hyperlaneTxReceipts = await this.currentSubmitter.submitTxs(
+      transformedHyperlaneTxs,
+    );
+    this.logger.info(
+      `✅ Successfully submitted ${hyperlaneTxReceipts.length} transactions to the ${this.currentSubmitter.txSubmitterType} submitter.`,
+    );
+
+    this.currentSubmitter = undefined;
+    this.currentTransformer = undefined;
```
yorhodes, line 159: calling `submitTxs` twice would have quite different behavior with this implementation

yorhodes (commented):
submitter and transformer implementations look good
I still think we need to streamline the builder interface a bit to be extensible to future implementations

typescript/sdk/src/providers/transactions/submitter/GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,98 @@
...
+    public readonly multiProvider: MultiProvider,
+    public readonly chain: ChainName,
+    public readonly props: GnosisSafeTxSubmitterProps,
+  ) {
+    this.multiProvider = multiProvider;
+    this.chain = chain;
+    this.props = props;
+  }
+
+  public async submitTxs(hyperlaneTxs: HTX[]): Promise<HTR[]> {
```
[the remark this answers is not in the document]
nbayindirli: Yep I was trying to conform too closely to the interface here. Instead made submitTx optional and added support to batch all txs to the gnosis safe

typescript/sdk/src/providers/transactions/submitter/GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,98 @@
...
+    const safeService: SafeApiKit.default = getSafeService(
+      this.chain,
+      this.multiProvider,
+    );
+    const nextNonce: number = await safeService.getNextNonce(
+      this.props.safeAddress,
+    );
+    const safeTransaction: EthSafeTransaction.default =
+      await safe.createTransaction({
+        safeTransactionData: { to, value: '0', data },
```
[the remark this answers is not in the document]
nbayindirli: Indeed– now supported

typescript/sdk/src/providers/transactions/submitter/GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,98 @@
...
+      });
+    const safeTransactionData: SafeTransactionData = safeTransaction.data;
+    const safeTxHash: string = await safe.getTransactionHash(safeTransaction);
+    const senderAddress: Address = await this.multiProvider.getSignerAddress(
+      this.chain,
+    );
+    const safeSignature: EthSafeSignature = await safe.signTransactionHash(
+      safeTxHash,
+    );
+    const senderSignature: string = safeSignature.data;
```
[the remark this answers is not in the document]
nbayindirli: Ah yes, definitely agreed. Added this as optional, and will fallback to mp's if not present

typescript/sdk/src/providers/transactions/submitter/GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    this.logger.debug(
+      `Submitting transaction proposal to ${this.props.safeAddress} on ${this.chain}: ${safeTxHash}`,
+    );
+
+    return (await safeService.proposeTransaction({
```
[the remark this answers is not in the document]
nbayindirli: Removed

typescript/sdk/package.json
```diff
@@ -7,6 +7,8 @@
     "@cosmjs/stargate": "^0.31.3",
     "@hyperlane-xyz/core": "3.10.0",
     "@hyperlane-xyz/utils": "3.10.0",
+    "@safe-global/api-kit": "^1.3.0",
+    "@safe-global/protocol-kit": "^1.2.0",
```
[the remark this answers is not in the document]
nbayindirli: Also noting we'd have a fairly significant SDK + Infra refactor of the gnosisSafe util if we wanted to use the latest 2.3.0 + 3.0.2 releases:
- https://www.npmjs.com/package/@safe-global/api-kit
- https://www.npmjs.com/package/@safe-global/protocol-kit

typescript/sdk/src/providers/transactions/HyperlaneTx.ts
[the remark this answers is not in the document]
nbayindirli: Spoke offline- current suite of Submitters and Transformers are now EtherV5-specific and protocol-type is now specified at the builder-level

typescript/sdk/src/providers/transactions/HyperlaneTxReceipt.ts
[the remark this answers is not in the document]
nbayindirli: Spoke offline- see above

typescript/sdk/src/providers/transactions/submitter/TxSubmitterTypes.ts
```diff
@@ -0,0 +1,15 @@
+import { Address } from '@hyperlane-xyz/utils';
+
+export enum TxSubmitterType {
+  SIGNER = 'Signer',
```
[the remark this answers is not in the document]
nbayindirli: I agree here– Signer alone did not seem like the superset

typescript/sdk/src/providers/transactions/transformer/InterchainAccountTxTransformer.ts
```diff
@@ -0,0 +1,133 @@
...
+    module: 'transactions',
+  });
+
+  public readonly interchainAccount: InterchainAccount;
+
+  constructor(
+    public readonly multiProvider: MultiProvider,
+    public readonly originChain: ChainName,
+    public readonly destinationChain: ChainName,
+    public readonly props: InterchainAccountTxTransformerProps,
```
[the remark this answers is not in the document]
nbayindirli: Keen– also added to props

typescript/sdk/src/providers/transactions/submitter/SignerTxSubmitter.ts
```diff
@@ -0,0 +1,54 @@
...
+    const hyperlaneReceipts: HTR[] = [];
+    for (const hyperlaneTx of hyperlaneTxs) {
+      const hyperlaneReceipt = await this.submitTx(hyperlaneTx);
+      hyperlaneReceipts.push(hyperlaneReceipt);
+    }
+    return hyperlaneReceipts;
+  }
+
+  public async submitTx(hyperlaneTx: HTX): Promise<HTR> {
+    const receipt: ContractReceipt = await this.multiProvider.sendTransaction(
```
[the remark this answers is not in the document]
yorhodes: sgtm

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
nbayindirli: Hmm– I'm not sure how we'd avoid maintaining state. If we'd like to have 1 or more transformers, we must keep track of what transformers have augmented the original txn set

Note: The transformer stack datastructure is now present

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
nbayindirli: We previously could not, but this is now possible with the recent change

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
nbayindirli: This was with the assumption we would only transform once at a time. We now support multiple transforms for a single submit operation

nbayindirli:
> submitter and transformer implementations look good I still think we need to streamline the builder interface a bit to be extensible to future implementations

Agreed– with the newest update you'll find:
- Comment resolution
- Functionality to allow multiple transforms to occur for a single submit
- Utilization of `TypedTransaction` & `TypedTransactionReceipt`, enabling the builder to be used cross-protocol (EV5, CW, Sealevel, etc.)

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface TxSubmitterInterface<
+  TX extends TypedTransaction,
+  TR extends TypedTransactionReceipt,
+> {
+  /**
+   * Defines the type of tx submitter.
+   */
+  txSubmitterType: TxSubmitterType;
+  multiProvider: MultiProvider;
```
paulbalaji, line 18: as multiprovider is EVM specific, should there be an EvmTxSubmitterInterface that contains this instead??

i.e. if we build upon above comment
```ts
export interface EvmTxSubmitterInterface
  extends TxSubmitterInterface<ProtocolType.Ethereum> {
  multiProvider: MultiProvider;
}
```

each submitter then implements EvmTxSubmitterInterface

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  TypedTransaction,
+  TypedTransactionReceipt,
+} from '../../ProviderType.js';
+
+import { TxSubmitterType } from './TxSubmitterTypes.js';
+
+export interface TxSubmitterInterface<
+  TX extends TypedTransaction,
+  TR extends TypedTransactionReceipt,
```
paulbalaji, line 12: you could instead just take `TProtocol extends ProtocolType` and infer the tx/receipt like https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3a08e31b6d603a4eec691da4218a965c7c0a8283/typescript/sdk/src/crud/AbstractCrudModule.ts#L21

e.g. 
```ts
// TX
ProtocolTypedTransaction<TProtocol>

//TR
ProtocolTypedReceipt<TProtocol>
```

paulbalaji (commented):
suggested use of `ProtocolTypedTransaction` which could be useful across the submitter/builder classes

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * Should execute all transactions and return their receipts.
+   * @param txs The array of transactions to execute
+   */
+  submitTxs(txs: TX[]): Promise<TR[] | void>;
+  /**
+   * Should execute a transaction and return its receipt.
+   * @param tx The transaction to execute
+   */
+  submitTx?(tx: TX): Promise<TR | void>;
```
yorhodes, line 29: imo we should not "include the type in the name"
ie remove the `Txs` suffixes

typescript/sdk/src/providers/transactions/transformer/TxTransformer.ts
```diff
@@ -0,0 +1,22 @@
...
+   * Should transform all transactions of type TX into transactions of type TX.
+   * @param txs The array of transactions to transform
+   */
+  transformTxs(txs: TX[]): Promise<TX[]>;
+
+  /**
+   * Should transform a transaction of type TX into a transaction of type TX.
+   * @param tx The transaction to transform
+   */
+  transformTx?(tx: TX): Promise<TX>;
```
yorhodes, line 21: ditto here re: `Txs` suffixes

typescript/sdk/src/utils/gnosisSafe.ts
yorhodes, line 1: this will have to be adjusted once @jmrossy 's registry changes land

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+  protected readonly logger: Logger = rootLogger.child({
+    module: 'submitter-builder',
+  });
+
+  private currentSubmitter?: TxSubmitterInterface<TX, TR>;
+
+  constructor(
+    private readonly currentTransformers: Stack<
+      TxTransformerInterface<TX>
+    > = new Stack<TxTransformerInterface<TX>>(),
```
yorhodes, line 46: I dont think we want to force consumers to use this stack interface either

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    const txReceipts = await this.currentSubmitter.submitTxs(transformedTxs);
+    this.logger.info(
+      `✅ Successfully submitted ${transformedTxs.length} transactions to the ${this.currentSubmitter.txSubmitterType} submitter.`,
+    );
+
+    this.currentSubmitter = undefined;
+
+    return txReceipts ?? [];
```
yorhodes, line 107: when would this be undefined?

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+        `🔄 Transformed ${transformedTxs.length} transactions with the ${currentTransformer.txTransformerType} transformer...`,
+      );
+    }
+
+    const txReceipts = await this.currentSubmitter.submitTxs(transformedTxs);
+    this.logger.info(
+      `✅ Successfully submitted ${transformedTxs.length} transactions to the ${this.currentSubmitter.txSubmitterType} submitter.`,
+    );
+
+    this.currentSubmitter = undefined;
```
yorhodes, line 105: why assign to undefined here?

typescript/sdk/src/providers/transactions/submitter/ethersV5/ImpersonatedAccountTxSubmitter.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  public async submitTx(
+    tx: EthersV5Transaction,
+  ): Promise<EthersV5TransactionReceipt> {
+    const signer = await impersonateAccount(this.props.address);
+    this.multiProvider.setSigner(this.chain, signer);
+    const receipt: ContractReceipt = await this.multiProvider.sendTransaction(
+      this.chain,
+      tx.transaction,
+    );
```
yorhodes, line 58: should we inherit this behavior from JsonRPC submitter?

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
yorhodes: also why have the transformers be readonly but the submitter be mutating?

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Submits a set of transactions to the builder.
+   * @param txs The transactions to submit
+   */
+  public async submit(txs: TX[]): Promise<TR[]> {
+    assert(
+      this.currentSubmitter,
+      'Must specify submitter to submit transactions.',
+    );
```
yorhodes, line 84: I would enforce this by requiring a submitter in the constructor instead of the `for` interface

typescript/sdk/src/providers/transactions/submitter/ethersV5/GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+    const nextNonce: number = await safeService.getNextNonce(
+      this.props.safeAddress,
+    );
+    const safeTransactionBatch: MetaTransactionData[] = txs.map(
+      ({ transaction }: EthersV5Transaction) => {
+        const { to, data, value } = transaction;
+        assert(
+          to && data,
+          'Invalid EthersV5Transaction: Missing required metadata.',
```
yorhodes, line 80:
```suggestion
          'Invalid EthersV5Transaction: Missing required field to or data.',
```

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    let transformedTxs = txs;
+    while (this.currentTransformers.size > 0) {
+      const currentTransformer: TxTransformerInterface<TX> =
+        this.currentTransformers.pop();
+      transformedTxs = await currentTransformer.transformTxs(transformedTxs);
+      this.logger.info(
+        `🔄 Transformed ${transformedTxs.length} transactions with the ${currentTransformer.txTransformerType} transformer...`,
+      );
+    }
```
yorhodes, line 98: I dont really understand why we would mutate here?
what if we want to submit two sets of transactions with the same set of transformations and underlying submitter?

typescript/sdk/src/utils/gnosisSafe.ts
yorhodes: if we are not going to remove the corresponding infra code now, please make a ticket capturing this

yorhodes (commented):
mostly looks good, please replace the stack-typescript implemenetation

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
yorhodes:
```typescript
import { PopulatedTransaction } from "ethers";

interface TxSubmitter<ReturnType> {
    submit(txs: PopulatedTransaction[]): ReturnType;
}

type TxTransformer = TxSubmitter<PopulatedTransaction[]>;

const NoopTransformer: TxTransformer = {
  submit: (txs: PopulatedTransaction[]) => txs
}

class SubmitterBuilder<ReturnType> implements TxSubmitter<ReturnType> {
    constructor(
        private readonly inner: TxSubmitter<ReturnType>,
        private readonly transformer: TxTransformer = NoopTransformer
    ) {}

    add(transformer: TxTransformer): SubmitterBuilder<ReturnType> {
        return new SubmitterBuilder(this, transformer);
    }

    submit(txs: PopulatedTransaction[]): ReturnType {
        return this.inner.submit(this.transformer.submit(txs));
    }
}
```
already shared this but for posterity

typescript/sdk/package.json
```diff
@@ -17,6 +19,7 @@
     "cross-fetch": "^3.1.5",
     "ethers": "^5.7.2",
     "pino": "^8.19.0",
+    "stack-typescript": "^1.0.4",
```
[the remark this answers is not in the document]
nbayindirli: This seemed like a better option than implementing a generic stack impl of our own– but definitely not against adding that if you feel strongly

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
nbayindirli: This is a great callout– keeping MP here was an oversight on my part. Moving this out in the next commit

typescript/sdk/package.json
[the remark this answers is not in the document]
yorhodes: would the author accept a bribe to dependency inject our system? the payoff could be quite large for changing the contents of a governance transaction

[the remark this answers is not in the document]
yorhodes: in any case the native typescript Array is a stand-in replacement for your usage of this library so I do feel quite strongly that this is a very unnecessary dependency
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop

typescript/sdk/src/utils/gnosisSafe.ts
nbayindirli: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3684

typescript/sdk/src/providers/transactions/submitter/TxSubmitter.ts
paulbalaji: np - it was only merged into main the other day

typescript/sdk/src/providers/transactions/TransactionTypes.ts
```diff
@@ -0,0 +1,7 @@
+import {
+  EthersV5Transaction,
+  EthersV5TransactionReceipt,
+} from '../ProviderType.js';
+
+export type EV5Tx = EthersV5Transaction['transaction'];
+export type EV5Receipt = EthersV5TransactionReceipt['receipt'];
```
yorhodes, line 7: these are already exported as types by the ethers library so not sure why we want aliases?

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,92 @@
...
+    ...txs: ProtocolTypedTransaction<TProtocol>['transaction'][]
+  ): Promise<ProtocolTypedReceipt<TProtocol>['receipt'][] | void> {
+    this.logger.info(
+      `Submitting ${txs.length} transactions to the ${this.currentSubmitter.txSubmitterType} submitter...`,
+    );
+
+    let transformedTxs = txs;
+    while (this.currentTransformers.length > 0) {
+      const currentTransformer: TxTransformerInterface<TProtocol> =
+        this.currentTransformers.pop()!;
```
yorhodes, line 78: afaict this still means we cant use the same builder twice for the same submitter/transforms combo but at this point happy to move on

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -0,0 +1,92 @@
...
+ *    txs
+ *  );
+ *  txReceipts = eV5builder.for(
+ *    new EV5ImpersonatedAccountTxSubmitter(chainA)
+ *  ).submit(txs);
+ *  txReceipts = eV5builder.for(
+ *    new EV5JsonRpcTxSubmitter(chainC)
+ *  ).submit(txs);
+ */
+export class TxSubmitterBuilder<TProtocol extends ProtocolType> {
```
yorhodes, line 32:
```suggestion
export class TxSubmitterBuilder<TProtocol extends ProtocolType> implements TxSubmitterInterface<TProtocol>{
```

typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  public async submit(...txs: EV5Tx[]): Promise<void> {
+    const safe: Safe.default = await getSafe(
+      this.chain,
+      this.multiProvider,
+      this.props.safeAddress,
+    );
+    const safeService: SafeApiKit.default = getSafeService(
+      this.chain,
+      this.multiProvider,
```
yorhodes, line 48: I take back what I said about having a separate signer address, given that we attach the signer on the multiprovider in these utils

typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5GnosisSafeTxSubmitter.ts
```diff
@@ -0,0 +1,89 @@
...
+import { getSafe, getSafeService } from '../../../../utils/gnosisSafe.js';
+import { MultiProvider } from '../../../MultiProvider.js';
+import { EV5Tx } from '../../TransactionTypes.js';
+import { TxSubmitterType } from '../TxSubmitterTypes.js';
+
+import { EV5TxSubmitterInterface } from './EV5TxSubmitterInterface.js';
+
+interface EV5GnosisSafeTxSubmitterProps {
+  safeAddress: Address;
+  signerAddress?: Address;
```
yorhodes, line 23: actually we should omit this in favor of multi provider signer, my b

yorhodes (approved):

typescript/sdk/src/providers/transactions/TransactionTypes.ts
nbayindirli: Chatted offline– this was originally just a shorthand (sugar), but fine using the default EV5 types instead

typescript/sdk/src/providers/transactions/submitter/ethersV5/ImpersonatedAccountTxSubmitter.ts
nbayindirli: Thought through this a bit more– and I have decided to extend this by `EV5JsonRpcTxSubmitter`. Now, `EV5ImpersonatedAccountTxSubmitter`'s submit() calls to its parent's function after impersonation

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
nbayindirli: Since I'll be exposing this soon after, I went ahead and updated transform() to take 1 or more transformers and reassign currentTransformers. Then, submit reads through the transformers sequentially, preserving state.

nbayindirli:
Note that this no longer builds due to registry work– merging this PR is therefore now blocked by migrating registry.ts to the SDK: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3689

PR #3646, opened 2024-04-20
Migrate SDK consts to Registry [2/3] - Infra

jmrossy:
_NOTE: CI won't pass until parts 1 and 3 are included too_

### Description

- Update the HelloWorld package to read metadata from the registry lib
- Update the Infra package to read/write to a LocalRegistry

### Drive-by changes

- Use consistent, newest yaml lib version

### Related issues

https://github.com/hyperlane-xyz/issues/issues/917

### Backward compatibility

Yes

### Testing

Not very well tested, unexpected issues with infra scripts are possible


typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -1,17 +1,30 @@
...
-} from '@hyperlane-xyz/sdk';
+import { ChainMap, ChainMetadata, ChainName } from '@hyperlane-xyz/sdk';
 
 import { getChainMetadatas } from '../../../src/config/chain.js';
+import { getChain } from '../../registry.js';
 
-// The `Mainnets` from the SDK are all supported chains for the mainnet3 environment.
 // These chains may be any protocol type.
-export const supportedChainNames = Mainnets;
+export const supportedChainNames = [
```
jmrossy, line 7: Listing chains explicitly here like we already do for testnet, otherwise the tests break due to extra chains in the registry that aren't in the agent config

typescript/infra/tsconfig.json
```diff
@@ -6,8 +6,8 @@
     "outDir": "./dist/",
     "rootDir": "./",
     "noUnusedLocals": false,
-    "module": "es2022",
-    "moduleResolution": "node",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
jmrossy, line 10: This was left at es2022+node after my ESM PR to accommodate a bug in the Safe Global libraries but I need to update it now to utilize conditional exports from the registry pkg. This makes infra consistent with the rest of the monorepo. I found a different workaround for the Safe bug.

typescript/infra/config/environments/mainnet3/chains.ts
tkporter: This works for me and is probably preferable anyways. I imagine that the registry will eventually have e.g. PI mainnets in there that we don't want to run any validators / relayers on, so explicitly calling out which mainnets we want to run infra for makes sense

typescript/infra/config/registry.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function getChainMetadata(): ChainMap<ChainMetadata> {
+  return getRegistry().getMetadata();
+}
+
+export function getChainAddresses(): ChainMap<ChainAddresses> {
+  return getRegistry().getAddresses();
+}
+
+export function getEnvChains(env: DeployEnvironment): ChainName[] {
```
tkporter, line 67: I'm a little worried that this could lead to confusion / weird things with this function possibly returning a different set of chains than the list that's in the environment-specific configs (e.g. the `supportedChains` here in mainnet3 https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3646/files#diff-966e3e77cfd1880c5d9e2e5b69c9b9d8526b8d9006f3766ec21d33f226afba5dR7)

Is this intended to be used in a different way, e.g. instead of getting the list of chains that we want mainnet3 infrastructure to be aware of (e.g. "core" chains that we may wanna run validators / relayers for), this is just to get all "mainnet" chains in the registry? Wondering if there's any way to make the different more clear

typescript/infra/scripts/agent-utils.ts
```diff
@@ -371,29 +370,49 @@ export function getModuleDirectory(
...
-        getContractAddressesSdkFilepath(),
-        `${deployEnvToSdkEnv[environment]}.json`,
-      )
-    : path.join(getModuleDirectory(environment, module), 'addresses.json');
+  return REGISTRY_MODULES.includes(module) && environment !== 'test';
+}
+
+// Where non-registry module addresses are dumped.
+// This package must die in fire.
+function getInfraLandfillPath(environment: DeployEnvironment, module: Modules) {
```
tkporter, line 382: 😆

tkporter (approved):
genrally LGTM - really just looked at some of the new stuff in infra

typescript/helloworld/src/scripts/deploy.ts
yorhodes, line 1: fwiw I think most of this package's scripts are unused

typescript/infra/README.md
```diff
@@ -0,0 +1,8 @@
+# @hyperlane-xyz/infra
+
+Various scripts and utilities for managing and deploying Hyperlane infrastructure.
+
+## Tips
+
+- To enable more verbose logging, see the log env vars mentioned in the [root readme](../../README.md)
```
yorhodes, line 7: thanks!

typescript/infra/scripts/agent-utils.ts
yorhodes, line 1: thanks for cleaning this up
not sure why it's in agent-utils tho lol :P

typescript/infra/scripts/agent-utils.ts
```diff
@@ -371,29 +361,49 @@ export function getModuleDirectory(
...
+
+export function writeAddresses(
+  environment: DeployEnvironment,
+  module: Modules,
+  addressesMap: ChainMap<Record<string, Address>>,
+) {
+  if (isRegistryModule(environment, module)) {
+    for (const [chainName, addresses] of Object.entries(addressesMap)) {
+      getRegistry().updateChain({ chainName, addresses });
+    }
```
yorhodes, line 393: we still need an interface for warp routes here

yorhodes: this is actually quite a nice migration path though :)

typescript/infra/src/utils/safe.ts
```diff
@@ -1,16 +1,23 @@
...
 
-import { ChainName, MultiProvider, chainMetadata } from '@hyperlane-xyz/sdk';
+import { ChainName, MultiProvider } from '@hyperlane-xyz/sdk';
+
+import { getChain } from '../../config/registry.js';
+
+// Due to https://github.com/safe-global/safe-core-sdk/issues/419
+// See also https://github.com/safe-global/safe-core-sdk/issues/514
+const SafeApiKit = _SafeApiKit.default;
+const Safe = _Safe.default;
```
yorhodes, line 12: hmm, cc @nbayindirli

typescript/infra/tsconfig.json
yorhodes: I think we are pulling the Safe stuff into the SDK

yorhodes (approved):
this honestly wasnt as bad as I thought it was gonna be
thanks for slogging through it!

nbayindirli (approved):

nbayindirli (approved):
LGTM– note that build is failing on CI

typescript/infra/src/utils/safe.ts
nbayindirli, line 12: Since these are unused, could we instead update the imports to
```
import SafeApiKit from '@safe-global/api-kit';
import Safe, { EthersAdapter } from '@safe-global/protocol-kit';
```
and use `SafeApiKit.default` & `Safe.default` below?

jmrossy: I'll give that a try! If typescript accepts it then yes let's do that

PR #3648, opened 2024-04-22
feat: add `hyperlane hook read` + `hyperlane ism read` to cli

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/1183
resolves https://github.com/hyperlane-xyz/issues/issues/1187

- read hook/ism from a given address
- can specify format (json or yaml)
- can specify optional output file path (format overridden by format of output file)
- added changeset
- drive-by deleting `read-config.ts` script from infra, no need for duplicate logic

```
$ hyperlane hook read

Reads onchain Hook configuration for a given address

Options:
      --log        Log output format        [string] [choices: "pretty", "json"]
      --verbosity  Log verbosity level
                     [string] [choices: "debug", "info", "warn", "error", "off"]
      --help       Show help                                           [boolean]
  -c, --chains     A path to a JSON or YAML file with chain configs
                                     [string] [default: "./configs/chains.yaml"]
      --chain      The specific chain to perform operations with.
                                                             [string] [required]
      --address    Address of the Hook to read.              [string] [required]
  -f, --format     Output file format
                            [string] [choices: "json", "yaml"] [default: "yaml"]
  -o, --output     Output file path                                     [string]
```

```
$ hyperlane ism read

Reads onchain ISM configuration for a given address

Options:
      --log        Log output format        [string] [choices: "pretty", "json"]
      --verbosity  Log verbosity level
                     [string] [choices: "debug", "info", "warn", "error", "off"]
      --help       Show help                                           [boolean]
  -c, --chains     A path to a JSON or YAML file with chain configs
                                     [string] [default: "./configs/chains.yaml"]
      --chain      The specific chain to perform operations with.
                                                             [string] [required]
      --address    Address of the Interchain Security Module to read.
                                                             [string] [required]
  -f, --format     Output file format
                            [string] [choices: "json", "yaml"] [default: "yaml"]
  -o, --output     Output file path                                     [string]
```


typescript/cli/src/commands/options.ts
```diff
@@ -147,3 +147,19 @@ export const dryRunOption: Options = {
...
+  type: 'string',
+  description: 'The specific chain to perform operations with.',
+};
+
+export const addressCommandOption: Options = {
+  type: 'string',
+  description: 'An onchain address.',
+};
+
+export const concurrencyCommandOption: Options = {
```
paulbalaji, line 161: TODO: move this into chainmetadata

typescript/cli/src/commands/options.ts
```diff
@@ -147,3 +147,19 @@ export const dryRunOption: Options = {
...
+
+export const chainCommandOption: Options = {
+  type: 'string',
+  description: 'The specific chain to perform operations with.',
+};
+
+export const addressCommandOption: Options = {
+  type: 'string',
+  description: 'An onchain address.',
+};
```
jmrossy, line 159: The description's maybe too vague here. Maybe this could take an optional `description` param?

typescript/cli/src/commands/options.ts
jmrossy: Are you imagining this would be a property of the RPC config or something else?

typescript/cli/src/commands/options.ts
```diff
@@ -147,3 +147,19 @@ export const dryRunOption: Options = {
...
+
+export const concurrencyCommandOption: Options = {
+  type: 'number',
+  description: 'Option to override the default concurrency level.',
```
jmrossy, line 163: Also finding this param pretty vague.

typescript/cli/src/hook/read.ts
```diff
@@ -0,0 +1,54 @@
...
+import { FileFormat, writeFileAtPath } from '../utils/files.js';
+
+/**
+ * Read Hook config for a specified chain and address, logging or writing result to file.
+ */
+export async function readHookConfig({
+  chain,
+  address,
+  chainConfigPath,
+  concurrency = 20,
```
jmrossy, line 16: should we be duplicating the 20 const here when it's already in the param default?

typescript/cli/src/hook/read.ts
```diff
@@ -0,0 +1,54 @@
...
+}): Promise<void> {
+  const customChains = readChainConfigsIfExists(chainConfigPath);
+  const multiProvider = getMultiProvider(customChains);
+
+  // output file format overrides provided format
+  format = output?.endsWith('.json')
+    ? 'json'
+    : output?.endsWith('.yaml')
+    ? 'yaml'
+    : format;
```
jmrossy, line 35: Would the `readYamlOrJson` or related utils in typescript/cli/src/utils/files.ts help with this? If not, consider making a util there for this logic

typescript/cli/src/hook/read.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  if (
+    multiProvider.getChainMetadata(chain).protocol === ProtocolType.Ethereum
```
jmrossy, line 38:
```suggestion
    multiProvider.getProtocol(chain) === ProtocolType.Ethereum
```

typescript/cli/src/hook/read.ts
```diff
@@ -0,0 +1,54 @@
...
+      logBlue(`Hook Config at ${address} on ${chain}:`);
+      log(stringConfig);
+    } else {
+      writeFileAtPath(output, stringConfig + '\n');
+      logBlue(`Hook Config written to ${output}.`);
+    }
+    return;
+  }
+
+  logRed('Unsupported chain. Currently this command supports EVM chains only.');
```
jmrossy, line 53: Very small nit about style: instead of nesting all the business logic and falling back to this error, what if you inverted it to check up front to reduce nesting?

typescript/cli/src/ism/read.ts
```diff
@@ -0,0 +1,54 @@
...
+  chainConfigPath: string;
+  concurrency: number;
+  format: FileFormat;
+  output?: string;
+}): Promise<void> {
+  const customChains = readChainConfigsIfExists(chainConfigPath);
+  const multiProvider = getMultiProvider(customChains);
+
+  // output file format overrides provided format
+  format = output?.endsWith('.json')
```
jmrossy, line 31: ditto

typescript/sdk/src/hook/read.ts
```diff
@@ -1,5 +1,6 @@
 import { assert } from 'console';
 import { ethers, providers } from 'ethers';
+import { stringify as yamlStringify } from 'yaml';
```
jmrossy, line 3: Yaml is currently a dev dep of the SDK, not a runtime dep, meaning this would fail if someone didn't happen to have the yaml lib bundled already. You can make it a runtime dep, it's small so nbd.

typescript/sdk/src/ism/read.ts
```diff
@@ -69,8 +70,16 @@ export class EvmIsmReader implements IsmReader {
     this.provider = this.multiProvider.getProvider(chain);
   }
 
-  public static stringifyConfig(config: IsmConfig, space?: number): string {
-    return JSON.stringify(config, ethersBigNumberSerializer, space);
+  public static stringifyConfig(
```
jmrossy, line 73: You can DRY this up with the other stringify method by using a generic for the config type. E.g.

`stringifyConfig<C>( config:C...)`

typescript/sdk/src/hook/read.ts
```diff
@@ -80,8 +81,16 @@ export class EvmHookReader implements HookReader {
...
+  public static stringifyConfig(
+    config: HookConfig,
+    space?: number,
+    format: 'json' | 'yaml' = 'yaml',
+  ): string {
+    const json = JSON.stringify(config, ethersBigNumberSerializer, space);
+    if (format === 'json') {
+      return json;
+    }
+    return yamlStringify(JSON.parse(json), null, space);
```
jmrossy, line 93: Why are you running it through JSON first? yaml can take arbitrary data just like JSON.stringify can

jmrossy (commented):
Mostly looks good, just a few small things to fix

typescript/cli/src/commands/options.ts
paulbalaji: yup exactly, on the rpc config

typescript/cli/src/commands/options.ts
paulbalaji: good call, I originally had more verbose descriptions initially but dumbed down to de-dupe

typescript/cli/src/commands/hook.ts
```diff
@@ -0,0 +1,54 @@
...
+      chains: chainsCommandOption,
+      chain: {
+        ...chainCommandOption,
+        demandOption: true,
+      },
+      address: {
+        ...addressCommandOption,
+        demandOption: true,
+      },
+      concurrency: concurrencyCommandOption,
```
yorhodes, line 39: ditto, this should be via chain config

typescript/sdk/src/ism/read.ts
yorhodes: looks like the type isn't used at all so could also make the config type `object`

yorhodes: ditto on deduplicating

typescript/cli/src/hook/read.ts
paulbalaji: I would normally do that, but kept it this way for easy future additions of non-evm readers

typescript/sdk/src/ism/read.ts
paulbalaji: I think I'll probably do that by moving this to utils - would be useful for core/warp modules later too

typescript/sdk/src/hook/read.ts
paulbalaji: was a quick way to get the bignumber reducer working for yaml ouput - but haven't yet dug into exactly why the current reducer doesn't play nice

typescript/sdk/src/hook/read.ts
paulbalaji: good catch

typescript/cli/src/hook/read.ts
paulbalaji: I'm easy though, happy to make it nice now and when time comes to add more readers it can be updated again

typescript/cli/src/commands/hook.ts
```diff
@@ -0,0 +1,54 @@
...
+      format: fileFormatOption,
+      output: outputFileOption(),
```
yorhodes, line 41: can we make these global somehow?

yorhodes: also would expect default to be `stdout`

typescript/cli/src/hook/read.ts
yorhodes: agreed, can we pull this into a shared handler for all of these commands?

typescript/sdk/src/hook/read.ts
yorhodes: afaict this isnt specific to `EvmHookReader` or `HookReader` at all
can we pull this out into a generic object util in the CLI? and keep yaml out of the SDK?

typescript/cli/src/hook/read.ts
```diff
@@ -0,0 +1,54 @@
...
+    const hookReader = new EvmHookReader(multiProvider, chain, concurrency);
+    const config = await hookReader.deriveHookConfig(address);
+    const stringConfig = EvmHookReader.stringifyConfig(config, 2, format);
+    if (!output) {
+      logBlue(`Hook Config at ${address} on ${chain}:`);
+      log(stringConfig);
+    } else {
+      writeFileAtPath(output, stringConfig + '\n');
+      logBlue(`Hook Config written to ${output}.`);
+    }
```
yorhodes, line 49: ahh I see
presumably this `stdout` or `filepath` output can be reused across commands as well?

yorhodes (commented):
nice work

typescript/cli/src/hook/read.ts
paulbalaji: yeah i can tidy up some of the utilities here to take stdout as an option

typescript/sdk/src/hook/read.ts
jmrossy: > was a quick way to get the bignumber reducer working for yaml ouput 
That's a good reason, feel free to leave it as is but maybe add a comment

paulbalaji:
TODO:
- [ ] hoist `format`/`output` cli args to top-level
	- [ ] tidy up existing uses of them
- [ ] better handling of json/yaml/stdout options

jmrossy:
> TODO:
> 
>     * [ ]  hoist `format`/`output` cli args to top-level
>       
>       * [ ]  tidy up existing uses of them
> 
>     * [ ]  better handling of json/yaml/stdout logic

@paulbalaji Actually would you mind leaving things as is? I'll be removing a lot of that for the registry work

paulbalaji:
> @paulbalaji Actually would you mind leaving things as is? I'll be removing a lot of that for the registry work

absolutely - makes my life easier :D

typescript/cli/src/commands/hook.ts
jmrossy: We can but I'm already reworking the common args so let's leave it as is

jmrossy (approved):

jmrossy:
Consider updating the changeset to cover the utils change :)

paulbalaji:
> Consider updating the changeset to cover the utils change :)

I split it out into separate changesets for the parts that made sense

PR #3658, opened 2024-04-22
✨ Hyperlane Starknet

EvolveArt:
### Description

Adds support for starknet based chains in Hyperlane.


rust/chains/hyperlane-starknet/abis/Mailbox.contract_class.json
yorhodes, line 1: gitignore?

rust/chains/hyperlane-starknet/src/bindings.rs
yorhodes, line 1: ditto, ignore?

rust/utils/run-locally/src/starknet/cli.rs
aroralanuk, line 1: nice

aroralanuk (commented):
I took a cursory look at the changes and LGTM but @daniel-savu is closer to the agent code so he should be the one to approve.

rust/chains/hyperlane-cosmos/Cargo.toml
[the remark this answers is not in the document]
EvolveArt: Should I make them in another PR tho? Just smol improvements

ameten:
Required checks are failing. It will be great to fix them.

rust/main/chains/hyperlane-starknet/src/provider.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[instrument(err, skip(self))]
+    async fn is_contract(&self, address: &H256) -> ChainResult<bool> {
+        Ok(true)
+    }
+
+    #[instrument(err, skip(self))]
+    async fn get_balance(&self, address: String) -> ChainResult<U256> {
+        let eth_token_address =
+            felt!("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7");
+
```
ameten, line 131: I wonder why this address is hardcoded?

rust/main/chains/hyperlane-starknet/src/provider.rs
```diff
@@ -0,0 +1,156 @@
...
+                    sender: H256::zero(),
+                    recipient: None,
+                    raw_input_data: None,
+                    receipt: Some(TxnReceiptInfo {
+                        gas_used: U256::from_big_endian(
+                            invoke_receipt.actual_fee.amount.to_bytes_be().as_slice(),
+                        ),
+                        cumulative_gas_used: U256::zero(),
+                        effective_gas_price: None,
+                    }),
```
ameten, line 114: It will be great to populate all these fields

rust/main/chains/hyperlane-starknet/src/provider.rs
EvolveArt: this is the ETH token address, it's always the same will move it as a constant

rust/main/chains/hyperlane-starknet/src/provider.rs
EvolveArt: Unfortunately that's all the info available in the receipt on starknet

rust/main/chains/hyperlane-starknet/src/mailbox.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.contract.address.into()
+    }
+}
+
+#[async_trait]
+impl Mailbox for StarknetMailbox {
+    #[instrument(skip(self))]
+    async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32> {
+        let block_number =
+            get_block_height_for_reorg_period(&self.provider.rpc_client(), reorg_period).await?;
```
daniel-savu, line 146: nit: could you refactor this function to return the latest block number if `reorg_period` is None? This way you'd always have a `BlockId::Number(b)` too call the contract with and can simplify the code below

rust/main/chains/hyperlane-starknet/src/mailbox.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    #[instrument(skip(self), fields(msg=%message, metadata=%bytes_to_hex(metadata)))]
+    async fn process_estimate_costs(
+        &self,
+        message: &HyperlaneMessage,
+        metadata: &[u8],
+    ) -> ChainResult<TxCostEstimate> {
+        let _contract_call = self.process_contract_call(message, metadata, None).await?;
+
+        Ok(TxCostEstimate::default())
```
daniel-savu, line 242: How would one actually estimate the gas cost of a call? Is there some RPC / provider method for this?

rust/main/chains/hyperlane-starknet/src/merkle_tree_hook.rs
```diff
@@ -0,0 +1,198 @@
...
+                .await
+                .map_err(Into::<HyperlaneStarknetError>::into)?,
+        };
+
+        let mut branch = tree
+            .branch
+            .iter()
+            .map(|b| H256::from_slice(b.value.to_bytes_be().as_slice()))
+            .collect::<Vec<H256>>();
+        branch.resize(TREE_DEPTH, H256::zero());
```
daniel-savu, line 157: curious why this resizing was added, looking at other implementations I don't think this is done anywhere else

rust/main/chains/hyperlane-starknet/src/provider.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    #[instrument(err, skip(self))]
+    async fn is_contract(&self, address: &H256) -> ChainResult<bool> {
+        Ok(true)
+    }
+
+    #[instrument(err, skip(self))]
+    async fn get_balance(&self, address: String) -> ChainResult<U256> {
+        // ETH token address
```
daniel-savu, line 129: noob question but is eth the fee currency on starknet? Don't get why it's used here

rust/main/chains/hyperlane-starknet/src/utils.rs
```diff
@@ -0,0 +1,290 @@
...
+    contracts::{
+        interchain_security_module::ModuleType as StarknetModuleType,
+        mailbox::Bytes as MailboxBytes, mailbox::Message,
+        validator_announce::Bytes as ValidatorAnnounceBytes,
+    },
+    HyperlaneStarknetError,
+};
+
+/// Polls a function until it returns true or the max poll count is exceeded.
+pub async fn assert_poll<F, Fut>(f: F, polling_time_ms: u64, max_poll_count: u32)
```
daniel-savu, line 29: there's a similar function to this that we use everywhere else in the hyperlane monorepo, would be better to reuse than one to avoid code duplication. It's called `call_and_retry_n_times`, from `hyperlane-core/src/rpc_clients/retry.ts`

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
```diff
@@ -0,0 +1,225 @@
...
+            .contract
+            .announce(&EthAddress(validator), &storage_location, signature);
+
+        let gas_estimate = tx
+            .estimate_fee()
+            .await
+            .map_err(|e| HyperlaneStarknetError::AccountError(e.to_string()))?
+            .overall_fee;
+
+        let max_cost = gas_estimate * FieldElement::TWO;
```
daniel-savu, line 102: assuming the estimate is doubled as a buffer for any errors? Does the caller get refunded any unused gas?

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
```diff
@@ -0,0 +1,225 @@
...
+                            .collect::<Result<Vec<String>, ParseCairoShortStringError>>()
+                    })
+                    .collect::<Result<Vec<Vec<String>>, ParseCairoShortStringError>>()
+                    .map(|strings_vec| {
+                        strings_vec
+                            .into_iter()
+                            .map(|inner_vec| inner_vec.join(""))
+                            .collect::<Vec<String>>()
+                    })
+            })
```
daniel-savu, line 172: could you add a comment to clarify what's going on here? Not sure I understand

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        Ok(storage_locations)
+    }
+
+    #[instrument(ret, skip(self))]
+    async fn announce_tokens_needed(&self, announcement: SignedType<Announcement>) -> Option<U256> {
+        let validator = bytes_to_hex(&announcement.value.validator.to_vec());
+
+        let Ok((_, max_cost)) = self.announce_contract_call(announcement).await else {
+            trace!("Unable to get announce contract call");
```
daniel-savu, line 184: these logs would be better suited as `warn`s IMO

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    async fn announce(&self, announcement: SignedType<Announcement>) -> ChainResult<TxOutcome> {
+        let (contract_call, _) = self.announce_contract_call(announcement).await?;
+        let tx = contract_call
+            .send()
+            .await
+            .map_err(|e| HyperlaneStarknetError::AccountError(e.to_string()))?;
+        let invoke_tx_receipt =
+            get_transaction_receipt(&self.provider.rpc_client(), tx.transaction_hash).await;
```
daniel-savu, line 205: I'm seeing a pattern where txs are sent with `send` and then the receipt is fetched, maybe a good idea to refactor these into one `send_and_confirm` function or similar

daniel-savu (commented):
still got some reviewing left but leaving my feedback so far

rust/main/chains/hyperlane-starknet/src/provider.rs
EvolveArt: yup, there is no native token so the fee token is a contract

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
EvolveArt: yes it does! that's the usual good practice on starknet because gas estimate is not that precise

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -231,7 +231,7 @@ where
             .query_with_meta()
             .await?
             .into_iter()
-            .map(|(event, meta)| (Indexed::new(H256::from(event.message_id)), meta.into()))
+            .map(|(event, meta)| (H256::from(event.message_id).into(), meta.into()))
```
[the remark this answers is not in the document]
EvolveArt: reverted

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -0,0 +1,440 @@
...
+        self.get_finalized_block_number().await
+    }
+
+    /// Note: This call may return duplicates depending on the provider used
+    #[instrument(err, skip(self))]
+    async fn fetch_logs_in_range(
+        &self,
+        range: RangeInclusive<u32>,
+    ) -> ChainResult<Vec<(Indexed<HyperlaneMessage>, LogMeta)>> {
+        let key = get_selector_from_name("Dispatch").unwrap(); // safe to unwrap
```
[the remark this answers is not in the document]
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -0,0 +1,440 @@
...
+            address: Some(self.contract.address),
+            keys: Some(vec![vec![key]]),
+        };
+
+        let chunk_size = range.end() - range.start() + 1;
+
+        let mut events: Vec<(Indexed<HyperlaneMessage>, LogMeta)> = self
+            .contract
+            .provider
+            .get_events(filter, None, chunk_size.into())
```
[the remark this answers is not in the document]
EvolveArt: chunk size is defined as the block range so shouldnt be any issue there, been testing with a lot of different chunk sizes in prod

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -0,0 +1,440 @@
...
+            .await
+            .map_err(Into::<HyperlaneStarknetError>::into)?
+            .events
+            .into_iter()
+            .map(|event| {
+                // TODO: remove unwraps
```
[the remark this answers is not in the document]
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -0,0 +1,440 @@
...
+                    block_number: event.block_number.unwrap(),
+                    block_hash: H256::from_slice(
+                        event.block_hash.unwrap().to_bytes_be().as_slice(),
+                    ),
+                    transaction_id: H256::from_slice(
+                        event.transaction_hash.to_bytes_be().as_slice(),
+                    )
+                    .into(),
+                    transaction_index: 0,   // TODO: what to put here?
+                    log_index: U256::one(), // TODO: what to put here?
```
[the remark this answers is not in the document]
EvolveArt: ok so its fixed to one then, zero being the eth transfer

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -0,0 +1,440 @@
...
+                    address: H256::from_slice(event.from_address.to_bytes_be().as_slice()),
+                    block_number: event.block_number.unwrap(),
+                    block_hash: H256::from_slice(
+                        event.block_hash.unwrap().to_bytes_be().as_slice(),
+                    ),
+                    transaction_id: H256::from_slice(
+                        event.transaction_hash.to_bytes_be().as_slice(),
+                    )
+                    .into(),
+                    transaction_index: 0,   // TODO: what to put here?
```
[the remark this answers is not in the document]
EvolveArt: unfortunately we can't get this information easily

rust/main/chains/hyperlane-starknet/src/ism/aggregation_ism.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[async_trait]
+impl AggregationIsm for StarknetAggregationIsm {
+    #[instrument(err)]
+    async fn modules_and_threshold(
+        &self,
+        message: &HyperlaneMessage,
+    ) -> ChainResult<(Vec<H256>, u8)> {
+        let message = &StarknetMessage {
```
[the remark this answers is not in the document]
EvolveArt: done

note that we need an impl by crate since in the back the abi types generation doesnt merge types accross contracts, this is unfortunate but also allows for more control

rust/main/chains/hyperlane-starknet/src/ism/interchain_security_module.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let response = tx
+            .call()
+            .await
+            .map_err(Into::<HyperlaneStarknetError>::into)?;
+
+        // We can't simulate the `verify` call in Starknet because
+        // it's not marked as an entrypoint. So we just use the query interface
+        // and hardcode a gas value - this can be inefficient if one ISM is
+        // vastly cheaper than another one.
```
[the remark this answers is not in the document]
EvolveArt: unfortunately we are in the same situation there

rust/main/chains/hyperlane-starknet/src/mailbox.rs
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/utils.rs
EvolveArt: replaced

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
EvolveArt: added

rust/main/chains/hyperlane-starknet/src/validator_announce.rs
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/ism/aggregation_ism.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Create a reference to a AggregationISM at a specific Starknet address on some
+    /// chain
+    pub fn new(
+        conn: &ConnectionConf,
+        locator: &ContractLocator,
+        signer: Signer,
+    ) -> ChainResult<Self> {
+        let account = build_single_owner_account(
+            &conn.url,
+            signer.local_wallet(),
```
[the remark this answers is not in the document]
EvolveArt: that's a lack of the abigen types library of starknet, will look to replace

rust/main/chains/hyperlane-starknet/src/ism/interchain_security_module.rs
```diff
@@ -0,0 +1,164 @@
...
+        let module = self
+            .contract
+            .module_type()
+            .call()
+            .await
+            .map_err(Into::<HyperlaneStarknetError>::into)?;
+        Ok(to_hpl_module_type(module))
+    }
+
+    #[instrument]
```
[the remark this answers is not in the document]
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/mailbox.rs
EvolveArt: done

rust/main/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -94,58 +94,25 @@ pub struct MockDispatchInner {
     pub metadata: String,
 }
```
daniel-savu, line 95: would rather leave the files in `run-locally/src/cosmos` unchanged, are these changes intended?

rust/main/utils/run-locally/src/starknet/utils.rs
```diff
@@ -0,0 +1,256 @@
...
+    let mock_hook = cli.deploy(declarations.hpl_test_mock_hook, vec![]);
+
+    // deploy mailbox
+    println!("Deploying mailbox");
+    let mailbox = cli.deploy(
+        declarations.hpl_mailbox,
+        vec![
+            domain.to_string(),
+            deployer.clone(),
+            ism_multisig.clone(),
```
daniel-savu, line 173: should `ism_agregate` be the default mailbox ISM instead? For better testing coverage

rust/main/hyperlane-core/Cargo.toml
```diff
@@ -17,8 +17,10 @@ bigdecimal.workspace = true
 borsh.workspace = true
 bs58.workspace = true
 bytes = { workspace = true, features = ["serde"] }
+cainome = { workspace = true, optional = true }
 config = { workspace = true, optional = true }
 convert_case.workspace = true
+cosmrs = { workspace = true, optional = true }
```
daniel-savu, line 23: we're aiming to remove all chain-specific dependencies from `hyperlane-core` since we're running into increasingly more dependency conflicts as we add new VMs. The logic specific to these crates (`cainome`, `cosmrs`, `starknet`) looks to only be conversion-related, which can live in e.g. the `hyperlane-*` crate specific to that VM. I believe this is why we never actually implemented the `From` trait for cosmos TxResponses and instead use `tx_response_to_outcome`; may need to do something similar for starknet

rust/main/hyperlane-core/Cargo.toml
```diff
@@ -64,4 +67,6 @@ ethers = [
     "dep:primitive-types",
 ]
 solana = ["dep:solana-sdk"]
+cosmos = ["dep:cosmrs"]
+starknet = ["dep:starknet", "dep:cainome"]
```
daniel-savu, line 71: these features should be removed as well as a result

rust/main/utils/run-locally/src/starknet/mod.rs
```diff
@@ -0,0 +1,587 @@
...
+    //             "LINKING NODES: {} -> {:?}",
+    //             node.domain,
+    //             targets.iter().map(|v| v.domain).collect::<Vec<_>>()
+    //         );
+    //     }
+
+    //     for target in targets {
+    //         link_networks(&katanad, linker, validator, node, target);
+    //     }
+    // }
```
daniel-savu, line 367: in the cosmos implementation the `link_networks` step is used to configure the IGP oracles for each origin chain, and also configure the routing ISM which I'm not sure has coverage in the starknet e2e (but would very much like to have it covered if you're down). So the commented lines can be removed

rust/main/hyperlane-core/Cargo.toml
EvolveArt: makes sense, done

rust/main/utils/run-locally/src/cosmos/mod.rs
EvolveArt: yes this code was literally not up to date

EvolveArt: could probably be fixed in another pr tho

rust/main/utils/run-locally/src/starknet/mod.rs
EvolveArt: since this test is for starknet <-> starknet testing and there is no IGP support yet on starknet it doesnt make sense, so removed

rust/main/utils/run-locally/src/starknet/utils.rs
EvolveArt: right

rust/main/utils/run-locally/config.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "pragmadevnet",
```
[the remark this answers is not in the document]
EvolveArt: removed, it was for deployment purpose

rust/main/utils/run-locally/src/starknet/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    //         "Relayer has indexed {} gas payments, expected {}",
+    //         gas_payments_scraped,
+    //         expected_gas_payments
+    //     );
+    //     return Ok(false);
+    // }
+
+    let delivered_messages_scraped = fetch_metric(
+        &relayer_metrics_port.to_string(),
+        "hyperlane_operations_processed_count",
```
ameten, line 532: I wonder why we don't use `hyperlane_messages_processed_count` metric as Ethereum+Sealevel E2E tests use?

rust/main/utils/run-locally/src/starknet/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    // if starting_relayer_balance <= ending_relayer_balance {
+    //     log!(
+    //         "Expected starting relayer balance to be greater than ending relayer balance, but got {} <= {}",
+    //         starting_relayer_balance,
+    //         ending_relayer_balance
+    //     );
+    //     return Ok(false);
+    // }
+
+    log!("Termination invariants have been meet");
```
ameten, line 558:
```suggestion
    log!("Termination invariants have been met");
```

rust/main/utils/run-locally/src/starknet/mod.rs
EvolveArt: we do the same as for cosmos so we use `hyperlane_operations_processed_count`

rust/main/chains/hyperlane-starknet/src/ism/multisig_ism.rs
```diff
@@ -0,0 +1,148 @@
...
+            sender: StarknetU256::from_bytes_be(&message.sender.to_fixed_bytes()),
+            destination: message.destination,
+            recipient: StarknetU256::from_bytes_be(&message.recipient.to_fixed_bytes()),
+            body: StarknetBytes {
+                size: message.body.len() as u32,
+                data: message.body.iter().map(|b| *b as u128).collect(),
+            },
+        }
+    }
+}
```
ameten, line 112: It looks like we have this kind of conversion implemented is many places. I wonder if it would be possible to avoid code duplication?

rust/main/Cargo.toml
```diff
@@ -129,6 +134,15 @@ solana-account-decoder = "=1.14.13"
 solana-client = "=1.14.13"
 solana-sdk = "=1.14.13"
 solana-transaction-status = "=1.14.13"
+solana-zk-token-sdk = "=1.14.13"
+spl-associated-token-account = { version = "=1.1.2", features = [
+  "no-entrypoint",
+] }
+spl-noop = { version = "=0.1.3", features = ["no-entrypoint"] }
+spl-token = { version = "=3.5.0", features = ["no-entrypoint"] }
+spl-token-2022 = { version = "=0.5.0", features = ["no-entrypoint"] }
```
ameten, line 143: I wonder why we need this solana related dependencies here? It looks like starknet package does not use them.

rust/main/chains/hyperlane-starknet/src/ism/multisig_ism.rs
EvolveArt: yes well that's a whole point of having it under a From implementation

EvolveArt: as I mentionned in a previous comment, unfortunately due to the way cainome (abigen for starknet) works we have to re-write this everytime because the underlying struct is not the same

rust/main/chains/hyperlane-starknet/src/indexer.rs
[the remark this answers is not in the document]
daniel-savu: what if a message is dispatched by another contract that emits logs before calling the mailbox - would it still be fixed to one?

rust/main/chains/hyperlane-starknet/src/ism/aggregation_ism.rs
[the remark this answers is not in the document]
daniel-savu: cool, if you created a tracker ticket I'm happy to resolve this conversation

rust/main/chains/hyperlane-starknet/src/ism/aggregation_ism.rs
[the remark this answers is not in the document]
daniel-savu: > since in the back the abi types generation doesnt merge types accross contracts

wah that's pretty crazy, fair enough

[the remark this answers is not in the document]
daniel-savu: makes sense why you hadn't implemented the `From` in the first place

rust/main/chains/hyperlane-starknet/src/mailbox.rs
daniel-savu: nice, I also meant that inside `get_block_height_for_reorg_period` you could modify the `ReorgPeriod::None` arm to be
```rust
    let block_height = match reorg_period {
        ReorgPeriod::Blocks(blocks) => {
            let tip = provider.latest_block_height().await?;
            let block_height = tip - blocks.get() as u64;
            Some(block_height)
        }
        ReorgPeriod::None => provider.latest_block_height().await?;
```

since this doesn't affect the other callers of this function AFAICT, and the signature would change to this, to no longer return an `Option`:
```
pub(crate) async fn get_block_height_for_reorg_period(
    provider: &WasmGrpcProvider,
    reorg_period: &ReorgPeriod,
) -> ChainResult<u64> {
```

rust/main/chains/hyperlane-starknet/src/utils.rs
```diff
@@ -0,0 +1,305 @@
...
+type TransactionReceiptResult = ChainResult<MaybePendingTransactionReceipt>;
+
+/// Polls the rpc client until the transaction receipt is available.
+pub async fn get_transaction_receipt(
+    rpc: &Arc<AnyProvider>,
+    transaction_hash: FieldElement,
+) -> TransactionReceiptResult {
+    // there is a delay between the transaction being available at the client
+    // and the sealing of the block, hence sleeping for 100ms
+    call_and_retry_n_times(
```
daniel-savu, line 40: btw `call_and_retry_n_times` will sleep 2 seconds between retries. Compared to the 100ms in the comment I'm not sure if it affects any latency assumptions you had. Worth either updating the comment or adding an optional `sleep_duration` parameter to `call_and_retry_n_times` that defaults to the currently used `RPC_RETRY_SLEEP_DURATION`

rust/main/utils/run-locally/src/cosmos/mod.rs
daniel-savu: It should be as easy as overwriting these files with what's on the `main` branch right? Let's undo any `run-locally/src/cosmos` from this PR to keep it starknet specific

rust/main/chains/hyperlane-starknet/src/provider.rs
EvolveArt: right but it does involve another rpc call

daniel-savu: that's ok since we'd need this data for the explorer integration 
cc @ameten

rust/main/chains/hyperlane-starknet/src/utils.rs
EvolveArt: good catch, added the extra parameter

rust/main/utils/run-locally/src/cosmos/mod.rs
EvolveArt: done

rust/main/chains/hyperlane-starknet/src/mailbox.rs
EvolveArt: changed!

rust/main/chains/hyperlane-starknet/src/indexer.rs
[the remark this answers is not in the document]
EvolveArt: should be yes

rust/main/chains/hyperlane-starknet/src/ism/aggregation_ism.rs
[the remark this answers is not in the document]
EvolveArt: should I create it on this repo?

rust/main/chains/hyperlane-starknet/src/provider.rs
EvolveArt: Unfortunately many fields you referenced are only accesed when building the tx but not through the receipt

EvolveArt: I've filled most of the fields that were low hanging fruits. 

A few remarks for the missing ones:
- `gas_price` this is a tricky one since the gas price used by the tx is the one defined by the gateway, so its not available on the rpc (yet), we can add it when it is
- We dont have access to the `resource_bounds` field so gas related fields cannott be populated unfortunately, also there is no notion of gas limit really on starknet neither of priority fees
- Last thing is about the recipient, so on Starknet there is native account abstraction meaning that based on the type of account it is we will have to decode what's the contract being called, can be done but a bit dirty and not maintanable, I can create an issue for further investigation

PR #3659, opened 2024-04-22
Migrate SDK consts to Registry [3/3] - CLI

jmrossy:
### Description

Update the CLI to read and write from registries

### Related issues

https://github.com/hyperlane-xyz/issues/issues/917

### Backward compatibility

No, this changes several CLI arguments related to configs and artifacts

### Testing

TODO


typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -0,0 +1,132 @@
...
+      this.primaryRegistry = new LocalRegistry({
+        uri: primaryRegistryUri,
+        logger: logger.child({ registry: 'primary-local' }),
+      });
+    }
+
+    this.overrideRegistry = new LocalRegistry({
+      uri: overrideRegistryUri,
+      logger: logger.child({ registry: 'override-local' }),
+    });
```
yorhodes, line 58: why have this in the constructor?
I was expecting this to be a very thin wrapper over `registries: IRegistry[]` and just `fold(objMerge())` over them

typescript/cli/src/context.ts
```diff
@@ -1,145 +1,52 @@
...
-    coreArtifactsPath?: string;
-    promptMessage?: string;
-  };
   keyConfig?: KeyConfig;
   skipConfirmation?: boolean;
-  warpConfig?: {
-    warpConfigPath?: string;
-    promptMessage?: string;
-  };
 }
```
yorhodes, line 18: love it

typescript/cli/src/context.ts
```diff
@@ -1,145 +1,52 @@
...
 
 interface CommandContextBase {
   chains: ChainName[];
-  customChains: ChainMap<ChainMetadata>;
   multiProvider: MultiProvider;
+  registry: IRegistry;
```
yorhodes, line 23: can we merge chains and registry as well? or too far :P

yorhodes (commented):
my quick 2c but overall love getting rid of the separate local CLI config interface 🚀

typescript/cli/src/utils/keys.ts
```diff
@@ -14,34 +12,32 @@ const IMPERSONATED_KEY_TYPE = 'address';
...
-    keyConfig,
-    skipConfirmation,
-  );
-
-  return await addressToImpersonatedSigner(key);
+}: {
+  key?: string;
+  skipConfirmation?: boolean;
+}) {
+  key ||= await retrieveKey(IMPERSONATED_KEY_TYPE, skipConfirmation);
```
yorhodes, line 38: impersonated key type should never prompt for private key fwiw

typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -0,0 +1,154 @@
...
+    logger ||= rootLogger.child({ module: 'MergedRegistry' });
+    super({ uri: '__merged_registry__', logger });
+
+    if (!registryUris.length)
+      throw new Error('At least one registry URI is required');
+
+    this.registries = registryUris.map((uri, index) => {
+      // If not provided, allow the GithubRegistry to use its default
+      if (isHttpsUrl(uri)) {
+        return new GithubRegistry({ uri, logger: logger!.child({ index }) });
```
yorhodes, line 46: is this a TODO comment?

typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -0,0 +1,154 @@
...
+import {
+  isHttpsUrl,
+  objKeys,
+  objMerge,
+  rootLogger,
+} from '@hyperlane-xyz/utils';
+
+export interface MergedRegistryOptions {
+  registryUris: Array<string>;
+  isDryRun?: boolean;
```
yorhodes, line 27: surprised to see this here?

typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -0,0 +1,154 @@
...
+        return new GithubRegistry({ uri, logger: logger!.child({ index }) });
+      } else {
+        return new LocalRegistry({ uri, logger: logger!.child({ index }) });
+      }
+    });
+
+    this.isDryRun = !!isDryRun;
+  }
+
+  async listRegistryContent(): Promise<RegistryContent> {
```
yorhodes, line 55: nice
might be overkill but you could pull out the `map(registry => r.fn).reduce(objMerge))` into a helper with `fn keyof IRegistry` input

typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async addWarpRoute(config: WarpCoreConfig): Promise<void> {
+    return this.multiRegistryWrite(
+      async (registry) => await registry.addWarpRoute(config),
+      'adding warp route',
+    );
+  }
+
+  protected async multiRegistryWrite(
```
yorhodes, line 129: I would expect in practice users only want to write to 1 registry at a time (and probably local registry)

typescript/cli/src/utils/keys.ts
```diff
@@ -14,34 +12,32 @@ const IMPERSONATED_KEY_TYPE = 'address';
...
+  const signer = await addressToImpersonatedSigner(key);
```
yorhodes, line 39: is key address here? I know this isnt your code but it looks confusing

yorhodes (approved):
love all the 🟥🟥🟥

typescript/cli/src/registry/MergedRegistry.ts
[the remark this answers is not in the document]
jmrossy: No because it uses the LocalRegistry (fs)

typescript/cli/src/config/warp.ts
```diff
@@ -104,11 +97,19 @@ export async function createWarpRouteDeployConfig({
...
     errorRed(
       `Warp route deployment config is invalid, please see https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/examples/warp-route-deployment.yaml for an example`,
     );
     throw new Error('Invalid multisig config');
   }
 }
+
+// Note, this is different than the function above which reads a config
+// for a DEPLOYMENT. This gets a config for using a warp route (aka WarpCoreConfig)
+export function readWarpRouteConfig(filePath: string): WarpCoreConfig {
```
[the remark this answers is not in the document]
jmrossy: For the transfer command

typescript/cli/src/send/transfer.ts
```diff
@@ -47,48 +41,41 @@ export async function sendTestTransfer({
...
-  const { signer, multiProvider, customChains, coreArtifacts, warpCoreConfig } =
-    await getContext({
-      chainConfigPath,
-      coreConfig: { coreArtifactsPath },
-      keyConfig: { key },
-      warpConfig: { warpConfigPath },
-    });
+  const { chainMetadata } = context;
+
+  const warpCoreConfig = readWarpRouteConfig(warpConfigPath);
```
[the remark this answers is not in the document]
jmrossy: It does indeed

typescript/cli/src/utils/keys.ts
jmrossy: It prompts for an address

typescript/cli/src/registry/MergedRegistry.ts
jmrossy: I need it to have this registry behave differently when writing during a dry run, which is still TODO.
We can't simply point to the local registry to a diff root path because we may still need to read the data there

typescript/cli/src/registry/MergedRegistry.ts
jmrossy: Not sure I agree but anyway we don't have to decide that detail right now. It's skipping the GithubRegistry writes atm

typescript/cli/src/utils/keys.ts
jmrossy: It is a bit confusing. Yes, the key param was overloaded to actually be an address for dry runs. 
We could likely improve this but I'm going to leave things as they are for now

typescript/cli/src/registry/MergedRegistry.ts
jmrossy: It's not, I'll rephrase it

Issue #3484, opened 2024-03-25
Accommodate non-standard native token decimals in IGP storage oracles


yorhodes:
## Problem

The neutron native token has [6 decimals](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/254466f11aaf375c39131f0d140dfddd1e586d4b/typescript/sdk/src/consts/chainMetadata.ts#L552), but all the other supported chains have 18 decimals on the native token. This causes issues in the [gas oracle configuration](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f606e3ea2254e7fedde173b755690160465f3839/typescript/infra/config/environments/mainnet3/gas-oracle.ts#L20) which assumes that gas prices are formatted as gwei. 

## Solution

Remove gwei as a universal unit of gas price and instead use the chain's native token's decimals to format gas prices, or the chain's version of wei (pick one). This will allow us to support chains with different native token decimals.

Issue #3386, opened 2024-03-11
Lower Neutron gas prices used by relayer


tkporter:
## Problem

- Since https://governance.neutron.org/proposals/A/31, Neutron gas prices are significantly lower
- This is reflected in the Cosmos chain registry as well

## Solution

- change the gas price here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/config/mainnet3_config.json#L490 and redeploy the neutron context relayer
- check w/ rossy to see if the explorer has also updated

PR #3662, opened 2024-04-23
Update IGP gas oracles with new heuristics, get Cosmos gas prices from registry

tkporter:
### Description

The original and main motivation of this PR was to feel confident when running `deploy.ts -m igp` off of main. Prior to this, running it off main could result in misconfiguring gas oracles, particularly when it came to non-EVM remotes. See https://github.com/hyperlane-xyz/issues/issues/1178 for some context.

This PR does a number of things:
- Fetches Cosmos gas prices from the Cosmos registry
- the `gasPrices.json` now lets you specify the decimals relating to the amount in there. There are 2 reasons for this: it's easier for us to deal with gwei values for EVM chains (as we often end up manually changing these instead of just relying on the script). The second reason is covered in the next bullet
- Handles decimals when it comes to gas prices. The existing tooling assumed gas prices would always be denominated in wei. An extra weird situation arises with non-EVM chains: chains like Neutron have gas prices that are quotes in the smallest denomination (ie the equivalent of "wei" for NTRN), but the gas price is not an integer and is very small, e.g. `0.0053`. This poses issues because the IGP contracts require the gas price that's configured onchain to be an integer. Simply rounding up from 0.0053 to 1 results in almost 2 orders of magnitude difference :(. So the solution here is to have some logic that accommodates really tiny gas prices by scaling up the configured gas price and scaling down the configured token exchange rate.
- Applies some "non-linear" gas pricing heuristics. See https://github.com/hyperlane-xyz/issues/issues/1204 and https://discord.com/channels/935678348330434570/935678739663192184/1232293417309044818 for some context
- Running `update-agent-config.ts` will fetch gas prices for Cosmos chains from the Cosmos registry and ensure they're accurate

### Drive-by changes

Some logging messages when applying IGP gas oracle changes to make it more clear what the change actually does

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3386
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3484
- Fixes https://github.com/hyperlane-xyz/issues/issues/1178
- Fixes https://github.com/hyperlane-xyz/issues/issues/1206
- Fixes https://github.com/hyperlane-xyz/issues/issues/1204

### Backward compatibility

Fully backward compatible

### Testing

Ran deploy.ts for the IGP on all chains!


typescript/infra/src/govern/HyperlaneIgpGovernor.ts
```diff
@@ -52,6 +52,7 @@ export class HyperlaneIgpGovernor extends HyperlaneAppGovernor<
         break;
       }
       case IgpViolationType.GasOracles: {
+        break;
```
paulbalaji, line 55: meant to remove?

typescript/sdk/src/consts/environments/mainnet.json
paulbalaji, line 1: is there any meaningful change here or just a reordering?

typescript/sdk/src/gas/HyperlaneIgpDeployer.ts
```diff
@@ -111,14 +114,25 @@ export class HyperlaneIgpDeployer extends HyperlaneDeployer<
         !actual.gasPrice.eq(desired.gasPrice) ||
         !actual.tokenExchangeRate.eq(desired.tokenExchangeRate)
       ) {
-        this.logger.debug(
-          `-> ${remote} ${serializeDifference(actual, desired)}`,
+        this.logger.info(
+          `-> ${remote} - ${serializeDifference(actual, desired)}`,
```
paulbalaji, line 118:
```suggestion
          `-> ${remote}: ${serializeDifference(actual, desired)}`,
```
maybe colon instead to avoid any possible confusion of +ve/-ve diffs

typescript/infra/config/environments/mainnet3/gasPrices.json
```diff
@@ -1,20 +1,74 @@
...
+    "decimals": 9
+  },
+  "neutron": {
+    "amount": "0.0053",
+    "decimals": 1
+  },
+  "injective": {
+    "amount": "700000000",
+    "decimals": 1
+  }
```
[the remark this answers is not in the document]
tkporter: yep you're exactly right, it just pads zeroes. For neutron, see the PR description:
> Handles decimals when it comes to gas prices. The existing tooling assumed gas prices would always be denominated in wei. An extra weird situation arises with non-EVM chains: chains like Neutron have gas prices that are quotes in the smallest denomination (ie the equivalent of "wei" for NTRN), but the gas price is not an integer and is very small, e.g. 0.0053. This poses issues because the IGP contracts require the gas price that's configured onchain to be an integer. Simply rounding up from 0.0053 to 1 results in almost 2 orders of magnitude difference :(. So the solution here is to have some logic that accommodates really tiny gas prices by scaling up the configured gas price and scaling down the configured token exchange rate.

or the comment here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3662/files#diff-ac1f1ed84dad24fd8eeff4f7b636db47ba358af2aae94141c70bc757e58a59a3R27

It's super weird but somehow it's possible for the gas price to be < 1 and still be denominated in the smallest unit of NTRN - cosmos seemingly allows floating point gas prices

typescript/sdk/src/consts/environments/mainnet.json
tkporter: Just reordering, I think it hadn't been updated since we moved toward a json serializer that uses a deterministic ordering

typescript/infra/config/environments/mainnet3/gasPrices.json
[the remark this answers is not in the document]
paulbalaji: ahhhh gotcha, I did read that in the description but didn't quite piece it together til now - thanks!

paulbalaji (approved):

Issue #3667, opened 2024-04-24
CLI warp route deploys do not support Hook configuration


ltyu:
## Problem
Currently, the CLI only accepts a [string for hook](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f8b6ea467a169044257fb1247688b55ea6dbc78f/typescript/sdk/src/router/schemas.ts#L13), but it should accept an object similar to the `interchainSecurityModule`.

## Solution
- Consider updating the hook property to accept an object similar to the `interchainSecurityModule`.
- use `ts-to-zod` schema generator with the types. 
- Also fix [this commented todo](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3671#discussion_r1583307660)

PR #3671, opened 2024-04-25
Derive WarpConfig from TokenRouter address

### LINKED ISSUE #3579: Derive `WarpConfig` from `TokenRouter` address
Blocked by:
- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3129
- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3436

Concretely:
- If a CrudModule (see https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3574 for context) doesn't exist already for core already, create it
- Given an address to a warp route contract, generate a `ERC20RouterConfig` or `ERC721RouterConfig` (worth confirming w/ Yorke that these are the types we actually want)

### Related CLI Issues
- https://github.com/hyperlane-xyz/issues/issues/1189

ltyu:
### Description
Add Reader to derive WarpConfig from TokenRouter address

### Related issues

- Fixes #3579 

### Backward compatibility
Yes

### Testing

Unit Tests


typescript/sdk/src/token/read.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  async deriveWarpRouteConfig(
+    address: Address,
+  ): Promise<Omit<ERC20RouterConfig, 'type' | 'gas'>> {
+    const fetchedBaseMetadata = await this.fetchBaseMetadata(address);
+    const fetchedTokenMetadata = await this.fetchTokenMetadata(
+      fetchedBaseMetadata.wrappedTokenAddress,
+    );
+
+    const results: Omit<ERC20RouterConfig, 'type' | 'gas'> = {
```
ltyu, line 59: these 2 fields are omitted because 1) `type` can't be derived from the contracts. 2) Not sure if we need to return 'gas' for now. iiuc, this value comes from `destinationGas()` look up

.changeset/eight-cherries-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Add Derive WarpConfig from TokenRouter address
```
paulbalaji, line 5:
```suggestion
Add EvmERC20WarpRouterReader to derive WarpConfig from TokenRouter address
```

typescript/sdk/src/token/read.ts
```diff
@@ -0,0 +1,138 @@
...
+  | 'owner'
+  | 'wrappedTokenAddress'
+  | 'hook'
+  | 'interchainSecurityModule',
+  string
+>;
+
+interface WarpRouteReader {
+  deriveWarpRouteConfig(address: Address): Promise<Partial<ERC20RouterConfig>>;
+}
```
paulbalaji, line 27: the return type here needs to be more generic as an EvmERC721WarpRouteReader would need to implement the same interface but that returns `ERC721RouterConfig`. i understand this is tricky though for warp vs core/hook/ism because there's so many different variations of a warpconfig

so I suggest we remove this interface and in the warp crud module implementations we manually include `EvmERC20WarpRouterReader` and `EvmERC721WarpRouterReader` instead of a top level interface, as many of these types/routes end up being EVM specific

typescript/sdk/src/token/read.ts
paulbalaji: make `Omit<ERC20RouterConfig, 'type' | 'gas'>` a new type e.g. `DerivedERC20WarpRouteConfig` a la https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/055558a2661d94f8a9779f850a25d879ed413378/typescript/sdk/src/ism/read.ts#L44-L46

typescript/sdk/src/token/read.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Fetches the metadata for a token address.
+   *
+   * @param tokenAddress - The address of the token.
+   * @returns A partial ERC20 metadata object containing the token name, symbol, total supply, and decimals.
+   */
+  async fetchTokenMetadata(
+    tokenAddress: Address,
+  ): Promise<Partial<ERC20Metadata>> {
```
paulbalaji, line 127: it doesn't look like you're actually returning a Partial because the only missing field `scale` is optional anyway

typescript/sdk/src/token/read.ts
```diff
@@ -0,0 +1,138 @@
...
+import { DEFAULT_CONTRACT_READ_CONCURRENCY } from '../consts/crud.js';
+import { EvmHookReader } from '../hook/read.js';
+import { EvmIsmReader } from '../ism/read.js';
+import { MultiProvider } from '../providers/MultiProvider.js';
+import { ChainName } from '../types.js';
+
+type WarpRouteBaseMetadata = Record<
+  | 'mailbox'
+  | 'owner'
+  | 'wrappedTokenAddress'
```
paulbalaji, line 19: where is `wrappedTokenAddress` coming from? can't see it in the config types

typescript/sdk/src/token/read.ts
ltyu: Fixed

typescript/sdk/src/token/read.ts
ltyu: Good point. Will remove interface

.changeset/eight-cherries-develop.md
ltyu: Fixed

typescript/sdk/src/token/read.ts
ltyu: It should be `token` instead. Updated

typescript/sdk/src/token/read.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (
+      fetchedBaseMetadata.interchainSecurityModule !==
+      ethers.constants.AddressZero
+    ) {
+      results.interchainSecurityModule =
+        await this.evmIsmReader.deriveIsmConfig(
+          fetchedBaseMetadata.interchainSecurityModule,
+        );
+    }
+    // @todo add after https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3667 is fixed
```
paulbalaji, line 67: should we have a ticket to track this particular todo after #3667?

paulbalaji (approved):

Issue #3594, opened 2024-04-12
Validator client changes for posting ECDSARegistry signature to s3 bucket


aroralanuk:
The validator client right now posts the validator signature in a s3 bucket for anyone to announce them on chain. Similarly, we can do it for the ECDSA signature required for the Eigenlayer ServiceManager contract to register the operator.

aroralanuk:
Switched to using the CLI

PR #3674, opened 2024-04-29
feat: posting ECDSARegistry signature to s3 bucket for validators

aroralanuk:
### Description

- add announce_to_avs to validator code to publish their avs signature

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3594


rust/agents/validator/src/validator.rs
```diff
@@ -286,6 +293,39 @@ impl Validator {
         }
     }
 
+    async fn announce_to_avs(&self) -> Result<()> {
```
tkporter, line 296: The tx to submit this onchain isn't done by the agent right?

Wondering if we should make this function and `write_operator_registration` more consistent in some way - as someone who isn't familiar with eigenlayer, it's not immediately clear that both are related / are related to EL

rust/hyperlane-base/src/settings/staking_config.rs
```diff
@@ -0,0 +1,18 @@
+use hyperlane_core::H160;
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, Default)]
+pub struct StakingConf {
+    /// Address of the HyperlaneServiceManager contract
+    pub service_managers: HashMap<u32, H160>,
```
tkporter, line 7: will there be stuff in here other than the service_managers?

rust/agents/validator/src/settings.rs
```diff
@@ -40,6 +41,8 @@ pub struct ValidatorSettings {
     pub reorg_period: u64,
     /// How frequently to check for new checkpoints
     pub interval: Duration,
+
+    pub staking_config: StakingConf,
```
tkporter, line 45: Do you expect to ever have this be set via settings?

if we won't ever have a different config other than the `default_staking_config`, imo you could just not even have this in ValidatorSettings, and just create it in `impl BaseAgent for Validator {`

rust/agents/validator/src/validator.rs
```diff
@@ -122,6 +124,7 @@ impl BaseAgent for Validator {
             agent_metrics,
             chain_metrics,
             core_metrics: metrics,
+            staking_config: settings.staking_config,
```
tkporter, line 127: imo instead of storing a config in the Validator that you can then lookup
```
let service_manager = self
            .staking_config
            .service_managers
            .get(&avs_domain)
```
it'd be nicer to just do that lookup in here when building the Validator once, and then store the `service_manager` as the config in the Validator

rust/hyperlane-base/src/types/local_storage.rs
```diff
@@ -100,6 +106,19 @@ impl CheckpointSyncer for LocalStorage {
         Ok(())
     }
 
+    async fn write_operator_registration(
```
tkporter, line 109: I see what you mean about room to dedupe - imo this is fine, don't worry too much about duping here, especially if we're trying to just get stuff out the door

rust/hyperlane-core/src/types/operator_registration.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::utils::{domain_separator, fmt_domain};
+use crate::{Signable, SignedType, H160, H256};
+
+/// Eigenlayer AVSDirectory registration details
+#[derive(Builder, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct OperatorRegistration {
+    /// The domain of the chain the AVS is deployed on
+    pub domain: u32,
+    /// The EIP-712 typehash for the `Registration` struct used by the contract
+    #[builder(default = "default_operator_avs_registration_typehash()")]
```
tkporter, line 16: nifty!

rust/hyperlane-core/src/utils.rs
```diff
@@ -50,6 +50,27 @@ pub fn announcement_domain_hash(address: H256, domain: impl Into<u32>) -> H256 {
     )
 }
 
+/// Computes the domain separator for a given domain (for eigenlayer)
+pub fn domain_separator(domain: u32, service_manager_address: H160) -> H256 {
```
tkporter, line 54: any reason not to put these functions in operator_registration?

rust/hyperlane-core/src/utils.rs
```diff
@@ -50,6 +50,27 @@ pub fn announcement_domain_hash(address: H256, domain: impl Into<u32>) -> H256 {
...
+            .chain(domain.to_be_bytes())
+            .chain(&service_manager_address)
+            .finalize()
+            .as_slice(),
+    )
+}
+
+/// Compute the expiry for operator registration EIP-1271 signatures
+pub fn operator_registration_signature_expiry() -> u32 {
+    60 * 60 * 24 * 7 // 1 week
```
tkporter, line 71: is this supposed to be a unix timestamp? or what is this relative to

tkporter (commented):
directionally lgtm!

rust/agents/validator/src/validator.rs
```diff
@@ -286,6 +293,39 @@ impl Validator {
...
+            "Announcing to AVS, location: {:?} for validator: {:?}",
+            announcment_location, address
+        );
+
+        let avs_domain = 17000; // holesky test
+        let service_manager = self
+            .staking_config
+            .service_managers
+            .get(&avs_domain)
+            .unwrap();
```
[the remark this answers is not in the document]
aroralanuk: fixed

rust/agents/validator/src/validator.rs
aroralanuk: yes, that makes sense. Fixed

rust/hyperlane-base/src/settings/staking_config.rs
aroralanuk: Yes, I imagine we'll have more config once we have native staking which will also have some form of registration announcement in a json form.

rust/agents/validator/src/settings.rs
aroralanuk: given the above comment, I was anticipating we may need to set it via settings as you mentioned

rust/hyperlane-core/src/utils.rs
aroralanuk: moved

rust/hyperlane-core/src/types/operator_registration.rs
aroralanuk: thanks!

rust/hyperlane-core/src/utils.rs
aroralanuk: good point, I'm just using max u32 for now.

Issue #3448, opened 2024-03-20
Epic: basic self relaying via the CLI / explorer / warp UI


tkporter:
**High level outcomes**
- Only MessageIdMultisig ISMs, Routing ISMs, Aggregation ISMs, and Null ISM for now
- Able to self relay using the warp route UI, explorer, and CLI
- Everything is built in a protocol-agnostic manner, but we just implement the "view calls" etc that are involved for EVM for now

**Motivation:**
- We often just use combinations of aggregation, routing, and multisig ISMs
- We just focus on the MessageIdMultisig ISM type, as creating a merkle proof for a MerkleRootMultisig is heavyweight and atm only reasonable to do in the relayer. As it stands now, it would likely require indexing all messages, which is unreasonable to do in the CLI or explorer
- We want to lower the time to success in testing that e.g. a warp route works
- Allowing self-relaying if the relayer is having issues is a nice option for users
- We're not going with a shared WASM implementation between Rust and Typescript for a few reasons:
  - The way the agents are written, the metadata builder logic includes some very agent-specific things, e.g. a connection to the DB, metrics, a merkle tree, etc https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/base.rs#L272-L282
  - Rewriting this on the Rust side would be high effort and contention for agent eng resources is high
  - Metadata building isn't likely to change so duplicating logic between Rust and Typescript so we're unlikely to have any issues with them being inconsistent due to changes in the future

See [Notion](https://www.notion.so/hyperlanexyz/Self-relaying-015f388dfff940fdb228d55cfb23a7f4) for the full thought process around this

```[tasklist]
### OPTIONAL & dependent on prioritization
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3507
```

```[tasklist]
### Metadata building in the SDK
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3449
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3450
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3451
```

```[tasklist]
### Self-delivery of messages
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3452
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3453
[...]

PR #3693, opened 2024-05-01
feat: typescript relayer MVP

yorhodes:
### Description

- implement `HyperlaneRelayer` as a wrapper around `HyperlaneCore`
- add relaying to CLI status command
- add infra script relay for an environment

### Related issues

- Touches https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3448

### Testing

Manual


typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -72,64 +78,128 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     ](destinationId, recipient, body);
   };
 
-  protected getDestination(message: DispatchedMessage): ChainName {
+  onDispatch(
```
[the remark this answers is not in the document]
yorhodes: changed the interface so it's not EVM specific

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,203 @@
...
+    // parallelizable because configs are on different chains
+    const [ism, hook] = await Promise.all([
+      this.getRecipientIsmConfig(message),
+      this.getSenderHookConfig(message),
+    ]);
+    this.logger.debug({ ism, hook }, `Retrieved ISM and hook configs`);
+
+    const metadata = await pollAsync(
+      () => this.metadataBuilder.build(message, { ism, hook, dispatchTx }),
+      5 * 1000, // every 5 seconds
```
nambrot, line 128: Pull out into constants or parameters on the constructor

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -207,27 +292,39 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+    sourceTx: TransactionReceipt | ViemTxReceipt,
   ): DispatchedMessage[] {
     return HyperlaneCore.getDispatchedMessages(sourceTx);
   }
 
   async getDispatchTx(
     originChain: ChainName,
     messageId: string,
-  ): Promise<ethers.ContractReceipt | ViemTxReceipt> {
+    blockNumber?: number,
```
nambrot, line 303: Probably out of scope for this PR, but at some point it would be useful to have an indexing abstraction in typescript to search over a range

typescript/sdk/src/core/HyperlaneCore.ts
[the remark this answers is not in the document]
nambrot: I'm wondering what the practical constraints of using `contract.on(filter)` are going to be. Does ethers basically do the indexing logic in their side? How does it deal with flaky RPCs?

nambrot (approved):
lgtm, caveat again that my review is as somebody pretty far away

typescript/cli/src/status/message.ts
```diff
@@ -37,27 +41,50 @@ export async function checkMessageStatus({
...
-    if (!origin) {
-      origin = await runSingleChainSelectionStep(
-        context.chainMetadata,
-        'Select the origin chain',
-      );
+    deliveredTx = await core.getProcessedReceipt(message);
+  } else {
+    logBlue(`Message ${messageId} was not yet delivered`);
+
+    if (!selfRelay) {
```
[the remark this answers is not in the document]
yorhodes: imo it's a nice feature in replace of providing destination

jmrossy:
@aroralanuk @yorhodes I see there's new activity here. Should I re-review this now?

aroralanuk:
> @aroralanuk @yorhodes I see there's new activity here. Should I re-review this now?

Yes, that will be great!

typescript/cli/src/status/message.ts
```diff
@@ -12,17 +14,19 @@ export async function checkMessageStatus({
   destination,
   origin,
   selfRelay,
+  dispatchTx,
 }: {
   context: CommandContext;
+  dispatchTx?: string;
   messageId?: string;
   destination?: ChainName;
```
jmrossy, line 22: Is this arg still used for any logic or just that bit of logging below? If not, recommend removing

typescript/cli/src/status/message.ts
```diff
@@ -37,27 +41,50 @@ export async function checkMessageStatus({
     chainAddresses,
     context.multiProvider,
   );
-  const mailbox = core.getContracts(destination).mailbox;
+
+  let dispatchedTx: TransactionReceipt;
+  if (!dispatchTx) {
```
jmrossy, line 46: These two var names feel too close and I got confused which was which

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -175,19 +165,129 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+  }
+
+  protected getRecipient(message: DispatchedMessage): IMessageRecipient {
+    return IMessageRecipient__factory.connect(
+      bytes32ToAddress(message.parsed.recipient),
+      this.multiProvider.getProvider(this.getDestination(message)),
+    );
+  }
+
+  estimateHandle(message: DispatchedMessage): Promise<ethers.BigNumber> {
```
jmrossy, line 223: Recommend against using ethers BN types in our class interfaces. It's too evm specific, and specific too an outdated lib. `string` would be fine. `bigint` would be better but too inconsistent with what we've done elsewhere

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -175,19 +165,129 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
...
+
+  async getHook(
+    originChain: ChainName,
+    senderAddress: Address,
+  ): Promise<Address> {
+    const provider = this.multiProvider.getProvider(originChain);
+    try {
+      const client = MailboxClient__factory.connect(senderAddress, provider);
+      const hook = await client.hook();
+      if (!eqAddress(hook, ethers.constants.AddressZero)) {
```
jmrossy, line 254: the `isZeroishAddress` util fn may be a more robust alternative for what you're trying to do here

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,221 @@
+import { TransactionReceipt } from '@ethersproject/providers';
```
jmrossy, line 1: Import ethers exports though the 'front door' of `'ethers'` rather than these sub-packages. This type should be under the `providers` export.

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,150 @@
...
+  private hookCache:
+    | Record<ChainName, Record<Address, DerivedHookConfigWithAddress>>
+    | undefined;
+  private ismCache:
+    | Record<ChainName, Record<Address, DerivedIsmConfigWithAddress>>
+    | undefined;
+
+  public readonly logger;
+
+  constructor(protected readonly core: HyperlaneCore, caching = true) {
```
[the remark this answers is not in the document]
jmrossy: I still think this is worth changing

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,221 @@
...
+export const RelayerCacheSchema = z.object({
+  hook: z.record(z.record(DerivedHookConfigWithAddressSchema)),
+  ism: z.record(z.record(DerivedIsmConfigWithAddressSchema)),
+});
+
+type RelayerCache = z.infer<typeof RelayerCacheSchema>;
+
+export class HyperlaneRelayer {
+  private multiProvider: MultiProvider;
+  private metadataBuilder: BaseMetadataBuilder;
```
jmrossy, line 44: There's pretty much no reason to make member args like these not `protected`, especially when the data is likely public in other places anyway

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  async getHookConfig(
+    chain: ChainName,
+    hook: Address,
+  ): Promise<DerivedHookConfig> {
+    const config = (this.cache?.hook[chain]?.[hook] ??
+      (await new EvmHookReader(this.multiProvider, chain).deriveHookConfig(
+        hook,
+      ))) as DerivedHookConfig | undefined;
```
jmrossy, line 71: Tough to read

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  async getIsmConfig(
+    chain: ChainName,
+    ism: Address,
+  ): Promise<DerivedIsmConfig> {
+    const config = (this.cache?.ism[chain]?.[ism] ??
+      (await new EvmIsmReader(this.multiProvider, chain).deriveIsmConfig(
+        ism,
+      ))) as DerivedIsmConfig | undefined;
```
jmrossy, line 91: ditto

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,221 @@
...
+    const dispatchTx = await this.core.getDispatchTx(originChain, messageId);
+    return this.relayMessage(dispatchTx);
+  }
+
+  hydrate(cache: RelayerCache): void {
+    assert(this.cache, 'Caching not enabled');
+    this.cache = objMerge(this.cache, cache);
+  }
+
+  async hydrateDefaults(): Promise<void> {
```
jmrossy, line 178: I get the intention but I think `init` would be a more intuitive name for these fns for a new dev to guess that these must be called before the relayer can be used

typescript/sdk/src/index.ts
```diff
@@ -6,6 +6,8 @@ export {
   BaseSealevelAdapter,
   MultiProtocolApp,
 } from './app/MultiProtocolApp.js';
+export { S3Config, S3Receipt, S3Wrapper } from './aws/s3.js';
```
jmrossy, line 9: Tough to properly review the changes here with so much churn. I've asked this before: why is the index order shuffling so much when different devs change it? Why isn't prettier and/or the vscode ts import organizer action being consistent?

typescript/sdk/src/metadata/ChainMetadataManager.ts
```diff
@@ -106,6 +106,15 @@ export class ChainMetadataManager<MetaExt = {}> {
     return chainMetadata;
   }
 
+  getMaxBlockRange(chainNameOrId: ChainNameOrId): number {
+    const metadata = this.getChainMetadata(chainNameOrId);
+    return Math.max(
+      ...metadata.rpcUrls.map(
+        ({ pagination }) => pagination?.maxBlockRange ?? 2000,
```
jmrossy, line 113: Move this 2k to a const. Also IMO 2k is too big and many public providers won't support it. I'd go with 1k

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -277,6 +277,18 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
   }
 
+  /**
+   * Get the latest block range for a given chain's RPC provider
+   */
+  async getLatestBlockRange(
+    chainNameOrId: ChainNameOrId,
+    rangeSize = this.getMaxBlockRange(chainNameOrId),
+  ): Promise<{ fromBlock: number; toBlock: number }> {
+    const toBlock = await this.getProvider(chainNameOrId).getBlock('latest');
```
jmrossy, line 287: I'm slightly concerned that this strategy of establishing a block range with specific block numbers could cause us to miss useful data in the latest block. E.g. if this is called, then some slow other call is made, then the block range is used. By then, the 'latest' we got isn't the real latest. Not a blocking issue, just a concern

typescript/utils/src/addresses.ts
```diff
@@ -372,3 +372,8 @@ export function ensure0x(hexstr: string) {
 export function strip0x(hexstr: string) {
   return hexstr.startsWith('0x') ? hexstr.slice(2) : hexstr;
 }
+
+export const fromHexString = (hexstr: string) =>
+  Buffer.from(strip0x(hexstr), 'hex');
```
jmrossy, line 377: Redundant with `addressToBytes` above

typescript/utils/src/addresses.ts
```diff
@@ -372,3 +372,8 @@ export function ensure0x(hexstr: string) {
...
+
+export const toHexString = (buf: Buffer) => ensure0x(buf.toString('hex'));
```
jmrossy, line 379: redundant with `bytesTo*` methods above

typescript/utils/package.json
```diff
@@ -11,6 +11,7 @@
     "yaml": "^2.4.1"
   },
   "devDependencies": {
+    "@ethersproject/bytes": "^5.7.0",
```
jmrossy, line 14: Redundant with `ethers` which is already a runtime dep

jmrossy (commented):
Overall I think this is a great new feature but the design of it will have broad implications and so probably merits more scrutiny. More practically speaking, there are a bunch of valid older comments that haven't been addressed yet (plus my new ones) so still some work needed I think.

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: imo `private` is about encapsulation rather than "privacy" but agree they can be protected

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: its not necessary for them to be called, but speeds up metadata building when a message is being processed (skipping derivation)

yorhodes: agree that the naming is confusing here and probably leaks too much "implementation" detail

typescript/sdk/src/metadata/ChainMetadataManager.ts
yorhodes: I thought 2k was the max for most

typescript/sdk/src/providers/MultiProvider.ts
yorhodes: fair point, though what is an alternative? I think most RPC providers will vomit if you give a from block but to block as latest with "range too big"

typescript/utils/src/addresses.ts
yorhodes: `addressToBytes` enforces a specific length which we do not want here

typescript/utils/src/addresses.ts
yorhodes: ditto, I dont think they are redundant

typescript/utils/src/addresses.ts
jmrossy: @yorhodes then it probably doesn't belong in `addresses.ts` right?

typescript/sdk/src/hook/read.ts
```diff
@@ -115,8 +119,8 @@ export class EvmHookReader implements HookReader {
   async deriveMerkleTreeConfig(
     address: Address,
   ): Promise<WithAddress<MerkleTreeHookConfig>> {
-    const hook = MerkleTreeHook__factory.connect(address, this.provider);
-    assert((await hook.hookType()) === OnchainHookType.MERKLE_TREE);
+    // const hook = MerkleTreeHook__factory.connect(address, this.provider);
```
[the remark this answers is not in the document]
aroralanuk: stale

typescript/infra/scripts/relay.ts
```diff
@@ -0,0 +1,40 @@
...
+async function main() {
+  const { environment } = await getArgs().argv;
+  const { core } = await getHyperlaneCore(environment);
+  const relayer = new HyperlaneRelayer(core);
+
+  const chains = ['optimism', 'arbitrum', 'polygon', 'celo', 'base'];
+
+  try {
+    const contents = await readFile(CACHE_PATH, 'utf-8');
+    const cache = JSON.parse(contents);
```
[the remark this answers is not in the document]
aroralanuk: fixed already

typescript/cli/src/status/message.ts
aroralanuk: We're using it for core.getDispatchTx too

typescript/sdk/src/core/HyperlaneCore.ts
aroralanuk: switched to string

typescript/sdk/src/providers/MultiProvider.ts
aroralanuk: not sure how I can fix this. Leaving this one out.

typescript/utils/src/addresses.ts
aroralanuk: already been moved to strings.ts in util

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,244 @@
...
+      () => this.metadataBuilder.build({ message, ism, hook, dispatchTx }),
+      waitTime,
+      12, // 12 attempts
+    );
+
+    this.logger.info({ message, metadata }, `Relaying message ${message.id}`);
+    return this.core.deliver(message, metadata);
+  }
+
+  async relayMessages(
```
nbayindirli, line 176: should probably remove this since it's not used

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -0,0 +1,244 @@
...
+  ): Promise<ethers.ContractReceipt[]> {
+    const messages = HyperlaneCore.getDispatchedMessages(dispatchTx);
+    return Promise.all(
+      messages.map((message, index) =>
+        this.relayMessage(dispatchTx, index, message),
+      ),
+    );
+  }
+
+  async relayMessageId(
```
nbayindirli, line 187: ditto

PR #3697, opened 2024-05-02
add defaultDescription to yargs --key option

nbayindirli:
### Description

* Adds defaultDescription to yargs --key option to avoid printing sensitive data

### Drive-by changes

* None

### Related issues

* None

### Backward compatibility

* Yes

### Testing

* Manual

### DIFF
--- a/typescript/cli/src/commands/options.ts
+++ b/typescript/cli/src/commands/options.ts
@@ -45,6 +45,7 @@ export const keyCommandOption: Options = {
 Dry-run: An address to simulate transaction signing on a forked network`,
   alias: 'k',
   default: ENV.HYP_KEY,
+  defaultDescription: '***REDACTED***',
 };
 
 /* Command-specific options */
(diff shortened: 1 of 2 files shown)


typescript/cli/src/commands/options.ts
```diff
@@ -45,6 +45,7 @@ export const keyCommandOption: Options = {
 Dry-run: An address to simulate transaction signing on a forked network`,
   alias: 'k',
   default: ENV.HYP_KEY,
+  defaultDescription: '***REDACTED***',
```
jmrossy, line 48:
```suggestion
  defaultDescription: 'HYP_KEY env var',
```

jmrossy: How about this to signal to user that's where the default will come from?

nbayindirli: Something similar sure– how about `process.env.HYP_KEY`?

jmrossy (approved):

PR #3698, opened 2024-05-02
Fix CLI dry run flag handling

jmrossy:
### Description

- Fix dry run flag
- Prevent invalid selection in warp config create
- Improve logging of warp validation errors


typescript/cli/src/commands/options.ts
```diff
@@ -123,7 +123,7 @@ export const dryRunOption: Options = {
   type: 'string',
   description:
     'Chain name to fork and simulate deployment. Please ensure an anvil node instance is running during execution via `anvil`.',
-  alias: ['d'],
+  alias: 'd',
```
nbayindirli, line 126: I believe this had brackets because @yorhodes and I decided to have both `-d` and `-dr` for the `--dry-run` shorthand. Can we add that back in?

nbayindirli (approved):
Thanks for this– lgtm aside from small comment on dry-run option aliases

typescript/cli/src/commands/options.ts
jmrossy: Yeah I removed the -dr alias because I've seen weird quirks in yargs for aliases longer than one character. Are you okay with just having the one alias?

Issue #3451, opened 2024-03-20
Implement metadata building for Aggregation ISMs


tkporter:
Note this should be built out in a **protocol agnostic** manner, but we expect to only implement chain interactions (view calls, etc) for EVM for now.

In the SDK:
- [ ] Implement a metadata builder for an aggregation ISM. See [Rust implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/aggregation.rs#L121-L161) for inspiration and the Solidity library for [Aggregation ISM metadata](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/isms/libs/AggregationIsmMetadata.sol#L10).
  - [ ] Pick `threshold` metas that we were able to successfully build metadata for. Note that we won't be supporting the building of MerkleRootMultisig ISM metadata (because this requires data that's hard to get for self-relaying), and aggregation ISMs often include these. So we should be tolerant of us not being able to build sub module ISM metadata. In the agents, we have some logic to use the cheapest valid metadatas, but this is 100% optional, imo it's fine here to just arbitrarily pick the `threshold` sub modules that we were able to successfully build metadata for
  - [ ] Similar to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3450, we'll need to make sure that we don't recurse infinitely, e.g. that imagine someone configuring an aggregation ISM that points to itself

Issue #3449, opened 2024-03-20
Create a metadata builder in the SDK, implement metadata building for `MessageIdMultisig` and `Null` ISM types


tkporter:
Note this should be built out in a **protocol agnostic** manner, but we expect to only implement chain interactions (view calls, etc) for EVM for now.

In the SDK:
- [ ] Create a non-module-specific metadata builder that has a simple entrypoint for building metadata for a message. Feel free to use the [Rust implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/base.rs#L187-L198) as inspiration. This would:
  - [ ] Get the recipientIsm using the destination chain's Mailbox
  - [ ] figure out what module type it is and then call into a module-type-specific metadata builder
- [ ] Build out the `MessageIdMultisig`-specific metadata building. See [Rust implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/multisig/message_id_multisig.rs#L20-L86) for inspiration and the [Solidity metadata lib](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/isms/libs/MessageIdMultisigIsmMetadata.sol#L12-L16)
  - This would look at the origin chain's validator announce contract to get the storage location of each validator
  - We'll probably want to rip out some of the S3-related logic that exists in infra today (e.g. [S3Validator](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/src/agents/aws/validator.ts#L44)) and put this into the SDK
  - We'd then look at the storage locations of each validator in S3 to get the required signed checkpoints
  - Note that the nonce of the message != the checkpoint index. This means we need some way to figure out the checkpoint index for a particular message. In the agents, we index merkle tree insertion events so this is easy, we just perform a db read. Without this luxury, we'll need some other way.
    - After thinking about this a bit [here](https://www.notion.so/hyperlanexyz/Self-relaying-015f388dfff940fdb228d55cfb23a7f4?pvs=4#555c14eddce7449f9093ef4ec2e6ac6c), I think the easiest option for now will be to get the merkle tree index of the message by looking at the transaction that dispatched the message. Part of this task is figuring out a decent interface for building metadata and being able to do this
- [x] #3608
- [x] #3609

PR #3701, opened 2024-05-05
Multisig and Aggregation ISM Metadata Encoding

yorhodes:
### Description

- Implement multisig metadata encoding/decoding
- Implement aggregation metadata encoding/decoding
- Generates test metadata fixtures from solidity unit tests
- Test encoders using fixtures

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3449
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3451

### Backward compatibility

Yes

### Testing

Unit Tests


typescript/sdk/src/ism/metadata/aggregation.ts
```diff
@@ -0,0 +1,62 @@
...
+// null indicates that metadata is NOT INCLUDED for this submodule
+// empty or 0x string indicates that metadata is INCLUDED but NULL
+export interface AggregationIsmMetadata {
+  submoduleMetadata: Array<string | null>;
+}
+
+const RANGE_SIZE = 4;
+
+// adapted from rust/agents/relayer/src/msg/metadata/aggregation.rs
+export class AggregationIsmMetadataBuilder {
```
jmrossy, line 12: I'm not suggesting you change this because I realize the SDK is already super class-heavy but if a class 
1. has no state
2. has all static methods
3. isn't using inheritance in some way

Then it's not a class, it's a namespace and could have been an object. Just food for thought.

typescript/sdk/src/ism/metadata/aggregation.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  static encode(metadata: AggregationIsmMetadata): string {
+    const rangeSize = this.rangeIndex(metadata.submoduleMetadata.length);
+
+    let encoded = Buffer.alloc(rangeSize, 0);
+    metadata.submoduleMetadata.forEach((meta, index) => {
+      if (meta === null) {
+        return;
+      }
```
jmrossy, line 24:
```suggestion
      if (!meta) return;
```

typescript/sdk/src/ism/metadata/multisig.test.ts
```diff
@@ -0,0 +1,71 @@
+import { expect } from 'chai';
+import { readFileSync, readdirSync } from 'fs';
+
+import { SignatureLike } from '@hyperlane-xyz/utils';
+
+import { ModuleType } from '../types.js';
+
+import { MultisigMetadata, MultisigMetadataBuilder } from './multisig.js';
+
+type Fixture = {
```
jmrossy, line 10: Nit: Could be something like
```
MultisigFixture = Fixture<MultisigMetadata>;

type Fixture<T> = {
  decoded: T;
  encoded: string;
};
```

typescript/utils/src/addresses.ts
```diff
@@ -372,3 +372,8 @@ export function ensure0x(hexstr: string) {
 export function strip0x(hexstr: string) {
   return hexstr.startsWith('0x') ? hexstr.slice(2) : hexstr;
 }
+
+export const fromHexString = (hexstr: string) =>
+  Buffer.from(strip0x(hexstr), 'hex');
+
+export const toHexString = (buf: Buffer) => ensure0x(buf.toString('hex'));
```
jmrossy, line 379: These feel redundant with addressToBytes and bytesToProtocolAddress above. I realize they're are subtle differences. If we really want to keep these, please give them more descriptive names to clarify when these should be used instead

typescript/utils/package.json
```diff
@@ -11,6 +11,7 @@
     "yaml": "^2.4.1"
   },
   "devDependencies": {
+    "@ethersproject/bytes": "^5.7.0",
```
jmrossy, line 14: Why add this when ethers is already a runtime dep?

jmrossy (commented):
Mostly looks good, just some small things. Don't forget to add your new classes to the SDK index file and to add a changeset.

solidity/test/isms/MultisigIsm.t.sol
```diff
@@ -30,32 +34,72 @@ abstract contract AbstractMultisigIsmTest is Test {
...
     function getMetadata(
         uint8 m,
         uint8 n,
         bytes32 seed,
         bytes memory message
     ) internal returns (bytes memory) {
-        uint32 domain = mailbox.localDomain();
-        uint256[] memory keys = addValidators(m, n, seed);
-        uint256[] memory signers = ThresholdTestUtils.choose(m, keys, seed);
+        string memory structured = "structured";
```
nambrot, line 45: This function could use some comments or breaking out into separate functions

solidity/test/isms/AggregationIsm.t.sol
```diff
@@ -32,34 +37,42 @@ contract AggregationIsmTest is Test {
...
             } else {
-                uint64 offset = 0;
-                offsets = bytes.concat(offsets, abi.encodePacked(offset));
+                offsets = bytes.concat(offsets, abi.encodePacked(uint64(0)));
+                vm.serializeString(structured, i.toString(), "null");
             }
         }
-        return abi.encodePacked(offsets, metametadata);
+
+        string memory path = string(
```
nambrot, line 68: The fixture creation process should be documented either in README or at this call-site (or where it is consumed)

nambrot (approved):
Just some requested documentation/comments would be good. And then my approval should be taken with a grain-of-salt since i'm further away now

typescript/utils/src/addresses.ts
yorhodes: moving to the strings utils (for arbitrary length strings)

typescript/sdk/src/ism/metadata/aggregation.ts
yorhodes: agreed
this PR was bisected from code where this was not just static

Issue #3450, opened 2024-03-20
Implement metadata building for Routing ISMs


tkporter:
Note this should be built out in a **protocol agnostic** manner, but we expect to only implement chain interactions (view calls, etc) for EVM for now.

In the SDK:
- [x] Build out a Routing ISM metadata builder. See [Rust implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/metadata/routing.rs#L16-L27) for inspiration. This should:
  - [x] Find the ISM that is being routed to, and build metadata on that. The way we do this in the agents is by having the "base" metadata builder (i.e. the non-module-specific one described in https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3449) itself be a metadata builder that takes in an ISM address, and the base metadata builder is responsible for calling into module-specific metadata builders. So we can call into the base metadata builder from here
  - [x] Make sure that there is infinite loop protection - e.g. we want to make sure that someone can't make a routing ISM that routes to itself and then we just recurse infinitely

PR #3702, opened 2024-05-05
feat(sdk): ISM metadata building from message context

yorhodes:
### Description

- Uses derived hook and ISM config and dispatchTx of message to implement metadata fetching

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3450

### Backward compatibility

Yes

### Testing

E2E testing BaseMetadataBuilder


typescript/sdk/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "The official SDK for the Hyperlane Network",
   "version": "3.10.0",
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.74.0",
```
jmrossy, line 6: Another 0.5Mb JS lib in this already big SDK :(
https://bundlephobia.com/package/@aws-sdk/client-s3@3.569.0

The tree-shaking ESM enables mostly mitigates this but still would be better to avoid if we can

typescript/sdk/src/aws/validator.ts
```diff
@@ -0,0 +1,103 @@
...
+  async getAnnouncement(): Promise<Announcement> {
+    const resp = await this.s3Bucket.getS3Obj<S3Announcement>(ANNOUNCEMENT_KEY);
+    if (!resp) {
+      throw new Error('No announcement found');
+    }
+
+    return resp.data.value;
+  }
+
+  async getCheckpoint(index: number) {
```
jmrossy, line 76: I think the lint rules will require you to be explicit about return types for SDK functions. At least it should be enforcing that.

typescript/sdk/src/hook/read.ts
```diff
@@ -115,8 +119,8 @@ export class EvmHookReader implements HookReader {
   async deriveMerkleTreeConfig(
     address: Address,
   ): Promise<WithAddress<MerkleTreeHookConfig>> {
-    const hook = MerkleTreeHook__factory.connect(address, this.provider);
-    assert((await hook.hookType()) === OnchainHookType.MERKLE_TREE);
```
jmrossy, line 119: Just noticed this file imports the wrong `assert` fn. For cross-compat, it should be importing our own from utils.

typescript/sdk/src/ism/metadata/aggregation.ts
```diff
@@ -9,7 +24,54 @@ export interface AggregationIsmMetadata {
...
+    message: DispatchedMessage,
+    context: {
+      ism: WithAddress<AggregationIsmConfig>;
+      hook: DerivedHookConfigWithAddress;
+      dispatchTx: TransactionReceipt;
+    },
+    maxDepth = 10,
+  ): Promise<string> {
+    assert(maxDepth > 0, 'Max depth reached');
+    const promises = await Promise.allSettled(
```
jmrossy, line 50: Just confirming you intended to use `allSettled` here vs `all`?

typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -1,18 +1,35 @@
...
+  eqAddress,
+  eqAddressEvm,
   fromHexString,
+  rootLogger,
   strip0x,
   toHexString,
 } from '@hyperlane-xyz/utils';
 
-import { ModuleType } from '../types.js';
+import '../../../../utils/dist/types.js';
```
jmrossy, line 24: Is this meant to be here?

typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -42,7 +225,7 @@ export class MultisigMetadataBuilder {
     return toHexString(buf);
   }
 
-  static decodeSimplePrefix(metadata: string) {
+  private static decodeSimplePrefix(metadata: string) {
```
jmrossy, line 228: IMO private is very rarely the right choice. protected is sufficient

typescript/sdk/src/ism/metadata/aggregation.ts
yorhodes: yep, we can accommodate some failures

typescript/sdk/src/ism/metadata/multisig.ts
yorhodes: nope

typescript/sdk/package.json
jmrossy: We will want it there yeah. We can punt.

typescript/sdk/src/aws/s3.ts
```diff
@@ -17,37 +17,60 @@ export class S3Wrapper {
...
     if (!match) throw new Error('Could not parse bucket url');
     return new S3Wrapper(match[1], match[2], undefined);
   }
 
-  constructor(bucket: string, region: string, folder: string | undefined) {
+  constructor(
+    bucket: string,
+    region: string,
+    folder: string | undefined,
+    caching = false,
```
nambrot, line 32: These kind of things I'd personally prefer as an object (as it becomes harder to understand at the call-site)

typescript/sdk/src/hook/read.ts
```diff
@@ -115,8 +119,8 @@ export class EvmHookReader implements HookReader {
...
+    // const hook = MerkleTreeHook__factory.connect(address, this.provider);
```
nambrot, line 122: Are these comments intended?

typescript/sdk/src/ism/metadata/aggregation.ts
nambrot: Optional: Might need to add a timeout here?

typescript/sdk/src/ism/metadata/builder.ts
```diff
@@ -0,0 +1,140 @@
...
+  // assumes that all post dispatch hooks are included in dispatchTx logs
+  async build(
+    message: DispatchedMessage,
+    context: {
+      dispatchTx: TransactionReceipt;
+      hook: DerivedHookConfigWithAddress;
+      ism: DerivedIsmConfigWithAddress;
+    },
+    maxDepth = 10,
+  ): Promise<string> {
```
nambrot, line 75: This is maybe premature, but would love to have Result type in here which makes surfacing errors to consumers a lot easier (like no quorum for example)

typescript/sdk/src/ism/metadata/builder.ts
```diff
@@ -0,0 +1,140 @@
...
+    assert(maxDepth > 0, 'Max depth reached');
+    this.logger.debug(
+      { maxDepth, context },
+      `Building ${context.ism.type} metadata`,
+    );
+
+    if (context.ism.type === IsmType.TRUSTED_RELAYER) {
```
nambrot, line 82: This is this in this separate if clause, and not in the switch?

typescript/sdk/src/aws/s3.ts
yorhodes: yeah FWIW i just copied this from infra but will convert to object like @jmrossy suggested

typescript/sdk/src/hook/read.ts
yorhodes: they are unnecessary RPC round trips

typescript/sdk/src/aws/validator.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    if (isS3CheckpointWithId(s3Object.data)) {
+      return s3Object.data;
+    } else {
+      throw new Error('Failed to parse checkpoint');
+    }
+  }
+
+  async getLatestCheckpointIndex(): Promise<number> {
```
jmrossy, line 88: When this lands I can cut some code from the widgets lib that tries to get this for the Timeline widget :)

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -121,6 +127,38 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     }
   }
 
+  async send(
```
jmrossy, line 130: sendMessage

jmrossy: Will you be updating the CLI to use this instead? Also would be great to see this in MultiProtocolCore

jmrossy: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/send/message.ts#L105

typescript/sdk/src/ism/HyperlaneIsmFactory.hardhat-test.ts
```diff
@@ -73,15 +88,15 @@ const randomIsmConfig = (depth = 0, maxDepth = 2): IsmConfig => {
...
+      relayer: relayerAddress ?? randomAddress(),
     };
     return config;
   } else {
     throw new Error(`Unsupported ISM type: ${moduleType}`);
   }
 };
 
-describe('HyperlaneIsmFactory', async () => {
+describe.skip('HyperlaneIsmFactory', async () => {
```
jmrossy, line 99: Intentional? Maybe leave a TODO comment

typescript/sdk/src/ism/metadata/builder.ts
```diff
@@ -0,0 +1,139 @@
...
+  constructor(core: HyperlaneCore) {
+    this.multisigMetadataBuilder = new MultisigMetadataBuilder(core);
+    this.aggregationMetadataBuilder = new AggregationMetadataBuilder(this);
+    this.routingMetadataBuilder = new RoutingMetadataBuilder(this);
+    this.nullMetadataBuilder = new NullMetadataBuilder(core.multiProvider);
+    this.multiProvider = core.multiProvider;
+  }
+
+  // assumes that all post dispatch hooks are included in dispatchTx logs
+  async build(context: MetadataContext, depth = 10): Promise<string> {
```
jmrossy, line 67: depth -> maxDepth

typescript/sdk/src/ism/metadata/builder.ts
```diff
@@ -0,0 +1,139 @@
...
+      case IsmType.OP_STACK:
+      case IsmType.PAUSABLE:
+        return this.nullMetadataBuilder.build({ ...context, ism });
+
+      case IsmType.MERKLE_ROOT_MULTISIG:
+      case IsmType.MESSAGE_ID_MULTISIG:
+        const merkleTreeHook = deepFind(
+          hook,
+          (v): v is WithAddress<MerkleTreeHookConfig> =>
+            v.type === HookType.MERKLE_TREE && v.address !== undefined,
```
jmrossy, line 87: `&& !!v.address` is sufficient

typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -1,39 +1,217 @@
...
+  private validatorCache: Record<ChainName, Record<string, S3Validator>> = {};
+
+  constructor(
+    protected readonly core: HyperlaneCore,
+    protected readonly logger = rootLogger.child({
+      module: 'MultisigMetadataBuilder',
+    }),
+  ) {}
+
+  private async s3Validators(
```
jmrossy, line 61: IMO there's pretty much no reason any of the members + methods in these classes should be private instead of protected. Would be short-sighted

typescript/sdk/src/ism/metadata/null.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (context.ism.type === IsmType.TRUSTED_RELAYER) {
+      const destinationSigner = await this.multiProvider.getSignerAddress(
+        context.message.parsed.destination,
+      );
+      assert(
+        eqAddress(destinationSigner, context.ism.relayer),
+        `Destination signer ${destinationSigner} does not match trusted relayer ${context.ism.relayer}`,
+      );
+    }
+    return '0x';
```
jmrossy, line 27: Do we definitely want this instead of a zero address or something else? If so, consider making it a const like NULL_METADATA_SOMETHING and exporting it?

typescript/utils/src/arrays.ts
```diff
@@ -14,3 +16,7 @@ export function chunk<T extends Sliceable>(str: T, size: number) {
 export function exclude<T>(item: T, list: T[]) {
   return list.filter((i) => i !== item);
 }
+
+export function randomChoice<T>(list: T[]) {
```
jmrossy, line 20: randomChoice -> randomElement

jmrossy (approved):
Some suggestions but mostly lgtm

typescript/sdk/src/ism/metadata/null.ts
yorhodes: the length of this does indeed matter
can pull it into a const

typescript/sdk/src/ism/metadata/builder.ts
yorhodes: ngl I find this a little less readable

jmrossy: Fair enough. What's the expected behavior if address is `null` or `''`? Because the !! cover all 3 whereas this does not

yorhodes: hmm good point
we shouldnt return true if its null or ''

PR #3710, opened 2024-05-06
Increase gas limit buffer

daniel-savu:
### Description

Gas estimation seems to be inaccurate for `multicall` sometimes, leading to failed sub-txs. This PR bumps the value added to the gas estimate - hoping it's enough to cover all cases.

### Drive-by changes

Calls `set_next_attempt_after` before adding a message to the confirm queue in `submit_single_operation`. This was causing some messages to fail confirmation because it was checking too soon, and be re-added to the prep queue.


yorhodes (approved):

rust/config/mainnet_config.json
```diff
@@ -1,6 +1,7 @@
 {
   "chains": {
     "ancient8": {
+      "batchContractAddress": "0x4C97D35c668EE5194a13c8DE8Afc18cce40C9F28",
```
yorhodes, line 4: why not call this multicall3?

daniel-savu: I wanna keep it generic to all execution envs since we may soon add support for cosmwasm

yorhodes: gotcha
FYI cosmwasm has native batching

tkporter (approved):

PR #3713, opened 2024-05-06
Fixes for CLI dry-runs

jmrossy:
### Description

- Fixes type of dry-run arg in deploy command definitions
- Runs verifyAnvil during dry-run context setup


typescript/cli/src/deploy/warp.ts
```diff
@@ -227,7 +227,7 @@ async function executeDeploy(params: DeployParams) {
 
   logGreen('✅ Hyp token deployments complete');
 
-  log('Writing deployment artifacts');
+  if (!isDryRun) log('Writing deployment artifacts');
```
yorhodes, line 230: I thought we had a dry run artifacts format?

yorhodes (approved):

typescript/cli/src/deploy/warp.ts
jmrossy: the registry tooling doesn't yet have support for configurable output sub-directories and we wouldn't want the dry-run to write over real artifacts.
What makes this tricky is the _read_ still happens from the normal place but he _write_ must happen somewhere else. Punted on this.

jmrossy: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3715

PR #3717, opened 2024-05-06
Workaround TS bug in Safe protocol-lib

jmrossy:
### Description

Replace SDK's `gnosisSafe.ts` utils file with a pure JS version to avoid TS checks on Safe lib which has an incorrectly placed TS file in it's lib dist folder:

Screenshot 2024-05-06 at 3 08 38 PM

### Drive-by changes

Fix assert import

## Related issues

https://github.com/safe-global/safe-core-sdk/issues/805

### DIFF
--- a/typescript/infra/scripts/safe-delegate.ts
+++ b/typescript/infra/scripts/safe-delegate.ts
@@ -4,9 +4,8 @@ import { LedgerSigner } from '@ethersproject/hardware-wallets';
 import '@ethersproject/hardware-wallets/thirdparty';
 import { AddSafeDelegateProps } from '@safe-global/api-kit';
 
-// prettier-ignore
 // @ts-ignore
-import { getSafeDelegates, getSafeService } from '@hyperlane-xyz/sdk/src/utils/gnosisSafe.js';
+import { getSafeDelegates, getSafeService } from '@hyperlane-xyz/sdk';
 
 import { getChains } from '../config/registry.js';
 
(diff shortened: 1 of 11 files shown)


typescript/infra/scripts/safe-delegate.ts
```diff
@@ -4,7 +4,9 @@ import { LedgerSigner } from '@ethersproject/hardware-wallets';
 import '@ethersproject/hardware-wallets/thirdparty';
 import { AddSafeDelegateProps } from '@safe-global/api-kit';
 
-import { getSafeDelegates, getSafeService } from '@hyperlane-xyz/sdk';
+// prettier-ignore
+// @ts-ignore
+import { getSafeDelegates, getSafeService } from '@hyperlane-xyz/sdk/src/utils/gnosisSafe.js';
```
jmrossy, line 9: These reach directly over because these utils aren't exported from the SDK anymore. We shouldn't export anything that doesn't have proper typing.

nbayindirli (approved):
ugh

typescript/sdk/src/index.ts
```diff
@@ -314,6 +303,17 @@ export {
   defaultViemProviderBuilder,
   protocolToDefaultProviderBuilder,
 } from './providers/providerBuilders.js';
+export { TxSubmitterInterface } from './providers/transactions/submitter/TxSubmitterInterface.js';
```
jmrossy, line 306: I don't know why different devs seem to get different results for the ordering here.

Issue #3736, opened 2024-05-08
Parse user-address from HYP_KEY instead of using a defined user-address for dry-runs


nbayindirli:

nbayindirli:
Fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3743

PR #3743, opened 2024-05-08
ensure HYP_KEY & --key are only used for PKs & let user pass --from-address from dry-runs

nbayindirli:
### Description

* This change ensures `HYP_KEY` and `--key` are only used for a developer's private key, and allows a user to pass in a `--from-address` for dry-runs (everything defaults to `HYP_KEY`, if sourced)

### Drive-by changes

- Improved DX by removing incorrect log statement after insufficient balance check
```
Before:
  ✅ Signer is valid
  ? WARNING: 0x2d54EC30Be0D2583b0d781e10B31905c3c54616d has low balance on alfajores. At least 1.0 recommended but found 0.0 CELO Continue? yes
  ✅ Balances are sufficient

After:
  ✅ Signer is valid
  ? WARNING: 0x2d54EC30Be0D2583b0d781e10B31905c3c54616d has low balance on alfajores. At least 1.0 recommended but found 0.0 CELO Continue? yes
```

### Related issues

* Fixes: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3736

### Backward compatibility

* Yes

### Testing

* Manual


typescript/cli/src/deploy/utils.ts
```diff
@@ -76,13 +76,17 @@ export async function runPreflightChecksForChains({
...
     multiProvider,
     signer,
     chainsToGasCheck ?? chains,
     minGas,
   );
-  logGreen('✅ Balances are sufficient');
+  logGreen(
+    lowBalanceExists
+      ? '⚠️ Balances may not be sufficient'
+      : '✅ Balances are sufficient',
```
jmrossy, line 88: Is this redundant with the error assertGasBalances already shows?

typescript/cli/src/registry/MergedRegistry.ts
```diff
@@ -92,7 +92,7 @@ export class MergedRegistry extends BaseRegistry implements IRegistry {
   }): Promise<void> {
     return this.multiRegistryWrite(
       async (registry) => await registry.addChain(chain),
-      `adding chain ${chain.chainName}`,
+      `Adding chain ${chain.chainName}`,
```
jmrossy, line 95: The reason it was lower case is because it gets weaved into another log statement (see below). E.g. with this change you'll have 'Done Adding'

typescript/cli/src/deploy/utils.ts
nbayindirli: Somewhat, but I kinda prefer it. If you'd like, we could just omit logging anything if insufficient:
```
if (!lowBalanceExists) logGreen( '✅ Balances are sufficient');
```

typescript/cli/src/registry/MergedRegistry.ts
nbayindirli: Is it worth adding a little logic to update that first letter if it's appended or not?

Would just like to avoid the following logging:
```
✅ Balances are sufficient
All systems ready, captain! Beginning deployment...
Deploying ISM factory contracts
Deploying to alfajores from https://alfajores.celoscan.io/address/0x16F4898F47c085C41d7Cc6b1dc72B91EA617dcBb
Deploying to fuji from https://testnet.snowtrace.io/address/0x16F4898F47c085C41d7Cc6b1dc72B91EA617dcBb
skipping updating chain alfajores at github registry
updating chain alfajores at local registry at ./
done updating chain alfajores at local registry
skipping updating chain fuji at github registry
updating chain fuji at local registry at ./
done updating chain fuji at local registry
ISM factory contracts deployed
Deploying core contracts to alfajores, fuji
```

.changeset/olive-apricots-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
```
cmcewen, line 2: is this only a patch? theoretically if someone was using an address as the HYP_KEY before, this would break right?

typescript/cli/src/deploy/utils.ts
cmcewen: what action would someone take if they see balances may not be sufficient? feels like it would make me think i should do something but not sure what i should do

typescript/cli/src/registry/MergedRegistry.ts
cmcewen: this is a taste thing - but i think drive bys often work better as separate PRs because you might get comments that aren't related to the PR title or main body of work

typescript/cli/src/utils/balances.ts
```diff
@@ -8,9 +8,10 @@ export async function assertNativeBalances(
   signer: ethers.Signer,
   chains: ChainName[],
   minBalanceWei: string,
-) {
+): Promise<boolean> {
```
cmcewen, line 11: i think things named assert i expect to throw usually, vs this change makes the function now return a bool. and i have to think a bit whether lowBalanceExists means the balance is too low, or it's not - vs something like `balanceIsSufficient` or something like that i just find easier to reason (true - balance is sufficient, false -  balance is not enough)

typescript/cli/src/utils/keys.ts
```diff
@@ -36,6 +36,8 @@ export async function getImpersonatedSigner({
   skipConfirmation?: boolean;
 }) {
   key ||= await retrieveKey(IMPERSONATED_KEY_TYPE, skipConfirmation);
+  if (key.length != ETHEREUM_ADDRESS_LENGTH)
+    key = privateKeyToSigner(key).address;
```
cmcewen, line 40: don't know if this is your code initially - but it seems like we're overloading the key to be both a private key and an address here? again i like explicitness (privateKey vs address)

typescript/cli/src/deploy/utils.ts
yorhodes: consider fauceting

typescript/cli/src/registry/MergedRegistry.ts
jmrossy: By all means feel free to improve this! I agree it's slightly off-putting but I wouldn't stress over it :)

typescript/cli/src/deploy/utils.ts
nbayindirli: nice callout– we should advise an action. If you take a look at the Drive-by changes section, you can see the current output, which includes:
```
? WARNING: 0x2d54EC30Be0D2583b0d781e10B31905c3c54616d has low balance on alfajores. At least 1.0 recommended but found 0.0 CELO Continue? yes
```

as you've pointed out below, definitely out of scope for this PR, but as a user I'd most enjoy the option to fund my wallet, and re-check balances without having to exit the script (current result of entering `no`)

typescript/cli/src/registry/MergedRegistry.ts
nbayindirli: > this is a taste thing - but i think drive bys often work better as separate PRs because you might get comments that aren't related to the PR title or main body of work

agreed– this was a misuse of the "Drive-by changes" section for the PR. These were definitely drive-bys, but not solely related to the dry-run change. I think there is good use in keeping the Drive-by change section in our PR template, so long as those changes are tangentially related



> @nbayindirli By all means feel free to improve this! I agree it's slightly off-putting but I wouldn't stress over it :)

Will leave it be for now 👍

.changeset/olive-apricots-develop.md
nbayindirli: You're right– the HYP_KEY will no longer support user addresses after this change. Updating to `minor`

typescript/cli/src/utils/balances.ts
nbayindirli: Completely agree– changing the return and not refactoring the name was poor retrofitting.

Updated to `nativeBalancesAreSufficient` where `sufficient` represents if a low balance is found or not

typescript/cli/src/utils/keys.ts
nbayindirli: this PR is the result of that conversation– where **`HYP_KEY` can now only be used to source a private key**.

In addition, a developer may provide a private key or an address to `--key` when running a dry-run (@yorhodes and I decided this would be useful for those dry-running for an arbitrary address, or not wanting to provide their PK just yet)

jmrossy (approved):
Still not 100% sure we have the ideal setup here but this is an improvement so let's ship it!

typescript/cli/src/utils/keys.ts
yorhodes: I would suggest to go further with the separation
if a user wants to pass an address for dry run, they should pass an --address or --from flag
passing different types to the --key flag is still confusing

yorhodes (changes requested):
we should adjust both the env var interface and the command line interface to separate concerns

nbayindirli:
> we should adjust both the env var interface and the command line interface to separate concerns

strongly agreed– i'm going to keep our `HYP_KEY` & `--key` notation strictly for PKs and additionally support `--from` for user addresses

typescript/cli/src/deploy/utils.ts
nbayindirli: Decided on just logging '✅ Balances are sufficient' when sufficient to keep this change moving

typescript/cli/src/commands/deploy.ts
```diff
@@ -83,12 +85,15 @@ const coreCommand: CommandModuleWithWriteContext<{
...
     'dry-run': dryRunOption,
+    'from-address': fromAddressCommandOption,
   },
   handler: async ({ context, targets, ism, hook, agent, dryRun }) => {
     logGray(
       `Hyperlane permissionless core deployment${dryRun ? ' dry-run' : ''}`,
     );
-    logGray('------------------------------------------------');
+    logGray(
+      `----------------------------------------${dryRun ? '--------' : ''}`,
```
yorhodes, line 95: find this slightly odd

typescript/cli/src/utils/balances.ts
```diff
@@ -25,23 +26,32 @@ export async function assertNativeBalances(
           message: `WARNING: ${error} Continue?`,
         });
         if (!isResume) throw new Error(error);
+        sufficient = false;
```
yorhodes, line 29: does this violate the `isResume` prompt?

typescript/cli/src/context/context.ts
```diff
@@ -86,11 +93,12 @@ export async function getDryRunContext(
...
   await forkNetworkToMultiProvider(multiProvider, chain);
-  const { key: impersonatedKey, signer: impersonatedSigner } =
-    await getImpersonatedSigner({
-      key,
-      skipConfirmation,
-    });
+
+  const { impersonatedKey, impersonatedSigner } = await getImpersonatedSigner({
+    fromAddress,
+    key,
```
yorhodes, line 99: what if `fromAddress` and `key` are incompatible?

yorhodes (approved):
mostly lgtm
thanks for separating the interface

PR #3744, opened 2024-05-08
Default to home directory for local registry

yorhodes:

### DIFF
--- a/typescript/cli/src/commands/options.ts
+++ b/typescript/cli/src/commands/options.ts
@@ -29,7 +29,7 @@ export const registryUriCommandOption: Options = {
 export const overrideRegistryUriCommandOption: Options = {
   type: 'string',
   description: 'Path to a local registry to override the default registry',
-  default: './',
+  default: '~/.hyperlane',
 };
 
 export const skipConfirmationOption: Options = {


typescript/cli/src/commands/options.ts
```diff
@@ -29,7 +30,7 @@ export const registryUriCommandOption: Options = {
 export const overrideRegistryUriCommandOption: Options = {
   type: 'string',
   description: 'Path to a local registry to override the default registry',
-  default: './',
+  default: `${os.homedir()}/.hyperlane`,
```
jmrossy, line 33: I disagree about this change, especially since the folder will be hidden by default. This is where deployment artifacts will be written. We'd be putting them where most people won't find them.

cmcewen: i think this is a question of whether people will search for artifacts based only on CLI commands or look at the file system. i think defaulting here to this dot directory makes the UX more resilient to new terminal windows, different directories, and won't add to someone's git repo accidentally, but it's true you won't get the visual feedback to see things appear.

yorhodes: I also think this makes docs easier to write ie `~/.hyperlane`

yorhodes: > We'd be putting them where most people won't find them.

We also log where we are writing artifacts to in every deploy command

jmrossy (approved):

PR #3745, opened 2024-05-08
Derive more defaults in chain config

yorhodes:
### Description

- Attempt to detect RPC URL, chain name, and chain ID

### Testing

```sh
Creating a new chain config
? Detected rpc url as http://localhost:8545, is this correct? no
? Enter http or https rpc url http://localhost:8546
? Detected chain name as anvil8546, is this correct? yes
? Detected chain id as 31338, is this correct? yes
? Do you want to set block or gas properties for this chain config?(optional) (y/N)
```


typescript/cli/src/config/chain.ts
```diff
@@ -38,19 +39,27 @@ export async function createChainConfig({
...
+      message: 'Enter chain name (one word, lower case)',
+    });
+  }
+
   const metadata: ChainMetadata = {
     name,
-    chainId: parseInt(chainId, 10),
-    domainId: parseInt(domainId, 10),
+    chainId,
+    domainId: chainId,
```
nbayindirli, line 58: ❤️

typescript/cli/src/config/chain.ts
```diff
@@ -38,19 +39,27 @@ export async function createChainConfig({
...
+
   const rpcUrl = await input({ message: 'Enter http or https rpc url' });
+  const network = await new ethers.providers.JsonRpcProvider(
+    rpcUrl,
+  ).getNetwork();
+
+  let { chainId, name } = network;
+  if (name === 'unknown') {
+    name = await input({
+      message: 'Enter chain name (one word, lower case)',
```
nbayindirli, line 51: possible nit: does input auto-add a colon at the end here?

nbayindirli (approved):
lgtm

typescript/cli/src/config/chain.ts
yorhodes: idk but would prefer not to change this everywhere

jmrossy: Should we really rely on the name provided by the network given that we've invented our own naming type:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/metadata/customZodTypes.ts#L20

PR #3749, opened 2024-05-09
Allow relaying of a specific message

nambrot:
### Description

Allows a specific message to be relayed in infra


yorhodes:
@nambrot can we close this?

Issue #3695, opened 2024-05-02
Implement IsmCrudModule `update()`


paulbalaji:
follows-on from https://github.com/hyperlane-xyz/issues/issues/1152

Concretely:
- given an ISM config, read the current state and return a set of txs required to modify the current config to the target state

### Related CLI command
https://github.com/hyperlane-xyz/issues/issues/1188

```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3800
```

PR #3759, opened 2024-05-10
feat: implement `update()` for `EvmIsmModule`

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3695 https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3800

- [x] introduces ism update logic
- [x] port over ism factory tests to ism module


typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+    const signer = this.multiProvider.getSigner(destination);
+    const provider = this.multiProvider.getProvider(destination);
+    const owner = await DomainRoutingIsm__factory.connect(
+      existingIsmAddress,
+      provider,
+    ).owner();
+    const isOwner = eqAddress(await signer.getAddress(), owner);
+
+    // reconfiguring existing routing ISM
+    if (existingIsmAddress && isOwner && !delta.mailbox) {
```
jmrossy, line 207: Can we break up this big fn? Maybe by splitting off these if/else clause contents?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      deployer,
+    } = params;
+    const overrides = multiProvider.getTransactionOverrides(destination);
+    let routingIsm: DomainRoutingIsm | DefaultFallbackRoutingIsm;
+
+    // filtering out domains which are not part of the multiprovider
+    config.domains = objFilter(
+      config.domains,
+      (domain, config): config is IsmConfig => {
+        const domainId = multiProvider.tryGetDomainId(domain);
```
jmrossy, line 486: ditto

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+        try {
+          return factories.domainRoutingIsmFactory.interface.parseLog(log);
+        } catch (e) {
+          return undefined;
+        }
+      })
+      .filter(
+        (log): log is ethers.utils.LogDescription =>
+          !!log && log.name === 'ModuleDeployed',
+      );
```
jmrossy, line 591: I see you copied this line for line (including the TODO) from HyperlaneIsmFactory. Could we DRY this up or is the plan to kill HyperlaneIsmFactory soon anyway?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+  }): Promise<Address> {
+    const {
+      chain,
+      factory,
+      values,
+      logger,
+      threshold = values.length,
+      multiProvider,
+    } = params;
+    const sorted = [...values].sort();
```
jmrossy, line 664: Well done! I see people make the mistake of sorting the original array very often

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+          factories: this.args.addresses,
+          deployer: this.deployer,
+        });
+        isms[originDomain] = ism.address;
+        const tx = await routingIsm.set(
+          originDomain,
+          isms[originDomain],
+          overrides,
+        );
+        await this.multiProvider.handleTx(destination, tx);
```
ltyu, line 234: Should update transactions suppose to be handled here, or should they be returned to caller `update()`?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      }
+      // unenrolling domains if needed
+      for (const originDomain of delta.domainsToUnenroll) {
+        logger.debug(
+          `Unenrolling originDomain ${originDomain} from preexisting routing ISM at ${existingIsmAddress}...`,
+        );
+        const tx = await routingIsm.remove(originDomain, overrides);
+        await this.multiProvider.handleTx(destination, tx);
+      }
```
ltyu, line 243: ditto

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+  }): Promise<IRoutingIsm> {
+    const { destination, config, mailbox, existingIsmAddress, logger } = params;
+    const overrides = this.multiProvider.getTransactionOverrides(destination);
+    let routingIsm: DomainRoutingIsm | DefaultFallbackRoutingIsm;
+
+    // filtering out domains which are not part of the multiprovider
+    config.domains = objFilter(
+      config.domains,
+      (domain, config): config is IsmConfig => {
+        const domainId = this.multiProvider.tryGetDomainId(domain);
```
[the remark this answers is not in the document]
paulbalaji: that's the one!

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +150,541 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      destination,
+      config,
+      origin,
+      mailbox,
+      multiProvider,
+      factories,
+      deployer,
+    } = params;
+    if (typeof config === 'string') {
+      // @ts-ignore
```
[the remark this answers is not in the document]
paulbalaji: honestly i'm not too sure - but can be removed as part of tidying up the Config type in this module to exclude `Address`

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: i thought about DRYing it up but as you guessed, yeah I want to get to a state where we can nuke the factory/deployers in favour of these modules

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -38,44 +75,71 @@ export class EvmIsmModule extends HyperlaneModule<
...
-    await this.creator.update({
-      destination,
-      config,
-      existingIsmAddress: this.args.addresses.deployedIsm,
-    });
+
+    if (typeof config === 'string') {
+      this.logger.debug('Skipping update for config of type Address.');
+      return [];
+    }
```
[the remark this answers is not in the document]
paulbalaji: spent 15mins looking into this and it may conflict with other modules, so I'll wait til we have ism+core+hook+warp merged before PR'ing this

Created issue to track https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3773

jmrossy (dismissed):

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,81 +1,181 @@
...
     deployer: HyperlaneDeployer<any, any>;
-    factories: HyperlaneContracts<ProxyFactoryFactories>;
+    factories: HyperlaneAddresses<ProxyFactoryFactories>;
     multiProvider: MultiProvider;
   }): Promise<EvmIsmModule> {
+    const { chain, config, deployer, factories, multiProvider } = params;
     const destination = multiProvider.getChainName(chain);
-    const ismCreator = new EvmIsmCreator(deployer, multiProvider, factories);
-    const deployedIsm = await ismCreator.deploy({
+    const deployedIsm = await EvmIsmModule.deploy({
```
yorhodes, line 173: why have a `deploy` separate from `create`?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
       config,
     });
   }
+
+  protected async updateRoutingIsm(params: {
+    destination: ChainName;
+    config: RoutingIsmConfig;
+    origin?: ChainName;
+    mailbox?: Address;
+    existingIsmAddress: Address;
```
yorhodes, line 195: imo these params dont really make sense here
what do origin or mailbox mean in this context?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+    logger: Logger;
+  }): Promise<IRoutingIsm> {
+    const { destination, config, mailbox, existingIsmAddress, logger } = params;
+
+    // 1. Filter for known domains
+    const { domains, safeConfigDomains } = EvmIsmModule.filterRoutingIsmDomains(
+      { config, multiProvider: this.multiProvider },
+    );
+    config.domains = domains;
```
yorhodes, line 204: imo in future we should adapt the configs themselves to accommodate non-evm keys

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      mailbox,
+    );
+
+    // Step 3: Get owner and signer
+    const signer = this.multiProvider.getSigner(destination);
+    const provider = this.multiProvider.getProvider(destination);
+    const owner = await DomainRoutingIsm__factory.connect(
+      existingIsmAddress,
+      provider,
+    ).owner();
```
yorhodes, line 222: would expect contracts to already be connected when the module.update is called

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+    const isOwner = eqAddress(await signer.getAddress(), owner);
```
yorhodes, line 223: I think we should remove this branching entirely
the update tx is the same whether or not we can submit it here

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+
+    if (existingIsmAddress && isOwner && !delta.mailbox) {
+      // 4. If possible, reconfigure existing ISM
+      return await this.reconfigureExistingRoutingIsm({
```
yorhodes, line 227: nit: unnecessary await

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+        destination,
+        config.owner,
+      );
+
+      if (config.type === IsmType.FALLBACK_ROUTING) {
+        if (!mailbox) {
+          throw new Error(
+            'Mailbox address is required for deploying fallback routing ISM',
+          );
+        }
```
yorhodes, line 254: shouldn't fallback routing config have a mailbox address on it?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+
+        // 6. If fallback routing ISM, reinitialise in-place with new submodules
+        return await this.initializeFallbackRoutingIsm({
```
yorhodes, line 257: nit: unnecessary await

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      );
+
+      if (config.type === IsmType.FALLBACK_ROUTING) {
+        if (!mailbox) {
+          throw new Error(
+            'Mailbox address is required for deploying fallback routing ISM',
+          );
+        }
+
+        // 6. If fallback routing ISM, reinitialise in-place with new submodules
```
yorhodes, line 256: nit: reinitialize
also dont prefer "initialize" nomenclature in "update"

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+          owner,
+          safeConfigDomains,
+          submoduleAddresses,
+          existingIsmAddress,
+          signer,
+          logger,
+        });
+      } else {
+        // 7. Else, deploy a new domain routing ISM
+        return await EvmIsmModule.deployDomainRoutingIsm({
```
yorhodes, line 268: nit: unnecessary await

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+          destination,
+          owner,
+          safeConfigDomains,
+          submoduleAddresses,
+          existingIsmAddress,
+          signer,
+          logger,
+        });
+      } else {
+        // 7. Else, deploy a new domain routing ISM
```
yorhodes, line 267: nit: numbering is not actually consecutive
7 runs even if 6 doesnt

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+          submoduleAddresses,
+          overrides: this.multiProvider.getTransactionOverrides(destination),
+          multiProvider: this.multiProvider,
+          factories: this.args.addresses,
+        });
+      }
+    }
+  }
+
+  protected async reconfigureExistingRoutingIsm(params: {
```
yorhodes, line 281: how is this different from update?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+        destination,
+        config: config.domains[origin],
+        origin,
+        mailbox,
+        multiProvider: this.multiProvider,
+        factories: this.args.addresses,
+        deployer: this.deployer,
+      });
+      const tx = await routingIsm.set(originDomain, ism.address, overrides);
+      await this.multiProvider.handleTx(destination, tx);
```
yorhodes, line 314: I thought we would always return this from the update fn

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      );
+      const tx = await routingIsm.remove(originDomain, overrides);
+      await this.multiProvider.handleTx(destination, tx);
+    }
+
+    // Transfer ownership if needed
+    if (delta.owner) {
+      logger.debug(`Transferring ownership of routing ISM...`);
+      const tx = await routingIsm.transferOwnership(delta.owner, overrides);
+      await this.multiProvider.handleTx(destination, tx);
```
yorhodes, line 330: ditto for all these as update txs?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+
+    for (const origin of Object.keys(config.domains)) {
+      const ism = await EvmIsmModule.deploy({
+        destination,
+        config: config.domains[origin],
+        origin,
+        mailbox,
+        multiProvider: this.multiProvider,
+        factories: this.args.addresses,
+        deployer: this.deployer,
```
yorhodes, line 389: seems like we have to pass a lot of context through
should we make the deploy fn non-static?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+    const signer = multiProvider.getSigner(destination);
+    const multisigIsmFactory =
+      config.type === IsmType.MERKLE_ROOT_MULTISIG
+        ? factories.staticMerkleRootMultisigIsmFactory
+        : factories.staticMessageIdMultisigIsmFactory;
+    const factory: StaticThresholdAddressSetFactory =
+      StaticThresholdAddressSetFactory__factory.connect(
+        multisigIsmFactory,
+        signer,
+      );
```
yorhodes, line 536: can we leverage the connected factories if this is not static?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+        destination,
+        config,
+        mailbox,
+      });
+
+      const owner = await resolveOrDeployAccountOwner(
+        this.multiProvider,
+        destination,
+        config.owner,
+      );
```
yorhodes, line 247: hoping to remove this ICA resolution from these abstractions
we can resolve the local owner address at a higher level

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      `Deploying ${ismType} to ${destination} ${
+        origin ? `(for verifying ${origin})` : ''
+      }`,
+    );
+
+    let contract: DeployedIsmType[typeof ismType];
+    switch (ismType) {
+      case IsmType.MESSAGE_ID_MULTISIG:
+      case IsmType.MERKLE_ROOT_MULTISIG:
+        contract = await EvmIsmModule.deployMultisigIsm({
```
yorhodes, line 436: any reason to assign to contract as opposed to just returning here?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+
+    const submoduleAddresses = Object.values(isms);
+
+    if (config.type === IsmType.FALLBACK_ROUTING) {
+      // deploying new fallback routing ISM
+      if (!mailbox) {
+        throw new Error(
+          'Mailbox address is required for deploying fallback routing ISM',
+        );
+      }
```
yorhodes, line 604: lets just have this on the config or require along with the ism factories

yorhodes (dismissed):
nice start, I think we need to remove some more `isOwner` branching

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: the `deploy` is an implementation detail of `create` that can get reused in `update` - would keep this split here to distinguish between returning a full module and returning just an individual deployed ism (which may be top-level or a submodule of some other ism)

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: this is unchanged from ismfactory - but origin can be removed and I can move mailbox into a required property of the evmismmodule along with the factories

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,81 +1,181 @@
...
+        StaticAggregationHookFactory__factory.connect(
+          args.addresses.staticAggregationHookFactory,
+          signer,
+        ),
+      domainRoutingIsmFactory: DomainRoutingIsmFactory__factory.connect(
+        args.addresses.domainRoutingIsmFactory,
+        signer,
+      ),
+    };
   }
```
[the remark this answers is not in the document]
paulbalaji: kinda but needed to do something like this instead

```ts
    this.factories = connectContracts<ProxyFactoryFactories>(
      attachContracts(this.args.addresses, proxyFactoryFactories),
      multiProvider.getSigner(args.chain),
    );
```

made a new helper for this

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: I didn't want each function to be super long so we have update() -> if condition: updateRoutingIsm() -> if condition: reconfigureExistingRoutingIsm() 

strictly speaking the logic in `updateRoutingIsm` could go into `update` and `reconfigureExistingRoutingIsm` should be renamed to `updateRoutingIsm` - but then this adds

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: originally because there was some more work after the switch but now it's all separate fns can just return

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+    const {
+      destination,
+      config,
+      origin,
+      mailbox,
+      logger,
+      multiProvider,
+      factories,
+      deployer,
+    } = params;
```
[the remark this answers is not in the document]
paulbalaji: what do you mean?

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: does that mean we should move the `transferOwnershipOfContracts` in here to update() as well?
image

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: yup

but then it's a question of how we want the module interaction to flow if Module.create() is no longer static

paulbalaji: yes - but then the flow for module users would be 

```
const module = Module.create()
module.deploy()
module.create()
module.update()
...
```

maybe the actual interface we desire is not create/read/update but deploy/read/update, and we do Module.new() to instantiate a class for an undeployed module?

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: since we're not getting the calldata from the interface - no longer need to connect contracts for an update

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      await this.multiProvider.handleTx(destination, tx);
+    }
+
+    return routingIsm;
+  }
+
+  protected async initializeFallbackRoutingIsm(params: {
+    destination: ChainName;
+    owner: Address;
+    safeConfigDomains: Domain[];
```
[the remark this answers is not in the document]
paulbalaji: honestly i'm not too sure - it's a relic/unchanged from the ismfactory

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+        `Deploying new ${threshold} of ${values.length} address set to ${chain}`,
+      );
+      const overrides = multiProvider.getTransactionOverrides(chain);
+      const hash = await factory['deploy(address[],uint8)'](
+        sorted,
+        threshold,
+        overrides,
+      );
+      await multiProvider.handleTx(chain, hash);
+      // TODO: add proxy verification artifact?
```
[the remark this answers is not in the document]
paulbalaji: could do something like this but not sure how to get the constructor arguments - if you know a quick way then I'll do it but otherwise don't want to spend too much time on this
image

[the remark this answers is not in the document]
paulbalaji: could also include verifier in the main class, and have the non-static callers of deployStaticAddressSet call the verifyContract method

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,700 @@
...
+
+    this.chainName = this.multiProvider.getChainName(this.args.chain);
+    this.domainId = this.multiProvider.getDomainId(this.chainName);
+
+    this.verifier = new ContractVerifier(
+      this.multiProvider,
+      {},
+      coreBuildArtifact,
+      ExplorerLicenseType.MIT,
+    );
```
paulbalaji, line 120: TODO: just grab the verifier from the deployer

typescript/sdk/src/ism/EvmIsmModule.ts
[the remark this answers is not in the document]
paulbalaji: the more i look into it, the less i'm sure this is even needed

it's filtering out the origin domains for a routing ism if it's not specified in the multiprovider, which seems not right? surely we want to configure it regardless if it's in the multiprovider or not on the destination chain that we are trying to deploy to?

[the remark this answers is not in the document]
yorhodes: indeed, this is probably some tech debt
I agree with your assessment that we do want to configure it regardless
I think the problem is maybe this unknown chain name/domain ID stuff

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -146,6 +146,15 @@ export function attachContractsMapAndGetForeignDeployments<
...
 }
 
+export function attachAndConnectContracts<F extends HyperlaneFactories>(
+  addresses: HyperlaneAddresses<F>,
+  factories: F,
+  connection: Connection,
+): HyperlaneContracts<F> {
+  const contracts = attachContracts(addresses, factories);
+  return connectContracts(contracts, connection);
+}
```
paulbalaji, line 156: saw your comment on a commit and moving into the PR @yorhodes 
> also how much does this function really save us in verbosity?

I like having it to reduce having an extra "const x = BLAH" before assigning to factories
image

typescript/sdk/src/deploy/types.ts
```diff
@@ -27,6 +27,13 @@ export type OwnableConfig<Keys extends PropertyKey = PropertyKey> = z.infer<
   ownerOverrides?: Partial<Record<Keys, Address>>;
 };
 
+export function extractOwnerAddress(config: Owner): Address {
+  if (typeof config === 'string') {
+    return config;
+  }
+  return config.owner;
+}
```
paulbalaji, line 35: this is how I'm extracting an owner address from an `Owner` from an `IsmConfig` - trusting that the caller of the ISM module has done the ICA stuff if they need to on their end

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -203,6 +203,23 @@ export interface EthersV5Transaction
...
+  return {
+    annotation: params.annotation,
+    type: ProviderType.EthersV5,
+    transaction: {
+      chainId: params.chainId,
+      to: params.to,
+      data: params.data,
+    },
+  };
+}
```
paulbalaji, line 221: > im surprised this type is not an extension of EthersV5Transaction

do you mean something like this?
```ts
export interface AnnotatedEthersV5Transaction extends EthersV5Transaction {
  annotation: string;
}
```

basically just syntactic sugar I guess, alternatively could add something like this
```ts
export type AnnotatedEthersV5Transaction = Annotated<EthersV5Transaction>;
```

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,676 @@
...
+    }
+
+    const delta = await routingModuleDelta(
+      this.chainName,
+      this.args.addresses.deployedIsm,
+      targetConfig,
+      this.multiProvider,
+      this.factories,
+      this.args.addresses.mailbox,
+    );
```
paulbalaji, line 234: i've tried to remove as much ICA logic as I can but this helper method does make a call to `resolveOrDeployAccountOwner`

how far do we go in re-ICAing stuff like this?

pretty much every ism utils function requires doing the `resolveOrDeployAccountOwner` thing

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,690 @@
...
-      config,
-      existingIsmAddress: this.args.addresses.deployedIsm,
+    const currentConfig = await this.read();
+    const configMatches = await moduleMatchesConfig(
+      this.chainName,
+      this.args.addresses.deployedIsm,
+      currentConfig,
+      this.multiProvider,
+      this.factories,
+    );
```
paulbalaji, line 131: this also calls `resolveOrDeployAccountOwner` under the hood

typescript/sdk/src/ism/EvmIsmModule.ts
yorhodes: we can punt the full migration to later but its unnecessary complexity everywhere

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -86,4 +186,610 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      .filter(
+        (log): log is ethers.utils.LogDescription =>
+          !!log && log.name === 'ModuleDeployed',
+      );
+
+    if (dispatchLogs.length === 0) {
+      throw new Error('No ModuleDeployed event found');
+    }
+
+    const moduleAddress = dispatchLogs[0].args['module'];
```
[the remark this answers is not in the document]
yorhodes: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/20eccf3d21772ba4986487dac6d904a62503e21b/typescript/sdk/src/core/HyperlaneCore.ts#L232 is quite similar
maybe we can have a `findMatchingLog` util or something

typescript/sdk/src/deploy/types.ts
yorhodes: I removed this from the "runtime" config in my PR as well https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3820
imo we shouldnt pollute all of the deployers with ICA resolution code

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+  }): Promise<boolean> {
+    return moduleMatchesConfig(
+      chain,
+      ism.serialize().deployedIsm,
+      config,
+      multiProvider,
+      factoryContracts,
+      mailbox,
+      configured,
+    );
```
yorhodes, line 186: I dont understand why we are using this in these tests

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+  }
+
+  async function logAndSendTransactions(txs: AnnotatedEthersV5Transaction[]) {
+    for (const tx of txs) {
+      if (tx.annotation) {
+        console.log(tx.annotation);
+      }
+      await multiProvider.sendTransaction(chain, tx.transaction);
+    }
+  }
```
yorhodes, line 208: if Annotated extended the base tx type, you could move this logging and type into the multi provider in a backward compatible manner

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+          console.error('Failed to match random ism config', e);
+          console.error('EXPECTED');
+          console.error(stringifyObject(config as object, 'yaml', 2));
+          console.error('ACTUAL');
+          console.error(
+            stringifyObject((await ism.read()) as object, 'yaml', 2),
+          );
+          process.exit(1);
+        }
+      });
```
yorhodes, line 284: the catch stuff seems useful for debugging these tests but not necessary in the published tests

yorhodes: we didnt really discuss this but I expected these tests to basically be
```typescript
const ismConfig = randomIsmConfig();
const ismModule = await createIsmModule({config});
const derivedConfig = await ismModule.read();
expect(derivedConfig).to.deep.equal(ismConfig);
```

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -203,6 +203,25 @@ export interface EthersV5Transaction
...
```
yorhodes, line 223: if this is unused, please remove

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+          type: IsmType.MESSAGE_ID_MULTISIG,
+          threshold: 1,
+          validators: [randomAddress()],
+        };
+
+        // txs required to add test5 domain
+        const updateTxs = await ism.update(exampleRoutingConfig);
+
+        // adding test5 domain is no-op, so no txs should be returned
+        expect(updateTxs.length).to.equal(0);
```
yorhodes, line 324: why is it no-op?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+        // apply the config to the new ISM
+        const applyTxs = await ism.update(exampleRoutingConfig);
+        await logAndSendTransactions(applyTxs);
+
+        // add config for a domain the multiprovider doesn't have
+        exampleRoutingConfig.domains['test5'] = {
+          type: IsmType.MESSAGE_ID_MULTISIG,
+          threshold: 1,
+          validators: [randomAddress()],
+        };
```
yorhodes, line 318: imo we should still deploy ISMs even if the multi provider doesnt have the domain?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
yorhodes, line 1: seems like a lot of the update test cases share a similar pattern/template
can we dedupe a bit?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+        const matches = await ismModuleMatchesConfig({
+          ism,
+          config: exampleRoutingConfig,
+          configured: false,
+        });
+        expect(matches).to.be.true;
+
+        // apply the config to the new ISM
+        const applyTxs = await ism.update(exampleRoutingConfig);
+        await logAndSendTransactions(applyTxs);
```
yorhodes, line 311: why would this do anything if we just `create`d with this config?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+          configured: false,
+        });
+        expect(matches).to.be.true;
+
+        // initial ISM address
+        const initialIsmAddress = ism.serialize().deployedIsm;
+
+        // apply the config to the new ISM
+        const applyTxs = await ism.update(exampleRoutingConfig);
+        await logAndSendTransactions(applyTxs);
```
yorhodes, line 346: ditto, why would this do anything? what am I missing

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+  MultisigIsmConfig,
+  RoutingIsmConfig,
+  RoutingIsmDelta,
+} from './types.js';
+import { moduleMatchesConfig, routingModuleDelta } from './utils.js';
+
+type ExtraArgs = {
+  deployedIsm: Address;
+  mailbox: Address;
+};
```
yorhodes, line 70: is `deployedIsm` similar to `WithAddress<IsmConfig>`
can we dedupe these interfaces

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+  DeployedIsmType,
+  IsmConfig,
+  IsmType,
+  MultisigIsmConfig,
+  RoutingIsmConfig,
+  RoutingIsmDelta,
+} from './types.js';
+import { moduleMatchesConfig, routingModuleDelta } from './utils.js';
+
+type ExtraArgs = {
```
yorhodes, line 67: dont love the name "ExtraArgs"

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
-      }
+      HyperlaneAddresses<ProxyFactoryFactories> & ExtraArgs
     >,
   ) {
     super(args);
+
     this.reader = new EvmIsmReader(multiProvider, args.chain);
-    this.creator = new EvmIsmCreator(deployer, multiProvider, args.addresses);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { mailbox, deployedIsm, ...addresses } = args.addresses;
```
yorhodes, line 99:
```suggestion
    const { mailbox, deployedIsm, ..._addresses } = args.addresses;
```
does this work? I think it can be unused if you prefix with `_`

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
   }
 
   public async read(): Promise<IsmConfig> {
-    return await this.reader.deriveIsmConfig(this.args.addresses.deployedIsm);
+    return this.reader.deriveIsmConfig(this.args.addresses.deployedIsm);
   }
 
-  public async update(config: IsmConfig): Promise<EthersV5Transaction[]> {
-    throw new Error('Method not implemented.');
+  // whoever calls update() needs to ensure that targetConfig has a valid owner
```
yorhodes, line 114: what does "valid owner" mean here?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+  public async update(
+    targetConfig: IsmConfig,
+  ): Promise<AnnotatedEthersV5Transaction[]> {
+    const currentConfig =
+      typeof this.args.config === 'string'
+        ? this.args.addresses.deployedIsm
+        : await this.read();
```
yorhodes, line 121: imo `this.read` should handle these branches

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
 
-    const destination = this.multiProvider.getChainName(this.args.chain);
-    await this.creator.update({
-      destination,
-      config,
-      existingIsmAddress: this.args.addresses.deployedIsm,
+    // Update the config in case it's a custom ISM
+    this.args.config = targetConfig;
+
+    // moduleMatchesConfig expects any domain filtering to have been done already
```
yorhodes, line 126: we shouldnt augment our tests to accommodate this legacy function's behavior

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+        targetConfig.type === IsmType.FALLBACK_ROUTING)
+    ) {
+      // filter for known domains
+      const { availableDomains } = this.filterRoutingIsmDomains({
+        config: targetConfig,
+      });
+      targetConfig.domains = availableDomains;
+    }
+
+    // this calls resolveOrDeployAccountOwner under the hood
```
yorhodes, line 139: I know this isnt your code but was never a fan of this

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+    // If configs match, no updates needed
+    if (configMatches) {
+      return [];
+    }
+
+    // Else, we have to figure out what an update for this ISM entails
+
+    // If target config is a custom ISM, just update the address
+    // if config -> custom ISM, update address
+    // if custom ISM -> custom ISM, update address
```
yorhodes, line 158: I dont understand these comments

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+      // if config -> config, AND types are different, do a new deploy
+      currentConfig.type !== targetConfig.type ||
+      // unless the new ISM is a Routing or Pausable ISM, do a new deploy
+      (targetConfig.type !== IsmType.ROUTING &&
+        targetConfig.type !== IsmType.FALLBACK_ROUTING &&
+        targetConfig.type !== IsmType.PAUSABLE)
+    ) {
+      const contract = await this.deploy({
+        config: targetConfig,
+      });
```
yorhodes, line 178: I would have expected `this.deploy` to handle reuse of addresses on the module

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+          annotation: 'Transferring ownership of pausable ISM...',
+          chainId: this.domainId,
+          to: this.args.addresses.deployedIsm,
+          data: Ownable__factory.createInterface().encodeFunctionData(
+            'transferOwnership(address)',
+            [targetOwner],
+          ),
+        });
+        return [tx];
+      }
```
yorhodes, line 215: can we consolidate owner checking for all Ownable ISMs?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
         ...factories,
-        deployedIsm: deployedIsm.address,
+        mailbox,
+        deployedIsm: ethers.constants.AddressZero,
       },
       chain,
       config,
     });
+    const deployedIsm = await module.deploy({ config });
+    module.args.addresses.deployedIsm = deployedIsm.address;
```
yorhodes, line 282: I find this slightly odd
why wouldn't module.deploy assign to this?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+      const tx = createAnnotatedEthersV5Transaction({
+        annotation: `Transferring ownership of routing ISM...`,
+        chainId: this.domainId,
+        to: deployedIsmAddress,
+        data: routingIsmInterface.encodeFunctionData(
+          'transferOwnership(address)',
+          [delta.owner],
+        ),
+      });
+      updateTxs.push(tx);
```
yorhodes, line 348: would be nice for this not to be specific to routing ism branch

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,89 +1,727 @@
...
+      ),
+    });
+    return [updateTx];
+  }
+
+  protected async deployRoutingIsmSubmodules({
+    config,
+  }: {
+    config: RoutingIsmConfig;
+  }): Promise<Address[]> {
```
yorhodes, line 382: why wouldnt this return an object? similar to the `WithAddress<RoutingIsmConfig>` type

yorhodes (commented):
leaving an early review to get you some feedback
will have to do another pass

typescript/sdk/src/consts/testChains.ts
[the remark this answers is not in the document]
paulbalaji: it was something I grabbed from @ltyu's warp module PR to be able to apply the update TXs in the tests: https://canary.discord.com/channels/935678348330434570/1240004125597372578/1242157611281944679

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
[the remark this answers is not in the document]
paulbalaji: not enough is duped to the ism factory to be worth the time to dedupe, though the test cases have overlap the implementations of the tests are different enough

paulbalaji: sure wtal

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+describe('EvmIsmModule', async () => {
+  let multiProvider: MultiProvider;
+  let ismFactoryDeployer: HyperlaneProxyFactoryDeployer;
+  let exampleRoutingConfig: RoutingIsmConfig;
+  let mailboxAddress: Address;
+  let newMailboxAddress: Address;
+
+  let fundingAccount: SignerWithAddress;
+
+  const chain = TestChainName.test4;
```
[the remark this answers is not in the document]
paulbalaji: see prev comments

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: usage of this fn and adding the `configured` arg are all downstream of this related discussion https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3759/files#r1609006616

assuming create() does the deploy()+update() then we can replace `moduleMatchesConfig` with deep equals comparator

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: good shout

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: relates back to these two
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3759/files#r1608999641
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3759/files#r1609006616

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: see prev deploy VS deploy+update comments

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: see prev deploy VS deploy+update comments

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: this is just an address, i was actually considering removing the config arg from the abstract module

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji:
```suggestion
    const { mailbox: _, deployedIsm: __, ...addresses } = args.addresses;
```
no I'd have to do something like this instead, because `_mailbox` and `_deployedIsm` are not properties of `args.addresses`

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: just means this module is not responsible for verifying if the owner in the config actually exists or ICA been setup

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: good catch

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: custom ISM is when the given config is just a string

if we're going from a string -> config, need to a full deploy

if we're going config -> config, do the update logic

if we're going config -> string or string -> string, no need to do any further logic besides updating the address of the ism in state

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: not sure what you mean by this

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: you mean `WithAddress<RoutingIsmConfig>[]` instead?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: i suppose we could, makes the ISM part of this https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3836 easier

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: scratch the above - the main reason is because we should be able to reuse the .deploy function to deploy submodules without overwriting the top-level ism address

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: copied from warp module pr

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
paulbalaji: had a quick attempt at this but the problem is multiprovider accepts either a PopulatedTx or a Promise<PopulatedTx> and i'm not sure we want to await just to grab an annotation out for logging

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: make this into a "mutable ism" subset

typescript/sdk/src/providers/ProviderType.ts
ltyu: maybe referring to the lines below?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -0,0 +1,720 @@
...
+  async function impersonateAccount(account: Address): Promise<MultiProvider> {
+    await hre.ethers.provider.send('hardhat_impersonateAccount', [account]);
+    await fundingAccount.sendTransaction({
+      to: account,
+      value: hre.ethers.utils.parseEther('1.0'),
+    });
+    return MultiProvider.createTestMultiProvider({
+      signer: hre.ethers.provider.getSigner(account),
+    });
+  }
```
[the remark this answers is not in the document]
paulbalaji: from our sync: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/bd6ca01fd4dd4c6d8491c36b633308829008c8f9/typescript/sdk/src/utils/fork.ts#L70

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: renamed to `IsmModuleAddresses`

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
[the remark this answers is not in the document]
paulbalaji: existing one is for anvil and doesn't quite work in our hh test, leaving for now and will extract out when required in future test

ltyu:
do we need a changeset :)

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+    // const verificationInput = buildVerificationInput(
+    //   proxyFactoryImplementations[factoryName],
+    //   contract.address,
+    //   constructorArgs,
+    // );
+    // await this.deployer.verifyContract(
+    //   this.chainName,
+    //   verificationInput,
+    //   logger,
+    // );
```
yorhodes, line 466: I think you can pull in the verify logic to the `deployStaticAddressSet` helper

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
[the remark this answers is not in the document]
yorhodes: anvil aliases the hardhat RPCs so should be compatible fwiw

typescript/sdk/src/ism/EvmIsmModule.ts
yorhodes: interesting, okay

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+  HyperlaneAddresses<ProxyFactoryFactories> & IsmModuleAddresses
 > {
-  protected logger = rootLogger.child({ module: 'EvmIsmModule' });
-  protected reader: EvmIsmReader;
+  protected readonly logger = rootLogger.child({ module: 'EvmIsmModule' });
+  protected readonly reader: EvmIsmReader;
+  protected readonly factories: HyperlaneContracts<ProxyFactoryFactories>;
+
+  // Adding these to reduce how often we need to grab from MultiProvider.
+  public readonly chainName: string;
```
yorhodes, line 83: nit: use `chain: ChainName` to be consistent?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+  // We use domainId here because MultiProvider.getDomainId() will always
+  // return a number, and EVM the domainId and chainId are the same.
+  public readonly domainId: number;
 
   protected constructor(
     protected readonly multiProvider: MultiProvider,
     protected readonly deployer: HyperlaneDeployer<any, any>,
     args: HyperlaneModuleArgs<
```
yorhodes, line 91: nit: these seem more like params than args

typescript/utils/src/objects.ts
```diff
@@ -136,3 +136,82 @@ export function stringifyObject(
...
+    }
+    return newObj;
+  } else if (typeof obj === 'string') {
+    return obj.toLowerCase();
+  }
+
+  return obj;
+}
+
+export function configDeepEquals(v1: any, v2: any): boolean {
```
yorhodes, line 159: we should use https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
   }
 
   public async read(): Promise<IsmConfig> {
-    return await this.reader.deriveIsmConfig(this.args.addresses.deployedIsm);
+    const config =
+      typeof this.args.config === 'string'
+        ? this.args.addresses.deployedIsm
+        : await this.reader.deriveIsmConfig(this.args.addresses.deployedIsm);
+
+    return normalizeConfig(config);
```
yorhodes, line 116: don't we want the addresses here?

typescript/sdk/src/ism/EvmIsmModule.ts
yorhodes: can you replace "custom ISM" with "address ISM" then?

typescript/sdk/src/utils/logUtils.ts
yorhodes, line 1: 🙏

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+        config: targetConfig,
+      });
+
+      this.args.addresses.deployedIsm = contract.address;
+      return [];
+    }
+
+    // Explicitly narrow the type of targetConfig to RoutingIsmConfig or PausableIsmConfig
+    // This is safe because we've already checked that the type is in MutableIsmType
+    targetConfig = targetConfig as RoutingIsmConfig | PausableIsmConfig;
```
yorhodes, line 177: instead of casting, use type assertions

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+
+    // At this point, only the 3 ownable/mutable ISM types remain: PAUSABLE, ROUTING, FALLBACK_ROUTING
```
yorhodes, line 179: prefer switch case or add an else throw in the event that we add more types to this

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+          logger,
+        });
+
+        this.args.addresses.deployedIsm = newIsm.address;
+      }
+    }
+
+    // if it's either of the routing ISMs, update their submodules
+    let updateTxs: AnnotatedEthersV5Transaction[] = [];
+    if (targetConfig.type !== IsmType.PAUSABLE) {
```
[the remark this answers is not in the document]
yorhodes: prefer checking the positive case

typescript/sdk/src/ism/types.ts
```diff
@@ -39,6 +39,13 @@ export enum IsmType {
   TRUSTED_RELAYER = 'trustedRelayerIsm',
 }
 
+// ISM types that can be updated in-place
+export const MutableIsmType = [
```
yorhodes, line 43: nit: use CAP_CASE
https://github.com/airbnb/javascript#naming--uppercase

typescript/sdk/src/ism/utils.ts
```diff
@@ -38,6 +39,46 @@ import {
 
 const logger = rootLogger.child({ module: 'IsmUtils' });
 
+// Determines the domains to enroll and unenroll to update the current ISM config
+// to match the target ISM config.
+export function calculateDomainRoutingDelta(
```
yorhodes, line 44: if you use `Sets` you can leverage these
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/465112db6fddb3b598d6da39c13491ff1bc7e700/typescript/utils/src/sets.ts to greatly simplify this functino

typescript/sdk/src/ism/utils.ts
```diff
@@ -170,11 +211,19 @@ export async function moduleMatchesConfig(
   multiProvider: MultiProvider,
   contracts: HyperlaneContracts<ProxyFactoryFactories>,
   mailbox?: Address,
+  // Less stringent checks on Routing/Pausable ISMs if not configured yet
+  configured = true,
```
yorhodes, line 215: I don't think we should be modifying this function, nor using it in your code

yorhodes (approved):
mostly looks good
I think we can simplify the normalization and deep equals a bit

typescript/sdk/src/ism/utils.ts
paulbalaji: nice catch, leftover from previous iterations - I'm not using this fn anymore

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: also updated `domainId` below to `Domain`

typescript/utils/src/objects.ts
paulbalaji: NICEEEEEE

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -1,59 +1,691 @@
...
+    }
+
+    // Else, we have to figure out what an update for this ISM entails
+
+    // If target config is a custom ISM, just update the address
+    // if config -> custom ISM, update address
+    // if custom ISM -> custom ISM, update address
+    if (typeof targetConfig === 'string') {
+      // TODO: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3773
+      this.args.addresses.deployedIsm = targetConfig;
```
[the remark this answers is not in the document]
paulbalaji: the way i see it, the ism module is always associated with some onchain address

for fresh contracts, we go through the static create() which uses the zeroAddress until it gets deployed/initialised

for any existing contracts, we can skip the create() and instantiate a new EvmIsmModule from an existing address - and get all the read/update/redeploy logic benefits there too

i considered making it an optional field in `addresses` so it doesn't have to be set to zero in the create(), however I preferred having the strict expectation that this value is set in all the other functions

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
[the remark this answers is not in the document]
paulbalaji: I did try using the existing ones and even writing a couple of `hardhat_XYZ` helpers, but I think the issue is that because it's hh unit testing, you have to go via `hre.ethers.provider`

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: leaving a comment to verify in the helper

typescript/sdk/src/ism/utils.ts
paulbalaji: I'm not totally sure - we'd have to do a deep comparison of configs per domain as well and Sets do a dumb comparison instead of the deep equals we'd want

times like this i wish everyone used C# so we could make everything structs and implement comparators and reuse libs for free :P

typescript/sdk/src/ism/EvmIsmModule.ts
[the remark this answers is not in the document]
yorhodes: I also like it not being optional

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -384,12 +385,16 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
-    const txReq = await this.prepareTx(chainNameOrId, await tx);
+    const { annotation, ...tx } = await txProm;
+    if (annotation) {
+      this.logger.info(annotation);
+    }
+    const txReq = await this.prepareTx(chainNameOrId, tx);
     const signer = this.getSigner(chainNameOrId);
     const response = await signer.sendTransaction(txReq);
-    this.logger.info(`Sent tx ${response.hash}`);
+    this.logger.info(`Sent tx ${response.hash} + ${annotation}`);
```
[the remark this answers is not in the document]
paulbalaji: was playing around with how to log it

typescript/utils/src/objects.ts
```diff
@@ -154,3 +155,32 @@ export function stringifyObject(
   }
   return yamlStringify(JSON.parse(json), null, space);
 }
+
+// Function to recursively remove 'address' properties and lowercase string properties
```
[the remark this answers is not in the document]
yorhodes: I meant to restrict the type from object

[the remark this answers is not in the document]
paulbalaji: ahh I see, will include in follow-up pr

Issue #3454, opened 2024-03-20
Relayer startup is slow


tkporter:
## Problem

- Now that we have tens of thousands of messages from domains, it takes a couple minutes after starting up a relayer to get to a point where the serial submitter has been made aware of messages close to the tip.
- This is probably a mix of:
  - The current logic of "read from the DB starting at index 0, then 1, etc" in various tasks results in a ton of DB reads, so we're maybe IO bound here
  - I think it probably doesn't help that we run tokio in the current thread mode (https://docs.rs/tokio/latest/tokio/runtime/enum.RuntimeFlavor.html#variant.CurrentThread, https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/main.rs#L23) - would it help if we moved to multithread? As long as rocksdb allows for concurrent DB reads, I think this should help us out, as we tend to have a task per domain that reads all the messages / merkle tree insertions

## Solution

- Fast delivery as outlined here https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3414 is closely related, particularly moving away from the whole nonce-by-nonce reading in the db
- Maybe moving away from current thread flavor on tokio would help?

tkporter:
I think part of the problem is also that the interface to rocksdb isn't async. So we block when performing rocksdb IO, and we sometimes do this in loops

From https://ryhl.io/blog/async-what-is-blocking/:
> To give a sense of scale of how much time is too much, a good rule of thumb is no more than 10 to 100 microseconds between each .await. That said, this depends on the kind of application you are writing.
I wonder if it makes sense to move our DB operations to a spawn_blocking closure or something?

There seem to be places where we probably block for wayyy longer than 100 microseconds, like when we call this for the first time upon startup, and it'll loop through tens of thousands of message nonces without ever hitting an .await: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dcb67e97da6e0e3d9abd1533dc2a5ca2ff9b4617/rust/agents/relayer/src/msg/processor.rs#L119-L148

nambrot:
as part of this, it might be nice to allow relayer operators to opt out of merkle tree processing

tkporter:
[...]

Issue #3239, opened 2024-02-08
Instrument Tokio Tasks with `tokio-metrics`


daniel-savu:
As a start to profiling our code to find inefficiencies / bottlenecks, we should instrument all the tokio tasks in the agents codebase with [tokio-metrics](https://crates.io/crates/tokio-metrics). There is an off-the-shelf grafana dashboard for these metrics, [here](https://grafana.com/grafana/plugins/bsull-console-datasource/). Most interestingly, we'll be able to sort by the tasks taking up the most resources:

![task-list](https://github.com/hyperlane-xyz/hyperlane-monorepo/assets/23065004/56fc4e28-3c71-4ebb-975b-75a4abc7bf44)

tkporter:
@daniel-savu should we close this or keep it around because we don't have an easy way to explore the metrics?

PR #3760, opened 2024-05-10
feat: Relayer tokio task instrumentation

daniel-savu:
### Description

- Started off adding tokio-metrics but then realised those are quite general, so while we do have instrumentation it's not exposed in our metrics endpoint
- switched to adding [tokio-console](https://github.com/tokio-rs/console/tree/main), which does give insight into the lifetime of specific tasks, so we can check which ones take up a long time during relayer startup. These are only visible at the `dependencyTrace` log level, so don't affect performance in the `hyperlane` context.

### Related issues

- Helps debug https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3454 and any future performance issues
- Does half the work for https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3239 (still need to expose these in the metrics endpoint and import the grafana template)


rust/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
```
daniel-savu, line 1: We need a build flag since tokio tracing instrumentation is only available in their unstable release

tkporter (approved):

PR #3761, opened 2024-05-10
fix: ignore warnings from fx-portal lib

paulbalaji:
does what the title says, making use of https://www.npmjs.com/package/hardhat-ignore-warnings
image

### DIFF
--- a/solidity/hardhat.config.cts
+++ b/solidity/hardhat.config.cts
@@ -2,6 +2,7 @@ import '@nomiclabs/hardhat-ethers';
 import '@nomiclabs/hardhat-waffle';
 import '@typechain/hardhat';
 import 'hardhat-gas-reporter';
+import 'hardhat-ignore-warnings';
 import 'solidity-coverage';
 
 /**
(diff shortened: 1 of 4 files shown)


cmcewen (commented):
i need a few more reps in the repo before approving PRs but overall would say LGTM!

solidity/hardhat.config.cts
```diff
@@ -30,4 +31,10 @@ module.exports = {
     bail: true,
     import: 'tsx',
   },
+  warnings: {
+    // turn off all warnings for libs:
+    'fx-portal/**/*': {
```
nbayindirli, line 36: neat

nbayindirli (approved):
lgtm

solidity/hardhat.config.cts
paulbalaji: note that we'd have to specify the path for each lib we want to ignore for, didn't do every single lib because I feel it's safer to be selective about what we ignore

yorhodes (approved):

PR #3772, opened 2024-05-14
chore: update clusters to use new prometheus version

tkporter:
### Description

GKE was complaining that the kube-state-metrics version we were running with was hitting deprecated APIs, I expect this to fix it.

Some minor changes to the infra setup meant that now the push gateway has a different service name, kinda annoyingly with a double `prometheus` in the name. I updated key funder to push to the new gateway

### Drive-by changes

We no longer use GKE autopilot, so started using the node exporter too


daniel-savu (approved):
LGTM! Do we actually need the node exporter enabled? Looks like it's used for exposing machine metrics, which we weren't using ([repo](https://github.com/prometheus/node_exporter))

tkporter:
Yeah I intentionally enabled it, I figure it's better to have those metrics available

PR #3784, opened 2024-05-14
feat: util for parsing warp route messages

yorhodes:
### Description

Adds token message parser

### Backward compatibility

Yes

### Testing

Manual warp route send


typescript/utils/src/messages.ts
```diff
@@ -67,3 +74,19 @@ export function parseMessage(message: string): ParsedMessage {
...
 }
+
+export function parseTokenMessage(messageBody: string): ParsedTokenMessage {
+  const RECIPIENT_OFFSET = 0;
+  const AMOUNT_OFFSET = 32;
+  const buf = fromHexString(messageBody);
+  const recipient = toHexString(
+    buf.slice(RECIPIENT_OFFSET, RECIPIENT_OFFSET + 32),
+  );
+  const amount = ethers.BigNumber.from(
```
jmrossy, line 85: Can we avoid ethers BigNumber here? Most of the utils package code uses `bignumber.js`. You may remember a while ago we considered a few options (including native BigInt) and opted for bignumber.js because it's one of the few that supports decimals.

typescript/utils/src/messages.ts
```diff
@@ -67,3 +74,19 @@ export function parseMessage(message: string): ParsedMessage {
   const body = utils.hexlify(buf.slice(BODY_OFFSET));
   return { version, nonce, origin, sender, destination, recipient, body };
 }
+
+export function parseTokenMessage(messageBody: string): ParsedTokenMessage {
```
jmrossy, line 78: Just curious, does this work for all protocols?

jmrossy (approved):

typescript/utils/src/messages.ts
yorhodes: yes!

typescript/utils/src/messages.ts
yorhodes: switched to bigint

PR #3785, opened 2024-05-14
dx: update Kurtosis cloud link text

nbayindirli:


yorhodes (approved):

typescript/cli/src/deploy/agent.ts
```diff
@@ -65,7 +65,7 @@ export async function runKurtosisAgentDeploy({
   const kurtosisCloudUrl = getKurtosisCloudUrl(base64EncodedPackageConfig);
 
   const kurtosisCloudLink = terminalLink(
-    'Cmd+Click or Ctrl+Click here',
+    'Kurtosis Cloud Link 🔗 (cmd+click or ctrl+click here)',
```
yorhodes, line 68: could use the unicode symbols
Command (or Cmd) ⌘
Control (or Ctrl) ⌃

nbayindirli: Personally I think `⌘+click or ⌃+click`, `⌘click or ⌃click`, etc. are less readable than `cmd+click or ctrl+click` since we are asking for a "click" and not another key or character input

PR #3811, opened 2024-05-19
chore(RocksDB): optimize config for read-heavy workload

daniel-savu:
### Description

Rocks DB can be configured to be significantly more performant when doing reads. I added several of the recommended ones, leaving out `Compression::None` since that may require a migration. These will result in higher memory usage, which is easy to get more of if needed, at our current scale.

| Parameter                   | Suggested Value                | Default Value          | Expected Read Latency Improvement   |
|-----------------------------|---------------------------------|------------------------|------------------------|
| Block Cache Size            | 64MB                            | 8MB         | 2-5x              |
| Bloom Filters               | `NewBloomFilterPolicy(10)` (whole-file-based) | No Bloom filter        | 30-50%                 |
| Optimize for Point Lookups  | 64MB                            | Not set                | 30-50%                 |
| LRU Cache Sharding          | 6                               | 4                      | 10-30%                 |
| Readahead Size              | 512KB                           | 0                      | 20-30%                 |

### DIFF
--- a/rust/hyperlane-base/src/db/rocks/mod.rs
+++ b/rust/hyperlane-base/src/db/rocks/mod.rs
@@ -2,7 +2,7 @@ use std::path::PathBuf;
 use std::{io, path::Path, sync::Arc};
 
 use hyperlane_core::{ChainCommunicationError, HyperlaneProtocolError};
-use rocksdb::{Options, DB as Rocks};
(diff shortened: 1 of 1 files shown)


daniel-savu:
Just benchmarked this on RC and it takes just as long for processors to finish logging `Fast forwarded current sequence`, so I'll close. 

But one thing to still consider is our access patterns: a single full sequential read on startup, and then mostly writes that are append-only

PR #3792, opened 2024-05-15
chore(sdk): add zod validation on submitter/transformers popTx + refactor ICA interface

nbayindirli:
### Description

* adds zod validation on submitter/transformer populated transactions
* refactors ICA transformer interface to reuse getCallRemote types
* runs stopImpersonatingAccount after submitting impersonated txs

### Drive-by changes

* None

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3786

### Backward compatibility

* Yes

### Testing

* Manual (no i/o yet; testing TK via CLI)

PR #3817, opened 2024-05-20
chore(sdk): add zod validation on submitter/transformers popTx + refactor ICA interface

nbayindirli:
See comments and approvals here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3792


yorhodes (approved):
see comments on other PR

PR #3821, opened 2024-05-20
fix: minor change was breaking in registry export

cmcewen:


yorhodes (commented):
thanks

yorhodes (approved):

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -16,15 +16,15 @@ const {
 export const ethereumMainnetConfigs: ChainMap<ChainMetadata> = {
   ...defaultEthereumMainnetConfigs,
   bsc: {
-    ...getChain('bsc'),
+    ...(getChain('bsc') as ChainMetadata),
```
tkporter, line 19: nit: any reason to use `as` instead of `getChain('bsc')!`?

jmrossy (approved):

Issue #3267, opened 2024-02-15
Hook indexing learns about new messages as they are indexed


tkporter:
ContractSync::sync tasks should:

- if it's indexing messages, have a broadcast channel that it sends newly indexed messages through. Tbd whether all logs should be sent via this or just ones that the DB says are new?
- if it's indexing hooks (e.g. IGP, merkle tree), receive from these broadcast channels to learn about a new message log & LogMeta. Then these should look for hook events in that tx

tkporter:
Relevant: https://github.com/hyperlane-xyz/issues/issues/977

daniel-savu:
- when you index a message, send over a channel to the IGP task, an indication that IGP may have occurred in the same tx. 
    - to get this event from a tx hash, can query for the tx receipt - generalizes well across EEs and no need to impl for sealevel (we do sequence based there)
- nice to have: also send a notification to the merkle tree indexer task to read the insertion

tkporter:
Some thoughts on rolling out as it was mentioned yesterday we should have an explicit path toward seeing if this is effective -

Imo we already have decent metrics to be able to determine the success of this.

Because of low confidence in IGP indexing, atm we've whitelisted all known warp routes to not enforce IGP payments in the relayer (because we know that in order to send a message, they'll have paid the IGP):
- hyperlane context https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/agent.ts#L139-L156
- neutron context https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/agent.ts#L271-L278

As part of rolling this out, we should remove all these configurations except for the Neutron-origin ones (see https://github.com/hyperlane-xyz/issues/issues/1257, maybe Injective is affected too?) so that we start enforcing gas payments on EVM chains. If we see unprocessable messages from EVM origins for known metric app contexts, then we know something is up.

We could maybe go further and add a metric to count IGP payments indexed via this new method for metric app contexts, but imo this is overkill. As long as logs make it clear when an IGP payment is indexed via this new method and we have the above metrics / alerts to know if we find ourselves still having IGP indexing issues with known metric app contexts, imo we're fine

PR #3833, opened 2024-05-21
feat(evm): collaborative indexing through txid sharing

daniel-savu:
### Description

- improved indexing reliability by sharing channels between indexing tasks, such that they inform one another when they come across a hyperlane-related transaction
- the receiving end will then get the receipt of that tx and filter it for events relevant to it
- this adds some flexibility to indexing since we can stop relying on watermark cursor entirely: as long as we have one task using sequence indexing, all others can rely on being informed by their channels about what txs to index
- this PR is only for the EVM chains, but can be easily extended

### Drive-by changes

- Enforces gas payments for all chains in e2e (previously this only used to be the case for sealevel)
- The channel size is configured per type, similarly to the indexing cursor, in `contract_sync/cursors/mod.rs`
- Reduces log verbosity is some places, e.g. by manually implementing `Debug` to exclude fields that aren't relevant to debugging (maybe there's even a macro annotation to exclude, but I haven't checked)
- Renames `MpmcReceiver` -> `BroadcastReceiver`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3267

### Testing

e2e tested with the watermark cursor manually disabled


rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -141,16 +244,38 @@ pub trait ContractSyncer<T>: Send + Sync {
...
+    /// If this syncer is also a broadcaster, return the channel to receive txids
+    fn get_new_receive_tx_channel(&self) -> Option<BroadcastReceiver<H512>>;
+}
+
+#[derive(new)]
+/// Options for syncing events
+pub struct SyncOptions<T> {
+    // Keep as optional fields for now to run them simultaneously.
+    // Might want to refactor into an enum later, where we either index with a cursor or rely on receiving
+    // txids from a channel to other inedxing tasks
```
tkporter, line 261: nit typo `inedxing`

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -13,8 +13,15 @@ pub enum CursorType {
     RateLimited,
 }
 
+// H256 * 1M = 32MB per origin chain worst case
+// With one such channel per origin chain.
+const TX_ID_CHANNEL_CAPACITY: Option<usize> = Some(1_000_000);
+
 pub trait Indexable {
     fn indexing_cursor(domain: HyperlaneDomainProtocol) -> CursorType;
+    fn broadcast_channel_size() -> Option<usize> {
```
tkporter, line 22: nit: may be nice to `///` document these functions

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -27,100 +31,199 @@ const SLEEP_DURATION: Duration = Duration::from_secs(5);
...
-}
 
-impl<T, D, I> ContractSync<T, D, I>
-where
-    T: Debug + Send + Sync + Clone + Eq + Hash + 'static,
-    D: HyperlaneLogStore<T>,
-    I: Indexer<T> + 'static,
-{
+    fn get_new_receive_tx_channel(&self) -> Option<BroadcastReceiver<H512>> {
+        // Create a new channel if it doesn't exist
```
tkporter, line 70: nit: is this accurate? made me think that a new broadcast channel was created altogether, but i guess it's just a new receiver end - led me to misunderstand how it works for a few mins

rust/agents/relayer/src/relayer.rs
```diff
@@ -318,6 +319,18 @@ impl BaseAgent for Relayer {
             .instrument(info_span!("Relayer server"));
         tasks.push(server_task);
 
+        let txid_receivers = self
+            .message_syncs
+            .iter()
+            .filter_map(|(domain, sync)| {
+                let maybe_rx = sync.get_new_receive_tx_channel();
+                if maybe_rx.is_none() {
+                    warn!(?domain, "No txid receiver for chain");
```
tkporter, line 328: nit: wonder if we should just panic or something to make sure this never happens? as this is a startup invariant I assume we'll wanna keep. Don't feel strongly here

rust/agents/relayer/src/relayer.rs
```diff
@@ -403,6 +424,7 @@ impl Relayer {
     async fn run_interchain_gas_payment_sync(
         &self,
         origin: &HyperlaneDomain,
+        mut rxs: HashMap<HyperlaneDomain, BroadcastReceiver<H512>>,
```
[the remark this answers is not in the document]
tkporter: thinking about it more, because we clone the entire hashmap for each origin chain, iiuc we end up creating n^2 broadcast receivers (instead of `n`, where n = the number of origin chains) when running all the IGP syncs. Because they don't get cleaned up (`rxs` is moved into the task), the receivers for all the other origins won't ever read from the channel, so the data will stick around in memory https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging. So we should def change this to just the broadcast receiver

We could also maybe just make it a non-optional receiver too if we really wanna make sure that we always have a receiver here?

rust/agents/relayer/src/relayer.rs
```diff
@@ -318,6 +319,18 @@ impl BaseAgent for Relayer {
             .instrument(info_span!("Relayer server"));
         tasks.push(server_task);
```
tkporter, line 320: thinking a little about how the broadcast receiver works, I'm realizing that the way we use `MpmcChannel` for the relayer server may mean that we don't clean up the MessageRetryRequests even when they're served (until we have `ENDPOINT_MESSAGES_QUEUE_SIZE = 1_000` of them). This is because we only ever use `receiver` as something to clone here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-core/src/types/channel.rs#L31. So when we call `run_destination_submitter` on line 344 for each destination chain, we end up creating `n` new receivers, for a total of n+1 receivers. We read from the new `n` receivers in the destination tasks but never the original reciever that lives in the MpmcChannel

Imo this low prio & urgency to fix and doesn't need to be in this PR - maybe something good in a small separate PR or we can put it in the backlog? Will defer to you whether you think it's worth doing rn

rust/agents/scraper/src/agent.rs
```diff
@@ -195,7 +195,7 @@ impl Scraper {
             .await
             .unwrap();
         let cursor = sync.cursor(index_settings.clone()).await;
-        tokio::spawn(async move { sync.sync("message_dispatch", cursor).await }).instrument(
+        tokio::spawn(async move { sync.sync("message_dispatch", cursor.into()).await }).instrument(
```
tkporter, line 198: I know we didn't chat about this but imo it makes sense to also move the scraper and validator to using hook indexing, wdyt?

tkporter (commented):
Very nice

rust/agents/relayer/src/relayer.rs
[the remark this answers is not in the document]
daniel-savu: Great point, changed to passing a single receiver. The receiver is intentionally non-optional, since e.g. the message sync won't be receving anything, but just broadcasting

rust/chains/hyperlane-ethereum/src/contracts/interchain_gas.rs
```diff
@@ -124,6 +128,46 @@ where
...
+        let logs: Vec<_> = receipt
+            .logs
+            .into_iter()
+            .filter_map(|log| {
+                let raw_log = RawLog {
+                    topics: log.topics.clone(),
+                    data: log.data.to_vec(),
+                };
+                let log_meta: EthersLogMeta = (&log).into();
+                let gas_payment_filter = GasPaymentFilter::decode_log(&raw_log).ok();
```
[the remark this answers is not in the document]
daniel-savu: I assume so, `decode_log` is defined on `EthEvent`, which is derived for `GasPaymentFilter` in the abi interface file with this annotation:
```
    #[ethevent(
        name = "GasPayment",
        abi = "GasPayment(bytes32,uint32,uint256,uint256)"
    )]
```

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -33,6 +32,17 @@ pub(crate) struct BackwardSequenceAwareSyncCursor<T> {
     index_mode: IndexMode,
 }
 
+impl<T> Debug for BackwardSequenceAwareSyncCursor<T> {
```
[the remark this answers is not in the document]
daniel-savu: verbosity is the reason yep. I think this makes a good case for parsing agent logs in e2e -> if we really need something for debugging purposes, we'll try parsing it and realise it's not there

rust/chains/hyperlane-ethereum/src/contracts/merkle_tree_hook.rs
```diff
@@ -142,6 +147,41 @@ where
             .as_u32()
             .saturating_sub(self.reorg_period))
     }
+
+    async fn fetch_logs_by_tx_hash(
```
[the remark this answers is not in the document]
daniel-savu: done

rust/agents/relayer/src/relayer.rs
daniel-savu: unwrapping only makes sense if we want to enforce hook indexing. My assumption in this PR is that it's optional and we can alternate between this and cursors. If we stop relying on watermark cursor, can change this to a panic

rust/agents/relayer/src/relayer.rs
daniel-savu: fixed, that's a great point and explains why there isn't an `mpmc` channel in tokio

rust/agents/scraper/src/agent.rs
daniel-savu: we can't enable hook indexing for the validator bc we don't have a message indexer, but as discussed it's probably fine without because the checkpoint submitter already uses sequences.

I did enable it for the scraper 👍

rust/hyperlane-base/src/contract_sync/mod.rs
tkporter: I definitely didn't realize tx meant transmitter 😅
Bc we use tx in many places to mean transaction might be nice to keep tx for transaction in the future

rust/utils/run-locally/src/main.rs
```diff
@@ -412,7 +403,6 @@ fn main() -> ExitCode {
     while !SHUTDOWN.load(Ordering::Relaxed) {
         if config.ci_mode {
             // for CI we have to look for the end condition.
-            // if termination_invariants_met(&config, starting_relayer_balance)
             if termination_invariants_met(
```
tkporter, line 406: after https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3870 is merged, it may be nice to follow up here to make sure that we hit the hook indexing code path

tkporter (approved):
Approving but I believe https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3833#discussion_r1630942387 may need to be changed before merging

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -27,100 +31,197 @@ const SLEEP_DURATION: Duration = Duration::from_secs(5);
...
-        let stored_logs = self
+        let stored_logs_metric = self
             .metrics
             .stored_events
             .with_label_values(&[label, chain_name]);
 
         loop {
-            indexed_height.set(cursor.latest_queried_block() as i64);
+            if let Some(broadcaster) = opts.tx_id_broadcaster.as_mut() {
+                let mut rx = broadcaster.subscribe();
```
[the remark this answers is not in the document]
daniel-savu: changed, you're right

rust/agents/relayer/src/relayer.rs
```diff
@@ -305,8 +306,8 @@ impl BaseAgent for Relayer {
         }
 
         // run server
-        let mpmc_channel = MpmcChannel::<MessageRetryRequest>::new(ENDPOINT_MESSAGES_QUEUE_SIZE);
-        let custom_routes = relayer_server::routes(mpmc_channel.sender());
+        let broadcast_tx = Sender::<MessageRetryRequest>::new(ENDPOINT_MESSAGES_QUEUE_SIZE);
```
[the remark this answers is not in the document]
daniel-savu: renamed to just `sender`, the type itself should be informing enough

rust/agents/relayer/src/relayer.rs
```diff
@@ -403,6 +412,7 @@ impl Relayer {
     async fn run_interchain_gas_payment_sync(
         &self,
         origin: &HyperlaneDomain,
+        broadcast_tx: Option<Sender<H512>>,
```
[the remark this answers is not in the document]
daniel-savu: my original reasoning was to allow `clone`ing receivers if needed, but we need a single receiver so I changed this to be a `Receiver`

rust/utils/run-locally/src/main.rs
daniel-savu: will do, but I did test manually with the watermark cursor disabled and it worked

Issue #3740, opened 2024-05-08
[TxS] add `submit` command


nbayindirli:

nbayindirli:
Fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3818

Issue #3771, opened 2024-05-13
Use BigInt for large numbers in schemas


ltyu:
## Problem
Currently, some of our config parameters accepts [strings](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ff221f66a0a53bca4fa0def7e5c4597aa287424e/typescript/sdk/src/token/schemas.ts#L10) to represent large numbers. This prevents us from doing math on such numbers, without converting them using a BigNumber-ish library. 

Instead, we should use javascript's BigInt to represent them. 

## Solution
- Update schema parameters in their respective schema to use `z.bigint` for any value that can be greater than js' `2^53 - 1` limit such as `totalSupply`
- The other value add with using bigint is that zod-to-json will allow validation/autocomplete within our registry and elsewhere.

mdimran29:
hi, i like to work on it if it is still available. 
i notice some related PR are merged. could you confirm what still  remaining ?

PR #3834, opened 2024-05-21
chore(sdk): add further zod support to SDK

nbayindirli:
### Description

* adds further zod support to SDK, namely for areas relating to `transactions`

### Drive-by changes

* No

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3740

### Backward compatibility

- Yes

### Testing

- None


typescript/sdk/src/index.ts
```diff
@@ -303,20 +303,30 @@ export {
...
 export { TxSubmitterType } from './providers/transactions/submitter/TxSubmitterTypes.js';
+export {
+  EV5GnosisSafeTxSubmitterPropsSchema,
+  EV5ImpersonatedAccountTxSubmitterPropsSchema,
+} from './providers/transactions/submitter/ethersV5/schemas.js';
 export {
   EV5GnosisSafeTxSubmitterProps,
   EV5ImpersonatedAccountTxSubmitterProps,
-} from './providers/transactions/submitter/ethersV5/EV5TxSubmitterTypes.js';
+} from './providers/transactions/submitter/ethersV5/types.js';
```
jmrossy, line 319: I raised this question with @yorhodes and @ltyu too:
Is there value in having the schema and types be split across files this way, effectively doubling the file count for this kind of code? It seems to me that they're tightly coupled so there's no reason not to co-locate them.

typescript/sdk/src/providers/transactions/schemas.ts
```diff
@@ -1,7 +1,17 @@
...
 
+import { ZHash } from '../../metadata/customZodTypes.js';
+
 export const PopulatedTransactionSchema = z.object({
   to: z.string(),
   data: z.string(),
   chainId: z.number(),
 });
+
+export const BigNumberSchema = z.any();
```
jmrossy, line 11: I know this has been discussed before but can you clarify why we would want to allow non-primitive, lib-specific BigNumbers in our configs?

typescript/sdk/src/middleware/account/schemas.ts
```diff
@@ -10,3 +14,11 @@ export const AccountConfigSchema = z.object({
   routerOverride: ZHash.optional(),
   ismOverride: ZHash.optional(),
 });
+
+export const GetCallRemoteSettingsSchema = z.object({
```
jmrossy, line 18: Maybe rename this? Having trouble understanding what this is for from the name. Or leave a comment

typescript/sdk/src/providers/transactions/transformer/TxTransformerTypes.ts
```diff
@@ -1,3 +1,3 @@
 export enum TxTransformerType {
-  ICA = 'Interchain Account',
+  ICA = 'interchainAccount',
```
jmrossy, line 2: 👍

jmrossy: I'd be in favor of renaming the enum key value from `ICA` -> `InterchainAccount` for clarity

jmrossy (commented):
This mostly lgtm! Just a few small questions please

typescript/sdk/src/index.ts
nbayindirli: I am 100% on board with this. Faster to develop and fewer files. Perhaps we can rename all files to `zodTypes.ts`

Obviously– issue here is it would be a large refactor. The reason I did it this way is bc we were already doing it this way

@yorhodes wdyt

typescript/sdk/src/providers/transactions/schemas.ts
nbayindirli: afaik we do not want this.. ideally we would have this: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3771

line 11 exists here just to unblock me with getting these zod schemas completed

typescript/sdk/src/middleware/account/schemas.ts
nbayindirli: Added comment `/* For InterchainAccount::getCallRemote() */`

typescript/sdk/src/providers/transactions/transformer/TxTransformerTypes.ts
nbayindirli: Keen– will make update now

typescript/sdk/src/providers/transactions/schemas.ts
yorhodes: the workaround should be z.string() for any "bignumberish" types before 3771

typescript/sdk/src/providers/transactions/transformer/TxTransformerTypes.ts
```diff
@@ -1,3 +1,3 @@
 export enum TxTransformerType {
-  ICA = 'Interchain Account',
+  INTERCHAIN_ACCOUNT = 'interchainAccount',
```
cmcewen, line 2: is this exported outside of the SDK? would this be a breaking change?

typescript/sdk/src/index.ts
cmcewen: rossy your comment makes sense but in the context of this PR i do think it's good for noah to land it following existing convention, and then we can maybe just do a chore refactor of all of it at once

jmrossy: @cmcewen Yep agreed, just meant to propose the change

typescript/sdk/src/providers/transactions/transformer/TxTransformerTypes.ts
nbayindirli: Yes, and yes in isolation it is breaking. It's currently only consumed by my submit changes on CLI-side, which is not exposed via a command yet. This PR makes that change CLI-side [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/27be8100d7db710277cfb166f3e1219f146907e0/typescript/cli/src/submit/submit.ts#L78)

typescript/sdk/src/index.ts
nbayindirli: Issue to capture: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3844

typescript/sdk/src/providers/transactions/schemas.ts
```diff
@@ -1,7 +1,18 @@
...
   data: z.string(),
+  value: BigNumberSchema.optional(),
   chainId: z.number(),
 });
+
+export const CallDataSchema = z.object({
+  to: ZHash,
+  data: z.string(),
+  value: BigNumberSchema.optional(),
+});
```
yorhodes, line 18: why have 2 different schemas which share 3 of same fields?
also `to` is `ZHash` on `CallData` but `z.string` on `PopulatedTransaction`

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
yorhodes, line 1: these names are a little longer than I personally prefer but more of a taste thing

typescript/sdk/src/providers/transactions/transformer/ethersV5/schemas.ts
```diff
@@ -0,0 +1,8 @@
+import { GetCallRemoteSettingsSchema } from '../../../../middleware/account/schemas.js';
+
+export const EV5InterchainAccountTxTransformerPropsSchema =
+  GetCallRemoteSettingsSchema.pick({
+    chain: true,
+    config: true,
+    hookMetadata: true,
```
yorhodes, line 7: any intended use for hookMetadata atm?

yorhodes (commented):
would be great to have some super basic unit tests for your schemas

typescript/sdk/src/providers/transactions/transformer/ethersV5/schemas.ts
nbayindirli: None, just conforming to getCallRemote since that's out of scope for this change

typescript/sdk/src/providers/transactions/schemas.ts
nbayindirli: > why have 2 different schemas which share 3 of same fields?

to emulate the current state of the repository. we can eventually refactor everything to just use a single in-house type. captured here: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3845

> also to is ZHash on CallData but z.string on PopulatedTransaction

did not see this– updated PopulatedTransaction to use ZHash

paulbalaji (approved):

jmrossy (approved):

jmrossy:
Please add a changeset before merging

typescript/sdk/src/providers/transactions/submitter/builder/schemas.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod';
+
+import { TransformerMetadataSchema } from '../../transformer/schemas.js';
+import { SubmitterMetadataSchema } from '../schemas.js';
+
+export const SubmissionStrategySchema = z.object({
+  chain: z.string(),
```
ltyu, line 7: should this also use `ZChainName`?

Issue #3827, opened 2024-05-21
Infra-based services found in the monorepo image should have access to a registry


tkporter:
## Problem

It seems like infra-based services, like kathy, key funder, or warp route monitoring, won't work when ran from a monorepo docker image at the moment. The presence of a registry within the image is required, which isn't the case at the moment

## Solution

- Figure out the best solution for this - should we just clone the registry inside the image? What version of the registry should be used?
- We should be able to deploy new infra-based services

PR #3841, opened 2024-05-23
fix: fetch the registry's `main` branch in the monorepo Docker image

tkporter:
### Description

Infra services require the presence of a registry. This is expected to be found at the env var REGISTRY_URI. This PR fetches the latest registry from `main` and uses that as the registry in the monorepo image for infra services / scripts to make use of. To prevent caching of this git clone, we use a build-time arg such that if the arg changes, any later layers in the image will be ran. See https://stackoverflow.com/a/73501716

### Related issues

- Partially fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3827 - we'll need a deploy to fully close it out

### DIFF
--- a/.github/workflows/monorepo-docker.yml
+++ b/.github/workflows/monorepo-docker.yml
@@ -79,4 +79,3 @@ jobs:
           # To always fetch the latest registry, we use the date as the cache key
           build-args: |
             REGISTRY_CACHE=${{ env.TAG_DATE }}
-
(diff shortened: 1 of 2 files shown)


paulbalaji:
do we always want to use `main` or should we be pinning against a specific commit?

tkporter:
@paulbalaji there was a bit of a convo here https://discord.com/channels/935678348330434570/1242227398548783229/1242827232565788714 but also an offline one in PL yesterday - for now we're gonna go with just going off `main` because it's just used for registry consts, and the SLA we have also requires the consts in registry `main` to not break consumers (e.g. the CLI that reads from `main`)

I think if we start feeling pain around breakage the next step would be to pin

paulbalaji (approved):

Issue #3732, opened 2024-05-08
Ensure key funder desired balances are configured for all supported EVM chains


tkporter:
## Problem

See https://www.notion.so/hyperlanexyz/Hybrid-HaaS-67195982e71b42de92bc83038498583a?pvs=4#0d1f51d99ef0459eb1e1c92a686a1c07 for more detailed context

Copying from there:

### Current State
    
The key funder’s current state:

- It’s easy to forget to add a new chain to the key funder
    - It’s configured here, and there’s no check to make sure that all chains are specified there https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f8b33d1270b06670594100685fc346d504229967/typescript/infra/config/environments/mainnet3/funding.ts#L25-L49
- It’s easy to forget to create an alert
- Very specific to our infrastructure setup (e.g. it’s aware of contexts, key roles, etc)
- Doesn’t support non-EVM chains
- Fragile and complex

## Solution

Ultimately, we want to do a refactor to make key funder’s implementation far less specific to our infrastructure setup. This is **out of scope** for now though.

Instead, we’ll focus on how to make key funder operations easier

This issue specifically tracks:
- creating a CI check or stricter types that ensure that desired balances for each supported EVM chain is configured

PR #3843, opened 2024-05-23
chore: update infra service images (key funder, kathy)

tkporter:
### Description

- Updates key funder and kathy images on testnet4 and mainnet3
- Sorts desired balances and is more explicit about requirements for kathy chains (https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3732 will expand on this to make sure we keep this up to date)
- I found it surprising that if the kathy balance wasn't specified in the config, then key funder would default to funding the full relayer amount. Made this not be the case


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -575,11 +575,19 @@ class ContextFunder {
...
+      if (desiredKathyBalance === undefined) {
+        logger.warn({ chain }, 'No desired balance for Kathy, not funding');
+        desiredBalanceEther = '0';
+      } else {
+        desiredBalanceEther = this.desiredKathyBalancePerChain[chain];
+      }
+    } else {
+      desiredBalanceEther = this.desiredBalancePerChain[chain];
+    }
+    let desiredBalance = ethers.utils.parseEther(desiredBalanceEther);
```
paulbalaji, line 590: if role is not kathy, could we still get undefined desiredBalanceEther? ie do we still need this 
```suggestion
    let desiredBalance = ethers.utils.parseEther(desiredBalanceEther ?? '0');
```

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -575,11 +575,19 @@ class ContextFunder {
   }
 
   private getDesiredBalanceForRole(chain: ChainName, role: Role): BigNumber {
-    const desiredBalanceEther =
-      role === Role.Kathy && this.desiredKathyBalancePerChain[chain]
-        ? this.desiredKathyBalancePerChain[chain]
-        : this.desiredBalancePerChain[chain];
-    let desiredBalance = ethers.utils.parseEther(desiredBalanceEther ?? '0');
+    let desiredBalanceEther: string;
```
tkporter, line 578:
```suggestion
    let desiredBalanceEther: string | undefined;
```

paulbalaji (approved):

PR #3846, opened 2024-05-24
chore: Remove nambrot from codeowners

nambrot:
It had to happen #RIP, my getting review requested ended up too noisy to miss actual tags in issues


paulbalaji (approved):
image

PR #3852, opened 2024-05-27
feat(hyp-ethereum): gas escalator middleware

daniel-savu:
### Description

Wraps evm providers in a gas escalator middleware as part of the `build_with_signer` step, so signed txs can have their gas price bumped if stuck in the mempool.

Gas price is increased every 60s by 12.5%, and is capped at the `max_fee_per_gas` set in tx overrides.

### Drive-by changes

- changes to our fork of `ethers`:
    - removes the `M: Clone` bound, since it's not actually required and conflicts with the middleware we use
    - increases the level of the `escalated` log in the middleware to `debug` (so we can see it working in prod without enabling `trace`)

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2959

### Backward compatibility

Yes

### Testing

e2e


rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
```diff
@@ -24,7 +25,10 @@ use hyperlane_core::{
 };
 
 use crate::signer::Signers;
-use crate::{ConnectionConf, EthereumFallbackProvider, RetryingProvider, RpcConnectionConf};
+use crate::{
+    ConnectionConf, EthereumFallbackProvider, RetryingProvider, RpcConnectionConf,
+    TransactionOverrides,
+};
```
tkporter, line 31: Something else to be aware of is that the PendingTransaction we get back and wait for here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/chains/hyperlane-ethereum/src/tx.rs#L56 isn't totally compatible with the gas escalator. If the initial tx is replaced by the gas escalator, in `track_pending_tx` we'll still be looking for the first transaction. I guess this isn't so bad because we'd time out on the initial tx, and our op_submitter logic would eventually realize that the message had been processed, but still things get a bit funky

tkporter (commented):
After diving a bit deeper, imo like many things in ethers, it feels like not many people actually use these middlewares in production / long-running services, and that it may need some work to meet our bar

The worst thing is that it doesn't seem to have EIP-1559 support at all (see https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3852#discussion_r1617509153), and it returns errors if you try to use this middleware with 1559 txs. This feels like a dealbreaker imo in its current state - either we'd need to change it to just ignore 1559 txs but still allow them, or we'd need to change it to accommodate 1559 txs.

Though pls correct me if I'm wrong here because I would've expected e2e to fail if this were the case? Maybe the transactions submitted by the relayer on e2e are all legacy transactions?

Some other things that make me nervous:
1. In our fork, it holds the lock across potentially a lot of async operations, which feels questionable: https://github.com/hyperlane-xyz/ethers-rs/blob/master/ethers-middleware/src/gas_escalator/mod.rs#L177

In upstream this seems to have been fixed fwiw https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/gas_escalator/mod.rs

2. The task can silently error https://github.com/hyperlane-xyz/ethers-rs/blob/master/ethers-middleware/src/gas_escalator/mod.rs#L223 (I think it's still possible to hit this edge case https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2959#issuecomment-2135459848). Upstream at least it indicates that's happening https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/gas_escalator/mod.rs#L343

3. We'd probably want some way to test things to certainly hit the code path that results in a gas escalation - not sure if you did this, but being able to test this in e2e or at least ad-hoc would be nice to derisk things

daniel-savu:
E2E fails because validators can't send the announce txs. The error is `invalid chain id for signer`, originating [here](https://github.com/ethereum/go-ethereum/blob/c11aac249ddca275a12a1da0e66d1f4c6c21f4f6/core/types/transaction_signing.go#L277) I think, since e2e txs are using EIP-1559. 

This means that the chain ID in the txData is different from the chain ID in the signature for some reason. Should look into why

rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
```diff
@@ -274,3 +284,21 @@ where
...
+    tx_overrides: &TransactionOverrides,
+) -> GasEscalatorMiddleware<M, GeometricGasPrice>
+where
+    M: Middleware + 'static,
+{
+    // Increase the gas price by 12.5% every 60 seconds
+    // (These are the default values from ethers doc comments)
+    const COEFFICIENT: f64 = 1.125;
+    const EVERY_SECS: u64 = 60u64;
+    let escalator = GeometricGasPrice::new(COEFFICIENT, EVERY_SECS, tx_overrides.max_fee_per_gas);
```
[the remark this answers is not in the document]
daniel-savu: I see, do you recommend hardcoding a value here then?

rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
```diff
@@ -234,15 +244,23 @@ pub trait BuildableWithProvider {
...
-    let provider = NonceManagerMiddleware::new(provider, address);
+    Ok(SignerMiddleware::new(provider, signer))
+}
+
+async fn wrap_with_nonce_manager<M: Middleware>(
+    provider: M,
+    signer: Signers,
+) -> Result<NonceManagerMiddleware<M>, M::Error> {
+    let provider_chain_id = provider.get_chainid().await?;
+    let signer = ethers::signers::Signer::with_chain_id(signer, provider_chain_id.as_u64());
```
tkporter, line 259: I think we can remove this now and just pass in `signer_address: &Address` into this function. Bc we just create the signer here only for the address, and the signer is actually used in `wrap_with_signer`

rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
```diff
@@ -210,10 +211,19 @@ pub trait BuildableWithProvider {
         M: Middleware + 'static,
     {
         Ok(if let Some(signer) = signer {
-            let signing_provider = wrap_with_signer(provider, signer)
+            // - The signing provider is used for sending txs, which may end up stuck in the mempool due to
+            // gas pricing issues. So we first wrap the provider in a gas escalator middleware.
+            // - When txs reach the gas escalator, they will already have been signed by the signer middleware,
```
tkporter, line 216: nit: I think this comment is now stale -- iiuc when txs reach the gas escalator they have not been signed yet

tkporter (approved):
some tiny suggestions

rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
daniel-savu: updated and also mentioned the placing of the nonce manager as the outermost middleware

rust/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs
daniel-savu: not really something we can fix, the gas escalator unit test also has this issue

Issue #2846, opened 2023-10-26
Arbitrum/Orbit L2 => L1 Hook and ISM


nambrot:
This ticket encapsulates a hook and ISM that allows a Hyperlane message to be transported via Arbitrum's `ArbSys` precompile contract on L2 to L1. See more under https://docs.arbitrum.io/for-devs/cross-chain-messsaging

You will want to use a [CCIP-read based ISM](https://docs.hyperlane.xyz/docs/protocol/ISM/ccip-read-ISM) to allow the fetching of the `OutboxProof` data from the virtual precompile of the Arbitrum chain that can then be passed by any Hyperlane relayer to process the message on L1.

Submission of this bounty should include a demo on a testnet or mainnet with tx hashes.

exp7l:
is this available to be worked on?

nambrot:
Also yes!

shreyaspapi:
@nambrot is it still being worked on? Or would like to give it a try.

nambrot:
Looks like no progress has been made, so I'd say you go for it

avious00:
this is available for anyone to claim

avious00:
@exp7l want to claim this after you finish L1?

PR #3853, opened 2024-05-27
feat: support Arbitrum L2->L1 post dispatch hook

aroralanuk:
### Description

- Contract support for the enabling postDispatch hook with the Arbitrum nitro bridge from L2 to L1
  - asynchronously via executeTransaction call to verifyMessageId first and then the relayer calling the verify message with no metadata (note: this supports msg.value)
  - synchronously via a single verify call which in turn calls executeTransaction on outbox to gets the message verified in the verifyMessageId (note: this doesn't support msg.value as ism.verify isn't payable)
  - Added a script for deploying the hook and ISM since the sdk doesn't support it yet.

### Drive-by changes

- changing the type from "rate_limited_hook" to "RATE_LIMITED" to maintain consistency

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2846

### Backward compatibility

Yes

### Testing

Unit and e2e with arbitrumsepolia->sepolia


solidity/package.json
yorhodes, line 1: revert

solidity/remappings.txt
```diff
@@ -3,4 +3,5 @@
 @eth-optimism=../node_modules/@eth-optimism
 ds-test/=lib/forge-std/lib/ds-test/src/
 forge-std/=lib/forge-std/src/
-fx-portal/=lib/fx-portal/
\ No newline at end of file
+fx-portal/=lib/fx-portal/
+@arbitrum=../node_modules/@arbitrum
```
yorhodes, line 7: nit: move up next to other node_modules paths

solidity/test/isms/ArbL2ToL1Ism.t.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    function deployArbHook() public {
+        vm.selectFork(arbitrumFork);
+
+        l2Mailbox = new TestMailbox(ARBITRUM_DOMAIN);
+        hook = new ArbL2ToL1Hook(
+            address(l2Mailbox),
+            MAINNET_DOMAIN,
+            TypeCasts.addressToBytes32(
+                0xc005dc82818d67AF737725bD4bf75435d065D239
```
yorhodes, line 50: what is this?

solidity/test/isms/ArbL2ToL1Ism.t.sol
```diff
@@ -0,0 +1,84 @@
...
+    TestRecipient internal testRecipient;
+    bytes internal testMessage =
+        abi.encodePacked("Hello from the other chain!");
+    bytes internal encodedMessage;
+    bytes32 internal messageId;
+
+    function setUp() public {
+        arbitrumFork = vm.createFork(vm.rpcUrl("arbitrum"));
+
+        testRecipient = new TestRecipient();
```
yorhodes, line 36: this is necessarily on ethereum? we should have one deployed already

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -0,0 +1,79 @@
...
+        bytes calldata,
+        bytes memory payload
+    ) internal override {
+        uint256 leafNum = arbSys.sendTxToL1(
+            TypeCasts.bytes32ToAddress(ism),
+            payload
+        );
+
+        // TODO: if too expensive, remove this
+        // (uint256 size, , ) = arbSys.sendMerkleTreeState();
```
yorhodes, line 75: I dont see why this is necessary? https://github.com/OffchainLabs/nitro-contracts/blob/90037b996509312ef1addb3f9352457b8a99d6a6/src/bridge/IOutbox.sol#L69
where is this required on the L1?
is this relevant? https://github.com/OffchainLabs/nitro-contracts/blob/1cab72ff3dfcfe06ceed371a9db7a54a527e3bfb/src/precompiles/ArbSys.sol#L111

solidity/test/isms/ArbL2ToL1Ism.t.sol
aroralanuk: stale

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: not necessary, removed

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    function hookType() external pure override returns (uint8) {
+        return uint8(IPostDispatchHook.Types.ARB_L2_TO_L1);
+    }
+
+    function _quoteDispatch(
+        bytes calldata,
+        bytes calldata
+    ) internal pure override returns (uint256) {
+        return 120_000; // estimate based on on-chain gas usage
```
yorhodes, line 63: should we make this an immutable so we can configure it without source code changes?

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // ============ Internal functions ============
+
+    /// @inheritdoc AbstractMessageIdAuthHook
+    function _sendMessageId(
+        bytes calldata metadata,
+        bytes memory payload
+    ) internal override {
+        arbSys.sendTxToL1{value: metadata.msgValue(0)}(
```
yorhodes, line 73: isnt this `msgValue(0)`in conflict with supporting value from L2 to L1?

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
```diff
@@ -113,5 +96,29 @@ abstract contract AbstractMessageIdAuthorizedIsm is
         emit ReceivedMessage(messageId);
     }
 
+    function _statefulVerify(
```
yorhodes, line 99: why pull this into a separate function?

yorhodes: afaict it was already stateful

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
```diff
@@ -113,5 +96,29 @@ abstract contract AbstractMessageIdAuthorizedIsm is
...
+            VERIFIED_MASK_INDEX
+        );
+        // rest 255 bits contains the msg.value passed from the hook
+        if (verified) {
+            uint256 _msgValue = verifiedMessages[messageId].clearBit(
+                VERIFIED_MASK_INDEX
+            );
+            if (_msgValue > 0) {
+                verifiedMessages[messageId] -= _msgValue;
+                payable(message.recipientAddress()).sendValue(_msgValue);
```
yorhodes, line 117: I guess this means that the value can be sent non-atomically before the message arrives

yorhodes: imo this is fine but we should clearly document this behavior

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    // ============ External Functions ============
+
+    /// @inheritdoc IInterchainSecurityModule
+    function verify(
+        bytes calldata metadata,
+        bytes calldata message
+    ) external override returns (bool) {
+        return
+            _statefulVerify(metadata, message) ||
```
yorhodes, line 67: why is this `||`?

yorhodes: afaict `_statefulVerify` does not verify authorized sender so this could be exploited

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
```diff
@@ -0,0 +1,144 @@
...
+            l2Sender == TypeCasts.bytes32ToAddress(authorizedHook),
+            "ArbL2ToL1Ism: l2Sender != authorizedHook"
+        );
+
+        bytes32 messageId = message.id();
+        {
+            // for parsing the message id from the verifyMessageId calldata
+            bytes32 convertedBytes;
+            assembly {
+                convertedBytes := mload(add(data, 36))
```
yorhodes, line 116: please avoid assembly
how about this?
```solidity
(bytes4, bytes32 parsedMessageId) = abi.decode(data, (bytes4, bytes32));
```

yorhodes (commented):
I think we need to be slightly careful about this "caching" layer
we only want to write to the cache after *someone* has successfully called the outbox.executeTransaction

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: done

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: That's just the default value here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/befc38daba7d8c44615f0fb12ea1a9993c2c9b67/solidity/contracts/hooks/libs/StandardHookMetadata.sol#L59

Using msg.value - GAS_QUOTE now

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
aroralanuk: I want to reuse it in the ArbL2ToL1Ism.

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
aroralanuk: added

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
aroralanuk: the first condition is when someone has already called; the second is when we call executeTransaction internally. _statefulVerify doesn't need, verifyMessageId already did it.

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: back to default 0

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
aroralanuk: switched to slicing here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3853/commits/b5a78f3edb1d48650dbea47cd7dd92117f79aa0a

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
```diff
@@ -72,17 +72,23 @@ abstract contract AbstractMessageIdAuthorizedIsm is
...
+    // ============ Public Functions ============
+
+    /**
+     * @notice Release the value to the recipient if the message is verified.
+     * @param message Message to release value for.
+     */
+    function releaseValueToRecipient(
+        bytes calldata message
+    ) public returns (bool) {
+        bool verified = isVerified(message);
```
yorhodes, line 89: I dont think this function should be responsible for verifying
it might be slightly more readable if you move this `verified` branch to the `verify` function?
you are also reading from the same storage slot twice here

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
```diff
@@ -0,0 +1,143 @@
...
+                    bytes
+                )
+            );
+
+        // check if the sender of the l2 message is the authorized hook
+        require(
+            l2Sender == TypeCasts.bytes32ToAddress(authorizedHook),
+            "ArbL2ToL1Ism: l2Sender != authorizedHook"
+        );
+        require(data.length == 36, "ArbL2ToL1Ism: invalid data length"); // this data is an abi encoded call of verifyMessageId(bytes32 messageId)
```
yorhodes, line 110: nit: move comment to separate line above

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
```diff
@@ -0,0 +1,143 @@
...
+        bytes32 messageId = message.id();
+        bytes32 convertedBytes;
+        assembly {
+            convertedBytes := mload(add(data, 36))
```
yorhodes, line 114: add a comment about what 36 is (4 byte selector + 32 byte message id)

yorhodes (approved):

solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
aroralanuk: fixed

Issue #3688, opened 2024-05-01
[feat][cli] Create CLI command `hyperlane core config`


nambrot:
- Use interactive hook+ism config builder
- write core config to file system
```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3860
```

nbayindirli:
fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3857

PR #3857, opened 2024-05-28
feat(cli): add 'hyperlane core config' command

nbayindirli:
### Description

* adds new `hyperlane core config` command
* this combines previous `hl config create ism` & `hl config create hook` control flows
* adds default ISM behaviour to aggregate `IsmType.TRUSTED_RELAYER` and the previous default fallback, `IsmType.MESSAGE_ID_MULTISIG`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3688

### Backward compatibility

* yes

### Testing

* [x] manual
* [ ] ci-test

### DIFF
--- a/typescript/cli/src/commands/config.ts
+++ b/typescript/cli/src/commands/config.ts
@@ -1,12 +1,8 @@
 import { CommandModule } from 'yargs';
 
 import { createChainConfig, readChainConfigs } from '../config/chain.js';
-import { createHooksConfigMap } from '../config/hooks.js';
-import { createIsmConfigMap, readIsmConfig } from '../config/ism.js';
-import {
-  createMultisigConfig,
-  readMultisigConfig,
-} from '../config/multisig.js';
+import { readIsmConfig } from '../config/ism.js';
+import { readMultisigConfig } from '../config/multisig.js';
 import {
   createWarpRouteDeployConfig,
   readWarpRouteDeployConfig,
@@ -40,8 +36,6 @@ const createCommand: CommandModule = {
   builder: (yargs) =>
     yargs
       .command(createChainConfigCommand)
-      .command(createIsmConfigCommand)
-      .command(createHookConfigCommand)
       .command(createWarpRouteDeployConfigCommand)
       .version(false)
       .demandCommand(),
@@ -53,42 +47,6 @@ const createChainConfigCommand: CommandModuleWithContext<{}> = {
   describe: 'Create a new, minimal Hyperlane chain config (aka chain metadata)',
   handler: async ({ context }) => {
     await createChainConfig({ context });
-    process.exit(0);
-  },
-};
-
-const createIsmConfigCommand: CommandModuleWithContext<{
-  out: string;
-  advanced: boolean;
-}> = {
-  command: 'ism',
-  describe: 'Create a basic or advanced ISM config for a validator set',
-  builder: {
-    out: outputFileCommandOption('./configs/ism.yaml'),
-    advanced: {
-      type: 'boolean',
-      describe: 'Create an advanced ISM configuration',
-      default: false,
-    },
-  },
-  handler: async ({ context, out, advanced }) => {
-    if (advanced) {
-      await createIsmConfigMap({ context, outPath: out });
-    } else {
-      await createMultisigConfig({ context, outPath: out });
-    }
-    process.exit(0);
-  },
-};
-
-const createHookConfigCommand: CommandModuleWithContext<{ out: string }> = {
-  command: 'hooks',
-  describe: 'Create a new hooks config (required & default)',
-  builder: {
-    out: outputFileCommandOption('./configs/hooks.yaml'),
-  },
-  handler: async ({ context, out }) => {
-    await createHooksConfigMap({ context, outPath: out });
     process.exit(0);
   },
 };
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -1,0 +1,49 @@
+import { CommandModule } from 'yargs';
+
+import { createHooksConfigMap } from '../config/hooks.js';
+import { createIsmConfigMap } from '../config/ism.js';
+import { CommandModuleWithContext } from '../context/types.js';
+import { log } from '../logger.js';
+
+import { outputFileCommandOption } from './options.js';
+
+/**
+ * Parent command
+ */
+export const coreCommand: CommandModule = {
+  command: 'core',
+  describe: 'Manage core Hyperlane contracts & configs',
+  builder: (yargs) => yargs.command(config).version(false).demandCommand(),
+  handler: () => log('Command required'),
+};
+
+export const config: CommandModuleWithContext<{
+  ismAdvanced: boolean;
+  ismOut: string;
+  hooksOut: string;
+}> = {
+  command: 'config',
+  describe: 'Create a core configuration, including ISMs and hooks.',
+  builder: {
+    ismAdvanced: {
+      type: 'boolean',
+      describe: 'Create an advanced ISM & hook configuration',
+      default: false,
+    },
+    ismOut: outputFileCommandOption('./configs/ism.yaml'),
+    hooksOut: outputFileCommandOption('./configs/hooks.yaml'),
+  },
+  handler: async ({ context, ismAdvanced, ismOut, hooksOut }) => {
+    await createIsmConfigMap({
+      context,
+      outPath: ismOut,
+      shouldUseDefault: !ismAdvanced,
+    });
+    await createHooksConfigMap({
+      context,
+      outPath: hooksOut,
+    });
+
+    process.exit(0);
+  },
+};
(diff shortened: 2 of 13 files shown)


typescript/cli/src/commands/core.ts
```diff
@@ -0,0 +1,55 @@
...
+import { log } from '../logger.js';
+
+import { outputFileCommandOption } from './options.js';
+
+/**
+ * Parent command
+ */
+export const coreCommand: CommandModule = {
+  command: 'core',
+  describe: 'Manage core Hyperlane contracts',
```
jmrossy, line 17: contracts and configs?

typescript/cli/src/commands/core.ts
```diff
@@ -0,0 +1,55 @@
...
+  builder: (yargs) => yargs.command(config).version(false).demandCommand(),
+  handler: () => log('Command required'),
+};
+
+export const config: CommandModuleWithContext<{
+  ismAdvanced: boolean;
+  ismOut: string;
+  hooksOut: string;
+}> = {
+  command: 'config',
```
jmrossy, line 27: This breaks from the structure all the other config commands are using. i.e. config -> warp instead of warp -> config.

I know folks were generally more in favor of this structure but we should do one of the other, not both

jmrossy (commented):
I'm missing some context on the core stuff going on but the code changes mostly lgtm. Just concerned about the inconsistent command structure

typescript/cli/src/commands/core.ts
nbayindirli: this is why I provided the deprecation notice logs– I will have a final PR that removes all of the old naming conventions (tracked by this issue: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3860)

@yorhodes can you chime in here as well on the overall restructuring

yorhodes: yes @jmrossy we are migrating to `hyperlane <module> <action>` structure
I guess I agree we should wholesale migrate, though maybe a nice middleground is "silent" aliasing? ie not showing up in --help but maintaining backwards compatibility with the old paths?

typescript/cli/src/commands/config.ts
yorhodes, line 1: maybe a good compromise is to just alias this command to the new one and remove the implementation here netirely

typescript/cli/src/config/hooks.ts
```diff
@@ -151,43 +158,52 @@ export async function createHookConfig(
   context: CommandContext,
   chain: ChainName,
   remotes: ChainName[],
+  hookTypeOverrides?: HookType[],
 ): Promise<HookConfig> {
+  let hookTypeOverride: HookType | undefined;
+  if (hookTypeOverrides) {
+    if (hookTypeOverrides.length === 1) hookTypeOverride = hookTypeOverrides[0];
+    if (hookTypeOverrides.length > 1) hookTypeOverride = HookType.AGGREGATION;
+  }
```
yorhodes, line 167: instead of polluting this function with the "overrides", can we just have a default `HookConfig` that we return before entering this function?

typescript/cli/src/config/hooks.ts
```diff
@@ -307,17 +323,27 @@ export async function createAggregationConfig(
...
+  hookTypeOverrides?: HookType[],
 ): Promise<HookConfig> {
-  const hooksNum = parseInt(
-    await input({
-      message: 'Enter the number of hooks to aggregate (number)',
-    }),
-    10,
-  );
+  let hooksNumOverride: number | undefined;
+  if (hookTypeOverrides) hooksNumOverride = hookTypeOverrides.length;
```
yorhodes, line 329: ditto here
lets pull the overrides out above this function

typescript/cli/src/config/ism.ts
```diff
@@ -154,47 +156,58 @@ export async function createIsmConfigMap({
...
   origins: ChainName[],
+  moduleTypeOverrides?: IsmType[],
 ): Promise<ZodIsmConfig> {
+  let moduleTypeOverride: IsmType | undefined;
+  if (moduleTypeOverrides) {
+    if (moduleTypeOverrides.length === 1)
+      moduleTypeOverride = moduleTypeOverrides[0];
+    if (moduleTypeOverrides.length > 1)
+      moduleTypeOverride = IsmType.AGGREGATION;
+  }
```
yorhodes, line 167: ditto here
lets pull the overrides out above this function

typescript/cli/src/config/ism.ts
```diff
@@ -249,24 +267,34 @@ async function createTrustedRelayerConfig(): Promise<ZodIsmConfig> {
...
-  const isms = parseInt(
-    await input({
-      message: 'Enter the number of ISMs to aggregate (number)',
-    }),
-    10,
-  );
+  let ismsOverride: number | undefined;
+  if (moduleTypeOverrides) ismsOverride = moduleTypeOverrides.length;
+  const isms =
+    ismsOverride ??
```
yorhodes, line 275: ditto here
lets pull the overrides out above this function

typescript/cli/src/config/ism.ts
```diff
@@ -249,24 +267,34 @@ async function createTrustedRelayerConfig(): Promise<ZodIsmConfig> {
...
+    parseInt(
+      await input({
+        message: 'Enter the number of ISMs to aggregate (number)',
+      }),
+      10,
+    );
 
   const threshold = parseInt(
     await input({
-      message: 'Enter the threshold of ISMs to for verification (number)',
```
yorhodes, line 262: 🙏

yorhodes (approved):

yorhodes (changes requested):
revoking approval until we make the fix

typescript/cli/src/commands/core.ts
jmrossy: Recommend breaking and migrating. Small numbers here:
Screenshot 2024-05-29 at 5 13 10 PM

nbayindirli: Agreed– I've now removed the stale commands in favor of the new ones.

typescript/cli/src/commands/config.ts
nbayindirli: per above convo I removed entirely and did **not** alias. if you'd like to have aliases please let me know

typescript/cli/src/config/hooks.ts
nbayindirli: i'd like to use as much of the existing flow as possible, and this gives us to ability to specify HookTypes/IsmTypes directly from the CLI or config in the future

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0c55e1259e3b07fb7a1458623aaa1f148c79427d/typescript/cli/src/config/hooks.ts#L164-L165
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0c55e1259e3b07fb7a1458623aaa1f148c79427d/typescript/cli/src/config/hooks.ts#L381-L389
are you alright with this instead?

nbayindirli: I see your comments below to pull out into an above function– going to consider this satisfactory

yorhodes: > i'd like to use as much of the existing flow as possible, and this gives us to ability to specify HookTypes/IsmTypes directly from the CLI or config in the future

this seems like strictly using **less** of the existing flow because you modified this function and changed the CLI interface

nbayindirli: updated to this:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2663f26b77b373a887b478ee57e1b1a5b52820a1/typescript/cli/src/config/ism.ts#L137-L139

typescript/cli/src/config/ism.ts
```diff
@@ -114,9 +117,11 @@ export function isValildIsmConfig(config: any) {
 export async function createIsmConfigMap({
   context,
   outPath,
+  shouldUseDefault = false,
```
yorhodes, line 120: would expect this to default to true

typescript/cli/src/config/ism.ts
```diff
@@ -236,17 +250,22 @@ export async function createMultisigConfig(
...
-    message: 'Enter relayer address',
-  });
+async function createTrustedRelayerConfig(
+  context: CommandContext,
+): Promise<ZodIsmConfig> {
+  const relayer = await detectAndConfirmOrPrompt(
+    async () => context.signer?.getAddress(),
+    'For trusted relayer ISM, enter',
+    'relayer address',
+  );
```
yorhodes, line 260: nice :D

typescript/cli/src/config/ism.ts
```diff
@@ -300,3 +320,7 @@ export async function createRoutingConfig(
     domains: domainsMap,
   };
 }
+
+async function createDefaultIsmConfig(context: CommandContext) {
+  return await createTrustedRelayerConfig(context);
+}
```
yorhodes, line 326: nit: unnecessary await :P
also given this is just 1 line with no transform I would skip this indirection

yorhodes (approved):

typescript/cli/src/config/ism.ts
nbayindirli: the final straw lmao: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/861af3197c75a9e2f864e03fc4d66271d955827f/.eslintrc#L26

nbayindirli: removed extra function

PR #3861, opened 2024-05-29
feat: implement `create()` for `EvmHookModule`

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/1153

- enables creation of new Hooks through the EvmHookModule
- introduce `EvmModuleDeployer`
	- separate from `HyperlaneDeployer`
	- contains some basic methods to deploy a contract/proxy
	- reduces module necessity HyperlaneDeployer

IN PROGRESS:
- [x] tests
	- [x] figure out why randomly generated routing/fallbackrouting hooks fail
	- [x] figure out why protocol fee hooks fail
image

### DIFF
--- a/.changeset/shy-countries-heal.md
+++ b/.changeset/shy-countries-heal.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Implement `create()` for the `EvmHookModule`
--- a/.changeset/shy-countries-heal.md
+++ b/.changeset/shy-countries-heal.md
@@ -2,4 +2,5 @@
 '@hyperlane-xyz/sdk': minor
 ---
 
-Implement `create()` for the `EvmHookModule`
+- Enables creation of new Hooks through the `EvmHookModule`.
+- Introduces an `EvmModuleDeployer` to perform the barebones tasks of deploying contracts/proxies.
--- a/typescript/sdk/src/deploy/EvmModuleDeployer.ts
+++ b/typescript/sdk/src/deploy/EvmModuleDeployer.ts
@@ -1,0 +1,288 @@
+import { ethers } from 'ethers';
+import { Logger } from 'pino';
+
+import {
+  StaticAddressSetFactory,
+  StaticThresholdAddressSetFactory,
+  TransparentUpgradeableProxy__factory,
+} from '@hyperlane-xyz/core';
+import { buildArtifact as coreBuildArtifact } from '@hyperlane-xyz/core/buildArtifact.js';
+import { Address, rootLogger } from '@hyperlane-xyz/utils';
+
+import { HyperlaneContracts, HyperlaneFactories } from '../contracts/types.js';
+import { MultiProvider } from '../providers/MultiProvider.js';
+import { ChainMap, ChainName } from '../types.js';
+
+import { isProxy, proxyConstructorArgs } from './proxy.js';
+import { ContractVerifier } from './verify/ContractVerifier.js';
+import {
+  ContractVerificationInput,
+  ExplorerLicenseType,
+} from './verify/types.js';
+import { getContractVerificationInput } from './verify/utils.js';
+
+export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
+  public verificationInputs: ChainMap<ContractVerificationInput[]> = {};
+
+  constructor(
+    protected readonly multiProvider: MultiProvider,
+    protected readonly factories: Factories,
+    protected readonly logger = rootLogger.child({
+      module: 'EvmModuleDeployer',
+    }),
+    protected readonly contractVerifier = new ContractVerifier(
+      multiProvider,
+      {},
+      coreBuildArtifact,
+      ExplorerLicenseType.MIT,
+    ),
+  ) {}
+
+  // Deploys a contract from a factory
+  public async deployContractFromFactory<F extends ethers.ContractFactory>({
+    chain,
+    factory,
+    contractName,
+    constructorArgs,
+    initializeArgs,
+  }: {
+    chain: ChainName;
+    factory: F;
+    contractName: string;
+    constructorArgs: Parameters<F['deploy']>;
+    initializeArgs?: Parameters<Awaited<ReturnType<F['deploy']>>['initialize']>;
+  }): Promise<ReturnType<F['deploy']>> {
+    this.logger.info(
+      `Deploy ${contractName} on ${chain} with constructor args (${constructorArgs.join(
+        ', ',
+      )})`,
+    );
+    const contract = await this.multiProvider.handleDeploy(
+      chain,
+      factory,
+      constructorArgs,
+    );
+
+    if (initializeArgs) {
+      this.logger.debug(`Initialize ${contractName} on ${chain}`);
+      const overrides = this.multiProvider.getTransactionOverrides(chain);
+      const initTx = await contract.initialize(...initializeArgs, overrides);
+      await this.multiProvider.handleTx(chain, initTx);
+    }
+
+    const verificationInput = getContractVerificationInput(
+      contractName,
+      contract,
+      factory.bytecode,
+    );
+    this.addVerificationArtifacts({ chain, artifacts: [verificationInput] });
+
+    // try verifying contract
+    try {
+      await this.contractVerifier?.verifyContract(chain, verificationInput);
+    } catch (error) {
+      // log error but keep deploying, can also verify post-deployment if needed
+      this.logger.debug(`Error verifying contract: ${error}`);
+    }
+
+    return contract;
+  }
+
+  /**
+   * Deploys a contract with a specified name.
+   *
+   * This is a generic function capable of deploying any contract type, defined within the `Factories` type, to a specified chain.
+   *
+   * @param {ChainName} chain - The name of the chain on which the contract is to be deployed.
+   * @param {K} contractKey - The key identifying the factory to use for deployment.
+   * @param {string} contractName - The name of the contract to deploy. This must match the contract source code.
+   * @param {Parameters<Factories[K]['deploy']>} constructorArgs - Arguments for the contract's constructor.
+   * @param {Parameters<Awaited<ReturnType<Factories[K]['deploy']>>['initialize']>?} initializeArgs - Optional arguments for the contract's initialization function.
+   * @param {boolean} shouldRecover - Flag indicating whether to attempt recovery if deployment fails.
+   * @returns {Promise<HyperlaneContracts<Factories>[K]>} A promise that resolves to the deployed contract instance.
+   */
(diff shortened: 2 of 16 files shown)


typescript/sdk/src/ism/EvmIsmModule.ts
yorhodes, line 1: why change this?

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,599 @@
...
+    if (!config.oracleConfig) {
+      this.logger.debug('No oracle config provided, skipping...');
+      return gasOracle;
+    }
+
+    this.logger.info(`Configuring gas oracle from ${this.chainName}...`);
+    const configsToSet: Array<StorageGasOracle.RemoteGasDataConfigStruct> = [];
+
+    for (const [remote, desired] of Object.entries(config.oracleConfig)) {
+      // TODO: add back support for non-EVM remotes.
```
yorhodes, line 547: is there an issue for this?

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,599 @@
...
+    mailbox,
+    proxyAdmin,
+    multiProvider,
+  }: {
+    chain: ChainNameOrId;
+    config: HookConfig;
+    deployer: HyperlaneHookDeployer;
+    factories: HyperlaneAddresses<ProxyFactoryFactories>;
+    mailbox: Address;
+    proxyAdmin: Address;
```
yorhodes, line 123: can we collapse addresses into one param?

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,599 @@
...
+    config,
+    deployer,
+    factories,
+    mailbox,
+    proxyAdmin,
+    multiProvider,
+  }: {
+    chain: ChainNameOrId;
+    config: HookConfig;
+    deployer: HyperlaneHookDeployer;
```
yorhodes, line 120: why does this need to be passed in?

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,599 @@
...
+    proxyAdmin: Address;
+    multiProvider: MultiProvider;
+  }): Promise<EvmHookModule> {
+    // instantiate new EvmHookModule
+    const module = new EvmHookModule(multiProvider, deployer, {
+      addresses: {
+        ...factories,
+        mailbox,
+        proxyAdmin,
+        deployedHook: ethers.constants.AddressZero,
```
yorhodes, line 132: can we deploy before we instantiate so that we don't mutate this field?

typescript/sdk/src/hook/EvmHookModule.ts
yorhodes: my sense is we need to replace the multiprovider with one of 
- multi protocol provider
- registry 
- chain metadata manager or something here

typescript/sdk/src/hook/EvmHookModule.ts
yorhodes: it seems like a lot of code is redundant between this create fn and the hook deployer
when will we dedupe?

yorhodes (commented):
thanks for getting this up

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: bug fix - if your given object contains more fields than required, those extra fields are also passed into `addresses` from here `const { mailbox: _, deployedIsm: __, ...addresses } = args.addresses;`, which then breaks the expectations inside attachAndConnectContracts

this way we're very strict about "give any object containing these fields, and i'll extract exactly the ones I care about"

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: only using fns from the abstract class, providing `HyperlaneHookDeployer` is a quick way to have a deployer type-compliant with the relevant hook/igp contracts

paulbalaji: naturally there will be a lot of dupe to avoid reinventing the wheel, part of the reason the updates are so difficult is because it involves understanding what's going on in the deployer and then doing a split

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: this comment was already there before

i think though that if the multiprovider is setup with chainmetadata that includes non-evm chains, then we're still fine - main issue is if we can't do the chainname<>domainId mapping not whether we have evm/non-evm

i think in infra before, we would automatically filter out non-evm chains, which is what led to the issue?

typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -0,0 +1,288 @@
...
+  }: {
+    chain: ChainName;
+    factory: StaticThresholdAddressSetFactory | StaticAddressSetFactory;
+    values: Address[];
+    logger: Logger;
+    threshold?: number;
+    multiProvider: MultiProvider;
+  }): Promise<Address> {
+    const address = await factory['getAddress(address[],uint8)'](
+      values,
```
ltyu, line 253: should we `sort()` the values as according to https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6620fe6367cc8cbc50d2edab4b057ccec1da41ba/solidity/contracts/libs/StaticAddressSetFactory.sol#L48

paulbalaji: yeah @yorhodes and i had a chat about this before - we ended up deciding that sorting should be something done by the caller/whoever writes the configuration in the first place - so that it simplifies the logic on the deployer/update/checking end of making sure it matches what we expect

ltyu (approved):
mostly looks good to me. awesome that youre able to bring over the verifier logic

ltyu:
> ## [Codecov](https://app.codecov.io/gh/hyperlane-xyz/hyperlane-monorepo/pull/3861?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperlane-xyz) Report
> All modified and coverable lines are covered by tests ✅
> 
> > Please [upload](https://docs.codecov.com/docs/codecov-uploader) report for BASE (`cli-2.0@cc65ee8`). [Learn more](https://docs.codecov.io/docs/error-reference?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=hyperlane-xyz#section-missing-base-commit) about missing BASE report.
> 
> Additional details and impacted files

don't forget to add this :D

typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -0,0 +1,288 @@
...
+   * Deploys a contract with a specified name.
+   *
+   * This is a generic function capable of deploying any contract type, defined within the `Factories` type, to a specified chain.
+   *
+   * @param {ChainName} chain - The name of the chain on which the contract is to be deployed.
+   * @param {K} contractKey - The key identifying the factory to use for deployment.
+   * @param {string} contractName - The name of the contract to deploy. This must match the contract source code.
+   * @param {Parameters<Factories[K]['deploy']>} constructorArgs - Arguments for the contract's constructor.
+   * @param {Parameters<Awaited<ReturnType<Factories[K]['deploy']>>['initialize']>?} initializeArgs - Optional arguments for the contract's initialization function.
+   * @param {boolean} shouldRecover - Flag indicating whether to attempt recovery if deployment fails.
```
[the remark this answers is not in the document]
paulbalaji: good catch thanks - I'll refresh the comments in this file

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,644 @@
...
+        config,
+        this.multiProvider.getSignerOrProvider(this.args.chain),
+      );
+    }
+
+    switch (config.type) {
+      case HookType.MERKLE_TREE:
+        return this.deployer.deployContract({
+          chain: this.chain,
+          contractKey: config.type,
```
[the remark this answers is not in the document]
paulbalaji: yup - can do!

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,644 @@
...
+    storageGasOracle: StorageGasOracle;
+    config: IgpConfig;
+  }): Promise<InterchainGasPaymaster> {
+    const deployerAddress = await this.multiProvider.getSignerAddress(
+      this.chain,
+    );
+
+    const igp = await this.deployer.deployProxiedContract({
+      chain: this.chain,
+      contractKey: 'interchainGasPaymaster',
```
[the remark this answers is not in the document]
paulbalaji: nice yes we can - the factories are keyed on the enum values anyway

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -1,53 +1,644 @@
...
+    // set hooks
+    await this.multiProvider.handleTx(
+      this.chain,
+      routingHook.setHooks(hooksToSet, this.txOverrides),
+    );
+
+    // transfer ownership
+    await this.multiProvider.handleTx(
+      this.chain,
+      routingHook.transferOwnership(config.owner, this.txOverrides),
```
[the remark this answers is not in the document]
paulbalaji: > In the legacy transferOwnershipOfContracts() of HyperlaneDeployer, we check config.ownerOverrides, is this being phased out?

so in a similar way to sorting of the static address set, I'm assuming in this module that the caller is responsible for resolving this sort of thing - so here we only look at the config.owner

> Kind of an optimization, but also in transferOwnershipOfContracts(), we only transfer if config.owner != deployer to save gas. wondering if we should bring this fn into the new Deployer

neat - i think we do this check at the end of the update() in ism/hook - so maybe worth moving out to a separate fn in the deployer just for owner transfer. could do it in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3888 or a follow-up PR

.changeset/shy-countries-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Implement `create()` for the `EvmHookModule`
```
paulbalaji, line 5:
```suggestion
- Enables creation of new Hooks through the `EvmHookModule`.
- Introduces an `EvmModuleDeployer` to perform the barebones tasks of deploying contracts/proxies.
```

PR #3870, opened 2024-05-30
feat: batching igp

daniel-savu:
### Description

Deducts IGP cost from senders of messages within a batch, based on the total gas used by the batch and the estimates for submitting messages individually.

The formula is `gas_used_by_operation = gas_used_by_tx *  (operation_estimated_gas / total_estimated_cost)`

Note that the individual estimate have a buffer value added to them (currently 75k), which will slightly skew proportions, though by a negligible amount.

For example, given these estimates in a batch: 150k, 160k, 180k, if we add 80k to each, we go from e.g. 0.312 for the first message to 0.319 -> 2.25% error which isn't so bad. The error can be larger if operations within a batch have very different estimates, but realistically within a 5% range based on back-of-the-napkin calculations.

### Drive-by changes

- The batching feature introduced an bug whereby gas expenditure would only be deducted in the `confirm` step. Now this is done in the `submit` step to account for txs that revert
- Sealevel e2e can now be disabled for faster iteration! I've not done the cleanest job, but even the fact that we have this will reduce the time to run e2e locally by more than half. To use it, set `SEALEVEL_ENABLED=false` when running `run-locally`
- e2e uses a new utility called `get_matching_lines`, that can be used to count (and in the future _parse_) logs, to reconstruct the state of the relayer and have more expressive correctness checks. This is used to make sure that gas is deducted for all messages, including those in batches.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3709

### Backward compatibility

Yes

### Testing

E2E, using `get_matching_lines`


rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -331,6 +326,46 @@ impl PendingOperation for PendingMessage {
...
+            return;
+        };
+        if let Err(e) = self
+            .ctx
+            .origin_gas_payment_enforcer
+            .record_tx_outcome(&self.message, submission_outcome.clone())
+        {
+            error!(error=?e, "Error when recording tx outcome");
+        }
+        match gas_used_by_batch_operation(
```
tkporter, line 345: nit: maybe worth clarifying somewhere that this is used by non-batched operations too, took me a bit

rust/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -85,11 +101,39 @@ pub trait PendingOperation: Send + Sync + Debug + TryBatchAs<HyperlaneMessage> {
...
+
+/// Calculate the gas used by an operation in a batch, by looking at the total cost of the batch,
+/// and the estimated cost of the operation compared to the sum of the estimates of all operations in the batch.
+pub fn gas_used_by_batch_operation(
+    batch_outcome: &TxOutcome,
+    batch_estimated_cost: U256,
+    operation_estimated_cost: U256,
+) -> ChainResult<U256> {
+    let gas_used_by_batch = FixedPointNumber::try_from(batch_outcome.gas_used)?;
+    let gas_limit = FixedPointNumber::try_from(operation_estimated_cost)?;
```
tkporter, line 129: nit: i had a bit of a hard time following this logic bc the variable names of the fixed versions didn't match the param names, might be easier to read if they stay similar

tkporter (approved):
approving to unblock but think https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3870#discussion_r1629608369 may need changed

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: meant to rename this and forgot, thanks for pointing out

tkporter (approved):

PR #3872, opened 2024-05-30
fix: backport rusoto timeout change to v2

daniel-savu:
Backport of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3283

Applies the fix in
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/2384 everywhere an `HttpClient` is constructed via rusoto.

It lowers the S3 timeout to 15s based on tips in [this thread](https://github.com/hyperium/hyper/issues/2136#issuecomment-589488526), to avoid `Error during dispatch: connection closed before message completed` errors. Note that we'll probably still run into these issues, but less frequently
([source](https://github.com/rusoto/rusoto/issues/1766#issuecomment-639299961)).

### DIFF
--- a/rust/hyperlane-base/src/settings/signers.rs
+++ b/rust/hyperlane-base/src/settings/signers.rs
@@ -1,16 +1,15 @@
-use std::time::Duration;
-
 use async_trait::async_trait;
 use ed25519_dalek::SecretKey;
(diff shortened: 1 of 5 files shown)


yorhodes (approved):
seems like you went from 20s to 15s?

Issue #3774, opened 2024-05-14
Ensure infra scripts / services use RPC URLs from GCP secrets


tkporter:
## Problem

It seems like following the registry code changes, we no longer use RPC URLs from GCP secrets

## Solution

Make sure that we use the private RPCs when running local deploys, kathy, key funder, etc

yorhodes:
We should declare a `LocalRegistry` instance for each environment in infra
Then we can use `MergedRegistry` to combine that with the existing `MergedRegistry` that @jmrossy migrated infra to
if network round trips go the github remote registry are a concern, we can ignore the github registry entirely and just merge the local clone with the environment overrides

jmrossy:
> We should declare a `LocalRegistry` instance for each environment in infra Then we can use `MergedRegistry`

I don't totally follow why we'd want separate LocalRegistries for that but RE MergedRegistry I guess this is reason reason number 2 to have it live outside the CLI. I'll migrate it to the Registry repo this week.

tkporter:
done

PR #3875, opened 2024-05-31
fix: use secret RPC urls in infra, apply environment-specific registry overrides

tkporter:
### Description

A while back, it seems secret RPC URLs stopped being used by monorepo services (see https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3179/files#diff-ba3b7bea2b5c046acd88d25ed382df875b49d77e061f2ef8347ffcc741c0353eR34). This adds them back.

Some things to note on the approach:
- Sadly, due to the way some of the infra config generation works, we're forced to use synchronous interfaces for getting chain metadata in a bunch of places. This is why `getRegistry` returns a `FileSystemRegistry`, which is a `SynchronousRegistry`.
- Applying secrets / overrides requires us to do things async for two reasons -- one is just that fetching from GCP secrets is an async operation, the other is that a `MergedRegistry` is not a `SynchronousRegistry`. The approach I've taken is that there are two ways of getting a registry - using `getRegistry` will give the synchronous version that can continue to be used by our sync config generation, but it won't have any environment-specific overrides (tx overrides or secrets). To get an async registry with environment-specific overrides, the `getRegistry` function on the EnvironmentConfig is used.
- Overrides are now more succinct as the merging happens later in the pipeline

### Drive-by changes

- Started moving away from using the gcloud CLI under the hood for GCP operations and toward using the GCP secret SDK
- Updated the mainnet3 key funder to fix an incident, but will follow up with a fresh deploy once this is mergd

### Related issues

- These are the code changes for https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3774, but I'll follow up with fresh deploys for the monorepo services


typescript/infra/src/utils/gcloud.ts
```diff
@@ -43,6 +44,29 @@ export async function fetchGCPSecret(
...
+  });
+  const [secretVersion] = await client.accessSecretVersion({
+    name: `projects/${GCP_PROJECT_ID}/secrets/${secretName}/versions/latest`,
+  });
+  const secretData = secretVersion.payload?.data;
+  if (!secretData) {
+    throw new Error(`Secret ${secretName} missing payload`);
+  }
+
+  // Handle both string and Uint8Array
```
paulbalaji, line 59: so this works for both?

```
secretValue
```

as well as 

```
{
 'x': 'secretValue',
 'y': 'secretOtherValue'
}
```

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -381,7 +381,7 @@ class ContextFunder {
       [Role.Kathy]: '',
     };
     const roleKeysPerChain: ChainMap<Record<FundableRole, BaseAgentKey[]>> = {};
-    const chains = getEnvironmentConfig(environment).chainMetadataConfigs;
+    const chains = getEnvironmentConfig(environment).supportedChainNames;
```
paulbalaji, line 384: nit
```suggestion
    const { supportedChainNames } = getEnvironmentConfig(environment);
```

typescript/infra/src/deployment/deploy.ts
```diff
@@ -122,15 +123,18 @@ export async function writeAgentConfig(
   environment: DeployEnvironment,
 ) {
   const addresses = getAddresses(environment, Modules.CORE);
-  const addressesForEnv = objFilter(
+  const ethereumAddressesForEnv = objFilter(
```
paulbalaji, line 126: i know it's just renaming something that was there already, but how do we know it's always eth addresses only

afaict we don't actually filter on protocol type in the multiprovider, so if we want only ethereum ones we should filter for `tryGetProtocol` being `ProtocolType.Ethereum`

paulbalaji: the `isEthereumProtocolChain` import seems unused

paulbalaji (commented):
just a couple qs

typescript/infra/src/deployment/deploy.ts
tkporter: oops this is a relic of a path I went down that I abandoned, will undo this all

typescript/infra/src/utils/gcloud.ts
tkporter: the data type the SDK provides is a `string | Uint8Array` so this just converts to a string if we get a `Uint8Array`. Then if that string is a JSON or anything else it can be parsed later on

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
tkporter: applied locally and pushed

paulbalaji (approved):

PR #3879, opened 2024-06-03
feat: implement hyperlane core config to return CoreConfig

### LINKED ISSUE #3877: `hyperlane core config` does not return CoreConfig
## Problem
In `config()` of core.ts, ISM and hooks are created, but produces split ism and hook config files, as specified in `ismOut` and `hooksOut`.  This is problematic because it does not conform to a single CoreConfig, which is expected by `hyperlane core deploy`
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/db9f733e4f4b82f9fd85d2a1285d95ff22519209/typescript/sdk/src/core/types.ts#L12-L18

## Solution
Instead of producing separate ism and hook files, consider producing a single CoreConfig that conforms to the example:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/1d0d1bb36bda9568bae882695647c642c2281090/typescript/cli/examples/core-config.yaml#L1-L19

ltyu:
### Description
- Updates `hl core config` to output a single `CoreConfig` file
- Makes `CoreConfig` single chain ie. remove ChainMap
- Removes `nativeTokenAndDecimals()` for fee calculation
- Add description as a prameter for `outputFileCommandOption()`

### Related issues
- Fixes #3877

### Backward compatibility
No, Outputs CoreConfig as a single file and no longer expecting 2 individual files

### Testing
Manual

### DIFF
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -1,6 +1,10 @@
 import { CommandModule } from 'yargs';
 
-import { HookConfigSchema, IsmConfig } from '@hyperlane-xyz/sdk';
+import {
+  HookConfigSchema,
+  IsmConfig,
+  IsmConfigSchema,
+} from '@hyperlane-xyz/sdk';
 
 import { createHookConfig } from '../config/hooks.js';
 import { createIsmConfig, createTrustedRelayerConfig } from '../config/ism.js';
@@ -86,6 +90,7 @@ export const config: CommandModuleWithContext<{
     );
 
     // Validate
+    IsmConfigSchema.parse(defaultIsm);
     HookConfigSchema.parse(requiredHook);
     HookConfigSchema.parse(defaultHook);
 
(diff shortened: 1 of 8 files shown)


typescript/cli/src/commands/core.ts
```diff
@@ -43,22 +56,45 @@ export const config: CommandModuleWithContext<{
...
       context,
-      outPath: hooksOut,
-    });
+      'Select required hook type',
+    );
+
+    // Validate
+    IsmConfigSchema.parse(defaultIsm);
+    HookConfigSchema.parse(requiredHook);
+    HookConfigSchema.parse(defaultHook);
```
yorhodes, line 95: any reason not to parse with the CoreConfigSchema?
maybe you can put the hook parse inside of the createHookConfig fn and ditto for createIsmConfig

yorhodes (approved):
love it
please also make the `core read` spit out `CoreConfig` (if it does not already)

typescript/cli/src/commands/core.ts
ltyu: Will investigate and fix in https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3880

PR #3881, opened 2024-06-03
feat: implement hyperlane core read to return CoreConfig 

### LINKED ISSUE #3880: `hyperlane core read` does not return CoreConfig
## Problem
In `read()` of core.ts, ISM and hooks are read, but produces split ism and hook config files, as specified in `ismOut` and `hooksOut`.  This is problematic because it does not conform to a single CoreConfig, which is expected by `hyperlane core deploy`
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/db9f733e4f4b82f9fd85d2a1285d95ff22519209/typescript/sdk/src/core/types.ts#L12-L18

## Solution
- Consider modifying read() to accept input of a single `mailbox` address, instead of `ismAddress` and `hookAddress`
- Instead of producing separate ism and hook files, consider producing a single CoreConfig that conforms to the example:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/1d0d1bb36bda9568bae882695647c642c2281090/typescript/cli/examples/core-config.yaml#L1-L19

ltyu:
### Description
- Updates `hl core read` to output a single `CoreConfig` file

### Related issues
- Fixes #3880

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -1,6 +1,7 @@
 import { CommandModule } from 'yargs';
 
 import {
+  EvmCoreReader,
   HookConfigSchema,
   IsmConfig,
   IsmConfigSchema,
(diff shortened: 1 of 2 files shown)


yorhodes (approved):
🚢

PR #3883, opened 2024-06-04
refactor: rename chain and config commands

ltyu:
### Description
- Renames `chain` to `registry
  - `hl registry *` 
- Renames `config` to `configure`
  - `hl core configure`
  - `hl warp configure`

### Drive-by
- Update `hl core configure` to prompt user for `owner`
- Update key detection for cli `--key`
- bug fix in core mailbox deploy
  
### Backward compatibility
No

### Testing
Manual


nbayindirli (changes requested):
looks like this is failing build

nbayindirli (approved):
sweet ty– lgtm

PR #3887, opened 2024-06-04
feat: deploy to new chains (fraxtal, linea, sei)

paulbalaji:
https://github.com/hyperlane-xyz/issues/issues/1279

- adds `fraxtal`, `linea`, `sei`
- supports `osmosis` as a remote
- ~drive-by hack to skip reconfiguring the same gas oracle (cc @tkporter)~
- drive-by fix to bring the logic of estimating tx + adding 10% buffer into the ism factory (as originally seen in the multiprovider handleDeploy function)
- drive-by fix for creating custom safes
- drive-by support of `osmosis` as a remote chain
- drive-by addition of `osmosis` to the rust agent


paulbalaji:
do not be afraid! most of the diff number comes from
image

typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -213,6 +230,11 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
     ],
   },
 
+  osmosis: {
+    threshold: 1,
+    validators: ['0xea483af11c19fa41b16c31d1534c2a486a92bcac'],
```
tkporter, line 235: realizing I think we need Mitosis's validators here as well. not sure if Mitosis is using our default ISM or not, hopefully they're using their own ISM so this isn't something we need to go through multisig txs again for

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -373,12 +373,23 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
           destination,
           config.owner,
         );
-        const tx = await domainRoutingIsmFactory.deploy(
+        // estimate gas
+        const estimatedGas = await domainRoutingIsmFactory.estimateGas.deploy(
```
tkporter, line 377: do we need to worry about similar changes in this file in the cli v2 branch? or will this be effectively deprecated

asking bc it may be nice to have a more reusable function that does this estimation and 10% bumping, but if it's gonna be replaced by cli-2.0 changes anyways then it's not worth it

tkporter (approved):

typescript/sdk/src/consts/multisigIsm.ts
tkporter: following up on discord about this

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -77,19 +81,23 @@ export const hyperlaneContextAgentChainConfig: AgentChainConfig = {
...
     injective: true,
     inevm: true,
+    linea: true,
     mantapacific: true,
     mode: true,
     moonbeam: true,
     // At the moment, we only relay between Neutron and Manta Pacific on the neutron context.
     neutron: false,
     optimism: true,
+    osmosis: false,
```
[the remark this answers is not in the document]
paulbalaji: left it off because we were running into issues on the validator (see dan's recent fixes), we should be good but would rather do that with a new build after we've merged this + dan's fixes

typescript/sdk/src/consts/multisigIsm.ts
paulbalaji: probably when we get them - but same as before would wait til in flight fixes are merged and we get a fresh build

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
paulbalaji: I've made a PR already that will include this bump: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3957

The logic's already in the multiprovider, the ISM deploy functions here are an edge case really

having a reusable func is probs good though, but I'll save that for the cli-2.0 PR

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -157,15 +158,21 @@ export abstract class HyperlaneAppGovernor<
...
+              chain,
+              call,
+            );
+          }
+          call.submissionType = submissionType;
         }
-        call.submissionType = submissionType;
+      } catch (error) {
+        console.error(
+          `Error inferring call submission types for chain ${chain}: ${error}`,
```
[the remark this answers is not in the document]
paulbalaji: I think so, because if we have an issue with a particular chain then we stop for it altogether and can rerun for that specific chain afterwards

typescript/sdk/src/aws/validator.ts
```diff
@@ -71,6 +71,15 @@ export class S3Validator extends BaseValidator {
     return resp.data.value;
   }
 
+  async getSignedAnnouncement(): Promise<S3Announcement> {
+    const resp = await this.s3Bucket.getS3Obj<S3Announcement>(ANNOUNCEMENT_KEY);
```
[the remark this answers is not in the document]
paulbalaji: ah nice, didn't spot this while checking diffs

Issue #3577, opened 2024-04-09
Implement HookModule update() with support for IGP (and merkle) artifacts


yorhodes:
Blocked by:
- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3129

When updating in the HookCrudModule, we need to make sure that we don't end up deploying new contracts of things that we want to be indexing (e.g. the IGP and the merkle tree hook). So when we perform an update, we should first read the existing HookConfig, and use this to identify the existing addresses of the IGP and merkle tree hook. If it's easier to only support updating when the new config is of the same "tree shape" as the existing onchain config, that's fine. We probably also want to keep the same StorageGasOracle address and avoid deploying a new one.

We should specifically keep in mind that we will want to be able to update IGP configs / gas oracles using this entrypoint in the future

PR #3888, opened 2024-06-04
feat: implement `update()` for `EvmHookModule`

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3577

- enables updates of hooks
- includes functionality for updating IGP hook and gas oracles
- drive-by fix to ism module

TODO:

- [x] tests (non-igp)
- [x] tests (igp)


typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -416,10 +418,14 @@ describe('EvmIsmModule', async () => {
 
       // point to new mailbox
       ism.setNewMailbox(newMailboxAddress);
+      console.log(mailboxAddress);
```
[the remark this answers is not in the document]
paulbalaji: woops nice catch - i removed a bunch after testing but not these

ltyu:
> ### ⚠️ No Changeset found
> Latest commit: [cebab01](https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/cebab01075fc34492b354ee563349e0c2f192ed0)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/hyperlane-xyz/hyperlane-monorepo/new/pb/3577?filename=.changeset/seven-houses-grin.md&value=---%0A%22%40hyperlane-xyz%2Fsdk%22%3A%20patch%0A---%0A%0Afeat%3A%20implement%20%60update()%60%20for%20%60EvmHookModule%60%0A)

add this :D

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -184,29 +287,296 @@ export class EvmHookModule extends HyperlaneModule<
...
+      targetConfig.type === HookType.FALLBACK_ROUTING &&
       !configDeepEquals(
-        target.fallback,
-        (current as FallbackRoutingHookConfig).fallback,
+        targetConfig.fallback,
+        (currentConfig as FallbackRoutingHookConfig).fallback,
       )
     ) {
-      const hook = await this.deploy({ config: target });
+      const hook = await this.deploy({ config: targetConfig });
```
ltyu, line 572: iiuc, `deploy()` will transfer ownership to the `config.owner`. This is fine if the owner and the subsequent signer of this updateTx are the same. 

But i believe it will be an issue if `config.owner` is set to an address that cannot sign txs, such as SAFE, address(0), etc.

example flow for illustration:
- CLI calls this fn with `signer`
- fn deploys, and then transfers to `config.owner`
- CLI attempts to `sendTransaction(updateTxs)` using `signer`

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -184,29 +287,296 @@ export class EvmHookModule extends HyperlaneModule<
...
+      updateTxs.push(
+        ...(await this.updateStorageGasOracle({
+          gasOracle,
+          currentOracleConfig: currentConfig.oracleConfig,
+          targetOracleConfig: targetConfig.oracleConfig,
+          targetOverhead: targetConfig.overhead, // used to log example remote gas costs
+        })),
+      );
+    } else {
+      const newGasOracle = await this.deployStorageGasOracle({
```
ltyu, line 365: iiuc, deploy() will transfer ownership to the config.owner. This is fine if the owner and the subsequent signer of this updateTx are the same.

But i believe it will be an issue if config.owner is set to an address that cannot sign txs, such as SAFE, address(0), etc.

example flow for illustration:
- CLI calls this fn with signer
- fn deploys, and then transfers to config.owner
- CLI attempts to sendTransaction(updateTxs) using signer

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +121,103 @@ export class EvmHookModule extends HyperlaneModule<
...
+    }
+
+    // Lastly, check if the resolved owner is different from the current owner
+    const owner = await Ownable__factory.connect(
+      this.args.addresses.deployedHook,
+      this.multiProvider.getProvider(this.chain),
+    ).owner();
+
+    // Return an ownership transfer transaction if required
+    if (!eqAddress(targetConfig.owner, owner)) {
```
ltyu, line 208: In which case will this execute? i thought the owner is always transferred, if needed in the deploy fns?

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -184,29 +287,296 @@ export class EvmHookModule extends HyperlaneModule<
...
+      if (!remoteDomain) {
+        this.logger.warn(
+          `Skipping gas oracle update ${this.chain} -> ${remote}. Expected if the remote domain is not in the MultiProvider.`,
+        );
+        continue;
+      }
+
+      // only update if the oracle config has changed
+      if (!current || !configDeepEquals(current, target)) {
+        configsToSet.push({ remoteDomain, ...target });
```
ltyu, line 472: There's no validation in these Config schemas. This might lead to weird unexpected behavior. For example,

If `target` has `remoteDomain` key, it will overwrite the fetched `remoteDomain`. Might be intentional, but it's also not a valid `StorageGasOracleConfigSchema`. I believe its possible to have something like 
```typescript
const target: StorageGasOracleConfigSchema = { 
  gasPrice: "1000", 
  tokenExchangeRate: "1000", 
  blablablabla: 7777 
}
```

Consider using Schema.parse to 1) validate config schemas, and 2) filter out extra parameters

I probably have to do this for my PRs too ;)

ltyu (commented):
mostly makes sense. i'll need to have another peek at this

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +121,103 @@ export class EvmHookModule extends HyperlaneModule<
...
+    // If configs match, no updates needed
+    if (configDeepEquals(currentConfig, targetConfig)) {
+      return [];
+    }
+
+    // Else, we have to figure out what an update for this Hook entails
+
+    // If target config is an address Hook, just update the address
+    // if config -> address Hook, update address
+    // if address Hook -> address Hook, update address
```
paulbalaji, line 143: avoid "->" arrow notation, in favour of explicit sentences/words

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: the intent with these new modules is that they should read/create/update to the input given by the caller, to reduce the necessity of "owner" checks and implicit behavioural changes that plagued the current suite of deployers

so any `deploy()` here will ensure the owner of the new contract matches the given config's owner - this should mean that there are no follow-on updates required afterwards

so if the CLI does attempt to do _another_ update afterwards, it is expected for that to go through the multisig, as that is the configuration that was provided

in the current world we already have the case where we if do a new deploy for adding more chains, we have to go through a multisig to set up the new ISM as the default on the mailbox

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: this can be super useful down the line with checker/governance actions too

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: i'm aware it's not great in ts land with objects so i tried to be mindful about only selecting/using fields i require from an object - but I'll add schema parsing at the top of constructor/deploy/update functions in both ISM + Hook modules as it never hurts to be safe!

ltyu (commented):
I approve, but will be comfortable if another dev takes a look. @yorhodes

typescript/sdk/src/hook/EvmHookModule.hardhat-test.ts
```diff
@@ -156,15 +153,19 @@ function randomHookConfig(
 }
 
 describe('EvmHookModule', async () => {
+  const chain = TestChainName.test4;
+
+  // beforeEach
```
nbayindirli, line 158: nit: remove

typescript/sdk/src/hook/EvmHookModule.hardhat-test.ts
```diff
@@ -156,15 +153,19 @@ function randomHookConfig(
...
-  const chain = TestChainName.test4;
+  let fundingAccount: Signer;
   let proxyFactoryAddresses: HyperlaneAddresses<ProxyFactoryFactories>;
   let factoryContracts: HyperlaneContracts<ProxyFactoryFactories>;
+  let exampleRoutingConfig: DomainRoutingHookConfig | FallbackRoutingHookConfig;
 
   beforeEach(async () => {
-    const [signer] = await hre.ethers.getSigners();
+    const [signer, funder] = await hre.ethers.getSigners();
+    fundingAccount = funder;
```
nbayindirli, line 168: set `let signer: Signer;` in parent scope and

```suggestion
    [signer, funder] = await hre.ethers.getSigners();
```

nbayindirli (commented):
partial, still reviewing

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -184,29 +288,296 @@ export class EvmHookModule extends HyperlaneModule<
...
+      // update storage gas oracle
+      // Note: this will only update the gas oracle for remotes that are in the target config
+      updateTxs.push(
+        ...(await this.updateStorageGasOracle({
+          gasOracle,
+          currentOracleConfig: currentConfig.oracleConfig,
+          targetOracleConfig: targetConfig.oracleConfig,
+          targetOverhead: targetConfig.overhead, // used to log example remote gas costs
+        })),
+      );
```
nbayindirli, line 364: can we follow this same `updateTxs.push(...(await this.*));` pattern above in the `update()` switch-case? this would also update line 168 to `const updateTxs: AnnotatedEV5Transaction[] = [];`

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
...
+    // normalize the config to ensure it's in a consistent format for comparison
+    const currentConfig = normalizeConfig(await this.read());
+
+    // Update the config
+    this.args.config = targetConfig;
+
+    // If configs match, no updates needed
+    if (configDeepEquals(currentConfig, targetConfig)) {
+      return [];
+    }
```
nbayindirli, line 148: thoughts on declaring `let updateTxs: AnnotatedEV5Transaction[];` early on, and returning this instead of `[]`?

nit: remove comment and inline like `if (...) return updateTxs;`

nbayindirli (approved):
thanks for flushing this out!

most of my comments are re codestyle, and therefore readability/maintainability. i believe at this point in the SDK we should err on using variable & function names to convey purpose, probably resulting in a bunch of `private` helper functions in classes like these

one thing that seems useful for repurpose is capturing all the update TX types somewhere so they can be reused through the SDK, up to you if worth the effort rn

otherwise thanks for being so thorough and functionally looks good, approved

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
       : this.reader.deriveHookConfig(this.args.addresses.deployedHook);
   }
 
-  public async update(_config: HookConfig): Promise<AnnotatedEV5Transaction[]> {
-    throw new Error('Method not implemented.');
+  public async update(
+    targetConfig: HookConfig,
+  ): Promise<AnnotatedEV5Transaction[]> {
+    targetConfig = HookConfigSchema.parse(targetConfig);
```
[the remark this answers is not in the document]
paulbalaji: it's to conform to the abstract HyperlaneModule class, where `TConfig` in the `EvmHookModule` is `HookConfig`, we reuse the TConfig in a few places throughout the abstract class to aid typings
image

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
...
+
+    // Do not support updating to a custom Hook address
+    if (typeof targetConfig === 'string') {
+      throw new Error(
+        'Invalid targetConfig: Updating to a custom Hook address is not supported. Please provide a valid Hook configuration.',
+      );
```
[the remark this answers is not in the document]
paulbalaji: not really, because the error is specifically for the case when targetConfig is just a pure string instead of an object

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
...
+    }
+
+    // save current config for comparison
+    // normalize the config to ensure it's in a consistent format for comparison
+    const currentConfig = normalizeConfig(await this.read());
```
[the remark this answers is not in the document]
paulbalaji: this is probably a preference thing but I would much rather read a comment than try to infer from variable names

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
...
+    // Else, we have to figure out what an update for this Hook entails
+    // Check if we need to deploy a new Hook
+    if (
+      // if updating from an address/custom config to a proper hook config, do a new deploy
+      typeof currentConfig === 'string' ||
+      // if updating a proper hook config whose types are different, do a new deploy
+      currentConfig.type !== targetConfig.type ||
+      // if it is not a mutable Hook, do a new deploy
+      !MUTABLE_HOOK_TYPE.includes(targetConfig.type)
+    ) {
```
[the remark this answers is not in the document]
paulbalaji: this is a good candidate for a helper function

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -113,8 +123,102 @@ export class EvmHookModule extends HyperlaneModule<
...
+      updateTxs.push({
+        annotation: 'Transferring ownership of ownable Hook...',
+        chainId: this.domainId,
+        to: this.args.addresses.deployedHook,
+        data: Ownable__factory.createInterface().encodeFunctionData(
+          'transferOwnership(address)',
+          [targetConfig.owner],
+        ),
+      });
+    }
```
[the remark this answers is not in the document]
paulbalaji: I've thought about doing this but didn't want to overload this PR, because we can extract this out to a wider transfer ownership helper that we can reuse across all the modules instead of just in here

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: I usually prefer to `return []` to make it explicit what should happen

paulbalaji:
> i believe at this point in the SDK we should err on using variable & function names to convey purpose, probably resulting in a bunch of `private` helper functions in classes like these

yeah I can agree with that - however I'm not sure I agree with removing comments? I like to sprinkle them through bits of logic so that when read together it paints a clearer picture of what's happening where. yes, we should have var/fn names that indicate purpose but not to the point you _have_ to read the code. imo comments/naming both can help in different ways 

> one thing that seems useful for repurpose is capturing all the update TX types somewhere so they can be reused through the SDK, up to you if worth the effort rn

100% - again this was one of those things that I wanted to do after merging this in and post-tx submitter merging (shows how long this PR has been cooking lol). There's some common update patterns across the modules and I wanted to get a feel for how the modules get implemented before typing them

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: not sure I like the verbosity but ig idm too much

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: re inline if statements, i'm not really a fan of them at all but will adopt if it makes its way into the linter

keeping as-is because there's probably a bunch of other places that exist where we have a mix of inline and normal

PR #3898, opened 2024-06-05
feat: relayer dashboard template

daniel-savu:
### Description

Relayer metrics template for monitoring & alerting. Looks like so:
Screenshot 2024-05-29 at 18 08 26


paulbalaji (approved):
LGTM - tried importing into grafana myself and seemed to work fine

PR #3899, opened 2024-06-05
feat: setup turborepo

paulbalaji:
- use [turbo](https://turbo.build/repo/docs) for faster build times
	- automatic build task scheduling/dependency ordering
	- automatic build caching
	- faster build times
	- see turbo docs for more info
	- better integration with depot cache
- use depot's turbo cache https://depot.dev/docs/cache/reference/turbo
- use more depot runners to speed up recovery of yarn-build-with-cache action
	- depot runners will make use of depot's caching backend over github-native or buildjet


package.json
```diff
@@ -12,24 +12,25 @@
...
+    "turbo": "^2.0.1"
   },
   "dependencies": {
     "@changesets/cli": "^2.26.2"
   },
   "packageManager": "yarn@4.0.2",
   "private": true,
   "scripts": {
-    "build": "yarn workspaces foreach --all --parallel --topological run build",
+    "build": "turbo run build",
```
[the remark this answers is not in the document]
paulbalaji: yup! in the turbo.json you sort of just configure the jobs you want to run - and then turbo detects the topology and can do only the jobs necessary to fulfil the command you have

e.g. if you're messing with some infra test but make some modifications, `turbo test` within the infra package will auto do the relevant sdk build before the infra test

paulbalaji:
> I haven't experimented with Turbo before but if it is indeed faster than I'm in favor. I imagine the caching action configuration in the CI job defs would need to be tweaked to support this, otherwise I wouldn't expect any perf benefit in CI

Yup that's right, we wouldn't get any immediate benefit in CI jobs as it stands but I put this up as a draft to could play around in spare time and see how/if it improves the local workflow

mpsq (approved):
Nice

xeno097 (approved):
🙌🏼

.github/workflows/release.yml
```diff
@@ -9,6 +9,14 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+env:
+  LOG_LEVEL: DEBUG
```
antigremlin, line 13: Do we want it to always be set to DEBUG? Could we rely on the debug UI switch in GitHub actions?

turbo.json
```diff
@@ -0,0 +1,25 @@
+{
+  "ui": "stream",
+  "tasks": {
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": ["dist/**", "cache/**", "types/**", "release/**"]
```
antigremlin, line 6: This doesn't cover the outputs for some packages. For example:
- solidity has artifacts and typechain-generated typescript
- helloworld is similar
- starknet has release

Should we add package-specific `turbo.json` for these?

antigremlin (approved):
This is great, and nudges us to do more improvements. For example, once this is merged, we can add shared configs for TypeScript, ESLint, and Prettier.

.github/workflows/release.yml
paulbalaji: good catch - we don't need to default to DEBUG

turbo.json
paulbalaji: i was debating it in my head but overall I think it's the right call, added specific turbo.jsons for solidity/helloworld/starknet here 981b1092a3f7b2ecfb8acd75987cead9237960d0

antigremlin (approved):

paulbalaji:
got another PR here to fix the storage diff-check https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6290

PR #3900, opened 2024-06-05
fix: testnet4 holesky config

paulbalaji:
ran `yarn tsx ./scripts/agents/update-agent-config.ts -e testnet4` and updated testnet4 config


tkporter (approved):
ty

PR #3905, opened 2024-06-05
fix(cli): set default fallback owner to empty instead of prompting for warp config

nbayindirli:
### Description

- sets default fallback owner to empty instead of prompting for warp config

### Drive-by changes

- none

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3903

### Backward compatibility

- yes

### Testing

- ci-test
- manual

### DIFF
--- a/typescript/cli/src/config/ism.ts
+++ b/typescript/cli/src/config/ism.ts
@@ -252,11 +252,6 @@ export const createRoutingConfig = callWithConfigCreationLogs(
 
 export const createFallbackRoutingConfig = callWithConfigCreationLogs(
   async (context: CommandContext): Promise<IsmConfig> => {
-    const owner = await input({
-      message: 'Enter owner address for fallback routing ISM',
-    });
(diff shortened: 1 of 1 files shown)


typescript/cli/src/config/ism.ts
```diff
@@ -196,7 +191,7 @@ export async function createRoutingConfig(
   }
   return {
     type,
-    owner: ownerAddress,
+    owner: '',
```
jmrossy, line 194: Can you confirm, routing configs never have an owner now?

nbayindirli: only fallback has no owner:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/isms/routing/DefaultFallbackRoutingIsm.sol#L14-L30

new update/rebase reflects this

cmcewen (approved):

PR #3906, opened 2024-06-05
feat(cli): add 'source' to detectAndConfirmOrPrompt

nbayindirli:
### Description

- adds `source` field to detect and confirm prompt so users know where defaults are coming from

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- manual

### DIFF
--- a/typescript/cli/src/utils/chains.ts
+++ b/typescript/cli/src/utils/chains.ts
@@ -78,13 +78,16 @@ export async function detectAndConfirmOrPrompt(
   detect: () => Promise<string | undefined>,
   prompt: string,
   label: string,
+  source?: string,
 ): Promise<string> {
   let detectedValue: string | undefined;
   try {
     detectedValue = await detect();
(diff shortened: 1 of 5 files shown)


typescript/cli/src/utils/chains.ts
```diff
@@ -78,13 +78,16 @@ export async function detectAndConfirmOrPrompt(
...
   label: string,
+  source?: string,
 ): Promise<string> {
   let detectedValue: string | undefined;
   try {
     detectedValue = await detect();
     if (detectedValue) {
       const confirmed = await confirm({
-        message: `Detected ${label} as ${detectedValue}, is this correct?`,
+        message: `Detected ${label} as ${detectedValue}${
```
jmrossy, line 88: Nit: I think `Found...` is a bit easier to understand than `Detected` but def nbd

jmrossy (approved):

PR #3908, opened 2024-06-05
fix(cli): remove opStackIsm & pausableIsm from hyperlane core configure

nbayindirli:
### Description

- removes `opStackIsm` & `pausableIsm` from `hyperlane core configure`

### Drive-by changes

- none

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3907

### Backward compatibility

- yes

### Testing

- ci-test
- manual

### DIFF
--- a/typescript/cli/src/config/ism.ts
+++ b/typescript/cli/src/config/ism.ts
@@ -71,8 +71,6 @@ const ISM_TYPE_DESCRIPTIONS: Record<string, string> = {
   [IsmType.TRUSTED_RELAYER]: 'Deliver messages from an authorized address',
   [IsmType.TEST_ISM]:
     'ISM where you can deliver messages without any validation (WARNING: only for testing, do not use in production)',
-  [IsmType.OP_STACK]: '',
-  [IsmType.PAUSABLE]: '',
 };
 
 export async function createIsmConfig(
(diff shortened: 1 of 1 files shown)


typescript/cli/src/config/ism.ts
```diff
@@ -105,7 +103,7 @@ export async function createIsmConfig(
     return createTrustedRelayerConfig(context);
   }
 
-  throw new Error(`Invalid ISM type: ${moduleType}}`);
+  throw new Error(`Invalid ISM type: ${moduleType}`);
```
ltyu, line 106: ❤️

typescript/cli/src/config/ism.ts
```diff
@@ -71,8 +71,6 @@ const ISM_TYPE_DESCRIPTIONS: Record<IsmType, string> = {
   [IsmType.TRUSTED_RELAYER]: 'Deliver messages from an authorized address',
   [IsmType.TEST_ISM]:
     'ISM where you can deliver messages without any validation (WARNING: only for testing, do not use in production)',
-  [IsmType.OP_STACK]: '',
-  [IsmType.PAUSABLE]: '',
```
ltyu, line 75: isn't this just removing the description? the select is generated below in `createIsmConfig`, or am i missing something?

ltyu: maybe something like this?
```typescript
export async function createIsmConfig(
  context: CommandContext,
): Promise<IsmConfig> {
  const moduleType = await select({
    message: 'Select ISM type',
    choices: Object.entries(ISM_TYPE_DESCRIPTIONS).map(
      ([value, description]) => ({
        value,
        description,
      }),
    ),
    pageSize: 10,
  });
...
}
```

nbayindirli: `IsmType.OP_STACK` and `IsmType.PAUSABLE` are not handled by the if-else tree (another +1 to switch-cases :p ), so they `throw new Error(`Invalid ISM type: ${moduleType}`);` which we decided is poor UX. We do not want them present as options in the selection at all

paulbalaji (approved):

PR #3909, opened 2024-06-05
fix(cli): remove default for chain name

nbayindirli:
### Description

- simply prompts if not a lowercase one-word chain name, instead of sometimes returning incorrect chain name

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- manual w/ caldera: https://early-rabbits-throw.hub.caldera.xyz/

### DIFF
--- a/typescript/cli/src/config/chain.ts
+++ b/typescript/cli/src/config/chain.ts
@@ -1,7 +1,11 @@
 import { confirm, input } from '@inquirer/prompts';
 import { ethers } from 'ethers';
 
-import { ChainMetadata, ChainMetadataSchema } from '@hyperlane-xyz/sdk';
+import {
+  ChainMetadata,
+  ChainMetadataSchema,
+  ZChainName,
+} from '@hyperlane-xyz/sdk';
 import { ProtocolType } from '@hyperlane-xyz/utils';
 
 import { CommandContext } from '../context/types.js';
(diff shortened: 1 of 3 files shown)


typescript/cli/src/config/chain.ts
```diff
@@ -57,7 +57,8 @@ export async function createChainConfig({
       const clientName = await provider.send('web3_clientVersion', []);
       const port = rpcUrl.split(':').slice(-1);
       const client = clientName.split('/')[0];
-      return `${client}${port}`;
+      const chainName = `${client}${port}`;
```
yorhodes, line 60: I dont think the port will ever be letters
I did this so that anvil:8545 and anvil:8546 etc would not collide

nbayindirli: removed this entirely per our convo

typescript/cli/src/config/chain.ts
```diff
@@ -52,17 +52,10 @@ export async function createChainConfig({
...
-      const client = clientName.split('/')[0];
-      return `${client}${port}`;
-    },
-    'Enter (one word, lower case)',
-    'chain name',
-    'JSON RPC provider',
-  );
+  const name = await input({
+    message: 'Enter chain name (one word, lower case)',
+    validate: (chainName) => isLowercaseOneWord(chainName),
```
yorhodes, line 57: instead of regex, we should use our zod parser for chain names?

nbayindirli: love it– added `validate: (chainName) => ZChainName.safeParse(chainName).success`

jmrossy (approved):

cmcewen (approved):

Issue #3851, opened 2024-05-27
Implement XERC20 type token adapter


yorhodes:
## Problem

`TokenType.XERC20` and `TokenType.XERC20Lockbox` use the `EvmHypCollateral` token adapter type (used by `WarpCore` and in the UI). This nicely inherits token approval logic but fails when the destination chain is checked for sufficient collateral as neither of these `TokenRouter` variants actually custody the collateral token.

## Solution

- [ ] XERC20 adapter should check if the HypXERC20 has sufficient minting limits on the XERC20 for `handle`. This should take effect in `WarpCore.isDestinationCollateralSufficient`.
- [ ] XERC20 adapter should check if the HypXERC20 has sufficient burning limits on the XERC20 for `transferRemote`

jmrossy:
@AlexBHarley Here are a few pointers of where we'd need code additions:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/Token.ts#L208
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/Token.test.ts#L21
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/adapters/EvmTokenAdapter.ts

And just for reference:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/warp/WarpCore.ts#L411

PR #3911, opened 2024-06-05
feat(sdk): XERC20 token adapter

AlexBHarley:
### Description

`HypXERC20Adapter` that allows checking for mint and burn limits put in place on XERC20 contracts.

### Drive-by changes

- Couple methods added to the `XERC20` interface
- Corrects `HypXERC20Lockbox` and `HypXERC20` config during CLI deployments

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3851

### Backward compatibility

Yes

### Testing

CLI testing.


typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -44,6 +48,13 @@ export interface WarpCoreOptions {
   routeBlacklist?: RouteBlacklist;
 }
 
+function isXERC20Adapter<T>(x: ITokenAdapter<T>): x is IHypXERC20Adapter<T> {
+  return (
+    !!(x as any as IHypXERC20Adapter<T>).belowBurnLimit &&
+    !!(x as any as IHypXERC20Adapter<T>).belowMintLimit
+  );
+}
```
jmrossy, line 56: Rather than trying to introspect this way, branch based on the TokenStandard (i.e. if the standard is `XERC20` or `XERC20Lockbox`

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -433,6 +444,10 @@ export class WarpCore {
     }
 
     const adapter = destinationToken.getAdapter(this.multiProvider);
+    if (isXERC20Adapter(adapter)) {
```
jmrossy, line 447: Please restructure this so the `this.logger.debug` still runs

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -433,6 +444,10 @@ export class WarpCore {
...
+      return await adapter.belowMintLimit(originTokenAmount.amount);
```
jmrossy, line 448: For consistency, it might be nice for your new method to simply return the limit itself and let the caller compare like we do already below here

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -504,6 +519,17 @@ export class WarpCore {
     const amountError = this.validateAmount(originTokenAmount);
     if (amountError) return amountError;
 
+    const destinationCollateralError = await this.validateDestinationCollateral(
```
jmrossy, line 522: I'm open to validating collateral here, probably a good idea, but note we'll want to then remove the `isDestinationCollateralSufficient` call in useTokenTransfer in the warp UI to avoid running this twice.

jmrossy (commented):
Great progress!

typescript/sdk/src/warp/WarpCore.ts
AlexBHarley: Seems reasonable to me. 

IMO best to do all the validations we can before entering the transaction building flow and updating the UI to make it look like this will tx will work.

AlexBHarley: Also nice here because then the CLI benefits from these checks without changes

yorhodes (approved):
love it, great first PR 🚢

avious00:
congrats on your first official PR!

PR #3918, opened 2024-06-06
chore: restrict for how long the backward syncer can block

daniel-savu:

### DIFF
--- a/rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
+++ b/rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
@@ -14,7 +14,7 @@ use tracing::{debug, instrument, warn};
 
 use super::{LastIndexedSnapshot, TargetSnapshot};
 
-const MAX_BACWARD_SYNC_BLOCKING_TIME: Duration = Duration::from_secs(1);
+const MAX_BACKWARD_SYNC_BLOCKING_TIME: Duration = Duration::from_secs(5);
 
 /// A sequence-aware cursor that syncs backward until there are no earlier logs to index.
 pub(crate) struct BackwardSequenceAwareSyncCursor<T> {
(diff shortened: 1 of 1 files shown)


rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -9,10 +9,13 @@ use hyperlane_core::{
     HyperlaneSequenceAwareIndexerStoreReader, IndexMode, Indexed, LogMeta, SequenceIndexed,
 };
 use itertools::Itertools;
+use tokio::time::sleep;
 use tracing::{debug, instrument, warn};
 
 use super::{LastIndexedSnapshot, TargetSnapshot};
 
+const MAX_BACWARD_SYNC_BLOCKING_TIME: Duration = Duration::from_secs(1);
```
tkporter, line 17: nit: BACWARD -> BACKWARD

Wdyt about like 5 seconds? Because we're doing tokio::select! and I'm not sure how long one iteration of the skip_indexed loop is on the scraper, a little worried that 1 second could be too short to go through one iteration and we'd never make progress on the backward

Also wondering if this means we need to be careful about skip_indexed being cancel safe, maybe a comment there would be good

tkporter (approved):

rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
daniel-savu: > 5 seconds

was torn between 1 and 5 secs but thought 1 is more than enough for a couple of round trips to the db. Happy to go with 5

> cancellation safety

should be all good since we're cancelling a future rather than a task, which can only happen at an `await` point (https://blog.yoshuawuyts.com/async-cancellation-1/), so worst case we make the same `get_sequence_log_block_number` again when we resume

Issue #3541, opened 2024-04-03
Create CLI command `hyperlane warp deploy`


tkporter:
Blocked by:
- https://github.com/hyperlane-xyz/issues/issues/1156

Scope:
- Implement a command for the case of applying a hook config: `hyperlane warp apply`
- - CLI args:
    - `--config <path>` - A path to the full `ChainMap<TokenConfig>`. `-` means to read from stdin
    - `--artifacts <path>` - A path to the artifact file to read and update in the registry
    - `--chain <chain-name>` (optional): the name of a single chain to apply to
- As this is a deploy command, it'll first create and then update
- This takes in a ChainMap of warp configs (I guess `ChainMap<TokenConfig>` ?)

This could be done along with https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3479

PR #3920, opened 2024-06-06
feat: implement `hyperlane warp deploy`

ltyu:
### Description
- Update `hyperlane warp deploy` command

### Drive-by changes
- Move Core deploy to core.ts
- Update Module return args to match TokenType

### Related issues
- Fixes #[3541](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3541)

### Backward compatibility
yes, but old command no longer applies

### Testing
Manual


typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -84,7 +76,7 @@ export class EvmERC20WarpModule extends HyperlaneModule<
 
     return new EvmERC20WarpModule(multiProvider, {
       addresses: {
-        deployedTokenRoute: deployedContracts[config.type].address,
+        [config.type]: deployedContracts[config.type].address,
```
cmcewen, line 79: does this make type-ing harder to reason about?

typescript/cli/src/deploy/warp.ts
```diff
@@ -110,20 +112,30 @@ async function executeDeploy(params: DeployParams) {
...
+        multiProvider,
+      });
 
+      return evmERC20WarpModule.serialize();
+    }),
+  );
   logGreen('✅ Hyp token deployments complete');
 
-  const warpCoreConfig = await getWarpCoreConfig(params, deployedContracts);
+  const deployedContractMap = attachContractsMap(
```
cmcewen, line 133: can you do this inside the map above? or this function expects a map vs just attaching as single one at a time

ltyu: Yup, i will keep that in mind when i re-implement this in another PR

typescript/cli/src/commands/core.ts
```diff
@@ -34,6 +43,48 @@ export const coreCommand: CommandModule = {
...
+  chain: string;
+  config: string;
+  dryRun: string;
+  fromAddress: string;
+}> = {
+  command: 'deploy',
+  describe: 'Deploy Hyperlane contracts',
+  builder: {
+    chain: chainCommandOption,
+    config: outputFileCommandOption(
```
yorhodes, line 62: should this be input ?

yorhodes (approved):
lets revert the address shape change but otherwise lgtm

typescript/cli/src/commands/core.ts
ltyu: yes

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: reverted

typescript/cli/src/commands/core.ts
ltyu: Will address in a following PR. Requires refactoring of the input function, but also merging this for bug bash

PR #3927, opened 2024-06-07
ci: update action versions

paulbalaji:
- update to latest versions of actions
- notably docker tasks get a couple of major bumps
- rename the foundry action to its current name


paulbalaji:
> Looks good but can you check on the slither CI failure and see if it's because of these updates?

I did some experimenting last week trying to see which change in the workflow causes the problem, but I managed to hit the error even with an effectively unchanged workflow 

I'll re-add an mvp of that to the commit history on this PR for future viewing

jmrossy (approved):

PR #3929, opened 2024-06-10
fix: warp configure infinite loop

### LINKED ISSUE #3924: warp configure goes into an selection loop
## Problem
The command `hyperlane warp configure` forces the user to configure an ISM in simple mode. This causes an unexpected selection loop that is hard to escape.

## Solution
Consider removing ISM selection when calling `hyperlane warp configure` with out the `--ismAdvanced` flag. The flow should be
```
? Detected owner address as 0xa2558cA30cd99c2Ab0e2349C274a3736698bD60e from signer, is this correct? yes
? Select chains to connect ancient8, arbitrum
Configuring warp route for chain ancient8
? Select ancient8's token type collateral
? Detected mailbox address as 0xa2558cA30cd99c2Ab0e2349C274a3736698bD60e from hyperlane-registry, is this correct? yes
Configuring warp route for chain arbitrum
? Select arbitrum's token type collateral
? Detected mailbox address as 0x979Ca5202784112f4738403dBec5D0F3B9daabB9 from hyperlane-registry, is this correct? yes
```

ltyu:
### Description
- Fixes `hyperlane warp configure` ism selection looping by configuring a default Aggregation ISM with trusted relayer and fallback routing

### Related issues

- Fixes #3924 

### Backward compatibility
Yes

### Testing

Manual

### DIFF
--- a/typescript/cli/src/config/warp.ts
+++ b/typescript/cli/src/config/warp.ts
@@ -147,7 +147,7 @@ export async function createWarpRouteDeployConfig({
 
     const interchainSecurityModule = advanced
       ? await createAdvancedIsmConfig(context)
-      : await createDefaultWarpIsmConfig(owner);
+      : createDefaultWarpIsmConfig(owner);
 
     switch (type) {
       case TokenType.collateral:
@@ -209,7 +209,7 @@ export function readWarpRouteConfig(filePath: string): WarpCoreConfig {
  * @param owner - The address of the owner of the ISM.
  * @returns The default Aggregation ISM configuration.
  */
-async function createDefaultWarpIsmConfig(owner: Address): Promise<IsmConfig> {
+function createDefaultWarpIsmConfig(owner: Address): IsmConfig {
   return {
     type: IsmType.AGGREGATION,
     modules: [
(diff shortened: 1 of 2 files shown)


typescript/cli/src/config/warp.ts
```diff
@@ -205,14 +201,27 @@ export function readWarpRouteConfig(filePath: string): WarpCoreConfig {
...
-): Promise<IsmConfig> {
+/**
+ * Creates a default configuration for an ISM with a TRUSTED_RELAYER and FALLBACK_ROUTING.
+ *
+ * Properties relayer and owner are both set as input owner.
+ *
+ * @param owner - The address of the owner of the ISM.
+ * @returns The default Aggregation ISM configuration.
+ */
+async function createDefaultWarpIsmConfig(owner: Address): Promise<IsmConfig> {
```
jmrossy, line 212: Seems like this function isn't async anymore and need not return a promise, is that right?

jmrossy (commented):
LGTM just one small thing

typescript/cli/src/config/warp.ts
nbayindirli: +

nbayindirli (approved):
nice

typescript/cli/src/config/warp.ts
ltyu: yes indeed!

ltyu: Fixed

PR #3932, opened 2024-06-10
fix: slow relayer startup

daniel-savu:

### DIFF
--- a/rust/agents/relayer/src/msg/processor.rs
+++ b/rust/agents/relayer/src/msg/processor.rs
@@ -138,7 +138,10 @@ impl DirectionalNonceIterator {
     #[instrument]
     fn iterate(&mut self) {
         match self.direction {
-            NonceDirection::High => self.nonce = self.nonce.map(|n| n.saturating_add(1)),
+            NonceDirection::High => {
+                self.nonce = self.nonce.map(|n| n.saturating_add(1));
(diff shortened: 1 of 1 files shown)


rust/agents/relayer/src/msg/processor.rs
```diff
@@ -138,7 +138,10 @@ impl DirectionalNonceIterator {
     #[instrument]
     fn iterate(&mut self) {
         match self.direction {
-            NonceDirection::High => self.nonce = self.nonce.map(|n| n.saturating_add(1)),
+            NonceDirection::High => {
+                self.nonce = self.nonce.map(|n| n.saturating_add(1));
+                debug!(?self, "Iterating high nonce");
```
paulbalaji, line 143: What does it mean to iterate a high nonce

daniel-savu: the message processor uses a forward-backward strategy to find messages in the db (added by the indexer) and send them to the submitter. It stores a "high nonce" for the forward part, and a "low nonce" for the backward part. If the high nonce is being iterated, it means a new message was added to the db by the indexer, meaning we expect this message to be delivered soon. The nonces are initialized to the highest seen nonce (stored in the db), so if the relayer logs this shortly after startup, it means it's processing messages -> a sign of fast startup

paulbalaji (approved):

Issue #3720, opened 2024-05-07
Add a CLI command to generate an agent config from registry content


tkporter:
## Problem

See here for full context and design: https://www.notion.so/hyperlanexyz/Hybrid-HaaS-67195982e71b42de92bc83038498583a?pvs=4#a3831e1497b249028ae6bca6a70994a1

Here's the current state:
- `typecript/infra/scripts/update-agent-config.ts` is the current way in which we generate agent configs for our mainnet3 and testnet4 deploys
- Agent configs relating to these envs are shipped with the agents and live here https://github.com/hyperlane-xyz/hyperlane-monorepo/tree/main/rust/config
- We have CI steps that generate these using the current `main` of the registry https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4261df74bbbf9cfd281ba6139cd811f91e4166ed/.github/workflows/test.yml#L177-L225 and fail if they're different from the ones in the monorepo
- In the CLI 2.0, there's no way for CLI users to generate an agent config. This means that a CLI user who deployed to a new chain and has new chain metadata / addresses relating to their chain doesn't have a way to generate an agent config using this and run their own agents

## Solution

For now, we'll keep using infra's `update-agent-config.ts`, and we'll keep the existing CI steps for generating the mainnet3 and testnet4 agent configs. This is because of open design questions around how make sure that transaction overrides that live in infra still make their way into the agent config. This'll be tracked in a separate issue

Also note you'll want to build off the [`cli-2.0`](https://github.com/hyperlane-xyz/hyperlane-monorepo/tree/cli-2.0) branch and not `main`, unless cli-2.0 gets merged into main by the time this gets picked up

For this issue, we want:
- A CLI command that has very similar behavior as `update-agent-config.ts`. Something like `hyperlane registry agent-config`. It should be able to take `--chains <list of chains>`, where it generates an agent config for each of the provided chains, or alternatively be told to generate an agent config for either all testnets or mainnets.
- This command creates a config using the chain metadata and addresses in the CLI's registry. Look at other CLI commands to understand how to interact with a registry - the general idea is that the CLI uses the Github registry as a "base" registry, and then applies a local registry on top of that. It should be possible for anyone to run this CLI command pointing to a local copy of the registry and for this to spit out an agent config
- Should output the agent config to some file as a json

PR #3938, opened 2024-06-10
feat(cli): Support creating agent configs from CLI

Mo-Hussain:
### Description

- Add support for creating agent configs using the CLI
- registry agent-config command with a required --chains option
- This will pick up local registry data

Example usage:

`hyperlane registry agent-config --chains anvil8545`

### Related issues

- Fixes #[3720](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3720)

### Backward compatibility

Yes

### Testing

Manual


typescript/cli/agent-config-anvil8545.json
```diff
@@ -0,0 +1,32 @@
+{
```
Mo-Hussain, line 1: Added these config here temporarily so draft reviewer can eye ball

typescript/cli/src/config/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+  chains,
+  environment,
+  out,
+}: {
+  context: CommandContext;
+  chains?: string[];
+  environment?: ChainType;
+  out: string;
+}) {
+  logBlue(`\nCreating agent config...`);
```
nbayindirli, line 25: nit: avoid graves w/o interpolation

typescript/cli/src/config/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+  let agentChains;
+  if (chains) {
+    agentChains = chains;
+  } else {
+    const metadata = Object.values(chainMetadata).filter((c) => {
+      if (environment === 'mainnet') return !c.isTestnet;
+      else return !!c.isTestnet;
+    });
+    agentChains = metadata.map((c) => c.name);
+  }
```
nbayindirli, line 39: nit: taste thing
```suggestion
  let agentChains = chains;
  if (!agentChains) {
    const metadata = Object.values(chainMetadata).filter((c) => {
      if (environment === 'mainnet') return !c.isTestnet;
      else return !!c.isTestnet;
    });
    agentChains = metadata.map((c) => c.name);
  }
```

typescript/cli/src/config/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  // TODO: consider adding additional config used to pass in gas prices for Cosmos chains
+  const agentConfig = buildAgentConfig(
+    agentChains,
+    multiProvider,
+    addresses as ChainMap<HyperlaneDeploymentArtifacts>,
+    startBlocks,
+  );
+
+  writeYamlOrJson(out, agentConfig);
```
nbayindirli, line 73: please specify `'yaml'` as third argument

typescript/cli/src/commands/registry.ts
```diff
@@ -88,3 +92,88 @@ const addressesCommand: CommandModuleWithContext<{ name: string }> = {
...
+  alias: 'c',
+  // TODO choices: ... can we configure this so that it's a list of chain names including any new chains added to the local registry
+};
+
+const environmentCommandOption: Options = {
+  type: 'string',
+  description: 'The name of the environment to deploy to',
+  alias: 'e',
+  choices: ChainTypes,
+};
```
jmrossy, line 108: Recommend moving these to commands/options and de-duping with similar coreTargetsCommandOption

typescript/cli/src/commands/registry.ts
```diff
@@ -88,3 +92,88 @@ const addressesCommand: CommandModuleWithContext<{ name: string }> = {
...
+    const { registry } = context;
+
+    let chainNames;
+    if (chains) {
+      chainNames = chains.split(',');
+      const validChainNames = await registry.getChains();
+
+      const invalidChainNames = chainNames.filter(
+        (chainName) => !validChainNames.includes(chainName),
+      );
```
jmrossy, line 165: I think `multiProvider.hasChain` is what you're looking for here.
You can get the MP from context

typescript/cli/src/commands/registry.ts
```diff
@@ -1,10 +1,13 @@
...
 import { CommandModuleWithContext } from '../context/types.js';
-import { log, logBlue, logGray, logTable } from '../logger.js';
+import { log, logBlue, logGray, logRed, logTable } from '../logger.js';
 
-const ChainTypes = ['mainnet', 'testnet'];
-type ChainType = (typeof ChainTypes)[number];
+import { outputFileCommandOption } from './options.js';
+
+export const ChainTypes = ['mainnet', 'testnet'];
+export type ChainType = (typeof ChainTypes)[number];
```
jmrossy, line 10: This is a bigger question than you signed up for with this PR but do we definitely still need this mainnet/testnet differentiation? The direction we've been going with other tooling is to just treat chains equally. Feel free to punt on this

typescript/cli/src/config/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (indexFrom !== undefined) {
+        return indexFrom;
+      }
+
+      const mailbox = core.getContracts(chain).mailbox;
+      try {
+        const deployedBlock = await mailbox.deployedBlock();
+        return deployedBlock.toNumber();
+      } catch (err) {
+        // TODO: how to this in the CLI? should this be surfaced to the user as a warning, should it require an interaction?
```
jmrossy, line 56: Depends. Is this an error the command can tolerate? It may be better to simply print some more details and then fail here as it likely points to a config issue or a bad RPC url

jmrossy (commented):
Good progress!

typescript/cli/src/commands/registry.ts
Mo-Hussain: @daniel-savu do the agents rely on this differentiation?

typescript/cli/agent-config-anvil8545.json
```diff
@@ -0,0 +1,32 @@
...
+      "domainRoutingIsmFactory": "0x8004C1975cFd117e3aBeed4358500B4b7F9C3D5E",
+      "interchainSecurityModule": "0xe1a0292Ac8cf6b27C6E7b3254a9f1863a54E5829",
+      "domainRoutingIsm": "0xe1a0292Ac8cf6b27C6E7b3254a9f1863a54E5829",
+      "merkleTreeHook": "0x42599A4780784b92D02EfAA69DAC2DC401730eD9",
+      "protocolFee": "0xc3A82aFfe47c9F68a7D4e3B68a994E3192d2FCAB",
+      "testRecipient": "0xD7b09d3Ff66abd8bbf24c023A0Ae34298bc5CB6a",
+      "mailbox": "0xD8824cA6d7C87Ac747631121e45062F076609762",
+      "proxyAdmin": "0x96b768d7790A4dE644ab2Ada1A4D1aa79977ce11",
+      "validatorAnnounce": "0x326FcC17826b1B1EFC65Ea7Ea07c3A6Ea7E9d695",
+      "index": {
```
Mo-Hussain, line 26: @daniel-savu - what does this index do? Do we have a schema come where for this config?

daniel-savu: it's to tell the indexer which block to start scanning from, it's documented [here](https://docs.hyperlane.xyz/docs/operate/config-reference#chainschain_nameindexfrom)

typescript/cli/src/commands/registry.ts
daniel-savu: I'm not sure what this type is used for, can you share some context?

daniel-savu: fwiw the agent configs don't have different schemas between testnet/mainnet, just different values for the same config items

Mo-Hussain: This type will be used to decide whether to generate the config for testnet or mainnet chains

daniel-savu: as far as the agents are concerned, it doesn't matter whether the chain is a mainnet or testnet

typescript/cli/src/commands/registry.ts
```diff
@@ -105,52 +105,53 @@ const createAgentConfigCommand: CommandModuleWithContext<{
...
+      });
 
-  //   return yargs as any as Argv<{
-  //     chains?: string | undefined;
-  //     environment?: string | undefined;
-  //     out: string;
-  //   }>;
-  // },
-  handler: async ({ context, chains, environment, out }) => {
+    return yargs as Argv<{
```
Mo-Hussain, line 137: Doing a type assertion hear. This was the best way that I found to validate the --chains and --environment, ensuring that they are mutually exclusive.

Our `CommandModuleWithContext` type, which we need to pass context the the handle method, also means that the builder needs to return the context when defined as a callback in this way. If anyone has tackled type issues related to yargs,  open to discussing something different

typescript/cli/src/config/agent.ts
```diff
@@ -67,6 +70,16 @@ export async function createAgentConfig({
     startBlocks,
   );
 
+  try {
+    AgentConfigSchema.parse(agentConfig);
+  } catch (e) {
+    const error = fromError(e);
+    logRed(
+      `Agent config is invalid, this is possibly due to required contracts not being deployed. See details below:\n${error.toString()}`,
```
Mo-Hussain, line 78: Using `zod-validation-error` to show users a better error message, seems to be maintained.

An example error message

```
yarn hyperlane registry agent-config --chains anvil8545
Hyperlane CLI

Creating agent config...
Agent config is invalid, this is possibly due to required contracts not being deployed. See details below
Validation error: Required at "chains.anvil8545.interchainGasPaymaster"
```

.changeset/lovely-boxes-bow.md
```diff
@@ -0,0 +1,5 @@
+---
```
Mo-Hussain, line 1: Does this look right?

Mo-Hussain: Do minor

typescript/cli/src/commands/options.ts
```diff
@@ -106,6 +108,19 @@ export const agentConfigCommandOption = (
...
+export const chainTargetsCommandOption: Options = {
+  type: 'string',
+  description: 'Comma-separated list of chain names',
+  alias: 'c',
+};
+
+export const environmentCommandOption: Options = {
+  type: 'string',
+  description: 'The name of the environment to deploy to',
+  alias: 'e',
```
Mo-Hussain, line 120: Make an enum in follow up

typescript/cli/src/config/agent.ts
```diff
@@ -0,0 +1,86 @@
...
+        return indexFrom;
+      }
+
+      const mailbox = core.getContracts(chain).mailbox;
+      try {
+        const deployedBlock = await mailbox.deployedBlock();
+        return deployedBlock.toNumber();
+      } catch (err) {
+        warnYellow(
+          `Failed to get deployed block, defaulting to 0 for index for ${chain}`,
```
Mo-Hussain, line 58: Fail here, this is due to rpc issues or bad config

PR #3944, opened 2024-06-11
fix(cli): make native token required in chain config

nbayindirli:
### Description
- addresses P1 in https://www.notion.so/hyperlanexyz/CLI-Bug-Bash-5fc0833e51034baeba3beba9319160ff?pvs=4
- NOTE: Mimicking this behaviour: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/utils/balances.ts#L23

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4008

#### TODO:
- Once merged, this section of the quickstart guide should be removed: https://www.notion.so/hyperlanexyz/Hyperlane-Quickstart-Beta-4d9ff48720ee43fa8456a7d1517098dd?pvs=4#15534475fd8d4b398e51e9e0d527fb48
  - cc @avious00

### DIFF
--- a/typescript/cli/src/config/chain.ts
+++ b/typescript/cli/src/config/chain.ts
@@ -88,64 +88,10 @@ export async function createChainConfig({
     rpcUrls: [{ http: rpcUrl }],
   };
 
-  const wantAdvancedConfig = await confirm({
-    default: false,
-    message:
-      'Do you want to set block or gas properties for this chain config?',
-  });
-  if (wantAdvancedConfig) {
-    const wantBlockConfig = await confirm({
-      message: 'Do you want to add block config for this chain?',
-    });
-    if (wantBlockConfig) {
-      const blockConfirmation = await input({
-        message:
-          'Enter no. of blocks to wait before considering a transaction confirmed(0-500)',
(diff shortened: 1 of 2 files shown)


cmcewen (approved):
this makes sense but maybe think nativeToken should be required in the schema

jmrossy:
> this makes sense but maybe think nativeToken should be required in the schema

@cmcewen I'm open to that change but there are pros and cons. A benefit would be that NativeToken info is sometimes used in WarpCore, like for logic related to gas, so that would be more reliable if native token were mandatory. The main downside is that it makes the burden for a minimal chain config higher. Another downside is that on some Comsos chains, the line between what's **a** native token vs **the** native token is unclear.

.changeset/forty-lemons-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': minor
+---
+
+Use '--config' instead of '--warp' in `hyperlane warp send`.
```
ltyu, line 5: changeset should be different than the stack parent

PR #3956, opened 2024-06-13
ci: fix slither

paulbalaji:
- ci: try fix slither
- drive-by: fix a bunch of the warnings from running `yarn`


package.json
```diff
@@ -6,7 +6,7 @@
     "@trivago/prettier-plugin-sort-imports": "^4.2.1",
     "@typescript-eslint/eslint-plugin": "^7.4.0",
     "@typescript-eslint/parser": "^7.4.0",
-    "eslint": "^9.0.0",
+    "eslint": "^8.57.0",
```
jmrossy, line 9: Spoke with Paul who pointed out that the subpackages are still using 8.57. Better to be consistent.

jmrossy (approved):

PR #3957, opened 2024-06-13
chore: include 10% gas estimate bump in EvmIsmModule

paulbalaji:
- in the multiprovider we already do a 10% bump on the estimated gas, as a cover for unpredictable gas estimation
- we recently found that the ism factories do not do this, which caused problems in a recent hyperlane core deployment
- added this into the ism factories on `main`
- so we are ensuring that this change is properly reflected in the new ism module as well

### DIFF
--- a/typescript/sdk/src/deploy/EvmModuleDeployer.ts
+++ b/typescript/sdk/src/deploy/EvmModuleDeployer.ts
@@ -258,11 +258,20 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
         `Deploying new ${threshold} of ${values.length} address set to ${chain}`,
       );
       const overrides = multiProvider.getTransactionOverrides(chain);
-      const hash = await factory['deploy(address[],uint8)'](
+
+      // estimate gas
+      const estimatedGas = await factory.estimateGas['deploy(address[],uint8)'](
         values,
         threshold,
         overrides,
       );
+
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -259,11 +259,20 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
...
+      const estimatedGas = await factory.estimateGas['deploy(address[],uint8)'](
         values,
         threshold,
         overrides,
       );
+
+      // add 10% buffer
+      const hash = await factory['deploy(address[],uint8)'](values, threshold, {
+        ...overrides,
+        gasLimit: estimatedGas.add(estimatedGas.div(10)), // 10% buffer
```
jmrossy, line 273: Would `.mul(1.1)` work? Or maybe not because this is an Ethers BigNumber which doesn't support decimals? In which case, what happens if gas isn't divisible by 10?

paulbalaji: first time round i did try .mul(1.1) for simplicity but I ran into issues - I haven't thought about the not-divisible-by-10 problem before but we've also not run into it in the 4months we've had it in the multiprovider 🤷‍♂️

jmrossy (approved):

PR #3958, opened 2024-06-13
fix(cosmos): validator checkpoint & agent indexing improvements

daniel-savu:
- propagates RPC errors to the indexing logic, so errors cause sleeps instead of overwhelming the RPCs. This also greatly reduces cosmos log noisiness: from 9.5k logs per min, to 2.5k logs per min.
- checks for published checkpoints in reverse, to prioritize recent messages among those from before the validator was spun up
- fixes a validator bug where errors in checkpoint publishing would cause the entire checkpoint queue to be reiterated through, from scratch
- Builds on top of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3887 so we can test this out on osmosis


rust/chains/hyperlane-cosmos/src/mailbox.rs
```diff
@@ -376,15 +376,18 @@ impl Indexer<HyperlaneMessage> for CosmosMailboxIndexer {
...
-                    None
                 }
+                logs
             })
+            // Propagate errors from any of the queries. This will cause the entire range to be retried,
+            // including successful ones, but we don't have a way to handle partial failures in a range for now.
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
             .flatten()
-            .map(|(log, meta)| (log.into(), meta))
```
daniel-savu, line 387: I just copied over the existing implementation from the IGP indexer, this line does the same thing as `.into()` but not worth changing back now

rust/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -242,11 +239,9 @@ impl WasmIndexer for CosmosWasmIndexer {
         let client = self.provider.rpc().clone();
         debug!(?block_number, cursor_label, domain=?self.provider.domain, "Getting logs in block");
 
-        let (block, block_results) = tokio::join!(
```
daniel-savu, line 245: technically less efficient but cosmos RPCs don't handle load well anyway

tkporter (commented):
Did you intend for this PR to fix the tree issues we're seeing that we chatted about here? https://discord.com/channels/935678348330434570/1251219444332625963/1252280709565911120 or were you going to put it in a separate PR?

rust/chains/hyperlane-cosmos/src/merkle_tree_hook.rs
```diff
@@ -309,15 +309,18 @@ impl Indexer<MerkleTreeInsertion> for CosmosMerkleTreeHookIndexer {
...
-                Ok(logs) => Some(logs),
-                Err(err) => {
+            .map(|(logs, block_number)| {
+                if let Err(err) = &logs {
                     warn!(?err, ?block_number, "Failed to fetch logs for block");
-                    None
                 }
+                logs
             })
+            // Propagate errors from any of the queries. This will cause the entire range to be retried,
```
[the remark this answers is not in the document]
daniel-savu: since we propagate errors from indexing, the contract_sync actually gets to sleep [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/816924c388a15a04a68f67fd292383789c04cb94/rust/hyperlane-base/src/contract_sync/mod.rs#L160).

You can see that before I deployed this ([logs here](https://cloudlogging.app.goo.gl/k13MfkccZFPPgHud6)), that warning wasn't logged at all, so we'd just query the same range in a loop

rust/hyperlane-core/src/accumulator/incremental.rs
```diff
@@ -7,7 +7,7 @@ use crate::accumulator::{
     H256, TREE_DEPTH, ZERO_HASHES,
 };
 
-#[derive(BorshDeserialize, BorshSerialize, Debug, Clone, Copy, new, PartialEq, Eq)]
```
daniel-savu, line 10: Removed the `Copy` here since we should be explicit about it

rust/agents/validator/src/submit.rs
```diff
@@ -110,40 +113,38 @@ impl ValidatorSubmitter {
...
                 );
                 sleep(self.interval).await;
                 continue;
             }
 
-            tree = call_and_retry_indefinitely(|| {
-                let mut tree = tree;
+            call_and_retry_indefinitely(|| {
                 let self_clone = self.clone();
+                let tree = tree.clone();
```
tkporter, line 142: I think this may introduce a new bug

imagine the tree has size 100 and the latest checkpoint is 110, but an error occurs part way through `submit_checkpoints_until_correctness_checkpoint`. I think there could be two types of errors: (1) the tree is only able to be built up to size 105, or (2) the whole tree is built up to size 110, but when the signed checkpoints are submitted to s3, only some are successfully submitted to s3 and then an error occurs (e.g. 110 - 106 (reverse order now) are successful, but 105 isn't).

We'd get an error from `submit_checkpoints_until_correctness_checkpoint` and then retry, but we'd retry building upon the tree that was modified in that last call to `submit_checkpoints_until_correctness_checkpoint`. So in error case 1 we'd start building from 105 to 110, and we'd skip signing checkpoints 101 - 105 entirely. And in error case 2 we'd start building the tree from 110, meaning we permanently skip the checkpoints 105 - 101 that we didn't get to submit last time.

One way around this might be to copy tree here, use the copied tree in `submit_checkpoints_until_correctness_checkpoint`, and then after we know that submitting all the checkpoints in that function is successful, we write / commit that new tree back to the Arc<Mutex>'d tree

I think this bug is also present for the backward one :/ not sure my suggestion above makes sense for the backward case, so something else may be needed

daniel-savu: fixed by retrying indefinitely at the individual checkpoint submission step, so we're fixing both the merkle tree being out of sync with what's submitted to checkpoint storage, and the submitter rewinds due to rate limits

rust/agents/validator/src/submit.rs
```diff
@@ -182,7 +160,13 @@ impl ValidatorSubmitter {
         while tree.count() as u32 <= correctness_checkpoint.index {
             if let Some(insertion) = self
                 .message_db
-                .retrieve_merkle_tree_insertion_by_leaf_index(&(tree.count() as u32))?
+                .retrieve_merkle_tree_insertion_by_leaf_index(&(tree.count() as u32))
+                .unwrap_or_else(|_| {
+                    panic!(
+                        "Database is missing merkle insertion for index {}",
```
tkporter, line 166: nit: iiuc this returns a Result<Option<_>> where an Err is just a db error but Ok(None) indicates that the db is missing the merkle insertion for that index, which is expected in this code path - should the panic message indicate an error occurred (which is not expected) and share the err?

tkporter (approved):
lgtm 🙏 thanks for this

PR #3960, opened 2024-06-13
chore: add all new chains to `KnownHyperlaneDomain` struct

daniel-savu:
### Description

So validators can be run with `--originChainName <chainname>`, like in the docs [here](https://docs.hyperlane.xyz/docs/operate/validators/run-validators#running-the-binary)

### DIFF
--- a/rust/hyperlane-core/src/chain.rs
+++ b/rust/hyperlane-core/src/chain.rs
@@ -83,6 +83,18 @@ pub enum KnownHyperlaneDomain {
     Injective = 6909546,
     InEvm = 2525,
 
+    Ancient8 = 888888888,
+
+    Blast = 81457,
+
+    Mode = 34443,
+
+    Redstone = 690,
+
+    Viction = 88,
+
+    Zetachain = 7000,
+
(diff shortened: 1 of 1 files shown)


rust/hyperlane-core/src/chain.rs
```diff
@@ -83,6 +83,18 @@ pub enum KnownHyperlaneDomain {
...
+    Blast = 81457,
+
+    Mode = 34443,
+
+    Redstone = 690,
+
+    Viction = 88,
+
+    Zetachain = 7000,
+
```
paulbalaji, line 97: nit: can you create a ticket and tag me to reorder these + the below ones alphabetically (will make it easier for us to make sense of it as we keep adding chains)

paulbalaji (approved):

rust/hyperlane-core/src/chain.rs
daniel-savu: created here https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3961

PR #3966, opened 2024-06-14
feat(cli): CLI command with pre-flight checks for a validator

### LINKED ISSUE #3813: CLI command with pre-flight checks for a validator
## Problem

Partners should be able to run a CLI command that confirms their validator has been stood up correctly. This is part of the Hybrid HaaS process (https://www.notion.so/hyperlanexyz/Hybrid-HaaS-67195982e71b42de92bc83038498583a?pvs=4#ca11e1fe300940fc910ff5d8aff515ad) where we want to make sure that partners have set their validator up correctly before things are handed off to AW

## Solution

A CLI command that:
- Checks if a validator is announced, and outputs a link to the announced S3 bucket
- Checks that there are signed checkpoints in this S3 bucket
  - the latest checkpoint index should roughly match what's on chain (considering that due to reorg periods, validators don't necessarily sign the tip)

## Nice to have

It may be nice to think through how we may want to use this internally -- e.g. should it allow for multiple validator addresses to be supplied as well?
```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3976
```

Mo-Hussain:
### Description

- Add validator preFlightCheck command that requires a chain name and a comma separated list of validator addresses
- command will check if the validator has announced and log it's storage location and latest signed checkpoint index

example usage:

Happy path
```
$yarn hyperlane validator preflightCheck --chain ethereum --validators 0xCE327111035dd38698c92c3778884dBbB0ca8103,0x5AEd2fD5cC5F9749C455646C86B0Db6126CaFcbB
Hyperlane CLI

Latest check point index of incremental merkle tree: 6279

✅ Validator 0xCE327111035dd38698c92c3778884dBbB0ca8103 announced
storage location: s3://hyperlane-v3-validator-signatures-everstake-ethereum/us-east-2
latest checkpoint index: 6279
✅ Validator is signing the latest available checkpoint


✅ Validator 0x5AEd2fD5cC5F9749C455646C86B0Db6126CaFcbB announced
storage location: s3://hyperlane-hashkey-cloud-validator-mainnet/ap-southeast-1
latest checkpoint index: 6279
✅ Validator is signing the latest available checkpoint



✅ Validator pre flight check passed
```

unhappy path
```
$yarn hyperlane validator preflightCheck --chain ethereum --validators 0x87cF8A85465118AfF9Ec728CA157798201B1E368,0x68970726568F09422B4c988Ee03526F3993a702b                                          
Hyperlane CLI

Latest check point index of incremental merkle tree: 6279

✅ Validator 0x87cF8A85465118AfF9Ec728CA157798201B1E368 announced
storage location: s3://rc-mainnet3-ethereum-validator-2/us-east-1
latest checkpoint index: 537
❌ Validator is not signing the latest available checkpoint


❌ Validator 0x68970726568F09422B4c988Ee03526F3993a702b has not been announced


❌ Pre flight check failed:
Some validators have not been announced.
Some validators are not signing the latest available checkpoint
```

### Drive-by changes

- use preferred `errorRed` method in create agent config script

### Related issues

- Fixes #3813

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/cli/src/commands/options.ts
+++ b/typescript/cli/src/commands/options.ts
@@ -7,16 +7,6 @@ import { LogFormat, LogLevel } from '@hyperlane-xyz/utils';
 import { ENV } from '../utils/env.js';
 
 /* Global options */
-// TODO: add enum for aliases
-// const enum OptionAliases {
-//   'SkipConfirmation' = 'y',
-//   'RegistryUri' = 'r',
-//   'PrivateKey' = 'k',
-//   'OutputFile' = 'o',
-//   'InputFile' = 'i',
-//   'Chain' = 'c',
-//   'WarpConfig' = 'w',
-// }
 
 export const makeOptionRequired = (option: Options): Options => ({
   ...option,
--- a/typescript/cli/src/commands/options.ts
+++ b/typescript/cli/src/commands/options.ts
@@ -7,6 +7,21 @@ import { LogFormat, LogLevel } from '@hyperlane-xyz/utils';
 import { ENV } from '../utils/env.js';
 
 /* Global options */
+// TODO: add enum for aliases
+// const enum OptionAliases {
+//   'SkipConfirmation' = 'y',
+//   'RegistryUri' = 'r',
+//   'PrivateKey' = 'k',
+//   'OutputFile' = 'o',
+//   'InputFile' = 'i',
+//   'Chain' = 'c',
+//   'WarpConfig' = 'w',
+// }
+
+export const makeOptionRequired = (option: Options): Options => ({
+  ...option,
+  demandOption: true,
+});
 
 export const logFormatCommandOption: Options = {
   type: 'string',
@@ -148,6 +163,12 @@ export const dryRunCommandOption: Options = {
 export const chainCommandOption: Options = {
   type: 'string',
   description: 'The specific chain to perform operations with.',
+};
+
+export const validatorCommandOption: Options = {
+  type: 'string',
+  description: 'Comma separated list of validator addresses',
+  demandOption: true,
 };
 
 export const addressCommandOption = (
--- a/typescript/cli/src/commands/options.ts
+++ b/typescript/cli/src/commands/options.ts
@@ -8,7 +8,7 @@ import { ENV } from '../utils/env.js';
 
 /* Global options */
 
-export const makeOptionRequired = (option: Options): Options => ({
+export const demandOption = (option: Options): Options => ({
   ...option,
   demandOption: true,
 });
--- a/typescript/cli/src/commands/validator.ts
+++ b/typescript/cli/src/commands/validator.ts
@@ -2,12 +2,13 @@ import { CommandModule } from 'yargs';
 
 import {
   Address,
+  ProtocolType,
   isValidAddress,
   normalizeAddress,
 } from '@hyperlane-xyz/utils';
 
 import { CommandModuleWithContext } from '../context/types.js';
-import { log, logRed } from '../logger.js';
+import { errorRed, log } from '../logger.js';
 import { getValidatorAddress } from '../validator/address.js';
 import { checkValidatorSetup } from '../validator/preFlightCheck.js';
 
@@ -75,31 +76,44 @@ const preFlightCheckCommand: CommandModuleWithContext<{
     validators: validatorCommandOption,
   },
   handler: async ({ context, chain, validators }) => {
-    const { multiProvider } = context;
+    const { multiProvider, registry } = context;
 
     // validate chain
     if (!multiProvider.hasChain(chain)) {
-      logRed(`Chain ${chain} is not supported by the current configuration`);
+      errorRed(`❌ ${chain} is not supported by the current configuration`);
+      process.exit(1);
+    }
+
+    const chainMetadata = await registry.getChainMetadata(chain);
+    if (!chainMetadata) {
+      errorRed(`\n❌ Unable to fetch metadata for chain ${chain}. Exiting.`);
+      process.exit(1);
+    }
+
+    if (chainMetadata && chainMetadata.protocol !== ProtocolType.Ethereum) {
+      errorRed(
+        `\n❌ Validator pre flight check only supports EVM chains. Exiting.`,
+      );
       process.exit(1);
     }
 
     // validate validators addresses
-    // TODO: how do we handle non EVM addresses
-    // TODO: consider using set
     const validatorList = validators.split(',');
-    const invalidAddresses: string[] = [];
-    const validAddresses: Address[] = [];
+    const invalidAddresses: Set<Address> = new Set();
+    const validAddresses: Set<Address> = new Set();
 
     for (const address of validatorList) {
       if (isValidAddress(address)) {
-        validAddresses.push(normalizeAddress(address));
+        validAddresses.add(normalizeAddress(address));
       } else {
-        invalidAddresses.push(address);
+        invalidAddresses.add(address);
       }
     }
 
-    if (invalidAddresses.length > 0) {
-      logRed(`Invalid addresses: ${invalidAddresses.join(', ')}`);
+    if (invalidAddresses.size > 0) {
+      errorRed(
+        `❌ Invalid addresses: ${Array.from(invalidAddresses).join(', ')}`,
+      );
       process.exit(1);
     }
 
--- a/typescript/cli/src/commands/validator.ts
+++ b/typescript/cli/src/commands/validator.ts
@@ -1,8 +1,15 @@
 import { CommandModule } from 'yargs';
 
+import {
+  Address,
+  isValidAddress,
+  normalizeAddress,
+} from '@hyperlane-xyz/utils';
+
 import { CommandModuleWithContext } from '../context/types.js';
-import { log } from '../logger.js';
+import { log, logRed } from '../logger.js';
 import { getValidatorAddress } from '../validator/address.js';
+import { checkValidatorSetup } from '../validator/preFlightCheck.js';
 
 import {
   awsAccessKeyCommandOption,
@@ -10,13 +17,20 @@ import {
   awsKeyIdCommandOption,
   awsRegionCommandOption,
   awsSecretKeyCommandOption,
+  chainCommandOption,
+  makeOptionRequired,
+  validatorCommandOption,
 } from './options.js';
 
 // Parent command to help configure and set up Hyperlane validators
 export const validatorCommand: CommandModule = {
   command: 'validator',
   describe: 'Configure and manage Hyperlane validators',
-  builder: (yargs) => yargs.command(addressCommand).demandCommand(),
+  builder: (yargs) =>
+    yargs
+      .command(addressCommand)
+      .command(preFlightCheckCommand)
+      .demandCommand(),
   handler: () => log('Command required'),
 };
 
@@ -46,6 +60,50 @@ const addressCommand: CommandModuleWithContext<{
       bucket,
       keyId,
     });
+    process.exit(0);
+  },
+};
+
+const preFlightCheckCommand: CommandModuleWithContext<{
+  chain: string;
+  validators: string;
+}> = {
+  command: 'preflightCheck',
+  describe: 'Check the validator has announced correctly for a given chain',
+  builder: {
+    chain: makeOptionRequired(chainCommandOption),
+    validators: validatorCommandOption,
+  },
+  handler: async ({ context, chain, validators }) => {
+    const { multiProvider } = context;
+
+    // validate chain
+    if (!multiProvider.hasChain(chain)) {
+      logRed(`Chain ${chain} is not supported by the current configuration`);
+      process.exit(1);
+    }
+
+    // validate validators addresses
+    // TODO: how do we handle non EVM addresses
+    // TODO: consider using set
+    const validatorList = validators.split(',');
+    const invalidAddresses: string[] = [];
+    const validAddresses: Address[] = [];
+
+    for (const address of validatorList) {
+      if (isValidAddress(address)) {
+        validAddresses.push(normalizeAddress(address));
+      } else {
+        invalidAddresses.push(address);
+      }
+    }
+
+    if (invalidAddresses.length > 0) {
+      logRed(`Invalid addresses: ${invalidAddresses.join(', ')}`);
+      process.exit(1);
+    }
+
+    await checkValidatorSetup(context, chain, validAddresses);
     process.exit(0);
   },
 };
--- a/typescript/cli/src/commands/validator.ts
+++ b/typescript/cli/src/commands/validator.ts
@@ -3,8 +3,8 @@ import { CommandModule } from 'yargs';
 import {
   Address,
   ProtocolType,
-  isValidAddress,
-  normalizeAddress,
+  isValidAddressEvm,
+  normalizeAddressEvm,
 } from '@hyperlane-xyz/utils';
 
 import { CommandModuleWithContext } from '../context/types.js';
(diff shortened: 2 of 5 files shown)


typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,102 @@
...
+      storageLocations = await validatorAnnounce.getAnnouncedStorageLocations([
+        validator,
+      ]);
+      const s3StorageLocation = storageLocations[0][0];
+
+      const s3Validator = await S3Validator.fromStorageLocation(
+        s3StorageLocation,
+      );
+
+      const latestCheckpointIndex =
```
Mo-Hussain, line 76: This assumes that we will always be setting the storage location as an S3 bucket which I assume is not always the case.

Do we want support being able to read local checkpoint files for those running the validator locally?

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+    const [_, latestCheckpointIndex] = await merkleTreeHook[
+      'latestCheckpoint()'
+    ]();
+
+    merkleTreeLatestCheckpointIndex = latestCheckpointIndex;
+    logBlue(
+      `\nLatest check point index of incremental merkle tree: ${merkleTreeLatestCheckpointIndex}\n`,
+    );
+  } catch (err) {
+    warnYellow(
```
Mo-Hussain, line 36: Should we fail if this happens?

typescript/cli/src/commands/validator.ts
```diff
@@ -49,3 +64,60 @@ const addressCommand: CommandModuleWithContext<{
     process.exit(0);
   },
 };
+
+const preFlightCheckCommand: CommandModuleWithContext<{
+  chain: string;
+  validators: string;
+}> = {
+  command: 'preflightCheck',
```
jmrossy, line 72: Recommend renaming this to simply `check` users need not type a long (possibly case senstitive, can't remember) command name.

typescript/cli/src/commands/validator.ts
```diff
@@ -49,3 +64,60 @@ const addressCommand: CommandModuleWithContext<{
...
+  builder: {
+    chain: makeOptionRequired(chainCommandOption),
+    validators: validatorCommandOption,
+  },
+  handler: async ({ context, chain, validators }) => {
+    const { multiProvider, registry } = context;
+
+    // validate chain
+    if (!multiProvider.hasChain(chain)) {
+      errorRed(`❌ ${chain} is not supported by the current configuration`);
```
jmrossy, line 83:
```suggestion
      errorRed(`❌ No metadata found for ${chain}. Ensure it is included in your configured registry.`);
```

typescript/cli/src/commands/validator.ts
```diff
@@ -49,3 +64,60 @@ const addressCommand: CommandModuleWithContext<{
...
+      process.exit(1);
+    }
+
+    const chainMetadata = await registry.getChainMetadata(chain);
+    if (!chainMetadata) {
+      errorRed(`\n❌ Unable to fetch metadata for chain ${chain}. Exiting.`);
+      process.exit(1);
+    }
```
jmrossy, line 91: You don't need this check. If the MP knows the chain, then it must have it's chain metadata.

It's fine to query the registry if you prefer but you can also get the metadata synchronously from the MP via `multiprovider.getChainMetadata(chain)`

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  const confirmedValidators = new Set<Address>();
+  const unconfirmedValidators = new Set<Address>();
+
+  for (const validator of validators) {
+    const matches = announcedValidators.filter((address) =>
+      eqAddress(address, validator),
+    );
+    if (matches.length === 0) {
```
jmrossy, line 58: Tip:
`if ( announcedValidators.find( a => eqAddress(a, validator) )`

jmrossy (commented):
Mostly looks good, just a few small comments. Can't comment on the UX questions though, maybe ask folks more familiar with the requirements here

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  if (unconfirmedValidators.size > 0) {
+    errorSet.add('Some validators have not been announced.');
+  }
+
+  let failedToReadError = false;
+  for (const validator of confirmedValidators) {
+    let storageLocations;
+    try {
+      storageLocations = await validatorAnnounce.getAnnouncedStorageLocations([
```
tkporter, line 76: nit: instead of n requests and the `getAnnouncedValidators` above, you can get all the info in a single RPC. This is what the relayer does when it's fetching announcements when relaying. Might be useful to avoid rate limiting
```
const validatorStorageLocations = await validatorAnnounce.getAnnouncedStorageLocations(validators);
// Now iterate through validatorStorageLocations, where indices correspond to
// validators, and any validators that haven't been announced will have an empty array
// as their listed storage locations
```

Example output with cast:
```
$ cast call 0xCe74905e51497b4adD3639366708b821dcBcff96 'getAnnouncedStorageLocations(address[])(string[][])' --rpc-url https://ethereum.publicnode.com '[0x94438a7de38d4548ae54df5c6010c4ebc5239eae,0x0000000000000000000000000000000000000000]'
[["s3://dsrv-hyperlane-v3-validator-signatures-validator1-ethereum/eu-central-1"], []]
```

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+          );
+        }
+      } else {
+        warnYellow(
+          `❗️ Cannot compare validator checkpoint signatures to latest check point \n`,
+        );
+      }
+    } catch (err) {
+      errorRed(
+        `❌ Failed to fetch storage locations for validator ${validator}, this may be due to the storage location not being an S3 bucket\n\n`,
```
tkporter, line 116: would be good to share the announcement in here

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+            `Some validators are not signing the latest available checkpoint\n`,
+          );
+        } else {
+          logBlue(
+            `✅ Validator is signing the latest available checkpoint\n\n`,
+          );
+        }
+      } else {
+        warnYellow(
+          `❗️ Cannot compare validator checkpoint signatures to latest check point \n`,
```
tkporter, line 111:
```suggestion
          `❗️ No latest signed checkpoint found for validator\n`,
```
maybe to be more clear what happened

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+    );
+    if (matches.length === 0) {
+      unconfirmedValidators.add(validator);
+    } else {
+      confirmedValidators.add(validator);
+    }
+  }
+
+  const errorSet = new Set<string>();
+  errorSet.add('Validator pre flight check failed:');
```
tkporter, line 66: nit: found this slightly confusing as I'd expect the errorSet to be empty if there were no errors instead of having size 1 - maybe this could be just a log message on line 131?

typescript/cli/src/validator/preFlightCheck.ts
tkporter: for posterity, chatted offline and S3 only is fine given that we don't have any local storage stuff implemented in typescript and the main use case of this command will be for production validators

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      const s3Validator = await S3Validator.fromStorageLocation(
+        s3StorageLocation,
+      );
+
+      const latestCheckpointIndex =
+        await s3Validator.getLatestCheckpointIndex();
+
+      logBlue(
+        `✅ Validator ${validator} announced\nstorage location: ${s3StorageLocation}\nlatest checkpoint index: ${latestCheckpointIndex}`,
```
tkporter, line 89: an additional thing to print that would be nice:
our storage location format in the announcement is a bespoke uri format, but often when debugging I navigate to the s3 bucket's webpage URL to poke around. It looks like this: https://dsrv-hyperlane-v3-validator-signatures-validator1-ethereum.s3.eu-central-1.amazonaws.com

Could you add a function to `S3Validator` to get the URL formatted like that and also log it here? sorry for the scope increase

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,136 @@
...
+      failedToReadError = true;
+    }
+  }
+
+  for (const validator of unconfirmedValidators) {
+    errorRed(`❌ Validator ${validator} has not been announced\n`);
+  }
+
+  if (failedToReadError) {
+    errorSet.add('Failed to fetch storage locations for some validators.');
```
tkporter, line 127: love these error recap logs at the end

typescript/cli/src/commands/validator.ts
tkporter, line 72:
```suggestion
  command: 'preflight-check',
```
Seems the pattern so far has been `foo-bar` for multi word commands in the CLI

tkporter (commented):
great job, looks great - some smaller suggestions

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: As in make this an error instead of a warning?

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: I'm not sure what you mean, successful announcement is log on ln 85

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: We cant check whether the latest signed checkpoint for the validator is up to date because we were unable to read the latest checkpoint on `merkleTreeHook`, so I think `No latest signed checkpoint found for validator` also doesn't make it clear. 

`Latest checkpoint could not be read`?

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: I'll do this as a follow up

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: I did see this method but was unsure as to whether the response would guarantee ordering. Will make the change

typescript/cli/src/validator/preFlightCheck.ts
tkporter: ah I meant putting `err` in the message

typescript/cli/src/validator/preFlightCheck.ts
tkporter: Ahh sorry I misread this. In that case then indicating why we can't compare and saying that the onchain latest checkpoint couldn't be read makes sense to me

typescript/cli/src/validator/preFlightCheck.ts
tkporter: yep sorry I had meant to remove this comment in my review

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -24,111 +24,89 @@ export const checkValidatorSetup = async (
...
-    announcedValidators = await validatorAnnounce.getAnnouncedValidators();
-  } catch (err) {
-    errorRed(
-      `❌ Failed to fetch announced validators for ${chain}. Exiting.\n`,
-    );
-    process.exit(1);
+    validatorStorageLocations =
+      await validatorAnnounce.getAnnouncedStorageLocations(validatorsArray);
+  } catch (e) {
+    errorSet.add('Failed to read announced storage locations on chain.');
```
Mo-Hussain, line 48: Surface a separate error for failing to read on-chain, put e in the message also @tkporter?

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -24,63 +24,53 @@ export const checkValidatorSetup = async (
...
-        s3StorageLocation,
-      );
+      const s3StorageLocation = storageLocations[0];
+
+      let s3Validator: S3Validator;
+      try {
+        s3Validator = await S3Validator.fromStorageLocation(s3StorageLocation);
+      } catch (e) {
+        errorRed(
+          `❌ Failed to fetch storage locations for validator ${validator}, this may be due to the storage location not being an S3 bucket\n\n`,
```
Mo-Hussain, line 69: Surface a separate error for failing to S3 bucket

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -108,27 +98,18 @@ export const checkValidatorSetup = async (
         }
       } else {
         warnYellow(
-          `❗️ Cannot compare validator checkpoint signatures to latest check point \n`,
+          `❗️ Cannot compare validator checkpoint signatures to latest checkpoint in the incremental merkletree, merkletree checkpoint could not be read\n`,
```
Mo-Hussain, line 101: More informative error message

typescript/cli/src/commands/validator.ts
```diff
@@ -49,3 +64,58 @@ const addressCommand: CommandModuleWithContext<{
     process.exit(0);
   },
 };
+
+const preFlightCheckCommand: CommandModuleWithContext<{
+  chain: string;
+  validators: string;
+}> = {
+  command: 'check',
```
tkporter, line 72: maybe worth renaming the file from preFlightCheck.ts to check.ts too for consistency

typescript/cli/src/validator/preFlightCheck.ts
```diff
@@ -0,0 +1,117 @@
+import { MerkleTreeHook__factory } from '@hyperlane-xyz/core';
+import { HyperlaneCore, S3Validator } from '@hyperlane-xyz/sdk';
+import { Address } from '@hyperlane-xyz/utils';
+
+import { CommandContext } from '../context/types.js';
+import { errorRed, logBlue, logGreen, warnYellow } from '../logger.js';
+
+export const checkValidatorSetup = async (
```
tkporter, line 8: similar to what i said in the other PR, maybe useful to split up chunks in here to other functions just to help readability with all the different branching, but don't feel obligated to

tkporter (approved):
looks great 🙏

typescript/cli/src/commands/options.ts
```diff
@@ -150,6 +155,12 @@ export const chainCommandOption: Options = {
   description: 'The specific chain to perform operations with.',
 };
 
+export const validatorCommandOption: Options = {
+  type: 'string',
+  description: 'Comma separated list of validator addresses',
```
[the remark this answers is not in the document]
Mo-Hussain: Right now when we pass an argument to `builder`, we use this type `{ [key: string]: Options }`. We will likely run into type issues if we go with this option as we use our own type for commands

```
export type CommandModuleWithWriteContext<Args> = CommandModule<
  {},
  Args & { context: WriteCommandContext }
>;
```

We use this type because we want to access this `context` in the `handler` callback. I do agree that this option does give us some nice features. Happy to explore whether we can make this work as a follow up

typescript/cli/src/validator/preFlightCheck.ts
Mo-Hussain: I will be pulling out some of this logic into functions to be reused by the AVS validator check command

typescript/cli/src/commands/options.ts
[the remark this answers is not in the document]
tkporter: happy to leave this to future potential work / maybe kick the can to the product team as they've also chosen to go with comma separation

PR #3968, opened 2024-06-14
feat: add stdout.rows to pagesize calculation with DEFAULT_PAGE_SIZE

ltyu:
### Description
- Add process.stdout.rows to pageSize adjustment for chain selection
- Defaults to 15 if process.stdout.rows < 0

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/utils/chains.ts
+++ b/typescript/cli/src/utils/chains.ts
@@ -9,6 +9,7 @@ import { log, logRed, logTip } from '../logger.js';
 // A special value marker to indicate user selected
 // a new chain in the list
 const NEW_CHAIN_MARKER = '__new__';
+const DEFAULT_PAGE_SIZE = 15;
 
 export async function runSingleChainSelectionStep(
   chainMetadata: ChainMap<ChainMetadata>,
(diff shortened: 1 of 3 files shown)


typescript/cli/src/utils/chains.ts
```diff
@@ -18,7 +19,7 @@ export async function runSingleChainSelectionStep(
   const chain = (await select({
     message,
     choices,
-    pageSize: 30,
+    pageSize: process.stdout.rows - 2 || DEFAULT_PAGE_SIZE, // subtract 2, to exclude previous prompt text
```
jmrossy, line 22: There probably should be a minimum. If process.stdout.rows is 0, we would pass in a page size of -2 here

jmrossy (commented):
Nice change! Needs a tweak though I think

typescript/cli/src/utils/chains.ts
ltyu: Yeah, the intention for `DEFAULT_PAGE_SIZE` was to do that, but it looks like thats not working as intended. Fixed

ltyu: Added calculatePageSize fn

typescript/cli/src/utils/chains.ts
```diff
@@ -10,6 +10,15 @@ import { log, logRed, logTip } from '../logger.js';
 // a new chain in the list
 const NEW_CHAIN_MARKER = '__new__';
 
+// Returns a dynamic pageSize based on process.stdout.rows
+const calculatePageSize = (
```
jmrossy, line 14: Recommend moving this to a util file not specific to `chains`

ltyu: move to `cli-options.ts`, which will hold functions to manipulate the cli function options

PR #3971, opened 2024-06-15
Support priorityFee fetching from RPC and some better logging

nambrot:
### Description

Like title says, this PR allows the SmartProvider to fetch feeData from the RPC, if supported. Otherwise it falls back to the existing 1.5 gwei hardcoded value that ethers.js v5 has. On most L2s, that is a gross overestimate, resulting in so much wasted ETH.

### Drive-by changes

Some additional logging

### Backward compatibility

Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -142,9 +142,12 @@ async function executeDeploy(params: DeployParams) {
     ismFactoryDeployer,
   );
 
+  logGreen('✅ ISM deployments complete');
+  log(`Deploying Warp contracts`);
+
   const deployedContracts = await deployer.deploy(modifiedConfig);
 
-  logGreen('✅ Hyp token deployments complete');
+  logGreen('✅ Warp contract deployments complete');
 
   const warpCoreConfig = await getWarpCoreConfig(params, deployedContracts);
   if (!isDryRun) {
--- a/typescript/sdk/src/providers/SmartProvider/HyperlaneJsonRpcProvider.ts
+++ b/typescript/sdk/src/providers/SmartProvider/HyperlaneJsonRpcProvider.ts
@@ -32,6 +32,13 @@ export class HyperlaneJsonRpcProvider
     super(rpcConfig.connection ?? rpcConfig.http, network);
   }
 
+  prepareRequest(method: string, params: any): [string, any[]] {
+    if (method === ProviderMethod.MaxPriorityFeePerGas) {
+      return ['eth_maxPriorityFeePerGas', []];
+    }
+    return super.prepareRequest(method, params);
+  }
+
   async perform(method: string, params: any, reqId?: number): Promise<any> {
     if (this.options?.debug)
       this.logger.debug(
@@ -149,6 +156,10 @@ export class HyperlaneJsonRpcProvider
     }
 
     return combinedResults;
+  }
+
+  get maxPriorityFeePerGas(): Promise<BigNumber> {
+    return this.perform('eth_maxPriorityFeePerGas', {});
   }
 
   getBaseUrl(): string {
--- a/typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
+++ b/typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
@@ -1,4 +1,5 @@
-import { providers } from 'ethers';
+import { BigNumber, providers } from 'ethers';
+import { resolveProperties } from 'ethers/lib/utils.js';
 import { Logger } from 'pino';
 
 import {
@@ -95,6 +96,41 @@ export class HyperlaneSmartProvider
     }
 
     this.supportedMethods = [...supportedMethods.values()];
+  }
+
+  async getPriorityFee() {
+    try {
+      return BigNumber.from(await this.perform('maxPriorityFeePerGas', {}));
+    } catch (error) {
+      return BigNumber.from('1500000000');
+    }
+  }
+
+  async getFeeData(): Promise<providers.FeeData> {
+    // override hardcoded getFeedata
+    const { block, gasPrice } = await resolveProperties({
+      block: this.getBlock('latest'),
+      gasPrice: this.getGasPrice().catch((error) => {
+        // @TODO: Why is this now failing on Calaveras?
+        //console.log(error);
+        return null;
+      }),
+    });
+
+    let lastBaseFeePerGas = null,
+      maxFeePerGas = null,
+      maxPriorityFeePerGas = null;
+
+    if (block && block.baseFeePerGas) {
+      // We may want to compute this more accurately in the future,
+      // using the formula "check if the base fee is correct".
+      // See: https://eips.ethereum.org/EIPS/eip-1559
+      lastBaseFeePerGas = block.baseFeePerGas;
+      maxPriorityFeePerGas = await this.getPriorityFee();
+      maxFeePerGas = block.baseFeePerGas.mul(2).add(maxPriorityFeePerGas);
+    }
+
+    return { lastBaseFeePerGas, maxFeePerGas, maxPriorityFeePerGas, gasPrice };
   }
 
   static fromChainMetadata(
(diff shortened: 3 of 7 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -142,9 +142,12 @@ async function executeDeploy(params: DeployParams) {
     ismFactoryDeployer,
   );
 
+  logGreen('✅ ISM deployments complete');
+  log(`Deploying Warp contracts`);
+
   const deployedContracts = await deployer.deploy(modifiedConfig);
 
-  logGreen('✅ Hyp token deployments complete');
+  logGreen('✅ Warp contract deployments complete');
```
jmrossy, line 150:
```suggestion
  logGreen('✅ Warp Route contract deployments complete');
```

typescript/sdk/src/providers/SmartProvider/HyperlaneJsonRpcProvider.ts
```diff
@@ -151,6 +158,10 @@ export class HyperlaneJsonRpcProvider
     return combinedResults;
   }
 
+  get maxPriorityFeePerGas(): Promise<BigNumber> {
```
jmrossy, line 161: Avoid `get/set` class member keywords.

typescript/sdk/src/providers/SmartProvider/HyperlaneJsonRpcProvider.ts
```diff
@@ -32,6 +32,13 @@ export class HyperlaneJsonRpcProvider
     super(rpcConfig.connection ?? rpcConfig.http, network);
   }
 
+  prepareRequest(method: string, params: any): [string, any[]] {
+    if (method === ProviderMethod.MaxPriorityFeePerGas) {
+      return ['eth_maxPriorityFeePerGas', []];
+    }
```
jmrossy, line 38: Why special case this?

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -1,4 +1,5 @@
-import { providers } from 'ethers';
+import { BigNumber, providers } from 'ethers';
+import { resolveProperties } from 'ethers/lib/utils.js';
```
jmrossy, line 2: incorrect import. This will screw up tree shaking and can cause other problems. I would guess that importing `utils` and then calling `utils.resolveProperties` is what you're looking for

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -97,6 +98,41 @@ export class HyperlaneSmartProvider
     this.supportedMethods = [...supportedMethods.values()];
   }
 
+  async getPriorityFee() {
+    try {
+      return BigNumber.from(await this.perform('maxPriorityFeePerGas', {}));
+    } catch (error) {
+      return BigNumber.from('1500000000');
```
jmrossy, line 105: The catch will prevent the retry logic. But if this is the logic you want anyway, add logging and move the number to a const

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -97,6 +98,41 @@ export class HyperlaneSmartProvider
...
+    }
+  }
+
+  async getFeeData(): Promise<providers.FeeData> {
+    // override hardcoded getFeedata
+    const { block, gasPrice } = await resolveProperties({
+      block: this.getBlock('latest'),
+      gasPrice: this.getGasPrice().catch((error) => {
+        // @TODO: Why is this now failing on Calaveras?
+        //console.log(error);
```
jmrossy, line 115: cleanup

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -97,6 +98,41 @@ export class HyperlaneSmartProvider
...
+        return null;
+      }),
+    });
+
+    let lastBaseFeePerGas = null,
+      maxFeePerGas = null,
+      maxPriorityFeePerGas = null;
```
jmrossy, line 122: types?

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -97,6 +98,41 @@ export class HyperlaneSmartProvider
...
+
+    if (block && block.baseFeePerGas) {
```
jmrossy, line 124:
```suggestion
    if (block?.baseFeePerGas) {
```

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -97,6 +98,41 @@ export class HyperlaneSmartProvider
...
+    } catch (error) {
+      return BigNumber.from('1500000000');
+    }
+  }
+
+  async getFeeData(): Promise<providers.FeeData> {
+    // override hardcoded getFeedata
+    const { block, gasPrice } = await resolveProperties({
+      block: this.getBlock('latest'),
+      gasPrice: this.getGasPrice().catch((error) => {
```
jmrossy, line 113: Are you certain we want to swallow errors here?

typescript/sdk/src/token/deploy.ts
```diff
@@ -83,58 +83,69 @@ abstract class TokenDeployer<
...
-        ]);
-
-        return { name, symbol, totalSupply: totalSupply.toString(), decimals };
       }
-    }
 
-    return undefined;
+      return undefined;
+    } catch (error) {
+      console.error(`Could not derive token metadata`, configMap, error);
```
jmrossy, line 138: use `this.logger`, not console

typescript/sdk/src/token/deploy.ts
```diff
@@ -83,58 +83,69 @@ abstract class TokenDeployer<
...
-          erc20.decimals(),
-        ]);
-
-        return { name, symbol, totalSupply: totalSupply.toString(), decimals };
       }
-    }
 
-    return undefined;
+      return undefined;
+    } catch (error) {
```
jmrossy, line 137: It may have been intentional to let this error propagate before. Are you certain the deployment should proceed even if we can't derive metadata in the way we expected to? Also these changes seem unrelated to SmartProvider and should preferably be in a separate PR

typescript/sdk/src/providers/SmartProvider/HyperlaneJsonRpcProvider.ts
nambrot: Just extending the functionality from the base json rpc provider in

https://github.com/ethers-io/ethers.js/blob/v5/packages/providers/src.ts/json-rpc-provider.ts#L555

typescript/sdk/src/token/deploy.ts
yorhodes: heavy agree with this, this is inconsistent now with how the schema assesses whether we have "sufficient" metadata to deploy

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
nambrot: Doesn't this perform at the SmartProvider level? so the retries happen inside `perform`?

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
nambrot: Copied from the upstream implementation and would keep it as is to avoid deviations

typescript/sdk/src/token/deploy.ts
nambrot: This is a `static` function so this.logger is not available?

typescript/sdk/src/token/deploy.ts
nambrot: Yes sorry, re-throwing the error

nambrot:
Closing in favor of #3971

typescript/sdk/src/token/deploy.ts
jmrossy: Then use rootLogger

PR #3973, opened 2024-06-16
Do not consider xERC20 a collateral standard 

nambrot:
### Description

Here is another funky issue, @AlexBHarley added the xerc20 tokens to `TOKEN_COLLATERALIZED_STANDARDS` https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3911/files#diff-b98aa4b2eb3c8040524fa7f6d52d0fbbbc89fbac07485305664832f1742212aaR97 to be able to support collateral checking. However, that inadvertently broke our fungibility logic at https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/51bfff683e903a217efc15ae566c6e713c428b42/typescript/sdk/src/token/Token.ts#L416

Specifically, there is an assumption that any token config that is a collateral standard has `collateralAddressOrDenom`, which is not the case for `HypXERC20` since it doesn't really keep things in collateral (but conceptually it kinda does since mint/burn limits should be checked).

I don't feel strongly either way and would leave it to @jmrossy and @AlexBHarley  to hash it out, but for now, here is a PR that has more special cased (ugly) logic for destination "collateral" checking and removes it from `TOKEN_COLLATERALIZED_STANDARDS`

### DIFF
--- a/.changeset/olive-geckos-behave.md
+++ b/.changeset/olive-geckos-behave.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/sdk": patch
+---
+
+Do not consider xERC20 a collateral standard to fix fungibility checking logic while maintaining mint limit checking
--- a/typescript/sdk/src/token/TokenStandard.ts
+++ b/typescript/sdk/src/token/TokenStandard.ts
@@ -94,12 +94,15 @@ export const TOKEN_NFT_STANDARDS = [
 export const TOKEN_COLLATERALIZED_STANDARDS = [
   TokenStandard.EvmHypCollateral,
   TokenStandard.EvmHypNative,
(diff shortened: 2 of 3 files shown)


jmrossy (commented):
This solution lgtm **if** there's isn't a situation where we need that `isFungibleWith` in Token to trigger for the xERC20 standards. In other words, if a fee (IGP or local) will be quoted in them, then the calculation of max amounts (e.g. user hits max) may be broken by this.

AlexBHarley:
Closing in favour of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3977. XERC20 being collateral should have never been merged in hindsight, XERC20Lockbox sure, but not XERC20

PR #3977, opened 2024-06-17
fix: Do not consider xERC20 a collateral standard 2

AlexBHarley:
Amending https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3973 to ensure output of the CLI doesn't incorrectly add `collateralAddressOrDenom` to xERC20 configs.

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -183,10 +183,7 @@ async function getWarpCoreConfig(
         return wrappedToken;
       }
 
-      if (
-        config.type === TokenType.collateral ||
-        config.type === TokenType.XERC20
-      ) {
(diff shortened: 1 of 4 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -183,10 +183,7 @@ async function getWarpCoreConfig(
         return wrappedToken;
       }
 
-      if (
-        config.type === TokenType.collateral ||
-        config.type === TokenType.XERC20
-      ) {
+      if (config.type === TokenType.collateral) {
```
nambrot, line 186: @AlexBHarley Can you expand on this change here?

nambrot (approved):

PR #3983, opened 2024-06-17
feat: Script to generate Renzo Warp Route Configuration

nambrot:

### DIFF
--- a/typescript/infra/scripts/generate-renzo-warp-route-config.ts
+++ b/typescript/infra/scripts/generate-renzo-warp-route-config.ts
@@ -1,0 +1,150 @@
+import { writeFileSync } from 'fs';
+import { stringify as yamlStringify } from 'yaml';
+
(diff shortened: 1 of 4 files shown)


typescript/infra/scripts/generate-renzo-warp-route-config.ts
```diff
@@ -0,0 +1,150 @@
...
+  },
+  ethereum: {
+    threshold: 1,
+    validators: [
+      '0xc7f7b94a6BaF2FFFa54DfE1dDE6E5Fcbb749e04f', // Renzo
+      '0x1fd889337F60986aa57166bc5AC121eFD13e4fdd', // Everclear
+    ],
+  },
+};
+const zeroAddress = '0x0000000000000000000000000000000000000000';
```
yorhodes, line 87: this needed to be 01 right?

yorhodes (approved):

PR #3985, opened 2024-06-18
feat(cli): Support agent config generation for all chains

Mo-Hussain:
### Description

- Support generating agent config for all supported chains
- Add user prompts: prompt user whether they want to zero the IGP addresses if not provided, prompt user whether they want to generate agent config anyway when schema validation fails
- Support skipping prompts, useful for CI

Example usage:

Specific chains

```
$yarn hyperlane registry agent-config --chains anvil8545
Hyperlane CLI

Creating agent config...
interchainGasPaymaster address is missing for anvil8545
? Would you like to set the interchainGasPaymaster address to 0x0 for anvil8545? yes

Agent config is invalid, this is possibly due to required contracts not being deployed. See details below:
Validation error: Required at "chains.anvil8545.merkleTreeHook"
? Would you like to continue anyway? yes

Writing agent config to file ./configs/agent-config.json
✅ Agent config successfully written to ./configs/agent-config.json
```

All chains
```
$yarn hyperlane registry agent-config
Hyperlane CLI

Creating agent config...
?
No chains provided, would you like to generate the agent config for all supported chains? yes
interchainGasPaymaster address is missing for sketchpad
? Would you like to set the interchainGasPaymaster address to 0x0 for sketchpad? yes
interchainGasPaymaster address is missing for anvil8545
? Would you like to set the interchainGasPaymaster address to 0x0 for anvil8545? yes

Agent config is invalid, this is possibly due to required contracts not being deployed. See details below:
Validation error: Required at "chains.anvil8545.merkleTreeHook"; Invalid at "chains.sketchpad.interchainSecurityModule"
? Would you like to continue anyway? yes

Writing agent config to file ./configs/agent-config.json
✅ Agent config successfully written to ./configs/agent-config.json
```

Skip prompts:

```
$yarn hyperlane registry agent-config --skipPrompts
Hyperlane CLI

Creating agent config...

No chains provided, generating agent config for all supported chains

Agent config is invalid, this is possibly due to required contracts not being deployed. See details below:
Validation error: Required at "chains.anvil8545.merkleTreeHook"; Required at "chains.anvil8545.interchainGasPaymaster"; Invalid at "chains.sketchpad.interchainGasPaymaster"; Invalid at "chains.sketchpad.interchainSecurityModule"
Creating agent config anyway...

Writing agent config to file ./configs/agent-config.json
✅ Agent config successfully written to ./configs/agent-config.json
```

### Testing

Manual

### DIFF
--- a/typescript/cli/src/commands/registry.ts
+++ b/typescript/cli/src/commands/registry.ts
@@ -113,17 +113,11 @@ const createAgentConfigCommand: CommandModuleWithContext<{
       false,
       'The path to output an agent config JSON file.',
     ),
-    skipPrompts: {
-      type: 'boolean',
-      description: 'Skip user prompts',
-      default: false,
-    },
   },
   handler: async ({
     context,
     chains,
     out,
-    skipPrompts,
   }: {
     context: CommandContext;
     chains?: string;
@@ -152,7 +146,6 @@ const createAgentConfigCommand: CommandModuleWithContext<{
       context,
       chains: chainNames,
       out,
-      skipPrompts,
     });
     process.exit(0);
   },
--- a/typescript/cli/src/commands/registry.ts
+++ b/typescript/cli/src/commands/registry.ts
@@ -3,7 +3,7 @@ import { CommandModule } from 'yargs';
 import { createAgentConfig } from '../config/agent.js';
 import { createChainConfig } from '../config/chain.js';
 import { CommandContext, CommandModuleWithContext } from '../context/types.js';
-import { log, logBlue, logGray, logRed, logTable } from '../logger.js';
+import { errorRed, log, logBlue, logGray, logTable } from '../logger.js';
 
 import {
   chainTargetsCommandOption,
@@ -100,12 +100,12 @@ const addressesCommand: CommandModuleWithContext<{ name: string }> = {
  * agent-config command
  */
 const createAgentConfigCommand: CommandModuleWithContext<{
-  chains: string;
+  chains?: string;
   out: string;
+  skipPrompts: boolean;
 }> = {
   command: 'agent-config',
   describe: 'Create a new agent config',
-
   builder: {
     chains: chainTargetsCommandOption,
     out: outputFileCommandOption(
@@ -113,32 +113,47 @@ const createAgentConfigCommand: CommandModuleWithContext<{
       false,
       'The path to output an agent config JSON file.',
     ),
+    skipPrompts: {
+      type: 'boolean',
+      description: 'Skip user prompts',
+      default: false,
+    },
   },
   handler: async ({
     context,
     chains,
     out,
+    skipPrompts,
   }: {
     context: CommandContext;
-    chains: string;
+    chains?: string;
     out: string;
+    skipPrompts: boolean;
   }) => {
     const { multiProvider } = context;
 
-    const chainNames = chains.split(',');
-    const invalidChainNames = chainNames.filter(
-      (chainName) => !multiProvider.hasChain(chainName),
-    );
-    if (invalidChainNames.length > 0) {
-      logRed(
-        `Invalid chain names: ${invalidChainNames
-          .join(', ')
-          .replace(/, $/, '')}`,
+    let chainNames: string[] | undefined;
+    if (chains) {
+      chainNames = chains.split(',');
+      const invalidChainNames = chainNames.filter(
+        (chainName) => !multiProvider.hasChain(chainName),
       );
-      process.exit(1);
+      if (invalidChainNames.length > 0) {
+        errorRed(
+          `❌ Invalid chain names: ${invalidChainNames
+            .join(', ')
+            .replace(/, $/, '')}`,
+        );
+        process.exit(1);
+      }
     }
 
-    await createAgentConfig({ context, chains: chainNames, out });
+    await createAgentConfig({
+      context,
+      chains: chainNames,
+      out,
+      skipPrompts,
+    });
     process.exit(0);
   },
 };
--- a/typescript/cli/src/config/agent.ts
+++ b/typescript/cli/src/config/agent.ts
@@ -1,8 +1,10 @@
 import { confirm } from '@inquirer/prompts';
+import { ethers } from 'ethers';
 import { fromError } from 'zod-validation-error';
 
 import { ChainAddresses } from '@hyperlane-xyz/registry';
 import {
+  AgentConfig,
   AgentConfigSchema,
   ChainMap,
   HyperlaneCore,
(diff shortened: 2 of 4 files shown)


tkporter (approved):
Lgtm! but should def move to the existing skipConfirmation

typescript/cli/src/commands/registry.ts
```diff
@@ -98,45 +98,60 @@ const addressesCommand: CommandModuleWithContext<{ name: string }> = {
...
   describe: 'Create a new agent config',
-
   builder: {
     chains: chainTargetsCommandOption,
     out: outputFileCommandOption(
       './configs/agent-config.json',
       false,
       'The path to output an agent config JSON file.',
     ),
+    skipPrompts: {
```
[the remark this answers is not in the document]
jmrossy: Trevor is right

typescript/cli/src/config/agent.ts
```diff
@@ -1,36 +1,64 @@
...
+  skipPrompts?: boolean;
 }) {
   logBlue('\nCreating agent config...');
 
   const { registry, multiProvider, chainMetadata } = context;
   const addresses = await registry.getAddresses();
 
-  const core = HyperlaneCore.fromAddressesMap(addresses, multiProvider);
+  if (!chains) {
+    if (skipPrompts) {
```
jmrossy, line 35: IIRC @yorhodes and @nbayindirli had some kind of new utility fn for this pattern of 'prompt if not provided'

Mo-Hussain: `export async function detectAndConfirmOrPrompt(
  detect: () => Promise<string | undefined>,
  prompt: string,
  label: string,
  source?: string,
): Promise<string> {
  let detectedValue: string | undefined;
  try {
    detectedValue = await detect();
    if (detectedValue) {
      const confirmed = await confirm({
        message: `Detected ${label} as ${detectedValue}${
          source ? ` from ${source}` : ''
        }, is this correct?`,
      });
      if (confirmed) {
        return detectedValue;
      }
    }
    // eslint-disable-next-line no-empty
  } catch (e) {}
  return input({ message: `${prompt} ${label}:`, default: detectedValue });
}`

I think you are referring to this method. I'm not sure if this use case is the right fit. 

Here we infer from the absence of chains, that all chains should be used and we have the user confirm (y/n) instead of prompting them for an input.

I'm open to having the users pass in a explicit option for all chains if that feels like a better pattern. Thoughts @nbayindirli @yorhodes?

tkporter: fwiw whatever we do go with, it's important that it doesn't require interaction as we'll be running this in CI

typescript/cli/src/config/agent.ts
```diff
@@ -19,19 +23,80 @@ export async function createAgentConfig({
...
+}
+
+function filterChainAddresses(
+  addresses: ChainMap<ChainAddresses>,
+  chains?: string[],
+) {
+  if (!chains) {
+    return addresses;
+  }
+  return objFilter(addresses, (chain, _): _ is ChainAddresses => {
```
tkporter, line 88: we have a nice helper `pick` that does exactly this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/utils/src/objects.ts#L89

typescript/cli/src/config/agent.ts
```diff
@@ -19,19 +23,80 @@ export async function createAgentConfig({
...
+
+  const core = HyperlaneCore.fromAddressesMap(chainAddresses, multiProvider);
+  const startBlocks = await getStartBlocks(chainAddresses, core, chainMetadata);
+
+  if (!skipConfirmation) {
+    await handleMissingInterchainGasPaymaster(chainAddresses);
+  }
+
+  const agentConfig: AgentConfig = buildAgentConfig(
+    chains ?? Object.keys(chainAddresses),
```
tkporter, line 46: iiuc because of `filterChainAddresses`, we may be able to just pass in `Object.keys(chainAddresses)` here?

typescript/cli/src/config/agent.ts
```diff
@@ -42,34 +107,57 @@ export async function createAgentConfig({
...
-    addresses as ChainMap<HyperlaneDeploymentArtifacts>,
-    startBlocks,
-  );
+async function handleMissingInterchainGasPaymaster(
+  chainAddresses: ChainMap<ChainAddresses>,
+) {
+  for (const [chain, addressesRecord] of Object.entries(chainAddresses)) {
+    if (!addressesRecord.interchainGasPaymaster) {
+      warnYellow(`interchainGasPaymaster address is missing for ${chain}`);
+      const zeroIGPAddress = await confirm({
```
tkporter, line 125: I think we may need to skip the interaction if `--yes` is used - iirc hybrid chains won't have an IGP and we'll want to still be able to eventually generate agent configs in CI even if that's the case. We can just default to using the zero address in this case

wonder if a helper function that wraps `confirm` where you can pass in whether or not to auto-confirm would be nice - it could print the question but then another log right after saying that it's auto-confirming

typescript/cli/src/config/agent.ts
```diff
@@ -19,19 +23,80 @@ export async function createAgentConfig({
...
-  const startBlocks = await promiseObjAll(
-    objMap(addresses, async (chain, _) => {
-      // If the index.from is specified in the chain metadata, use that.
+  const chainAddresses = filterChainAddresses(addresses, chains);
+
+  const core = HyperlaneCore.fromAddressesMap(chainAddresses, multiProvider);
+  const startBlocks = await getStartBlocks(chainAddresses, core, chainMetadata);
+
+  if (!skipConfirmation) {
+    await handleMissingInterchainGasPaymaster(chainAddresses);
```
jmrossy, line 42: This is probably okay but my expectation would have been to pass `skip` into `handleMissingInterchainGasPaymaster` and have it fail if it's both not provided and skip is true. See https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/rossy/cli-warp-output-shape/typescript/cli/src/deploy/core.ts#L48

typescript/cli/src/config/agent.ts
```diff
@@ -19,19 +23,80 @@ export async function createAgentConfig({
...
+  }
+
+  const agentConfig: AgentConfig = buildAgentConfig(
```
jmrossy, line 45: Nit, this `: AgentConfig` type def is redundant with the fn's return type def right?

typescript/cli/src/config/agent.ts
jmrossy, line 125: Nit: I think the convention would be `zeroIgpAddress` though `isZeroIgpAdress` or `useZero...` would be more clear imo

jmrossy (approved):
Approved with minor comments

PR #3988, opened 2024-06-18
feat: reduce extraneous entries in generated registry/agent artifacts

paulbalaji:
- filters out remote domain metadata from configs written to disk
```
    "ancient8": {
      // this gets removed
      "arbitrum": {
        "aggregationHook": "0xblahblahblah",
        "interchainSecurityModule": "0xblahblah"
        ...etc
      },
      // keep the usual entries like this
      "aggregationHook": "0x1EF4ED658d542524d1D547ba2F94d3B038a55b8f",
      "batchContractAddress": "0x4C97D35c668EE5194a13c8DE8Afc18cce40C9F28",
      "blockExplorers": [
...etc
```
- no longer include secret overrides in the configs written to disk
	- i.e. use public registry RPCs and chain metadata (want our output to be like that of update-agent-config, but with updated contract addresses only)

### DIFF
--- a/typescript/infra/scripts/agent-utils.ts
+++ b/typescript/infra/scripts/agent-utils.ts
@@ -44,6 +44,7 @@ import { Role } from '../src/roles.js';
 import {
   assertContext,
   assertRole,
+  filterRemoteDomainMetadata,
   getInfraPath,
   inCIMode,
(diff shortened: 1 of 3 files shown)


tkporter (approved):
nice

PR #3990, opened 2024-06-18
feat: reintroduce `ism read` and `hook read` commands

paulbalaji:
the original functionality of these commands got lost across a couple of refactors

1. `ism read`/`hook read` was merged to become `core read`, which still had the purpose of "gimme config at this address"
2. then with a recent change, `core read` became "gimme the ism/hook configs relating to this mailbox address"

This PR adds the original functionality back of returning the ISM/Hook configs for *any* ISM/Hook address (on EVM)

drive-by:
- update mainnet config
- fix typescript linting errors/warnings

### DIFF
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -120,7 +120,7 @@ export const read: CommandModuleWithContext<{
   config: string;
 }> = {
   command: 'read',
-  describe: 'Reads onchain ISM & Hook configurations for given addresses',
+  describe: 'Reads onchain Core configuration for a given mailbox address',
   builder: {
     chain: {
       ...chainCommandOption,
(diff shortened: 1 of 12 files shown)


paulbalaji:
I debated adding this back as `config read`, but the existing config commands seem to be more about manipualting local configs and don't touch onchain stuff. However, the closest in functionality to these two is `core read`, and if we follow that pattern of `<module> read` I think these commands make sense - but open to discuss

typescript/cli/src/commands/core.ts
```diff
@@ -120,7 +120,7 @@ export const read: CommandModuleWithContext<{
   config: string;
 }> = {
   command: 'read',
-  describe: 'Reads onchain ISM & Hook configurations for given addresses',
+  describe: 'Reads onchain Core configuration for a given mailbox address',
```
yorhodes, line 123: thank you!

yorhodes (approved):

PR #3943, opened 2024-06-11
fix(cli): use registry to getWarpArtifactsPaths for warp send

nbayindirli:
### Description

- use registry to getWarpArtifactsPaths for `hl warp send`
- TODO: Expose `getWarpArtifactsPaths` from registry

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- ci-test

PR #3992, opened 2024-06-18
feat(cli): add `warp --symbol` flag

yorhodes:
### Description

- This allows users to interact read warp routes from the registry with symbol identifier rather than chain/address

### Drive-by

Rename `wei` to `amount`

### Backward compatibility

No, outputs chain map

### Testing

- Manual

(single)
```sh
$ yarn hyperlane warp read --symbol EZETH
$ yarn hyperlane warp send --symbol EZETH
```

(multiple)
```sh
$ yarn hyperlane warp read --symbol USDC
$ yarn hyperlane warp send --symbol USDC
```

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -1,7 +1,8 @@
 import { stringify as yamlStringify } from 'yaml';
 import { CommandModule } from 'yargs';
 
-import { EvmERC20WarpRouteReader } from '@hyperlane-xyz/sdk';
+import { ChainMap, EvmERC20WarpRouteReader } from '@hyperlane-xyz/sdk';
+import { objMap, promiseObjAll } from '@hyperlane-xyz/utils';
 
 import { createWarpRouteDeployConfig } from '../config/warp.js';
 import {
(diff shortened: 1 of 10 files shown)


typescript/cli/src/commands/warp.ts
```diff
@@ -100,42 +106,86 @@ export const configure: CommandModuleWithContext<{
...
+        warpCoreConfig.tokens.map((t) => [t.chainName, t.addressOrDenom!]),
+      );
+    } else if (chain && address) {
+      addresses = {
+        [chain]: address,
+      };
+    } else {
+      logGreen(`Please specify either a symbol or chain and address`);
+      process.exit(0);
+    }
```
nbayindirli, line 172: this seems to overlap with
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b0828b3d0eb95921a8753bb0c08b36b37ca0af15/typescript/cli/src/send/transfer.ts#L119-L130

can we consolidate somehow?

nbayindirli (changes requested):
this overlaps significantly with https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3943/files (still draft)

only feedback is
1. we should probably support 2+ assets with same name (test for ETH on `yourchain` and `fuji` returned `No (or multiple) warp routes found for symbol ETH`),
2. should consider moving symbol logic into shared fn for use throughout cli

nbayindirli:
> this overlaps significantly with https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3943/files (still draft)
> 
> only feedback is
> 
> 1. we should probably support 2+ assets with same name (test for ETH on `yourchain` and `fuji` returned `No (or multiple) warp routes found for symbol ETH`),
> 2. should consider moving symbol logic into shared fn for use throughout cli

I see 1 is addressed now so 2 is only concern

typescript/cli/src/commands/warp.ts
nbayindirli: cc @jmrossy since we discussed this in-depth last week

ltyu (approved):

typescript/cli/src/commands/warp.ts
yorhodes: I will add the same symbol option to this command though

nbayindirli (approved):
🚢

nbayindirli:
may be an issue parsing `cli/test-configs/anvil/deployments/warp_routes/FAKE/anvil1-anvil2-config.yaml` when running `warp send` in ci-test now

PR #4003, opened 2024-06-19
feat: relayer queue operation labels

daniel-savu:
### Description

- adds status labels to operations in the relayer queues. In practice, the most useful ones are those for when a message is retried due to an error. Labels are defined as enums to keep the cardinality finite, the size of prometheus metrics could get very large
- whenever a message is pushed to a queue, the interface requires specifying the new status as a `Option<PendingOperationStatus>`. If `None`, the status is left as before (such as in the case when an operation isn't ready to be retried).
- a limitation is that labels aren't stored in the DB, so messages are loaded with the `FirstPrepareAttempt` label. I just triggered retries for destinations I wanted to get labels for, but we could persist to the db in a follow-up PR
- Example prep queue labels for inevm:
Screenshot 2024-06-21 at 18 05 58

### Drive-by changes

- removes `make_op_try` as it doesn't really add anything that couldn't be done with regular Rust
- bumps the `strum` version because there was a bug in the previous one, preventing inner enum fields from being converted to strings

### Backward compatibility

<!--
Yes

### Testing

Manual


rust/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -106,6 +114,74 @@ pub trait PendingOperation: Send + Sync + Debug + TryBatchAs<HyperlaneMessage> {
...
+    #[strum(to_string = "Delivery transaction reverted or reorged")]
+    /// Delivery transaction reverted or reorged
+    RevertedOrReorged,
+}
+
+#[derive(Display, Debug, Clone)]
+/// Reasons for repreparing an operation
+pub enum ConfirmReason {
+    #[strum(to_string = "Submitted by this relayer")]
+    /// Error checking message delivery status
```
tkporter, line 174: nit: this & the `AlreadySubmitted` comment look wrong

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -322,7 +353,14 @@ impl PendingOperation for PendingMessage {
                 message_id=?self.message.id(),
                 "Transaction attempting to process message either reverted or was reorged"
             );
-            self.on_reprepare()
+            let span = info_span!(
```
tkporter, line 356: curious why this span but not for any other cases? wonder if these would already be in a span / if this should be for the other cases too

but probably a good idea not to instrument return values bc i'm sure
```
        if !self.is_ready() {
            return PendingOperationResult::NotReady;
        }
```
gets called a bunch and would be noisy

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -414,10 +458,21 @@ impl PendingMessage {
...
-    fn on_reprepare(&mut self) -> PendingOperationResult {
+    fn on_reprepare<E: Debug>(
+        &mut self,
+        err: E,
+        reason: ReprepareReason,
+    ) -> PendingOperationResult {
         self.inc_attempts();
         self.submitted = false;
-        PendingOperationResult::Reprepare
+        warn!(error = ?err, "{}", reason.clone());
```
tkporter, line 468: I assume info about the PendingMessage will be in some span that's shown in this log?

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -414,10 +458,21 @@ impl PendingMessage {
...
+        PendingOperationResult::Reprepare(reason)
+    }
+
+    fn on_reconfirm<E: Debug>(&mut self, err: E, reason: &str) -> PendingOperationResult {
+        self.inc_attempts();
+        warn!(error = ?err, "{}", reason.clone());
```
tkporter, line 474: Wonder if adding a string in here to make this log line easy to search for may be useful, e.g. `"Reconfirming message: {}"` or something - same for the on_reprepare

rust/agents/relayer/src/msg/op_queue.rs
```diff
@@ -98,8 +105,12 @@ impl OpQueue {
     /// Get the metric associated with this operation
     fn get_operation_metric(&self, operation: &dyn PendingOperation) -> IntGauge {
         let (destination, app_context) = operation.get_operation_labels();
-        self.metrics
-            .with_label_values(&[&destination, &self.queue_metrics_label, &app_context])
+        self.metrics.with_label_values(&[
+            &destination,
+            &self.queue_metrics_label,
+            &operation.status().to_string(),
```
tkporter, line 111: very clean that this is pretty much the only change to plug things together w/ updated metric labels

rust/agents/relayer/src/msg/pending_message.rs
tkporter, line 468: if `err` is `""` does this mean that logs will have `error: ""`? Wonder if there's any clearer way to indicate no error actually happened - I often search "error" in logs and think it'd be nice to only show that when an error for sure happened

tkporter (approved):
Very nice, this is great

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: ah so it's because I wanted to remove the `warn` on the line above, but the `warn` had two fields in the logs which I can't easily pass to `on_reprepare`. So I figured I'd just include those two fields in the span, and the span requires there to also be a string description, and if a warning ends up being printed in `on_reprepare`, those two fields will also be visible. I pushed a change that better reflects this now

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: not right now, but I'll include this

rust/agents/relayer/src/msg/pending_message.rs
daniel-savu: and to clarify, indeed the span doesn't show up when values are returned, just when logs are emitted

tkporter: ah cool, sgtm

PR #4005, opened 2024-06-19
fix: reduce loop test igpconfig

tiendn:
### Description
Refactor `testIgpConfig` in `testUtils.ts` for Enhanced Readability and Maintainability by reduce loop

### Drive-by changes
1. **Constant Usage:** Introduced `OVERHEAD_COST` as a constant to replace the magic number `60000`, making the purpose of this value clearer and the code easier to update.
2. **Improved Readability:** By using descriptive names and constants, the function is now easier to understand at a glance, which is particularly beneficial for new contributors or during code reviews.
3. **Dynamic Configuration Logic:** Ensured that the configuration for `overhead` and `oracleConfig` is dynamically generated for each chain, excluding the current chain to simulate realistic test scenarios.

These changes do not alter the logic but make the codebase cleaner and more maintainable.


typescript/sdk/src/test/testUtils.ts
```diff
@@ -63,25 +63,31 @@ const TEST_ORACLE_CONFIG = {
...
+      });
+      return [
+        local,
+        {
+          owner,
+          oracleKey: owner,
+          beneficiary: owner,
+          overhead: Object.fromEntries(overhead),
+          oracleConfig: Object.fromEntries(oracleConfig),
+        },
```
paulbalaji, line 89: nit: would prefer not to update the maps directly

```ts
const TEST_OVERHEAD_COST = 60000;

export function testIgpConfig(
  chains: ChainName[],
  owner = nonZeroAddress,
): ChainMap<IgpConfig> {
  return Object.fromEntries(
    chains.map((local) => {
      const overhead: IgpConfig['overhead'] = {};
      const oracleConfig: IgpConfig['oracleConfig'] = {};
      exclude(local, chains).map((remote) => {
        overhead[remote] = TEST_OVERHEAD_COST;
        oracleConfig[remote] = TEST_ORACLE_CONFIG;
      });
      return [
        local,
        {
          owner,
          oracleKey: owner,
          beneficiary: owner,
          overhead,
          oracleConfig,
        },
      ];
    }),
  );
}
```

tiendn: oh, sr. Thank you for your advising

PR #4007, opened 2024-06-19
feat: Add smart provider log setter

### LINKED ISSUE #4006: (cli-2.0) SmartProvider logs pollute output
## Problem
In deriveTokenTyp() of EvmERC20WarpRouteReader, Provider errors flood the terminal because of the try-catch TokenType derivation. This is problematic because they look like real errors, but are expected. 

## Solution
- Consider adding logic to turn off the errors temporarily within this method, and then re-enabling it afterwards.
- Modify the SmartProvider to add configurability to when to show these errors
- Filter these errors out somehow

ltyu:
### Description
- Add logic to enable setting of log levels to the SmartProvider. 
- Add logic to set log level to `silent` in the beginning of `deriveTokenType()`, and then resetting it after.

### Related issues
- Fixes #4006 

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -6,11 +6,17 @@ import {
   HypERC20__factory,
 } from '@hyperlane-xyz/core';
 import {
+  HyperlaneSmartProvider,
   MailboxClientConfig,
   TokenRouterConfig,
   TokenType,
 } from '@hyperlane-xyz/sdk';
-import { Address, eqAddress, rootLogger } from '@hyperlane-xyz/utils';
+import {
+  Address,
+  eqAddress,
+  getLogLevel,
+  rootLogger,
+} from '@hyperlane-xyz/utils';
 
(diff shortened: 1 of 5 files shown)


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -203,4 +219,15 @@ export class EvmERC20WarpRouteReader {
...
     return { name, symbol, decimals, totalSupply: totalSupply.toString() };
   }
+
+  /**
+   * Conditionally sets the log level for a smart provider.
+   *
+   * @param level - The log level to set, e.g. 'debug', 'info', 'warn', 'error'.
+   */
+  protected setSmartProviderLogLevel(level: string) {
+    if (this.provider instanceof HyperlaneSmartProvider) {
```
jmrossy, line 229: Recommend against instanceof in production code.
Better would be to check for the existence of a `setLogLevel` fn on the provider, even if it requires a ts-ignore here.

jmrossy (approved):

PR #3678, opened 2024-04-29
feat: WarpModule with update() IsmConfig

ltyu:
### Description
Implement WarpModule update() with IsmConfig

### Related issues
- Fixes https://github.com/hyperlane-xyz/issues/issues/1156

### Backward compatibility
Yes

### Testing
Unit Tests

PR #4009, opened 2024-06-19
fix(cli): ezETH deploy fixes

nambrot:
### Description

some minor fixes to make the ezEth deploy work. Does not include the modifications to contract verification retries that I made.


typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -466,14 +466,14 @@ export class EvmIsmModule extends HyperlaneModule<
...
     // deploy the submodules first
-    const submoduleAddresses: Address[] = await Promise.all(
-      Object.keys(config.domains).map(async (origin) => {
-        const { address } = await this.deploy({
-          config: config.domains[origin],
-        });
-        return address;
-      }),
-    );
+    const submoduleAddresses: Address[] = [];
```
paulbalaji, line 469: curious what wasn't working and why changing promise.all to for loops fixes things

nambrot: Promise.all is not a good idea, because concurrency support in signers is very poor, so you'll end up signing multiple transactions with the same nonce

paulbalaji:
may have to pull in changes from `cli-2.0`, as the cli-e2e ci job has been renamed

typescript/cli/src/utils/balances.ts
```diff
@@ -40,18 +40,16 @@ export async function gasBalancesAreSufficient(
...
-      const gasPrice = await provider.getGasPrice();
-      const minBalanceWei = gasPrice.mul(minGas).toString();
-      sufficient = await nativeBalancesAreSufficient(
-        multiProvider,
-        signer,
-        [chain],
-        minBalanceWei,
-      );
-    }),
-  );
```
yorhodes, line 55: I dont see why this concurrency was a problem

typescript/sdk/src/ism/EvmIsmModule.ts
yorhodes: (not a good idea for a single chain and single signer)

typescript/cli/src/utils/balances.ts
```diff
@@ -40,18 +40,16 @@ export async function gasBalancesAreSufficient(
...
+  for (const chain of chains) {
+    const provider = multiProvider.getProvider(chain);
+    const gasPrice = await provider.getGasPrice();
+    const minBalanceWei = gasPrice.mul(minGas).toString();
+    sufficient = await nativeBalancesAreSufficient(
+      multiProvider,
+      signer,
+      [chain],
+      minBalanceWei,
+    );
```
yorhodes, line 52: also now realizing this only effectively checks the last chain...

typescript/sdk/src/token/deploy.ts
```diff
@@ -64,11 +66,16 @@ abstract class TokenDeployer<
     }
   }
 
-  async initializeArgs(_: ChainName, config: TokenRouterConfig): Promise<any> {
+  async initializeArgs(
+    chain: ChainName,
+    config: TokenRouterConfig,
+  ): Promise<any> {
+    const signer = await this.multiProvider.getSigner(chain).getAddress();
```
yorhodes, line 73: does this raise or lower determinism :P

yorhodes (commented):
we should fix the balance sufficient check

typescript/cli/src/utils/balances.ts
nambrot: It was because it would prompt for multiple confirmations at the same time

typescript/sdk/src/token/deploy.ts
nambrot: My read is that this is part of the API of `RouterDeployer`

typescript/cli/src/utils/balances.ts
nbayindirli: if any of the many chains have insufficient balance, the `sufficient` flag becomes `false`– why does this only check last chain?

typescript/sdk/src/token/deploy.ts
ltyu: I will need to add transferOwnership() in my WarpModule #3678 because i use `deployContracts()` directly

ltyu (approved):

typescript/cli/src/utils/balances.ts
```diff
@@ -40,18 +40,16 @@ export async function gasBalancesAreSufficient(
...
-        [chain],
-        minBalanceWei,
-      );
-    }),
-  );
+  for (const chain of chains) {
+    const provider = multiProvider.getProvider(chain);
+    const gasPrice = await provider.getGasPrice();
+    const minBalanceWei = gasPrice.mul(minGas).toString();
+    sufficient = await nativeBalancesAreSufficient(
```
ltyu, line 47: nativeBalancesAreSufficient() allows a user to continue if any of the chains have insufficient balance. why do we want to let the user continue? It will fail anyways because of insufficient balance

nambrot: because our estimates might be wrong, or a user has topped it up in the meantime.

typescript/cli/src/utils/balances.ts
yorhodes: if `chains.length == 2 && isSufficient(chains[0]) == false && isSufficient(chains[1]) == true` then sufficient is true after this loop ends?

yorhodes: this is also true for the `nativeBalancesAreSufficient` helper, which also should not prompt and assign to a single variable across concurrent promises

ltyu:
Screenshot 2024-06-26 at 6 07 17 PM

typescript/cli/src/deploy/utils.ts
```diff
@@ -55,13 +56,20 @@ export async function runPreflightChecksForChains({
...
+      multiProvider,
+      signer,
+      chainsToGasCheck ?? chains,
+      minGas,
+    )
+  ).every((sufficient) => sufficient);
+
+  if (allSufficient) logGreen('✅ Balances are sufficient');
+  else
+    logRed(
```
nambrot, line 70: Like mentioned offline, wouldn't do this. If you get the warning and then top up later, we will still show this warning. What actually makes more sense to me, is to fetch all the imbalances all at once, list those, and then ask the user once if they want to continue

PR #4014, opened 2024-06-20
feat(infra): allow specifying chains to deploy validators for in infra

paulbalaji:
- allow someone to provide a specific set of chains to deploy validators for

deploying validator:
image

deploying relayer:
image

### DIFF
--- a/typescript/infra/scripts/agents/utils.ts
+++ b/typescript/infra/scripts/agents/utils.ts
@@ -31,18 +31,19 @@ export class AgentCli {
     // make all the managers first to ensure config validity
     for (const role of this.roles) {
       switch (role) {
-        case Role.Validator:
-          for (const chain of this.agentConfig.contextChainNames[role]) {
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/agents/utils.ts
```diff
@@ -32,6 +33,12 @@ export class AgentCli {
       switch (role) {
         case Role.Validator:
           for (const chain of this.agentConfig.contextChainNames[role]) {
+            // check if current chain is in provided set of chains
+            if (!this.chains?.includes(chain)) {
```
tkporter, line 37: is the intended behavior to default to not skipping any chains if `this.chains` is undefined? iiuc as written it will always require this.chains to be truthy, which feels cumbersome

paulbalaji: no that's an oversight from me, will fix

tkporter (approved):
🫡

PR #4020, opened 2024-06-20
feat: add xerc20 limit lookups to warp read

yorhodes:
### Description

Adds xerc20 mint and burn limit fetching to `warp read --symbol`

### Backward compatibility

Yes

### Testing

Manual
Screenshot 2024-06-20 at 3 19 13 PM


typescript/cli/src/commands/warp.ts
```diff
@@ -144,6 +151,45 @@ export const read: CommandModuleWithContext<{
...
+                : await HypXERC20__factory.connect(
+                    router,
+                    provider,
+                  ).wrappedToken();
+
+            const xerc20 = IXERC20__factory.connect(xerc20Address, provider);
+            const mint = await xerc20.mintingCurrentLimitOf(router);
+            const burn = await xerc20.burningCurrentLimitOf(router);
+
+            const formattedLimits = objMap({ mint, burn }, (_, v) =>
```
jmrossy, line 181: objMap seems like an odd choice here

typescript/cli/src/commands/warp.ts
```diff
@@ -144,6 +151,45 @@ export const read: CommandModuleWithContext<{
         context.registry,
         symbol,
       );
+
+      // TODO: merge with XERC20TokenAdapter and WarpRouteReader
+      const xerc20Limits = await Promise.all(
```
jmrossy, line 156: Not sure if this is still the case but my intention was to keep these command definition files simple and dumb, in case we ever wanted to swap out yargs for something else. So I'd recommend moving this business logic to somewhere else

jmrossy (approved):
LGTM just some style comments

typescript/cli/src/commands/warp.ts
yorhodes: why?

PR #4022, opened 2024-06-20
chore(ci): fix ci workflow filters on main

paulbalaji:
chore(ci): fix ci workflow filters on main

generally copying the path/filters for `pull_request` into `push` so that we correctly run the right workflows

### DIFF
--- a/.github/workflows/monorepo-docker.yml
+++ b/.github/workflows/monorepo-docker.yml
@@ -4,11 +4,6 @@ on:
     branches: [main]
     tags:
       - '**'
-    paths:
-      # For now, because this image is only used to use `infra`, we just build for infra changes
-      - 'typescript/infra/**'
-      - 'Dockerfile'
-      - '.dockerignore'
   pull_request:
     paths:
       # For now, because this image is only used to use `infra`, we just build for infra changes
--- a/.github/workflows/monorepo-docker.yml
+++ b/.github/workflows/monorepo-docker.yml
@@ -4,6 +4,11 @@ on:
     branches: [main]
     tags:
       - '**'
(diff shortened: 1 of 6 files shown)


.github/workflows/monorepo-docker.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches: [main]
     tags:
       - '**'
+    paths:
+      # For now, because this image is only used to use `infra`, we just build for infra changes
+      - 'typescript/infra/**'
```
tkporter, line 9: does this mean that we wouldn't build an image for a commit to main unless it touches this? or for agents, anything in the rust dir?

if so, curious what motivated this, I've found it to be kinda nice to have these built for every commit to main. Some reasons:
- it's easy to find an image, you don't need to search for a specific commit, you just grab the latest
- at least atm while we don't have registry pinning it's kinda nice to have these be built so we have an up to date registry in the monorepo images

The only downside I see is cost and afaik I haven't seen tension on the costs here

paulbalaji: i think for me at a high level I feel a new image is only required if there were changes - so you don't run into an issue trying to see if two images had any actual underlying diff, but I get your point that it's easy to grab an image tag from any new build this way

it's not really an urgent thing, more of a drive-by while fixing the rust/rust-skipped/test/test-skipped workflows - happy to revert the docker build changes

tkporter (approved):

.github/workflows/monorepo-docker.yml
tkporter: gotcha, makes sense, thanks

PR #4031, opened 2024-06-21
feat: pin monorepo to a specific registry version

paulbalaji:
feat: pin monorepo to a specific registry version
- introducing a `.registryrc`, which can be a commit hash, branch or tag
- introduce a new action just for checking out a registry
	- remove registry caching because it's quicker now to just checkout the specified version


jmrossy (commented):
Seems cool but just one question: Will the potentially conflicting sources of truth for registry versions cause confusion? There's this new .registryrc but also the version(s) used in package.json files

jmrossy (approved):

PR #4036, opened 2024-06-22
chore: some throughput tweaks

tkporter:
### Description

Context: https://discord.com/channels/935678348330434570/1254036693217054794


rust/Cargo.lock
```diff
@@ -7037,6 +7037,7 @@ dependencies = [
  "num-traits",
  "once_cell",
  "prometheus",
+ "rand 0.5.6",
```
paulbalaji, line 7040: why is the version here different to the one in the Cargo.toml?

typescript/infra/src/agents/index.ts
```diff
@@ -139,7 +139,7 @@ export abstract class AgentHelmManager {
             rpcConsensusType: this.rpcConsensusType(chain),
             protocol: metadata.protocol,
             blocks: { reorgPeriod },
-            maxBatchSize: 4,
+            maxBatchSize: 16,
```
daniel-savu, line 142: are we bumping this to 32 everywhere following our conversation earlier?

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -285,9 +285,9 @@ const releaseCandidate: RootAgentConfig = {
     rpcConsensusType: RpcConsensusType.Fallback,
     docker: {
       repo,
-      tag: '0d12ff3-20240620-173353',
+      tag: '9535087-20240623-174819',
     },
-    whitelist: releaseCandidateHelloworldMatchingList,
+    // whitelist: releaseCandidateHelloworldMatchingList,
```
daniel-savu, line 290: should we add back this comment?
```
    // We're temporarily (ab)using the RC relayer as a way to increase
    // message throughput.
```

daniel-savu (approved):
lgtm!

typescript/infra/src/agents/index.ts
tkporter: will do

rust/Cargo.lock
tkporter: Screen Shot 2024-06-24 at 2 24 42 PM

Whoa i made a typo! and we have this same typo in a bunch of other places that I've just fixed. Wild!

It shows up as a compilation warning but we have a lot of noise from some Solana warnings that it was ignored

tkporter: nice catch

PR #4040, opened 2024-06-24
feat(infra): deploy to a set of specific chains

paulbalaji:
- feat(infra): deploy to a set of specific chains
- allows users to specify more than 1 target chain to deploy to
- provides suggestions of available chains to deploy (given local registry)

image


tkporter (approved):
Nice

PR #4041, opened 2024-06-24
feat: pin monorepo to a specific registry version (cherry-pick #4031)

paulbalaji:
- feat: pin monorepo to a specific registry version (cherry-pick #4031)
- drive-by agent config update


jmrossy (approved):
LGTM, let's merge it along with https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4035

PR #4043, opened 2024-06-24
feat: persist status labels

daniel-savu:
### Description

Builds on top of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4003 to persist updates to an operation's status to the db, so we no longer default to `FirstPrepareAttempt` upon restarting the relayer

### Drive-by changes

Moves `hyperlane_db` from `hyperlane_base` to `hyperlane_core` so it can be added to the `PendingOperation` trait. `PendingMessage` already stores a reference to the db in `ctx`, so I figured it doesn't break any assumptions.

### Backward compatibility
Yes

### Testing

Unit tests encoding of `PendingOperationStatus`, need to manually tests e2e


rust/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -186,9 +186,21 @@ async fn receive_task(
...
-            .await;
+        let status = match op.origin_db().retrieve_status_by_message_id(&op.id()) {
+            Ok(status) => status,
+            Err(e) => {
+                warn!(error=?e, "Failed to retrieve status for message");
+                None
+            }
+        };
+        let status = status.unwrap_or_else(|| {
+            debug!(
```
tkporter, line 197: nit: iiuc we'd print this for each new message, wondering if we should make this trace? just seeing as this isn't super important to know when debugging

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -127,6 +128,13 @@ impl PendingOperation for PendingMessage {
     }
 
     fn set_status(&mut self, status: PendingOperationStatus) {
+        if let Err(e) = self
+            .ctx
+            .origin_db
+            .store_status_by_message_id(&self.message.id(), &self.status)
+        {
+            warn!(message_id = ?self.message.id(), err = %e, "Persisting `status` failed for message");
```
tkporter, line 136: would maybe nice to put the status in there

rust/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -129,7 +134,35 @@ pub enum PendingOperationStatus {
...
+}
+
+impl Decode for PendingOperationStatus {
+    fn read_from<R>(reader: &mut R) -> Result<Self, HyperlaneProtocolError>
+    where
+        R: std::io::Read,
+        Self: Sized,
+    {
+        // Deserialize from JSON and read from the reader, to avoid having to implement the encoding / decoding manually
+        serde_json::from_reader(reader).map_err(|_| {
```
tkporter, line 156: is there any way to get err in what's returned / maybe surface what the problematic bytes were?

rust/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -114,7 +119,7 @@ pub trait PendingOperation: Send + Sync + Debug + TryBatchAs<HyperlaneMessage> {
     fn set_retries(&mut self, retries: u32);
 }
 
-#[derive(Debug, Display, Clone)]
+#[derive(Debug, Display, Clone, Serialize, Deserialize, PartialEq)]
 /// Status of a pending operation
 pub enum PendingOperationStatus {
```
tkporter, line 124: Can we note here and ReprepareReason / ConfirmReason that we should only add new variants otherwise we break reading from the DB?

tkporter (approved):
lgtm, minor suggestions to just to discourage db incompat in the future

rust/hyperlane-core/src/traits/pending_operation.rs
daniel-savu: 💯

rust/agents/relayer/src/msg/pending_message.rs
```diff
@@ -142,6 +150,16 @@ impl PendingOperation for PendingMessage {
         self.ctx.destination_mailbox.domain()
     }
 
+    fn retrieve_status_from_db(&self) -> Option<PendingOperationStatus> {
+        match self.ctx.origin_db.retrieve_status_by_message_id(&self.id()) {
+            Ok(status) => status,
+            Err(e) => {
+                warn!(error=?e, "Failed to retrieve status for message");
```
daniel-savu, line 157: @tkporter FYI this will initially cause lots of warning logs, until the relayer retries every message once

tkporter: is it bc nothing will have been persisted yet?

I think `retrieve_status_by_message_id` gives Result<Option<_>> and it gives Ok(None) if there was no error reading the db but it just doesn't have an entry

PR #4046, opened 2024-06-24
fix: only evaluate dynamic revert reasons in reverting branch

yorhodes:
### Description

Optimize non-reverting branch gas usage caused by building dynamic revert reason strings

### Testing

Unit Tests


aroralanuk (approved):
Nice, what's the estimate of gas savings here?

PR #4048, opened 2024-06-25
chore: fix type typing

tiendn:
### Description
Fix type typo in `deploy.ts`

**Summary**
This pull request addresses a minor typo in the deploy.ts script. The type assertion for configMap was set `unknown`. It should have a name, or define exactly type

**Changes**
Corrected the type assertion for configMap from `ChainMap<unknown>` to `ChainMap<Config>`.

**Affected File**
`typescript/infra/scripts/deploy.ts`

**Detailed Explanation**

```diff
+ type Config = object;
- configMap: config as ChainMap<unknown>,
+ configMap: config as ChainMap<Config>,
```
This change ensures that the configMap is correctly typed, improving type safety and code readability.

**Testing**

**Additional Notes**

### DIFF
--- a/typescript/.changeset/clever-gifts-confess.md
+++ b/typescript/.changeset/clever-gifts-confess.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/infra': major
+---
+
+fix type typo ChainMap<unknown>
--- a/typescript/.changeset/clever-gifts-confess.md
+++ b/typescript/.changeset/clever-gifts-confess.md
@@ -1,5 +1,5 @@
 ---
-'@hyperlane-xyz/infra': major
+'@hyperlane-xyz/infra': patch
 ---
 
 fix type typo ChainMap<unknown>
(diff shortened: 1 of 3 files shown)


typescript/.changeset/clever-gifts-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/infra': major
```
paulbalaji, line 2: i would class this as a patch, not a major

typescript/infra/scripts/deploy.ts
```diff
@@ -51,6 +51,8 @@ import {
 } from './agent-utils.js';
 import { getEnvironmentConfig, getHyperlaneCore } from './core-utils.js';
 
+type Config = object;
```
jmrossy, line 54: This is essentially as generic as the `unknown` type before. I do appreciate the attempt but the type would need to be defined entirely and accurately for us to accept the submission

jmrossy (changes requested):

typescript/infra/scripts/deploy.ts
tiendn: Oh, okay

PR #4052, opened 2024-06-25
fix(ci): use the correct commit sha for agent/monorepo image tags

paulbalaji:
- use the correct commit hash for the agent and monorepo image tag
- the actual checked out commit was correct, but this now ensures that the TAG_SHA matches the checked out commit hash
- drive-by ensure that workflow gets triggered on changes to itself


daniel-savu (approved):
🤌

Issue #3915, opened 2024-06-06
Agent debug logs should always be present when configured


tkporter:
## Problem

Despite running the agents with `HYP_LOG_LEVEL=debug`, some debug logs refuse to show up, while others are present. 

Some examples:
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-base/src/contract_sync/mod.rs#L85 (https://cloudlogging.app.goo.gl/nYDSrsRtVguag4s1A)
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs#L131 (https://cloudlogging.app.goo.gl/nkWbSrmUtrq94sGG9) 
- some logs from hyperlane-db
- https://discord.com/channels/935678348330434570/1255176165736579183/1255182819966193778 in the scraper
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-base/src/contract_sync/mod.rs#L155 this log

## Solution

- We probably have some weird config getting in the way, or maybe some `#[instrument]` annotations are missing or misconfigured. Poke through reasons why they may not be showing up, and make sure they are
- We should be able to rely on debug logs in our code always showing up when we add them

PR #3981, opened 2024-06-17
feat(cli): merge in cli-2.0 (v4.0.0)

cmcewen:
CLI v4.0.0

PR #4054, opened 2024-06-25
fix: agent log verbosity

daniel-savu:
### Description

tbh I don't know why, but this makes all the logs we were previously missing show up in e2e:
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b99920eccae22de9fc9d735af87cf0e2b306655a/rust/hyperlane-base/src/contract_sync/mod.rs#L155
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f68319f46e6c140088e53c4a018bb23f12a8d8bd/rust/hyperlane-base/src/db/rocks/hyperlane_db.rs#L93

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3915

### Testing

E2E, by counting occurrences of the two logs linked above in relayer log output. Before this change, those logs would be missing completely


rust/hyperlane-base/src/settings/trace/mod.rs
```diff
@@ -63,7 +63,7 @@ impl TracingConfig {
     pub fn start_tracing(&self, metrics: &CoreMetrics) -> Result<console_subscriber::Server> {
         let mut target_layer = Targets::new().with_default(self.level);
 
-        if self.level < Level::DependencyTrace {
+        if self.level == Level::Trace {
```
tkporter, line 66: not sure if it's bc of this change or not but seems like we now (or still ?) get some noisy logs

from e2e:
Screen Shot 2024-06-26 at 2 07 15 PM

Wonder if we should set hyper to Level::Warn in the `if self.level < Level::Trace` block on line 82 or something - or maybe all of these?

rust/hyperlane-base/src/settings/trace/mod.rs
```diff
@@ -63,7 +63,7 @@ impl TracingConfig {
...
             // Reduce log noise from trusted libraries that we can reasonably assume are working correctly
             target_layer = target_layer
                 .with_target("hyper", Level::Info)
```
tkporter, line 69: lmaooo this is totally the one that's at fault

From https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#filtering-with-targets:
> A Targets filter consists of one or more [target](https://docs.rs/tracing-core/0.1.32/x86_64-unknown-linux-gnu/tracing_core/metadata/struct.Metadata.html#method.target) prefixes, paired with [LevelFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.LevelFilter.html)s. If a span or event’s [target](https://docs.rs/tracing-core/0.1.32/x86_64-unknown-linux-gnu/tracing_core/metadata/struct.Metadata.html#method.target) begins with one of those prefixes, and its [level](https://docs.rs/tracing-core/0.1.32/x86_64-unknown-linux-gnu/tracing_core/metadata/struct.Level.html) is at or below the [LevelFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.LevelFilter.html) enabled for that prefix, then the span or event will be enabled.

`hyper` prefix is conflicting with some of our targets that start with `hyperlane_base` and `hyperlane_core` 🤦‍♂️ 

I wonder if we can change this to `hyper::` and that'd fix it? And given that I wonder if we should just stick with `if self.level < Level::DependencyTrace`

tkporter (approved):
🚀

paulbalaji:
will require pulling `main` after https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3981 gets merged

PR #4055, opened 2024-06-25
fix(cli): RPC errors while sending test messages

ltyu:
### Description
This PR fixes RPC errors when trying to send a test message using the CLI for older chains that don't have tokenType (e.g., `hyperlane send message --relay --origin fuji --destination betaop`)

Additional details: https://discord.com/channels/935678348330434570/1254873902480359435/1255215832959811687

### Backward compatibility
Yes

### Testing
- tested with `hyperlane warp deploy` and then `hyperlane warp send --relay --warp $HOME/.hyperlane/deployments/warp_routes/ETH/alfajores-betaop-config.yaml`
- tested `hyperlane send message --relay --origin fuji --destination betaop`
- tested `hyperlane send message --relay --origin alfrajores --destination betaop`
- tested `hyperlane send message --relay --origin holesky --destination betaop`
- tested `hyperlane core read --mailbox 0xEf9F292fcEBC3848bF4bB92a96a04F9ECBb78E59 --chain alfajores `

### DIFF
--- a/typescript/sdk/src/hook/EvmHookReader.test.ts
+++ b/typescript/sdk/src/hook/EvmHookReader.test.ts
@@ -204,7 +204,7 @@ describe('EvmHookReader', () => {
 
     // top-level method infers hook type
     const hookConfig = await evmHookReader.deriveHookConfig(mockAddress);
-    expect(hookConfig).to.deep.equal({});
+    expect(hookConfig).to.be.undefined;
   });
 
   /*
--- a/typescript/sdk/src/hook/EvmHookReader.test.ts
+++ b/typescript/sdk/src/hook/EvmHookReader.test.ts
@@ -186,6 +186,27 @@ describe('EvmHookReader', () => {
     expect(config).to.deep.equal(hookConfig);
   });
 
+  it('should return an empty config if deriving fails', async () => {
+    const mockAddress = generateRandomAddress();
+    const mockOwner = generateRandomAddress();
+
+    // Mocking the connect method + returned what we need from contract object
+    const mockContract = {
+      // No type
+      owner: sandbox.stub().resolves(mockOwner),
+    };
+    sandbox
+      .stub(MerkleTreeHook__factory, 'connect')
+      .returns(mockContract as unknown as MerkleTreeHook);
+    sandbox
+      .stub(IPostDispatchHook__factory, 'connect')
+      .returns(mockContract as unknown as IPostDispatchHook);
+
+    // top-level method infers hook type
+    const hookConfig = await evmHookReader.deriveHookConfig(mockAddress);
+    expect(hookConfig).to.deep.equal({});
+  });
+
   /*
     Testing for more nested hook types can be done manually by reading from existing contracts onchain.
     Examples of nested hook types include:
--- a/typescript/sdk/src/hook/EvmHookReader.ts
+++ b/typescript/sdk/src/hook/EvmHookReader.ts
@@ -88,13 +88,12 @@ export class EvmHookReader implements HookReader {
   async deriveHookConfig(address: Address): Promise<DerivedHookConfig> {
     try {
       const hook = IPostDispatchHook__factory.connect(address, this.provider);
-
-      const onchainHookType = await hook.hookType();
-      this.logger.debug('Deriving HookConfig', { address, onchainHookType });
+      this.logger.debug('Deriving HookConfig', { address });
 
       // Temporarily turn off SmartProvider logging
       // Provider errors are expected because deriving will call methods that may not exist in the Bytecode
       this.setSmartProviderLogLevel('silent');
+      const onchainHookType = await hook.hookType();
 
       switch (onchainHookType) {
         case OnchainHookType.ROUTING:
--- a/typescript/sdk/src/hook/EvmHookReader.ts
+++ b/typescript/sdk/src/hook/EvmHookReader.ts
@@ -46,7 +46,9 @@ import {
 export type DerivedHookConfig = WithAddress<Exclude<HookConfig, Address>>;
 
 export interface HookReader {
-  deriveHookConfig(address: Address): Promise<WithAddress<HookConfig>>;
+  deriveHookConfig(
+    address: Address,
+  ): Promise<WithAddress<HookConfig> | undefined>;
   deriveMerkleTreeConfig(
     address: Address,
   ): Promise<WithAddress<MerkleTreeHookConfig>>;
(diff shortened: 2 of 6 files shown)


typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -345,4 +359,16 @@ export class EvmHookReader implements HookReader {
       type: HookType.PAUSABLE,
     };
   }
+
+  /**
+   * Conditionally sets the log level for a smart provider.
+   *
+   * @param level - The log level to set, e.g. 'debug', 'info', 'warn', 'error'.
+   */
+  protected setSmartProviderLogLevel(level: string) {
```
ltyu, line 368: this also exists in EvmWarpReader. Will DRY it out in a separate PR

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -85,33 +86,46 @@ export class EvmHookReader implements HookReader {
...
+        default:
+          throw new Error(
+            `Unsupported HookType: ${OnchainHookType[onchainHookType]}`,
+          );
+      }
+    } catch (e) {
+      this.logger.debug(
+        `Hook deriving failed for hook: ${address} with error ${e}`,
+      );
+      return {} as DerivedHookConfig;
```
ltyu, line 126: @paulbalaji this changes the existing behavior of derive. Previously, this throws, but now we return an empty HookConfig. can you think of any issues this could cause?

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -85,33 +86,46 @@ export class EvmHookReader implements HookReader {
...
-        return this.deriveProtocolFeeConfig(address);
-      // ID_AUTH_ISM could be OPStackHook, ERC5164Hook or LayerZeroV2Hook
-      // For now assume it's OP_STACK
-      case OnchainHookType.ID_AUTH_ISM:
-        return this.deriveOpStackConfig(address);
-      default:
-        throw new Error(
-          `Unsupported HookType: ${OnchainHookType[onchainHookType]}`,
-        );
+    try {
```
ltyu, line 89: This `try-catch` will catch all errors including the `hookType()` call. This is only needed for contracts that exist before the `hookType` enum was created such fuji mailbox's defaultHook.

typescript/sdk/src/hook/EvmHookReader.test.ts
```diff
@@ -186,6 +186,27 @@ describe('EvmHookReader', () => {
...
+    const mockContract = {
+      // No type
+      owner: sandbox.stub().resolves(mockOwner),
+    };
+    sandbox
+      .stub(MerkleTreeHook__factory, 'connect')
+      .returns(mockContract as unknown as MerkleTreeHook);
+    sandbox
+      .stub(IPostDispatchHook__factory, 'connect')
+      .returns(mockContract as unknown as IPostDispatchHook);
```
jmrossy, line 203: Does the sandbox need to be reset after the test? Maybe mocha does it automatically, I can't quite remember

typescript/sdk/src/hook/EvmHookReader.ts
jmrossy: Any reason we can't do that in this PR?

jmrossy (commented):
Most lgtm but I'd like to see your question to Paul answered. It does seem relevant that this reader would fail open now, meaning its consumers can't differentiate from an empty config and a failure to read.

typescript/sdk/src/hook/EvmHookReader.ts
ltyu: mostly because I haven't figured out a good place to put this yet. 

I need to change the CoreReader to do the same, so i'll need to add this fn in it too.

typescript/sdk/src/hook/EvmHookReader.test.ts
ltyu: this is done in the `after()` above

ltyu:
> Most lgtm but I'd like to see your question to Paul answered. It does seem relevant that this reader would fail open now, meaning its consumers can't differentiate from an empty config and a failure to read.

Yup, good point, and something I needed help considering. The prior deriveHookConfig() only throws if we are trying to derive an unsupported hookType. In this case, we return `{}` if 

1) hookType is unsupported
2) Any other failures e.g., hook does not have a `hookType()` at all, so the prior call will fail.

Returning a `{}` can be advantageous because this enables the consumer to continue. For example, if `hyperlane core read` tries to derive the mailbox's defaultHook, but it doesn't have one, it can continue and return the rest of it's Core config, and leave defaultHook as `{}`.

typescript/sdk/src/hook/EvmHookReader.ts
paulbalaji: I'm onboard with returning an object instead of throwing an error, but `{}` is not a valid `DerivedHookConfig` so we may hit problems where we're trying to read a field we expect e.g. `address` on an empty object

I'd be more in favour of changing the function signature to return `Promise<DerivedHookConfig | undefined>`, and then updating callers of `deriveHookConfig` to handle it how they see fit

(and ditto for the other readers in the follow-up PR)

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -85,33 +86,46 @@ export class EvmHookReader implements HookReader {
...
+        case OnchainHookType.ID_AUTH_ISM:
+          return this.deriveOpStackConfig(address);
+        default:
+          throw new Error(
+            `Unsupported HookType: ${OnchainHookType[onchainHookType]}`,
+          );
+      }
+    } catch (e) {
+      this.logger.debug(
+        `Hook deriving failed for hook: ${address} with error ${e}`,
```
nbayindirli, line 124: let's also add `onchainHookType` like
```suggestion
        `Failed to derive ${onchainHookType} hook type (${address}): ${e}`,
```

paulbalaji: yeah I prefer this over my comment

typescript/sdk/src/hook/EvmHookReader.ts
nbayindirli: ++

typescript/sdk/src/hook/EvmHookReader.test.ts
```diff
@@ -186,6 +186,27 @@ describe('EvmHookReader', () => {
...
+
+    // top-level method infers hook type
+    const hookConfig = await evmHookReader.deriveHookConfig(mockAddress);
+    expect(hookConfig).to.deep.equal({});
```
paulbalaji, line 207: need to update test

paulbalaji (approved):

jmrossy:
> Yup, good point, and something I needed help considering.

@ltyu My recommendation for a situation like this would be to either return a null-ish value (as was done here now it seems) or a result monad. Nam's a big fan of result monad. I like them too but should point out we don't really use them in the SDK/CLI yet so it would come with some required boilerplate or a lib.

PR #4066, opened 2024-06-27
chore: add shebang to build.sh to specify bash as the interpreter

assafmo:
the `[[` command isn't available on e.g. zsh


tkporter (approved):
thanks 🙏

paulbalaji (approved):

PR #4067, opened 2024-06-27
fix(e2e): hardcode kathy chains for e2e in send-test-messages.ts

paulbalaji:
resolves https://github.com/hyperlane-xyz/issues/issues/1291
- remove test4 from kathy test
- drive-by tidy-up of `test4` metadata
- resets e2e test timeout back to original figure
- reuse original e2e rust cache

gonna keep the original commits on here as a trail of thought

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -188,7 +188,7 @@ jobs:
 
   e2e-matrix:
     runs-on: larger-runner
-    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main') || github.event_name == 'merge_group'
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main' || github.base_ref == 'cli-2.0') || github.event_name == 'merge_group'
     needs: [yarn-build]
     strategy:
       matrix:
@@ -215,7 +215,7 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v2-${{ runner.os }}-rust-cache"
+          prefix-key: "v1-${{ runner.os }}-rust-cache"
           shared-key: ${{ matrix.e2e-type }}
           workspaces: |
             ./rust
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -215,7 +215,7 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v2-${{ runner.os }}-rust-cache"
+          prefix-key: "v1-${{ runner.os }}-rust-cache"
```
paulbalaji, line 218: revert https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3981/commits/45dc4f99c90dccbce5c46e9e77a58b576f95f5a2

.github/workflows/test.yml
```diff
@@ -266,7 +266,7 @@ jobs:
         working-directory: ./rust
         env:
           E2E_CI_MODE: 'true'
-          E2E_CI_TIMEOUT_SEC: '3600'
+          E2E_CI_TIMEOUT_SEC: '600'
```
paulbalaji, line 269: revert https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3981/commits/cf253328a07f1e0da340afcbd0edead43aa48c9e and https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3981/commits/d51813a1da224626a63e42507299a895f9f0dcdd

.github/workflows/test.yml
```diff
@@ -188,7 +188,7 @@ jobs:
 
   e2e-matrix:
     runs-on: larger-runner
-    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main') || github.event_name == 'merge_group'
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main' || github.base_ref == 'cli-2.0') || github.event_name == 'merge_group'
```
paulbalaji, line 191: required to trigger e2e/cli-e2e for this PR

daniel-savu (approved):
not sure if we have doc or helper comments for the kathy CLI, but we should mention that it's hardcoded to only the three test networks now

PR #4071, opened 2024-06-27
feat(infra): Add xERC20 limits monitoring capabilities

### LINKED ISSUE #4047: Migrate warp monitoring to CLI & xERC20 limit coverage
### Context
- We have existing warp route monitoring for all our deployed warp routes monitoring collateral & synthetic imbalances. 
- We also have a quickly shipped CLI command https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4020 for checking the mint/burn limits on xERC20 warp routes for Renzo
- We want to productionize the limit checking and have it live in a grafana dashboard that we can share with Renzo

### Solution
1. Update warp route monitoring to reflect the latest SDK/CLI structure including WarpCore
2. Merge in or recreate the limit checking ability within warp route monitoring
3. Expose warp route monitoring via CLI command
4. Create a public grafana dashboard w alerts for Renzo to monitor when limits are hit

Mo-Hussain:
### Description

- Add support for pushing limits metrics for xERC20 wrap routes
- For now only push limits metrics for xERC20 wrap routes, not yet supporting warp route balances 

### Related issues

 - Fixes #4047 
### Testing

Manual

### DIFF
--- a/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
+++ b/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
@@ -90,7 +90,7 @@ async function main(): Promise<boolean> {
   const tokenConfig: WarpRouteConfig =
     readWarpRouteConfig(filePath).data.config;
 
-  // await checkTokenBalances(checkFrequency, tokenConfig);
+  await checkTokenBalances(checkFrequency, tokenConfig);
 
   if (
     Object.values(tokenConfig).some((token) => token.type === TokenType.XERC20)
@@ -366,7 +366,6 @@ async function checkXERC20Limits(checkFrequency: number) {
 
   setInterval(async () => {
     try {
-      logger.info('Checking xERC20 limits');
       const xERC20Limits = await getXerc20Limits(warpCoreConfig, multiProvider);
       logger.info('xERC20 Limits:', xERC20Limits);
       updateXERC20LimitsMetrics(xERC20Limits);
--- a/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
+++ b/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
@@ -3,6 +3,11 @@ import { ethers } from 'ethers';
 import { Gauge, Registry } from 'prom-client';
 import yargs from 'yargs';
 
+import {
+  HypXERC20Lockbox__factory,
+  HypXERC20__factory,
+  IXERC20__factory,
+} from '@hyperlane-xyz/core';
 import { ERC20__factory } from '@hyperlane-xyz/core';
 import {
   ChainMap,
@@ -10,8 +15,11 @@ import {
   CosmNativeTokenAdapter,
   CwNativeTokenAdapter,
   MultiProtocolProvider,
+  MultiProvider,
   SealevelHypCollateralAdapter,
+  TokenStandard,
   TokenType,
+  WarpCoreConfig,
   WarpRouteConfig,
   WarpRouteConfigSchema,
 } from '@hyperlane-xyz/sdk';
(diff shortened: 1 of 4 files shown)


typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -240,4 +252,146 @@ export function updateTokenBalanceMetrics(
...
+        };
+      }),
+  );
+
+  return result;
+}
+
+async function checkXERC20Limits(checkFrequency: number) {
+  const registry = await getRegistry();
+  const symbol = 'EZETH';
```
tkporter, line 347: A couple things about this version:
- doesn't work for things other than EZETH - imo we should make this configurable before merging
- breaks if there is more than one EZETH deployment

In the future, I wonder if we should make use of warp route IDs, which were added kinda recently to identify specific warp route deployments.

Some context on the IDs:
- basically option 1 as described here https://discord.com/channels/935678348330434570/1240001006050213918/1242481517352063038
- implemented here https://github.com/hyperlane-xyz/hyperlane-registry/pull/43/files#diff-a7ae5c67bef044df975c964ec7c0b32839ae904c5c669de9ae55dfb0f8674e87R62 and https://github.com/hyperlane-xyz/hyperlane-registry/pull/43/files#diff-b8ce851525e7bc41ba0aaec5c12987e029ae77a4fd9c0c60d81c766ee73130e1R54-R57

Iiuc the format of a warp ID is `${token}/${sortedChains.join('-')}`, e.g. `EZETH/arbitrum-base-blast-bsc-ethereum-linea-mode-optimism` (can be found by looking at the names / path of https://github.com/hyperlane-xyz/hyperlane-registry/tree/main/deployments/warp_routes/EZETH)

We could pass the warp ID as a flag to the deploy script that ensures the warp ID exists, and then that could pass the warp ID as a flag to this script via helm.

However this does give us a couple issues:
- these IDs are super verbose and cumbersome to write / remember
- atm the name of the infra file that's passed into the deploy script with the config is used to figure out what to name the helm / k8s deployment and resources (https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/scripts/warp-routes/helm.ts#L25). moving to using the warp ID instead won't work bc the resource name requirements in k8s are pretty conservative - `EZETH/arbitrum-base-blast-bsc-ethereum-linea-mode-optimism` would def be too long and it'd complain about caps and `/`. So we'd need to figure out a different naming setup
- It's a little awkward having the xerc20 path involve the registry and the other non-xerc20 path not involve the registry
- Honestly am not certain if warp IDs are a notion that will stick around long or not. We can get an indication from the product team here tho

So I see our options as the following, in rough order of lowest lift -> highest lift:
1. Just leave this as-is, don't accommodate any other xERC20s other than EZETH
2. leave mostly as-is, but get 'EZETH' here from `EZETH-deployments.yaml` so it's a little more dynamic / configurable
3. kick the can further - for now just shamelessly duplicate and copy https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/deployments/warp_routes/EZETH/arbitrum-base-blast-bsc-ethereum-linea-mode-optimism-config.yaml into `EZETH-deployments.yaml` here in infra (with the appropriate schema). We don't get anything from the registry, just the file passed in. Then later we can revisit to do something more "correct" which is probably close to what I describe above
4. Implement something close to what I describe above now, knowing that we'll have deviating paths for xerc20 vs not, and accepting we'll have to make some changes to our infra setup to get this working
5. full on commit to moving to the registry for everything with a setup similar to what I describe above for both the xerc20 and erc20 path

I'm kinda partial to (3) to get this merged and over the line with minimal effort while still being configurable / usable for other tokens in the future, and then we schedule follow up work to do (5) in a way that feels right. But am open to chatting about all these, I don't feel very strongly

Mo-Hussain: I guess if we do 5 eventually, 3 now feels like a move and work in the wrong direction

tkporter (approved):
Nice 🚀

Issue #4049, opened 2024-06-25
Key funder: make IGP claim threshold configurable via CLI args


tkporter:
## Problem

Funding amounts are configurable, IGP claim thresholds aren't. This means we need to make a code change when we add a new chain to start claiming from IGPs 😔

## Solution

- Similar to what we do with desired balances, set claim thresholds via the CLI

## Nice to have

- Maybe default to a claim threshold based off the relayer desired balances? Dunno

PR #4074, opened 2024-06-28
feat(keyfunder): make IGP claim thresholds configurable via command line args

paulbalaji:
hit successor to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3284
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4049

- igp claim thresholds are moved to command line args instead, like desired balances
	- making it easier to tweak IGP claim thresholds
- default claim threshold is 20% of desired balance for that chain
	- roughly follows what I've seen with the other balance vs igp claim thresholds
- adding thresholds for all chains we've configured a desired relayer balance for
- drive-by fixing keyfunder logging to include the relevant metadata

testing:
- ran locally and successfully performed igp claims on [blast](https://blastscan.io/tx/0x314d83d16e0406f77731befc7e8387dc946af01f7dc2200a426ff9748ad76d27), [linea](https://lineascan.build/tx/0xa5f4731ebcff073dacc6c15326196002bafd9578059688fd2f29db1a2742cb14), [viction](https://www.vicscan.xyz/tx/0x39921d3018a7b019518aa7a55a45ff29cce379bb9aabf0910c14423bd88689d5) & more


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -520,35 +509,51 @@ class ContextFunder {
   }
 
   private async attemptToClaimFromIgp(chain: ChainName) {
-    const igpClaimThresholdEther = igpClaimThresholdPerChain[chain];
+    const igpClaimThresholdEther =
+      this.igpClaimThresholdPerChain[chain] ||
+      this.desiredBalancePerChain[chain];
```
Mo-Hussain, line 514: Why do we want to fall back to desired balance per chain here?

Mo-Hussain (commented):
Just so context, why do we want to maintain a desired balance for these contracts?

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -520,35 +509,51 @@ class ContextFunder {
   }
 
   private async attemptToClaimFromIgp(chain: ChainName) {
-    const igpClaimThresholdEther = igpClaimThresholdPerChain[chain];
+    const igpClaimThresholdEther =
```
Mo-Hussain, line 512: nit: the intention here was not obvious for me on initial reading. 

Maybe something like 
```
let igpClaimThreshold = this.igpClaimThresholdPerChain[chain]
if (!igpClaimThreshold) {
   // set igpClaimThreshold to a 5th of the desiredBalancePerChain 
}
```

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -520,35 +509,51 @@ class ContextFunder {
...
+      this.igpClaimThresholdPerChain[chain] ||
```
Mo-Hussain, line 513: Will this `igpClaimThresholdPerChain` include the hardcoded config if none is passed as an arg?

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -340,11 +319,14 @@ class ContextFunder {
       },
     );
 
-    logger.info('Successfully read keys for context from file', {
-      filePath,
-      readOnlyKeysPerChain,
-      context,
-    });
+    logger.info(
```
Mo-Hussain, line 322: Has the interface for these methods changed?

paulbalaji:
> Just so context, why do we want to maintain a desired balance for these contracts?

so we don't claim every time the keyfunder is run - only when it goes above a certain threshold where it's worth the effort to claim

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: I think it changed a while back but we haven't updated these - none of the metadata here was being logged originally

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: not directly - what happens is that we read the config for a given environment and then run a keyfunder with the respective command line args based on the config

when we deploy/update keyfunder in k8s, what happens is we update the args configured in the cron job

so `igpClaimThresholdPerChain` here should only include the chains we pass in, to avoid being hardcoded to a specific environment

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: it ends up making the code more verbose than it needs to imo, here it's kind of concise and to-the-point that if we don't have a config for a given chain we skip it

would rather just leave a comment for future readers

paulbalaji: e.g.
```ts
    const igpClaimThresholdEther = this.igpClaimThresholdPerChain[chain];
    let igpClaimThreshold: BigNumber;
    if (!igpClaimThresholdEther) {
      logger.warn(
        { chain },
        `No IGP claim threshold for chain ${chain}, inferring threshold from desired balance...`,
      );

      const desiredBalanceEther = this.desiredBalancePerChain[chain];
      if (!desiredBalanceEther) {
        logger.warn(
          { chain },
          `No desired balance for chain ${chain}, skipping IGP claim`,
        );
        return;
      }

      igpClaimThreshold = ethers.utils.parseEther(desiredBalanceEther).div(5);
    } else {
      igpClaimThreshold = ethers.utils.parseEther(igpClaimThresholdEther);
    }
```

Mo-Hussain: Comment for future readers is good :)

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
Mo-Hussain: I ask because where possible, I think it would be good to fall back on pre configured defaults like
```
 igpClaimThresholdPerChain: {
    arbitrum: '0.1',
    ancient8: '0.1',
    avalanche: '2',
    base: '0.1',
    blast: '0.1',
    bsc: '0.3',
    celo: '5',
    ethereum: '0.2',
    fraxtal: '0.1',
    gnosis: '5',
    inevm: '3',
    linea: '0.1',
    mantapacific: '0.1',
    mode: '0.1',
    moonbeam: '5',
    optimism: '0.1',
    polygon: '20',
    polygonzkevm: '0.1',
    redstone: '0.1',
    scroll: '0.1',
    sei: '2',
    viction: '2',
    zetachain: '20',
  },
```

instead of a 5th of the desired balance. Is that what happens in this change?

Mo-Hussain (approved):

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: discussed offline - explained how keyfunder config related to args

PR #4084, opened 2024-07-02
chore: bump Solana-origin nautilus fees

tkporter:
### Description

Solana-origin fees to Nautilus are now about 25 cents. It used to be sub-cent. Nautilus origin is about 71 cents (37.889581133408400000 ZBC) at the moment, so this is left untouched.

### DIFF
--- a/rust/sealevel/environments/mainnet2/gas-oracle-configs.json
+++ b/rust/sealevel/environments/mainnet2/gas-oracle-configs.json
@@ -4,7 +4,7 @@
         "gasOracle": {
             "type": "remoteGasData",
             "tokenExchangeRate": "100000000000000000000",
-            "gasPrice": "3000000000",
+            "gasPrice": "300000000000",
             "tokenDecimals": 18
         }
     },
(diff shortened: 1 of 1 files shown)


rust/sealevel/environments/mainnet2/gas-oracle-configs.json
```diff
@@ -4,16 +4,16 @@
...
             "tokenDecimals": 18
         }
     },
     {
         "domain": 22222,
         "gasOracle": {
             "type": "remoteGasData",
-            "tokenExchangeRate": "4700000000000000",
-            "gasPrice": "1000000000",
+            "tokenExchangeRate": "6010517751000000",
```
daniel-savu, line 15: is this SOL/ZBC?  Already seeing a ratio of 7813 but probably doesn't matter much

daniel-savu (approved):

rust/sealevel/environments/mainnet2/gas-oracle-configs.json
tkporter: tbh I did a view call to see what was currently configured and it was `6010517751000000`, and i just adjusted the gas price to be a factor higher instead of trying to have an accurate token ratio. I believe Nam did the same for BSC and atm the end cost is what matters more so this was easiest lol

PR #4089, opened 2024-07-02
feat(relayer): ListOperations api endpoint

daniel-savu:
### Description

Adds a `list_operations` API endpoint to the relayer, which returns all operations in its prepare queue.

Example output from `GET http://0.0.0.0:9092/list_operations?destination_domain=13371`

```
[
  {
    "app_context": null,
    "message": {
      "body": [
        18,
        52
      ],
      "destination": 13371,
      "nonce": 1,
      "origin": 13373,
      "recipient": "0x000000000000000000000000927b167526babb9be047421db732c663a0b77b11",
      "sender": "0x000000000000000000000000927b167526babb9be047421db732c663a0b77b11",
      "version": 3
    },
    "num_retries": 13,
    "status": {
      "Retry": "CouldNotFetchMetadata"
    },
    "submitted": false,
    "type": "PendingMessage"
  },
  {
    "app_context": null,
    "message": {
      "body": [
        18,
        52
      ],
      "destination": 13371,
      "nonce": 2,
      "origin": 13372,
      "recipient": "0x000000000000000000000000927b167526babb9be047421db732c663a0b77b11",
      "sender": "0x000000000000000000000000927b167526babb9be047421db732c663a0b77b11",
      "version": 3
    },
    "num_retries": 13,
    "status": {
      "Retry": "CouldNotFetchMetadata"
    },
    "submitted": false,
    "type": "PendingMessage"
  }
]
```

### Drive-by changes

Makes `PendingOperation` serializable, which means switching from JSON format to e.g. CSV (for easily filtering / aggregating in excel) should be very easy.

### Backward compatibility

Yes

### Testing

Manual and Unit Tests


rust/Cargo.toml
```diff
@@ -182,6 +182,7 @@ tracing-error = "0.2"
 tracing-futures = "0.2"
 tracing-subscriber = { version = "0.3", default-features = false }
 tracing-test = "0.2.2"
+typetag = "0.2"
```
daniel-savu, line 185: `Serialize` isn't object safe so we can't normally use it for our `Box<dyn PendingOperation>` types, but this crate works around that by creating a `Serialize`-able enum whose each item is the struct `impl` of the `PendingOperation` trait.

daniel-savu:
Slightly annoying how the operation status is serialized:
```
    "status": {
      "Retry": "CouldNotFetchMetadata"
    },
```
whereas I would've wanted the value to have no nesting: `status: "Retry(CouldNotFetchMetadata)"`, because in csv you get a new column for each nested field within status, e.g. `status/Retry` and `status/FirstPrepareAttempt`. Not that big of a deal though

tkporter:
@daniel-savu are there any alternative enum representations https://serde.rs/enum-representations.html that would help? maybe the internal one where we can have `"type": "Retry"` instead?

tkporter (approved):
nice 🔥 

might be nice in the future to include gas payments but for now I think this is sufficient

daniel-savu:
> @daniel-savu are there any alternative enum representations https://serde.rs/enum-representations.html that would help? maybe the internal one where we can have `"type": "Retry"` instead?

turns out there is a way, since individual enum member can be `untagged` (https://github.com/serde-rs/serde/pull/2403), and for `Retry(ReprepareReason)` it wouldn't matter if `retry` isn't part of the serialization. But this actually breaks DB storage so we may want to think about this twice - probably still ok to do now, while we don't have that many GBs of messages

PR #4091, opened 2024-07-02
fix: use sortMapEntries in writeYaml

### LINKED ISSUE #4057: CLI artifact fields should be ordered alphabetically
See request [here](https://github.com/hyperlane-xyz/hyperlane-registry/pull/73?notification_referrer_id=NT_kwDOAArYX7IxMTIyNzg2NjYwMzo3MTA3NTE#issuecomment-2189170498)

And some more context [here](https://github.com/hyperlane-xyz/hyperlane-registry/pull/39)

bertux:
### Description

It includes usage of sortMapEntries option in yamlStringify function called in writeYaml function.

### Drive-by changes

None

### Related issues

- Fixes #4057 

### Backward compatibility

Yes
Because I've converted the argument 2 into the indent option as documented in yaml library.

### Testing

Manual by trying to reproduce the issue in the same way.

### DIFF
--- a/.changeset/five-timers-reflect.md
+++ b/.changeset/five-timers-reflect.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
+---
+
+Fix the missing sorting in the YAML file generated
--- a/typescript/cli/src/config/chain.ts
+++ b/typescript/cli/src/config/chain.ts
@@ -95,7 +95,7 @@ export async function createChainConfig({
 
   const parseResult = ChainMetadataSchema.safeParse(metadata);
   if (parseResult.success) {
-    logGreen(`Chain config is valid, writing sorted to registry:`);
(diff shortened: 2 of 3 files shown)


bertux:
Manual test didn't sort when trying `yarn hyperlane registry init` it generated the same content than before.

bertux:
The log test succeeded so the stringify syntax is correct but it has to be used also in the registry package when generating the YAML again.

bertux:
@avious00 I can't further without more guidance because I've not found the source code where the chain metadata is written so I can't add sorting there and I've added the unsorted keyword in the log to warn about it's still unsorted.
All other files written by utils/file.ts are sorted.
Let me know how you would like to continue with this PR.

bertux:
For the registry part I've created the PR at https://github.com/hyperlane-xyz/hyperlane-registry/pull/85
Now I only wait for your review.

jmrossy (approved):
Looks good!

PR #4093, opened 2024-07-03
feat(infra): strongly type keyfunder config

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3732

- strongly type the keyfunder config
	- new chains added to an environment's supported chains list must be configured
- opens up the path to strongly type more of the env config down the line

missing 1 chain:
image

missing multiple chains:
image


tkporter (approved):
wooo lovely

PR #4094, opened 2024-07-03
feat(cli): Add hyperlane warp apply

ltyu:
### Description
- Adds `hyperlane warp apply`

### Related issues
- Fixes https://github.com/hyperlane-xyz/issues/issues/1190

### Backward compatibility
Yes

### Testing
Manual

To test: `yarn hyperlane warp apply --warp`

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -50,6 +50,7 @@ export const warpCommand: CommandModule = {
   describe: 'Manage Hyperlane warp routes',
   builder: (yargs) =>
     yargs
+      .command(apply)
       .command(deploy)
       .command(init)
       .command(read)
@@ -58,6 +59,13 @@ export const warpCommand: CommandModule = {
       .demandCommand(),
 
   handler: () => log('Command required'),
+};
+
+export const apply = {
+  command: 'apply',
+  describe: 'Update Warp Route contracts',
+  builder: {},
+  handler: async () => {},
 };
 
 export const deploy: CommandModuleWithWriteContext<{
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -9,11 +9,9 @@ import {
 } from '@hyperlane-xyz/core';
 import {
   ChainMap,
-  EvmERC20WarpModule,
   EvmERC20WarpRouteReader,
   TokenStandard,
   WarpCoreConfig,
-  proxyFactoryFactoriesAddresses,
 } from '@hyperlane-xyz/sdk';
 import { objMap, promiseObjAll } from '@hyperlane-xyz/utils';
 
@@ -27,7 +25,7 @@ import {
   CommandModuleWithWriteContext,
 } from '../context/types.js';
 import { evaluateIfDryRunFailure } from '../deploy/dry-run.js';
-import { runWarpRouteDeploy } from '../deploy/warp.js';
+import { runWarpRouteApply, runWarpRouteDeploy } from '../deploy/warp.js';
 import { log, logGray, logGreen, logRed, logTable } from '../logger.js';
 import { sendTestTransfer } from '../send/transfer.js';
 import { indentYamlOrJson, writeYamlOrJson } from '../utils/files.js';
@@ -83,6 +81,8 @@ export const apply: CommandModuleWithWriteContext<{
     },
   },
   handler: async ({ context, config, symbol, warp }) => {
+    logGray(`Hyperlane Warp Apply`);
+    logGray('------------------------------------------------');
     let warpCoreConfig: WarpCoreConfig;
     if (symbol) {
       warpCoreConfig = await selectRegistryWarpRoute(context.registry, symbol);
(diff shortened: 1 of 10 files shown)


typescript/cli/src/commands/warp.ts
```diff
@@ -60,6 +63,75 @@ export const warpCommand: CommandModule = {
...
+              deployedTokenRoute: warpCoreReduced[chain as any].addressOrDenom!,
+            },
+          },
+        );
+        return evmERC20WarpModule.update(config);
+      }),
+    );
+
+    // Loop through each chain, and then loop through each update. ugh.
+    // @TODO maybe use concurrency for each chain
```
paulbalaji, line 124: promise per chain should be fine, as long as we catch errors and log them appropriately and display the chains that a deployment was successful or failed on

typescript/cli/src/commands/warp.ts
```diff
@@ -60,6 +63,75 @@ export const warpCommand: CommandModule = {
...
 };
 
+export const apply: CommandModuleWithWriteContext<{
+  config: string;
+  symbol?: string;
+  warp: string;
+}> = {
+  command: 'apply',
+  describe: 'Update Warp Route contracts',
+  builder: {
```
paulbalaji, line 73: should we also allow a `--chains` filter so you can deploy to a specific subset of the chains defined in the config? e.g. say you do a deploy across chains but it fails on a particular chain, you can redeploy just on a particular chain

ltyu: I like where you head is. 

I think i'll leave this to a future update if its requested. As of now, if 1 chain fails, and then we run update again, it should skip the other chains because the config diffs will find the successful chains to be the same.

paulbalaji: ok cool sg

typescript/cli/src/commands/warp.ts
```diff
@@ -60,6 +62,46 @@ export const warpCommand: CommandModule = {
...
+      demandOption: false,
+    },
+    warp: {
+      ...warpCoreConfigCommandOption,
+      demandOption: false,
+    },
+  },
+  handler: async ({ context, config, symbol, warp }) => {
+    logGray(`Hyperlane Warp Apply`);
+    logGray('------------------------------------------------');
```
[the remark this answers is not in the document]
ltyu: no style guide. i just copy and pasted from somewhere 😅

typescript/sdk/src/deploy/schemas.ts
```diff
@@ -13,3 +13,7 @@ export const ProxyFactoryFactoriesSchema = z.object({
   staticAggregationHookFactory: z.string(),
   domainRoutingIsmFactory: z.string(),
 });
+
+export type proxyFactoryFactoriesAddresses = z.infer<
```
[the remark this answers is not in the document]
ltyu: fixed

typescript/cli/src/commands/warp.ts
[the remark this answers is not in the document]
ltyu: Created a TODO for it

typescript/cli/src/deploy/warp.ts
```diff
@@ -313,3 +327,57 @@ async function getWarpCoreConfig(
...
+          }
+
+          logGreen(`Warp config updated on ${chain} chain.`);
+        } else {
+          logGreen(
+            `Warp config on ${chain} chain is the same as target. No updates needed.`,
+          );
+        }
+      } catch (e) {
+        logRed(`Warp config on ${chain} chain failed to update.`, e);
```
paulbalaji, line 379: nit, from what I see in a couple other commands:
```suggestion
          logGreen(`Warp config updated on ${chain}.`);
        } else {
          logGreen(
            `Warp config on ${chain} is the same as target. No updates needed.`,
          );
        }
      } catch (e) {
        logRed(`Warp config on ${chain} failed to update.`, e);
```

paulbalaji (approved):

typescript/cli/src/deploy/warp.ts
ltyu: fixed

typescript/cli/src/deploy/warp.ts
```diff
@@ -313,3 +327,57 @@ async function getWarpCoreConfig(
...
+  } = params;
+
+  // Addresses used to get static Ism factories
+  const addresses = await registry.getAddresses();
+
+  // Convert warpCoreConfig.tokens[] into a mapping of { [chainName]: Config }
+  // This allows O(1) reads within the loop
+  const warpCoreByChain = Object.fromEntries(
+    warpCoreConfig.tokens.map((token) => [token.chainName, token]),
+  );
```
jmrossy, line 345: This is fine but realistically if n < 1000-ish if really doesn't matter if it's O(1), O(n), or even O(n^2)

typescript/cli/src/deploy/warp.ts
```diff
@@ -313,3 +327,57 @@ async function getWarpCoreConfig(
...
+  // Attempt to update Warp Routes
+  // Can update existing or deploy new contracts
+  logGray(`Comparing target and onchain Warp configs`);
+  await promiseObjAll(
+    objMap(warpDeployConfig, async (chain, config) => {
+      try {
+        // Update Warp
+        config.ismFactoryAddresses = addresses[
+          chain
+        ] as ProxyFactoryFactoriesAddresses;
```
jmrossy, line 356: Why is a cast needed here?

typescript/cli/src/deploy/warp.ts
```diff
@@ -313,3 +327,57 @@ async function getWarpCoreConfig(
...
+
+          logGreen(`Warp config updated on ${chain}.`);
+        } else {
+          logGreen(
+            `Warp config on ${chain} is the same as target. No updates needed.`,
+          );
+        }
+      } catch (e) {
+        logRed(`Warp config on ${chain} failed to update.`, e);
+      }
```
jmrossy, line 380: Are you certain it's better to continue updating the rest if there's a failure instead of stopping and letting the user debug the issue?

jmrossy (commented):
Mostly lgtm, just a few comments

typescript/cli/src/deploy/warp.ts
ltyu: Not certain. But i'm open to changing this as requirements change. As of now, there are none.

The reason I did this was because we are only doing ism updates, which are mostly isolated for each chain. When updating an ISM we may also deploy one, which is committed right away. Then after deploying, we set a warp route's ISM as a transaction.

If we revert here because an update failed in another chain, we've potentially deployed an ISM for no reason.

typescript/cli/src/deploy/warp.ts
ltyu: Because addresses[chain] returns a Record<string, string> which seems to be no specific enough
```
Type 'Record<string, string>' is missing the following properties from type '{ staticMerkleRootMultisigIsmFactory: string; staticMessageIdMultisigIsmFactory: string; staticAggregationIsmFactory: string; staticAggregationHookFactory: string; domainRoutingIsmFactory: string; }': staticMerkleRootMultisigIsmFactory, staticMessageIdMultisigIsmFactory, staticAggregationIsmFactory, staticAggregationHookFactory, domainRoutingIsmFactoryts(2739)
```

typescript/cli/src/deploy/warp.ts
ltyu: True that

ltyu: I mostly did it because i like reasoning with similar data shapes. I thought it would be hard to compare an array with an object

PR #4097, opened 2024-07-04
feat: validator dashboard template

daniel-savu:
### Description

Replaces the old (and probably deprecated) validator dashboard template with a new one that shows metrics that can be alerted on:
Screenshot 2024-07-04 at 14 25 53

### Drive-by changes

Uses the "External Sharing" grafana feature to generate a json that removes AW-specific data from the relayer template


tkporter (approved):
ngl didn't even realize we were checking these in already

PR #4098, opened 2024-07-04
chore: Revert "feat(hyp-ethereum): gas escalator middleware (#3852)"

daniel-savu:
This reverts commit c3c002a8fe4870c7e462cd5073365a7f9a2235bc.

### Description

The gas escalator took up to 28gb of memory, and even though it was only ever released on RC, it must have taken up too many cluster resources, ending up starving the `hyperlane` relayer. Reverting for now until we have a better idea

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2725,7 +2725,7 @@ dependencies = [
 [[package]]
 name = "ethers"
 version = "1.0.2"
-source = "git+https://github.com/hyperlane-xyz/ethers-rs?tag=2024-06-27-2#df3510cdfdd18ee6e6e9b0d7bad61f405c8abc49"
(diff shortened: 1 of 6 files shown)


tkporter (approved):

daniel-savu:
@tkporter suggested that since we build a bunch of short-lived providers, we may end up with "zombie" gas escalator tasks, that are spawned each time the escalator middleware. Note how the join handle isn't stored anywhere [here](https://github.com/hyperlane-xyz/ethers-rs/blob/df3510cdfdd18ee6e6e9b0d7bad61f405c8abc49/ethers-middleware/src/gas_escalator/mod.rs#L203).

Proposed next steps are to capture the `JoinHandle` and implement `Drop` on `GasEscalatorMiddleware` to clean up the escalator task. We'd measure this with `tokio-console` in e2e by running with `dependencyTrace` and checking if the number of tokio tasks scales with the number of messages.

PR #4099, opened 2024-07-05
fix: incorrect use of blacklist variable

lyfsn:
### Description

Otherwise, the whitelist will not work; any whitelist parameter will function as a blocklist.


tkporter (approved):
Wow, very surprised (and ashamed) to see this! Great find. Thank you for the fix!

PR #4101, opened 2024-07-05
feat(sdk): Handle subdirectories for the folder in S3Validator

### LINKED ISSUE #4100: Update S3 Validator to handle subdirectories for the folder
Our S3Validator class assumes that the S3 URI will have single directory where validators will save checkpoints. Some validators are saving checkpoints in subdirectories.

Our agents are able to handle subdirectories and can read the checkpoints. We want to update our S3Validator class to be aligned with the agents and work in the same way

Mo-Hussain:
### Description

- Set the folder in the `S3Config` to include any subdirectories and not just the first folder
- This brings the SDK inline with current behavior in the agents

### Related issues

- Fixes #4100 

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/aws/validator.ts
+++ b/typescript/sdk/src/aws/validator.ts
@@ -39,7 +39,7 @@ export class S3Validator extends BaseValidator {
         const s3Config = {
           bucket: pieces[0],
           region: pieces[1],
-          folder: pieces[2],
+          folder: pieces.slice(2).join('/'),
           caching: true,
         };
         const s3Bucket = new S3Wrapper(s3Config);
--- a/.changeset/tiny-lions-kiss.md
+++ b/.changeset/tiny-lions-kiss.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': patch
+---
+
+Handle subdirectories for the folder in S3Validator class


typescript/sdk/src/aws/validator.ts
```diff
@@ -39,7 +39,7 @@ export class S3Validator extends BaseValidator {
         const s3Config = {
           bucket: pieces[0],
           region: pieces[1],
-          folder: pieces[2],
+          folder: pieces.slice(2).join('/'),
```
paulbalaji, line 42: might be my smooth brain but can you give me an example of before/after this fix pls

jmrossy (approved):
lgtm

typescript/sdk/src/aws/validator.ts
Mo-Hussain: The existing implementation: Coinsummer are storing the objects in a sub directory, `s3://cs-avs-hyperlane/us-east-1/validator/avs1/ethereum`, we are looking in `validator` and not finding anything.

new implementation will look in `validator/avs1/ethereum`

daniel-savu (approved):
nice, just ran and it we have 30 validators running correctly according to the script (including coinsummer) 🔥

PR #4106, opened 2024-07-08
feat(relayer): new status label for IGP not found

daniel-savu:
### Description

Adds new status label to relayer opqueue for operations whose gas payments were not found. This now creates a distinction between operations without a payment and those with insufficient payment.

Helps better uncover cases where hook indexing misses events because there is no receipt yet

### DIFF
--- a/rust/agents/relayer/src/msg/gas_payment/mod.rs
+++ b/rust/agents/relayer/src/msg/gas_payment/mod.rs
@@ -93,11 +93,12 @@ impl GasPaymentEnforcer {
             message_id: msg_id,
             destination: message.destination,
         };
-        let Some(current_payment) = self
+        let current_payment_option = self
             .db
-            .retrieve_gas_payment_by_gas_payment_key(gas_payment_key)?
-        else {
-            return Ok(GasPolicyStatus::NoPaymentFound);
+            .retrieve_gas_payment_by_gas_payment_key(gas_payment_key)?;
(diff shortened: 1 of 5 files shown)


rust/agents/relayer/src/msg/gas_payment/mod.rs
```diff
@@ -80,15 +87,18 @@ impl GasPaymentEnforcer {
...
             message_id: msg_id,
             destination: message.destination,
         };
-        let current_payment = self
+        let Some(current_payment) = self
             .db
-            .retrieve_gas_payment_by_gas_payment_key(gas_payment_key)?;
+            .retrieve_gas_payment_by_gas_payment_key(gas_payment_key)?
+        else {
+            return Ok(GasPolicyStatus::NoPaymentFound);
```
tkporter, line 100: does this break the use case of configuring the relayer with GasPaymentEnforcementPolicy::None?

I think we'll want to keep the ability to configure that for some / all messages

daniel-savu: hmm good point. Changed it to return:
- `PolicyMet` if the policy passes, regardless of there being a gas payment or not
- `NoPaymentFound`, if the policy rejects it and there is no gas payment
- `PolicyNotMet`, if the policy rejects it in spite of there being a gas payment

tkporter (approved):

PR #4120, opened 2024-07-10
chore: set env-test timeout

paulbalaji:
- chore: set env-test timeout
- to avoid machines spinning idly for hours/days in jobs like this https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/9872063807/job/27261501530?pr=3888#step:6:44

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -404,7 +404,7 @@ jobs:
 
   env-test:
     runs-on: ubuntu-latest
-    timeout-minutes: 2
+    timeout-minutes: 3
     needs: [yarn-build]
     strategy:
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -404,6 +404,7 @@ jobs:
 
   env-test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -404,6 +404,7 @@ jobs:
 
   env-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 2
```
paulbalaji, line 407:
```suggestion
    timeout-minutes: 3
```

.github/workflows/test.yml
```diff
@@ -404,6 +404,7 @@ jobs:
 
   env-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 3
```
Mo-Hussain, line 407: Do we have a sense for how long this usually takes to run?

paulbalaji: yeah they take about 1-2mins usually, so added an extra 1 min buffer here

Mo-Hussain: sg!

Mo-Hussain (approved):

PR #4121, opened 2024-07-10
feat(utils): Set errorMessage argument for assert to required

Mo-Hussain:
### Description

- Set errorMessage argument for assert to required

### DIFF
--- a/typescript/sdk/src/ism/EvmIsmReader.ts
+++ b/typescript/sdk/src/ism/EvmIsmReader.ts
@@ -7,7 +7,6 @@ import {
   OPStackIsm__factory,
   PausableIsm__factory,
   StaticAggregationIsm__factory,
-  TestIsm__factory,
   TrustedRelayerIsm__factory,
 } from '@hyperlane-xyz/core';
 import {
@@ -44,6 +43,10 @@ export interface IsmReader {
     address: Address,
   ): Promise<WithAddress<MultisigIsmConfig>>;
   deriveNullConfig(address: Address): Promise<WithAddress<NullIsmConfig>>;
+  assertModuleType(
+    moduleType: ModuleType,
+    expectedModuleType: ModuleType,
+  ): void;
 }
 
 export class EvmIsmReader implements IsmReader {
@@ -98,7 +101,7 @@ export class EvmIsmReader implements IsmReader {
       this.provider,
     );
     const owner = await ism.owner();
-    assert((await ism.moduleType()) === ModuleType.ROUTING);
+    this.assertModuleType(await ism.moduleType(), ModuleType.ROUTING);
 
     const domains: RoutingIsmConfig['domains'] = {};
     const domainIds = await ism.domains();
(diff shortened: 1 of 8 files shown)


typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -230,8 +238,9 @@ export class EvmIsmReader implements IsmReader {
 
     // if it has VERIFIED_MASK_INDEX, it's AbstractMessageIdAuthorizedIsm which means OPStackIsm
     const opStackIsm = OPStackIsm__factory.connect(address, this.provider);
+    this.assertModuleType(await opStackIsm.moduleType(), ModuleType.NULL);
```
Mo-Hussain, line 241: @paulbalaji - pulled this guy out of the try catch as I'm not sure we want to catch it with that error message

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -214,7 +222,7 @@ export class EvmIsmReader implements IsmReader {
     try {
       const paused = await pausableIsm.paused();
       const owner = await pausableIsm.owner();
-      assert((await pausableIsm.moduleType()) === ModuleType.NULL);
+      this.assertModuleType(await pausableIsm.moduleType(), ModuleType.NULL);
```
Mo-Hussain, line 225: @paulbalaji - Is this guy redundant because `const paused = await pausableIsm.paused();` will always through an error before this is run?

.eslintrc
```diff
@@ -24,6 +24,13 @@
     "no-ex-assign": ["error"],
     "no-constant-condition": ["off"],
     "no-return-await": ["error"],
+    "no-restricted-syntax": [
```
Mo-Hussain, line 27: I realise we could also make the error message a required arg for the assert function

```
export function assert<T>(
  predicate: T,
  errorMessage?: string,
): asserts predicate {
  if (!predicate) {
    throw new Error(errorMessage ?? 'Error');
  }
}
```

typescript/sdk/src/ism/EvmIsmReader.ts
paulbalaji: i'd say to solve both the comments here do something like this just once right at the top of the deriveNullConfig function, so we do the null assert once at the top and then we go into checking which null ism it is

```ts
    const ism = IInterchainSecurityModule__factory.connect(address, this.provider);
    this.assertModuleType(await ism.moduleType(), ModuleType.NULL);
```

.eslintrc
jmrossy: Right, so why not make that change first?

jmrossy (commented):
LGTM but like you suggested, it does seem simpler to let the TS compiler handle this instead of a custom lint rule

.eslintrc
Mo-Hussain: I arrived at this right at the end after realising assert was something we implemented

Mo-Hussain: I've made the change to assert

jmrossy (approved):

PR #4127, opened 2024-07-11
feat(infra): Add script to create safes and add safe addresses for new chains

Mo-Hussain:
### Description

- Add an infra script to create a safe on specific chain using the `safeSigners.json` config where the safe signers are defined

### Testing

Manual


typescript/infra/scripts/safes/create-safe.ts
```diff
@@ -0,0 +1,87 @@
...
+  const safeAddress = await safe.getAddress();
+
+  console.log(`Safe address: ${safeAddress}`);
+  console.log(
+    `Safe url: ${safeTxServiceUrl}/home?safe=${chain}:${safeAddress}`,
+  );
+  console.log('url may not be correct, please check by following the link');
+}
+
+const getDeployerPrivateKey = async () => {
```
tkporter, line 64: we have a helper to get this more easily: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/scripts/agent-utils.ts#L310

```
const key = await getKeyForRole('mainnet3', Contexts.Hyperlane, Role.Deployer);
// fetch the key contents
await key.fetch();
// now `key.privateKey` can be used
```

typescript/infra/scripts/safes/create-safe.ts
```diff
@@ -0,0 +1,87 @@
...
+} from '../agent-utils.js';
+
+const OWNERS_FILE_PATH = 'config/environments/mainnet3/safe/safeSigners.json';
+
+async function main() {
+  const { chain, safeTxServiceUrl, threshold } = await withThreshold(
+    withSafeTxServiceUrlRequired(withChainRequired(getArgs())),
+  ).argv;
+
+  const chainMetadata = await getChainMetadata();
```
tkporter, line 24: could maybe use `getChain(chain)` (also from registry.ts) as it looks like we just use a single chain

typescript/infra/package.json
```diff
@@ -18,6 +18,8 @@
     "@hyperlane-xyz/sdk": "4.1.0",
     "@hyperlane-xyz/utils": "4.1.0",
     "@nomiclabs/hardhat-etherscan": "^3.0.3",
+    "@safe-global/protocol-kit": "^4.0.2",
+    "@safe-global/safe-core-sdk-types": "^5.0.2",
```
tkporter, line 22: is this required? doesn't seem used but unsure if protocol-kit requires its presence

tkporter (approved):

typescript/infra/package.json
Mo-Hussain: Will double check

tkporter (approved):

PR #4136, opened 2024-07-12
chore: update app contexts

tkporter:
### Description

- updating the ezETH metric app context on the relayer
- adding merkly addresses to the app contexts
- temporarily changing the gas payment enforcement for merkly ETH contracts to just require that some payment was made, this way we won't run into issues with https://github.com/hyperlane-xyz/issues/issues/1294 over the weekend. In practice it's just been the merkly eth contracts that cause this problem when they run out of liquidity
- removed some warp route gas payment non-enforcement after we have hook indexing


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -233,6 +226,32 @@ const metricAppContexts = [
     name: 'renzo_ezeth',
     matchingList: routerMatchingList(renzoEzEthAddresses),
   },
+  {
+    name: 'renzo_ezeth_old',
+    // There's an old message to Base that's stuck around, we
+    // just care about this one for now.
```
paulbalaji, line 232: do we need to keep this in the config forever or just until it gets delivered? wondering if it's worth just letting the RC cover this and not commiting this

tkporter: nah I was hoping to remove it after it's processed. Just wanted it in the hyperlane context so that we can always easily see that the one message is still stuck around

paulbalaji (approved):

PR #4141, opened 2024-07-12
fix(utils): use browser compatible `deepEquals`

paulbalaji:
- fix: use browser compatible `deepEquals`
- removes use of node-specific deepEquality check
- add a few unit tests for `deepEqual` and `deepCopy`


nbayindirli (commented):
i believe we already have the functionality you need in utils:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d00f2ffc00d87ffb3e695a99278ee5dcae0e2dbf/typescript/utils/src/objects.ts#L12-L14

paulbalaji:
> i believe we already have the functionality you need in utils:
> 
> https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d00f2ffc00d87ffb3e695a99278ee5dcae0e2dbf/typescript/utils/src/objects.ts#L12-L14

this does not do what we need because then it relies on the ordering of items in the object to matter - we want objects to be semantically equal and not string equals

jmrossy:
@paulbalaji I'm gonna replace this with lodash since we've already got a bunch of transitive deps on lodash so there's no need for a new lib here

jmrossy (approved):

PR #4142, opened 2024-07-12
fix(sdk): update cosmos chain schema

paulbalaji:
- update cosmos chain schema to match latest spec https://github.com/cosmos/chain-registry/blob/c6cfe11be261b20ce5881ea9ba38d855130c22a5/chain.schema.json#L451-L460
- drive-by fixing of agg test

### DIFF
--- a/typescript/sdk/src/utils/cosmos.ts
+++ b/typescript/sdk/src/utils/cosmos.ts
@@ -274,7 +274,15 @@ export const CosmosChainSchema = z
               .object({
                 primary_color_hex: z
                   .string()
-                  .regex(new RegExp('^#[0-9a-fA-F]{6}$'))
+                  .min(1)
+                  .regex(new RegExp('^#([0-9a-fA-F]{6}|[0-9a-fA-F]{8})$'))
+                  .optional(),
+                background_color_hex: z
+                  .string()
+                  .min(1)
+                  .regex(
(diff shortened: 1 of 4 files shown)


nbayindirli (approved):

typescript/sdk/src/utils/cosmos.ts
```diff
@@ -274,7 +274,15 @@ export const CosmosChainSchema = z
               .object({
                 primary_color_hex: z
                   .string()
-                  .regex(new RegExp('^#[0-9a-fA-F]{6}$'))
+                  .min(1)
```
jmrossy, line 277: This feels like it's likely to keep breaking. Do we definitely need this in the monorepo?

jmrossy (approved):

typescript/sdk/src/utils/cosmos.ts
paulbalaji: I'll wait til post-hackathon to chat to trevor about this, as I believe he added it a while back in need of parsing some of these fields in the cosmos registry

tkporter: Maybe theres a way to import this via npm from the cosmos registry?

If not, we just need the gas price stuff really. So if we'd prefer to modify this to be a subset of the full spec, that works for me too

Issue #4144, opened 2024-07-12
feat(sdk): Bring over enroll router logic into WarpModule


ltyu:
## Problem
Currently the WarpModule is incomplete such that it does not handle enrollment, both on create(), and update(). 

This is problematic because https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4078 requires the enrollment step, which is in the deep in the [deployer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d00f2ffc00d87ffb3e695a99278ee5dcae0e2dbf/typescript/sdk/src/router/HyperlaneRouterDeployer.ts#L37)

## Solution
- Port over the enrollment logic such that it is called in 1) Module.create(), and 2) update()

PR #4145, opened 2024-07-12
feat(sdk): Implement enroll router logic into WarpModule

ltyu:
### Description
- Adds (optional) `remoteRouters` into WarpDeployConfig so update() can use it to compare with onchain config
- Adds enroll router logic into WarpModule

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4144

### Testing
Unit Tests


typescript/sdk/src/router/schemas.ts
```diff
@@ -17,8 +17,17 @@ export const ForeignDeploymentConfigSchema = z.object({
   foreignDeployment: z.string().optional(),
 });
 
+export const RemoteRouterSchema = z.record(
+  z.string(), // domain
+  z.string(), // router
```
nbayindirli, line 22: Why not name them here?

typescript/sdk/src/token/EvmERC20WarpModule.hardhat-test.ts
```diff
@@ -33,13 +33,22 @@ import { ProxyFactoryFactories } from '../deploy/contracts.js';
...
 import { ChainMap } from '../types.js';
 
 import { EvmERC20WarpModule } from './EvmERC20WarpModule.js';
 import { TokenType } from './config.js';
 import { TokenRouterConfig } from './schemas.js';
 
+const randomRemoteRouters = (n: number) => {
+  const routers: RemoteRouter = {};
+  for (let domain = 0; domain < n; domain++) {
+    routers[domain] = randomAddress();
```
nbayindirli, line 47: just sanity-checking here that `domain` is always just an integer index? What else might domain be?

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -74,7 +81,64 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+   * @param expectedConfig - The expected token router configuration.
+   * @returns A array with a single Ethereum transaction that need to be executed to enroll the routers
+   */
+  async updateRemoteRouters(
+    actualConfig: TokenRouterConfig,
+    expectedConfig: TokenRouterConfig,
+  ): Promise<AnnotatedEV5Transaction[]> {
+    const updateTransactions: AnnotatedEV5Transaction[] = [];
+
+    if (expectedConfig.remoteRouters) {
```
nbayindirli, line 107: might increase readability to short-circuit with the if rather than punch everything out a tab, like
```suggestion
    if (!expectedConfig.remoteRouters) {
        logger.trace('updateRemoteRouters(): Nothing to update: expectedConfig.remoteRouters undefined.');
        return [];
    }
    
    // remaining code block here
```

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -182,12 +246,24 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
       addresses: {
         deployedTokenRoute: deployedContracts[config.type].address,
       },
       chain,
       config,
     });
+
+    // Enroll Remote Routers
+    if (config.remoteRouters && !isObjEmpty(config.remoteRouters)) {
+      const enrollRemoteTx = await warpModule.updateRemoteRouters(
```
nbayindirli, line 259: even though it's always a single tx, this should still be named `enrollRemoteTxs`

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -182,12 +246,24 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+        await warpModule.read(),
+        config,
+      );
+
+      await multiProvider.sendTransaction(chain, enrollRemoteTx[0]); // updateRemoteRouters is always a single tx
```
nbayindirli, line 264: and if we want to get fancy to remove the comment:
```suggestion
      const onlyTxIndex = 0;
      await multiProvider.sendTransaction(chain, enrollRemoteTx[onlyTxIndex]);
```

nbayindirli (approved):
lgtm.

my comments are either clarifying or stylistic– let me know your thoughts

typescript/sdk/src/token/EvmERC20WarpModule.hardhat-test.ts
paulbalaji: domain/domainId is always an integer, it's chainId that can be either an integer or string

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -74,7 +81,64 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+   * Updates the remote routers for the Warp Route contract.
+   *
+   * @param actualConfig - The on-chain router configuration, including the remoteRouters array.
+   * @param expectedConfig - The expected token router configuration.
+   * @returns A array with a single Ethereum transaction that need to be executed to enroll the routers
+   */
+  async updateRemoteRouters(
+    actualConfig: TokenRouterConfig,
+    expectedConfig: TokenRouterConfig,
+  ): Promise<AnnotatedEV5Transaction[]> {
```
paulbalaji, line 104: the array type is useful for the abstract update() func, but since this is quite specific to the warp module and we're always going to return just one transaction, I'm not opposed to it returning `Promise<AnnotatedEV5Transaction>` instead

typescript/sdk/src/token/EvmERC20WarpModule.ts
paulbalaji: +1

paulbalaji (commented):
code mostly lgtm, just left an option that we could make updateRemoteRouters return `tx` instead of `tx[]`

typescript/sdk/src/router/schemas.ts
ltyu: What did you have in mind to name it?

I want it match the contract's `mapping(uint domain => bytes32 router)`

In this case, the structure for this is object is
```
{ 
    domain1: 'some router address as bytes'),
    domain2: 'some router address as bytes')
    domain3: 'some router address as bytes')
}
```

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: I actually prefer the array. imho, it's an additional mental load to remember that it's a single array, but i think it reduces the codesize in this case

1. If we use an object, we have to return an empty object in one of the branches (see Noah's [suggestion](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4145/files/dbdb18235322a3cb09a40aaa06a9bf6df8525240#r1680195452)), which means the calling function will have to handle the empty object. In this case, we will have an empty object in the [transactions](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dbdb18235322a3cb09a40aaa06a9bf6df8525240/typescript/sdk/src/token/EvmERC20WarpModule.ts#L86-L89)
2. If we keep it as an array, we can spread an empty array, which results in a noop.

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: updated.

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: Fixed

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: Fixed

typescript/sdk/src/router/schemas.ts
nbayindirli: something simple like
```suggestion
  const RemoteRouterDomain = z.string();
  const RemoteRouterRouter = z.string().startsWith('0x');

  export const RemoteRouterSchema = z.record(
    RemoteRouterDomain,
    RemoteRouterRouter,
  );
```

typescript/sdk/src/token/EvmERC20WarpModule.ts
nbayindirli: can we have 2 fns , `updateRemoteRouters(): tx[]` and `updateRemoteRouter: tx` where the first calls into the second, and the caller can decide whichever they'd like to call?

re codesize: not sure we should index on this outside of contract development (iff additional code == better readability+maintainability)

typescript/sdk/src/router/schemas.ts
ltyu: Partially updated. going to remove the explicit "Domain" schema type as its not really needed. Also removed the comments

typescript/sdk/src/token/EvmERC20WarpModule.ts
nambrot: Driveby, but agree with @ltyu, a single function that returns arrays is simplest. Definitely disagree that additional code is better readability (all things being equal)

ltyu: im going to keep it as is.

these are helper functions specifically for the `update()` and the interface for each of the current and future ( e.g., hook updates) will be the same, which is that they all return arrays.

typescript/sdk/src/router/schemas.ts
ltyu: after reading the discord comments. i think its fair to say that the suggested code is clearer. 

will do that instead :)

PR #4146, opened 2024-07-14
chore(sdk): add rpcUrl, chainId, and method(params) to smart provider logging

nbayindirli:
### Description

- adds rpcUrl, chainId, and method(params) to smart provider logging

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- manual


typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
paulbalaji, line 1: do we want to manually add the metadata like this or include it as an obj to the debug log lines?

e.g.
```ts
this.logger.debug({...rpcUrl, chainId, method}, 'slow response....')
```

nbayindirli: (referring to 266 below): forsure– updated to remove string interpolation

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -264,26 +264,40 @@ export class HyperlaneSmartProvider
...
         if (result.status === ProviderStatus.Success) {
           return result.value;
         } else if (result.status === ProviderStatus.Timeout) {
           this.logger.debug(
-            `Slow response from provider #${pIndex}.${
-              !isLastProvider ? ' Triggering next provider.' : ''
-            }`,
+            `Slow response from provider:`,
+            { ...providerMetadata },
+            isLastProvider ? '' : 'Triggering next provider.',
```
paulbalaji, line 280: object should be the first arg, everything after the msg is interpolation values https://getpino.io/#/docs/api?id=loggerdebugmergingobject-message-interpolationvalues

paulbalaji: there's probably a good follow-up task to update all the other this.logger calls to do the same - i believe we didn't really update this after the switch to pino

nbayindirli: ugh i _really_ dislike this format. data followed by message makes very little sense to me

with that said, why do we care? the point is to be able to grep logs easily for, e.g., `Slow response from provider:` and not `Slow response from ${chain} provider ${index}:`

nbayindirli: so to further clarify– we are eliminating string interpolation hyperlane-side here, and what happens pino-side should not affect our ability to grep logs

nbayindirli: i made the requested update, and we can follow up on this as a chat offline

paulbalaji (approved):

Issue #3770, opened 2024-05-13
chore: CLI Semantics (cleanup string/console output)


nbayindirli:
As features have been added to the CLI, certain string/logging patterns have broken or no longer come across in a professional light.

This semantics issue is to capture the following work first mentioned in [this PR](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3743), along with any other console-related work for whatever the end-developer will see/experience while interacting with the CLI, and to standardize those practices across the codebase:
- [Update MergedRegistry log styling to be consistent with rest of CLI](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3743/commits/9c84695754700319d93149028190d6b75a6ff129)
- Update command-level logging to take place in context (before additional output)
- Update logging to be relevant for current command + underlining on string length

nbayindirli:
![Image](https://github.com/hyperlane-xyz/hyperlane-monorepo/assets/15343884/c94ae620-5ece-48dd-a452-1ea158c01140)

nbayindirli:
After running, `hyperlane core configure` there is double newline at EOF

nbayindirli:
Typo in "? Enter protocol fee in for protocol fee hook" after running `hyperlane core configure` > `protocolFee`

nbayindirli:
`protocolFee` and `interchainGasPaymaster` do not end in `Hook`

nbayindirli:
for `hyperlane warp config` do not list raw enum string for token type selection, and make sure every option has a description

nbayindirli:
bold **default** in `? Select default hook type merkleTreeHook`
bold **required** in `? Select required hook type protocolFee`

nbayindirli:
when running `hl core deploy`, include log like "If you do not see your network in the below list, please be sure to run `hl config create` beforehand (note: this will eventually be `hl registry configure`

nbayindirli:
When there is no `HYP_KEY` sourced, let users know at the beginning of the CLI flow

nbayindirli:
When a deployment plan is presented to a user during `hyperlane * deploy`, emit a log with the location of the core config being read

nbayindirli:
Standardize SDK/CLI logging: https://www.loggly.com/use-cases/what-is-structured-logging-and-how-to-use-it/

nbayindirli:
See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4149#discussion_r1679158055

PR #4149, opened 2024-07-15
feat(cli): add blockExplorers option on registry init

nbayindirli:
### Description

- adds blockExplorers option on registry init

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- manual


typescript/cli/src/config/chain.ts
```diff
@@ -111,6 +114,42 @@ export async function createChainConfig({
   }
 }
 
+async function addBlockExplorerConfig(metadata: ChainMetadata): Promise<void> {
+  const wantBlockExplorerConfig = await confirm({
+    default: false,
+    message: 'Do you want to add a block explorer config for this chain',
```
paulbalaji, line 120: what style are we going for in the cli, should we phrase this as "Do you want to configure a block explorer for this chain"?

typescript/cli/src/config/chain.ts
```diff
@@ -111,6 +114,42 @@ export async function createChainConfig({
...
+    });
+    const family = (await select({
+      message: 'Select the type (family) of block explorer:',
+      choices: Object.entries(ExplorerFamily).map(([_, value]) => ({ value })),
+      pageSize: 10,
+    })) as ExplorerFamily;
+    const apiKey =
+      (await input({
+        message:
+          "Optional: Enter an API key for the explorer, or 'enter' to skip. Please be sure to remove this field if you intend to add your config to the Hyperlane registry:",
```
paulbalaji, line 140: just to disambiguate "enters"
```suggestion
          "Optional: Provide an API key for the explorer, or hit 'enter' to skip. Please be sure to remove this field if you intend to add your config to the Hyperlane registry:",
```

paulbalaji (commented):
code change lgtm

paulbalaji:
don't forget changelog

typescript/cli/src/config/chain.ts
nbayindirli: i'm alright with both, this was to emulate the current format in the other config functions.

i've added you comment here to capture in the future: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3770#issuecomment-2231431779

typescript/cli/src/config/chain.ts
nbayindirli: done

paulbalaji (approved):

PR #4158, opened 2024-07-17
feat: start sending resource request metrics to grafana; bump scraper CPU requests

tkporter:
### Description

- Starts sending `kube_pod_container_resource_requests` to grafana
- Useful for alerts that depend on the requests for a pod
- Deployed prom servers on testnet4 and mainnet3 with this config change
- Found that the scraper CPU requests are too low - bumped to 2 CPU - seemingly upon startup there's a pretty big CPU spike, so adjusted the requests to be within this spike.
- 
Screen Shot 2024-07-17 at 12 36 29 PM

### DIFF
--- a/typescript/infra/src/infrastructure/monitoring/prometheus.ts
+++ b/typescript/infra/src/infrastructure/monitoring/prometheus.ts
@@ -78,7 +78,7 @@ async function getPrometheusConfig(
             {
               action: 'keep',
               regex:
-                '(container.*|optics.*|Optics.*|prometheus.*|ethereum.*|hyperlane.*|kube_pod_status_phase|kube_pod_container_status_restarts_total)',
+                '(container.*|optics.*|Optics.*|prometheus.*|ethereum.*|hyperlane.*|kube_pod_status_phase|kube_pod_container_status_restarts_total|kube_pod_container_resource_requests)',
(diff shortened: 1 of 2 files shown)


typescript/infra/src/infrastructure/monitoring/prometheus.ts
```diff
@@ -78,7 +78,7 @@ async function getPrometheusConfig(
             {
               action: 'keep',
               regex:
-                '(container.*|optics.*|Optics.*|prometheus.*|ethereum.*|hyperlane.*|kube_pod_status_phase|kube_pod_container_status_restarts_total)',
+                '(container.*|optics.*|Optics.*|prometheus.*|ethereum.*|hyperlane.*|kube_pod_status_phase|kube_pod_container_status_restarts_total|kube_pod_container_resource_requests)',
```
daniel-savu, line 81: Where can I find a list of all the available metrics here?

daniel-savu (approved):

typescript/infra/src/infrastructure/monitoring/prometheus.ts
tkporter: I port forwarded to the on-cluster prometheus server to play around, but also https://github.com/kubernetes/kube-state-metrics

PR #4159, opened 2024-07-17
fix: more conservative batching gas estimates

daniel-savu:
### Description

Estimating gas isn't very accurate for Multicall txs that allow failures. This PR now estimates gas as: `max(batch_tx_estimate, individual_tx_estimates_sum)`, where `individual_tx_estimates_sum` estimates each individual tx and adds a conservative overhead that `Multicall.aggregate3` adds per subtx.

The aggregate3 overhead is calcualted based on some observed tenderly gas profiling results as follows.
- For [4 txs](https://dashboard.tenderly.co/shared/simulation/63e85ac7-3ea9-475c-8218-a7c1dd508366/gas-usage):
    - aggregate3 gas: 805,160
    - process costs: 290,091 + 168,711 + 168,711 + 168,711 = 796,224
    - overhead: 805,160 - 796,224 = 8,936 => 2,234 per subtx
- For [2 txs](https://dashboard.tenderly.co/tx/arbitrum/0xad644e431dc53c3fc0a074a749d118ff5517346c3f28d8e2513610cc9ab5c91a/gas-usage):
    - aggregate3 gas: 201,710
    - process costs: 103,011 + 92,512 = 195,523
    - overhead: 201,710 - 195,523 = 6,187 => 3,093 per subtx
- So will assume an overhead of 3500 `aggregate3` gas per message in batch

### Drive-by changes

This slightly changes the semantics of gas overrides. If a gas limit is configured, it will only guarantee that that's the minimum limit a tx can have. Now we use the `max(estimated_gas, gas_limit_override)` as the actual gas limit.
 
### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1295

### Backward compatibility

Yes

### Testing

 I think it's ok to release to release and check the new log, to see the individual estimate vs the batch estimate and double-check which one is picked in the tx explorer

### DIFF
--- a/rust/chains/hyperlane-ethereum/src/contracts/multicall.rs
+++ b/rust/chains/hyperlane-ethereum/src/contracts/multicall.rs
@@ -1,12 +1,20 @@
 use std::sync::Arc;
 
+use derive_new::new;
 use ethers::{abi::Detokenize, providers::Middleware};
 use ethers_contract::{builders::ContractCall, Multicall, MulticallResult, MulticallVersion};
-use hyperlane_core::{utils::hex_or_base58_to_h256, HyperlaneDomain, HyperlaneProvider};
+use hyperlane_core::{utils::hex_or_base58_to_h256, HyperlaneDomain, HyperlaneProvider, U256};
+use tracing::warn;
 
-use crate::{ConnectionConf, EthereumProvider};
+use crate::{tx::apply_gas_estimate_buffer, ConnectionConf, EthereumProvider};
 
 const ALLOW_BATCH_FAILURES: bool = true;
+
+/// Conservative estimate picked by withdrawing the gas used by individual calls from the total cost of `aggregate3`
+/// based on:
+/// - https://dashboard.tenderly.co/shared/simulation/63e85ac7-3ea9-475c-8218-a7c1dd508366/gas-usage
+/// - https://dashboard.tenderly.co/tx/arbitrum/0xad644e431dc53c3fc0a074a749d118ff5517346c3f28d8e2513610cc9ab5c91a/gas-usage
+const MULTICALL_OVERHEAD_PER_CALL: u64 = 3500;
 
 pub async fn build_multicall<M: Middleware + 'static>(
     provider: Arc<M>,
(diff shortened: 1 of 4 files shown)


rust/chains/hyperlane-ethereum/src/contracts/multicall.rs
```diff
@@ -34,16 +42,45 @@ pub async fn build_multicall<M: Middleware + 'static>(
...
+{
     // clear any calls that were in the multicall beforehand
     multicall.clear_calls();
 
-    calls.into_iter().for_each(|call| {
+    let mut individual_estimates_sum = Some(U256::zero());
+    let overhead_per_call: U256 = MULTICALL_OVERHEAD_PER_CALL.into();
+
+    for call in calls.into_iter() {
+        match call.estimate_gas().await {
```
tkporter, line 66: do we need to always estimate gas again here? I was hoping we'd be able to tap into the gas estimate we already have for a message from the prepare step

I think we should already have it in the ContractCall from this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4fa75a4bd825372ac51fb75ebc66151fe6cdd304/rust/chains/hyperlane-ethereum/src/contracts/mailbox.rs#L418

Maybe we could save on RPCs easily by using `call.tx.gas()` if it's Some, but otherwise falling back to estimating?

tkporter: Also am realizing that when we call `process_contract_call` in `process_batch`, it calls `add_gas_overrides` with the provided Some tx_gas_estimate. This would previously not result in an estimation if tx_gas_estimate was Some, but now that we do the max of the tx_gas_estimate and a new estimate_gas, this isn't the case. So this comment is stale https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4fa75a4bd825372ac51fb75ebc66151fe6cdd304/rust/chains/hyperlane-ethereum/src/contracts/mailbox.rs#L300, and I think we end up calling estimateGas twice per message when batching, and once now when submitting a non-batch. I think it'd probably be nice to cut down on extra estimateGas calls if we don't need them

rust/chains/hyperlane-ethereum/src/contracts/multicall.rs
```diff
@@ -34,16 +42,45 @@ pub async fn build_multicall<M: Middleware + 'static>(
     Ok(multicall)
 }
 
-pub fn batch<M: Middleware, D: Detokenize>(
+#[derive(new)]
+pub struct MulticallContractCall<M: Middleware, D: Detokenize> {
+    pub call: ContractCall<M, D>,
```
tkporter, line 47: instead of the wrapper type I think it's possible to set `call.gas(gas_limit)`

rust/chains/hyperlane-ethereum/src/contracts/multicall.rs
```diff
@@ -34,16 +41,40 @@ pub async fn build_multicall<M: Middleware + 'static>(
...
+    let overhead_per_call: U256 = MULTICALL_OVERHEAD_PER_CALL.into();
+
     calls.into_iter().for_each(|call| {
+        individual_estimates_sum = match call.tx.gas() {
+            Some(gas_estimate) => {
+                let gas_estimate: U256 = gas_estimate.into();
+                individual_estimates_sum.map(|sum| sum + gas_estimate + overhead_per_call)
+            }
+            None => {
+                warn!(call=?call.tx, "Failed to estimate gas for call in batch");
```
tkporter, line 65: curious if you think it's worth just estimating here if this happens?
I'm happy either way. I think the log message / comment below are stale -- maybe something like "unknown gas limit for batched call, falling back to estimating gas for entire batch"

rust/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -92,14 +96,18 @@ where
     M: Middleware + 'static,
     D: Detokenize,
 {
+    // either use the pre-estimated gas limit or estimate it
+    let estimated_gas_limit: U256 = match tx.tx.gas() {
```
tkporter, line 100: We don't set the ContractCall's gas prior to `process_contract_call` calling this function

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f800297259a0932cfbce9e988648efc9254848fd/rust/chains/hyperlane-ethereum/src/contracts/mailbox.rs#L301-L312

So we end up hitting the None branch and estimating gas again for each process call in a batch or a non-batched message when it's submitted, and only hit the Some branch for a batch tx. Is this intentional?

I guess not the end of the world but still feels a bit wasteful / confusing. Maybe `process_contract_call` can set the gas if `tx_gas_estimate` is Some? This used to be the intention of `add_gas_overrides` but now that the behavior has changed a bit, this setting it in `process_contract_call` may make sense

daniel-savu: Ended up doing the batch tx gas estimation in the `batch` constructor function, seems cleaner. So now I'm doing both the estimate for the batch and the summing when building the batch `ContractCall`, and I no longer need to abuse the semantics of gas overrides. 

I did keep the logic to pick the `max` between the override and the gas estimate though, to avoid override misconfigs preventing us from submitting.

So now the submission flow of a batch will set the gas limit based on `max(override, max(individual_op_sum, batch_estimate))`, whereas the flow of a single message per tx will just be `max(override, op_estimate)`

tkporter (approved):

PR #4162, opened 2024-07-17
chore(sdk): support DefaultFallbackRoutingIsm in sdk metadata builder

nambrot:
### Description

Support DefaultFallbackRoutingIsm in sdk metadata builder

### Backward compatibility

Yes

### Testing

Manually processed https://explorer.hyperlane.xyz/message/0x7d5f8450965a97735e39007f7cb9db135852123d589d81f81c113d84e3a70ceb


typescript/sdk/src/ism/metadata/routing.ts
```diff
@@ -56,3 +57,49 @@ export class RoutingMetadataBuilder implements MetadataBuilder {
...
+      this.baseMetadataBuilder.multiProvider.getProvider(
+        context.message.parsed.destination,
+      );
+    const fallbackIsm = DefaultFallbackRoutingIsm__factory.connect(
+      context.ism.address,
+      destinationProvider,
+    );
+    const defaultIsmAddress = await fallbackIsm.module(
+      context.message.parsed.origin,
+    );
```
yorhodes, line 91: this is not necessarily default ISM right?

typescript/sdk/src/ism/metadata/routing.ts
```diff
@@ -56,3 +57,49 @@ export class RoutingMetadataBuilder implements MetadataBuilder {
...
+
+    const ismReader = new EvmIsmReader(
+      this.baseMetadataBuilder.multiProvider,
+      context.message.parsed.destination,
+    );
+    const defaultIsmConfig = await ismReader.deriveIsmConfig(defaultIsmAddress);
```
yorhodes, line 97: this is clever I guess but kinda awkward that we have the reader inside the metadata builder itself?
does the reader not support deriving the fallback tree already?

yorhodes (approved):

typescript/sdk/src/ism/metadata/routing.ts
nambrot: I can see how this is a bit confusing, but in this flow specifically it is, since its not in the domains mapping, thus implied that the fallbackIsm will return to you the default ISM

typescript/sdk/src/ism/metadata/routing.ts
nambrot: We do not currently do that

typescript/sdk/src/ism/metadata/routing.ts
yorhodes: I see, when `isRouted == false`

typescript/sdk/src/ism/metadata/routing.ts
yorhodes: this is basically "lazy" ISM derivation
this requires a lot of network round trips so will contribute significantly to latency relaying messages with a default fallback ISM

Issue #4130, opened 2024-07-11
fix(sdk): contract verification rate limiting failures


nbayindirli:

nbayindirli:
fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4168

PR #4163, opened 2024-07-17
fix(sdk): account for unparsable scanner responses during contract verification

nbayindirli:
### Description

- for certain failures, etherscan may return an html response we cannot parse with JSON.parse(). One such example is when our `sourceCode` is too large, due to including all build artifacts during contract verification.
- while the underlying issue is `Request Entity Too Large` for this failure, our verifier was crashing with:
```
   `Error verifying contract: SyntaxError: Unexpected token '<', "<html>
    <h"... is not valid JSON`
```
- this change accounts for unparsable scanner responses during contract verification, and gives our users insight into the underlying failure

### Drive-by changes

- minor log update
- also accounts for bug where we printed "Verification failed" every time we'd check if a contract had been verified or not before verifying

### Related issues

- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4130

### Backward compatibility

- yes

### Testing

- manual
example output:
```
Pending https://bscscan.com/tx/0x47d9c0fd641b8689deb6cee2cb5f5d3651eb246be0a1de59d523b655c621140b (waiting 1 blocks for confirmation)
📝 Verifying implementation at 0xfFd1c0E31d1B18b51D1cfdBafbC679A97ACfb3Fa...
{
  failure: 'Request Entity Too Large',
  status: 413,
  chain: 'bsc',
  apiUrl: 'https://api.bscscan.com/api',
  family: 'etherscan'
} Failed to parse response from explorer.
Error verifying contract: Error: Failed to parse response from explorer (https://api.bscscan.com/api, bsc): Request Entity Too Large (413)
```

### DIFF
--- a/typescript/sdk/src/deploy/verify/ContractVerifier.ts
+++ b/typescript/sdk/src/deploy/verify/ContractVerifier.ts
@@ -132,8 +132,8 @@ export class ContractVerifier {
       );
       throw new Error(
         `Failed to parse response from explorer (${apiUrl}, ${chain}): ${
-          response.statusText ?? 'UNKNOWN STATUS TEXT'
-        } (${response.status ?? 'UNKNOWN STATUS'})`,
+          response.statusText || 'UNKNOWN STATUS TEXT'
+        } (${response.status || 'UNKNOWN STATUS'})`,
       );
     }
 
--- a/typescript/sdk/src/deploy/verify/ContractVerifier.ts
+++ b/typescript/sdk/src/deploy/verify/ContractVerifier.ts
@@ -108,38 +108,58 @@ export class ContractVerifier {
         method: 'GET',
       });
     } else {
-      response = await fetch(url.toString(), {
+      const init: RequestInit = {
         method: 'POST',
         headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
         body: params,
-      });
+      };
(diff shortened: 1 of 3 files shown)


typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -108,38 +108,58 @@ export class ContractVerifier {
...
+          chain,
+          apiUrl,
+          family,
+        },
+        'Failed to parse response from explorer.',
+      );
+      throw new Error(
+        `Failed to parse response from explorer (${apiUrl}, ${chain}): ${
+          response.statusText ?? 'UNKNOWN STATUS TEXT'
+        } (${response.status ?? 'UNKNOWN STATUS'})`,
```
ltyu, line 136: probably don't need the nullish check. Maybe better to use `||`

according to the [spec](https://fetch.spec.whatwg.org/#response-class), this can return an empty byte, but never undefined.

ltyu (approved):
LGTM!

typescript/sdk/src/deploy/verify/ContractVerifier.ts
nbayindirli: nice– updated

Issue #4123, opened 2024-07-10
fix(sdk): mark contracts as proxies


paulbalaji:
## Problem

Sometimes when deploying and verifying contracts, we are able to verify both the proxy contract and implementation contract but miss the step of marking the proxy as a proxy.

## Solution

Look into why we fail to mark some contracts as proxy, and also make it more fault tolerant of explorer API issues

a good starting point to investigate: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/c5ab4425a7699f1e112f5164368f94c91bbdc7eb/typescript/sdk/src/deploy/verify/ContractVerifier.ts#L198-L204

avious00:
@nbayindirli does this also cover general verification messages https://www.notion.so/hyperlanexyz/deployed-contracts-are-not-verified-193673098724483797a54381029916fc

nbayindirli:
although not confirmed, my hunch here is that we're not explicitly setting the `expectedimplementation` field on our `verifyproxycontract` request

nbayindirli:
example output from proxy request on BSC:
```
📝 Verifying proxy at 0x170FAfC9958276a49486B27fAD3Ca7aEdBcedbc3...
🚀 url: URL {
  href: 'https://api.bscscan.com/api',
  origin: 'https://api.bscscan.com',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'api.bscscan.com',
  hostname: 'api.bscscan.com',
  port: '',
  pathname: '/api',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
🚀 init: {
  method: 'POST',
  headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
  body: URLSearchParams {
    'module' => 'contract',
    'action' => 'verifyproxycontract',
    'address' => '0x170FAfC9958276a49486B27fAD3Ca7aEdBcedbc3',
    'apikey' => 'ZZJIDRMAKBWG2H4E6E1UJN6YQR2F993ARZ' }
}
🚀 response: Response {
  size: 0,
  timeout: 0,
  [Symbol(Body internals)]: {
    body: Gunzip {
      _writeState: [Uint32Array],
      _events: [Object],
      _readableState: [ReadableState],
      _writableState: [WritableState],
      allowHalfOpen: true,
      _maxListeners: undefined,
      _eventsCount: 5,
      bytesWritten: 0,
      _handle: [Zlib],
[...]

PR #4165, opened 2024-07-18
fix(sdk): remove contract verification check; add expectedimplementation; enhance logging

nbayindirli:
### Description

- removes contract verification check
- adds `expectedimplementation` method params for `verifyproxycontract`
- enhances logging and renames variables for clarity

### Drive-by changes

- none

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4123

### Backward compatibility

- yes

### Testing

- manual

### DIFF
--- a/typescript/sdk/src/deploy/verify/ContractVerifier.ts
+++ b/typescript/sdk/src/deploy/verify/ContractVerifier.ts
@@ -198,33 +198,6 @@ export class ContractVerifier {
     return responseJson.result;
   }
 
-  // private async isAlreadyVerified(
-  //   chain: ChainName,
-  //   input: ContractVerificationInput,
-  //   verificationLogger: Logger,
-  // ): Promise<boolean> {
-  //   try {
-  //     const result = await this.submitForm(
-  //       chain,
-  //       ExplorerApiActions.GETSOURCECODE,
-  //       verificationLogger,
-  //       {
-  //         address: input.address,
-  //       },
-  //     );
-  //     return !!result[0]?.SourceCode;
-  //   } catch (error: any) {
-  //     // This specific error message is thrown when a contract is not verified.
-  //     if (!error.message.includes(ExplorerApiErrors.NOT_VERIFIED.toString())) {
-  //       verificationLogger.debug(
-  //         { error },
-  //         'Error checking if contract is already verified',
-  //       );
-  //     }
-  //     return false;
-  //   }
-  // }
-
   private async verifyProxy(
     chain: ChainName,
     input: ContractVerificationInput,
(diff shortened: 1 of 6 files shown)


typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -183,32 +191,39 @@ export class ContractVerifier {
...
-      return !!result[0]?.SourceCode;
-    } catch (error) {
-      verificationLogger.debug(
-        { error },
-        'Error checking if contract is already verified',
-      );
-      return false;
-    }
-  }
+  // private async isAlreadyVerified(
```
cmcewen, line 201: what is your goal in commenting this out? do you expect to use this again in the future?

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -256,14 +267,22 @@ export class ContractVerifier {
     chain: ChainName,
     input: ContractVerificationInput,
     verificationLogger: Logger,
-  ): Promise<{ proxyGuid: any }> {
+  ): Promise<string> {
     try {
-      return this.submitForm(
+      const proxyGuid = await this.submitForm(
```
cmcewen, line 272: nit: i find it confusing that two different sounding functions (this.markProxy and this.submitForm) return a variable with the same name (proxyGuid)

typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -358,17 +377,17 @@ export class ContractVerifier {
...
-        chain,
-        input.address,
-      );
-      verificationLogger.debug(
-        `Contract already verified at ${addressUrl}#code`,
-      );
-      await sleep(200); // 5 calls/s (https://info.etherscan.com/api-return-errors/)
-      return;
-    }
+    // if (await this.isAlreadyVerified(chain, input, verificationLogger)) {
```
cmcewen, line 380: same comment about comment 😄

typescript/sdk/src/deploy/verify/ContractVerifier.ts
nbayindirli: I expect us to reward the function in the future, but also commented code is dead code. I'll remove it and we can come back to this PR if we really need that code.

typescript/sdk/src/deploy/verify/ContractVerifier.ts
nbayindirli: `submitForm()` only returns the response for whatever the request is (in this case, `ExplorerApiActions.VERIFY_PROXY`).

In the context of `markProxy()`, this is always `proxyGuid` if the response returns a 200

typescript/sdk/src/deploy/verify/ContractVerifier.ts
nbayindirli: removed as well :)

PR #4169, opened 2024-07-18
fix(relayer): don't submit reverting messages in batches

daniel-savu:
### Description

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4159

Some messages in a Multicall would revert because previous messages would invalidate them temporarily. The IGP allowance of those messages would be deducted, and if this happened more than once, those messages would become stuck.

This PR fixes those unnecesary spends on reverting `process` txs by using `.call()` to simulate the outcome of a Multicall tx. Reverting messages are removed from the batch and submitted serially, and if the batch is only left with one non-reverting message, even that one is submitted serially (rather than through a batch of one message).

Messages that did get submitted have their igp allowance deducted and status updated in `update_sent_ops_state`.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1294
### Testing

Tested by manually setting the result of `call` to failed for all messages, to see if they are submitted serially. In prod, we can track the `Submitted transaction batch` log, which now includes any excluded messages from the batch.


rust/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -328,6 +330,70 @@ where
...
         fill_tx_gas_params(tx, self.provider.clone(), &tx_overrides).await
     }
+
+    async fn simulate_batch(
+        &self,
+        multicall: &mut Multicall<M>,
+        contract_calls: Vec<ContractCall<M, ()>>,
+    ) -> ChainResult<BatchSimulation<M>> {
+        let batch = multicall::batch::<_, ()>(multicall, contract_calls.clone()).await;
+        let batch = self.add_gas_overrides(batch, None).await?;
```
tkporter, line 340: Do we need to call this here? Because we're just calling I think it's fine to skip this, I believe these are mostly in service of prepping a call to be a full on tx. E.g. we don't need the gas price / gas limit set and could cut out the extra RPCs
Feels more appropriate in `submit_tx`

rust/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -328,6 +330,70 @@ where
...
+        let call_results = batch.call().await?;
+
+        // successful calls within the batch are guaranteed to still be successful after
```
tkporter, line 343: at the risk of being pedantic I think this is an assumption of ours and edge cases could exist 😛

rust/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -480,48 +484,84 @@ struct OperationBatch {
...
+                Self::handle_batch_result(self.operations, batch_result, confirm_queue).await
             }
             Err(e) => {
                 warn!(error=?e, batch=?self.operations, "Error when submitting batch. Falling back to serial submission.");
+                self.operations
             }
+        };
+
+        if !excluded_ops.is_empty() {
+            debug!(excluded_ops=?excluded_ops, "Operations would revert in batch. Submitting individually.");
```
tkporter, line 498: nit: seems like the Err case above will result in two logs:
```
Error when submitting batch. Falling back to serial submission.
Operations would revert in batch. Submitting individually.
```
Do we want the second log if the first one's printed? could be a little confusing

rust/hyperlane-core/src/traits/mailbox.rs
```diff
@@ -60,3 +71,23 @@ pub trait Mailbox: HyperlaneContract + Send + Sync + Debug {
...
     fn process_calldata(&self, message: &HyperlaneMessage, metadata: &[u8]) -> Vec<u8>;
 }
+
+/// The result of processing a batch of messages
+#[derive(new, Debug)]
+pub struct BatchResult {
+    /// The outcome of executing the batch, if one was sent
+    pub outcome: Option<TxOutcome>,
+    /// Indexes of excluded calls from the batch (i.e. that were not executed)
+    pub excluded_call_indexes: Vec<usize>,
```
tkporter, line 81: nit: wonder if this should be `failed_indexes` bc the exclusion is based on how BatchResult is used and BatchResult isn't really aware of this

tkporter (approved):
lgtm!
I do think that the submission logic is starting to get a bit confusing as we add more logic / special cases. I just mention this to keep in mind for future changes - this is good to merge imo

daniel-savu:
> I do think that the submission logic is starting to get a bit confusing as we add more logic / special cases. I just mention this to keep in mind for future changes - this is good to merge imo

@tkporter I agree, there's some tech debt that was added when we shipped batching, and the extra logic for fixing edge cases doesn't help. I see it being hard to onboard or contribute to this part of the codebase because of that, but agree we probably have higher priority things to do for now

rust/agents/relayer/src/msg/op_submitter.rs
tkporter: sgtm!

PR #4178, opened 2024-07-23
feat: remove outbox param from `ArbL2ToL1Ism`

aroralanuk:
### Description

- outbox contract can be derived from arbitrum's bridge contract so it's not useful to pass in

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Yes

### Testing

Unit

### DIFF
--- a/solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
+++ b/solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
@@ -13,7 +13,7 @@ pragma solidity >=0.8.0;
  @@@@@@@@@       @@@@@@@@@
 @@@@@@@@@       @@@@@@@@*/
 
-// ============ Internal Imports ============
+// ============ Internal Imports =============
 
 import {IInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
 import {TypeCasts} from "../../libs/TypeCasts.sol";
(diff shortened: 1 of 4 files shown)


solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
```diff
@@ -45,15 +46,12 @@ contract ArbL2ToL1Ism is
...
-        address _bridge,
-        address _outbox
-    ) CrossChainEnabledArbitrumL1(_bridge) {
+    constructor(address _bridge) CrossChainEnabledArbitrumL1(_bridge) {
         require(
             Address.isContract(_bridge),
             "ArbL2ToL1Ism: invalid Arbitrum Bridge"
         );
-        arbOutbox = IOutbox(_outbox);
+        arbOutbox = IOutbox(IBridge(_bridge).activeOutbox());
```
yorhodes, line 54: can this change?

yorhodes (approved):

solidity/contracts/isms/hook/ArbL2ToL1Ism.sol
aroralanuk: no, it's in the initializer here: https://github.com/OffchainLabs/nitro-contracts/blob/69399a22fdc3102120d5ac0737d0830580218aeb/src/bridge/Bridge.sol#L25

PR #4179, opened 2024-07-23
feat: new validators

Mo-Hussain:
### Description

- Add new validators
- Add new safes
- make mapViolationsToCalls run in parallel

### Drive-by changes

- improve `create-safe.ts` script to fetch and print the transaction service url with instruction to add it to registry


typescript/infra/scripts/safes/create-safe.ts
```diff
@@ -54,10 +54,18 @@ async function main() {
...
   console.log(`Safe address: ${safeAddress}`);
-  console.log(
-    `Safe url: ${safeTxServiceUrl}/home?safe=${chain}:${safeAddress}`,
-  );
+  console.log(`Safe url: ${safeHomeUrl}/home?safe=${chain}:${safeAddress}`);
   console.log('url may not be correct, please check by following the link');
+
+  try {
+    const response = await fetch(`${safeHomeUrl}/home`);
+    const resultsJson = await response.json();
```
tkporter, line 62: does this work?
`https://safe.mainnet.frax.com/home` doesn't give me JSON

Annoyingly I feel like it changes with each deployment, e.g. for frax the endpoint is `https://safe.mainnet.frax.com/cgw/v1/chains`, for the official gnosis safe it's `https://safe-client.safe.global/v1/chains` :(

typescript/infra/scripts/safes/create-safe.ts
```diff
@@ -54,10 +54,18 @@ async function main() {
...
+    console.log(
+      `Add the transaction service url ${resultsJson.results.transactionService} as gnosisSafeTransactionServiceUrl to the metadata.yml in the registry`,
```
tkporter, line 64: I think `results` is an array and we'd need to filter for the specific chain for UIs that support multiple chains, like `https://safe-client.safe.global/v1/chains`

tbh if this feels like a big hassle to make it accurate, I think it's fine to just leave the programmatic transaction service finding to a later time

tkporter (approved):

typescript/infra/scripts/safes/create-safe.ts
Mo-Hussain: Hmm yes it can vary per deployment

PR #4180, opened 2024-07-23
feat: Relayer heap profiling

daniel-savu:
### Description

During the scalability audit we've identified relayer memory usage as a cause of concern. This PR runs the profiler each time e2e is run and outputs a report called `dhat-heap.json` in the current dir. That file can be visualised with a dhat viewer, such as the one hosted by the author [here](https://nnethercote.github.io/dh_view/dh_view.html).

It adds a conditionally compiled heap memory profiler ([dhat-rs](https://docs.rs/dhat/latest/dhat/)), which is only included when both the `memory-profiling` feature is enabled and the `memory-profiling` cargo profile is used.

The report can order the profiling results by total lifetime allocated bytes, peak allocated bytes, total allocation count, etc. Below is the result of sorting by total lifetime allocated bytes, which reveals that the biggest memory hogs are the hook indexing channel [senders](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-base/src/contract_sync/mod.rs#L52) (one per origin chain), taking up 100+ MB. Its size can most likely be reduced from 1M to 1k.

Screenshot 2024-07-23 at 14 32 21

### Drive-by changes

Adds the remaining IDs in the channel as a field to the hook indexing log, to know if it's safe to lower the channel size.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1293

### Backward compatibility

Yes, since everything is conditionally compiled / isolated in a new cargo profile.

### Testing

Manual

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -5,3 +5,4 @@ kathydb
 hyperlane_db
 config/test_config.json
 validator_db_anvil*
+dhat-heap*
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2129,6 +2129,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "dhat"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "98cd11d84628e233de0ce467de10b8633f4ddaecafadefc86e13b84b8739b827"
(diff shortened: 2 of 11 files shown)


tkporter (approved):
nice 🔥

PR #4190, opened 2024-07-23
feat: use multiple `mpsc` channels instead of `broadcast` for async interface

daniel-savu:
### Description

Follow up to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4180

Lowers txid broadcast channel capacity from 1M to 30k and refactors it to use a vec of `mpsc` under the hood.

### Context
The heap profiler showed that each new chain takes up 100MB of memory because of the txid broadcast channel. In addition to this, the tokio broadcast channel is a ring buffer where new items simply overwrite old items, making it hard to safely lower the capacity. 

As such, to be able to safely lower the capacity of this channel, usage of `tokio`'s `broadcast` channel is replaced with a `Vec` of `mpsc`s, which offer an async `send` interface for backpressure. So instead of overwriting old items, senders will block on sends, which makes sense in our case as we absolutely don't want to miss items. The submitter is slower than the indexer based on empirical observation, so blocking on channel size doesn't bottleneck throughput.

I noticed the max number of channel items we can reach is 29k, by running the RC relayer after an 8 day break, during which around 160k messages were submitted. Lowering the capacity to 30k should cause us to only encounter backpressure in very rare cases, and the footprint of this channel would lower from 100MB to 3MB.

### Drive-by changes

The delivery indexer in the scraper is no longer passed a txid receiver, since message deliveries are not expected to occur in the same tx as message dispatches.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1293

### Testing

E2E, by adding invariants to make sure txid indexing works

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -173,7 +173,7 @@ tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"] }
 thiserror = "1.0"
 time = "0.3"
 tiny-keccak = "2.0.2"
-tokio = { version = "1", features = ["parking_lot", "tracing"] }
+tokio = { version = "1.37", features = ["parking_lot", "tracing"] }
 tokio-metrics = { version = "0.3.1", default-features = false }
 tokio-test = "0.4"
 toml_edit = "0.19.14"
--- a/rust/hyperlane-base/src/contract_sync/broadcast.rs
+++ b/rust/hyperlane-base/src/contract_sync/broadcast.rs
@@ -1,0 +1,32 @@
+use derive_new::new;
+use eyre::Result;
+use hyperlane_core::H512;
+use tokio::sync::mpsc::{Receiver as MpscReceiver, Sender as MpscSender};
+
+#[derive(Debug, Clone, new)]
+/// Wrapper around a vec of mpsc senders that broadcasts messages to all of them.
+/// This is a workaround to get an async interface for `send`, so senders are blocked if any of the receiving channels is full,
+/// rather than overwriting old messages (as the `broadcast` channel ring buffer implementation does).
+pub struct BroadcastMpscSender<T> {
+    capacity: usize,
+    #[new(default)]
+    sender: Vec<MpscSender<T>>,
+}
+
+impl BroadcastMpscSender<H512> {
+    /// Send a message to all the receiving channels.
+    // This will block if at least of the receiving channels is full
+    pub async fn send(&self, txid: H512) -> Result<()> {
+        for sender in &self.sender {
+            sender.send(txid).await?
+        }
+        Ok(())
+    }
+
+    /// Get a receiver channel that will receive messages broadcasted by all the senders
+    pub fn get_receiver(&mut self) -> MpscReceiver<H512> {
+        let (sender, receiver) = tokio::sync::mpsc::channel(self.capacity);
+        self.sender.push(sender);
+        receiver
+    }
+}
--- a/rust/hyperlane-base/src/contract_sync/broadcast.rs
+++ b/rust/hyperlane-base/src/contract_sync/broadcast.rs
@@ -34,7 +34,7 @@ impl BroadcastMpscSender<H512> {
     }
 
     /// Get a receiver channel that will receive messages broadcasted by all the senders
-    pub async fn get_receiver(&mut self) -> MpscReceiver<H512> {
+    pub async fn get_receiver(&self) -> MpscReceiver<H512> {
         let (sender, receiver) = tokio::sync::mpsc::channel(self.capacity);
 
         self.sender.lock().await.push(sender);
@@ -42,7 +42,7 @@ impl BroadcastMpscSender<H512> {
     }
 
     /// Utility function map an option of `BroadcastMpscSender` to an option of `MpscReceiver`
-    pub async fn map_get_receiver(maybe_self: Option<&mut Self>) -> Option<MpscReceiver<H512>> {
+    pub async fn map_get_receiver(maybe_self: Option<&Self>) -> Option<MpscReceiver<H512>> {
         if let Some(s) = maybe_self {
             Some(s.get_receiver().await)
         } else {
--- a/rust/hyperlane-base/src/contract_sync/broadcast.rs
+++ b/rust/hyperlane-base/src/contract_sync/broadcast.rs
@@ -24,7 +24,7 @@ pub struct BroadcastMpscSender<T> {
 
 impl BroadcastMpscSender<H512> {
     /// Send a message to all the receiving channels.
-    // This will block if at least of the receiving channels is full
+    // This will block if at least one of the receiving channels is full
     pub async fn send(&self, txid: H512) -> Result<()> {
         let senders = self.sender.lock().await;
         for sender in &*senders {
--- a/rust/hyperlane-base/src/contract_sync/broadcast.rs
+++ b/rust/hyperlane-base/src/contract_sync/broadcast.rs
@@ -1,7 +1,12 @@
+use std::sync::Arc;
+
 use derive_new::new;
 use eyre::Result;
 use hyperlane_core::H512;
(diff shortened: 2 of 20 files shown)


rust/Cargo.toml
```diff
@@ -173,7 +173,7 @@ tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"] }
 thiserror = "1.0"
 time = "0.3"
 tiny-keccak = "2.0.2"
-tokio = { version = "1", features = ["parking_lot", "tracing"] }
```
daniel-savu, line 176: needed to bump tokio's version to get `recv.len()` for the mpsc receiver

rust/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -175,11 +177,11 @@ where
                 );
 
                 if let Some(tx) = self.broadcast_sender.as_ref() {
-                    logs.iter().for_each(|(_, meta)| {
-                        if let Err(err) = tx.send(meta.transaction_id) {
+                    for (_, meta) in &logs {
```
daniel-savu, line 180: this could be parallelized but given how fast the channels get populated given they can reach 30k items, I think the receivers are the bottleneck here

rust/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -13,8 +13,8 @@ pub enum CursorType {
     RateLimited,
 }
 
-// H512 * 1M = 64MB per origin chain
-const TX_ID_CHANNEL_CAPACITY: Option<usize> = Some(1_000_000);
+// H512 * 30k =~ 2MB per origin chain
+const TX_ID_CHANNEL_CAPACITY: Option<usize> = Some(30_000);
```
daniel-savu, line 17: Setting this to 30k since that's the max we've seen empirically. This will get rounded up to `32,768` - overall a 97% reduction in memory usage

rust/Cargo.toml
```diff
@@ -173,7 +173,7 @@ tendermint-rpc = { version = "0.32.0", features = ["http-client", "tokio"] }
...
+tokio = { version = "1.37", features = ["parking_lot", "tracing"] }
```
daniel-savu, line 176: needed to bump tokio's version to get recv.len() for the mpsc receiver

rust/hyperlane-base/src/contract_sync/broadcast.rs
```diff
@@ -0,0 +1,52 @@
...
+use hyperlane_core::H512;
+use tokio::sync::{
+    mpsc::{Receiver as MpscReceiver, Sender as MpscSender},
+    Mutex,
+};
+
+#[derive(Debug, Clone, new)]
+/// Wrapper around a vec of mpsc senders that broadcasts messages to all of them.
+/// This is a workaround to get an async interface for `send`, so senders are blocked if any of the receiving channels is full,
+/// rather than overwriting old messages (as the `broadcast` channel ring buffer implementation does).
```
yorhodes, line 14: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.blocking_send is this relevant?

rust/hyperlane-base/src/contract_sync/broadcast.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Otherwise it would be possible to call `get_receiver` and create new receiver-sender pairs, whose sender is later dropped
+    /// because the other `BroadcastMpscSender`s have no reference to it. The receiver would then point to a closed
+    /// channel. So all instances of `BroadcastMpscSender` have to point to the entire set of senders.
+    #[new(default)]
+    sender: Arc<Mutex<Vec<MpscSender<T>>>>,
+}
+
+impl BroadcastMpscSender<H512> {
+    /// Send a message to all the receiving channels.
+    // This will block if at least of the receiving channels is full
```
tkporter, line 27:
```suggestion
    // This will block if at least one of the receiving channels is full
```

rust/hyperlane-base/src/contract_sync/broadcast.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub async fn send(&self, txid: H512) -> Result<()> {
+        let senders = self.sender.lock().await;
```
tkporter, line 29: noting that if send blocks that means the lock is still held. Given the way `get_receiver` is used this is unlikely to be a deadlock issue. Just flagging in case you see anything that could potentially go wrong

rust/utils/run-locally/src/invariants.rs
```diff
@@ -65,12 +65,16 @@ pub fn termination_invariants_met(
     const STORING_NEW_MESSAGE_LOG_MESSAGE: &str = "Storing new message in db";
     const LOOKING_FOR_EVENTS_LOG_MESSAGE: &str = "Looking for events in index range";
     const HYPER_INCOMING_BODY_LOG_MESSAGE: &str = "incoming body completed";
+
+    const TX_ID_INDEXING_LOG_MESSAGE: &str = "Found log(s) for tx id";
```
tkporter, line 69: curious if you think we should use a shared const in the log site too. Don't feel strongly

rust/utils/run-locally/src/invariants.rs
```diff
@@ -93,6 +97,19 @@ pub fn termination_invariants_met(
         log_counts.get(LOOKING_FOR_EVENTS_LOG_MESSAGE).unwrap() > &0,
         "Didn't find any logs about looking for events in index range"
     );
+    let total_tx_id_log_count = log_counts.get(TX_ID_INDEXING_LOG_MESSAGE).unwrap();
+    assert!(
+        // there are 3 txid-indexed events:
+        // - relayer: merkle insertion and gas payment
+        // - scraper: gas payment
```
tkporter, line 104: `total_tx_id_log_count` just looks at the relayer right?

if we expect a merkle insertion and at least one gas payment, should we make this `*total_tx_id_log_count as u64 >= config.kathy_messages * 2` ?

tkporter (approved):

rust/utils/run-locally/src/invariants.rs
daniel-savu: you're right that it does only use relayer logs, but some logs cover multiple events when batches are sent - e.g. `Found log(s) for tx id num_logs = 2`. So there's a chance there are fewer `total_tx_id_log_count` than total messages, if `batch_count * 2 < message_count`

rust/utils/run-locally/src/invariants.rs
daniel-savu: I think this log would live in `hyperlane-base` and somehow that might create a circular dependency

tkporter: I don't think hyperlane-base depends on run-locally, but not a big deal

tkporter: or is it that other crates would then need to import from hyperlane-base

rust/utils/run-locally/src/invariants.rs
tkporter: ah I see

I guess this could this be flaky then / it's just a soft check that things work?

rust/hyperlane-base/src/contract_sync/broadcast.rs
daniel-savu: excellent spot, probably not an issue indeed but fixed it

rust/utils/run-locally/src/invariants.rs
daniel-savu: could be flaky and is more of a soft check, if it fail randomly I'll just change it to be nonzero

rust/utils/run-locally/src/invariants.rs
daniel-savu: turned out it works to just define in it `hyperlane-base`, don't know why I remember it not working

Issue #4191, opened 2024-07-23
fix(cli): Remove large configs in core read commands


ltyu:
## Problem
Currently, any read operation (e.g., `yarn hyperlane core read --mailbox 0x2f9DB5616fa3fAd1aB06cB2C906830BA63d135e3  --chain fraxtal`) will print the results on screen, and/or write to a config file (e.g., core-config.yaml). When files are too big, and the terminal is too small, then the results get truncated. 

This is problematic because users 1) cannot see the results, and 2) do not even see that the results are written to the config because the message gets cleared.

## Solution
Always hiding large results, and write to a config file

PR #4192, opened 2024-07-23
fix(cli): hide large configs in core read command

ltyu:
### Description
This PR adds a helper function that logs the config terminal "preview", _only_ if the number of lines is < 250 for `hyperlane core read`.

### Related issues
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4191

Yes

### Testing
Manual
- `yarn build && yarn hyperlane core read --mailbox 0x2f9DB5616fa3fAd1aB06cB2C906830BA63d135e3  --chain fraxtal `
- And tried with a deployed one to test out lines < 250

### DIFF
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -1,4 +1,3 @@
-import { stringify as yamlStringify } from 'yaml';
 import { CommandModule } from 'yargs';
 
 import { EvmCoreReader } from '@hyperlane-xyz/sdk';
@@ -12,7 +11,7 @@ import { runCoreDeploy } from '../deploy/core.js';
 import { evaluateIfDryRunFailure } from '../deploy/dry-run.js';
 import { errorRed, log, logGray, logGreen } from '../logger.js';
 import {
-  indentYamlOrJson,
+  logYamlIfUnderMaxLines,
   readYamlOrJson,
   writeYamlOrJson,
 } from '../utils/files.js';
@@ -146,7 +145,7 @@ export const read: CommandModuleWithContext<{
       const coreConfig = await evmCoreReader.deriveCoreConfig(mailbox);
       writeYamlOrJson(configFilePath, coreConfig, 'yaml');
       logGreen(`✅ Core config written successfully to ${configFilePath}:\n`);
-      log(indentYamlOrJson(yamlStringify(coreConfig, null, 2), 4));
+      logYamlIfUnderMaxLines(coreConfig, 250);
     } catch (e: any) {
       errorRed(
         `❌ Failed to read core config for mailbox ${mailbox} on ${chain}:`,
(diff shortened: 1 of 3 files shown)


typescript/cli/src/commands/core.ts
```diff
@@ -146,7 +145,7 @@ export const read: CommandModuleWithContext<{
       const coreConfig = await evmCoreReader.deriveCoreConfig(mailbox);
       writeYamlOrJson(configFilePath, coreConfig, 'yaml');
       logGreen(`✅ Core config written successfully to ${configFilePath}:\n`);
-      log(indentYamlOrJson(yamlStringify(coreConfig, null, 2), 4));
+      logYamlIfUnderMaxLines(coreConfig, 250);
```
nbayindirli, line 148: store 250 as a const like `MAX_READ_LINE_OUTPUT` to be shared between core and warp

nbayindirli (changes requested):
need to apply to `warp` as well

ltyu:
> need to apply to `warp` as well

@nbayindirli Agreed. im going to do this in 4.3 release.
- It requires a little bit more thought since we need to always write to output parameter, which is optional right now. We may end up doing the same as Core, but going to leave that to discussion to decide if its desirable.
- warp configs are arguably not as big. At least they can fit on the screen

typescript/cli/src/utils/files.ts
```diff
@@ -221,3 +228,22 @@ export function indentYamlOrJson(str: string, indentLevel: number): string {
...
+ * Logs the YAML representation of an object if the number of lines is less than the specified maximum.
+ *
+ * @param obj - The object to be converted to YAML.
+ * @param maxLines - The maximum number of lines allowed for the YAML representation.
+ * @param margin - The number of spaces to use for indentation (default is 2).
+ */
+export function logYamlIfUnderMaxLines(
+  obj: any,
+  maxLines: number,
+  margin = 2,
```
nbayindirli, line 242:
```suggestion
  maxLines: number = MAX_READ_LINE_OUTPUT,
  margin: number = 2,
```

nbayindirli (approved):

typescript/cli/src/commands/core.ts
```diff
@@ -146,7 +146,7 @@ export const read: CommandModuleWithContext<{
       const coreConfig = await evmCoreReader.deriveCoreConfig(mailbox);
       writeYamlOrJson(configFilePath, coreConfig, 'yaml');
       logGreen(`✅ Core config written successfully to ${configFilePath}:\n`);
-      log(indentYamlOrJson(yamlStringify(coreConfig, null, 2), 4));
+      logYamlIfUnderMaxLines(coreConfig, MAX_READ_LINE_OUTPUT);
```
nbayindirli, line 149: with the below update, just
```suggestion
      logYamlIfUnderMaxLines(coreConfig);
```

nbayindirli (approved):

Issue #4061, opened 2024-06-26
hyperlane core read shows provider errors


ltyu:
## Problem
`hyperlane core read --mailbox 0xEf9F292fcEBC3848bF4bB92a96a04F9ECBb78E59 --chain alfajores` display provider errors. This is problematic because it can confuse users thinking they have an actual error
![Screenshot 2024-06-26 at 12 03 32 PM](https://github.com/hyperlane-xyz/hyperlane-monorepo/assets/6251863/056fc304-85e0-4f8a-8acb-ad1fe16b83bf)

## Solution
- Consider silencing the provider before trying to derive Ism?

- Dry out `setSmartProviderLogLevel()` as mentioned in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4055#discussion_r1653406785

nbayindirli:
fixed by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4194

PR #4194, opened 2024-07-24
fix(sdk): silence provider logs in deriveIsmConfig

nbayindirli:
### Description

- silences provider logs in deriveIsmConfig, like deriveHookConfig
- created HyperlaneReader class for readers to extend in order to reuse `setSmartProviderLogLevel`

### Drive-by changes

- none

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4061

### Backward compatibility

- yes

### Testing

- manual
- ci

### DIFF
--- a/typescript/sdk/src/hook/EvmHookReader.ts
+++ b/typescript/sdk/src/hook/EvmHookReader.ts
@@ -147,9 +147,9 @@ export class EvmHookReader extends HyperlaneReader implements HookReader {
         this.logger.debug(`${customMessage}:\n\t${e}`);
       }
       throw new Error(`${customMessage}:\n\t${e}`);
+    } finally {
+      this.setSmartProviderLogLevel(getLogLevel()); // returns to original level defined by rootLogger
     }
-
-    this.setSmartProviderLogLevel(getLogLevel()); // returns to original level defined by rootLogger
 
     return derivedHookConfig;
   }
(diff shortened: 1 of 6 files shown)


typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -60,37 +61,57 @@ export class EvmIsmReader implements IsmReader {
...
+          break;
+        case ModuleType.NULL:
+          derivedIsmConfig = await this.deriveNullConfig(address);
+          break;
+        case ModuleType.CCIP_READ:
+          throw new Error('CCIP_READ does not have a corresponding IsmType');
+        default:
+          throw new Error(`Unknown ISM ModuleType: ${moduleType}`);
+      }
+    } catch (e: any) {
```
ltyu, line 106: I think i prefer using `finally` because the error is being rethrown. In the case that the caller decides to do something else with the error related to the provider, the logging will be silent.

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -1,4 +1,4 @@
-import { ethers, providers } from 'ethers';
+import { ethers } from 'ethers';
```
ltyu, line 1: add a changeset

typescript/sdk/src/ism/EvmIsmReader.ts
nbayindirli: good callout– will update for hook reader as well

ltyu (approved):
lgtm!

PR #4196, opened 2024-07-24
[For docker image purposes] rc relayer profiling

daniel-savu:


daniel-savu:
Closing - the image was built successfully with `larger-runner` but isn't running well in RC for some reason

PR #4202, opened 2024-07-24
fix(cli): fix collateralVault 'token' input request for yield routes

nbayindirli:
### Description

- fixes collateralVault 'token' input request for yield routes, where we were asking for a token instead of a vault address

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- yes

### Testing

- ci


typescript/cli/src/config/warp.ts
```diff
@@ -166,6 +165,18 @@ export async function createWarpRouteDeployConfig({
...
         break;
+      case TokenType.collateralVault:
+        result[chain] = {
+          mailbox,
+          type,
+          owner,
+          isNft,
+          interchainSecurityModule,
+          token: await input({
+            message: `Enter the erc-4626 vault address on chain ${chain}`,
```
ltyu, line 176: nit: we're trying to move away from saying `...on chain ${chain}` because it's kinda awkward for chains that have "chain" in them e.g., "cheesechain"

```suggestion
            message: `Enter the ERC-4626 vault address on ${chain}`,
```

ltyu (approved):

ltyu (changes requested):
Changeset

typescript/cli/src/config/warp.ts
nbayindirli: I tend include "noun ${var}" so if var is null or undefined it is still clear what should have been there

ltyu (approved):

PR #4203, opened 2024-07-25
feat: implementation of yield rebasing warp route contracts

aroralanuk:
### Description

- To reflect the rebased underlying asset, `_transferRemote` is overridden because we need to pass in the ERC4626 vault shares which on the synthetic `HypERC4626VaultYield` which change in value as the underlying vault shares accrue yield or receive a drawdown.
- The exchangeRate (i.e., assets/shares) is encoded as metadata and sent to the synthetic token to calculate the value of the synthetic token.
- On synthetic transfers, the overridden `ERC20.transfer()` function calculates the needed shares for the amount and transfers them instead.

Design decisions

- Do we need to rebase frequently? Yes, otherwise the exchangeRate on the synthetic token will be stale compared to the the actual vault claimable yield, which means a user who deposits later and right before the  infrequent rebase will get disproportionate share of the yield. You can rebase by calling `rebase()` on the collateral token which makes a transferRemote call with 0 amount.
- Why exchangeRate and not underlyingDepositedAmount on the synthetic? If a user withdraws (ie calls transferRemote on the synthetic) while a rebase is inflight, the underlyingDepositedAmount will get an inaccurate update (will be overreported and hence will derive inaccurate balance amounts on the synthetic. The exchangeRate doesn't change if the user is depositing or withdrawing.
- Why override the `transfer` on the synthetic? Otherwise if a user A sends 100 USDC worth of shares to user B which has a exchange rate of 1.05, user A would have unwittingly sent 105 #tokens (5 excess).
- What happens in a drawdown of the value? Once the user calls withdraws the the withdrawal they get out is accordingly discounted. If a user transfers on the synthetic, it will still gets the stale exchangeRate until rebase.
- What if the owner of the vault charges a fee on the yield? In that case, you need to override the totalAssets to reflect the assets - feesByOwner and the warp route implementation will stay the same. One example implementation is `MockERC4626YieldSharing` and we have adequate tests to cover the functionality.
- Rounding error? We're rounding down the exchangeRate as per the general recommendation for ERC4626 vault math.
- Why not make the synthetic token an ERC4626 itself, given that there's some overlapping functionality? Even though, the assetsToShares and assetsToShares functions are the same, the synthetic token itself doesn't have an underlying token so it doesn't confer to the 4626 interface.
- What if there are multiple synthetic tokens for a single rebasing collateral contract and what happens is the exchange rate is not in sync between the synthetics?

Things to keep in mind for 

- How do I keep the exchangeRate in sync with the underlying vault? Keep calling rebase as a cron job frequently, or make a hook call in deposit, redeem, and withdraw function to automatically resync the exchangeRate.
- 4626 math pitfalls: https://www.zellic.io/blog/exploring-erc-4626/

### Drive-by changes

- Renaming `HypERC20CollateralVaultDeposit` -> `HypERC4626OwnerYieldCollateral`

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1334

### Backward compatibility

Yes

### Testing

Unit tests


solidity/contracts/token/extensions/HypERC20Rebasing.sol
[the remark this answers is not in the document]
yorhodes: will be tricky to document how to use this

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+    function _transferTo(
+        address _recipient,
+        uint256 _amount,
+        bytes calldata
+    ) internal virtual override {
+        // withdraw with the specified amount of shares
+        vault.redeem(_amount, _recipient, address(this));
+    }
+
+    function rebase(uint32 _destinationDomain) public payable {
```
yorhodes, line 94: document that this is for updating the exchange rate when there is vault activity but no interchain transfer activity to update the virtual exchange rate

solidity/contracts/token/extensions/HypERC20Rebasing.sol
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // Override so that we update the exchange rate
+    function _transferTo(
+        address _recipient,
+        uint256 _amount,
+        bytes calldata metadata
+    ) internal virtual override {
+        super._transferTo(_recipient, _amount, metadata);
+        uint256 _exchangeRate = abi.decode(metadata, (uint256));
```
yorhodes, line 59: maybe restrict exchange rate updates to only when the origin is the collateral domain?

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+
+        emit SentTransferRemote(_destination, _recipient, _shares);
+    }
+
+    /**
+     * @dev Deposits into the vault and increment assetDeposited
+     * @param _amount amount to deposit into vault
+     */
+    function _depositIntoVault(uint256 _amount) internal returns (uint256) {
+        wrappedToken.approve(address(vault), _amount);
```
yorhodes, line 77: if we are infinite approving in the `initialize` this seems unnecessary?

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+    ) internal virtual override {
+        // withdraw with the specified amount of shares
+        vault.redeem(_amount, _recipient, address(this));
+    }
+
+    function rebase(uint32 _destinationDomain) public payable {
+        // force a rebase with an empty transfer to 0x1
+        _transferRemote(
+            _destinationDomain,
+            address(0x1).addressToBytes32(),
```
yorhodes, line 98: gas nit: `bytes32(0x1)`

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+    function initialize(
+        address _hook,
+        address _interchainSecurityModule,
+        address _owner
+    ) public override initializer {
+        wrappedToken.approve(address(vault), type(uint256).max);
+        _MailboxClient_initialize(_hook, _interchainSecurityModule, _owner);
+    }
+
+    function _transferRemote(
```
yorhodes, line 38: does this take effect on all external `transferRemote` overloads?

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+import "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
+import {TokenMessage} from "../libs/TokenMessage.sol";
+import {HypERC20Collateral} from "../HypERC20Collateral.sol";
+import {TypeCasts} from "../../libs/TypeCasts.sol";
+
+/**
+ * @title Hyperlane ERC20 Token Collateral with deposits collateral to a vault
+ * @author ltyu
+ */
+contract HypERC20RebasingCollateral is HypERC20Collateral {
```
yorhodes, line 13: how does this differ from the existing vault collateral implementation? we should rename for clarity or replace the existing one

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
```diff
@@ -0,0 +1,105 @@
...
+        bytes memory _hookMetadata,
+        address _hook
+    ) internal virtual override returns (bytes32 messageId) {
+        // Can't override _transferFromSender only because we need to pass shares in the token message
+        _transferFromSender(_amount);
+        uint256 _shares = _depositIntoVault(_amount);
+        uint256 _exchangeRate = PRECISION.mulDiv(
+            vault.totalAssets(),
+            vault.totalSupply()
+        );
```
yorhodes, line 52: can we somehow reuse the 4626 implementation here and elsewhere?

solidity/contracts/token/extensions/HypERC20Rebasing.sol
```diff
@@ -0,0 +1,91 @@
...
+        return sharesToAssets(_balance);
+    }
+
+    function assetsToShares(uint256 _amount) public view returns (uint256) {
+        return _amount.mulDiv(PRECISION, exchangeRate);
+    }
+
+    function sharesToAssets(uint256 _shares) public view returns (uint256) {
+        return _shares.mulDiv(exchangeRate, PRECISION);
+    }
```
yorhodes, line 90: can we reuse the 4626 implementation here somehow?

solidity/contracts/token/extensions/HypERC20Rebasing.sol
```diff
@@ -0,0 +1,91 @@
...
+        address _recipient,
+        uint256 _amount,
+        bytes calldata metadata
+    ) internal virtual override {
+        super._transferTo(_recipient, _amount, metadata);
+        uint256 _exchangeRate = abi.decode(metadata, (uint256));
+        exchangeRate = _exchangeRate;
+    }
+
+    // Override to send shares instead of assets
```
yorhodes, line 63: this seems dangerous to me
can we document in comments why we prefer virtualizing shares rather than assets on this contract?

solidity/contracts/token/extensions/HypERC20Rebasing.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    constructor(
+        uint8 _decimals,
+        address _mailbox
+    ) HypERC20(_decimals, _mailbox) {
+        _disableInitializers();
+        exchangeRate = 1e10;
+    }
+
+    // Gotta override to send shares instead of assets
```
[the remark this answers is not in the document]
aroralanuk: Yes

solidity/contracts/token/extensions/HypERC20Rebasing.sol
aroralanuk: done

solidity/contracts/token/extensions/HypERC20Rebasing.sol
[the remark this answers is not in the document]
aroralanuk: supports multi-synthetic now

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
aroralanuk: good point, I removed infinite approval.

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
aroralanuk: done

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
aroralanuk: yes

solidity/contracts/token/extensions/HypERC20RebasingCollateral.sol
aroralanuk: Even though, the assetsToShares and assetsToShares functions are the same, the synthetic token itself doesn't have an underlying token so it doesn't confer to the 4626 interface. It's a lot of surface area for not a lot of benefit.

solidity/contracts/token/extensions/HypERC20Rebasing.sol
aroralanuk: ditto as the last comment, resolving as duplicate

solidity/contracts/token/extensions/HypERC20Rebasing.sol
aroralanuk: discussed in standup, from description:

"Why exchangeRate and not underlyingDepositedAmount on the synthetic? If a user withdraws (ie calls transferRemote on the synthetic) while a rebase is inflight, the underlyingDepositedAmount will get an inaccurate update (will be overreported and hence will derive inaccurate balance amounts on the synthetic. The exchangeRate doesn't change if the user is depositing or withdrawing."

solidity/contracts/test/ERC4626/ERC4626Test.sol
yorhodes, line 1: revert this?

yorhodes (commented):
imo Yield can be omitted from all of these names?

solidity/contracts/token/extensions/HypERC4626VaultYield.sol
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    function _handle(
+        uint32 _origin,
+        bytes32 _sender,
+        bytes calldata _message
+    ) internal virtual override {
+        if (_origin == collateralDomain) {
+            uint256 _exchangeRate = abi.decode(_message.metadata(), (uint256));
+            exchangeRate = _exchangeRate;
```
yorhodes, line 70: why have an intermediate variable?

solidity/contracts/token/extensions/HypERC4626VaultYield.sol
```diff
@@ -0,0 +1,103 @@
...
+
+        emit SentTransferRemote(_destination, _recipient, _amountOrId);
+    }
+
+    function _handle(
+        uint32 _origin,
+        bytes32 _sender,
+        bytes calldata _message
+    ) internal virtual override {
+        if (_origin == collateralDomain) {
```
yorhodes, line 68: this is fine, but in general adding branches is a significant complexity increase
if the coverage tooling was working I would guess you arent covering both branches

solidity/contracts/token/extensions/HypERC4626VaultYield.sol
```diff
@@ -0,0 +1,103 @@
...
+    uint256 public constant PRECISION = 1e10;
+    uint32 public immutable collateralDomain;
+    uint256 public exchangeRate; // 1e10
+
+    constructor(
+        uint8 _decimals,
+        address _mailbox,
+        uint32 _collateralDomain
+    ) HypERC20(_decimals, _mailbox) {
+        collateralDomain = _collateralDomain;
```
yorhodes, line 29: this makes it slightly more annoying to configure

solidity/contracts/token/extensions/HypERC4626VaultYield.sol
yorhodes, line 1: should this maybe just be called `HypERC4626`?

solidity/contracts/token/extensions/HypERC4626VaultYieldCollateral.sol
yorhodes, line 1: what about `HypERC4626Collateral`

solidity/contracts/token/extensions/HypERC4626OwnerYieldCollateral.sol
yorhodes, line 1: what about `HypERC4626OwnerCollateral`

solidity/test/token/HypERC4626VaultYieldTest.t.sol
yorhodes, line 1: I dont believe you are testing the non collateral domain handle behavior

solidity/contracts/test/ERC4626/ERC4626Test.sol
aroralanuk: used here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/24cc280c363871860fe69fbdfe176ee118c94af2/solidity/test/token/HypERC4626VaultYieldTest.t.sol#L295

yorhodes: how is that related to importing Ownable?

solidity/test/token/HypERC4626VaultYieldTest.t.sol
aroralanuk: I'm doing it here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/24cc280c363871860fe69fbdfe176ee118c94af2/solidity/test/token/HypERC4626VaultYieldTest.t.sol#L316

solidity/contracts/client/Router.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity >=0.6.11;
 
+import "forge-std/console.sol";
+
```
yorhodes, line 5: revert?

solidity/contracts/test/ERC4626/ERC4626Test.sol
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity >=0.8.0;
+
 import "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/interfaces/IERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
```
yorhodes, line 7: revert?

solidity/contracts/test/ERC4626/ERC4626Test.sol
aroralanuk: I had removed Ownable from here.

solidity/contracts/token/extensions/HypERC4626Collateral.sol
```diff
@@ -0,0 +1,111 @@
...
+    function rebase(uint32 _destinationDomain) public payable {
+        // force a rebase with an empty transfer to 0x1
+        _transferRemote(
+            _destinationDomain,
+            NULL_RECIPIENT,
+            0,
+            msg.value,
+            bytes(""),
+            address(0)
+        );
```
yorhodes, line 109: does this work for the use case where we set the destination domain to be a vanilla `HypNative` or `HypCollateral` ?

yorhodes (approved):

solidity/contracts/token/extensions/HypERC4626Collateral.sol
aroralanuk: Positive. Unless the wrappedToken in `HypCollateral` has a revert on transfer for 0 amount, it's compatible. (Safe to assume this).

PR #4205, opened 2024-07-25
chore(sdk): Remove dead SDK logos

jmrossy:
### Description

Logos were moved to the registry. Existing logos in monorepo are stale and no longer needed.

### Backward compatibility

No but very unlikely anyone but us was using them


typescript/sdk/package.json
nbayindirli, line 1: ❤️

nbayindirli (approved):

PR #4209, opened 2024-07-26
chore: Upgrade CosmJS libs to 0.32.4

jmrossy:
### Description

- Upgrade CosmJS libs to 0.32.4
- Remove explorer dep from ccip-server

### Related issues

https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/issues/202

### Backward compatibility

No, major version bump required because CosmJS 0.32 is not backwards compatible with 0.31

### Testing

Tested in Warp UI

### DIFF
--- a/typescript/ccip-server/src/services/__mocks__/HyperlaneService.ts
+++ b/typescript/ccip-server/src/services/__mocks__/HyperlaneService.ts
@@ -1,7 +1,7 @@
-import { MessageTx } from 'hyperlane-explorer/src/types';
+import { MessageTx } from '../explorerTypes';
 
 class HyperlaneService {
-  async getOriginBlockByMessageId(id: string): Promise<MessageTx> {
+  async getOriginBlockByMessageId(_id: string): Promise<MessageTx> {
     return {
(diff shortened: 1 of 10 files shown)


typescript/ccip-server/src/services/__mocks__/HyperlaneService.ts
```diff
@@ -1,7 +1,7 @@
-import { MessageTx } from 'hyperlane-explorer/src/types';
+import { MessageTx } from '../explorerTypes';
 
 class HyperlaneService {
-  async getOriginBlockByMessageId(id: string): Promise<MessageTx> {
+  async getOriginBlockByMessageId(_id: string): Promise<MessageTx> {
```
nbayindirli, line 4: assuming `id` is a field on `MessageTx`, can we make this `messageId: string` instead of `_id`?

nbayindirli: nit^

nbayindirli (approved):

PR #4219, opened 2024-07-30
feat: default to mailbox address in registry

yorhodes:
### Description

```sh
yarn hyperlane core read --chain etherereum
```
will fetch mailbox address from registry

### Drive-by changes

None

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/cli/src/commands/core.ts
+++ b/typescript/cli/src/commands/core.ts
@@ -115,8 +115,8 @@ export const init: CommandModuleWithContext<{
 
 export const read: CommandModuleWithContext<{
   chain: string;
-  mailbox: string;
   config: string;
+  mailbox?: string;
 }> = {
   command: 'read',
   describe: 'Reads onchain Core configuration for a given mailbox address',
@@ -128,7 +128,6 @@ export const read: CommandModuleWithContext<{
     mailbox: {
       type: 'string',
       description: 'Mailbox address used to derive the core config',
-      demandOption: true,
     },
     config: outputFileCommandOption(
       './configs/core-config.yaml',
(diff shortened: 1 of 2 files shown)


typescript/cli/src/commands/core.ts
```diff
@@ -137,6 +136,16 @@ export const read: CommandModuleWithContext<{
     ),
   },
   handler: async ({ context, chain, mailbox, config: configFilePath }) => {
+    if (!mailbox) {
+      const addresses = await context.registry.getChainAddresses(chain);
+      mailbox = addresses?.mailbox;
+      if (!mailbox) {
+        throw new Error(
```
jmrossy, line 143: @nbayindirli @ltyu was there a util for this pattern of 'if not given, prompt, then fail if still missing'? I thought someone made one but can't quite remember.

typescript/cli/src/commands/core.ts
```diff
@@ -137,6 +136,16 @@ export const read: CommandModuleWithContext<{
     ),
   },
   handler: async ({ context, chain, mailbox, config: configFilePath }) => {
+    if (!mailbox) {
```
jmrossy, line 139: I'm nitpicking but the structure I initially intended for these command defs was for the code under /src/commands to just dumb proxies to implementations elsewhere. My reasoning was that `yargs` is not that great and we may want to swap it out eventually. I used yargs because we were already using it elsewhere.

typescript/cli/src/commands/core.ts
yorhodes: I think I made this `detectAndConfirmOrPrompt`

typescript/cli/src/commands/core.ts
yorhodes: given that this PR didnt introduce this structure would prefer not to change it here

typescript/cli/src/commands/core.ts
yorhodes: in this particular case I dont think we want to burden the user with confirming a mailbox address? like if I do `core read --chain ethereum` I want to know the ethereum core info, not be asked to confirm the ethereum mailbox address?

nbayindirli (approved):
lgtm

avious00 (approved):
LGTM

PR #4225, opened 2024-07-31
feat(cli): add submit functionality support to warp apply

nbayindirli:
### Description

- adds submit functionality support to warp apply
- enables dynamic submission of transactions to vanilla json rpc, gnosis safe, and impersonated accounts while dry-running
- allows easy support of ICA tx submissions in the future

### Drive-by changes
- just updated `ApplyParams` to `WarpApplyParams`
- e2e tests written in ts

### Related issues

- https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4242

### Backward compatibility

- yes

### Testing
- [x] Single warp route transfer ownership from EOA to Safe
- [x] Single warp route transfer from Safe to EOA
- [x] 2 warp route transfer ownership from EOA to safe
- [x] 2 warp route transfer from Safe to EOA (sepolia and basesepolia)

Multichain enrollments through their respective Safes:
- [x] Deploy to sepolia with address to Signer
- [x] Transfer Owner to safe
- [x] Warp apply to extend a synthetic Route to base sepolia and set owner to safe
- [x] Approve safe txs to enroll each other
- [x] Send a test message

E2e Testing


typescript/cli/tests/warp.zs-test.ts
jmrossy, line 1: Why 'zs' test (instead of maybe zx-test)?

typescript/cli/tests/warp.zs-test.ts
```diff
@@ -0,0 +1,138 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { Wallet } from 'ethers';
+import { $ } from 'zx';
+
+import { WarpCoreConfig, WarpRouteDeployConfig } from '@hyperlane-xyz/sdk';
+import { Address, assert, eqAddress } from '@hyperlane-xyz/utils';
+
+import { HypNative__factory } from '../../../solidity/dist/factories/contracts/token/HypNative__factory.js';
```
jmrossy, line 8: I'd expect this to be exported from `@hyperlane-xyz/core`. Is the direct import path necessary?

typescript/cli/tests/warp.zs-test.ts
```diff
@@ -0,0 +1,138 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
```
jmrossy, line 1: nbd but you can avoid these inconsistent import paths from `ethers` by reaching into the export namespaces like `import {providers}... providers.JsonRpcProvider`

typescript/cli/tests/warp.zs-test.ts
```diff
@@ -0,0 +1,138 @@
...
+const WARP_CORE_CONFIG_PATH = `${REGISTRY_PATH}/deployments/warp_routes/ETH/anvil1-config.yaml`;
+
+const ANVIL_KEY =
+  '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80';
+
+/**
+ * Deploys the Hyperlane core contracts to the specified chain using the provided config.
+ */
+async function hyperlaneCoreDeploy(coreInputPath: string) {
+  await $`yarn workspace @hyperlane-xyz/cli run hyperlane core deploy \
```
jmrossy, line 30: In cases like this, the fn need not be async. Just return the promise that $ gives you. The only reason to explicitly unwrap the promise like this is if you need it this way for error handling

typescript/cli/tests/warp.zs-test.ts
```diff
@@ -0,0 +1,138 @@
...
+        --yes`;
+}
+
+/**
+ * Retrieves the deployed Warp address from the Warp core configuration.
+ * @param warpCorePath - The file path to the Warp core configuration.
+ * @returns The deployed Warp address.
+ */
+function getDeployedWarpAddress(warpCorePath: string) {
+  const warpCoreConfig: WarpCoreConfig = readYamlOrJson(warpCorePath);
```
jmrossy, line 57: This is already simple so maybe not worth reconsidering but if the warp outputs are being written to a filesystem registry as I'd expect, you could instead use the registry's tooling for reading/writing these kinds of things. Nbd though if this was simpler overall

package.json
```diff
@@ -15,7 +15,8 @@
     "tsx": "^4.7.1"
   },
   "dependencies": {
-    "@changesets/cli": "^2.26.2"
+    "@changesets/cli": "^2.26.2",
+    "zx": "^8.1.4"
```
jmrossy, line 19: Did you mean for this to be at the root instead of in the CLI's package.json?

typescript/cli/src/submit/submit.ts
```diff
@@ -44,7 +44,7 @@ async function getSubmitter<TProtocol extends ProtocolType>(
         ...submitterMetadata,
       });
     case TxSubmitterType.GNOSIS_SAFE:
-      return new EV5GnosisSafeTxSubmitter(multiProvider, {
+      return EV5GnosisSafeTxSubmitter.create(multiProvider, {
```
jmrossy, line 47: Why does this one use a factory-like instantiator while the others above don't?

jmrossy (commented):
Mostly looking good so far!

typescript/cli/tests/warp.zs-test.ts
ltyu: incorrectly named

typescript/cli/tests/warp-apply.zx-test.ts
```diff
@@ -0,0 +1,52 @@
...
+import { assert, eqAddress } from '@hyperlane-xyz/utils';
+
+import { getDeployedWarpAddress, updateWarpOwner } from './commands/helpers.js';
+import {
+  hyperlaneCoreDeploy,
+  hyperlaneWarpApply,
+  hyperlaneWarpDeploy,
+} from './commands/warp.js';
+
+/// To run: 1) start an anvil, 2) yarn run tsx tests/warp.zs-test.ts inside of cli/
```
ltyu, line 13: This a more of a PoC. Will clean up & integrate this more (CI, use chai/expect, etc.) for e2e testing work

typescript/cli/src/submit/submit.ts
ltyu: Need to do async calls to validate safes. iiuc, constructors can't be async

typescript/cli/package.json
```diff
@@ -44,6 +45,7 @@
     "lint": "eslint . --ext .ts",
     "prettier": "prettier --write ./src ./examples",
     "test": "mocha --config .mocharc.json",
+    "test:e2e": "./scripts/e2e-test.sh",
```
ltyu, line 48: These tests are not enabled in CI yet. Will do this: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4388

typescript/cli/src/commands/warp.ts
```diff
@@ -93,15 +102,34 @@ export const apply: CommandModuleWithWriteContext<{
       process.exit(0);
     }
     const warpDeployConfig = await readWarpRouteDeployConfig(config);
+    let chainSubmissionStrategy;
+    if (strategyUrl) {
+      chainSubmissionStrategy = readChainSubmissionStrategy(strategyUrl);
+    }
```
jmrossy, line 108:
```suggestion
    const chainSubmissionStrategy = strategyUrl ? readChainSubmissionStrategy(strategyUrl) : undefined;
```

typescript/cli/src/commands/warp.ts
```diff
@@ -93,15 +102,34 @@ export const apply: CommandModuleWithWriteContext<{
...
     process.exit(0);
   },
 };
 
+/**
+ * Retrieves a chain submission strategy from the provided filepath.
+ * @param submissionStrategyFilepath a filepath to the submission strategy file
+ * @returns a formatted submission strategy
+ */
+export function readChainSubmissionStrategy(
```
jmrossy, line 124: I still think we should keep these `command/*` files void of business logic, just the HCI stuff. So suggest moving this

typescript/cli/src/context/context.ts
```diff
@@ -134,7 +134,7 @@ export async function getDryRunContext(
  * and an override one (such as a local directory)
  * @returns a new MergedRegistry
  */
-function getRegistry(
+export function getRegistry(
```
jmrossy, line 137: I see you're using this in a test helper. I'm _slightly_ concerned someone else will see this is exported and think it's the way to get a registry object (instead of via the command context). Could you have the test get the registry via the already exported `getContext` fn? If not np

typescript/cli/src/tests/warp-apply.e2e-test.ts
```diff
@@ -0,0 +1,132 @@
+import { expect } from 'chai';
+import { Wallet } from 'ethers';
+import { beforeEach } from 'mocha';
```
jmrossy, line 3: Is this import necessary? I thought the before/after/it fns just worked

typescript/cli/src/tests/warp-apply.e2e-test.ts
```diff
@@ -0,0 +1,132 @@
...
+const CHAIN_NAME_2 = 'anvil2';
+
+const BURN_ADDRESS = '0x0000000000000000000000000000000000000001';
+const EXAMPLES_PATH = './examples';
+const CORE_CONFIG_PATH = `${EXAMPLES_PATH}/core-config.yaml`;
+const WARP_CONFIG_PATH = `${EXAMPLES_PATH}/warp-route-deployment.yaml`;
+const WARP_CORE_CONFIG_PATH_1 = `${REGISTRY_PATH}/deployments/warp_routes/ETH/anvil1-config.yaml`;
+
+describe('WarpApply e2e tests', async function () {
+  this.timeout(0); // No limit timeout since these tests can take a while
```
jmrossy, line 29: I believe you can hook this `.timeout` call to the back of the describe fn, which looks more natural IMHO

typescript/cli/src/tests/warp-apply.e2e-test.ts
```diff
@@ -0,0 +1,132 @@
...
+    );
+    const { stdout } = await updateOwner(
+      BURN_ADDRESS,
+      CHAIN_NAME_1,
+      warpConfigPath,
+      WARP_CORE_CONFIG_PATH_1,
+    );
+
+    expect(stdout).to.include(
+      'Warp config on anvil1 is the same as target. No updates needed.',
```
jmrossy, line 76: We should probably abstract error messages + codes into a const file if we're going to start testing them like this. You don't have to do that now

jmrossy (commented):
Mostly looks good, just a few things remaining

typescript/cli/src/deploy/warp.ts
```diff
@@ -475,16 +489,30 @@ export async function runWarpRouteApply(params: ApplyParams): Promise<void> {
...
+            return logGreen(
               `Warp config on ${chain} is the same as target. No updates needed.`,
             );
+
+          const transactionReceipts = await submitter.submit(...transactions);
+
+          if (transactionReceipts && transactionReceipts.length > 0) {
+            if (params.chainSubmissionStrategy)
+              return logGreen(
+                `✅ Warp config update successfully submitted to safe (${params.chainSubmissionStrategy}) on ${chain}:\n\n`,
```
yorhodes, line 502:
```suggestion
                `✅ Warp config update successfully submitted (${params.chainSubmissionStrategy}) on ${chain}:\n\n`,
```

yorhodes: feels a little odd to have this branch in here
when would we be calling `submitter.submit` without a submission strategy?

typescript/cli/src/deploy/warp.ts
```diff
@@ -597,13 +649,27 @@ async function enrollRemoteRouters(
...
+        );
+
+      const transactionReceipts = await submitter.submit(...mutatedConfigTxs);
+
+      if (transactionReceipts && transactionReceipts.length > 0) {
+        if (warpApplyParams.chainSubmissionStrategy)
+          return logGreen(
+            `✅ Router enrollment update successfully submitted to safe (${warpApplyParams.chainSubmissionStrategy}) on ${chain}:\n\n`,
+            indentYamlOrJson(yamlStringify(transactionReceipts, null, 2), 4),
+          );
```
yorhodes, line 665: ditto
also can you help me understand why we have this code in multiple places?

yorhodes (approved):
theres a lot here but mostly lgtm
thanks for improving the e2e tests though next time maybe separate PRs if possible :)

typescript/cli/src/tests/warp-apply.e2e-test.ts
ltyu: I would like each test to run with a freshly deployed warp route. One example is that the tests might burn the owner so subsequent updates may not work.

typescript/cli/src/tests/warp-apply.e2e-test.ts
ltyu: Definitely considered this! However, i'm not sure if i will keep this style of testing. still investigating here, but will keep this for now

typescript/cli/src/commands/warp.ts
ltyu: I agree. will fix

typescript/cli/src/tests/warp-apply.e2e-test.ts
jmrossy: RIght, but what I'm asking is, if you simply delete this line, does the test work the same?

typescript/cli/src/deploy/warp.ts
ltyu: you're right. we don't need this extra branch

typescript/cli/src/context/context.ts
ltyu: Fixed to use getContext

typescript/cli/src/commands/warp.ts
ltyu: Fixed

typescript/cli/src/tests/warp-apply.e2e-test.ts
ltyu: Ohhhhhhhhhh....

Yes they do :)

typescript/cli/src/commands/warp.ts
ltyu: Moved this particular fn out. The other parameter related branching above this is a bigger lift so i'll leave them for now

typescript/cli/src/deploy/warp.ts
ltyu: Removed

ltyu: Because we want to submit a set of tx for each chain. 

If we DRY this by putting all the txs into a single transactions array and then submitting it at the end, we'd have to group them by chain, then look up the strategy associated for that chain, and then send it.

ltyu: We also submit the safe txs as a batch[]. we could do them individually in a loop. although i like it DRY-er, i'll leave this for now

typescript/cli/src/tests/warp-apply.e2e-test.ts
ltyu: will fix in a separate PR

ltyu:
Merging this to fix e2e tests in a separate PR

PR #4237, opened 2024-07-31
chore(root): remove ROOT dir from global search in vscode

nbayindirli:


jmrossy (approved):
From Noah on Discord:
```
This is so when I search I don't end up in the node_modules version of the sdk etc
And end up in the /typescript/sdk version
```

PR #4240, opened 2024-08-01
feat(cli): Prompt for isTestnet when defining chain

jmrossy:
### Description

Prompt for chain testnet/mainnet during chain definition flow

Screenshot 2024-08-01 at 2 13 48 PM

### Testing

Locally


nbayindirli (approved):
thanks!

PR #4241, opened 2024-08-01
feat: transfer and call prototype

yorhodes:
### Description

Compose warp routes with ICAs to warp an asset to the destination chain and then perform an abi encoded call from the account with the warped asset


yorhodes:
docs updated https://docs.hyperlane.xyz/docs/guides/transfer-and-call

PR #4244, opened 2024-08-02
chore: agent metric log tweaks

tkporter:
### Description

- some small tweaks as I looked into a recent issue with metrics


tkporter:
yep they work now 😛

PR #4245, opened 2024-08-02
feat: fully testing ICA ownership with tooling

tkporter:
### Description

Some tweaks to get checker tooling working with ICA ownership:
- changed the `inferCallSubmissionTypes` logic to no longer mutate `this.calls`. Some weird side effects were occurring, especially where the ICA logic would try to pop and push from `this.calls` while also looping over it. This is also why I made it serial again. I ended up moving away from mutating this.calls entirely during the iteration, so I guess moving back to doing it concurrently can be done again. Happy to do this if it feels good to - the only reason I haven't is this wasn't a place where we had huge inefficiencies and it'd complicate the code a little
- Got it working where a Safe on chain A owns an ICA on chain B

These are the flows I tested:
- sepolia is the "owner chain"
- deployed a warp route between sepolia and alfajores. Changed the alfajores owner to an ICA owned by the sepolia deployer key. Tested transferring ownership back from this to the deployer.
- Did the exact same as above, but now with the ICA being owned by a Safe on sepolia
- Transferred ownership of Alfajores core contracts to and from the sepolia-deployer-owned ICA

These are some raw notes that may be of interest of what I did https://www.notion.so/hyperlanexyz/ICA-playground-996cf28aea1649a18051afb8bb82acb2

This doesn't auto-deploy ICAs. An ICA is expected to have been deployed already, and then just configured in `owners.ts`. Check-deploy figures out that it's an ICA and does all the rest. This is fine imo in the short term as I only imagine us using ICAs on two chains (inevm and viction) for the time being.


typescript/infra/src/govern/multisend.ts
```diff
@@ -17,10 +17,11 @@ export class SignerMultiSend extends MultiSend {
 
   async sendTransactions(calls: CallData[]) {
     for (const call of calls) {
-      const receipt = await this.multiProvider.sendTransaction(
-        this.chain,
-        call,
-      );
+      const estimate = await this.multiProvider.estimateGas(this.chain, call);
+      const receipt = await this.multiProvider.sendTransaction(this.chain, {
+        gasLimit: estimate.mul(11).div(10), // 10% buffer
```
paulbalaji, line 22: nice catch

typescript/sdk/src/middleware/account/InterchainAccount.ts
```diff
@@ -61,29 +61,39 @@ export class InterchainAccount extends RouterApp<InterchainAccountFactories> {
         `Origin chain (${config.origin}) metadata needed for deploying ICAs ...`,
       );
     }
-    const localRouter = this.router(this.contractsMap[chain]);
-    const routerAddress =
+    const destinationRouter = this.router(this.contractsMap[destinationChain]);
```
paulbalaji, line 64: the renaming makes things a lot clearer here

paulbalaji (commented):
mostly lgtm

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -184,37 +186,44 @@ export abstract class HyperlaneAppGovernor<
...
+        for (const call of this.calls[chain]) {
+          let inferredCall: InferredCall;
+
+          inferredCall = await this.inferCallSubmissionType(chain, call);
+          if (inferredCall.type === SubmissionType.MANUAL) {
+            inferredCall = await this.inferICAEncodedSubmissionType(
+              chain,
+              call,
+            );
+          }
```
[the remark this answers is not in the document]
tkporter: not sure how to make this an if/else tbh (apart from maybe putting this ICA inferring elsewhere, lmk if you'd prefer that), but I added a comment. Lmk if you were thinking of something that I'm not getting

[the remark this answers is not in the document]
paulbalaji: nvm it's me misreading one of the lines - comment helps as well though!

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -281,15 +318,41 @@ export abstract class HyperlaneAppGovernor<
       chain: ChainName,
       submitterAddress: Address,
     ): Promise<boolean> => {
+      // The submitter needs to have enough balance to pay for the call.
+      // Surface a warning if the submitter's balance is insufficient, as this
+      // can result in fooling the tooling into thinking thinking otherwise valid
```
Mo-Hussain, line 323: nit: thinking is repeated

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -248,31 +261,55 @@ export abstract class HyperlaneAppGovernor<
...
           to: callRemote.to,
           data: callRemote.data,
           value: callRemote.value,
           description: `${call.description} - interchain account call from ${origin} to ${chain}`,
         };
-        const subType = await this.inferCallSubmissionType(origin, encodedCall);
+        const { type: subType } = await this.inferCallSubmissionType(
+          origin,
+          encodedCall,
+          async (chain: ChainName, submitterAddress: Address) => {
```
Mo-Hussain, line 279: Does this need to be async?

Mo-Hussain (approved):

typescript/infra/src/govern/HyperlaneAppGovernor.ts
tkporter: nice catch, this was a relic of a previous iteration where it did some async stuff, ill rm

PR #4252, opened 2024-08-05
chore: migrate `test1` from `13371` to `133711`

paulbalaji:
- chore: migrate `test1` from `13371` to `133711`
- due to clash with immutablezkevm
- drive-by registry update + agent config update


rust/agents/scraper/migration/src/m20230309_000001_create_table_domain.rs
```diff
@@ -321,7 +329,7 @@ const DOMAINS: &[RawDomain] = &[
     RawDomain {
         name: "test1",
         token: "ETH",
-        domain: 13371,
+        domain: 133711,
```
daniel-savu, line 332: how about some prefix rather than a suffix so we can apply it to all test chains?

paulbalaji: works for me

daniel-savu (approved):

jmrossy:
@paulbalaji in the future lets include a changeset here since it affects an SDK export (`testChains`)

PR #4262, opened 2024-08-06
Add jq as prerequisite since yarn install won't work without it

ameten:
### Description

Adds `jq` as prerequisite in README.md since `yarn install` won't work without it.

### Testing

Checked how README.md looks in GitHub.


tkporter (approved):

README.md
```diff
@@ -26,6 +26,10 @@ To read more about interchain applications, how the protocol works, and how to i
 
 ## Working on Hyperlane
 
+### Prerequisites
```
daniel-savu, line 29: Great point to add a prerequisites step, with this I'd now move the other dependency install steps (`Foundry`, `Node`, `Workspaces`) under `Prerequisites` by changing their markdown header from `###` to `####`. 

Also I think `jq` should have its own `####` subsection with some tips for installing directly from the readme, or linking to a guide like https://jqlang.github.io/jq/download/. (Just so it matches the other dependency installation steps)

ameten: It makes sense. Let me fix it.

ameten: I moved Foundry and Node to prerequisites. Workspaces (yarn install and yarn build) looks like part of day to day development process.

I can move Workspaces to prerequisites as well, if it does not look right.

@daniel-savu

daniel-savu: good spot, you're right 🫡

daniel-savu (approved):

PR #4264, opened 2024-08-06
feat: new mainnet ICA deployment & tooling

tkporter:
### Description

- Corresponding registry PR: https://github.com/hyperlane-xyz/hyperlane-registry/pull/130
- Decided to deploy new ICA routers instead of going through the big hassle of using the existing one
- Added a new script to make it easier to get / deploy an ICA


typescript/sdk/src/router/ProxiedRouterChecker.ts
```diff
@@ -11,7 +11,7 @@ export abstract class ProxiedRouterChecker<
 > extends HyperlaneRouterChecker<Factories, App, Config> {
   async checkOwnership(chain: ChainName): Promise<void> {
     const config = this.configMap[chain];
-    let ownableOverrides = {};
+    let ownableOverrides = config.ownerOverrides;
```
paulbalaji, line 14: something for the impending future: i'm not sure if new modules use/respect ownerOverrides at all, worth adding to the cli reqs sheet

typescript/infra/scripts/get-owner-ica.ts
```diff
@@ -0,0 +1,82 @@
...
+  return withChainRequired(getEnvArgs())
+    .option('ownerChain', {
+      type: 'string',
+      description: 'Origin chain where the governing owner lives',
+      demandOption: true,
+    })
+    .option('owner', {
+      type: 'string',
+      description:
+        'Address of the governing owner. Defaults to the owner of the owner chain',
```
paulbalaji, line 21: not sure I fully follow the description, what's "owner of the owner chain"

typescript/infra/scripts/check-deploy.ts
```diff
@@ -74,7 +76,15 @@ async function check() {
     multiProvider,
   );
   const routerConfig = core.getRouterConfig(envConfig.owners);
-  const ica = InterchainAccount.fromAddressesMap(chainAddresses, multiProvider);
+  const icaChainAddresses = objFilter(
```
paulbalaji, line 79: was there a specific reason for narrowing down the chain addresses in check-deploy?

tkporter: yeah - some chains don't have ICA routers deployed on them, and the SDK would complain that we'd pass in chainAddresses in there that didn't define any interchainAccountRouters

tkporter: similar deal for the RouterConfigs, we'd pass in a bunch of router configs in there that exceeded the chains in chainAddresses, and this caused issues

typescript/sdk/src/router/ProxiedRouterChecker.ts
tkporter: great call

typescript/infra/scripts/get-owner-ica.ts
tkporter: was trying to say the owner in the owner.ts file relating to the ownerChain passed in, but this is confusing, changing this

paulbalaji (approved):

Mo-Hussain:
lgtm

PR #4266, opened 2024-08-06
feat(cli): Add existing mailbox confirmation to core deploy

ltyu:
### Description
Add check & confirm for existing mailbox to core deploy to allow users to decide if they want to deploy a new mailbox

### Drive-by changes
Add skip confirmation to core deploy

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/deploy/utils.ts
+++ b/typescript/cli/src/deploy/utils.ts
@@ -100,7 +100,7 @@ async function confirmExistingMailbox(
   const addresses = await context.registry.getChainAddresses(chain);
   if (addresses?.mailbox) {
     const isConfirmed = await confirm({
-      message: `Mailbox has already exists at ${addresses.mailbox}, registry artifacts will be overwritten`,
+      message: `Mailbox already exists at ${addresses.mailbox}. Are you sure you want to deploy a new mailbox and overwrite existing registry artifacts?`,
       default: false,
     });
 
--- a/typescript/cli/src/deploy/utils.ts
+++ b/typescript/cli/src/deploy/utils.ts
@@ -12,7 +12,7 @@ import {
 import { Address, ProtocolType } from '@hyperlane-xyz/utils';
 
 import { parseIsmConfig } from '../config/ism.js';
-import { WriteCommandContext } from '../context/types.js';
+import { CommandContext, WriteCommandContext } from '../context/types.js';
 import {
   log,
   logBlue,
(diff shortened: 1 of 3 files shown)


typescript/cli/src/deploy/utils.ts
```diff
@@ -86,12 +86,30 @@ export async function runDeployPlanStep({
...
 }
 
+async function confirmExistingMailbox(
+  context: CommandContext,
+  chain: ChainName,
+) {
+  const addresses = await context.registry.getChainAddresses(chain);
+  if (addresses?.mailbox) {
+    const isConfirmed = await confirm({
+      message: `Mailbox has already exists at ${addresses.mailbox}, registry artifacts will be overwritten`,
```
paulbalaji, line 103: something like this to better inform what's happening - probs needs some workshopping on the msg
```suggestion
      message: `Mailbox already exists at ${addresses.mailbox}. Are you sure you want to deploy a new mailbox and overwrite existing registry artifacts?`,
```

typescript/cli/src/deploy/utils.ts
```diff
@@ -86,12 +86,30 @@ export async function runDeployPlanStep({
...
+) {
+  const addresses = await context.registry.getChainAddresses(chain);
+  if (addresses?.mailbox) {
+    const isConfirmed = await confirm({
+      message: `Mailbox already exists at ${addresses.mailbox}. Are you sure you want to deploy a new mailbox and overwrite existing registry artifacts?`,
+      default: false,
+    });
+
+    if (!isConfirmed) {
+      throw Error('Deployment cancelled');
```
yorhodes, line 108: I find this slightly odd behavior, answering `no` is an error?

typescript/cli/src/deploy/utils.ts
```diff
@@ -86,12 +86,30 @@ export async function runDeployPlanStep({
   );
 
   if (skipConfirmation) return;
+  await confirmExistingMailbox(context, chain);
   const isConfirmed = await confirm({
     message: 'Is this deployment plan correct?',
   });
   if (!isConfirmed) throw new Error('Deployment cancelled');
```
yorhodes, line 93: I guess its consistent with this

typescript/cli/src/deploy/utils.ts
yorhodes: not sure if `detectAndConfirmOrPrompt` is useful here

yorhodes (approved):

typescript/cli/src/deploy/utils.ts
jmrossy: Alternatives would be to return some value here and check if in callers or just `process.exit(1)` if we def want to end execution here

PR #4268, opened 2024-08-06
feat: SOL (solana <> eclipse) warp route

daniel-savu:


rust/sealevel/client/src/core.rs
```diff
@@ -15,9 +19,33 @@ use crate::{
 use hyperlane_core::H256;
 use hyperlane_sealevel_igp::accounts::{SOL_DECIMALS, TOKEN_EXCHANGE_RATE_SCALE};
 
+pub(crate) fn adjust_gas_price_if_needed(chain_name: &str, ctx: &mut Context) {
```
tkporter, line 22: not a huge deal but because we so many txs, you could maybe put it in `Context::new`

rust/sealevel/client/src/router.rs
```diff
@@ -372,7 +378,7 @@ pub(crate) fn deploy_routers<
             existing_program_ids.as_ref(),
         );
 
-        // Add the router to the list of routers.
+        // // Add the router to the list of routers.
```
tkporter, line 381: nit: undo

rust/sealevel/client/src/warp_route.rs
```diff
@@ -255,23 +263,78 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
-                );
+                    .with_client(client)
+                    .send_with_payer();
 
                 let (mint_account, _mint_bump) =
                     Pubkey::find_program_address(hyperlane_token_mint_pda_seeds!(), &program_id);
-                // TODO: Also set Metaplex metadata?
-                init_txn.add(
+
+                // the metadata pointer should be initialized here, right before the mint initialization below
```
tkporter, line 282: nit: curious if this comment should still be here

rust/sealevel/client/src/warp_route.rs
```diff
@@ -255,23 +263,78 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+                // but this is via the CLI most likely
+                // `spl-token initialize-metadata [FLAGS] [OPTIONS] <TOKEN_MINT_ADDRESS> <TOKEN_NAME> <TOKEN_SYMBOL> <TOKEN_URI>`
+                // `spl-token update-metadata-address <TOKEN_MINT_ADDRESS> <METADATA_ADDRESS>`
+                let chain_url = client.url();
+                println!("chain_url: {:?}", chain_url);
```
tkporter, line 287: don't feel strongly if you prefer to keep them, but just calling out this println and others in this file in case you meant to rm. Some seem useful but some seem like debugging msgs

rust/sealevel/client/src/warp_route.rs
```diff
@@ -255,23 +263,78 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+
+                println!("initialized metadata pointer. Status: {status}");
+
+                let required_rent = client
+                    .get_minimum_balance_for_rent_exemption(SyntheticPlugin::MINT_ACCOUNT_SIZE)
+                    .expect("querying lamports for rent failed");
+                let mint_account_balance = client
+                    .get_balance(&mint_account)
+                    .expect("failed to get balance of the mint acc");
+                let required_balance_for_metadata_rent = required_rent - mint_account_balance;
```
tkporter, line 315: nit: maybe `required_transfer_for_metadata_rent`? could also use saturating_sub to avoid an accidental overflow / skip the tx if it's 0

Don't feel strongly if you'd rather move on

rust/sealevel/client/src/warp_route.rs
```diff
@@ -295,6 +358,57 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
         }
         .with_client(client)
         .send_with_payer();
+
+        // do the token-2022 cli initialization steps here
+        if let TokenType::Synthetic(token_metadata) = &app_config.token_type {
+            let (mint_account, _mint_bump) =
+                Pubkey::find_program_address(hyperlane_token_mint_pda_seeds!(), &program_id);
+            // the metadata initialization should happen here, but this is via the CLI most likely
```
tkporter, line 366:
```suggestion
            // the metadata initialization using the CLI
```

rust/sealevel/client/src/warp_route.rs
```diff
@@ -255,23 +263,78 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+                let status = cmd
+                    .stdout(Stdio::inherit())
+                    .stderr(Stdio::inherit())
+                    .status()
+                    .expect("Failed to run command");
+
+                println!("initialized metadata pointer. Status: {status}");
+
+                let required_rent = client
+                    .get_minimum_balance_for_rent_exemption(SyntheticPlugin::MINT_ACCOUNT_SIZE)
```
tkporter, line 310: I thought this was supposed to be the length we expect to have after the metadata initialize? like something closer to 600+ bytes

Iiuc this size is the one that the warp route initialize sets for the mint, which will already make it rent exempt with a transfer from the payer

rust/sealevel/client/src/warp_route.rs
```diff
@@ -295,6 +358,57 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+            // the metadata initialization should happen here, but this is via the CLI most likely
+            let mut cmd = Command::new(spl_token_binary_path.clone());
+            cmd.args([
+                "initialize-metadata",
+                mint_account.to_string().as_str(),
+                token_metadata.name.as_str(),
+                token_metadata.symbol.as_str(),
+                token_metadata.uri.as_deref().unwrap_or(""),
+                "-p",
+                "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb",
```
tkporter, line 375: nit: could put this into a const, it's used in a few places

rust/sealevel/client/src/warp_route.rs
```diff
@@ -539,3 +653,30 @@ pub fn parse_token_account_data(token_type: FlatTokenType, data: &mut &[u8]) {
         }
     }
 }
+
+pub fn install_spl_token_cli() {
```
tkporter, line 657: nice

tkporter (approved):
Feel free to ignore some of the recommendations as I understand the intent wasn't to make this very refined. Nice job!

rust/sealevel/client/src/core.rs
daniel-savu: I don't think we have a `chain` or `local_domain` available when we call `Context::new`

rust/sealevel/client/src/warp_route.rs
daniel-savu: this was just to aid my implementation, will remove all comments like this one

rust/sealevel/client/src/warp_route.rs
daniel-savu: you're right, this is the log from the deployment: `Funding mint account 8e7oRHQ2VVzgjncL8gX75PWigq9q3qfJMgxtA9mEwgfP with funding_amount 0 to reach total balance of 24616`. 

But fortunately the difference in rent is paid in the metadata initialization transaction, so I'll remove this logic entirely: https://explorer.eclipse.xyz/tx/VuT3asE5yr1P6g4rDtCcxUzoG7ZX6aHXpnrobNdrEik14wNKmtK9WFZB8mkVUrsJ1WDzzCG3NkBNSwW9unpgXox

PR #4269, opened 2024-08-07
Hacky bsc gas price override on v2

tkporter:

### DIFF
--- a/rust/chains/hyperlane-ethereum/src/mailbox.rs
+++ b/rust/chains/hyperlane-ethereum/src/mailbox.rs
@@ -282,6 +282,15 @@ where
                 tx = tx.gas_price(gas_price);
             }
         }
+        if self.domain.id() == 56 {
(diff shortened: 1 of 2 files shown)


rust/chains/hyperlane-ethereum/src/mailbox.rs
```diff
@@ -282,6 +282,15 @@ where
                 tx = tx.gas_price(gas_price);
             }
         }
+        if self.domain.id() == 56 {
+            if let Ok(gas_price) = std::env::var("BSC_GAS_PRICE")
```
daniel-savu, line 286: is this because we don't have gas overrides on v2?

daniel-savu (approved):

rust/chains/hyperlane-ethereum/src/mailbox.rs
tkporter: yeah :(

Issue #4227, opened 2024-07-31
`set-rpc-urls.ts` should make it easier to initialize secret RPC URLs with those from the registry


tkporter:
## Problem

When adding a chain, we've already crafted metadata with public RPC URLs. It's cumbersome to then need to run `scripts/secret-rpc-urls/set-rpc-urls.ts` for each chain and to manually copy in some of these RPC URLs.

## Solution

Not 100% sure what exactly it should look like, but some random thoughts on what this could involve:
- it could at the least suggest registry RPC URLs before prompting you for the RPC URLs you want
- would maybe be nice to let the command work with many chains - it's cumbersome to run this manually for each chain
- might even be nice to have a mode where it iterates through all the chains in the environment and spots any chains that don't yet have secret RPC URLs. It could auto-suggest registry URLs but also let you put in whatever you want

definition of success here would be lower time to adding RPC URLs to secrets for the first time

PR #4270, opened 2024-08-07
feat: add script to set rpc secrets from registry

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4227

- add `set-rpc-urls-from-registry` script to set secrets by reading in contents of registry
	- example usage: `yarn tsx scripts/secret-rpc-urls/set-rpc-urls-from-registry.ts -e mainnet3 -c base zircuit arbitrum`
- refactored `set-rpc-urls` to reuse logic across both scripts

### DIFF
--- a/typescript/infra/src/utils/rpcUrls.ts
+++ b/typescript/infra/src/utils/rpcUrls.ts
@@ -41,7 +41,7 @@ export async function setAndVerifyRpcUrls(
     await confirmSetSecrets(environment, chain, secretPayload);
     await testProvidersIfNeeded(chain, rpcUrlsArray);
     await updateSecretAndDisablePrevious(environment, chain, secretPayload);
-  } catch (error) {
+  } catch (error: any) {
     console.error(
       `Error occurred while setting RPC URLs for ${chain}:`,
       error.message,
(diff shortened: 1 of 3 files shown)


typescript/infra/src/utils/rpcUrls.ts
```diff
@@ -0,0 +1,139 @@
...
+
+import { disableGCPSecretVersion } from './gcloud.js';
+import { isEthereumProtocolChain } from './utils.js';
+
+export async function testProviders(rpcUrlsArray: string[]): Promise<boolean> {
+  let providersSucceeded = true;
+  for (const url of rpcUrlsArray) {
+    const provider = new ethers.providers.StaticJsonRpcProvider(url);
+    try {
+      const blockNumber = await provider.getBlockNumber();
```
tkporter, line 19: I think it happened to us recently where a provider would just hang - now that we'll be doing many chains in a single command, may be nice to timeout after a couple secs https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/a49f52cc9c6ab7efa092d52adfa61b724a6f9cb2/typescript/utils/src/async.ts#L17. But don't feel obligated to add now if you'd rather just see if it becomes an issue later

tkporter (approved):
I was originally thinking it'd be nice to have this part of the flow of `set-rpc-urls` if the secret doesn't yet exist, but also happy with this. Just mentioning in case it wasn't considered

typescript/infra/src/utils/rpcUrls.ts
paulbalaji: na that's a good call, haven't run into it myself but worth adding

PR #4278, opened 2024-08-07
fix(cli): Add output file option to warp & core apply

ltyu:
### Description
Add --output flag to output file in `hyperlane warp & core apply`

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -28,7 +28,11 @@ import { evaluateIfDryRunFailure } from '../deploy/dry-run.js';
 import { runWarpRouteApply, runWarpRouteDeploy } from '../deploy/warp.js';
 import { log, logGray, logGreen, logRed, logTable } from '../logger.js';
 import { sendTestTransfer } from '../send/transfer.js';
-import { indentYamlOrJson, writeYamlOrJson } from '../utils/files.js';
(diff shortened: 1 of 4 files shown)


typescript/cli/src/commands/warp.ts
```diff
@@ -79,25 +84,47 @@ export const apply: CommandModuleWithWriteContext<{
       ...warpCoreConfigCommandOption,
       demandOption: false,
     },
+    out: outputFileCommandOption(
+      undefined,
+      false,
+      'The path to output a warp route deployment config JSON or YAML file.',
+    ),
```
jmrossy, line 91: Is it confusing that the output will automatically get added to the registries but also optionally this file?

jmrossy (approved):
Approving but with one concern. I'll leave the product decisions to you, Noah, and Connor so do whatever you think is best

typescript/cli/src/commands/warp.ts
ltyu: Sorry, I should've marked this as draft as it was not ready. Still some discussion about the requirements of this

ltyu:
Closing this because the intention was to update WarpCore artifacts upon `owner` update. however, the artifacts do not have an `owner` so it is not relevant.

PR #4279, opened 2024-08-07
feat: Add long-running CLI relayer

nambrot:
### Description

Adds the CLI equivalent of #3693. Should mostly be used only for demo purposes with reliable RPCs in the registry. Includes some drive-by changes to the existing single-message self-relay.

### Drive-by changes

- Self-relay fixes

### Backward compatibility

Yes

### Testing

Manual


typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,37 @@
...
+import { CommandModuleWithContext } from '../context/types.js';
+
+import { agentTargetsCommandOption } from './options.js';
+import { MessageOptionsArgTypes } from './send.js';
+
+export const relayerCommand: CommandModuleWithContext<
+  MessageOptionsArgTypes & { chains?: string }
+> = {
+  command: 'relayer',
+  describe: 'Run a CLI relayer',
```
jmrossy, line 12:
```suggestion
  describe: 'Run a Hyperlane message self-relayer',
```

typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,37 @@
...
+  builder: {
+    chains: {
+      ...agentTargetsCommandOption,
+      demandOption: false,
+    },
```
jmrossy, line 17: demandOption isn't set on `agentTargetsCommandOption` right? So no need for this more verbose form

typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+  handler: async ({ context, chains }) => {
+    const chainsArray = chains
+      ? chains.split(',').map((_) => _.trim())
+      : undefined;
```
jmrossy, line 22: Seems like you meant to set demandOption to true then

typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,37 @@
...
+    const core = HyperlaneCore.fromAddressesMap(
+      chainAddresses,
+      context.multiProvider,
+    );
+
+    const relayer = new HyperlaneRelayer({ core });
+    relayer.start(chainsArray);
+    process.once('SIGINT', async () => {
+      relayer.stop(chainsArray);
+      console.log('Stopping relayer ...');
```
jmrossy, line 33: Use the `logger`, not console directly

typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,37 @@
...
+      : undefined;
+    const chainAddresses = await context.registry.getAddresses();
+    const core = HyperlaneCore.fromAddressesMap(
+      chainAddresses,
+      context.multiProvider,
+    );
+
+    const relayer = new HyperlaneRelayer({ core });
+    relayer.start(chainsArray);
+    process.once('SIGINT', async () => {
```
jmrossy, line 31: This method need not be async if you're not `await`ing anything

jmrossy (commented):
A nice addition, just needs some touchups

typescript/cli/src/commands/relayer.ts
nambrot: I think i misunderstood, but its fine for them not to pass it (in which case it runs the relayer everywhere)

typescript/cli/src/status/message.ts
```diff
@@ -44,22 +44,27 @@ export async function checkMessageStatus({
...
     } catch (e) {
       logRed(`Failed to infer dispatch transaction for message ${messageId}`);
+
+      dispatchTx = await input({
+        message: 'Provide dispatch transaction hash',
+      });
+      dispatchedReceipt = await context.multiProvider
+        .getProvider(origin)
+        .getTransactionReceipt(dispatchTx);
     }
```
yorhodes, line 64: I know you were fixing a bug but this doesnt make it much clearer

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -211,13 +211,25 @@ export class HyperlaneRelayer {
...
+        message,
+      );
+
+      this.logger.info(
+        `Message ${message.id} was processed in ${
+          this.multiProvider.tryGetExplorerTxUrl(destination, {
+            hash: processReceipt.transactionHash,
+          }) ?? 'tx ' + processReceipt.transactionHash
+        }`,
+      );
```
yorhodes, line 226: I would have expected the multi provider to already log this tbh

yorhodes (approved):

typescript/cli/src/status/message.ts
nambrot: Really, imo its more clear. What do you think would make it more clear?

typescript/sdk/src/core/HyperlaneRelayer.ts
nambrot: it does not

jmrossy (approved):

typescript/cli/src/status/message.ts
yorhodes: not having the receipt fetching in two places

typescript/cli/cli.ts
```diff
@@ -20,6 +20,7 @@ import {
   skipConfirmationOption,
 } from './src/commands/options.js';
```
ltyu, line 21: Add changeset

typescript/cli/src/commands/relayer.ts
```diff
@@ -0,0 +1,35 @@
...
+      ? chains.split(',').map((_) => _.trim())
+      : undefined;
+    const chainAddresses = await context.registry.getAddresses();
+    const core = HyperlaneCore.fromAddressesMap(
+      chainAddresses,
+      context.multiProvider,
+    );
+
+    const relayer = new HyperlaneRelayer({ core });
+    relayer.start(chainsArray);
```
ltyu, line 28: Would be nice to have a logger here to signify that the relayer has started. otherwise, it looks like its just hanging

ltyu (approved):
very cool

typescript/cli/src/status/message.ts
nambrot: I would say its clear, but might feel duplicative, but IMO they are different code paths. Happy to hear suggestions as to how to make this clearer

Issue #4069, opened 2024-06-27
Validator should be able to be configured to use GCP storage for checkpoints


tkporter:
## Problem

- GCP checkpoint storage is implemented and the relayer is seemingly capable of reading checkpoints
- Validator settings seem to not support this case though

## Solution

- Update the match that validators use to support using GCP storage https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/44cc9bf6b33f5288462ecc027e9d624ae682d552/rust/agents/validator/src/settings.rs#L151-L195
- Probably get rid of the Some(_) branch so this doesn't happen again
- Probably worth a run to confirm it actually works, and maybe another pass on the code

mawutory:
@tkporter can i pick this up? this would be my first.

daniel-savu:
we should also update out infra `announce-validators.ts` script to parse gcp paths and announce with that: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/bb470aec25c02086921d1305c78cc09a1cda2528/typescript/infra/scripts/announce-validators.ts#L49

but this is probably for another ticket

daniel-savu:
> @tkporter can i pick this up? this would be my first.

go for it @mawutory

mawutory:
thanks @daniel-savu here's what i believe is expected of me:

replace `Some(_)` with `Some("gcp")` having a similar structure as `Some("s3")`?

daniel-savu:
closed in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3156 and other follow ups

PR #4282, opened 2024-08-08
feat: Report git commit hash as agent metadata

ameten:
### Description

* We are storing git commit hash as an environment variable when we compile the source code of an agent.
* Validator uses git commit hash as a agent version in `metadata_latest.json` file.
* File `metadata_latest.json` is overwritten every time when validator is restarted.
* Git commit hash is extracted by `git` util, so, it should be available during build time.
* If the directory with the source code does not contain `.git` folder, git commit hash will be defaulted to `VERGEN_IDEMPOTENT_OUTPUT`.

Agent Metadata is propagated to Relayer and Scraper, but is not used there. We can use it in the future, reporting git hash commit in the logs, for example.

### Backward compatibility

Yes

### Testing

* Tested locally with E2E Tests.
* Tested against S3 by deploying docker image into k8s


rust/hyperlane-base/build.rs
```diff
@@ -0,0 +1,8 @@
+use anyhow::Result;
+use vergen::EmitBuilder;
+
+fn main() -> Result<()> {
+    EmitBuilder::builder().all_git().emit()?;
```
daniel-savu, line 5: nit: Looks like `EmitBuilder::builder().all_git().git_sha()` would suffice here, to avoid adding unnecessary env vars

rust/hyperlane-base/src/agent/metadata.rs
```diff
@@ -0,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+
+/// Metadata about agent
+#[derive(Debug, Deserialize, Serialize)]
+pub struct AgentMetadata {
```
daniel-savu, line 5: nit: we use the `derive_new` macro throughout the codebase to generate constructor boilerplate, you could derive it by adding `new` here imported from `derive_new::new`. Then you could remove the `impl AgentMetadata`

rust/hyperlane-base/src/agent/metadata.rs
daniel-savu, line 1: nit: `hyperlane-base` is consumed solely by the agents, so the `agent` naming doesn't add much. Do you think using the `agent` dir is preferable to having `metadata.rs` live directly in `hyperlane-base/src`?

rust/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -174,6 +175,15 @@ impl CheckpointSyncer for GcsStorageClient {
         Ok(())
     }
 
+    /// Write the agent metadata to this syncer
+    async fn write_metadata(&self, metadata: &AgentMetadata) -> Result<()> {
```
daniel-savu, line 179: Noticing we could reuse some logic across all `CheckpointSyncer` impls, and across all write operations (`write_checkpoint`, `write_metadata`, `write_announcement`). First the data is serialized with `serde_json`, then it's posted in a syncer-specific way. I'll create a ticket for this in case we ever decide to post additional data

daniel-savu (commented):
Great work!

- should make sure to include the e2e test. To also test the S3 integration, you can deploy a validator image to the RC context and manually check that the metadata file is posted. Also curious to check if the metadata file does get updated when the commit changes
- love that `VERGEN_SYSINFO_` allows including more debug info in the future, should we ever need to

rust/hyperlane-base/build.rs
ameten: It makes sense. 

I fixed it like this: 
```suggestion
    EmitBuilder::builder().git_sha(false).emit()?;
```

rust/hyperlane-base/src/agent/metadata.rs
ameten: Nice. Good catch!

rust/hyperlane-base/src/agent/metadata.rs
ameten: I did not know that. It makes sense. Fixed.

rust/hyperlane-base/src/types/gcs_storage.rs
ameten: Ok. I leave it as is then.

rust/hyperlane-base/src/metadata.rs
```diff
@@ -0,0 +1,10 @@
+use derive_new::new;
+use serde::{Deserialize, Serialize};
+
+/// Metadata about agent
+#[derive(Debug, Deserialize, Serialize, new)]
+pub struct AgentMetadata {
+    /// Latest git commit hash at the time when agent was built
+    /// If .git was not present at the time of build, field defaults to "VERGEN_IDEMPOTENT_OUTPUT"
```
tkporter, line 8: nit: feels like this info is most useful to know where the `git_sha` is created (i.e where `env!("VERGEN_GIT_SHA")` is called), wonder if a note about this at the `AgentMetadata::new` callsite would be useful, or maybe rolling that `env!("VERGEN_GIT_SHA")` bit into a non-derived `AgentMetadata::new()`

rust/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -7,6 +7,7 @@ use std::fmt;
 use ya_gcp::{storage::StorageClient, AuthFlow, ClientBuilder, ClientBuilderConfig};
 
 const LATEST_INDEX_KEY: &str = "gcsLatestIndexKey";
```
tkporter, line 9: not within the scope of this PR, but I didn't realize the keys for the GCS checkpoint syncer are different than s3, I'm a little worried this'll make debugging GCS stuff unnecessarily annoying. Because GCS isn't technically live (https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4069) @daniel-savu I wonder if we should also change this as part of 4069?

rust/utils/run-locally/src/invariants.rs
```diff
@@ -1,213 +1,7 @@
-use std::fs::File;
-use std::path::Path;
+pub use common::SOL_MESSAGES_EXPECTED;
```
tkporter, line 1: I like this drive-by refactor :)

tkporter (approved):
Great job, LGTM! Super excited to see a first Rust contribution like this that'll make all our lives much easier

rust/utils/run-locally/src/invariants.rs
ameten: Thank you!

rust/hyperlane-base/src/metadata.rs
ameten: I moved documentation comments around.

I would not move `env!("VERGEN_GIT_SHA")` inside the constructor of the struct instance since it reduces the flexibility. We may want to create an instance of the struct with a specific hash for testing purposes, for example.

PR #4283, opened 2024-08-08
fix(cli): Remove getUri from core read logging

ltyu:
### Description
This removes the `getUri()` from a specific log because the function throws for MergeRegistry and ends up hiding the real error message.

### Backward compatibility
Yes

### Testing
Manual


jmrossy (approved):

jmrossy:
@ltyu if you prefer, we can change the behavior in MergedRegistry here: https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/src/registry/MergedRegistry.ts#L39

PR #4284, opened 2024-08-08
feat: attributable fraud for signers

yorhodes:
### Description

Implements fraud proof attributions for checkpoint signers (validators). It is important to have a record of attribution in storage for slashing replay protection. We also maintain a [whitelist of merkle tree hooks](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b14f997810ebd7dbdff2ac6622a149ae77010ae3/solidity/contracts/libs/CheckpointLib.sol#L50-L54) such that validators cannot simply lie about this part of the checkpoint and circumvent fraud proofs.

### Backward compatibility

Yes

### Testing

Unit Tests


solidity/contracts/AttributeCheckpointFraud.sol
```diff
@@ -0,0 +1,123 @@
...
+        merkleTreeWhitelist[merkleTree] = true;
+    }
+
+    function unattributed(
+        Checkpoint calldata checkpoint,
+        bytes calldata signature
+    ) public view returns (bytes32 digest, address signer) {
+        digest = checkpoint.digest();
+        signer = ECDSA.recover(digest, signature);
+        require(
```
aroralanuk, line 50: why revert, shouldn't we just return it?

solidity/contracts/AttributeCheckpointFraud.sol
```diff
@@ -0,0 +1,123 @@
...
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+
+import {TREE_DEPTH} from "./libs/Merkle.sol";
+import {CheckpointLib, Checkpoint} from "./libs/CheckpointLib.sol";
+import {CheckpointFraudProofs} from "./CheckpointFraudProofs.sol";
+
+enum FraudType {
+    NOT_PROVEN,
+    Whitelist,
```
aroralanuk, line 14: nit: consistent formatting of enum

solidity/contracts/AttributeCheckpointFraud.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    CheckpointFraudProofs public immutable checkpointFraudProofs =
+        new CheckpointFraudProofs();
+
+    mapping(address => bool) public merkleTreeWhitelist;
+
+    mapping(address signer => mapping(bytes32 digest => FraudType))
+        public attributions;
+
+    function whitelist(address merkleTree) external onlyOwner {
```
aroralanuk, line 36: do we need a unwhitelist function too?

solidity/contracts/AttributeCheckpointFraud.sol
```diff
@@ -0,0 +1,123 @@
...
+contract AttributeCheckpointFraud is Ownable {
+    using CheckpointLib for Checkpoint;
+    using Address for address;
+
+    CheckpointFraudProofs public immutable checkpointFraudProofs =
+        new CheckpointFraudProofs();
+
+    mapping(address => bool) public merkleTreeWhitelist;
+
+    mapping(address signer => mapping(bytes32 digest => FraudType))
```
aroralanuk, line 33: will be support signer rotation in the future, will it be relevant here?

solidity/test/AttributeCheckpointFraud.t.sol
```diff
@@ -0,0 +1,248 @@
...
+            abi.encode(digest, keccak256(abi.encode(signer, uint(2))))
+        );
+        vm.store(address(acf), slot, bytes32(uint(fraudType)));
+    }
+
+    function test_unattributed(
+        Checkpoint calldata checkpoint,
+        uint256 privateKey
+    ) public {
+        vm.assume(
```
aroralanuk, line 58: why do we need this?

aroralanuk (approved):
Nice, quite straightforward.

solidity/contracts/AttributeCheckpointFraud.sol
yorhodes: intentionally cased this differently because its not really a valid type but is the default in storage

solidity/contracts/AttributeCheckpointFraud.sol
yorhodes: the intention was to prevent fraud from being attributed multiple times, I think I will just add a block.timestamp thing instead

solidity/contracts/AttributeCheckpointFraud.sol
yorhodes: if we add then validators will need to monitor this contract, imo its unnecessary

solidity/test/AttributeCheckpointFraud.t.sol
yorhodes: added a comment

PR #4285, opened 2024-08-08
fix(cli): incorrect command for new chains

ltyu:
### Description
Fixes the new chain message to display the correct command


jmrossy (approved):

jmrossy:
FWIW I still this `registry init` is not an intuitive command name for defining a chain metadata :P

PR #4170, opened 2024-07-22
feat: add `IStaticWeightedIsm` implementation

aroralanuk:
### Description

- Adding the static variation of the IWeightedIsm config where you configure validators with their corresponding weights and store it as bytecode.

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4160

### Backward compatibility

Yes

### Testing

Fuzz tests

PR #4289, opened 2024-08-09
fix: revert interface change for AbstractMultisig

aroralanuk:
### Description

PR #4170 added signatureCount to the AbstractMultisig interface for making sure the correct revert message has shown not enough signatures passed to ism.verify() but this is a breaking change so I've converted it into an internal function.

### DIFF
--- a/solidity/contracts/isms/multisig/AbstractWeightedMultisigIsm.sol
+++ b/solidity/contracts/isms/multisig/AbstractWeightedMultisigIsm.sol
@@ -47,6 +47,17 @@ abstract contract AbstractStaticWeightedMultisigIsm is
     ) public view virtual returns (ValidatorInfo[] memory, uint96);
 
     /**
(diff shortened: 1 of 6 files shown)


solidity/contracts/isms/multisig/AbstractWeightedMultisigIsm.sol
```diff
@@ -46,6 +46,17 @@ abstract contract AbstractStaticWeightedMultisigIsm is
...
+    /**
+     * @notice public function for reading number of signatures in metadata
+     * @param _metadata ABI encoded module metadata
+     * @return the number of signatures
+     */
+    function signatureCount(
+        bytes calldata _metadata
+    ) public pure virtual returns (uint256) {
+        return _signatureCount(_metadata);
+    }
```
yorhodes, line 58: why are you so insistent that this is on the public interface of the contract?

Issue #4228, opened 2024-07-31
IGP configs inside a fallback routing hook should not be checked N times if they are the same


tkporter:
## Problem

Our default hook includes a fallback routing hook where each domain points to the same config under the hood. The real config is:
- for known domains, points to an aggregation(pausable, merkle, igp)
- for unknown domains, points to the merkle tree hook

The logic in deploy.ts (and maybe check-deploy.ts too? unsure) will loop through every domain the fallback routing hook config and try to deploy / check the inner config, and each of these inner configs is identical. So we have an O(n^2) increase in complexity (arguably even O(n^3) because inside the IGP config, we also add another remote chain that we check). This takes a very very very long time, even if you do `--concurrentDeploy`

## Solution

- Maybe have a cached set of `(hook address, hook config)` to look up if a hook config has already been deployed/checked?

PR #4305, opened 2024-08-13
fix: short-circuit known configs in legacy hook deployer

paulbalaji:
fix: short-circuit known configs in legacy hook deployer

given we already do the process of recovering the same hooks for each domain, this short-circuits that even further to avoid the extra time taken to try, recover, move-on for each hook


tkporter (approved):
love it

fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4228

PR #4311, opened 2024-08-14
feat(governor): Support handling proxyAdmin violation in HyperlaneAppGovernor

Mo-Hussain:
### Description

- Support handling proxyAdmin violation, where the proxyAdmin of a proxy contract is not the expected proxyAdmin, `HyperlaneAppGovernor`
- `HyperlaneCoreGovernor` and `ProxiedRouterGovernor` implementation will use it

### Testing

Manual

### DIFF
--- a/typescript/infra/src/govern/HyperlaneAppGovernor.ts
+++ b/typescript/infra/src/govern/HyperlaneAppGovernor.ts
@@ -11,6 +11,7 @@ import {
   InterchainAccount,
   OwnableConfig,
   OwnerViolation,
+  ProxyAdminViolation,
 } from '@hyperlane-xyz/sdk';
 // @ts-ignore
 import { canProposeSafeTransactions } from '@hyperlane-xyz/sdk';
@@ -346,7 +347,11 @@ export abstract class HyperlaneAppGovernor<
         // Will throw if the transaction fails
         await multiProvider.estimateGas(chain, call, submitterAddress);
         return true;
-      } catch (e) {} // eslint-disable-line no-empty
+      } catch (e) {
+        console.error(
+          `Transaction from ${submitterAddress} failed on ${chain}: ${e}`,
+        );
+      }
       return false;
     };
 
@@ -431,5 +436,43 @@ export abstract class HyperlaneAppGovernor<
         description: `Transfer ownership of ${violation.name} at ${violation.contract.address} to ${violation.expected}`,
       },
     };
+  }
+
+  async handleProxyAdminViolation(violation: ProxyAdminViolation) {
+    const code = await this.checker.multiProvider
+      .getProvider(violation.chain)
+      .getCode(violation.proxyAdmin.address);
+
+    let call;
+    if (code !== '0x') {
+      // admin for proxy is ProxyAdmin contract
+      call = {
+        chain: violation.chain,
+        call: {
+          to: violation.actual,
+          data: violation.proxyAdmin.interface.encodeFunctionData(
+            'changeProxyAdmin',
+            [violation.proxy.address, violation.expected],
+          ),
+          value: BigNumber.from(0),
+          description: `Change proxyAdmin of transparent proxy ${violation.proxy.address} from ${violation.actual} to ${violation.expected}`,
+        },
+      };
+    } else {
+      // admin for proxy is EOA
+      call = {
+        chain: violation.chain,
+        call: {
+          to: violation.proxy.address,
+          data: violation.proxy.interface.encodeFunctionData('changeAdmin', [
+            violation.expected,
+          ]),
+          value: BigNumber.from(0),
+          description: `Change admin of ${violation.proxy.address} to ${violation.expected}`,
+        },
+      };
+    }
+
+    return call;
   }
 }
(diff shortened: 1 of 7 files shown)


typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -346,7 +347,11 @@ export abstract class HyperlaneAppGovernor<
         // Will throw if the transaction fails
         await multiProvider.estimateGas(chain, call, submitterAddress);
         return true;
-      } catch (e) {} // eslint-disable-line no-empty
+      } catch (e) {
```
Mo-Hussain, line 350: Helped with debugging

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -432,4 +437,42 @@ export abstract class HyperlaneAppGovernor<
       },
     };
   }
+
+  async handleProxyAdminViolation(violation: ProxyAdminViolation) {
+    const code = await this.checker.multiProvider
```
Mo-Hussain, line 442: Logic was lifted from https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d0ce06280fab6e7a5560dc0f026a264a7cf8adab/typescript/sdk/src/deploy/HyperlaneDeployer.ts#L502

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -432,4 +437,42 @@ export abstract class HyperlaneAppGovernor<
...
+          data: violation.proxyAdmin.interface.encodeFunctionData(
+            'changeProxyAdmin',
+            [violation.proxy.address, violation.expected],
+          ),
+          value: BigNumber.from(0),
+          description: `Change proxyAdmin of transparent proxy ${violation.proxy.address} from ${violation.actual} to ${violation.expected}`,
+        },
+      };
+    } else {
+      // admin for proxy is EOA
```
Mo-Hussain, line 462: @yorhodes - do we have proxy contracts where the admin is an EOA?

typescript/infra/src/govern/HyperlaneAppGovernor.ts
tkporter: might be a little noisy / concerning to someone less familiar with the implementation and just running tooling, as we expect this to happen a lot as part of deciding which submission type works. Maybe `rootLogger.debug` instead of `console.error` could help? and/or the message suggesting that the error is possibly expected

typescript/infra/src/govern/HyperlaneAppGovernor.ts
tkporter: This goes against the best practice https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/transparent/TransparentUpgradeableProxy.sol#L35-L39 which we've always followed - so imo we don't need to support this case

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -89,6 +93,10 @@ export abstract class HyperlaneAppChecker<
     const provider = this.multiProvider.getProvider(chain);
     const contracts = this.app.getContracts(chain);
 
+    const proxyAdminContract = proxyAdminAddress
```
tkporter, line 96: could this just be
```
const expectedProxyAdminContract = ProxyAdmin__factory.connect(expectedAdmin, provider)
```
So that we don't need to have this "prioritize proxyAdminAddress otherwise use the proxyAdmin contract from app` logic in two places

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -101,6 +109,8 @@ export abstract class HyperlaneAppChecker<
               name,
               expected: expectedAdmin,
               actual: actualAdmin,
+              proxyAdmin: proxyAdminContract,
```
tkporter, line 112: could we put "expected" in the `proxyAdmin` and `proxyAdminContract` names so this is a little more clear at a glance?

tkporter (approved):
Some suggestions but lgtm!

typescript/infra/src/govern/HyperlaneAppGovernor.ts
Mo-Hussain: I did lift this logic from the sdk 

[hyperlane-monorepo/typescript/sdk/src/deploy/HyperlaneDeployer.ts](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d0ce06280fab6e7a5560dc0f026a264a7cf8adab/typescript/sdk/src/deploy/HyperlaneDeployer.ts#L502)

where we still support this case

typescript/infra/src/govern/HyperlaneAppGovernor.ts
Mo-Hussain: Fair

typescript/infra/src/govern/HyperlaneAppGovernor.ts
tkporter: ah gotcha, I'm happy to keep the code path fwiw

PR #4312, opened 2024-08-14
feat(HypERC20App): Support proxiedFactories in HypERC20App, extend HypERC20Checker with ProxiedRouterChecker

Mo-Hussain:
### Description

- Extend `HypERC20Checker` with `ProxiedRouterChecker` so we can check for proxyAdmin violations (`checkProxiedContracts` is call in `checkChain`)
- Support passing `proxiedFactories` to `HypERC20App`  

### Fixes
- #4325 

### Testing

Manual

### DIFF
--- a/typescript/infra/scripts/check-deploy.ts
+++ b/typescript/infra/scripts/check-deploy.ts
@@ -13,6 +13,7 @@ import {
   InterchainQueryChecker,
   attachContractsMapAndGetForeignDeployments,
   hypERC20factories,
+  proxiedFactories,
 } from '@hyperlane-xyz/sdk';
 import { objFilter } from '@hyperlane-xyz/utils';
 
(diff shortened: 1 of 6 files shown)


typescript/infra/scripts/check-deploy.ts
```diff
@@ -153,18 +154,21 @@ async function check() {
...
-    const addresses = getWarpAddresses(warpRouteId);
-    const filteredAddresses = Object.keys(addresses) // filter out changes not in config
+    const warpAddresses = getWarpAddresses(warpRouteId);
+    const filteredAddresses = Object.keys(warpAddresses) // filter out changes not in config
       .filter((key) => key in config)
       .reduce((obj, key) => {
-        obj[key] = addresses[key];
+        obj[key] = {
+          ...warpAddresses[key],
+          proxyAdmin: chainAddresses[key].proxyAdmin,
```
yorhodes, line 163: this assumes that the proxyAdmin of the mailbox should always be the same as warp routes, is that desirable? (I think so)

typescript/infra/scripts/check-deploy.ts
```diff
@@ -153,18 +154,21 @@ async function check() {
...
+        };
         return obj;
-      }, {} as typeof addresses);
+      }, {} as typeof warpAddresses);
 
     const { contractsMap, foreignDeployments } =
       attachContractsMapAndGetForeignDeployments(
         filteredAddresses,
-        hypERC20factories,
+        { ...hypERC20factories, ...proxiedFactories },
```
yorhodes, line 171: nice

yorhodes (approved):

typescript/infra/scripts/check-deploy.ts
Mo-Hussain: We have taken that stance, one proxyAdmin per chain that owns all proxies, we've changed the admin contract for proxies that did not meet this requirement

PR #4313, opened 2024-08-14
chore: enroll worldchain, xlayer, cheesechain, zircuit validators

paulbalaji:
- chore: enroll worldchain, xlayer, cheesechain, zircuit validators
- fix: update cosmos zod schema
- drive-by: update core checker to check if the ism in registry matches the expected config


tkporter (approved):

typescript/sdk/src/utils/cosmos.ts
```diff
@@ -431,18 +741,22 @@ export const CosmosChainSchema = z
...
+            block_page: z.string().min(1).optional(),
           })
           .strict(),
       )
       .optional(),
-    keywords: z.array(z.string()).optional(),
+    keywords: z.array(z.string().min(1)).optional(),
     extra_codecs: z.array(z.enum(['ethermint', 'injective'])).optional(),
   })
-  .passthrough()
```
tkporter, line 445: can you keep `passthrough()`, we intentionally added this to be robust to schema additions on their side

tkporter (approved):

typescript/sdk/src/utils/cosmos.ts
tkporter: a comment about this would help too

paulbalaji: spoke offline but 

passthrough vs strict.and(intersection(any,any)) should be the same thing, will leave a comment

typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -518,8 +518,13 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
   },
 
   zircuit: {
-    threshold: 1,
-    validators: ['0x169ec400cc758fef3df6a0d6c51fbc6cdd1015bb'],
+    threshold: 2,
```
paulbalaji, line 521:
```suggestion
    threshold: 3,
```

PR #4320, opened 2024-08-14
feat: flexible voting client on symbiotic vault

yorhodes:
### Description

- implements https://flexiblevoting.com/docs/contracts/flex-voting-client on https://docs.symbiotic.fi/core-modules/vaults


solidity/contracts/dao/FlexibleVotingVault.sol
```diff
@@ -0,0 +1,35 @@
...
+        vault = IVault(_vault);
+    }
+
+    function _rawBalanceOf(
+        address _user
+    ) internal view override returns (uint256) {
+        return vault.activeBalanceOf(_user);
+    }
+
+    // betrays the client interface by asking for timestamp instead of block number
```
nambrot, line 21: I think you'll need to expand the docs here if you gonna do this

nambrot: Did you talk with ScopeLift about this?

alexkeating: I think this should be fine since the function is overriden and not using the underlying history struct which requires a block number

PR #4323, opened 2024-08-15
feat: panics upon startup when RPC issues occur include the offending chain

tkporter:
### Description

Some quick low hanging fruit to make debugging easier in the future. Today (https://discord.com/channels/935678348330434570/1273582453889699843) some chains were having RPC issues and it caused an agent deploy to panic upon startup. Debugging this was hard because the panic message just indicates an error is occurring, but doesn't say what the offending chain is!

Examples:
- https://cloudlogging.app.goo.gl/puQk7hm4fXyJhYKs8
- https://cloudlogging.app.goo.gl/5qZrbc2Ww4LWQXWC6
- https://cloudlogging.app.goo.gl/ZL7mKbBzzJaePtty9

All these panic happen when constructing the cursor, which isn't yet done in an indexing task. The quick and easy fix is to just use expect instead of unwrapIn

### DIFF
--- a/rust/agents/validator/src/validator.rs
+++ b/rust/agents/validator/src/validator.rs
@@ -216,7 +216,10 @@ impl Validator {
         let index_settings =
             self.as_ref().settings.chains[self.origin_chain.name()].index_settings();
         let contract_sync = self.merkle_tree_hook_sync.clone();
-        let cursor = contract_sync.cursor(index_settings).await;
+        let cursor = contract_sync
+            .cursor(index_settings)
+            .await
+            .expect(&format!("Error getting cursor"));
         tokio::spawn(async move {
             contract_sync
                 .clone()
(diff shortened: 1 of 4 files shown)


rust/agents/validator/src/validator.rs
```diff
@@ -216,7 +216,10 @@ impl Validator {
         let index_settings =
             self.as_ref().settings.chains[self.origin_chain.name()].index_settings();
         let contract_sync = self.merkle_tree_hook_sync.clone();
-        let cursor = contract_sync.cursor(index_settings).await;
+        let cursor = contract_sync
+            .cursor(index_settings)
+            .await
+            .expect(&format!("Error getting cursor"));
```
daniel-savu, line 222: nit: maybe worth still including the origin here in case we'll ever move to an omniscient validator

daniel-savu (approved):

PR #4329, opened 2024-08-16
chore: Enable no-restricted-imports eslint rule

jmrossy:
### Description

Incorrect imports from node-specific global libs, especially for `assert` are very common. This won't catch all possible mistakes but it will help avoid future problems like https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4327 

### Backward compatibility

Yes


ltyu (approved):
thanks for taking it 1 step further :)

paulbalaji (approved):

PR #4330, opened 2024-08-16
feat(sdk): Support proxyAdmin checks for admins not owned by AW

Mo-Hussain:

### DIFF
--- a/typescript/sdk/src/deploy/HyperlaneAppChecker.ts
+++ b/typescript/sdk/src/deploy/HyperlaneAppChecker.ts
@@ -25,6 +25,7 @@ import {
   AccessControlViolation,
   BytecodeMismatchViolation,
   CheckerViolation,
+  OwnableConfig,
   OwnerViolation,
   ProxyAdminViolation,
   TimelockControllerViolation,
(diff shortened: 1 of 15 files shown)


typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -81,41 +88,59 @@ export abstract class HyperlaneAppChecker<
...
+
+          // get owner of proxy admin contract
+          const actualProxyAdminContract = ProxyAdmin__factory.connect(
+            actualProxyAdminAddress,
+            provider,
+          );
+          const actualProxyAdminOwner = await actualProxyAdminContract.owner();
+          const expectedOwner = ownableOverrides?.['proxyAdmin'] ?? owner;
+
+          if (!eqAddress(actualProxyAdminOwner, expectedOwner)) {
```
Mo-Hussain, line 113: This will check any proxyAdmin (AW or third party) is owned by the expected owner in the config, we are pulling the proxyAdmin owner check out of `checkOwnership` and doing it here instead.

typescript/sdk/src/token/app.ts
```diff
@@ -42,12 +35,12 @@ export class HypERC20App extends GasRouterApp<
   }
 
   static fromAddressesMap(
-    addressesMap: HyperlaneAddressesMap<HypERC20Factories & ProxiedFactories>,
+    addressesMap: HyperlaneAddressesMap<HypERC20Factories>,
     multiProvider: MultiProvider,
   ): HypERC20App {
     const helper = appFromAddressesMapHelper(
       addressesMap,
-      { ...hypERC20factories, ...proxiedFactories },
```
Mo-Hussain, line 50: No longer passing proxied factories as we cannot assume that we own the relevant `proxyAdmins` and that we have the addresses i.e we don't have the renzo warp route router proxyAdmin addresses in config

PR #4334, opened 2024-08-18
feat: Add succinct lightclient ISM

ltyu:
TBD


solidity/contracts/hooks/DispatchedHook.sol
```diff
@@ -0,0 +1,40 @@
...
+     */
+    function _postDispatch(
+        bytes calldata,
+        bytes calldata message
+    ) internal virtual override {
+        dispatched[message.nonce()] = message.id();
+    }
+
+    /// @inheritdoc AbstractPostDispatchHook
+    function _quoteDispatch(
```
nambrot, line 34: This should probably be configurable to correctly meter for the relayer gas consumption of the ISM?

solidity/contracts/isms/ccip-read/SP1LightClientIsm.sol
```diff
@@ -0,0 +1,194 @@
...
+import {Mailbox} from "../../Mailbox.sol";
+import {DispatchedHook} from "../../hooks/DispatchedHook.sol";
+import {StorageProof} from "../../libs/StateProofHelpers.sol";
+import {ISuccinctProofsService} from "../../interfaces/ccip-gateways/ISuccinctProofsService.sol";
+
+/**
+ * @title SP1LightClientIsm
+ * @notice Uses Succinct to verify that a message was delivered via a Hyperlane Mailbox and tracked by DispatchedHook
+ */
+contract SP1LightClientIsm is AbstractCcipReadIsm, OwnableUpgradeable {
```
nambrot, line 33: I'm a bit out my comfort zone, but IIUC there should be some pretty general logic in here that could be extracted to a "StorageProofIsm"? Specifically, the storage proof logic seems independent of SP1 right?

solidity/contracts/isms/ccip-read/SP1LightClientIsm.sol
```diff
@@ -0,0 +1,194 @@
...
+// ============ Internal Imports ============
+
+import {AbstractCcipReadIsm} from "./AbstractCcipReadIsm.sol";
+import {Message} from "../../libs/Message.sol";
+import {Mailbox} from "../../Mailbox.sol";
+import {DispatchedHook} from "../../hooks/DispatchedHook.sol";
+import {StorageProof} from "../../libs/StateProofHelpers.sol";
+import {ISuccinctProofsService} from "../../interfaces/ccip-gateways/ISuccinctProofsService.sol";
+
+/**
```
nambrot, line 29: I would love some readme/mermaid diagrams explain the general architecture

typescript/ccip-server/src/config.ts
```diff
@@ -3,21 +3,15 @@ import dotenvFlow from 'dotenv-flow';
 dotenvFlow.config();
 
 const RPC_ADDRESS = process.env.RPC_ADDRESS as string;
-const LIGHT_CLIENT_ADDR = process.env.LIGHT_CLIENT_ADDR as string;
```
nambrot, line 6: I didn't realize that we have this already in the monorepo. I think we should have an opinion on whether they should be living in the monorepo. IMO at minimum it should be modular so that the default is not SP1?

nambrot (commented):
Just did a quick pass, probably would do another

solidity/contracts/isms/ccip-read/SP1LightClientIsm.sol
ltyu: Move generalized logic to StorageProofIsm

typescript/ccip-server/src/config.ts
ltyu: Can we do this in a separate PR considering the current scope? Seems like community-ism repo is not ready yet and needs to be organized (looks mostly to be used for contracts right now)

solidity/contracts/hooks/DispatchedHook.sol
ltyu: Add setDispatchFee

solidity/contracts/isms/ccip-read/SP1LightClientIsm.sol
ltyu: Added to PR description

PR #4338, opened 2024-08-19
feat(sdk): Support proxyAdmin checks for admins not owned by AW

Mo-Hussain:
### Description
- Support checking proxyAdmin owners for non AW proxy admins
- Stop assuming all proxyAdmins will be owned by AW
- This change was motivated by the need to have our checker tooling check Renzo Wapr routes, where Renzo owns the proxy routers

### Drive-by changes
- Clean up the interface for `ProxyAdminViolation`

### Testing

Manual

### DIFF
--- a/typescript/infra/scripts/check-deploy.ts
+++ b/typescript/infra/scripts/check-deploy.ts
@@ -167,7 +167,7 @@ async function check() {
           config[key].owner,
           envConfig.owners[key].owner,
         )
-          ? chainAddresses[key].proxyAdmin
+          ? chainAddresses[key]?.proxyAdmin
           : undefined;
 
         if (proxyAdmin) {
--- a/typescript/infra/scripts/check-deploy.ts
+++ b/typescript/infra/scripts/check-deploy.ts
@@ -15,7 +15,7 @@ import {
   hypERC20factories,
   proxiedFactories,
 } from '@hyperlane-xyz/sdk';
-import { objFilter } from '@hyperlane-xyz/utils';
+import { eqAddress, objFilter } from '@hyperlane-xyz/utils';
 
 import { Contexts } from '../config/contexts.js';
 import { DEPLOYER } from '../config/environments/mainnet3/owners.js';
@@ -161,8 +161,19 @@ async function check() {
       .reduce((obj, key) => {
         obj[key] = {
           ...warpAddresses[key],
-          proxyAdmin: chainAddresses[key].proxyAdmin,
         };
+
+        const proxyAdmin = eqAddress(
+          config[key].owner,
+          envConfig.owners[key].owner,
+        )
+          ? chainAddresses[key].proxyAdmin
+          : undefined;
+
+        if (proxyAdmin) {
+          obj[key].proxyAdmin = proxyAdmin;
+        }
+
         return obj;
       }, {} as typeof warpAddresses);
 
--- a/typescript/infra/scripts/check-deploy.ts
+++ b/typescript/infra/scripts/check-deploy.ts
@@ -163,6 +163,8 @@ async function check() {
           ...warpAddresses[key],
         };
 
+        // if the owner in the config is an AW account, set the proxyAdmin to the AW singleton proxyAdmin
+        // this will ensure that the checker will check that any proxies are owned by the singleton proxyAdmin
         const proxyAdmin = eqAddress(
           config[key].owner,
           envConfig.owners[key].owner,
--- a/typescript/infra/src/govern/HyperlaneAppGovernor.ts
+++ b/typescript/infra/src/govern/HyperlaneAppGovernor.ts
@@ -1,6 +1,7 @@
 import { BigNumber } from 'ethers';
 import prompts from 'prompts';
 
+import { ProxyAdmin__factory } from '@hyperlane-xyz/core';
 import { Ownable__factory } from '@hyperlane-xyz/core';
 import {
   ChainMap,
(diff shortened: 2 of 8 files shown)


typescript/infra/scripts/check-deploy.ts
```diff
@@ -161,8 +161,19 @@ async function check() {
       .reduce((obj, key) => {
         obj[key] = {
           ...warpAddresses[key],
-          proxyAdmin: chainAddresses[key].proxyAdmin,
         };
+
+        const proxyAdmin = eqAddress(
```
Mo-Hussain, line 166: Only pass proxyAdmin address if owner of config is AW

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -435,9 +436,13 @@ export abstract class HyperlaneAppGovernor<
   }
 
   async handleProxyAdminViolation(violation: ProxyAdminViolation) {
-    const code = await this.checker.multiProvider
-      .getProvider(violation.chain)
-      .getCode(violation.expectedProxyAdmin.address);
+    const provider = this.checker.multiProvider.getProvider(violation.chain);
+    const code = await provider.getCode(violation.expected);
+
+    const expectedProxyAdmin = ProxyAdmin__factory.connect(
```
Mo-Hussain, line 442: Instantiate the `expectedProxyAdmin` contract here instead of where violation is captured

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -446,17 +451,17 @@ export abstract class HyperlaneAppGovernor<
         chain: violation.chain,
         call: {
           to: violation.actual,
-          data: violation.expectedProxyAdmin.interface.encodeFunctionData(
+          data: expectedProxyAdmin.interface.encodeFunctionData(
```
Mo-Hussain, line 454: We could probably access this interface without instantiating the Contract

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -81,23 +81,14 @@ export abstract class HyperlaneAppChecker<
 
   async checkProxiedContracts(
     chain: ChainName,
-    proxyAdminAddress?: Address,
+    owner: Address,
+    ownableOverrides?: Record<string, Address>,
   ): Promise<void> {
     const expectedProxyAdminAddress =
-      proxyAdminAddress ?? this.app.getContracts(chain).proxyAdmin.address;
-    if (!expectedProxyAdminAddress) {
```
Mo-Hussain, line 88: The `expectedProxyAdminAddress` is no longer required. This is to support the case where we don't own the proxy and therefore will not know the proxyAdmin address before deployment i.e Renzo warp routes

if `expectedProxyAdminAddress` is not set, we will not check for proxyAdmin address violations. 

For AW owned proxies, we will check that the `proxyAdmin` is the singleton `proxyAdmin`

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -106,15 +97,36 @@ export abstract class HyperlaneAppChecker<
...
             contract.address,
           );
-          if (!eqAddress(actualProxyAdminAddress, expectedProxyAdminAddress)) {
+          const actualProxyAdminContract = ProxyAdmin__factory.connect(
+            actualProxyAdminAddress,
+            provider,
+          );
+          const actualProxyAdminOwner = await actualProxyAdminContract.owner();
+          const expectedOwner = ownableOverrides?.['proxyAdmin'] ?? owner;
+          if (!eqAddress(actualProxyAdminOwner, expectedOwner)) {
```
Mo-Hussain, line 106: Here we check if the `proxyAdmin` owner is the expected config owner. This will do this check for AW owned and non AW owned `proxyAdmins`. 

We need to do this here because for non AW owned `proxyAdmins`, we don't have the `proxyAdmins` address in config and the explicit ownership check will therefore not be done in `checkOwnership`

This does me that the ownership check for AW owned `proxyAdmin` is done twice but I think this is unavoidable and not a high cost

tkporter: Even with these comments it took me a bit to understand what was going on here, could you add comments to the code for future readers? e.g. describing why `expectedProxyAdminAddress` may be undefined, why we have all this logic, etc

tkporter, line 106: do we expect / want this code to run if `expectedProxyAdminAddress` is defined? if not, can we explicitly skip this?

I think we want to make sure that we don't ever create two OwnerViolations. If we call it once here and again later, I think that'd be a bug

Mo-Hussain: I suppose it is safe to assume that that if `expectedProxyAdminAddress` is defined that the ownership check will be done in `checkOwnership` so we can skip it here

typescript/sdk/src/router/ProxiedRouterChecker.ts
```diff
@@ -9,16 +9,29 @@ export abstract class ProxiedRouterChecker<
   App extends RouterApp<Factories>,
   Config extends ProxiedRouterConfig,
 > extends HyperlaneRouterChecker<Factories, App, Config> {
-  async checkOwnership(chain: ChainName): Promise<void> {
+  getOwnableOverrides(chain: ChainName): any {
```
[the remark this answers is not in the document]
Mo-Hussain: Whoops this was an oversight

tkporter (approved):
nice, the comments helped a lot, ty!

PR #4343, opened 2024-08-19
fix: Update renzo zircuit validator set

ltyu:
Should match default Ism: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/1abefa60e2a1ded528e852d6fbbf0255c4dfdc42/typescript/sdk/src/consts/multisigIsm.ts#L539-L542


tkporter (approved):
🫡

cmcewen (approved):

PR #4353, opened 2024-08-20
feat(infra): support renzo warp route for checker tooling

Mo-Hussain:
### Description

- enable checker tooling to check renzo warp route

### Drive-by changes
- pass `mailboxAddress` to `moduleMatchesConfig`, needed for `FALLBACK_ROUTER` checks

### Testing

Manual


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -295,7 +295,7 @@ const metricAppContexts = [
   },
   {
     name: 'renzo_ezeth',
-    matchingList: routerMatchingList(renzoEzEthAddresses),
+    matchingList: matchingList(renzoEzEthAddresses),
```
Mo-Hussain, line 298: not using router as the key

typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConifg.ts
```diff
@@ -0,0 +1,190 @@
...
+  linea: '0xb7092685571B49786F1248c6205B5ac3A691c65E',
+  ethereum: '0xD1e6626310fD54Eceb5b9a51dA2eC329D6D4B68A',
+  fraxtal: '0x8410927C286A38883BC23721e640F31D3E3E79F8',
+  zircuit: '0x8410927C286A38883BC23721e640F31D3E3E79F8',
+};
+
+export const getRenzoEZETHWarpConfig = async (
+  routerConfig: ChainMap<RouterConfig>,
+): Promise<ChainMap<TokenRouterConfig>> => {
+  const registry = await getRegistry();
```
Mo-Hussain, line 121: Elected to use the file system registry, was there a specific reason that you used github registry @nambrot

typescript/infra/scripts/generate-renzo-warp-route-config.ts
```diff
@@ -1,191 +1,12 @@
...
-              },
-            ];
-
-            return ret;
-          },
-        ),
-      ),
-    );
-
+  const tokenConfig = await getRenzoEZETHWarpConfig({});
```
Mo-Hussain, line 9: Call into helper here

typescript/infra/config/environments/mainnet3/agent.ts
tkporter: could we keep tracking the old addresses as `renzo_ezeth_old` below (instead of what's in `renzo_ezeth_old` atm) so we have confidence there are no in-flight messages?

We'll have to be careful rolling it out bc of https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/agent.ts#L313, so we should roll it out to rc first, and if we hit that issue we can remove helloworld for now

typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConifg.ts
tkporter: this seems fine to me, we'll just need to be conscious of the local registry being accurate

paulbalaji (approved):

typescript/infra/scripts/generate-renzo-warp-route-config.ts
Mo-Hussain: Doesn't need it passed in

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -33,7 +32,8 @@ import inevmEthereumUsdtAddresses from './warp/inevm-USDT-addresses.json';
 import injectiveInevmInjAddresses from './warp/injective-inevm-addresses.json';
 import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
 import merklyEthAddresses from './warp/merkly-eth-addresses.json';
-import renzoEzEthAddresses from './warp/renzo-ezETH-addresses.json';
+import renzoEzEthAddressesV1 from './warp/renzo-ezETH-addresses-v1.json';
```
tkporter, line 35: is this v2?

Mo-Hussain: Hmm I guess you are right, v2 introduced fraxtal

Mo-Hussain: contract addresses wont change

PR #4355, opened 2024-08-20
feat(sealevel): charge protocol fees

daniel-savu:
### Description

Adds protocol fees to the Sealevel implementation, by embedding the logic / custody into the mailbox program & PDA. The logic is very similar to what already exists in the IGP implementation.

The instructions to interact with the protocol fees implementation are:
- `Init` (now requires specifying the `protocol_fee` and `beneficiary` in addition to the other mailbox params)
- `ClaimProtocolFees` - invokable by anyone, and sending accumulated lamports to the `beneficiary`
- `SetProtocolFeeConfig` - invokable by the owner, sets new values for `protocol_fee` and / or `beneficiary`

### Drive-by changes

- The restriction applied on max message size has been removed to mirror the Solidity V3 implementation

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4356

### Backward compatibility

No - requires redeploying the mailbox

### Testing

Sealevel program integration tests, as is done for the other programs

### DIFF
--- a/rust/sealevel/programs/mailbox-test/src/functional.rs
+++ b/rust/sealevel/programs/mailbox-test/src/functional.rs
@@ -42,6 +42,7 @@ use crate::utils::{
 const LOCAL_DOMAIN: u32 = 13775;
 const REMOTE_DOMAIN: u32 = 69420;
 const PROTOCOL_FEE: u64 = 1_000_000_000;
+const MAX_PROTOCOL_FEE: u64 = 1_000_000_001;
 
 async fn setup_client() -> (
     BanksClient,
@@ -95,19 +96,25 @@ async fn setup_client() -> (
     (banks_client, payer, test_send_receiver, test_ism)
 }
 
+fn test_protocol_fee_config() -> ProtocolFee {
+    ProtocolFee {
+        fee: PROTOCOL_FEE,
+        beneficiary: Pubkey::new_unique(),
+    }
+}
+
 #[tokio::test]
 async fn test_initialize() {
     let program_id = mailbox_id();
     let (mut banks_client, payer, _, _) = setup_client().await;
-    let beneficiary = Pubkey::new_unique();
-
+    let protocol_fee_config = test_protocol_fee_config();
     let mailbox_accounts = initialize_mailbox(
         &mut banks_client,
         &program_id,
         &payer,
         LOCAL_DOMAIN,
-        PROTOCOL_FEE,
-        beneficiary,
+        MAX_PROTOCOL_FEE,
+        protocol_fee_config.clone(),
     )
     .await
     .unwrap();
@@ -121,10 +128,8 @@ async fn test_initialize() {
             outbox_bump_seed: mailbox_accounts.outbox_bump_seed,
             owner: Some(payer.pubkey()),
             tree: MerkleTree::default(),
-            protocol_fee: ProtocolFee {
-                fee: PROTOCOL_FEE,
-                beneficiary,
-            },
+            max_protocol_fee: MAX_PROTOCOL_FEE,
+            protocol_fee: protocol_fee_config,
         },
     )
     .await;
@@ -161,8 +166,8 @@ async fn test_initialize_errors_if_called_twice() {
         &program_id,
         &payer,
         LOCAL_DOMAIN,
-        PROTOCOL_FEE,
-        Pubkey::new_unique(),
+        MAX_PROTOCOL_FEE,
+        test_protocol_fee_config(),
     )
     .await
     .unwrap();
@@ -174,8 +179,8 @@ async fn test_initialize_errors_if_called_twice() {
         &program_id,
         &payer,
         LOCAL_DOMAIN + 1,
-        PROTOCOL_FEE,
-        Pubkey::new_unique(),
+        MAX_PROTOCOL_FEE,
+        test_protocol_fee_config(),
     )
     .await;
     assert_transaction_error(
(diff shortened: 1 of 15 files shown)


rust/sealevel/programs/mailbox/src/processor.rs
```diff
@@ -852,3 +861,61 @@ fn transfer_ownership(
...
+/// 1. `[]` The beneficiary.
+fn claim_protocol_fees(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
+    let accounts_iter = &mut accounts.iter();
+
+    // Account 0: Outbox PDA.
+    let outbox_info = next_account_info(accounts_iter)?;
+    let outbox = Outbox::verify_account_and_fetch_inner(program_id, outbox_info)?;
+
+    // Account 1: Beneficiary
+    let benecifiary_info = next_account_info(accounts_iter)?;
```
tkporter, line 878: nit: `benecifiary_info` -> `beneficiary_info`

rust/sealevel/programs/mailbox/src/accounts.rs
```diff
@@ -100,6 +100,8 @@ pub struct Outbox {
     pub owner: Option<Pubkey>,
     /// The merkle tree of dispatched messages.
     pub tree: MerkleTree,
+    /// The protocol fee configuration.
```
tkporter, line 103: I imagine we may want a max protocol fee that's set at init time, wdyt? I think it's a best practice for stuff like this

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/hooks/ProtocolFee.sol#L38

rust/sealevel/programs/mailbox-test/src/functional.rs
```diff
@@ -267,6 +309,164 @@ async fn test_dispatch_from_eoa() {
...
+    // ---- Test protocol fee claiming ----
+    let non_beneficiary = new_funded_keypair(&mut banks_client, &payer, 1000000000).await;
+
+    let beneficiary_balance_before = banks_client
+        .get_balance(protocol_fee.beneficiary)
+        .await
+        .unwrap();
+
+    // Accounts:
+    // 0. `[writeable]` The IGP.
```
tkporter, line 382: nit: mentions IGP here and the next line

rust/sealevel/programs/mailbox-test/src/functional.rs
```diff
@@ -267,6 +309,164 @@ async fn test_dispatch_from_eoa() {
...
+        .get_account(mailbox_accounts.outbox)
+        .await
+        .unwrap()
+        .unwrap();
+    let rent_exempt_balance = Rent::default().minimum_balance(outbox_account.data.len());
+    assert_eq!(outbox_account.lamports, rent_exempt_balance);
+}
+
+#[tokio::test]
+async fn test_set_protocol_fee_config() {
```
tkporter, line 420: could we also have a `test_set_protocol_fee_config_errors_if_owner_not_signer`

tkporter (commented):
This is great! Happy to see how easy it is to make changes like this. A couple nits / suggestions but lgtm

rust/sealevel/programs/mailbox/src/processor.rs
```diff
@@ -637,15 +645,18 @@ fn outbox_dispatch(
...
-
     let count = outbox
         .tree
         .count()
         .try_into()
         .expect("Too many messages in outbox tree");
+
+    let protocol_fee = outbox.protocol_fee.fee;
+    invoke(
+        &system_instruction::transfer(payer_info.key, outbox_info.key, protocol_fee),
```
tkporter, line 656: do you know if this will log that the transfer occurred?

I'm thinking about how we'll want to start indexing protocol fees in the scraper very soon. The IGP payments / messages / message deliveries all use a dedicated account where info about it is written.

I think for protocol fees, our options are:

1. put the protocol fee payment in the dispatched message account that's written
2. have a log using `msg!` that we can use to figure out how much protocol fee was paid in a given tx. Bc of hook indexing, we just need the tx hash and can find this using https://solana.com/docs/rpc/http/gettransaction which shows all the logs for the tx
3. the scraper could maybe just assume that a configured protocol fee was paid

2 feels the most reasonable to me - wdyt? If the system program doesn't already log this, wonder if we should add a `msg!` saying the protocol fee was paid and how much it was

tkporter (approved):
Lgtm!

rust/sealevel/programs/mailbox/src/processor.rs
daniel-savu: I checked several txs we made when deploying / sending warp transfers and couldn't find system transfer logs, so I doubt we can assume that.


Option 2 seems to be working well yep, hook indexing ftw:
```
curl --location 'https://mainnetbeta-rpc.eclipse.xyz' \
--header 'Content-Type: application/json' \
--data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getTransaction",
    "params": [
     "32wNq1e37Azb7K81qXBVodmpfZKuKxPpuYuBYaXt6SD2j5g4cNhLkk2z3zaZX9c486T82CWMkMHGjZ6esdyyYWWJ",
      "json"
    ]
}' | jq .result.meta.logMessages
```

daniel-savu: However when parsing logs, we need to make sure to drop any logs that occur before the `invoke` log of the outbox, and after the outbox returns (likely the `Dispatched message log`). Otherwise someone could create a contract that logs `"Protocol fee of {} paid from {} to {}"` and then invokes `outbox_dispatch`, and the scraper wouldn't know which log is the correct one

tkporter: yeah good call. I recall there being an edge case with logs fwiw https://github.com/coral-xyz/anchor/issues/1613 but I think this is fine as I can't imagine the incentive for someone to not have protocol fees indexed

tkporter (approved):

rust/sealevel/programs/mailbox/src/processor.rs
daniel-savu: interesting workaround is to make a self-CPI with call data as the log, since that's never truncated / removed
https://book.anchor-lang.com/anchor_in_depth/events.html#cpi-events

PR #4358, opened 2024-08-21
chore: update nautilus rpc url

tkporter:
### Description

for warp route monitoring
Note my confidence in the warp route deployment tooling in infra is low on v2 - so I just edited the sts

### DIFF
--- a/.github/workflows/monorepo-docker.yml
+++ b/.github/workflows/monorepo-docker.yml
@@ -8,6 +8,7 @@ on:
     paths:
       # For now, because this image is only used to use `infra`, we just build for infra changes
       - 'typescript/infra/**'
+      - 'typescript/sdk/**'
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/monorepo-docker.yml
```diff
@@ -8,6 +8,7 @@ on:
     paths:
       # For now, because this image is only used to use `infra`, we just build for infra changes
       - 'typescript/infra/**'
+      - 'typescript/sdk/**'
```
jmrossy, line 11: Should all package be included here (utils, core)?

jmrossy (approved):

.github/workflows/monorepo-docker.yml
tkporter: added them, ty

PR #4359, opened 2024-08-21
feat: deploy pzETH PoC, various tooling tweaks and fixes, and fix an edge case in the multisig ISM program

tkporter:
### Description

A few small tweaks:
- This now transfers the metadata pointer and metadata update authorities. e.g. on this spl-token I created, this is what it looks like, where there are distinct authorities for the metdata pointer and metadata extensions:
```
$ spl-token display 5kFYTUUG5N4uoxAEFLBB5n45eAzifAGKwcY1uoogn6SS -u mainnet-beta

SPL Token Mint
  Address: 5kFYTUUG5N4uoxAEFLBB5n45eAzifAGKwcY1uoogn6SS
  Program: TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
  Supply: 0
  Decimals: 9
  Mint authority: 5kFYTUUG5N4uoxAEFLBB5n45eAzifAGKwcY1uoogn6SS
  Freeze authority: (not set)
Extensions
  Metadata Pointer:
    Authority: 9bRSUPjfS3xS6n5EfkJzHFTRDa4AHLda8BU2pP4HoWnf
    Metadata address: 5kFYTUUG5N4uoxAEFLBB5n45eAzifAGKwcY1uoogn6SS
  Metadata:
    Update Authority: 6DjHX6Ezjpq3zZMZ8KsqyoFYo1zPSDoiZmLLkxD4xKXS # <-- this is because this was my default keypair in my config
    Mint: 5kFYTUUG5N4uoxAEFLBB5n45eAzifAGKwcY1uoogn6SS
    Name: Renzo Restaked LST
    Symbol: pzETH
    URI: https://raw.githubusercontent.com/hyperlane-xyz/hyperlane-registry/12660fd34d30e960a748d87408a8d88f634f7454/deployments/warp_routes/pzETH/ethereum-solana-metadata.json
```

So I added a step to transfer these away as well. I think it's possible to maybe initialize it right off the bat to the mint, but for now this works

- Made `adjust_gas_price_if_needed` idempotent - ran into an issue where this was called twice before a tx was submitted because of the warp route edge case where there's just a single SVM based chain
- my CLI config has the devnet deployer key `6DjHX6Ezjpq3zZMZ8KsqyoFYo1zPSDoiZmLLkxD4xKXS` set as the default instead of the mainnet deployer `9bRSUPjfS3xS6n5EfkJzHFTRDa4AHLda8BU2pP4HoWnf`. This caused some issues where some CLI commands assumed that the default key was the mainnet deployer. Added some flags to some commands to ensure this is correct
- Updated the Solana gas price to be accurate - it was charging thousands with the provided gas price - it was off by a few orders of magnitude

And one major fix:
- As part of our v3 update, a subtle ISM bug was introduced -- the index of the signed checkpoint was being set to the message's nonce, when it should've been from the metadata's merkle_index. This poses issues if the message nonce != the checkpoint index, which is what happened when I sent a test message from Ethereum to Solana. Yikes! I redeployed a new ISM program and set it as the default on the Mailbox, and all was fixed. I updated the tests to specifically cover this case


typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -439,10 +439,6 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
     ],
   },
 
-  eclipse: {
```
paulbalaji, line 442: how come?

rust/sealevel/client/src/warp_route.rs
```diff
@@ -356,26 +364,37 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
-                .stdout(Stdio::inherit())
-                .stderr(Stdio::inherit())
-                .status()
-                .expect("Failed to run command");
-            println!("set the mint authority to the mint account. Status: {status}");
+            // Burn the metadata pointer, metadata, and mint authorities by moving them to the mint
+            let authorities_to_transfer = &["metadata-pointer", "metadata", "mint"];
+
+            for authority in authorities_to_transfer {
+                println!("Transferring authority: {authority} to the mint account");
```
daniel-savu, line 371: worth displaying the mint account here as well?

rust/sealevel/client/src/warp_route.rs
```diff
@@ -356,26 +364,37 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+
+                let mut cmd = Command::new(spl_token_binary_path.clone());
+                cmd.args([
+                    "authorize",
+                    mint_account.to_string().as_str(),
+                    "metadata-pointer",
```
daniel-savu, line 377: should this be 
```suggestion
                    authority,
```

rust/sealevel/client/src/core.rs
```diff
@@ -26,6 +26,24 @@ pub(crate) fn adjust_gas_price_if_needed(chain_name: &str, ctx: &mut Context) {
             (PROCESS_DESIRED_PRIORITIZATION_FEE_LAMPORTS_PER_TX * 1_000_000)
         // Divide by the max compute units
         / DEFAULT_INSTRUCTION_COMPUTE_UNIT_LIMIT as u64;
+
+        if initial_instructions
```
daniel-savu, line 30: sucks that you ran into this, but also great that we got to battle test the per-chain fee logic, should make future deploys much less annoying

rust/sealevel/programs/ism/multisig-ism-message-id/src/processor.rs
```diff
@@ -249,7 +249,7 @@ fn verify(
                 merkle_tree_hook_address: metadata.origin_merkle_tree_hook,
                 mailbox_domain: message.origin,
                 root: metadata.merkle_root,
-                index: message.nonce,
+                index: metadata.merkle_index,
```
daniel-savu, line 252: maybe a good time to consider using multiple ISMs in e2e

daniel-savu (approved):

typescript/sdk/src/consts/multisigIsm.ts
tkporter: it was a duplicate entry, see https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4359/files#diff-69567aec038546b64be29d4a095e0707d7e0b82ff4696c32945c74458023adf3L145

rust/sealevel/client/src/warp_route.rs
tkporter: oops good catch, ty

rust/sealevel/programs/ism/multisig-ism-message-id/src/processor.rs
tkporter: fwiw this is the only ISM impl we have on Sealevel - I think if we were sending EVM <> Sealevel in e2e we maybe would've caught this

typescript/sdk/src/consts/multisigIsm.ts
paulbalaji: https://y.yarn.co/a026b9b9-2104-4b07-aade-afe5ee6d7674_text.gif

rust/sealevel/programs/ism/multisig-ism-message-id/src/processor.rs
daniel-savu: actually, since the merkle tree lives in the mailbox, I'm a bit unsure about how we could end up with a leaf index that's different from the message nonce, even EVM <> Sealevel

tkporter: For Sealevel origin we'd never get in this situation but for EVM -> Sealevel it's possible (which is what surfaced this)

Iiuc in our e2e tests when we send messages from the EVM we sometimes double insert or skip insertion into the merkle tree, so it'd test the case where nonce != merkle index

Issue #4341, opened 2024-08-19
Add default ISM relayer metric for reporting invalid validator signatures


yorhodes:
## Problem

The relayer builds metadata for the multisig ISMs by searching for a [quorum of checkpoints from validators for *some* root](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3f88aa6f6ed1e23a30d219a122b120702446bd82/rust/hyperlane-base/src/types/multisig.rs#L185-L202). It does not, however, indicate to the operators whether a root is inconsistent, with other validators, or with the relayer's merkle tree. 

## Solution

Relayer operators should know when any validator signs a fraudulent checkpoint, quorum or not. We should add a metric that reports unexpected checkpoints, ideally labelling with the default ISM context when relevant.

## Nice to Have

- [ ] include the correct root/message ID in the metric
- [ ] make a merkle proof to the correct root/message ID available

PR #4360, opened 2024-08-21
feat: support `IWeightedMultisigIsm` in relayer

aroralanuk:
### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1345


rust/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -89,6 +89,28 @@ pub trait MultisigIsmMetadataBuilder: AsRef<MessageMetadataBuilder> + Send + Syn
         let metas: Result<Vec<Vec<u8>>> = self.token_layout().iter().map(build_token).collect();
         Ok(metas?.into_iter().flatten().collect())
     }
+
+    async fn ism_validator_requirements(
+        &self,
+        ism_address: H256,
+        message: &HyperlaneMessage,
+    ) -> Result<(Vec<(H256, u64)>, u64)> {
```
yorhodes, line 97: would it make more sense to represent the validators and weights as a mapping offchain?

rust/agents/relayer/src/msg/metadata/multisig/weighted.rs
```diff
@@ -0,0 +1,88 @@
...
+        validators: &[(H256, u64)],
+        threshold_weight: u64,
+        message: &HyperlaneMessage,
+        checkpoint_syncer: &MultisigCheckpointSyncer,
+    ) -> Result<Option<MultisigMetadata>> {
+        MerkleRootMultisigMetadataBuilder::new(self.0.clone())
+            .fetch_metadata(validators, threshold_weight, message, checkpoint_syncer)
+            .await
+    }
+}
```
yorhodes, line 63: is there no way to just inherit the implementation?

rust/config/testnet_config.json
yorhodes, line 1: why are these changed?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -114,7 +119,7 @@ impl MultisigCheckpointSyncer {
         // Sort in descending order. The n'th index will represent
         // the highest index for which we (supposedly) have (n+1) signed checkpoints
         latest_indices.sort_by(|a, b| b.cmp(a));
-        if let Some(&highest_quorum_index) = latest_indices.get(threshold - 1) {
+        if let Some(&highest_quorum_index) = latest_indices.get(threshold_weight as usize - 1) {
```
yorhodes, line 122: when the threshold_weight is not an index I dont really understand this logic

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -151,7 +157,14 @@ impl MultisigCheckpointSyncer {
...
 
-        for validator in validators.iter() {
+        let sorted_validators: Vec<(usize, &(H256, u64))> = weighted_validators
+            .iter()
+            .enumerate()
+            .sorted_by_key(|(_, (_, weight))| std::cmp::Reverse(*weight))
+            .collect();
+
+        let mut cumulative_weight: u64 = 0;
+        for (_, (validator, weight)) in sorted_validators {
```
yorhodes, line 167: what is the unused tuple value here?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -197,8 +211,14 @@ impl MultisigCheckpointSyncer {
...
 
-                    // If we've hit a quorum, create a MultisigSignedCheckpoint
-                    if signature_count >= threshold {
+                    // If we've hit a quorum in weight, create a MultisigSignedCheckpoint
+                    if cumulative_weight >= threshold_weight {
+                        signed_checkpoints.sort_by_key(|sc| {
+                            weighted_validators
+                                .iter()
+                                .position(|(v, _)| v == &H256::from(sc.recover().unwrap()))
+                                .unwrap()
```
yorhodes, line 220: is this an optimization for the contracts encoded ordering? please leave a comment

typescript/infra/scripts/deploy.ts
yorhodes, line 1: why is this changed?

rust/agents/relayer/src/msg/metadata/multisig/base.rs
aroralanuk: I don't understand like validator[weight] = 13e9?

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: index

rust/agents/relayer/src/msg/metadata/multisig/weighted.rs
aroralanuk: I've tried it but the compiler was complaining, don't remember the exact issue

rust/config/testnet_config.json
aroralanuk: reverted

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: Yes, it's identical to the unweighted version

typescript/infra/scripts/deploy.ts
aroralanuk: reverted

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: added comment

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: yes, thanks for pointing it out. Will file it under "CRIT"

typescript/infra/config/environments/test/aggregationIsm.ts
```diff
@@ -1,14 +1,20 @@
...
+} from './multisigIsm.js';
 
 export const aggregationIsm = (validatorKey: string): AggregationIsmConfig => {
   return {
     type: IsmType.AGGREGATION,
     modules: [
       merkleRootMultisig(validatorKey),
       messageIdMultisig(validatorKey),
+      uniformlyWeightedMultisigIsm(merkleRootMultisig(validatorKey)),
+      uniformlyWeightedMultisigIsm(messageIdMultisig(validatorKey)),
```
yorhodes, line 16: nice, full coverage :)

rust/hyperlane-base/src/types/multisig.rs
yorhodes: should you use `iter()`?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -1,3 +1,4 @@
+use itertools::Itertools;
```
yorhodes, line 1: is this unused? remove pls

rust/config/testnet_config.json
yorhodes, line 1: revert these?

rust/chains/hyperlane-ethereum/src/contracts/validator_announce.rs
yorhodes, line 1: revert this

rust/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -362,6 +372,16 @@ impl BaseMetadataBuilder {
             .await
     }
 
+    // TODO
```
yorhodes, line 375: what is TODO? remove if this is stale

typescript/infra/config/environments/test/aggregationIsm.ts
yorhodes: imo we should test a version which is not uniformly weighted though

yorhodes (approved):

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: we still need a map with iter()?

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -114,7 +125,19 @@ impl MultisigCheckpointSyncer {
         // Sort in descending order. The n'th index will represent
         // the highest index for which we (supposedly) have (n+1) signed checkpoints
         latest_indices.sort_by(|a, b| b.cmp(a));
-        if let Some(&highest_quorum_index) = latest_indices.get(threshold - 1) {
+
+        // Find the highest index that meets the threshold weight
+        let mut cumulative_weight = 0;
+        let mut highest_quorum_index = None;
+        for (i, &index) in latest_indices.iter().enumerate() {
+            cumulative_weight += weighted_validators[i].1; // Add the weight of this validator
```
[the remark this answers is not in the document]
aroralanuk: done

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: no, it's used here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b025a3af7e8cf4b96c0eaae18afdff2398235195/rust/hyperlane-base/src/types/multisig.rs#L179

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -151,8 +178,15 @@ impl MultisigCheckpointSyncer {
...
 
-        for validator in validators.iter() {
-            let addr = H160::from(*validator);
+        let sorted_validators: Vec<(usize, &ValidatorWithWeight)> = weighted_validators
+            .iter()
+            .enumerate()
+            .sorted_by_key(|(_, vw)| std::cmp::Reverse(vw.weight))
+            .collect();
+
+        let mut cumulative_weight: u64 = 0;
```
tkporter, line 187: we shouldn't have a cumulative_weight defined at this level

the logic in this function is intended to be robust to validators in a set signing different roots, in which case it shouldn't be considered a quorum. That's why we have this `signed_checkpoints_per_root` HashMap keyed by roots. If everything works correctly, there should only ever be a single key, but this makes us tolerant to misbehaving validators.

Now seeing this we should probably extend this key to not be the merkle root but instead the digest of the checkpoint, as it's possible for misbehaving validators to modify other things like the merkle tree hook address / domain / message id. So instead of `let root = signed_checkpoint.value.root;`, something like `let root = signed_checkpoint.value.signing_hash();`

But anyways imo this should match the logic we had before, where instead of having a mutable `signature_count` that would be added to with each found checkpoint, we only look at the cumulatively found weights in the appropriate entry of `signed_checkpoints_per_root`

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -186,32 +220,40 @@ impl MultisigCheckpointSyncer {
...
 
-                    // If we've hit a quorum, create a MultisigSignedCheckpoint
-                    if signature_count >= threshold {
+                    // If we've hit a quorum in weight, create a MultisigSignedCheckpoint
+                    if cumulative_weight >= threshold_weight {
+                        // to conform to the onchain ordering of the set by address
+                        signed_checkpoints.sort_by_key(|sc| {
+                            weighted_validators
+                                .iter()
+                                .position(|vw| vw.validator == H256::from(sc.recover().unwrap()))
```
tkporter, line 241: We should avoid all unwraps at this part of the code. Just return an Err with `?` here and the line below. Even if logically we know that unwrap should never get hit because of prior logic, imo this is an unnecessary code smell and if the prior logic gets changed in the future, it won't always be obvious if it has implications further down and could result in a panic

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -186,32 +220,40 @@ impl MultisigCheckpointSyncer {
...
                         "Found signed checkpoint"
                     );
 
-                    // If we've hit a quorum, create a MultisigSignedCheckpoint
-                    if signature_count >= threshold {
+                    // If we've hit a quorum in weight, create a MultisigSignedCheckpoint
+                    if cumulative_weight >= threshold_weight {
+                        // to conform to the onchain ordering of the set by address
+                        signed_checkpoints.sort_by_key(|sc| {
+                            weighted_validators
```
tkporter, line 239: This feels like an unnecessarily expensive sorting

`sc.recover()` is pretty expensive and iiuc from the docs https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by_key we'll do this a bunch of times, so if an expensive operation is performed `sort_by_cached_key` is likely to be faster

But regardless this approach feels a bit wasteful, is there a way we can avoid the `sc.recover()` stuff?
Maybe by changing the type of `signed_checkpoints_per_root` to include a property with the intended index of the validator so it's easier to sort later

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -10,6 +11,18 @@ use hyperlane_core::{
 };
 
 use crate::{CheckpointSyncer, CoreMetrics};
+/// Weights are scaled by 1e10 as 100%
```
tkporter, line 14: does this matter from the agent's POV? feels like a smart contract impl detail that has no impact on agent operations

rust/agents/relayer/src/msg/metadata/multisig/weighted.rs
```diff
@@ -0,0 +1,108 @@
...
+            .context(CTX)?;
+
+        let validators: Vec<ValidatorWithWeight> = validators
+            .into_iter()
+            .map(|(validator, weight)| ValidatorWithWeight::new(validator, weight))
+            .collect();
+
+        Ok((validators, threshold))
+    }
+    fn token_layout(&self) -> Vec<MetadataToken> {
```
tkporter, line 93: nit: can you put a newline before this, easier to read with a line between fns

tkporter, line 93: Can we just make this function not take in `&self`? None of the impls rely on self. Then we can just do `MessageIdMultisigMetadataBuilder::token_layout()`

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -663,6 +663,27 @@ impl ChainConf {
         .context(ctx)
     }
 
+    /// Try to convert the chain setting into a Multisig Ism contract
+    pub async fn build_weighted_multisig_ism(
+        &self,
+        address: H256,
+        metrics: &CoreMetrics,
+    ) -> Result<Box<dyn WeightedMultisigIsm>> {
+        let ctx = "Building weighed multisig ISM";
```
tkporter, line 672: nit: weighed -> weighted

rust/chains/hyperlane-ethereum/src/ism/weighted_multisig_ism.rs
```diff
@@ -0,0 +1,123 @@
...
+{
+    contract: Arc<EthereumWeightedMultisigIsmInternal<M>>,
+    domain: HyperlaneDomain,
+}
+
+impl<M> EthereumWeightedMultisigIsm<M>
+where
+    M: Middleware + 'static,
+{
+    /// Create a reference to a mailbox at a specific Ethereum address on some
```
tkporter, line 53: nit: change mailbox

rust/chains/hyperlane-ethereum/src/ism/weighted_multisig_ism.rs
```diff
@@ -0,0 +1,123 @@
...
+        message: &HyperlaneMessage,
+    ) -> ChainResult<(Vec<(H256, u64)>, u64)> {
+        let (validator_addresses, threshold) = self
+            .contract
+            .validators_and_threshold_weight(RawHyperlaneMessage::from(message).to_vec().into())
+            .call()
+            .await?;
+        let validators: Vec<(H256, u64)> = validator_addresses
+            .iter()
+            .map(|x| (H256::from(x.signing_address), x.weight as u64))
```
tkporter, line 108: the weight can be a uint96 on chain right?

At the least we should use `try_into()` instead of using `as` (which can overflow) here / below with threshold / anywhere else

But also wondering if the weight type should be u128 in the agents to accommodate thresholds up to 96 bits?

rust/agents/relayer/src/msg/metadata/multisig/weighted.rs
```diff
@@ -0,0 +1,108 @@
...
+            .await
+    }
+}
+
+#[derive(Debug, Clone, Deref, new, AsRef)]
+pub struct WeightedMessageIdMultisigMetadataBuilder(MessageMetadataBuilder);
+
+#[async_trait]
+impl MultisigIsmMetadataBuilder for WeightedMessageIdMultisigMetadataBuilder {
+    async fn ism_validator_requirements(
```
tkporter, line 69: not a huge deal but there's more code duplication between the different multisig types now with this - both for the weighted and non weighted case. Helper functions may be useful

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -10,6 +11,18 @@ use hyperlane_core::{
...
+pub type Weight = u64;
```
tkporter, line 15: nit: imo type aliases like this can be more confusing than just using u64 but happy either way

rust/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
```diff
@@ -76,4 +76,30 @@ impl MultisigIsmMetadataBuilder for MerkleRootMultisigMetadataBuilder {
...
+            .validators_and_threshold(message)
+            .await
+            .context(CTX)?;
+
+        let unit_validators: Vec<ValidatorWithWeight> = validators
+            .into_iter()
+            .map(|v| ValidatorWithWeight::new(v, 1))
+            .collect();
+
+        Ok((unit_validators, threshold as Weight))
```
tkporter, line 103: nit: use `.into()` to signal this is safe

rust/agents/relayer/src/msg/metadata/multisig/message_id_multisig.rs
```diff
@@ -84,4 +84,30 @@ impl MultisigIsmMetadataBuilder for MessageIdMultisigMetadataBuilder {
...
+            .validators_and_threshold(message)
+            .await
+            .context(CTX)?;
+
+        let unit_validators: Vec<ValidatorWithWeight> = validators
+            .into_iter()
+            .map(|v| ValidatorWithWeight::new(v, 1))
+            .collect();
+
+        Ok((unit_validators, threshold as Weight))
```
tkporter, line 111: ditto

tkporter (commented):
in the future please wait to merge till someone from the Rust / agents side reviews this

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -114,7 +128,19 @@ impl MultisigCheckpointSyncer {
         // Sort in descending order. The n'th index will represent
         // the highest index for which we (supposedly) have (n+1) signed checkpoints
         latest_indices.sort_by(|a, b| b.cmp(a));
-        if let Some(&highest_quorum_index) = latest_indices.get(threshold - 1) {
+
+        // Find the highest index that meets the threshold weight
+        let mut cumulative_weight = 0;
+        let mut highest_quorum_index = None;
+        for (i, &index) in latest_indices.iter().enumerate() {
+            cumulative_weight += weighted_validators[i].weight; // Add the weight of this validator
```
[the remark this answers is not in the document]
tkporter: `latest_indices` also isn't guaranteed to have an element for each validator. If the validator didn't sign, there won't be a corresponding element https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/hyperlane-base/src/types/multisig.rs#L88

rust/hyperlane-base/src/types/multisig.rs
yorhodes: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4341 is related I think

rust/hyperlane-base/src/types/multisig.rs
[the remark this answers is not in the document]
aroralanuk: I've fixed it here but with return a tuple of (validator,index) from get_validator_latest_checkpoints_and_update_metrics https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/cc5483565dce6ea71ca0b97f4e6840d4ae8bd700/rust/hyperlane-base/src/types/multisig.rs#L263

Otherwise, just the index is not workable and this function's return isn't needed anywhere else.

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: It's nice to be explicit that we want the Weight here instead we want u128/u64 which doesn't tell why?

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: It's moreso context for someone looking at the logs with a weight = 12300000 and not understanding what the scale is. Most likely, looks like an absolute stake amount from this perspective, so it can be confusing.

rust/hyperlane-base/src/types/multisig.rs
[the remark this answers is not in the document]
tkporter: yeah something like this makes sense, though it'll be more clear to me when I do another review

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: Ok, does unwrap_or((true, 0)) work here then?

rust/agents/relayer/src/msg/metadata/multisig/weighted.rs
aroralanuk: It has &self in trait declaration in MultisigIsmMetadataBuilder

PR #4369, opened 2024-08-22
feat: Add check if an address is a contract in Cosmos

ameten:
### Description

Currently, integration into Cosmos make an assumption that the address where a message should be delivered is a contract. This change allows Relayer to actually check if the recipient is a contract.

### Backward compatibility

Yes

### Testing

* Tested using E2E tests for Cosmos.
* Tested using auxiliary unit test (it is marked as ignored since it connects to real gRPC of Neutron).

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/providers/grpc.rs
+++ b/rust/chains/hyperlane-cosmos/src/providers/grpc.rs
@@ -97,19 +97,8 @@ pub trait WasmProvider: Send + Sync {
         block_height: Option<u64>,
     ) -> ChainResult<Vec<u8>>;
 
-    /// Perform a wasm query against a specified contract address.
-    async fn wasm_query_to<T: Serialize + Sync + Send + Clone + Debug>(
-        &self,
-        to: String,
-        payload: T,
-        block_height: Option<u64>,
-    ) -> ChainResult<Vec<u8>>;
-
     /// Request contract info from the stored contract address.
     async fn wasm_contract_info(&self) -> ChainResult<ContractInfo>;
-
-    /// Request contract info from a specified contract address
-    async fn wasm_contract_info_to(&self, to: String) -> ChainResult<ContractInfo>;
 
     /// Send a wasm tx.
     async fn wasm_send<T: Serialize + Sync + Send + Clone + Debug>(
@@ -500,24 +489,11 @@ impl WasmProvider for WasmGrpcProvider {
         T: Serialize + Send + Sync + Clone + Debug,
     {
         let contract_address = self.get_contract_address()?;
-        self.wasm_query_to(contract_address.address(), payload, block_height)
-            .await
-    }
-
-    async fn wasm_query_to<T>(
-        &self,
-        to: String,
-        payload: T,
-        block_height: Option<u64>,
-    ) -> ChainResult<Vec<u8>>
-    where
-        T: Serialize + Send + Sync + Clone,
-    {
         let query_data = serde_json::to_string(&payload)?.as_bytes().to_vec();
         let response = self
             .provider
             .call(move |provider| {
-                let to = to.clone();
+                let to = contract_address.address().clone();
                 let query_data = query_data.clone();
                 let future = async move {
                     let mut client = WasmQueryClient::new(provider.channel.clone());
@@ -547,14 +523,10 @@ impl WasmProvider for WasmGrpcProvider {
 
     async fn wasm_contract_info(&self) -> ChainResult<ContractInfo> {
         let contract_address = self.get_contract_address()?;
-        self.wasm_contract_info_to(contract_address.address()).await
-    }
-
-    async fn wasm_contract_info_to(&self, to: String) -> ChainResult<ContractInfo> {
         let response = self
             .provider
             .call(move |provider| {
-                let to = to.clone();
+                let to = contract_address.address().clone();
                 let future = async move {
                     let mut client = WasmQueryClient::new(provider.channel.clone());
 
--- a/rust/chains/hyperlane-cosmos/src/providers/grpc.rs
+++ b/rust/chains/hyperlane-cosmos/src/providers/grpc.rs
@@ -456,6 +456,13 @@ impl WasmGrpcProvider {
             sequence: base_account.sequence,
         })
     }
+
+    fn get_contract_address(&self) -> Result<&CosmosAddress, ChainCommunicationError> {
+        let contract_address = self.contract_address.as_ref().ok_or_else(|| {
+            ChainCommunicationError::from_other_str("No contract address available")
+        })?;
+        Ok(contract_address)
+    }
 }
 
 #[async_trait]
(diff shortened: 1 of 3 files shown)


rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -534,6 +540,42 @@ impl WasmProvider for WasmGrpcProvider {
         Ok(response.data)
     }
 
+    async fn wasm_contract_info(&self) -> ChainResult<ContractInfo> {
```
daniel-savu, line 543: if you add a getter for `self.contract_address` on the trait, you could move this implementation inside the trait definition (to have it be the default implementation), since I doubt this method would ever need to be implemented differently

ameten: I would go in an opposite direction here: hide `get_contract_address` method inside `WasmGrpcProvider` and reuse it in all other methods of `WasmProvider` implemented by `WasmGrpcProvider`. This way the interface of trait `WastProvider` stays clean. I added a commit to illustrate it.

ameten: Discussed and agreed to remove unnecessary methods from the trait.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -89,7 +90,7 @@ impl HyperlaneProvider for CosmosProvider {
     }
 
     async fn is_contract(&self, _address: &H256) -> ChainResult<bool> {
-        // FIXME
+        let _ = self.grpc_client.wasm_contract_info().await?;
```
daniel-savu, line 93: should this value be returned here?

daniel-savu (approved):
📜

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: I took another look. It is better to report `Ok(false)` on error here since otherwise the message will be re-tried on prepare. I added a commit.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -88,9 +91,14 @@ impl HyperlaneProvider for CosmosProvider {
...
     }
 
-    async fn is_contract(&self, _address: &H256) -> ChainResult<bool> {
-        // FIXME
-        Ok(true)
+    async fn is_contract(&self, address: &H256) -> ChainResult<bool> {
+        match self.grpc_client.wasm_contract_info().await {
+            Ok(c) => Ok(true),
+            Err(e) => {
+                info!("provided address is not a contract, address: {:?}", address);
```
tkporter, line 98: nit: we have some existing logging where `is_contract` is consumed https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/relayer/src/msg/pending_message.rs#L214, imo it's preferable to leave it up to the consumer

Not sure if this was included just for debugging or meant to be checked in, but for the future we've tended to avoid formatting log strings with variable values so it's easier to search / interact with logs. E.g. this would be preferable:
```
info!(?address, "Provided address is not a contract");
```

rust/chains/hyperlane-cosmos/src/providers/grpc/tests.rs
```diff
@@ -0,0 +1,77 @@
...
+use url::Url;
+
+use hyperlane_core::config::OperationBatchConfig;
+use hyperlane_core::{ContractLocator, HyperlaneDomain, KnownHyperlaneDomain};
+
+use crate::address::CosmosAddress;
+use crate::grpc::{WasmGrpcProvider, WasmProvider};
+use crate::{ConnectionConf, CosmosAmount, RawCosmosAmount};
+
+#[ignore]
```
tkporter, line 12: Nice, feels like a happy medium between not having unit tests include fallible network interactions & still being able to test things if issues arise

Wonder if there's an argument for something closer to integration tests in the future where we'd spin up a local cosmos node or something

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -534,15 +521,43 @@ impl WasmProvider for WasmGrpcProvider {
...
+
+                    let request = tonic::Request::new(QueryContractInfoRequest { address: to });
+
+                    let response = client
+                        .contract_info(request)
+                        .await
+                        .map_err(ChainCommunicationError::from_other)?
+                        .into_inner()
+                        .contract_info
+                        .ok_or(ChainCommunicationError::from_other_str(
```
tkporter, line 541: wonder if this function returning a `ChainResult<Option<ContractInfo>>` would be possible / more clear as a return type so we can disambiguate genuine gRPC/network errors from the provided address not being a contract

That is, assuming their API doesn't give an error if there's no contract and instead it has contract_info here as None

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
```diff
@@ -534,15 +521,43 @@ impl WasmProvider for WasmGrpcProvider {
...
+            .provider
+            .call(move |provider| {
+                let to = contract_address.address().clone();
+                let future = async move {
+                    let mut client = WasmQueryClient::new(provider.channel.clone());
+
+                    let request = tonic::Request::new(QueryContractInfoRequest { address: to });
+
+                    let response = client
+                        .contract_info(request)
```
tkporter, line 536: I assume the response size here is generally pretty lightweight? `ContractInfo` seems modest https://docs.rs/terra-proto-rs/latest/terra_proto_rs/cosmwasm/wasm/v1/struct.ContractInfo.html but curious if the extension is ever very large

tkporter (approved):
LGTM, some suggestions but happy with whatever you choose

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: I like this approach (with `?address`). And provided that there is logging already, I'll remove this line.

rust/chains/hyperlane-cosmos/src/providers/grpc/tests.rs
ameten: I would support adding integration tests with isolated environment.

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
ameten: I tested it. Their API returns an error if an address is not a contract. 

So, we return an error in two cases: address is definitely not a contract (first error) and if `contract_info` is empty (it may mean that contract does not have the code to process the asset we are going to transfer to it).

It is strange that a contract account may have empty `contract_info`. I did not see such a case in my testing.

I suppose that `contract_info` is `Option<ContractInfo>` because optional fields is the best practice in gRPC/protobuf definitions, but it is just a conjecture.

rust/chains/hyperlane-cosmos/src/providers/grpc.rs
ameten: The extensions are empty in the test I've done.

PR #4370, opened 2024-08-22
feat(infra): Add check-warp-deploy cron job

Mo-Hussain:
### Description

- add helm chart, helper functions and boilerplate to support check-warp-deploy cron job

### Drive-by changes

- refactor check-deploy to call `getGovernor`

### Testing

Manual


typescript/infra/config/environments/mainnet3/warp/checkWarpDeploy.ts
```diff
@@ -0,0 +1,13 @@
+import { CheckWarpDeployConfig } from '../../../../src/config/funding.js';
+import { environment } from '../chains.js';
+
+export const checkWarpDeployConfig: CheckWarpDeployConfig = {
+  docker: {
+    repo: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
+    tag: '9fa5231-20240822-124838',
```
Mo-Hussain, line 7: Will set to latest in a follow up, deployment rely's on changes on this PR

typescript/infra/config/environments/mainnet3/warp/checkWarpDeploy.ts
```diff
@@ -0,0 +1,13 @@
+import { CheckWarpDeployConfig } from '../../../../src/config/funding.js';
+import { environment } from '../chains.js';
+
+export const checkWarpDeployConfig: CheckWarpDeployConfig = {
+  docker: {
+    repo: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
+    tag: '36f7e14-20240823-160646',
```
Mo-Hussain, line 7: Will set to latest in a follow up, as some of the changes in this PR are needed for the image

typescript/infra/helm/check-warp-deploy/templates/_helpers.tpl
```diff
@@ -0,0 +1,42 @@
+{{/*
```
Mo-Hussain, line 1: @tkporter - lifted most of this form key funder

typescript/infra/helm/check-warp-deploy/values.yaml
```diff
@@ -0,0 +1,12 @@
+image:
+  repository: gcr.io/hyperlane-labs-dev/hyperlane-monorepo
+  tag:
+hyperlane:
+  runEnv: mainnet3
+  chains: []
+cronjob:
+  schedule: '0 15 * * *'
```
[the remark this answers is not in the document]
Mo-Hussain: Yes, I'll add

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -278,7 +207,11 @@ const getGovernor = async (
           chain ?? fork
         } is non evm and it not compatible with warp checker tooling`,
       );
-      return;
+      throw Error(
```
Mo-Hussain, line 210: Any concerns with this function throwing @paulbalaji, It is called during chain deployments

paulbalaji: na let it throw, don't think it's a problem

typescript/infra/helm/check-warp-deploy/templates/cron-job.yaml
```diff
@@ -0,0 +1,32 @@
...
+            - ./typescript/infra/scripts/check/check-warp-deploy.ts
+            - -e
+            - {{ .Values.hyperlane.runEnv }}
+            - --pushMetrics
+            env:
+            - name: PROMETHEUS_PUSH_GATEWAY
+              value: {{ .Values.infra.prometheusPushGateway }}
+            envFrom:
+            - secretRef:
+                name: check-warp-deploy-env-var-secret
```
tkporter, line 32: nit: could you add a trailing newline

typescript/infra/helm/check-warp-deploy/templates/_helpers.tpl
tkporter: Not suggesting to do this now but some of these helm files would be nice to live in a common library that all the charts consume from. This and the external secrets stuff to avoid duplication

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -42,33 +39,36 @@ import {
...
-    context,
-    chain,
-    asDeployer,
-    warpRouteId,
-  } = argv ?? (await getCheckDeployArgs().argv);
+export function getCheckDeployArgs() {
+  return withWarpRouteId(withModule(getCheckBaseArgs()));
+}
 
+export const getGovernor = async (
```
tkporter, line 62: nit: any reason to make this an arrow function instead of a normal `async function`? I don't feel strongly but just for consistency in infra we've usually had top level fns like this stay using the `function` keyword

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -296,10 +229,38 @@ const getGovernor = async (
...
-    return;
+    throw Error(
+      `Checker or governor not implemented not implemented for ${module}`,
+    );
   }
 
   return governor;
 };
+
+export const logViolations = (violations: CheckerViolation[]) => {
```
tkporter, line 240: ditto

typescript/infra/scripts/check/check-warp-deploy.ts
```diff
@@ -1,29 +1,101 @@
...
+} from './check-utils.js';
 
-async function checkWarp() {
-  const argv = await getCheckArgs().argv;
+async function main() {
+  const { environment, asDeployer, chain, fork, context, pushMetrics } =
+    await getCheckWarpDeployArgs().argv;
+
+  const metricsRegister = new Registry();
+  const checkerViolationsGauge = new Gauge({
```
tkporter, line 19: should this use `checkerViolationsGaugeObj`? or should it replace `checkerViolationsGaugeObj`

typescript/infra/scripts/check/check-warp-deploy.ts
```diff
@@ -1,29 +1,101 @@
...
+      if (pushMetrics) {
+        await submitMetrics(
+          metricsRegister,
+          `check-warp-deploy-${environment}`,
+          {
+            overwriteAllMetrics: true,
+          },
+        );
+      }
     } catch (e) {
```
tkporter, line 90: Wondering if we should have some way to know that an error has occurred. We had a similar problem with key funder - we want to know when errors occur, but don't want to stop the funding of other keys/chains if an error does occur

There iirc we'd keep track if any errors were caught and would use that to figure out what code to process.exit with. I think we have some metrics on pod errors available in prometheus that can be used for an alert in this case

tkporter (approved):
very nice 🔥

typescript/infra/scripts/check/check-utils.ts
Mo-Hussain: No reason, I'll change

typescript/infra/scripts/check/check-warp-deploy.ts
Mo-Hussain: Good spot

typescript/infra/scripts/check/check-warp-deploy.ts
Mo-Hussain: This is a good idea, as an aside, what do you think about retries? I've observed some errors due to RPCs

tkporter: What kind of errors have you seen?

Was hoping that SmartProvider is used under the hood and handles the retries at that level

Mo-Hussain: I think theres a error after the smart provider has exhausted all RPCs right, I saw this error once

Issue #4349, opened 2024-08-20
Automate pre-deploy balance checking on AW deployer+relayer keys


paulbalaji:
Since the deployer and relayer key are known, we should script up a way to check balances on all chains before a deployment - instead of manually looking it up on the block explorers

PR #4379, opened 2024-08-23
feat: script to print balances of important keys

paulbalaji:
resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4349

- feat: script to print balances of important keys
- super useful as a pre-deploy spot check

e.g. on testnet4
image

### DIFF
--- a/typescript/infra/scripts/print-balances.ts
+++ b/typescript/infra/scripts/print-balances.ts
@@ -1,4 +1,4 @@
-import { BigNumber } from 'ethers';
+import { formatUnits } from 'ethers/lib/utils.js';
 
 import { Contexts } from '../config/contexts.js';
 import { Role } from '../src/roles.js';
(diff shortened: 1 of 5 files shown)


typescript/infra/scripts/print-balances.ts
```diff
@@ -0,0 +1,82 @@
...
+    process.exit(1);
+  });
+
+function convertToNativeTokenValue(value: BigNumber, decimals: number): string {
+  const divisor = BigNumber.from(10).pow(decimals);
+  const integerPart = value.div(divisor);
+  const fractionalPart = value.mod(divisor).toString().padStart(decimals, '0');
+  const truncatedFractionalPart = fractionalPart.slice(0, 3);
+  return `${integerPart.toString()}.${truncatedFractionalPart}`;
+}
```
paulbalaji, line 82: there's surely a better way ?

Mo-Hussain: Does ethers have some nice helpers?

Mo-Hussain: https://docs.ethers.org/v5/api/utils/display-logic/#utils-formatUnits

Mo-Hussain (approved):

typescript/infra/scripts/print-balances.ts
paulbalaji: hmmm it tidies up the code but need to figure out how to truncate 
image

paulbalaji: nvm got it

PR #4380, opened 2024-08-23
feat: Add ChainSubmissionStrategy and update hyperlane submit

ltyu:
### Description
- This PR is a prerequisite for https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4225
- Adds `ChainSubmissionStrategy` which is a ChainMap of SubmissionStrategy
- Moves the submissionStrategy logic out of the context to allow `--strategy` to be used for both cases (mostly to not having to parse and validate 2 schemas in `getSubmissionStrategy()`)
- Adds logic to assume that all `--transactions` are of the same chainId with explicit validation

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/config/submit.ts
+++ b/typescript/cli/src/config/submit.ts
@@ -3,7 +3,10 @@ import { stringify as yamlStringify } from 'yaml';
 import {
   PopulatedTransactions,
   PopulatedTransactionsSchema,
+  SubmissionStrategy,
 } from '@hyperlane-xyz/sdk';
+import { PopulatedTransaction } from '@hyperlane-xyz/sdk';
+import { MultiProvider } from '@hyperlane-xyz/sdk';
 import { assert, errorToString } from '@hyperlane-xyz/utils';
 
 import { WriteCommandContext } from '../context/types.js';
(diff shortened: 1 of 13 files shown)


typescript/cli/src/config/submit.ts
```diff
@@ -57,6 +62,29 @@ export async function runSubmit({
...
+ * @param transactions - The list of populated transactions.
+ * @returns The name of the chain that the transactions are submitted on.
+ * @throws If the transactions are not all on the same chain or chain is not found
+ */
+function getChainFromTxs(
+  multiProvider: MultiProvider,
+  transactions: PopulatedTransactions,
+) {
+  const firstTransaction = transactions.at(0);
+  assert(firstTransaction, 'Transactions empty');
```
yorhodes, line 78: maybe this can be part of the zod schema?

typescript/cli/src/config/submit.ts
```diff
@@ -57,6 +62,29 @@ export async function runSubmit({
...
+
+  const sameChainIds = transactions.every(
+    (t: PopulatedTransaction) => t.chainId === firstTransaction.chainId,
+  );
+  assert(sameChainIds, 'Transactions must be submitted on the same chains');
```
yorhodes, line 83: interesting, I guess we could use this as the means of multichain support too?

yorhodes (approved):

typescript/cli/src/config/submit.ts
ltyu: yeah i think we can assume all PopulatedTransactions have the same chainId

yorhodes: sorry I meant the non empty criteria

typescript/cli/src/config/submit.ts
ltyu: I will remove this when we allow multichain support in `hyperlane submit`. it's currently restricted to a single chain anyways

Issue #4384, opened 2024-08-26
Sort funds in cosmwasm native transfers


sampocs:
## Problem
CosmWasm `ExecuteContract` message require that the attached `funds` attached are sorted by denom ([ref](https://github.com/CosmWasm/wasmd/blob/1d0c66c4cfdf996d12d044638b685b9d09bbd7c1/x/wasm/types/tx.go#L131) and [ref](https://github.com/cosmos/cosmos-sdk/blob/19e0de5e0aab223b8a0dc53a0717ff0120dcb51a/types/coin.go#L270)). 

The current hyperlane-sdk [implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/092ce5fd392eeec2782949062ff084e25ee68e6e/typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts#L405-L414) does not have sorting, which causes both the message simulation and submission to fail.  

## Solution
Sort funds by denom

PR #4385, opened 2024-08-26
fix: sort cosmwasm funds by denom

sampocs:
### Description
CosmWasm `ExecuteContract` message require that the attached `funds` attached are sorted by denom ([ref](https://github.com/CosmWasm/wasmd/blob/1d0c66c4cfdf996d12d044638b685b9d09bbd7c1/x/wasm/types/tx.go#L131) and [ref](https://github.com/cosmos/cosmos-sdk/blob/19e0de5e0aab223b8a0dc53a0717ff0120dcb51a/types/coin.go#L270)). 

The current hyperlane-sdk [implementation](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/092ce5fd392eeec2782949062ff084e25ee68e6e/typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts#L405-L414) does not have sorting, which causes both the message simulation and submission to fail.  

### Drive-by changes
No

### Related issues
#4384 

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/.changeset/red-nails-trade.md
+++ b/.changeset/red-nails-trade.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Sorted cwNative funds by denom in transfer tx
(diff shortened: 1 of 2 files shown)


jmrossy (approved):
It's odd that this code has been working in the Warp UI until now (AFAIK) but seems like a safe change regardless 👍

sampocs:
Agreed haha. I assume most instances had a single denom, or possibly got just lucky with the ordering

jmrossy:
@sampocs is this issue blocking your work? It would be helpful to know as we decide when to publish the next SDK update

sampocs:
Nope not blocking! We just patched it on our end for the time being

PR #4387, opened 2024-08-27
feat: Add scraper to Cosmos E2E test

ameten:
### Description

Add scraper to Cosmos E2E test

* Add minimal functionality to scraper for cosmos so that scraper does not crash
* Add scraper to Cosmos E2E test

### Backward compatibility

Yes

### Testing

E2E Tests for Cosmos


rust/agents/scraper/migration/src/m20240823_000006_add_injective_neutron_osmosis.rs
```diff
@@ -0,0 +1,138 @@
...
+use sea_orm::ActiveValue::Set;
+use sea_orm::{ActiveModelTrait, DbErr, DeriveMigrationName, NotSet};
+use sea_orm_migration::{MigrationTrait, SchemaManager};
+use time::OffsetDateTime;
+
+use crate::async_trait;
+use crate::m20240823_000006_add_injective_neutron_osmosis::domain::ActiveModel;
+
+/// List of domain data we want to add into database.
+const DOMAINS: &[RawDomain] = &[
```
tkporter, line 11: fwiw to date our habits around adding domains haven't required us to create a new migration for new chains - calling it out not because I think our existing habits are good though 🙈

Our process for adding new chains has just been:
- add the new chain to the existing migration https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/agents/scraper/migration/src/m20230309_000001_create_table_domain.rs#L16. We haven't actually applied this migration against the prod db since we initially stood it up. So this change is really only applied in e2e, or if we were to stand up a new scraper DB in prod.
- when adding new domains into our prod db, we run a SQL update query directly against the db to add them 😅

I'm very happy to change this process as obviously it's sketchy, but wondering if you'd prefer to adjust this process as part of your work here or to change this later. I'm fine with just adding these domains into the existing migration file (less duplicated code for the up and down logic) and then we can work with Paul to figure out a better setup for the future when it comes to prod db changes - but if you think it's cleaner in general to have separate migrations, I'm happy with that too

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -82,12 +88,95 @@ impl HyperlaneChain for CosmosProvider {
...
+    async fn get_block_by_hash(&self, hash: &H256) -> ChainResult<BlockInfo> {
+        let tendermint_hash = Hash::from_bytes(Algorithm::Sha256, hash.as_bytes())
+            .expect("block hash should be of correct size");
+
+        // TODO add proper error handling
+        let response = self
+            .rpc_client
+            .block_by_hash(tendermint_hash)
+            .await
+            .map_err(|_| ChainCommunicationError::from_other_str("generic error"))?;
```
tkporter, line 100: nit: should make sure we propagate the actual error here / rm that TODO. Same in `get_txn_by_hash`

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -82,12 +88,95 @@ impl HyperlaneChain for CosmosProvider {
...
+
+        let received_hash = H256::from_slice(response.block_id.hash.as_bytes());
+
+        if &received_hash != hash {
+            return Err(ChainCommunicationError::from_other_str(
+                "received incorrect block",
```
tkporter, line 106: nit: may be nice to surface the hashes in the error, ditto for the tx hash in `get_txn_by_hash`

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -82,12 +88,95 @@ impl HyperlaneChain for CosmosProvider {
...
-    async fn get_txn_by_hash(&self, _hash: &H256) -> ChainResult<TxnInfo> {
-        todo!() // FIXME
+    async fn get_txn_by_hash(&self, hash: &H256) -> ChainResult<TxnInfo> {
+        let tendermint_hash = Hash::from_bytes(Algorithm::Sha256, hash.as_bytes())
+            .expect("transaction hash should be of correct size");
+
+        // TODO add proper error handling
+        let response = self
+            .rpc_client
+            .tx(tendermint_hash, true)
```
tkporter, line 132: does this need to be true? not sure we use the proof

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -82,12 +88,95 @@ impl HyperlaneChain for CosmosProvider {
...
+
+        let received_hash = H256::from_slice(response.hash.as_bytes());
+
+        if &received_hash != hash {
+            return Err(ChainCommunicationError::from_other_str(
+                "received incorrect transaction",
+            ));
+        }
+
+        let tx = Tx::from_bytes(&response.tx).map_err(|_| {
```
tkporter, line 144: might also be nice to propagate the error, share the tx hash if it's useful for debugging

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -82,12 +88,95 @@ impl HyperlaneChain for CosmosProvider {
...
+            ChainCommunicationError::from_other_str("could not deserialize transaction")
+        })?;
+
+        let signer = tx
+            .auth_info
+            .signer_infos
+            .get(0)
+            .expect("there should be at least one signer");
```
tkporter, line 152: while very unlikely I could see it being possible that some weird Cosmos chain has tweaked their chain to allow "system transactions" where there's no signer. Not sure if it'd result in `signer_infos` being empty or what, but I wonder if we should just Err here instead of panic?

tkporter (approved):
very nice!

rust/agents/scraper/migration/src/m20240823_000006_add_injective_neutron_osmosis.rs
ameten: I prefer to adjust it later. 😅

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: It is an initial PR just to make scraper to work with Cosmos and not to fail E2E test. I am going to revisit error handling later.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: I'll revisit error handling later.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: Fixed

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: I'll do error handling in future PRs.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: It is a very valid concern. There can be more edge cases like this. I'll handle them in the future PRs.

PR #4395, opened 2024-08-28
fix: forward value from ICA router to proxies

yorhodes:
### Description

Value sent to the hyperlane message recipient (InterchainAccountRouter) would not be forwarded to the proxy account for use in calls with value. Only value sent to the proxy account was supported.

### Related Issues

- Fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/4

### Backward compatibility

No, ICA routers would need upgrade/redeploy

### Testing

Unit Tests


aroralanuk (approved):
Looks good, just link the chainlight issue here. 

Also, a comment saying that our relayer doesn't support msg.value currently will be useful for someone testing the ICAs.

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -307,7 +307,8 @@ contract InterchainAccountRouter is Router {
             _sender,
             _ism.bytes32ToAddress()
         );
-        _interchainAccount.multicall(_calls);
+        // forward all balance to the interchain account for msg.value and non-call value transfers
+        _interchainAccount.multicall{value: address(this).balance}(_calls);
```
nambrot, line 311: Did you consider summing the msg.value of the calls and passing that in? This feels kinda blunt, but not sure of the full trade-offs

yorhodes: I did not consider that
in what case would we want to leave balance on the router?

nambrot: When composing it with ISMs that pass msg.value non-atomically

yorhodes: this solves value arriving non-atomically on the message recipeint (ICA Router) but actually prevents value from arriving to the ICA account itself
it would force all value to flow through this router which I dont think is desirable

PR #4402, opened 2024-08-29
fix: set profile overrides in root manifest

daniel-savu:
### Description

Fixes `cargo build-sbf` warnings by moving all profile overrides to `rust/Cargo.toml`, so they actually take effect.

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1368

### Backward compatibility

Yes

### Testing

Manually, by initializing `inbox.processed_count` to `u64::MAX - 1`, and then observing logs in e2e:
```
13: Program log: About to increment processed count. Current count: 18446744073709551615
14: Program log: panicked at 'attempt to add with overflow', sealevel/programs/mailbox/src/processor.rs:386:5
```
The panic occurs at `inbox.processed_count += 1;`

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -195,6 +195,73 @@ ya-gcp = { version = "0.11.1", features = ["storage"] }
 ## TODO: remove this
 cosmwasm-schema = "1.2.7"
 
+[profile.release.package.access-control]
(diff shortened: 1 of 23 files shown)


tkporter (approved):
🤝

Issue #4388, opened 2024-08-27
feat(cli): Enable cli typescript e2e tests


ltyu:
Currently, cli e2e TS tests are not enabled in CI because they cause the CI process to unexpectedly hang. Fix this

PR #4404, opened 2024-08-30
feat: Enable cli e2e tests in CI

ltyu:
### Description
This PR enables cli e2e ts tests and adds some tests for Warp apply

### Related issues
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4388

### Backward compatibility
Yes


typescript/cli/package.json
```diff
@@ -44,7 +44,8 @@
     "clean": "rm -rf ./dist",
     "lint": "eslint . --ext .ts",
     "prettier": "prettier --write ./src ./examples",
-    "test": "mocha --config .mocharc.json",
+    "test": "yarn test:e2e",
+    "test:e2e": "./scripts/all-test.sh",
```
jmrossy, line 48: Why separate these two?

typescript/cli/tmp/anvil2/warp-route-deployment-anvil2.yaml
```diff
@@ -0,0 +1,3 @@
+anvil2:
```
jmrossy, line 1: Did you mean to include these? If so, let's not put them in `cli/tmp`, the `tmp` makes it seem like they shouldn't be checked in

ltyu: sorry this was meant to be a draft to debug. Unfortunately i could not get this working with CI yet. The test hangs and then eventually fails (after 5 hours). 

I've concluded that the e2e tests do fails. My hunch is that it's anvil related, possibly conflicting with the foundry test script in the SDK. 

Will need more time to investigate the raw logs.

ltyu: Will remove

typescript/cli/package.json
ltyu: Fixed

.github/workflows/test.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           fi
 
   yarn-test:
-    runs-on: ubuntu-latest
+    runs-on: larger-runner
```
paulbalaji, line 117: oh dang, how come?

would it make sense to put the e2e on a separate job?

jmrossy: +1, these cost more than $1 each run. We want to run only the bare minimum on them

typescript/cli/src/tests/warp-apply.e2e-test.ts
```diff
@@ -35,7 +34,7 @@ const WARP_CORE_CONFIG_PATH_2 = `${REGISTRY_PATH}/deployments/warp_routes/ETH/an
 
 describe('WarpApply e2e tests', async function () {
   let chain2Addresses: ChainAddresses = {};
-  this.timeout(0); // No limit timeout since these tests can take a while
+  this.timeout(60000); // Long timeout since these tests can take a while
```
jmrossy, line 37: I would have expected even longer than one minute actually
Btw, you can use underscores in numbers these days for legibility so `60000` -> `60_000`

.github/workflows/test.yml
ltyu: Good to know! i'll try this without the larger-runner

typescript/cli/src/tests/warp-apply.e2e-test.ts
ltyu: Fixed

paulbalaji (approved):

jmrossy (approved):

PR #4405, opened 2024-08-30
feat: add `IWeightedIsm` relayer change pt 2

aroralanuk:


rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -111,10 +129,15 @@ impl MultisigCheckpointSyncer {
             return Ok(None);
         }
 
-        // Sort in descending order. The n'th index will represent
-        // the highest index for which we (supposedly) have (n+1) signed checkpoints
-        latest_indices.sort_by(|a, b| b.cmp(a));
-        if let Some(&highest_quorum_index) = latest_indices.get(threshold - 1) {
+        // Sort in descending order by index. The n'th index will represent
```
tkporter, line 132: nit: I think this comment about the n'th index is no longer correct

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
+
+            if cumulative_weight >= threshold_weight {
+                return Some(*index);
+            }
+        }
+
+        // If threshold is not met
+        debug!(
+            cumulative_weight = cumulative_weight,
+            threshold_weight = threshold_weight,
```
tkporter, line 315: might be nice to add `sorted_indices` and `weighted_validators` in here too

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
+            "Highest quorum index not found"
+        );
+        None
+    }
+
+    // signed_checkpoint.recover() is expensive, so we cache the result in recovered_addresses
+    fn fetch_or_recover(
```
tkporter, line 322: Hmm I think this function doesn't work for what we wanna do

The main problem is that the `let checkpoint_hash = signed_checkpoint.value.signing_hash();` is not unique for each validator. Like if we have 5 validators all behaving correctly, they'll each have signed the same hash. So going from signing hash -> validator address doesn't work - we'd end up putting the very first validator into the hashmap but not the other ones. This is why we have `let mut signed_checkpoints_per_root: HashMap<H256, Vec<SignedCheckpointWithMessageId>> = HashMap::new();`, where we can go from signing hash -> multiple validators

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
                     }
 
                     // Ensure that the signature is actually by the validator
-                    let signer = signed_checkpoint.recover()?;
+                    let signer =
+                        self.fetch_or_recover(&signed_checkpoint, &mut recovered_addresses)?;
```
tkporter, line 207: I think it's safest to always recover here without any kind of caching

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
                         debug!(
-                            validator = format!("{:#x}", validator),
+                            validator = format!("{:#x}", vw.validator),
                             index = index,
                             "Checkpoint signature mismatch"
                         );
                         continue;
                     }
 
+                    // let checkpoint_hash = signed_checkpoint.value.signing_hash();
```
tkporter, line 218: nit

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
                     signed_checkpoints.push(signed_checkpoint);
 
+                    let cumulative_weight: Weight = signed_checkpoints
+                        .iter()
+                        .filter_map(|sc| {
+                            let signer =
+                                self.fetch_or_recover(sc, &mut recovered_addresses).ok()?;
+                            weighted_validators
+                                .iter()
+                                .find(|vw| vw.validator == signer)
```
tkporter, line 232: I think we should try our best to not have an O(n) operation here. Especially bc we're iterating through signed_checkpoints, so we end up doing like O(n^2) to just get the weight

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
                     // Push the signed checkpoint into the hashmap
                     let root = signed_checkpoint.value.root;
                     let signed_checkpoints = signed_checkpoints_per_root.entry(root).or_default();
                     signed_checkpoints.push(signed_checkpoint);
 
+                    let cumulative_weight: Weight = signed_checkpoints
+                        .iter()
+                        .filter_map(|sc| {
+                            let signer =
+                                self.fetch_or_recover(sc, &mut recovered_addresses).ok()?;
```
tkporter, line 229: I should've been more clear in the last review but I'd prefer to not have any recovery for stuff like this, even if it's cached. I think it's more confusing than adding some additional data to `signed_checkpoints`

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
+                    // If we've hit a quorum in weight, create a MultisigSignedCheckpoint
+                    if cumulative_weight >= threshold_weight {
+                        // to conform to the onchain ordering of the set by address
+                        signed_checkpoints.sort_by_key(|sc| {
+                            self.fetch_or_recover(sc, &mut recovered_addresses)
+                                .ok()
+                                .and_then(|signer| {
+                                    weighted_validators
+                                        .iter()
+                                        .position(|vw| vw.validator == signer)
```
tkporter, line 257: ditto I don't think we should have O(n) operations / anything related to recovering in the sorting fn

I'd probably restructure this function a bit to be something like this instead of having `fetch_or_recover`:
1. Create
```
struct ValidatorSignedCheckpoint {
  signed_checkpoint: SignedCheckpointWithMessageId,
  validator: H256,
  weight: Weight,
  ism_index: usize,
}

#[derive(Default)]
struct ValidatorSignedCheckpoints {
  cumulative_weight: Weight,
  signed_checkpoints: Vec<ValidatorSignedCheckpoint>,
}

impl ValidatorSignedCheckpoints {
  fn push(...) { 
    // push into `self.signed_checkpoints`, update the cumulative weight
  }
}
```

2. Change `signed_checkpoints_per_root` to use this
```
let mut signed_checkpoints_per_root: HashMap<H256, ValidatorSignedCheckpoints> =
            HashMap::new();
```

3. Now `cumulative_weight` can come directly from the map entry in O(1)
4. Now the sorting to match the ISM order here can just use the `ism_index` for a basic O(nlog(n)):
```
validator_signed_checkpoints.signed_checkpoints.sort_by_key(|sc| sc.ism_index)
```

rust/hyperlane-base/src/types/multisig.rs
```diff
@@ -171,51 +203,218 @@ impl MultisigCheckpointSyncer {
...
+        threshold_weight: Weight,
+        sorted_indices: &[(H256, u32)],
+    ) -> Option<u32> {
+        let weight_dict: HashMap<H256, Weight> = weighted_validators
+            .iter()
+            .map(|v| (v.validator, v.weight))
+            .collect();
+
+        let mut cumulative_weight: u128 = 0;
+        let mut validators_included = HashSet::new();
```
tkporter, line 297: I actually don't think we need or want `validators_included` so that we can support a case where an ISM has duplicate validator addresses in there. Because we assume `weighted_validators` is accurate to what the ISM provides right

tkporter (commented):
thanks for making all those changes, think we're very close

rust/hyperlane-base/src/settings/chains.rs
```diff
@@ -663,6 +663,27 @@ impl ChainConf {
...
+        let ctx = "Building weighted multisig ISM";
+        let locator = self.locator(address);
+
+        match &self.connection {
+            ChainConnectionConf::Ethereum(conf) => {
+                self.build_ethereum(conf, &locator, metrics, h_eth::WeighedMultisigIsmBuilder {})
+                    .await
+            }
+            ChainConnectionConf::Fuel(_) => todo!(),
+            ChainConnectionConf::Sealevel(_) => todo!(),
```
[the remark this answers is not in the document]
aroralanuk: Got it, additionally I've added the error for fuel just as a precaution given that fuel might be live in the near future.

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: Don't understand this exactly, I only changed to include the validator alongside the index, how it is incorrect?

rust/hyperlane-base/src/types/multisig.rs
aroralanuk: I'm summing up in the weight_dict hashmap now, not sure that's what we want

PR #4406, opened 2024-08-30
feat: add error handling for blocks and proper sender

ameten:
### Description

* Use correct signer as sender
* Add error handling in getting block by hash

### Backward compatibility

Yes

### Testing

E2E Test for Cosmos (no implementation in invariant)

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/providers/mod.rs
+++ b/rust/chains/hyperlane-cosmos/src/providers/mod.rs
@@ -201,6 +201,11 @@ impl HyperlaneProvider for CosmosProvider {
             .map_err(|e| ChainCommunicationError::from_other_str("could not convert from proto"))?;
         let contract = H256::try_from(CosmosAccountId::new(&msg.contract))?;
 
+        // We calculate the sender and the nonce for the transaction.
+        // We use `payer` of the fees as the sender of the transaction and search for its
+        // signature information to find the nonce.
+        // If `payer` is not specified, we use the account which signed the transaction first, as
+        // the sender.
         let (sender, nonce) = tx
             .auth_info
             .fee
--- a/rust/chains/hyperlane-cosmos/src/providers/mod.rs
+++ b/rust/chains/hyperlane-cosmos/src/providers/mod.rs
@@ -96,10 +96,7 @@ impl CosmosProvider {
             })
             // If there were not any signer info with non-empty public key or no signers for the transaction,
             // we get None here
-            .map_or_else(
-                || Err(ChainCommunicationError::from_other_str("no signer info")),
-                |r| r,
-            )
+            .unwrap_or_else(|| Err(ChainCommunicationError::from_other_str("no signer info")))
     }
 
     fn convert_signer_info_into_account_id_and_nonce(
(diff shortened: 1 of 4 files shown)


rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -172,14 +204,22 @@ impl HyperlaneProvider for CosmosProvider {
...
-            .signer_infos
-            .get(0)
-            .expect("there should be at least one signer");
-        let sender = self.sender(signer)?;
-        let nonce = signer.sequence;
+            .fee
+            .payer
+            .as_ref()
+            .map(|payer| self.search_payer_in_signer_infos(&tx.auth_info.signer_infos, payer))
+            .map_or_else(
```
tkporter, line 213: ditto I think `unwrap_or_else` could be clearer

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -172,14 +204,22 @@ impl HyperlaneProvider for CosmosProvider {
             .map_err(|e| ChainCommunicationError::from_other_str("could not convert from proto"))?;
         let contract = H256::try_from(CosmosAccountId::new(&msg.contract))?;
 
-        // TODO choose correct signer (should it be the one paid for the transaction)
-        let signer = tx
+        let (sender, nonce) = tx
```
tkporter, line 207: a comment about the general approach here may be useful for future readers. Iiuc it'll try using the fee payer if it exists / is able to get a signer there, otherwise it falls back to getting the first signer?

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -172,14 +204,22 @@ impl HyperlaneProvider for CosmosProvider {
...
-        let sender = self.sender(signer)?;
-        let nonce = signer.sequence;
+            .fee
+            .payer
+            .as_ref()
+            .map(|payer| self.search_payer_in_signer_infos(&tx.auth_info.signer_infos, payer))
+            .map_or_else(
+                || {
+                    let signer_info = tx.auth_info.signer_infos.get(0).ok_or_else(|| {
+                        ChainCommunicationError::from_other_str("no signer info in default signer")
```
tkporter, line 216: something we'll have to think about is general error handling in the scraper and how fragile it is. Atm the scraper prefers to not insert any data to the db if there are errors getting things like blocks / txs relating to that piece data. Some context here https://github.com/hyperlane-xyz/issues/issues/1369

It doesn't have to be in this PR, but I think we may want to rethink this. E.g. if there's some special transaction on Cosmos that for some reason doesn't have any fee payer or signers (maybe some "system transaction"), atm the scraper will propagate this error such that the piece of data that really matters, e.g. a message that was indexed in that transaction, ultimately doesn't get inserted into the db.

I wonder instead if some things should just be optional in the DB - e.g. if the sender wasn't able to be found by our logic, that's fine and we just log it and write null to the DB, but we still write the Hyperlane message that occurred in that transaction to the DB

tkporter (approved):

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -78,16 +79,45 @@ impl CosmosProvider {
...
+            // After the following we have a single Ok entry and, possibly, many Err entries
+            .filter_ok(|(a, s)| payer == a)
+            // If we have Ok entry, use it since it is the payer, if not, use the first entry with error
+            .find_or_first(|r| match r {
+                Ok((a, s)) => payer == a,
+                Err(e) => false,
+            })
+            // If there were not any signer info with non-empty public key or no signers for the transaction,
+            // we get None here
+            .map_or_else(
```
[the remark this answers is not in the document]
ameten: It does. Thank you!

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: It is.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: It makes sense. I added a comment.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: I agree. We need to revisit it in more general settings thought.

PR #4407, opened 2024-08-30
fix(v2): set profile overrides in root manifest

daniel-savu:
### Description

Backport of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4402

Fixes `cargo build-sbf` warnings by moving all profile overrides to `rust/Cargo.toml`, so they actually take effect.


tkporter (approved):
🤴

PR #4408, opened 2024-08-30
feat(envTests): Support using secret rpcs for CI env tests

Mo-Hussain:
### Description

- Use Arbitrum and Optimism private RPCs that are set in Github Secrets for env-test
- Increase env-test timeout to 4 mins for viction
- Adapt fork script so that it create anvil instance by calling a script

### Testing

Manual

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -409,7 +409,7 @@ jobs:
       MAINNET3_ARBITRUM_RPC_URLS: ${{ secrets.MAINNET3_ARBITRUM_RPC_URLS }}
       MAINNET3_OPTIMISM_RPC_URLS: ${{ secrets.MAINNET3_OPTIMISM_RPC_URLS }}
 
-    timeout-minutes: 6
+    timeout-minutes: 10
     needs: [yarn-build]
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -405,7 +405,11 @@ jobs:
 
   env-test:
     runs-on: ubuntu-latest
-    timeout-minutes: 3
+    env:
+      MAINNET3_ARBITRUM_RPC_URLS: ${{ secrets.MAINNET3_ARBITRUM_RPC_URLS }}
+      MAINNET3_OPTIMISM_RPC_URLS: ${{ secrets.MAINNET3_OPTIMISM_RPC_URLS }}
+
+    timeout-minutes: 4
```
Mo-Hussain, line 412: Increase timeout for viction which can be slow

typescript/infra/fork.sh
```diff
@@ -16,9 +16,7 @@ trap 'jobs -p | xargs -r kill' EXIT
 # exit 1 on any subsequent failures
 set -e
 
-RPC_URL=`LOG_LEVEL=error yarn tsx ./scripts/print-chain-metadatas.ts -e $ENVIRONMENT | jq -r ".$CHAIN.rpcUrls[0].http"`
-
-anvil --fork-url $RPC_URL --fork-retry-backoff 3 --compute-units-per-second 200 --gas-price 1 --silent &
+LOG_LEVEL=error yarn tsx ./scripts/run-anvil.ts -e $ENVIRONMENT -c $CHAIN &
```
Mo-Hussain, line 19: Start anvil using this script to that we can use private RPCs, calling `./scripts/print-chain-metadatas.ts` with private RPCs will log them in CI logs

paulbalaji (approved):

PR #4409, opened 2024-08-30
feat: add error handling and refactoring

ameten:
### Description

* Add error handling for getting transaction by hash
* Add error handing for extracting contract address
* Extract a method for 

### Backward compatibility

Yes

### Testing

Cosmos E2E Testing (without invariant for scraper)

### DIFF
--- a/rust/chains/hyperlane-cosmos/src/providers/mod.rs
+++ b/rust/chains/hyperlane-cosmos/src/providers/mod.rs
@@ -141,6 +141,30 @@ impl CosmosProvider {
             .map(|(a, n)| CosmosAddress::from_account_id(a).map(|a| (a.digest(), n)))??;
         Ok((sender, nonce))
     }
+
+    /// Extract contract address from transaction.
+    /// Assumes that there is only one `MsgExecuteContract` message in the transaction
+    fn contract(tx: &Tx) -> Result<H256, ChainCommunicationError> {
+        use cosmrs::proto::cosmwasm::wasm::v1::MsgExecuteContract as ProtoMsgExecuteContract;
+
+        let any = tx
+            .body
+            .messages
+            .iter()
+            .find(|a| a.type_url == "/cosmwasm.wasm.v1.MsgExecuteContract")
+            .ok_or_else(|| {
+                ChainCommunicationError::from_other_str("could not find contract execution message")
+            })?;
+        let proto = ProtoMsgExecuteContract::from_any(any).map_err(|e| {
+            ChainCommunicationError::from_other_str(&format!(
+                "could not decode contract execution message into proto: {}",
+                e,
+            ))
+        })?;
+        let msg = MsgExecuteContract::try_from(proto)?;
+        let contract = H256::try_from(CosmosAccountId::new(&msg.contract))?;
+        Ok(contract)
+    }
 }
 
 impl HyperlaneChain for CosmosProvider {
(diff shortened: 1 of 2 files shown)


rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -170,55 +219,27 @@ impl HyperlaneProvider for CosmosProvider {
...
 
-        // TODO add proper error handling
         let response = self
             .rpc_client
             .tx(tendermint_hash, false)
             .await
-            .map_err(|_| ChainCommunicationError::from_other_str("generic error"))?;
+            .map_err(|e| {
+                ChainCommunicationError::from_other_str(&format!("error on RPC request: {}", e))
+            })?;
```
daniel-savu, line 228: I believe the error instance would already have the information conveyed by the string, so could be cleaner to just
```suggestion
            .map_err(Into::<HyperlaneCosmosError>::into)?;
```

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -116,6 +116,55 @@ impl CosmosProvider {
...
+        let (sender, nonce) = tx
+            .auth_info
+            .fee
+            .payer
+            .as_ref()
+            .map(|payer| self.search_payer_in_signer_infos(&tx.auth_info.signer_infos, payer))
+            .map_or_else(
+                || {
+                    let signer_info = tx.auth_info.signer_infos.get(0).ok_or_else(|| {
+                        ChainCommunicationError::from_other_str("no signer info in default signer")
```
daniel-savu, line 135: We started using chain-specific error structs a while ago where possible. If this error is added as an item under `CosmosError`, we'd get the conversion to `ChainCommunicationError` for free, and could reuse above as well (on line 107)

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -116,6 +116,55 @@ impl CosmosProvider {
...
+            .find(|a| a.type_url == "/cosmwasm.wasm.v1.MsgExecuteContract")
+            .ok_or_else(|| {
+                ChainCommunicationError::from_other_str("could not find contract execution message")
+            })?;
+        let proto = ProtoMsgExecuteContract::from_any(any).map_err(|e| {
+            ChainCommunicationError::from_other_str(&format!(
+                "could not decode contract execution message into proto: {}",
+                e,
+            ))
+        })?;
```
daniel-savu, line 163: `Prost::DecodeError` is an item on `CosmosError`, so this can just be
```suggestion
        let proto = ProtoMsgExecuteContract::from_any(any)..map_err(Into::<HyperlaneCosmosError>::into)?;
```

rust/chains/hyperlane-cosmos/src/providers/mod.rs
```diff
@@ -116,6 +116,55 @@ impl CosmosProvider {
...
+                    self.convert_signer_info_into_account_id_and_nonce(signer_info)
+                },
+                |p| p,
+            )
+            .map(|(a, n)| CosmosAddress::from_account_id(a).map(|a| (a.digest(), n)))??;
+        Ok((sender, nonce))
+    }
+
+    /// Extract contract address from transaction.
+    /// Assumes that there is only one `MsgExecuteContract` message in the transaction
```
daniel-savu, line 146: When would there not be just one `MsgExecuteContract` message in a transaction? Thinking that we'll eventually batch submissions in the same tx and not sure if that affects this assumption - but this can be handled at that time

daniel-savu (approved):

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: Thank you. Fixed.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: Yes, the challenge is several messages which execute different contracts. Current implementation does not identify the type of contract. The message which is sent to contract (json) contains only method and arguments, but it does not contain the type of contract. So, if we have several messages, we need to filter out execution of unrelated contracts as well. There is a lot to think about.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: Thank you. Fixed.

rust/chains/hyperlane-cosmos/src/providers/mod.rs
ameten: Fixed.

PR #4410, opened 2024-08-30
chore: add `mainnet` suffix to solana and eclipse

daniel-savu:
### Description

Because we're doing a new core deployment that includes protocol fees


paulbalaji (commented):
once https://github.com/hyperlane-xyz/hyperlane-registry/pull/171 is merged, you should update the `.registryrc` and the agent configs too (remember to update your local registry to that commit before running the config update script)

typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -522,7 +522,7 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
     validators: ['0xfa33391ee38597cbeef72ccde8c9e13e01e78521'],
   },
 
-  solana: {
+  solanamainnet: {
```
paulbalaji, line 525: > /home/runner/work/hyperlane-monorepo/hyperlane-monorepo/typescript/sdk/src/consts/multisigIsm.ts
  530:3  error  Expected object keys to be in ascending order. 'solanadevnet' should be before 'solanamainnet'  sort-keys

paulbalaji (approved):
could you also fix the typo here pls
> ./rust/chains/hyperlane-cosmos/src/libs/address.rs:38: enconding ==> encoding

tkporter (approved):
@paulbalaji we didn't have an issues with fuse using the same domain ID right

paulbalaji:
> @paulbalaji we didn't have an issues with fuse using the same domain ID right

reusing domain ids is gucci

PR #4411, opened 2024-09-02
fix: increase wait time for postgres to be ready in cosmos e2e

paulbalaji:
## fix: increase wait time for postgres to be ready in cosmos e2e

increase time waiting for postgres to be ready, to reduce this sort of flake https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10665491633/job/29559169324?pr=4410#step:13:2998 

based on the logs it looks like 10s is just on the border, adding an extra 5s to give headroom

## drive-by fix: disable fail-fast on e2e-matrix

currently only the non-cosmwasm tests are required to merge, therefore we should not fail the entire matrix if the cosmos job fails. This also means we can test that the issues we're seeing with the `cosmwasm` job do not affect the `non-cosmwasm` one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,6 +187,7 @@ jobs:
     if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main' || github.base_ref == 'cli-2.0') || github.event_name == 'merge_group'
     needs: [yarn-build]
     strategy:
+      fail-fast: false
       matrix:
         e2e-type: [cosmwasm, non-cosmwasm]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -187,6 +187,7 @@ jobs:
     if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.base_ref == 'main' || github.base_ref == 'cli-2.0') || github.event_name == 'merge_group'
     needs: [yarn-build]
     strategy:
+      fail-fast: false
```
daniel-savu, line 190: I do think that the cosmos e2e test should be required in CI, just so we don't accidentally break deploy tooling or agent support

paulbalaji: I think we should still have `fail-fast: false` to let both tests run til completion, but promote `cosmwasm` to a required check in the repo settings

daniel-savu (approved):

PR #4413, opened 2024-09-02
feat: more ergonomic RPC URL setting

tkporter:
### Description

- Updates `scripts/secret-rpc-urls/set-rpc-urls.ts` to:
  - Interactively prompt for what RPC URLs to use
  - If desired, update K8s consumers (secrets & reset pods)

Usage:
```
yarn tsx ./scripts/secret-rpc-urls/set-rpc-urls.ts -e mainnet3 --chain lisk
```

or even
```
yarn tsx ./scripts/secret-rpc-urls/set-rpc-urls.ts -e mainnet3 --chains lisk zircuit
```

Removes `set-rpc-urls-from-registry` in favor of this new command, which lets you do this as well.

Note that atm it doesn't consider warp monitors as dependent k8s resources. This is because we haven't really defined infrastructure as code for these -- the deployments that exist are kinda manually created by passing in a config file. This should change but didn't feel appropriate in this PR

The meatiest changes are in utils/rpcUrls.ts and utils/k8s.ts, the rest were kinda drivebys to accommodate this

### Drive-by changes

- Various tweaks to labels used throughout k8s to be a bit more uniform
- Refactored a lot of helm related scripts to use `HelmManager` as a superclass


typescript/infra/scripts/secret-rpc-urls/set-rpc-urls.ts
```diff
@@ -1,22 +1,28 @@
...
+  assertCorrectKubeContext,
+  getArgs,
+  withChains,
+} from '../agent-utils.js';
+import { getEnvironmentConfig } from '../core-utils.js';
 
 async function main() {
-  const { environment, chain, rpcUrls } = await withRpcUrls(
-    withChainRequired(getArgs()),
+  const { environment, chains } = await withChains(getArgs()).alias(
```
Mo-Hussain, line 10: We could use `withChainsRequired`?

typescript/infra/scripts/warp-routes/deploy-warp-monitor.ts
```diff
@@ -15,11 +16,10 @@ async function main() {
     .string('filePath')
     .parse();
 
-  await runWarpRouteHelmCommand(
-    HelmCommand.InstallOrUpgrade,
-    'mainnet3',
-    filePath,
-  );
+  await assertCorrectKubeContext(getEnvironmentConfig('mainnet3'));
```
Mo-Hussain, line 19: nit: thought on a enum for this?

typescript/infra/src/utils/helm.ts
```diff
@@ -91,3 +92,97 @@ export function getDeployableHelmChartName(helmChartConfig: HelmChartConfig) {
...
+      );
+      return true;
+    } catch (error) {
+      return false;
+    }
+  }
+
+  async getExistingK8sSecrets(): Promise<string[]> {
+    const [output] = await execCmd(
+      `kubectl get secret --selector=app.kubernetes.io/instance=${this.helmReleaseName} -o jsonpath='{.items[*].metadata.name}' -n ${this.namespace}`,
```
Mo-Hussain, line 172: Could look at using https://www.npmjs.com/package/@kubernetes/client-node in the future

typescript/infra/src/utils/k8s.ts
```diff
@@ -0,0 +1,138 @@
...
+            `⏳ ${resourceNames.length - missing.length} of ${
+              resourceNames.length
+            } ${resourceType}s up, waiting for ${missing.length} more`,
+          );
+          throw new Error(
+            `${resourceType}s not ready, ${missing.length} missing`,
+          );
+        }
+      },
+      2000,
```
Mo-Hussain, line 79: nit: could we make these constants that describe what they do?

Mo-Hussain (approved):
Nice work!

typescript/infra/scripts/secret-rpc-urls/set-rpc-urls.ts
paulbalaji: `withChainsRequired` doesn't itself exist yet but would be glad if it did!

typescript/infra/scripts/warp-routes/deploy-warp-monitor.ts
paulbalaji: +1, would be useful in other places

typescript/infra/src/utils/rpcUrls.ts
```diff
@@ -50,28 +61,167 @@ export async function setAndVerifyRpcUrls(
...
+      });
+    }
+
+    const selection = await select({
+      message: 'Select RPC URL',
+      choices,
+      pageSize: 30,
+    });
+
+    if (selection === SystemChoice.DONE) {
```
paulbalaji, line 180: a "select all" would be nice so i don't have to manually select each registry one on new chain deploys

typescript/infra/src/utils/rpcUrls.ts
```diff
@@ -1,46 +1,57 @@
...
+      environment,
+      chain,
+    );
+    const newRpcUrls = await inputRpcUrls(chain, currentSecrets);
+    console.log(`Selected RPC URLs: ${formatRpcUrls(newRpcUrls)}\n`);
+
+    const secretPayload = JSON.stringify(newRpcUrls);
+    await confirmSetSecretsInteractive(environment, chain, secretPayload);
     await updateSecretAndDisablePrevious(environment, chain, secretPayload);
+    await refreshDependentK8sResourcesInteractive(
```
paulbalaji, line 51: nit: would be nice to have:
- `--no-rotate` or `--set-only` to skip the refresh step
- `--rotate-only` to skip the set/update secrets steps

typescript/infra/scripts/warp-routes/deploy-warp-monitor.ts
tkporter: Spent a bunch of time on this just to hit a bunch of circular deps in infra

Would be happy to add this but I'm gonna call it for now, a bit rabbithole-y without a huge payoff

typescript/infra/src/utils/helm.ts
tkporter: yeah would support this for sure

typescript/infra/src/utils/rpcUrls.ts
tkporter: you can skip it interactively with the Y/n prompt, are you saying a CLI flag would be preferable?

paulbalaji: yeah just as an additional option to save some reading/clicks, but not required

PR #4414, opened 2024-09-02
feat: solana <> eclipse with protocol fees

daniel-savu:
### Description

Artifacts from the core deployment, as well as the SOL warp route between Solana and Eclipse.


rust/sealevel/environments/mainnet3/igp/eclipsemainnet/default/igp-accounts.json
```diff
@@ -0,0 +1,5 @@
+{
+  "salt": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "igp_account": "ABb3i11z7wKoGCfeRQNQbVYWjAm7jG7HzZnDLV4RKRbK",
+  "overhead_igp_account": ""
```
tkporter, line 4: should this be `3Wp4qKkgf4tjXz1soGyTSndCgBPLZFSrZkiDZ8Qp9EEj`? don't recall how this tooling works tbh. ditto for the solanamainnet one, wondering if `overhead_igp_account` needs to be set

tkporter (approved):
🦅

rust/sealevel/environments/mainnet3/igp/eclipsemainnet/default/igp-accounts.json
daniel-savu: will add for completeness but I didn't run into any issues without it and it was also not mentioned in the runbook

PR #4415, opened 2024-09-02
fix: use known safe multisend addresses in getSafe

paulbalaji:
- fix: use known safe addresses in getSafe
- also requires updating the MultiSend governor helper to handle cases where there is no known `MultiSend` contract
- hardcodes the `safeDeploymentsVersions` mapping, as it's not exposed by the safe protocol-kit
- introduce utilities in the MultiSend to delete specific or all pending transactions
- add helper scripts for safe tx delete operations


typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -13,22 +17,61 @@ export function getSafeService(chain, multiProvider) {
...
+    released: true,
+  });
+  const multiSendCallOnly = getMultiSendCallOnlyDeployment({
+    version: multiSendCallOnlyVersion,
+    network: domainId,
+    released: true,
+  });
+
+  // Use the safe address for multiSendAddress and multiSendCallOnlyAddress
+  // if the contract is not deployed
```
jmrossy, line 64: Nit, feels like this comment should be a few lines lower above 70

typescript/infra/src/govern/multisend.ts
```diff
@@ -58,28 +60,204 @@ export class SafeMultiSend extends MultiSend {
...
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          safeTxHash: safeTxHash,
+          signature: signature,
+        }),
+      });
+
+      if (res.status === 204) {
```
jmrossy, line 173: It's definitely 204 every time and not any other success codes (e.g. 200)?

jmrossy (approved):

typescript/infra/src/govern/multisend.ts
paulbalaji: that's what the spec seems to suggest, but happy to loosen up to any success code image

typescript/infra/src/govern/multisend.ts
```diff
@@ -51,35 +56,72 @@ export class SafeMultiSend extends MultiSend {
...
+        safeTransactionData,
+      });
+      await this.proposeSafeTransaction(safeSdk, safeService, safeTransaction);
+    }
+  }
+
+  // Helper function to propose a multi-send transaction
+  private async proposeMultiSendTransaction(
+    calls: CallData[],
+    safeSdk: any,
```
tkporter, line 96: nit: do these need to be `any`? seems like they come from some fns that have types

typescript/infra/src/govern/multisend.ts
```diff
@@ -51,35 +56,72 @@ export class SafeMultiSend extends MultiSend {
...
-    const safeTxHash = await safeSdk.getTransactionHash(safeTransaction);
-    const senderSignature = await safeSdk.signTransactionHash(safeTxHash);
+    await this.proposeSafeTransaction(safeSdk, safeService, safeTransaction);
+  }
 
-    const senderAddress = await this.multiProvider.getSignerAddress(this.chain);
-    await safeService.proposeTransaction({
+  // Helper function to propose a safe transaction
+  private async proposeSafeTransaction(
+    safeSdk: any,
```
tkporter, line 113: ditto

typescript/infra/src/govern/multisend.ts
```diff
@@ -51,35 +56,72 @@ export class SafeMultiSend extends MultiSend {
...
+
+  // Helper function to propose individual transactions
+  private async proposeIndividualTransactions(
+    calls: CallData[],
+    safeSdk: any,
+    safeService: any,
+  ) {
+    for (const call of calls) {
+      const safeTransactionData = createSafeTransactionData({ call });
+      const safeTransaction = await createSafeTransaction({
```
tkporter, line 83: nit: afaict this is also done in `proposeMultiSendTransaction`, could maybe roll this into the shared fn too. Don't feel strongly

typescript/infra/src/utils/safe.ts
```diff
@@ -0,0 +1,226 @@
...
+import {
+  ChainNameOrId,
+  MultiProvider,
+  getSafe,
+  getSafeService,
+} from '@hyperlane-xyz/sdk';
+import { Address, CallData } from '@hyperlane-xyz/utils';
+
+export async function getSafeAndService({
+  chain,
```
tkporter, line 12: I don't feel too strongly but noting that this param pattern where we have an object and destructure deviates from the pattern we have elsewhere. Slight preference to keep things consistent but will defer to you if you feel strongly this is better

typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -13,22 +17,61 @@ export function getSafeService(chain, multiProvider) {
   return new SafeApiKit.default({ txServiceUrl, ethAdapter });
 }
 
+// Default safe version to use if not specified
+const DEFAULT_SAFE_VERSION = '1.3.0';
+const safeVersionOverrides = {
+  ancient8: '1.1.1',
```
tkporter, line 23: there's no way to get this via the SDK I presume? as this is used for starting to interact with it?

tkporter (approved):
very nice

typescript/infra/src/utils/safe.ts
paulbalaji: more of a style preference based on usage in the newer parts of the sdk/cli, but happy to defer to the current infra style

typescript/infra/src/govern/multisend.ts
paulbalaji: they iterate over slightly different things, so thought to keep them separate for readability

`proposeIndividualTransactions` - per call, creates txdata, proposes N txs
`proposeMultiSendTransaction` - creates tx data incl all calls, proposes 1 tx

paulbalaji:
I know it was only a nit but the types were bothering me so I did some wrangling to fix them. It required unifying the `protocol-kit` version we use across both SDK/Infra to be the same now, see [`8b15145` (#4415)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4415/commits/8b151450c4ef0154928c3a40470be6c000d75857)

Updating types for the safe service api-kit were mostly fine in [`8490247` (#4415)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4415/commits/849024741dc6771d202bbcacef9127a333eb94bd)

typescript/infra/src/utils/safe.ts
```diff
@@ -0,0 +1,226 @@
...
+        verifyingContract: safeAddress,
+      },
+      primaryType: 'DeleteRequest',
+      message: {
+        safeTxHash: safeTxHash,
+        totp: totp,
+      },
+    };
+
+    const signature = await (signer as ethers.Wallet)._signTypedData(
```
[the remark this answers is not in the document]
paulbalaji: although the docs suggest it should be available, the `ethers.Signer` interface under the hood does not have this method. did some digging and think it's probably safer to cast to `JsonRpcSigner` instead

typescript/sdk/src/utils/gnosisSafe.js
paulbalaji: there was not in the protocol-kit but after adding the types I found a way in the api-kit which seems to be working! see [`ef00a34` (#4415)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4415/commits/ef00a34559814d2e6ae332815d3fe958069aa367)

tkporter (approved):

PR #4419, opened 2024-09-04
feat: add the ability to remove a gas config from the sealevel IGP

tkporter:
### Description

Needed to disable relaying to Nautilus on Solana


tkporter:
just gonna get this in as it's innocuous and v2

PR #4420, opened 2024-09-04
chore: increase env-test timeout

paulbalaji:
chore: increase env-test timeout
- viction is borderline 4mins, so upping it to 5 for more headroom


paulbalaji:
done as part of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4415

PR #4422, opened 2024-09-04
chore: unify rust cache

paulbalaji:
chore: unify rust cache

- reduces the number of unique caches
- increases cache reuse
- greatly reduces over-aggressive cache eviction
- migrate e2e rust cache to buildjet
	- benefits: https://buildjet.com/for-github-actions/blog/launch-buildjet-cache#buildjet-cache-benefits
	- 20GB/week free cache is 2x what GHA provides

could also consider trying the drop-in npm setup/caching that buildjet could provide

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -212,8 +212,8 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v1-${{ runner.os }}-rust-cache"
-          shared-key: ${{ matrix.e2e-type }}
+          prefix-key: "v3-rust"
+          cache-all-crates: true
           workspaces: |
             ./rust
 
@@ -300,8 +300,8 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v1-${{ runner.os }}-rust-cache"
-          shared-key: "cli-e2e"
+          prefix-key: "v3-rust"
+          cache-all-crates: true
           workspaces: |
             ./rust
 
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -363,8 +365,9 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          prefix-key: "v1-${{ runner.os }}-rust-cache"
-          shared-key: "cli-e2e"
+          prefix-key: "v3"
```
daniel-savu, line 368: This these same prefix and shared key are identical for all steps, could we remove one of them and just use `rust-cache` for all?

.github/workflows/test.yml
```diff
@@ -363,8 +365,9 @@ jobs:
...
+          shared-key: "rust-cache"
+          cache-all-crates: true
```
daniel-savu, line 370: don't think this will do much based on their docs :/ (https://github.com/Swatinem/rust-cache). What it adds as caching of any CI-specific rust crates, of which I think we have none?

daniel-savu (commented):
couldn't find uses of buildjet yet, assuming that's why the PR is still a draft

.github/workflows/test.yml
paulbalaji: going to be closing PR because moving it all to the same cache had a negative effect

paulbalaji:
going to be closing PR because moving it all to the same cache had a negative effect

daniel-savu:
> going to be closing PR because moving it all to the same cache had a negative effect

@paulbalaji maybe because it exceeded the free storage cap?

PR #4424, opened 2024-09-04
feat: migrate large runners to buildjet

paulbalaji:
feat: migrate large runners to buildjet
- includes migration to buildjet's cache
- benefits: https://buildjet.com/for-github-actions/blog/launch-buildjet-cache#buildjet-cache-benefits
	- 20GB/week free cache is 2x what GHA provides
- drive-by: add restore-keys for the yarn caches
- more buildjet investigation notes [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4445)


.github/workflows/test.yml
```diff
@@ -365,6 +373,7 @@ jobs:
         with:
           prefix-key: "v1-${{ runner.os }}-rust-cache"
           shared-key: "cli-e2e"
+          save-if: ${{ !startsWith(github.ref, 'refs/heads/gh-readonly-queue') }}
```
tkporter, line 376: what does this mean?

.github/workflows/test.yml
```diff
@@ -214,6 +218,9 @@ jobs:
         with:
           prefix-key: "v1-${{ runner.os }}-rust-cache"
           shared-key: ${{ matrix.e2e-type }}
+          cache-all-crates: "true"
+          cache-provider: "buildjet"
```
tkporter, line 222: curious if this is needed in other places too? iiuc there are a bunch of different rust cache steps in this file and in rust.yml

tkporter:
do you see any speedup here?

.github/workflows/test.yml
paulbalaji: it means don't write to cache if we're in the merge queue (since no other branches can reuse it)

.github/workflows/test.yml
paulbalaji: not for now imo, the goal here is to have the rust caches split across providers - so with this PR:

buildjet:
- e2e-cosmwasm
- e2e-non-cosmwasm

gha:
- rust.yml test-rs
- rust.yml lint-rs
- cli-e2e

paulbalaji:
> do you see any speedup here?

we shouldn't see any speedup between a warm cache run between the two, as the runner is unchanged and both GHA/buildjet have similar cache performance on GH runners

where I do expect to see a speedup is that once merged, we *should* get fewer cache evictions thus less cache misses thus faster general build times

paulbalaji:

daniel-savu (approved):
very noice

PR #4425, opened 2024-09-04
feat(infra): use private RPCs for warp monitor service

Mo-Hussain:
### Description
- support private RPCs for warp monitor service

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConifg.ts
+++ b/typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConifg.ts
@@ -1,14 +1,13 @@
 import {
   ChainMap,
   IsmType,
-  RouterConfig,
   TokenRouterConfig,
   TokenType,
   buildAggregationIsmConfigs,
 } from '@hyperlane-xyz/sdk';
 import { symmetricDifference } from '@hyperlane-xyz/utils';
 
-import { getRegistry } from '../../chains.js';
+import { getRegistry as getMainnet3Registry } from '../../chains.js';
 
 const lockbox = '0xC8140dA31E6bCa19b287cC35531c2212763C2059';
 const xERC20 = '0x2416092f143378750bb29b79eD961ab195CcEea5';
@@ -118,7 +117,7 @@ const ezEthSafes: Record<string, string> = {
 export const getRenzoEZETHWarpConfig = async (): Promise<
   ChainMap<TokenRouterConfig>
 > => {
-  const registry = await getRegistry();
+  const registry = await getMainnet3Registry();
 
   const validatorDiff = symmetricDifference(
     new Set(chainsToDeploy),
--- a/typescript/infra/helm/warp-routes/templates/_helpers.tpl
+++ b/typescript/infra/helm/warp-routes/templates/_helpers.tpl
@@ -81,5 +81,5 @@ The warp-routes container
   - {{ .Values.environment}}
   envFrom:
   - secretRef:
-      name: warp-routes-env-var-secret
+      name: {{ include "hyperlane.fullname" . }}-secret
 {{- end }}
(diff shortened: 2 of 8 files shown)


typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -330,17 +343,13 @@ async function getXERC20Limits(
...
             }
+            default:
+              throw new Error(`Unsupported token type ${token.type}`);
           }
-          break;
         }
+        default:
+          throw new Error(`Unsupported protocol type ${token.protocolType}`);
       }
-      return {
```
Mo-Hussain, line 337: Never to default to zero values eve though was not the cause of the issue

typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConifg.ts
```diff
@@ -1,14 +1,13 @@
...
   IsmType,
-  RouterConfig,
   TokenRouterConfig,
   TokenType,
   buildAggregationIsmConfigs,
 } from '@hyperlane-xyz/sdk';
 import { symmetricDifference } from '@hyperlane-xyz/utils';
 
-import { getRegistry } from '../../chains.js';
+import { getRegistry as getMainnet3Registry } from '../../chains.js';
```
Mo-Hussain, line 10: drive by to make the intentions clearer

typescript/infra/helm/warp-routes/templates/_helpers.tpl
```diff
@@ -70,4 +79,9 @@ The warp-routes container
   - "10000"
   - -f
   - {{ .Values.configFilePath }}
+  - -e
```
Mo-Hussain, line 82: Could probably hard code this in the HelmManager

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -88,13 +92,21 @@ async function main(): Promise<boolean> {
     )
     .demandOption('filePath')
     .string('filePath')
+    .describe('environment', 'deploy environment')
```
tkporter, line 95: nit: think you could use the `getArgs` fn and get this for ffree

typescript/infra/helm/warp-routes/templates/env-var-external-secret.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: warp-routes-env-var-external-secret
```
tkporter, line 4: I think we should probably have the name of this ExternalSecret and the secret name (line 14) be dynamic based on the name of the helm chart / the pod. Otherwise we risk having the many different warp route helm releases try to control the same resources, and bad things can happen. E.g. even if helm doesn't complain about this when doing a deploy, if we end up uninstalling a warp route it'd get rid of this external secret and then break all the other ones that depend on it

I notice that we do this in the key funder helm chart where you probably got this from though - we kinda get away with it there bc key funder is intended to be a singleton (but tbh it's probably a bad practice for us to have hardcoded the name there anyways)

so i'd make this name something like
```
{{ include "hyperlane.fullname" . }}-external-secret
```

and the secret name `{{ include "hyperlane.fullname" . }}-secret` or something

typescript/infra/helm/warp-routes/templates/env-var-external-secret.yaml
```diff
@@ -0,0 +1,45 @@
...
+    template:
+      type: Opaque
+      metadata:
+        labels:
+          {{- include "hyperlane.labels" . | nindent 10 }}
+        annotations:
+          update-on-redeploy: "{{ now }}"
+      data:
+        GCP_SECRET_OVERRIDES_ENABLED: "true"
+        GCP_SECRET_OVERRIDE_HYPERLANE_{{ .Values.hyperlane.runEnv | upper }}_KEY_DEPLOYER: {{ print "'{{ .deployer_key | toString }}'" }}
```
tkporter, line 24: just calling out that we don't necessarily need the deployer here, don't feel too strongly about it being here tbh though

tkporter (approved):
would like to see the externalsecret / secret names change, but otherwise lgtm! thanks for doing this

PR #4429, opened 2024-09-04
feat: cleanup caches for merged prs

paulbalaji:
- feat: cleanup caches for merged prs
- force delete based on the sample workflow https://github.com/actions/cache/blob/main/tips-and-workarounds.md#force-deletion-of-caches-overriding-default-cache-eviction-policy

see it in action here: https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10705157657/job/29679857473?pr=4429#step:3:30

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -32,11 +32,12 @@ jobs:
 
           REPO="${{ github.repository }}"
 
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            PR_NUMBER="${{ github.event.pull_request.number }}"
-          else
-            PR_NUMBER="${{ github.event.inputs.pr_number }}"
-          fi
+          PR_NUMBER=4422
+          # if [[ "${{ github.event_name }}" == "pull_request" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-cache.yml
```diff
@@ -0,0 +1,52 @@
+# Copied from https://github.com/actions/cache/blob/main/tips-and-workarounds.md#force-deletion-of-caches-overriding-default-cache-eviction-policy
+name: Cleanup Caches for Merged PRs
+
+on:
+  pull_request:
+    types: [closed, enqueued]
```
tkporter, line 6: tbh I'm not certain, I assume we still run CI for PRs in the merge queue and that's why it takes so long? in which case does it make sense to clean up the cache when enqueuing a PR?

paulbalaji: it's just how GHA caches work, that the merge queue cache is derived from the target branch, and is separate to that of the branch being merged

paulbalaji:
probs not required after moving to buildjet https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4424

PR #4435, opened 2024-09-05
feat: implement HyperlaneHaasGovernor

paulbalaji:
feat: implement HyperlaneHaasGovernor
- allows one-time governance of actions resulting from both ICA+Core checkers
- updates check-deploy to support the `-m haas` module
- also fixes a bug to do with submitting calls to the current SAFE owner when a new ICA owner has just been configured

example:
image
image


typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -80,6 +80,9 @@ export const ethereumChainOwners: ChainMap<OwnableConfig> = Object.fromEntries(
           validatorAnnounce: DEPLOYER, // unused
           testRecipient: DEPLOYER,
           fallbackRoutingHook: DEPLOYER,
+          // Ensure we know keep the Safe and ICA addresses for each chain somewhere in the config
+          safeAddress: safes[local],
```
tkporter, line 84: nit: agree plumbing this through here is the easiest, and also agree it's a bit of a hack 😄

I don't feel too strongly but wondering if it's easy to change some part of this to understand that these two keys are different from the others?

Like it'd be nice to have ownerOverrides.contractOverrides.foo or something. Or maybe some naming schema to make it even more clear that safeAddress is special? like `_safeAddress` or something

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -80,6 +80,9 @@ export const ethereumChainOwners: ChainMap<OwnableConfig> = Object.fromEntries(
           validatorAnnounce: DEPLOYER, // unused
           testRecipient: DEPLOYER,
           fallbackRoutingHook: DEPLOYER,
+          // Ensure we know keep the Safe and ICA addresses for each chain somewhere in the config
```
tkporter, line 83: nit: slight grammar mistake but could you also describe the problem for why we needed this

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -361,7 +374,8 @@ export abstract class HyperlaneAppGovernor<
     }
 
     // 2. Check if the call will succeed via Gnosis Safe.
-    const safeAddress = this.checker.configMap[chain].owner;
+    const safeAddress =
+      this.checker.configMap[chain].ownerOverrides?.safeAddress;
```
tkporter, line 378: one edge case I could see us hitting is that there may not always be a single Safe address on a given chain

We already have this today to some degree (we have some "old Safes" you can see in Notion that still own some contracts, though they generally own v2)

But I think in the not so far future we'll have multiple safes -- e.g. how there's gonna be a separate Safe that will own the Proxy Admin. In this situation we'll have this new Safe set as the `proxyAdmin` owner in the ownerOverrides

So wondering if we should check this plumbed in safeAddress in addition to checking `configMap[chain].owner;` instead of in lieu

typescript/infra/src/govern/HyperlaneHaasGovernor.ts
```diff
@@ -0,0 +1,100 @@
...
+  async checkChain(chain: ChainName) {
+    await this.icaChecker.checkChain(chain);
+    await this.coreChecker.checkChain(chain);
+  }
+
+  getCheckerViolations() {
+    return [...this.icaChecker.violations, ...this.coreChecker.violations];
+  }
+
+  async govern(confirm = true, chain?: ChainName) {
```
tkporter, line 82: didn't you say some filtering for dupes was needed? not sure i see it

tkporter (approved):
love it, some suggestions

typescript/infra/src/govern/HyperlaneHaasGovernor.ts
paulbalaji: yup, this was added to the base class https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4435/files#diff-07e7a8ab698661948a9707a204e898b70a3bba0c6d1fe64d8d633bb968ee5bb0R186-R194

typescript/infra/config/environments/mainnet3/owners.ts
paulbalaji: to resolve this + the multiple safe edge case, I think we could update the schema we use here to something like

ownerOverrides.safe.owner 
ownerOverrides.ica.owner

and then we could have variations for individual contracts e.g.

ownerOverrides.safe.proxyAdmin
ownerOverrides.ica.proxyAdmin
etc

imo there needs to be a bit more of a rework to support this multi-safe world so the above proposal could be a stopgap til we figure out how to do it better

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -413,11 +427,10 @@ export abstract class HyperlaneAppGovernor<
...
 
-    return {
-      type: SubmissionType.MANUAL,
-      chain,
-      call,
-    };
+    // If we're not able to make the call from a signer or Safe, we try
+    // to infer if it must be sent from an ICA controlled by a remote owner.
+    // This new inferred call will be unchanged if the call is not an ICA call after all.
+    return this.inferICAEncodedSubmissionType(chain, call);
```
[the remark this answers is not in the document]
paulbalaji: not sure but can look into it?

typescript/infra/src/govern/HyperlaneAppGovernor.ts
paulbalaji: answered in a diff comment

typescript/infra/src/govern/HyperlaneHaasGovernor.ts
```diff
@@ -0,0 +1,100 @@
...
+} from '@hyperlane-xyz/sdk';
+
+import {
+  AnnotatedCallData,
+  HyperlaneAppGovernor,
+} from './HyperlaneAppGovernor.js';
+import { HyperlaneCoreGovernor } from './HyperlaneCoreGovernor.js';
+import { ProxiedRouterGovernor } from './ProxiedRouterGovernor.js';
+
+export class HyperlaneHaasGovernor extends HyperlaneAppGovernor<any, any> {
```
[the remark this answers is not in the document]
paulbalaji: think there was a reason I didn't but I didn't try too hard, WTAL

typescript/infra/src/govern/HyperlaneAppGovernor.ts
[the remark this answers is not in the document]
paulbalaji: fixed

PR #4436, opened 2024-09-05
feat(sdk): Sort ISM config arrays

Mo-Hussain:
### Description

- Introduce sorting validator arrays and modules by type for ISM in the `normalizeConfig` method
- This will avoid deploying a new version of the same ISM (in practice) for updates
- Sort staticAddressSets by value before deploying
- This will enable us to use ISM recovery
- This will also avoid raising ISM mismatch violations for our checker tooling

### Testing

Unit Tests

### DIFF
--- a/typescript/sdk/src/deploy/EvmModuleDeployer.ts
+++ b/typescript/sdk/src/deploy/EvmModuleDeployer.ts
@@ -264,35 +264,40 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
     threshold?: number;
     multiProvider: MultiProvider;
   }): Promise<Address> {
-    values = values.sort();
+    const sortedValues = [...values].sort();
+
     const address = await factory['getAddress(address[],uint8)'](
-      values,
+      sortedValues,
       threshold,
     );
     const code = await multiProvider.getProvider(chain).getCode(address);
     if (code === '0x') {
       logger.debug(
-        `Deploying new ${threshold} of ${values.length} address set to ${chain}`,
+        `Deploying new ${threshold} of ${sortedValues.length} address set to ${chain}`,
       );
       const overrides = multiProvider.getTransactionOverrides(chain);
 
       // estimate gas
       const estimatedGas = await factory.estimateGas['deploy(address[],uint8)'](
-        values,
+        sortedValues,
         threshold,
         overrides,
       );
 
       // add 10% buffer
-      const hash = await factory['deploy(address[],uint8)'](values, threshold, {
-        ...overrides,
-        gasLimit: estimatedGas.add(estimatedGas.div(10)), // 10% buffer
-      });
+      const hash = await factory['deploy(address[],uint8)'](
+        sortedValues,
+        threshold,
+        {
+          ...overrides,
+          gasLimit: estimatedGas.add(estimatedGas.div(10)), // 10% buffer
+        },
+      );
 
       await multiProvider.handleTx(chain, hash);
     } else {
       logger.debug(
-        `Recovered ${threshold} of ${values.length} address set on ${chain}: ${address}`,
+        `Recovered ${threshold} of ${sortedValues.length} address set on ${chain}: ${address}`,
       );
     }
 
--- a/typescript/sdk/src/deploy/EvmModuleDeployer.ts
+++ b/typescript/sdk/src/deploy/EvmModuleDeployer.ts
@@ -264,6 +264,7 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
     threshold?: number;
     multiProvider: MultiProvider;
   }): Promise<Address> {
+    values = values.sort();
     const address = await factory['getAddress(address[],uint8)'](
       values,
       threshold,
--- a/typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
+++ b/typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
@@ -4,7 +4,7 @@ import { expect } from 'chai';
 import { Signer } from 'ethers';
 import hre from 'hardhat';
 
-import { Address, eqAddress, normalizeConfig } from '@hyperlane-xyz/utils';
+import { Address, eqAddress } from '@hyperlane-xyz/utils';
 
 import { TestChainName, testChains } from '../consts/testChains.js';
 import { HyperlaneAddresses, HyperlaneContracts } from '../contracts/types.js';
@@ -13,6 +13,7 @@ import { HyperlaneProxyFactoryDeployer } from '../deploy/HyperlaneProxyFactoryDe
 import { ProxyFactoryFactories } from '../deploy/contracts.js';
 import { MultiProvider } from '../providers/MultiProvider.js';
 import { randomAddress, randomInt } from '../test/testUtils.js';
+import { normalizeConfig } from '../utils/ism.js';
 
 import { EvmIsmModule } from './EvmIsmModule.js';
 import { HyperlaneIsmFactory } from './HyperlaneIsmFactory.js';
--- a/typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
+++ b/typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
@@ -4,7 +4,12 @@ import { expect } from 'chai';
 import { Signer } from 'ethers';
 import hre from 'hardhat';
 
-import { Address, eqAddress, normalizeConfig } from '@hyperlane-xyz/utils';
+import {
+  Address,
+  eqAddress,
+  normalizeConfig,
+  sortValidatorsInConfig,
+} from '@hyperlane-xyz/utils';
 
 import { TestChainName, testChains } from '../consts/testChains.js';
 import { HyperlaneAddresses, HyperlaneContracts } from '../contracts/types.js';
@@ -170,9 +175,18 @@ describe('EvmIsmModule', async () => {
 
   // expect that the ISM matches the config after all tests
   afterEach(async () => {
-    const normalizedDerivedConfig = normalizeConfig(await testIsm.read());
+    const derivedConfiig = await testIsm.read();
+    const normalizedDerivedConfig = normalizeConfig(derivedConfiig);
+    const sortedNormalizedDerivedConfig = sortValidatorsInConfig(
+      normalizedDerivedConfig,
+    );
     const normalizedConfig = normalizeConfig(testConfig);
-    assert.deepStrictEqual(normalizedDerivedConfig, normalizedConfig);
+    const sortedNormalizedConfig = sortValidatorsInConfig(normalizedConfig);
+
+    assert.deepStrictEqual(
+      sortedNormalizedDerivedConfig,
+      sortedNormalizedConfig,
+    );
   });
 
   // create a new ISM and verify that it matches the config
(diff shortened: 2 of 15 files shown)


typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -429,5 +505,99 @@ describe('EvmIsmModule', async () => {
           .true;
       });
     }
+
+    it(`reordering modules in an existing staticAggregationIsm should not trigger a redeployment`, async () => {
```
Mo-Hussain, line 509: @paulbalaji - should we expect this test to pass? it is not atm. I think this is a limitation of the sdk where a change in the order of the modules of the `staticAggregationIsm` with lead to redeployment.

One thing we can to do to mitigate this is sort the config by sorting the domains array by the ism type. This will work if there are not duplicate types an ism in the `staticAggregationIsm`. This should cover most cases

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -445,7 +450,7 @@ export class EvmIsmModule extends HyperlaneModule<
     const address = await EvmModuleDeployer.deployStaticAddressSet({
       chain: this.chain,
       factory: this.factories[factoryName],
-      values: config.validators,
+      values: [...config.validators].sort(),
```
Mo-Hussain, line 453: Brings us in line with the `IsmFactory`

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -146,10 +147,14 @@ export class EvmIsmModule extends HyperlaneModule<
 
     // save current config for comparison
     // normalize the config to ensure it's in a consistent format for comparison
-    const currentConfig = normalizeConfig(await this.read());
+    const currentConfig = sortValidatorsInConfig(
```
Mo-Hussain, line 150: Normalize and sort before doing the comparison

typescript/utils/src/objects.ts
```diff
@@ -175,3 +175,27 @@ export function normalizeConfig(obj: WithAddress<any>): any {
 
   return obj;
 }
+
+// write a function that will go through an object and sort any arrays it finds
+export function sortValidatorsInConfig(config: any): any {
```
Mo-Hussain, line 180: @paulbalaji - we could extend this to sort modules in `staticAggregationIsm` by the ISM type

paulbalaji:
so my initial thought was why not just sort at the source when reading instead of having to iterate over the object again:
- ism `validators` <https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/74a592e5867a375f1705654c1112d7c9ac784ddf/typescript/sdk/src/ism/EvmIsmReader.ts#L210>
- ism `modules` <https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/74a592e5867a375f1705654c1112d7c9ac784ddf/typescript/sdk/src/ism/EvmIsmReader.ts#L189>
- hook 'hooks'  <https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/74a592e5867a375f1705654c1112d7c9ac784ddf/typescript/sdk/src/hook/EvmHookReader.ts#L193>

here we can sort `modules` and `hooks` the address property, HOWEVER, we'd still have to iterate in the normalizeConfig function and it wouldn't help us with the config we are comparing the one onchain to because the raw config wouldn't have module/hook addresses for us to sort things with

so probably the bet course of action is to combine the sorting into the normalizeConfig function directly. I think it also means if we sort in the normalization we don't have to actually sort the array before deploying, but I would probably still leave that sort in there to make it easy to recover any static address sets that we deploy. to that end, this sort should go directly in the static address set deploy fn

so overall I'd say:
1. roll the `sortValidatorsInConfig` logic directly into the `normalizeConfig` function, and also sort `hooks` and `modules` with the same logic
2. avoid calling `.sort()` directly in any of the modules, instead do a sort on `values` directly in `deployStaticAddressSet` <https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/74a592e5867a375f1705654c1112d7c9ac784ddf/typescript/sdk/src/deploy/EvmModuleDeployer.ts#L242>

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
Mo-Hussain: @paulbalaji - can you look at this test! Should it pass in your opinion?

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -36,19 +36,27 @@ const randomMultisigIsmConfig = (m: number, n: number): MultisigIsmConfig => {
...
 
-const randomIsmConfig = (depth = 0, maxDepth = 2): IsmConfig => {
+function randomNonNestedModuleType(): ModuleType {
+  return NonNestedModuleTypes[randomInt(NonNestedModuleTypes.length)];
+}
+
+const randomIsmConfig = (depth = 0, maxDepth = 2) => {
   const moduleType =
-    depth == maxDepth ? ModuleType.MERKLE_ROOT_MULTISIG : randomModuleType();
+    depth === maxDepth ? randomNonNestedModuleType() : randomModuleType();
```
Mo-Hussain, line 58: If max depth is reached, return a random non nested ISM type to cap depth

typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -254,6 +254,7 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
     threshold?: number;
     multiProvider: MultiProvider;
   }): Promise<Address> {
+    values = values.sort();
```
paulbalaji, line 257: would copy what the ism factory does, as `.sort()` is in-place
```suggestion
    values = [...values].sort();
```

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -170,8 +189,10 @@ describe('EvmIsmModule', async () => {
 
   // expect that the ISM matches the config after all tests
   afterEach(async () => {
-    const normalizedDerivedConfig = normalizeConfig(await testIsm.read());
+    const derivedConfiig = await testIsm.read();
+    const normalizedDerivedConfig = normalizeConfig(derivedConfiig);
```
paulbalaji, line 193:
```suggestion
    const normalizedDerivedConfig = normalizeConfig(derivedConfig);
```

typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts
```diff
@@ -65,10 +73,21 @@ const randomIsmConfig = (depth = 0, maxDepth = 2): IsmConfig => {
...
+        let moduleType;
+
+        // Ensure that we do not add the same module type more than once per level
+        do {
+          moduleConfig = randomIsmConfig(depth + 1, maxDepth);
+          moduleType = moduleConfig.type;
+        } while (moduleTypes.has(moduleType));
+
+        moduleTypes.add(moduleType);
+        return moduleConfig;
```
paulbalaji, line 89: maybe `moduleType` could be an optional arg for `randomIsmConfig`, so we could iterate through each moduletype instead of the do-while

typescript/utils/src/objects.ts
```diff
@@ -175,3 +179,35 @@ export function normalizeConfig(obj: WithAddress<any>): any {
...
+  // Check if it's an object and not null
+  else if (typeof config === 'object' && config !== null) {
+    const sortedConfig: any = {};
+    for (const key in config) {
+      if (key === 'validators' && Array.isArray(config[key])) {
+        // Sort the validators array in lexicographical order (since they're already lowercase)
+        sortedConfig[key] = config[key].sort();
+      }
+      // if key is modules, sort the objects in the array by their 'type' property
+      else if (key === 'modules' && Array.isArray(config[key])) {
```
paulbalaji, line 198: what about hooks in the hookconfig? https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/74a592e5867a375f1705654c1112d7c9ac784ddf/typescript/sdk/src/hook/EvmHookReader.ts#L193

typescript/sdk/src/deploy/EvmModuleDeployer.ts
jmrossy: One of the worst design decisions in Javascript's history. I see people make this mistake very often

typescript/utils/src/objects.ts
```diff
@@ -175,3 +179,35 @@ export function normalizeConfig(obj: WithAddress<any>): any {
 
   return obj;
 }
+
+// write a function that will go through an object and sort any arrays it finds
+export function sortValidatorsAndModulesInConfig(config: any): any {
```
jmrossy, line 184: This file (and this whole package mostly) is for utilities agnostic to a particular concept or use case IMHO. So if you'd like to include this here, consider generalizing it to not be so specific to `validatorsAndModules`. Otherwise move it to the SDK

typescript/utils/src/objects.ts
```diff
@@ -175,3 +179,35 @@ export function normalizeConfig(obj: WithAddress<any>): any {
...
+  if (Array.isArray(config)) {
+    return config.map(sortValidatorsAndModulesInConfig);
+  }
+  // Check if it's an object and not null
+  else if (typeof config === 'object' && config !== null) {
+    const sortedConfig: any = {};
+    for (const key in config) {
+      if (key === 'validators' && Array.isArray(config[key])) {
+        // Sort the validators array in lexicographical order (since they're already lowercase)
+        sortedConfig[key] = config[key].sort();
```
jmrossy, line 195: Ditto as above, sort will modify the provided param here

typescript/utils/src/objects.ts
```diff
@@ -175,3 +179,35 @@ export function normalizeConfig(obj: WithAddress<any>): any {
...
+      }
+      // if key is modules, sort the objects in the array by their 'type' property
+      else if (key === 'modules' && Array.isArray(config[key])) {
+        sortedConfig[key] = config[key].sort((a: any, b: any) => {
```
jmrossy, line 199: ditto

typescript/sdk/src/deploy/EvmModuleDeployer.ts
Mo-Hussain: Sloppy of me, especially given that I'd done it the right way in a previous commit

typescript/utils/src/objects.ts
Mo-Hussain: @paulbalaji - I'm going to move `normalizeConfig` to the SDK as well given that it references an address

typescript/sdk/src/utils/ism.ts
```diff
@@ -11,3 +13,61 @@ export function multisigIsmVerificationCost(m: number, n: number): number {
   // @ts-ignore
   return multisigIsmVerifyCosts[`${n}`][`${m}`];
 }
+
+// Function to recursively remove 'address' properties and lowercase string properties
+export function normalizeConfig(obj: WithAddress<any>): any {
+  return sortValidatorsAndModulesInConfig(loweCaseConfig(obj));
+}
+
+function loweCaseConfig(obj: any): any {
```
paulbalaji, line 22: and also fix where it's called
```suggestion
function lowerCaseConfig(obj: any): any {
```

typescript/sdk/src/utils/ism.ts
```diff
@@ -11,3 +13,61 @@ export function multisigIsmVerificationCost(m: number, n: number): number {
...
+    return newObj;
+  } else if (typeof obj === 'string') {
+    return obj.toLowerCase();
+  }
+
+  return obj;
+}
+
+// write a function that will go through an object and sort any arrays it finds
+export function sortValidatorsAndModulesInConfig(config: any): any {
```
paulbalaji, line 41: nit maybe something like sortArraysInConfig, in case have to add more cases in the future
```suggestion
export function sortValidatorsAndModulesInConfig(config: any): any {
```

paulbalaji (approved):

jmrossy (approved):

PR #4437, opened 2024-09-05
fix: add fix for checking for correct messageId in `OPL2ToL1Ism` and `ArbL2ToL1Ism`  via external call

aroralanuk:
### Description

- In agreement with Chainlight team's recommendation, added a second isVerified() check after portal call to make sure an arbitrary call which passes the check for metadata length and messageId cannot be verified without calling is verifyMessageId(messageId) in `OPL2ToL1Ism` and `ArbL2ToL1`

### Drive-by changes

None

### Related issues

- fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/2

### Backward compatibility

No, but the contracts aren't deployed anywhere

### Testing

Unit testing


solidity/contracts/isms/hook/OPL2ToL1Ism.sol
```diff
@@ -66,10 +66,10 @@ contract OPL2ToL1Ism is
         bytes calldata metadata,
         bytes calldata message
     ) external override returns (bool) {
-        bool verified = isVerified(message);
```
yorhodes, line 69: I know this is what they recommended but I do not understand
I think once you add the `isAuthorized` check to the `_verifyWithPortalCall`, you will prevent anything other than a withdrawal tx where the data is a `verifyMessageId` call from the authorized hook

yorhodes (changes requested):

solidity/contracts/isms/hook/OPL2ToL1Ism.sol
aroralanuk: the `_crossChainSender()` which you use for determining if the sender is the l2Hook uses `xDomainMsgSender` here https://github.com/ethereum-optimism/optimism/blob/d887cfa990a39b30bb102e480661a1f09e0add67/packages/contracts-bedrock/src/universal/CrossDomainMessenger.sol#L122. This is reset once to 0x000....dead after the finalizeWithdrawalTransaction call is over. You can check the test here: https://github.com/ethereum-optimism/optimism/blob/d887cfa990a39b30bb102e480661a1f09e0add67/packages/contracts-bedrock/test/L1/L1CrossDomainMessenger.t.sol#L211.  

This means we cannot use isAuthorized in the `_verifyWithPortalCall` call and therefore need a second `isVerified` check

yorhodes (approved):

PR #4441, opened 2024-09-05
feat: improve Router enrollment with Checker/Governor

paulbalaji:
feat: improve Router enrollment with Checker/Governor
- aggregate violations into a single RouterViolation with the diff in router enrollments
- reduces number of txs to do the enrollment
- now supports ICA router enrollment of chains that we haven't enrolled
	- previously we would only iterate over the set of routers already onchain, and no attention was paid to new domains

example call for `ica`:
image

example call for `haas`:
image


typescript/infra/src/govern/ProxiedRouterGovernor.ts
```diff
@@ -54,19 +56,29 @@ export class ProxiedRouterGovernor<
...
   protected handleEnrolledRouterViolation(violation: RouterViolation) {
-    const remoteDomain = this.checker.multiProvider.getDomainId(
-      violation.remoteChain,
-    );
+    const domains: number[] = [];
+    const addresses: string[] = [];
+
+    for (const [remoteChain, expectedRouter] of Object.entries(
+      violation.routerDiff,
+    ) as [ChainName, Address][]) {
```
tkporter, line 64: nit: is the `as` needed? think the types should be fine

typescript/sdk/src/middleware/account/InterchainAccount.ts
```diff
@@ -31,6 +31,12 @@ export class InterchainAccount extends RouterApp<InterchainAccountFactories> {
     super(contractsMap, multiProvider);
   }
 
+  override async remoteChains(chainName: string): Promise<ChainName[]> {
```
tkporter, line 34: do you think this change should also be applied for `RouterApp`? I imagine other RouterApps will want to benefit from this. Feels a little weird to have InterchainAccount.remoteChains use the contractsMap but other RouterApps to use what's onchain

Or maybe if reading onchain is necessary to stay the same in RouterApp, there should be a RouterApp fn for reading it onchain and a different one for reading it from the contractsMap?

typescript/infra/src/govern/ProxiedRouterGovernor.ts
```diff
@@ -54,19 +56,29 @@ export class ProxiedRouterGovernor<
...
         to: violation.contract.address,
         data: violation.contract.interface.encodeFunctionData(
-          'enrollRemoteRouter',
-          [remoteDomain, violation.expected],
+          'enrollRemoteRouters',
+          [domains, addresses],
         ),
         value: BigNumber.from(0),
-        description: `Enroll router for remote chain ${violation.remoteChain} (${remoteDomain}) ${violation.expected} in ${violation.contract.address}`,
+        description: `Enroll missing routers for ${
```
tkporter, line 79: nit: iiuc this code path is also intended for when the routers change for whatever reason, not necessarily that routers are missing. I think this wording may be misleading for that situation, and it may be nice to surface the old router (though feel less strongly there, as we didn't do this before)

tkporter (approved):
very nice, some suggestions

typescript/infra/src/govern/ProxiedRouterGovernor.ts
paulbalaji: resolved

typescript/infra/src/govern/ProxiedRouterGovernor.ts
paulbalaji: fixed

typescript/sdk/src/router/RouterApps.ts
```diff
@@ -45,14 +45,14 @@ export abstract class RouterApp<
   // check onchain for remote enrollments
   override async remoteChains(chainName: string): Promise<ChainName[]> {
     const router = this.router(this.contractsMap[chainName]);
-    const domainIds = (await router.domains()).map((domain) => {
+    const onchainRemoteChainNames = (await router.domains()).map((domain) => {
```
jmrossy, line 48: 👍

typescript/sdk/src/router/types.ts
```diff
@@ -56,10 +57,11 @@ export enum RouterViolationType {
...
 export interface RouterViolation extends CheckerViolation {
   type: RouterViolationType.EnrolledRouter;
-  remoteChain: string;
   contract: Router;
-  actual: string;
-  expected: string;
+  routerDiff: ChainMap<{
+    actual: AddressBytes32;
+    expected: AddressBytes32;
+  }>;
```
jmrossy, line 64: This is fine but would it be better to decode the address from bytes to string to make it easier for consumers of these violations to work with?

paulbalaji: I think it's better to leave as-is here, so the governor (which works with this) doesn't have to do any translation from the values that its respective checker provides

PR #4442, opened 2024-09-05
chore: update scroll reorgperiod in agents config

daniel-savu:


tkporter (approved):
think you'll wanna update the .registryrc post registry commit too

PR #4443, opened 2024-09-05
chore: Update agent config

yorhodes:


cmcewen (approved):

paulbalaji (commented):
you should update `.registryrc` to `1d43e33fc84f486d0edf20a9e573f914e53fe94c` ~~then checkout that commit locally and rerun the update-agent script for mainnet3 & testnet4~~

otherwise agent-config job will fail

paulbalaji (approved):

PR #4445, opened 2024-09-05
feat: buildjet investigation

paulbalaji:
TL;DR:
- if we move to buildjet's 8vcpu runner, we can drop agent image build times from ~28mins to ~7mins
- cold-start e2e also drops from 24->15 mins (cosmwasm) and 33->22 mins (non-cosmwasm)
- warm e2e times 8->9 mins (cosmwasm) and 12->14 mins (non-cosmwasm)
    - but this is with cache still on GHA, moving runner+cache here to buildjet would save another 3 mins on both warm runs by my estimate
    - there is about a 20% reduction in the time taken for the actual agent tests step itself:
        - cosmwasm: 4m28 -> 3m40
        - non-cosmwasm: 9m27 -> 7m20

currently just a draft but some interesting observations:

1. currently cache has no impact on build time
	- rerunning builds for a commit or pushing empty commits should in theory lead to quicker builds with layer caching, but the main problem is that the `RUN` step of building the agents is not cached by the docker action
	- only RUNs with bind mounts get a checksum computed [(reference)](https://docs.docker.com/build/cache/invalidation/#general-rules)
	- disabling docker build caching could help save a bunch of cache space
3. migrating to `larger-runner` gives us 2x speedup (27->13m)
	- https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10730492851/job/29759029297?pr=4445
4. migrating to buildjet's larger runner equivalent (`buildjet-8vcpu-ubuntu-2204`) gives us 4x speedup (27->7m)
	- 2x speedup over `larger-runner`! (13->7m)
	- https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10731469866/job/29761608654?pr=4445
5. increasing to 32vcpu has no meaningful impact (7->6m)
	- https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10730867566/job/29760017932?pr=4445
6. decreasing to 4vcpu still takes less time than the default `larger-runner` (13->11m)
	- https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10730933793/job/29760194597?pr=4445

### DIFF
--- a/.github/workflows/agent-release-artifacts.yml
+++ b/.github/workflows/agent-release-artifacts.yml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   prepare:
-    runs-on: larger-runner
+    runs-on: ubuntu-latest
     outputs:
       tag_date: ${{ steps.taggen.outputs.TAG_DATE }}
       tag_sha: ${{ steps.taggen.outputs.TAG_SHA }}
--- a/.github/workflows/rust-docker.yml
+++ b/.github/workflows/rust-docker.yml
@@ -27,7 +27,7 @@ jobs:
         run: echo "::set-output name=defined::true"
 
   build-and-push-to-gcr:
-    runs-on: buildjet-32vcpu-ubuntu-2204
+    runs-on: buildjet-4vcpu-ubuntu-2204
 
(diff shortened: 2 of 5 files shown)


paulbalaji:

PR #4447, opened 2024-09-06
chore: tweaks to enroll non-evm warp routes

daniel-savu:


typescript/cli/src/deploy/utils.ts
```diff
@@ -55,12 +55,13 @@ export async function runPreflightChecksForChains({
   assertSigner(signer);
   logGreen('✅ Signer is valid');
 
-  await nativeBalancesAreSufficient(
-    multiProvider,
-    signer,
-    chainsToGasCheck ?? chains,
-    minGas,
-  );
+  // TODO: this should skip checks on non-evm chains
```
jmrossy, line 58: Why should we skip on non-evm?

daniel-savu: because the scripts just throw errors for non-evm chains - so whoever runs this script for non-evm chains should be warned that there's fewer chains, but still have the option to go ahead

typescript/cli/src/utils/balances.ts
```diff
@@ -14,12 +14,7 @@ export async function nativeBalancesAreSufficient(
 ) {
   const sufficientBalances: boolean[] = [];
   for (const chain of chains) {
-    // Only Ethereum chains are supported
-    if (multiProvider.getProtocol(chain) !== ProtocolType.Ethereum) {
-      logGray(`Skipping balance check for non-EVM chain: ${chain}`);
-      continue;
-    }
-    const address = multiProvider.getSigner(chain).getAddress();
```
manu-flint, line 22: Hey @daniel-savu 
I am trying to build the cli from source but due to line 22 being removed in this change, "address" becomes undefined and later in line 25 and 31 "address" is read. Resulting in a failed build, also there are a few unused variables which are logged during the build. 
Can you please find time to check it once, thank you in advance.
image

typescript/cli/src/deploy/utils.ts
```diff
@@ -55,12 +55,16 @@ export async function runPreflightChecksForChains({
   }
   logGreen('✅ Chains are valid');
 
-  await nativeBalancesAreSufficient(
-    multiProvider,
-    chainsToGasCheck ?? chains,
-    minGas,
-    skipConfirmation,
-  );
+  assertSigner(signer);
```
manu-flint, line 58: "signer " becomes undefined here

PR #4452, opened 2024-09-07
chore: revert to using a 10 min relayer `CONFIRM_DELAY`

daniel-savu:


tkporter (approved):
wonder if eventually we should make this dependent on the reorgPeriod of the chain

daniel-savu:
> wonder if eventually we should make this dependent on the reorgPeriod of the chain

for sure, I had mentioned this to Connor on Saturday. The downside of having a large catch-all confirm delay is that we hold more ops that required in the confirm queue - we'll have to see how significant of a memory footprint this'll have @tkporter

paulbalaji:

PR #4466, opened 2024-09-09
feat: interchain account mirror

yorhodes:
### Description

See [(internal) notion spec](https://www.notion.so/hyperlanexyz/Governance-Contracts-Spec-Internal-b93e36ac8bd9472b8aa60be6dac7975a?pvs=4) for motivation. 

Implements an origin-side "mirror" to Interchain Accounts that exposes a "transparent" local address space interface to a target contract on a destination chain. 

Introduces a factory for deploying these mirrors with Create2 for committing to a specific `destination`, `target`, and `owner` tuple.

Does not use OpenZeppelin `Ownable` for immutable Create2 derivation on `owner`.

### Drive-by changes

None

### Backward compatibility

Yes

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/middleware/InterchainAccountMirror.sol
+++ b/solidity/contracts/middleware/InterchainAccountMirror.sol
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+// ============ Internal Imports ============
+import {TypeCasts} from "../libs/TypeCasts.sol";
+import {InterchainAccountRouter} from "./InterchainAccountRouter.sol";
+import {CallLib} from "./libs/Call.sol";
+
+/**
+ * Format of data:
+ * [   0: ????] destination call data
+ * [ -72:  -40] destination call value
+ * [ -40:  -20] destination ICA router
+ * [ -20:     ] destination ISM
+ */
+library InterchainAccountMirrorCalldata {
+    function encode(
+        bytes calldata _data,
+        uint256 _value,
+        address _icaRouter,
+        address _ism
+    ) internal pure returns (bytes memory) {
+        return abi.encodePacked(_data, _value, _icaRouter, _ism);
+    }
+
+    function destinationValue(
+        bytes calldata data
+    ) internal pure returns (uint256) {
+        return uint256(bytes32(data[data.length - 72:data.length - 40]));
+    }
+
+    function destinationCalldata(
+        bytes calldata data
+    ) internal pure returns (bytes calldata) {
+        return data[:data.length - 40];
+    }
+
+    function destinationIcaRouter(
+        bytes calldata data
+    ) internal pure returns (address) {
+        return address(bytes20(data[data.length - 40:]));
+    }
+
+    function destinationIsm(
+        bytes calldata data
+    ) internal pure returns (address) {
+        return address(bytes20(data[data.length - 20:]));
+    }
+}
+
+contract InterchainAccountMirror {
+    using InterchainAccountMirrorCalldata for bytes;
+    using TypeCasts for address;
+
+    uint32 immutable destination;
+    address immutable target;
+
+    address immutable deployer;
+
+    InterchainAccountRouter immutable icaRouter;
+
+    constructor(
+        uint32 _destination,
+        address _target,
+        InterchainAccountRouter _icaRouter
+    ) {
+        deployer = msg.sender;
+        destination = _destination;
+        target = _target;
+        icaRouter = _icaRouter;
+    }
+
+    modifier onlyDeployer() {
+        require(msg.sender == deployer, "sender not deployer");
+        _;
+    }
+
+    // solhint-disable-next-line no-complex-fallback
+    fallback() external payable onlyDeployer {
+        CallLib.Call[] memory calls = new CallLib.Call[](1);
+        calls[0] = CallLib.Call({
+            to: target.addressToBytes32(),
+            value: msg.data.destinationValue(),
+            data: msg.data.destinationCalldata()
+        });
+        icaRouter.callRemoteWithOverrides{value: msg.value}(
+            destination,
+            msg.data.destinationIcaRouter().addressToBytes32(),
+            msg.data.destinationIsm().addressToBytes32(),
+            calls
+        );
+    }
+}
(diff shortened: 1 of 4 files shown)


solidity/contracts/middleware/InterchainAccountMirror.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.13;
+
+// ============ Internal Imports ============
+import {TypeCasts} from "../libs/TypeCasts.sol";
+import {InterchainAccountRouter} from "./InterchainAccountRouter.sol";
+import {CallLib} from "./libs/Call.sol";
+import {InterchainAccountMirrorCalldata} from "./libs/InterchainAccountMirrorCalldata.sol";
+
+contract InterchainAccountMirror {
```
nambrot, line 10: Obiouvlsy im not actually a naming czar, but ngl this sounds p unintuitive to me. I think the word proxy would make more sense than mirror

solidity/contracts/middleware/InterchainAccountMirror.sol
```diff
@@ -0,0 +1,52 @@
...
+    using InterchainAccountMirrorCalldata for bytes;
+
+    InterchainAccountRouter immutable icaRouter;
+
+    uint32 immutable destination;
+    bytes32 immutable target;
+
+    address immutable owner;
```
nambrot, line 18: Don't love this, but i admit its not as bad as I thought

solidity/contracts/middleware/InterchainAccountMirror.sol
```diff
@@ -0,0 +1,52 @@
...
+        owner = _owner;
+    }
+
+    modifier onlyOwner() {
+        require(msg.sender == owner, "InterchainAccountMirror: only owner");
+        _;
+    }
+
+    // solhint-disable-next-line no-complex-fallback
+    fallback() external payable onlyOwner {
```
nambrot, line 38: I think this is v nice, but IMO we should still expose the batch functions and hook overrides to be "safe"

solidity/contracts/middleware/libs/InterchainAccountMirrorCalldata.sol
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Format of data:
+ * [   0: ????] destination call data
+ * [ -96:  -64] destination call value
+ * [ -64:  -32] destination ICA router
+ * [ -32:     ] destination ISM
+ */
+library InterchainAccountMirrorCalldata {
+    uint32 internal constant VALUE_OFFSET = 96;
+    uint32 internal constant ICA_ROUTER_OFFSET = 64;
```
nambrot, line 13: \*\*shiver**

dont love it still, but remains your call

solidity/test/InterchainAccountMirror.t.sol
```diff
@@ -0,0 +1,78 @@
...
+        bytes32 target,
+        bytes32 destinationIcaRouter,
+        bytes32 destinationIsmAddress
+    ) public payable {
+        vm.expectEmit(true, true, true, false);
+        emit InterchainAccountMirrorDeployed(
+            destination,
+            target,
+            owner,
+            address(0x0) // assume CREATE2 works as expected
```
nambrot, line 42: I dont get this comment?

solidity/test/InterchainAccountMirror.t.sol
```diff
@@ -0,0 +1,78 @@
...
+
+import {Test} from "forge-std/Test.sol";
+
+import {InterchainAccountMirrorCalldata} from "../contracts/middleware/libs/InterchainAccountMirrorCalldata.sol";
+import {InterchainAccountMirror} from "../contracts/middleware/InterchainAccountMirror.sol";
+import {InterchainAccountMirrorFactory} from "../contracts/middleware/InterchainAccountMirrorFactory.sol";
+import {InterchainAccountRouter} from "../contracts/middleware/InterchainAccountRouter.sol";
+import {CallLib} from "../contracts/middleware/libs/Call.sol";
+
+contract InterchainAccountMirrorTest is Test {
```
nambrot, line 12: I think an e2e test with MockMailbox would be rather nice here

solidity/contracts/middleware/libs/InterchainAccountMirrorCalldata.sol
yorhodes: lmao
do you have an alternative?

solidity/contracts/middleware/InterchainAccountMirror.sol
yorhodes: proxy is already used for the other end and I didnt want to conflate the two

yorhodes: I personally spent a bit of time thinking about this and arrived at mirror

solidity/test/InterchainAccountMirror.t.sol
yorhodes: im just not testing the CREATE2 derivation

solidity/contracts/middleware/libs/InterchainAccountMirrorCalldata.sol
nambrot: Adding it to the immutables of the Mirror?

solidity/test/InterchainAccountMirror.t.sol
nambrot: Still took me a second to understand that you are not asserting the last argument, maybe say something like "Not asserting the create2 derivation here" or something like that

solidity/contracts/middleware/libs/InterchainAccountMirrorCalldata.sol
yorhodes: is the reason you dont like this simply the encoding is trickier to interpret? having the value be committed to in an immutable seems especially silly

solidity/contracts/middleware/InterchainAccountMirror.sol
yorhodes: ingress and egress proxy?

yorhodes: given this is not intended for external use atm the name does not seem important

PR #4468, opened 2024-09-10
fix: ensure no duplicate signatures verified for `AbstractWeightedMultisigIsm.verify()`

aroralanuk:
### Description

- There's a bug in the verify function of `AbstractWeightedMultisigIsm` that an attacker might use to bypass the verification of multiple signatures. The code tried to check the duplication of the signers if not found, but the code does not increment validatorIndex when the recovered signer matches to the stored signer. For instance:

- validatorsAndThresholdWeight returns [A, B, C, D]
- an attacker uses signatures as [sig from A, sig from A, sig from A, sig from A, ...] or [sig from B, sig from B, …]

Fix is to add a ++validatorIndex at the end of the for loop implies we don't allow the next signer to be the same as the signer we just verified.

### Drive-by changes

None

### Related issues

From Chainlight's audit findings

### Backward compatibility

We haven't deployed these contracts yet on testnet/mainnet

### Testing

Fuzz testing


solidity/test/isms/MultisigIsm.t.sol
yorhodes, line 1: thanks for adding this

yorhodes: good lesson to add "adversarial" test cases in addition to fuzzed inputs

solidity/contracts/isms/multisig/AbstractWeightedMultisigIsm.sol
```diff
@@ -90,6 +90,7 @@ abstract contract AbstractStaticWeightedMultisigIsm is
 
             // add the weight of the current validator
             _totalWeight += _validators[_validatorIndex].weight;
+            ++_validatorIndex;
```
yorhodes, line 93: should we rename `i` to be `signatureIndex` for clarity?

yorhodes (approved):

PR #4469, opened 2024-09-10
feat: remove extraneous caching

paulbalaji:
feat: remove yarn build cache
- saves 600MB of cache space PER COMMIT
- doesn't impact overall per-commit CI time
- better CI time in the general case due to fewer cache evictions on the e2e matrix job
- [with caching](https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10797745470) vs [no caching](https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/10797213927?pr=4469)

drive-by: disable docker caching for agent build
- currently cache has no impact on build time of the agent image
- rerunning builds for a commit or pushing empty commits should in theory lead to quicker builds with layer caching, but the main problem is that the `RUN` step of building the agents is not cached by the docker action
- only RUNs with bind mounts get a checksum computed [(reference)](https://docs.docker.com/build/cache/invalidation/#general-rules)
- disabling docker build caching could help save a bunch of cache space
- best way to 4x speed here is moving this job to a buildjet 8vcpu runner (read [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4445) for more info)

drive-by: TLC 
- update to latest docker build action
	- provides a nice build summary and can generate artifacts that help break down where build time is spent
- update check-env before build, given the current deprecation annotations
	- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- coverage step doesn't need to wait for `yarn-test` anymore

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,35 +54,6 @@ jobs:
             exit 1
           fi
 
-  yarn-build:
-    runs-on: ubuntu-latest
-    needs: [yarn-install]
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          submodules: recursive
-          fetch-depth: 0
-
-      - name: yarn-cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            **/node_modules
-            .yarn
-          key: ${{ runner.os }}-yarn-cache-${{ hashFiles('./yarn.lock') }}
-
-      - name: build-cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            ./*
-            !./rust
-          key: ${{ github.event.pull_request.head.sha || github.sha }}
-
-      - name: build
-        run: yarn build
-
   lint-prettier:
     runs-on: ubuntu-latest
     needs: [yarn-install]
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -272,20 +218,10 @@ jobs:
           E2E_KATHY_MESSAGES: '20'
           RUST_BACKTRACE: 'full'
 
-  e2e:
```
jmrossy, line 275: We don't need this anymore?

jmrossy (commented):
Nice!

.github/workflows/test.yml
paulbalaji: we originally had to it be backwards compatible when e2e was a single job, pre-matrix. nowadays in the branch protection settings we require the specific cases in the matrix to pass, as opposed to the overall e2e job

also given that we don't want to fail-fast in the e2e-matrix, we'll always hit this e2e job after the matrix completes and this step will always pass regardless of whether any job in the matrix failed. we could change this behaviour, but given our existing branch protection settings I think it's dead weight

jmrossy (approved):
These workflow defs are like greek to me to some extent but everything lgtm

daniel-savu (approved):
noice

PR #4470, opened 2024-09-10
fix: add 10% gas bump to initialize call in EvmModuleDeployer

paulbalaji:
fix: add 10% gas bump to initialize call in EvmModuleDeployer

### DIFF
--- a/typescript/sdk/src/deploy/EvmModuleDeployer.ts
+++ b/typescript/sdk/src/deploy/EvmModuleDeployer.ts
@@ -69,8 +69,18 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
 
     if (initializeArgs) {
       this.logger.debug(`Initialize ${contractName} on ${chain}`);
+      // Estimate gas for the initialize transaction
+      const estimatedGas = await contract.estimateGas.initialize(
+        ...initializeArgs,
+      );
+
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -69,8 +69,18 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
...
+      // Estimate gas for the initialize transaction
+      const estimatedGas = await contract.estimateGas.initialize(
+        ...initializeArgs,
+      );
+
+      // deploy with 10% buffer on gas limit
       const overrides = this.multiProvider.getTransactionOverrides(chain);
-      const initTx = await contract.initialize(...initializeArgs, overrides);
+      const initTx = await contract.initialize(...initializeArgs, {
+        gasLimit: estimatedGas.add(estimatedGas.div(10)),
```
jmrossy, line 80: Same feedback as last time, seems like these consts should be pulled out

jmrossy (approved):

typescript/sdk/src/deploy/EvmModuleDeployer.ts
paulbalaji: i have a post-it note to do it 😅 hoping to get a spare mo either this or next week

PR #4471, opened 2024-09-10
feat: deploy to everclear mainnet

paulbalaji:
feat: deploy to everclear mainnet
- registry PR https://github.com/hyperlane-xyz/hyperlane-registry/pull/187

PR #4473, opened 2024-09-10
feat: deploy to oort mainnet

paulbalaji:
feat: deploy to oort mainnet
- registry PR https://github.com/hyperlane-xyz/hyperlane-registry/pull/190
- builds on top of the everclear PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4471


typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -401,6 +401,11 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
     ],
   },
 
+  oortmainnet: {
+    threshold: 1,
+    validators: ['0x9b7ff56cd9aa69006f73f1c5b8c63390c706a5d7'],
+  },
```
jmrossy, line 407: These have always been in the SDK IIRC. I considered moving them to the registry but I believe @tkporter suggested they stay here. Is it time to revisit that? Seems like reducing the package touch points from 3 (registry, infra, SDK), to 2 would be a good thing

jmrossy (approved):

typescript/sdk/src/consts/multisigIsm.ts
tkporter: Yeah I think there have been a few recent discussions about moving this into the registry, unfortunately just hasn't been prioritized yet

tkporter (approved):

.registryrc
```diff
@@ -1 +1 @@
-7512eac66d258d6cd8e8f417665b8e2ef81ab353
+10e4db81aa24e28a2d17920e0eea87537a016a09
```
```suggestion
8e93ecb4fa207fc4822393a4c13627be3da89eaa
```

paulbalaji:
> Seems like reducing the package touch points from 3 (registry, infra, SDK), to 2 would be a good thing

+1, very down to get this moved to registry. Will see if I can do this in spare cycles

PR #4476, opened 2024-09-11
fix: add replay protection to `AbstractMessageIdAuthHook` and `AbstractMessageIdAuthorizedIsm`

aroralanuk:
### Description

Problem: 

Chainlight - "... the AbstractMessageIdAuthHook.postDispatch() does not verify whether the msg.sender is a Mailbox or whether the message is being replayed. Consequently, an attacker can resend the same interchain message immediately after Mailbox.dispatch() is called.

For example, suppose Alice sends an interchain message with 1 ether using the OP Stack ISM from L1. The Mailbox.dispatch() calls OPStackHook.postDispatch(), followed by the OPStackISM.verifyMessageId(messageId) being triggered in L2 by the deposit transaction. Under normal circumstances, when releaseValueToRecipient() is invoked, the verifiedMessages[messageId] (1 ether) will be transferred to the recipient.

However, an attacker can replay the message with 0 msg.value by calling OPStackHook.postDispatch() immediately after the user calls Mailbox.dispatch(). This causes OPStackISM.verifyMEssageId(messageId) to be called again, leading to verifiedMessages[messageId] being set 0 + VERIFIED_MASK. Since the verifiedMessages value is overwritten with 0, the user's funds become permanently locked within the ISM." 

Solution:

Add replay protection to both the `AbstractMessageIdAuthHook.postDispatch()` and `AbstractMessageIdAuthorizedIsm.verifyMessageId()` to ensure the same message cannot be replayed on either side. Why both:

- In the future, we may implement an ISM that doesn't require a verifyMessageId caching call before the verify call, and in that scenario, this attack vector can still be realized. 
- Even with replay protection on the hook side, some native or 3rd bridges may have the ability to retry messages (this is not the case for Optimism or Arbitrum) so it's useful to enable replay here too.

Adding replay protection on both sides might seem a bit redundant, but it's more of a precautionary measure, premeditating future hook-ISM design pitfalls.

### Drive-by changes

None

### Related issues

- fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/3

### Backward compatibility

No, but not deployed to any mainnet

### Testing

Unit tests

### DIFF
--- a/solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
+++ b/solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
@@ -123,6 +123,11 @@ abstract contract AbstractMessageIdAuthorizedIsm is
             msg.value < 2 ** VERIFIED_MASK_INDEX,
             "AbstractMessageIdAuthorizedIsm: msg.value must be less than 2^255"
         );
+        uint256 currentValue = verifiedMessages[messageId];
+        require(
+            !currentValue.isBitSet(VERIFIED_MASK_INDEX),
+            "AbstractMessageIdAuthorizedIsm: message already verified"
+        );
 
         verifiedMessages[messageId] = msg.value.setBit(VERIFIED_MASK_INDEX);
         emit ReceivedMessage(messageId);
--- a/solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
+++ b/solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
@@ -123,9 +123,8 @@ abstract contract AbstractMessageIdAuthorizedIsm is
             msg.value < 2 ** VERIFIED_MASK_INDEX,
             "AbstractMessageIdAuthorizedIsm: msg.value must be less than 2^255"
         );
-        uint256 currentValue = verifiedMessages[messageId];
         require(
-            !currentValue.isBitSet(VERIFIED_MASK_INDEX),
+            !isVerified(messageId),
             "AbstractMessageIdAuthorizedIsm: message already verified"
         );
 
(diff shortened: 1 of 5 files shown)


solidity/contracts/isms/hook/AbstractMessageIdAuthorizedIsm.sol
```diff
@@ -123,6 +123,11 @@ abstract contract AbstractMessageIdAuthorizedIsm is
             msg.value < 2 ** VERIFIED_MASK_INDEX,
             "AbstractMessageIdAuthorizedIsm: msg.value must be less than 2^255"
         );
+        uint256 currentValue = verifiedMessages[messageId];
+        require(
+            !currentValue.isBitSet(VERIFIED_MASK_INDEX),
+            "AbstractMessageIdAuthorizedIsm: message already verified"
+        );
```
yorhodes, line 130:
```suggestion
        require(
            !isVerified(messageId),
            "AbstractMessageIdAuthorizedIsm: message already verified"
        );
```

solidity/test/isms/OPStackIsm.t.sol
```diff
@@ -75,8 +75,8 @@ contract OPStackIsmTest is Test {
 
     function setUp() public {
         // block numbers to fork from, chain data is cached to ../../forge-cache/
-        mainnetFork = vm.createFork(vm.rpcUrl("mainnet"), 18_992_500);
-        optimismFork = vm.createFork(vm.rpcUrl("optimism"), 114_696_811);
+        mainnetFork = vm.createFork(vm.rpcUrl("mainnet"), 20_726_880);
+        optimismFork = vm.createFork(vm.rpcUrl("optimism"), 125_226_936);
```
yorhodes, line 79: why change these?

yorhodes (approved):

PR #4479, opened 2024-09-12
chore: tidy up worfklow triggers

paulbalaji:
chore: tidy up worfklow triggers

This will hopefully stop the double triggering we're seeing currently for the X/X-skipped workflows

rust.yml:
- run if there are changes to rust dir or to the workflow file
- always run in merge queue

rust-skipped.yml:
- only run if there are no changes to rust dir or to workflow file

test.yml:
- don't run if the only changes are MD files
- always run in merge queue

test.yml:
- run if the only changes are MD files

We always run the full rust.yml/test.yml workflows in the merge queue, because the path matching/filtering for push/pull_request isn't supported by the `merge_queue` argument. There's options to add a step to manually do it but considering our CI is a bit faster now it should be fine in the general case


.github/workflows/rust-skipped.yml
```diff
@@ -4,16 +4,16 @@ name: rust
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - 'rust/**'
-      - .github/workflows/rust.yml
+    paths:
+      - '**/*'
+      - '!rust/**'
```
paulbalaji, line 9:
```suggestion
      - '!rust/main/**'
      - '!rust/sealevel/**'
```

.github/workflows/rust-skipped.yml
```diff
@@ -4,16 +4,16 @@ name: rust
...
   pull_request:
     branches: [main]
-    paths-ignore:
-      - 'rust/**'
-      - .github/workflows/rust.yml
-  # Support for merge queues
-  merge_group:
+    paths:
+      - '**/*'
+      - '!rust/**'
```
paulbalaji, line 15:
```suggestion
      - '!rust/main/**'
      - '!rust/sealevel/**'
```

jmrossy (approved):

PR #4482, opened 2024-09-13
drive-by: move gasLimit buffer calculation to utils

paulbalaji:
drive-by: move gasLimit buffer calculation to utils


jmrossy (approved):
LGTM

Issue #4463, opened 2024-09-09
feat(cli): Remove mailbox confirmation for 


ltyu:
## Problem
Currently, `hyperlane warp deploy` asks 
<img width="837" alt="Screenshot 2024-09-09 at 1 55 43 PM" src="https://github.com/user-attachments/assets/dca2e3bc-34d4-40dc-bc1d-0096692b683d">

This confirmation step is superfluous because most users do not know what to do here and will default to 'yes'. 

## Solution
- Remove this confirmation step

## Nice to Have
- Remove this confirmation in all commands, if they exist

PR #4485, opened 2024-09-13
feat: remove mailbox confirmation prompt for `warp init` command

xeno097:
### Description

This PR removes the mailbox address confirmation prompt from the `warp init` command. The command now uses the default mailbox value retrieved from the registry and asks for user input only if a mailbox address is not found in the registry for the selected chain.

- Adds a `isAddressAndMatchesProtocol` function to validate that a provided string is a valid address and belongs to an expected protocol (Ethereum, Solana, ...) 
- Adds address validation for the user-provided mailbox address using the newly implemented `isAddressAndMatchesProtocol` function

Before:
image

After:
image

image

image

### Drive-by changes

- No

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4463

### Backward compatibility

- Yes

### Testing

- Manual testing

Notes:
- E2e test implementation was attempted but testing prompts have proven difficult and require further investigation/hacking to see if there is a way to test user prompts.

### DIFF
--- a/typescript/cli/src/config/warp.ts
+++ b/typescript/cli/src/config/warp.ts
@@ -131,14 +131,6 @@ export async function createWarpRouteDeployConfig({
   const result: WarpRouteDeployConfig = {};
   for (const chain of warpChains) {
     logBlue(`${chain}: Configuring warp route...`);
-    const type = await select({
-      message: `Select ${chain}'s token type`,
-      choices: TYPE_CHOICES,
-    });
-
-    // TODO: restore NFT prompting
-    const isNft =
-      type === TokenType.syntheticUri || type === TokenType.collateralUri;
 
     // default to the mailbox from the registry and if not found ask to the user to submit one
     const chainAddresses = await context.registry.getChainAddresses(chain);
@@ -149,6 +141,15 @@ export async function createWarpRouteDeployConfig({
         validate: isAddress,
         message: `Could not retrieve mailbox address from the registry for chain "${chain}". Please enter a valid mailbox address:`,
       }));
+
+    const type = await select({
+      message: `Select ${chain}'s token type`,
+      choices: TYPE_CHOICES,
+    });
+
+    // TODO: restore NFT prompting
+    const isNft =
+      type === TokenType.syntheticUri || type === TokenType.collateralUri;
 
     const interchainSecurityModule = advanced
       ? await createAdvancedIsmConfig(context)
--- a/typescript/cli/src/config/warp.ts
+++ b/typescript/cli/src/config/warp.ts
@@ -15,7 +15,7 @@ import {
 import {
   Address,
   assert,
-  isAddressAndMatchesProtocol,
+  isAddress,
   objMap,
   promiseObjAll,
 } from '@hyperlane-xyz/utils';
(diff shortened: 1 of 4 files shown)


typescript/cli/src/config/warp.ts
```diff
@@ -134,15 +140,24 @@ export async function createWarpRouteDeployConfig({
...
+    const chainMetadata = await context.registry.getChainMetadata(chain);
+    if (!chainMetadata) {
+      throw new Error(`Metadata not found for chain "${chain}"`);
+    }
+
+    let mailbox: Address;
+    if (chainAddresses) {
+      mailbox = chainAddresses.mailbox;
+    } else {
+      mailbox = await input({
```
yorhodes, line 155: I think we prefer `const` declarations when possible
```suggestion
    const mailbox = chainAddresses?.mailbox ?? await input({
```
can see the argument that the if/else is more readable

typescript/cli/src/config/warp.ts
```diff
@@ -134,15 +140,24 @@ export async function createWarpRouteDeployConfig({
...
-      'mailbox address',
-      'hyperlane-registry',
-    );
+    // default to the mailbox from the registry and if not found ask to the user to submit one
+    const chainAddresses = await context.registry.getChainAddresses(chain);
+
+    const chainMetadata = await context.registry.getChainMetadata(chain);
+    if (!chainMetadata) {
+      throw new Error(`Metadata not found for chain "${chain}"`);
+    }
```
yorhodes, line 149: this command did not require there to be chain metadata before your change so not the biggest fan of this only for input address parsing

yorhodes (commented):
nice work on your first PR, already looking good

typescript/cli/src/config/warp.ts
yorhodes: I understand the intent, just dont like that it is a "breaking" change
the CLI will fail for Solana/Cosmos warp routes in other commands so I do not see this as a significant improvement

typescript/cli/src/config/warp.ts
```diff
@@ -134,15 +140,15 @@ export async function createWarpRouteDeployConfig({
...
-    const mailbox = await detectAndConfirmOrPrompt(
-      async () => {
-        const addresses = await context.registry.getChainAddresses(chain);
-        return addresses?.mailbox;
-      },
-      `For ${chain}, enter the`,
-      'mailbox address',
-      'hyperlane-registry',
-    );
+    // default to the mailbox from the registry and if not found ask to the user to submit one
```
ltyu, line 143: Imho, this prompt should be moved up to right after the chain selection. This is because the prompt depends on the chain selection, and users should not have to go back and resolve chain-related prompts after something else (select token type). 

1. select chain
2. select token type
3. Prompt for mailbox address because chain does not have it.

I think it should be
1. select chain
2. Prompt for mailbox address because chain does not have it.
3. select token type

typescript/utils/src/addresses.ts
```diff
@@ -58,6 +58,15 @@ export function getAddressProtocolType(address: Address) {
   }
 }
 
+export function isAddressAndMatchesProtocol(
```
ltyu, line 61: is this still used? can remove if not

ltyu (commented):
great progress! 

You should also add a changeset by using `yarn changeset` in the root folder. let me know if you have trouble.

typescript/cli/src/config/warp.ts
xeno097: I agree, it makes more sense to have it there thanks for the suggestion!

ltyu (approved):
LGTM

PR #4487, opened 2024-09-15
fix: replace import console module with direct console

tiendn:
This PR addresses the restriction on using 'console' from module for follow [eslint rule](https://eslint.org/docs/latest/rules/no-restricted-imports). The changes include:

- Remove all import `console` module, use directly instead Such as: `import {log} from 'console` and `@hyperlane-xyz/utils`

#### Modified
- `services/HyperlaneService.ts`
- `scripts/annouce-validator.ts`

### Testing
- Everything work well. Just console

### Notes
- In the future, we should manage log on a unified logger instead of directly console

### DIFF
--- a/typescript/infra/scripts/announce-validators.ts
+++ b/typescript/infra/scripts/announce-validators.ts
@@ -3,6 +3,7 @@ import { readFileSync } from 'fs';
 import * as path from 'path';
 
 import { ChainName } from '@hyperlane-xyz/sdk';
+import { assert } from '@hyperlane-xyz/utils';
 
 import { getChains } from '../config/registry.js';
 import { InfraS3Validator } from '../src/agents/aws/validator.js';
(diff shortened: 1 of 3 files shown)


typescript/infra/scripts/announce-validators.ts
```diff
@@ -118,7 +117,7 @@ async function main() {
     const location = announcement.value.storage_location;
     const announcedLocations =
       await validatorAnnounce.getAnnouncedStorageLocations([address]);
-    assert(announcedLocations.length == 1);
+    console.assert(announcedLocations.length == 1);
```
jmrossy, line 120: Thanks for the fix but this is still not quite what it should be. The problem was the import. It should be `import {assert} from '@hyperlane-xyz/utils`

tiendn: Tks. I've fixed it!

jmrossy (approved):

PR #4500, opened 2024-09-16
fix: Sealevel hyp token adapter fixes

jmrossy:
### Description

- Fix arg validation for Sealevel HypNative adapters
- Increase Token.ts test coverage to Sealevel adapters
- Swallow getBalance errors for non-existent SVM accounts
- Allow extra properties in ChainMetadata objects

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4499

### Backward compatibility

Yes

### Testing

Tested in Warp UI

### DIFF
--- a/typescript/sdk/src/metadata/chainMetadataTypes.ts
+++ b/typescript/sdk/src/metadata/chainMetadataTypes.ts
@@ -249,11 +249,10 @@ export const ChainMetadataSchemaObject = z.object({
 });
 
 // Passthrough allows for extra fields to remain in the object (such as extensions consumers may want like `mailbox`)
-const ChainMetadataSchemaObjectExtensible =
-  ChainMetadataSchemaObject.passthrough();
+const ChainMetadataSchemaExtensible = ChainMetadataSchemaObject.passthrough();
 
 // Add refinements to the object schema to conditionally validate certain fields
-export const ChainMetadataSchema = ChainMetadataSchemaObjectExtensible.refine(
+export const ChainMetadataSchema = ChainMetadataSchemaExtensible.refine(
   (metadata) => {
     if (
       [ProtocolType.Ethereum, ProtocolType.Sealevel].includes(
--- a/typescript/sdk/src/metadata/chainMetadataTypes.ts
+++ b/typescript/sdk/src/metadata/chainMetadataTypes.ts
@@ -248,8 +248,12 @@ export const ChainMetadataSchemaObject = z.object({
     .describe('Properties to include when forming transaction requests.'),
 });
 
+// Passthrough allows for extra fields to remain in the object (such as extensions consumers may want like `mailbox`)
+const ChainMetadataSchemaObjectExtensible =
+  ChainMetadataSchemaObject.passthrough();
+
 // Add refinements to the object schema to conditionally validate certain fields
-export const ChainMetadataSchema = ChainMetadataSchemaObject.refine(
+export const ChainMetadataSchema = ChainMetadataSchemaObjectExtensible.refine(
   (metadata) => {
     if (
       [ProtocolType.Ethereum, ProtocolType.Sealevel].includes(
@@ -333,7 +337,9 @@ export const ChainMetadataSchema = ChainMetadataSchemaObject.refine(
     },
   );
 
-export type ChainMetadata<Ext = object> = z.infer<typeof ChainMetadataSchema> &
+export type ChainMetadata<Ext = object> = z.infer<
+  typeof ChainMetadataSchemaObject
+> &
   Ext;
 
 export type BlockExplorer = Exclude<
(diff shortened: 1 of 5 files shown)


typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -248,8 +248,12 @@ export const ChainMetadataSchemaObject = z.object({
     .describe('Properties to include when forming transaction requests.'),
 });
 
+// Passthrough allows for extra fields to remain in the object (such as extensions consumers may want like `mailbox`)
+const ChainMetadataSchemaObjectExtensible =
```
yorhodes, line 252: bit of a mouthful 😅

typescript/sdk/src/token/Token.ts
```diff
@@ -232,24 +214,46 @@ export class Token implements IToken {
...
+          mailbox,
+        },
         false,
       );
     } else if (standard === TokenStandard.SealevelHypSynthetic) {
+      assert(mailbox, `Mailbox required for Sealevel hyp tokens`);
+      assert(
+        collateralAddressOrDenom,
+        `collateralAddressOrDenom required for Sealevel hyp collateral tokens`,
+      );
```
yorhodes, line 248: I think this should be omitted here (synthetic branch)

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -108,11 +108,18 @@ export class SealevelTokenAdapter
...
   async getBalance(owner: Address): Promise<bigint> {
-    const tokenPubKey = this.deriveAssociatedTokenAccount(new PublicKey(owner));
-    const response = await this.getProvider().getTokenAccountBalance(
-      tokenPubKey,
-    );
-    return BigInt(response.value.amount);
+    try {
+      const tokenPubKey = this.deriveAssociatedTokenAccount(
+        new PublicKey(owner),
+      );
```
yorhodes, line 114: should we keep this derivation outside of the try/catch?

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -605,11 +619,18 @@ export class SealevelHypSyntheticAdapter extends SealevelHypTokenAdapter {
...
   override async getBalance(owner: Address): Promise<bigint> {
-    const tokenPubKey = this.deriveAssociatedTokenAccount(new PublicKey(owner));
-    const response = await this.getProvider().getTokenAccountBalance(
-      tokenPubKey,
-    );
-    return BigInt(response.value.amount);
+    try {
+      const tokenPubKey = this.deriveAssociatedTokenAccount(
+        new PublicKey(owner),
+      );
```
yorhodes, line 625: ditto
is there no easy way to dedupe this with the other adapters?

typescript/sdk/src/token/Token.ts
```diff
@@ -232,24 +214,46 @@ export class Token implements IToken {
...
-        sealevelAddresses!,
+        {
+          warpRouter: addressOrDenom,
+          token: collateralAddressOrDenom,
+          mailbox,
+        },
         false,
       );
     } else if (standard === TokenStandard.SealevelHypSynthetic) {
+      assert(mailbox, `Mailbox required for Sealevel hyp tokens`);
```
daniel-savu, line 244: nit: this `assert` is shared by all sealevel tokens so it could stay under `if (protocol === ProtocolType.Sealevel) {`

typescript/sdk/src/token/Token.ts
daniel-savu: I think it's at least needed for balance checks: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/5a0d68bdce143d31bd775e697aca11119c60fdea/typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts#L157

daniel-savu (approved):

typescript/sdk/src/metadata/chainMetadataTypes.ts
jmrossy: Renamed

typescript/sdk/src/token/Token.ts
jmrossy: I agree with Dan, I think it's needed b.c. Solana

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
jmrossy: Good idea, done.

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
jmrossy: There are differences btwn the getBalance implementations, I think we need them all.

typescript/sdk/src/token/Token.ts
jmrossy: I considered that but then TS won't know that we've confirmed the mailbox var isn't null. We'd still need the `!` symbols like before. Duplication vs type safety tradeoff. Opted for this approach but I can revert.

PR #4501, opened 2024-09-16
fix: warp apply throws if submitter transactionReceipts is undefined

ltyu:
### Description
Fix to add conditional for `transactionReceipts` so that the yaml is only parsed when it is not `undefined`. 

### Drive-by changes
E2e tests for warp apply strategy

### Backward compatibility
Yes

### Testing
Manual/Unit Tests

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -497,12 +497,10 @@ export async function runWarpRouteApply(
             });
           const transactionReceipts = await submitter.submit(...transactions);
 
-          if (transactionReceipts && transactionReceipts.length > 0) {
-            return logGreen(
-              `✅ Warp config update successfully submitted with ${submitter.txSubmitterType} on ${chain}:\n\n`,
-              indentYamlOrJson(yamlStringify(transactionReceipts, null, 2), 4),
-            );
-          }
+          return logGreen(
+            `✅ Warp config update successfully submitted with ${submitter.txSubmitterType} on ${chain}:\n\n`,
+            indentYamlOrJson(yamlStringify(transactionReceipts, null, 2), 4),
+          );
         } catch (e) {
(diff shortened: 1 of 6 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -677,11 +677,12 @@ async function enrollRemoteRouters(
...
-        `✅ Router enrollment update successfully submitted with ${submitter.txSubmitterType} on ${chain}:\n\n`,
-        indentYamlOrJson(yamlStringify(transactionReceipts, null, 2), 4),
-      );
+      if (transactionReceipts && transactionReceipts.length > 0) {
+        return logGreen(
+          `✅ Router enrollment update successfully submitted with ${submitter.txSubmitterType} on ${chain}:\n\n`,
+          indentYamlOrJson(yamlStringify(transactionReceipts, null, 2), 4),
+        );
+      }
     }),
```
yorhodes, line 686: does the `submit` return type allow for undefined? might make more sense to have the safe submitter return empty array or something

yorhodes (approved):

typescript/cli/src/deploy/warp.ts
xeno097: I agree, seems like the bug wasn't caught early because the `yamlStringify` function accepts as the first argument a param of type `any` while the `submit` function might not return a value

yorhodes (approved):

typescript/cli/src/deploy/warp.ts
ltyu: @xeno097 Yup, this was mostly due a change in the sdk version: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4415

I haven't dug into it deeply, but i seems the downgraded version returns undefined instead of an array. We'll handle it. Another reason is that we're using a .js file for our gnosis safe service, so no typing to help.

PR #4502, opened 2024-09-17
chore: Remove pined dependencies on cargo-platform and predicates-core

ameten:
### Description

Remove pined dependencies on cargo-platform and predicates-core since the pins are not needed.

### Backward compatibility

Yes

### Testing

Run E2E Tests


tkporter (approved):
Any idea why they were added to begin with?

ameten:
> Any idea why they were added to begin with?

Not at the moment. I should chat to the contributor of the change to figure it out.

Issue #4464, opened 2024-09-09
feat(cli): Allow user to decide to deploy a trusted relayer ISM


ltyu:
## Problem
Currently in the CLI command: `hyperlane warp init`, we initialize a `trustedRelayerIsm` in the config. This means that deployments will include a trusted relayer ism that is not useful outside of self-relaying. To go to production, we generally [ask](https://docs.hyperlane.xyz/docs/guides/remove-trusted-relayer) to remove this ISM. This is problematic because this can be an extra step that wastes time and gas.

## Solution
- Let the user pick whether they want to potentially deploy a `trustedRelayerIsm`
- a user should skip this prompt through `--yes`
- Update relevant docs for this

## Additional context
https://discord.com/channels/935678348330434570/1280579302773227571

PR #4504, opened 2024-09-17
feat: `warp init` simplified config

xeno097:
### Description

This PR updates the ISM configuration prompting for the `warp init` command allowing the user to choose if a trusted ISM should be used by default or a configured one.

Before:
image

image

After:
image

image

### Drive-by changes

- No

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4464

### Backward compatibility

- Yes

### Testing

- Manual


typescript/cli/src/config/warp.ts
```diff
@@ -107,6 +107,13 @@ export async function createWarpRouteDeployConfig({
   outPath: string;
   advanced: boolean;
 }) {
+  // Providing the global --yes flag will trigger the usage of a trusted ISM while the
+  // --advanced flag will force the user to manually configure an ISM. Both flags should
+  // not be set to true at the same time.
+  if (context.skipConfirmation && advanced) {
+    throw new Error('Arguments advanced and yes are mutually exclusive');
+  }
```
yorhodes, line 115: this feels like an annoying failure mode even though its more "explicit", can we just bias to one?

yorhodes (approved):

.changeset/long-swans-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': major
```
yorhodes, line 2: I dont think this is a major change actually?

yorhodes (changes requested):

typescript/cli/src/config/warp.ts
xeno097: Yep! I'll remove this check and if both flags have been provided then the `--advanced` flag will have priority

.changeset/long-swans-drive.md
xeno097: You are right I'll change it to minor

.changeset/long-swans-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': minor
+---
+
+Add prompt in `warp init` command to choose if a trusted relayer should be used instead of making the choice by default for the user adn enable the `--yes` flag to default to a trusted ISM
```
ltyu, line 5:
```suggestion
Add prompt in `warp init` command to choose if a trusted relayer should be used instead of making the choice by default for the user and enable the `--yes` flag to default to a trusted ISM
```

typescript/cli/src/config/warp.ts
```diff
@@ -144,9 +144,20 @@ export async function createWarpRouteDeployConfig({
...
-      ? await createAdvancedIsmConfig(context)
-      : createDefaultWarpIsmConfig(owner);
+    let createDefaultIsm: boolean;
+    if (advanced) {
+      createDefaultIsm = false;
+    } else if (context.skipConfirmation) {
+      createDefaultIsm = true;
+    } else {
+      createDefaultIsm = await confirm({
+        message: 'Use a trusted ISM for warp route',
```
ltyu, line 154:
```suggestion
        message: 'Do you want to use a trusted ISM for warp route?',
```

typescript/cli/src/config/warp.ts
```diff
@@ -144,9 +144,20 @@ export async function createWarpRouteDeployConfig({
...
+      });
+    }
+
+    const interchainSecurityModule = createDefaultIsm
```
ltyu, line 158: So right now, if the user selects 'no' for "Use a trusted ISM for warp route", then they will be prompted to go into "advanced mode" and have them configure an ISM. We want them to instead go through these paths:

1. if --advanced is specified: manually configure ISM. in other words, call `createAdvancedIsmConfig()`
2. if --advanced is not specified, prompt "Do you want to use a trusted ISM for warp route?"
  - if 'yes' , create _both_ trustedRelayer and defaultFallbackRouting. See `createDefaultWarpIsmConfig()`
  - If 'no' , only create a defaultFallbackRouting.

ltyu (commented):
Good progress. 
There is some conflicting (and confusing) flags that should not work together, in this case it is `--advanced` and `--yes`. One thing that can help is to force `--advanced` and `--yes` to be mutually exclusive by using `conflicts`

Here's where you can use it:
Screenshot 2024-09-19 at 1 46 20 PM

xeno097:
> Good progress. There is some conflicting (and confusing) flags that should not work together, in this case it is `--advanced` and `--yes`. One thing that can help is to force `--advanced` and `--yes` to be mutually exclusive by using `conflicts`
> 
> Here's where you can use it: Screenshot 2024-09-19 at 1 46 20 PM

Hey @ltyu before pushing the pr I tried the `conflicts` option but it fails even if the user does not provide it because we have default values for it, I implemented a manual check but changed it after a comment from @yorhodes here -> https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4504#discussion_r1763479106. I can re-add the manual check if we want to

ltyu:
> > Good progress. There is some conflicting (and confusing) flags that should not work together, in this case it is `--advanced` and `--yes`. One thing that can help is to force `--advanced` and `--yes` to be mutually exclusive by using `conflicts`
> > Here's where you can use it: Screenshot 2024-09-19 at 1 46 20 PM
> 
> Hey @ltyu before pushing the pr I tried the `conflicts` option but it fails even if the user does not provide it because we have default values for it, I implemented a manual check but changed it after a comment from @yorhodes here -> [#4504 (comment)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4504#discussion_r1763479106). I can re-add the manual check if we want to

ah make sense. i'm okay with it being biased to one of them.

typescript/cli/src/config/warp.ts
```diff
@@ -151,9 +151,28 @@ export async function createWarpRouteDeployConfig({
...
+     *  -- no flag is provided: the user must choose if the default ISM config should be used:
+     *    - yes: the default ISM config will be used (Trusted ISM + Default fallback ISM)
+     *    - no: the default fallback ISM will be used
+     */
+    let interchainSecurityModule: IsmConfig;
+    if (advanced) {
+      interchainSecurityModule = await createAdvancedIsmConfig(context);
+    } else if (context.skipConfirmation) {
+      interchainSecurityModule = createDefaultWarpIsmConfig(owner);
+    } else if (
```
ltyu, line 167: From a UX experience, i prefer all token related prompts to be together. In this case, a user selects their type, but then gets asked about ISMs, and then goes back to configuring the token. An example of this is for collateralVault
Screenshot 2024-09-25 at 11 52 32 AM

ltyu: i suggest moving the type selection to above the switch statement. but overall, good progress

xeno097: Agreed it makes more sense to have it there, thank you for the suggestion 🔥

ltyu (approved):
LGTM!

PR #4506, opened 2024-09-17
fix: obtain default multisend address correctly

paulbalaji:
fix: obtain default multisend address correctly
- should help decoding if we're using the expect set of multisend addresses
- also marks our multisend with `onlyCalls: true`

> The reason we can't use `defaultAddress` is because it's fixed to either `canonical` or `eip155` depending on the setting for `ethereum`, regardless of which actually is preferred for each given chain. however, the `networkAddresses` record has items in the correct order.
> See https://github.com/safe-global/safe-deployments/blob/b17349df0eb91cf3683eeae882c07f19188421d9/src/utils.ts#L29 for the implementation.

drive-by:
- bump up to latest safe-deployments patch

| Before | After |
|--------|-------|
| Before 1 | After 1 |
| Before 2 | After 2 |

### DIFF
--- a/typescript/infra/src/utils/safe.ts
+++ b/typescript/infra/src/utils/safe.ts
@@ -42,10 +42,12 @@ export async function createSafeTransaction(
   safeService: SafeApiKit.default,
   safeAddress: Address,
   safeTransactionData: MetaTransactionData[],
+  onlyCalls?: boolean,
 ): Promise<SafeTransaction> {
   const nextNonce = await safeService.getNextNonce(safeAddress);
   return safeSdk.createTransaction({
     safeTransactionData,
+    onlyCalls,
     options: { nonce: nextNonce },
   });
 }
--- a/typescript/sdk/src/utils/gnosisSafe.js
+++ b/typescript/sdk/src/utils/gnosisSafe.js
@@ -67,12 +67,10 @@ export async function getSafe(chain, multiProvider, safeAddress) {
     multiSend = getMultiSendDeployment({
       version: multiSendVersion,
       network: domainId,
-      released: true,
     });
     multiSendCallOnly = getMultiSendCallOnlyDeployment({
       version: multiSendCallOnlyVersion,
       network: domainId,
-      released: true,
     });
   }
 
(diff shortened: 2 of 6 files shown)


typescript/infra/src/utils/safe.ts
```diff
@@ -42,10 +42,12 @@ export async function createSafeTransaction(
   safeService: SafeApiKit.default,
   safeAddress: Address,
   safeTransactionData: MetaTransactionData[],
+  onlyCalls?: boolean,
 ): Promise<SafeTransaction> {
   const nextNonce = await safeService.getNextNonce(safeAddress);
   return safeSdk.createTransaction({
     safeTransactionData,
+    onlyCalls,
```
tkporter, line 50: So I understand correctly: we're still delegatecalling into the MultiSend contract, but the calls that are performed by the MultiSend contract can only be calls, not delegatecalls

tkporter (approved):
great digging!

typescript/infra/src/utils/safe.ts
paulbalaji: exactly

typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -83,9 +81,9 @@ export async function getSafe(chain, multiProvider, safeAddress) {
       [domainId]: {
         // Use the safe address for multiSendAddress and multiSendCallOnlyAddress
         // if the contract is not deployed or if the version is not found
-        multiSendAddress: multiSend?.defaultAddress || safeAddress,
+        multiSendAddress: multiSend?.networkAddresses[domainId] || safeAddress,
```
ltyu, line 84: iiuc, we expect our domainId to always "match" theirs?

ltyu (commented):
I did not review the infra stuff, but mostly lgtm

typescript/sdk/src/utils/gnosisSafe.js
paulbalaji: yup, because it's for EVM chains and we have chainId==domainId for those (and chainId is pretty standardised on evm)

paulbalaji: I'll add a comment to that effect in code

PR #4507, opened 2024-09-17
fix: Add recursive query string

ltyu:
### Description
- Adds `recursive=true` query string to the url in github-proxy

### DIFF
--- a/typescript/github-proxy/src/index.ts
+++ b/typescript/github-proxy/src/index.ts
@@ -1,7 +1,8 @@
 import { DISALLOWED_URL_MSG } from './errors.js';
 
-const GITHUB_API_ALLOWLIST =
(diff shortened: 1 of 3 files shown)


typescript/github-proxy/src/index.ts
```diff
@@ -12,7 +12,7 @@ export default {
       return new Response(DISALLOWED_URL_MSG, { status: 401 });
     }
 
-    const apiUrl = new URL(`${GITPUB_API_HOST}${apiUrlPath}`);
+    const apiUrl = new URL(`${GITPUB_API_HOST}${apiUrlPath}?recursive=true`);
```
jmrossy, line 15: Spoke offline about including all query params here. Agreed to allow just this one for now

jmrossy (approved):

jmrossy:
Please include a changeset

PR #4509, opened 2024-09-18
chore: update SVM ISMs

tkporter:
### Description

- Updates the enrolled sets on the default ISMs of eclipsemainnet and solanamainnet

### Drive-by changes

- stopped reading `type` completely for the multisig ISM configs -- this isn't needed whatsoever and caused issues bc the sealevel tooling would expect a number but infra would generate a string w/ the ISM type name
- to support string chain_ids (the case on Cosmos), added some logic
- removed `testnet3`, not needed anymore


rust/sealevel/client/src/router.rs
```diff
@@ -125,7 +126,19 @@ impl ChainMetadata {
...
-        self.domain_id.unwrap_or(self.chain_id)
+        self.domain_id.unwrap_or_else(|| {
+            // Try to parse as a number, otherwise panic, as the domain ID must
+            // be specified if the chain id is not a number.
+            self.chain_id
+                .as_u64()
+                .and_then(|v| v.try_into().ok())
+                .unwrap_or_else(|| {
+                    panic!(
+                        "Invalid chain_id cannot be used as the domain_id: {:?}",
```
daniel-savu, line 137: nit - as a CLI user I'd find something like this easier to understand: "`domain_id` is undefined, nor could it be inferred from `chain_id` due to a parsing error"

rust/sealevel/environments/mainnet3/chain-config.json
```diff
@@ -391,6 +391,9 @@
     },
     "protocol": "ethereum",
     "rpcUrls": [
+      {
+        "http": "https://still-red-moon.bsc.quiknode.pro/be390591d09c000af5f23b40a06b326ba2683a2d"
```
daniel-savu, line 395: is adding this url intended? If not, should remove from here and from the PR branch history

daniel-savu (approved):

rust/sealevel/environments/mainnet3/chain-config.json
tkporter: Not intentional, ty

PR #4510, opened 2024-09-18
feat: Fetch logs by transaction hash

ameten:
### Description

Cosmos indexers now can fetch logs by transaction hash. It allows to reprocess transaction which may be missed by range based indexers.

### Drive-by changes

Quite a bit of refactoring

### Related issues

Contributes into issue #4300

### Backward compatibility

Yes

### Testing

E2E Tests


rust/main/chains/hyperlane-cosmos/src/mailbox/delivery_indexer.rs
```diff
@@ -123,7 +123,21 @@ impl Indexer<H256> for CosmosMailboxDeliveryIndexer {
     }
 
     async fn get_finalized_block_number(&self) -> ChainResult<u32> {
-        self.indexer.get_finalized_block_number().await
+        self.provider.get_finalized_block_number().await
+    }
+
+    async fn fetch_logs_by_tx_hash(
```
tkporter, line 129: fwiw I don't expect these functions atm to be used for indexing deliveries or for message dispatches (just bc it's only meant to be used for indexing hook events), but having them doesn't hurt

rust/main/chains/hyperlane-cosmos/src/providers/rpc/provider.rs
```diff
@@ -0,0 +1,277 @@
...
+        // than indexing latency, so we do them sequentially.
+        let block = self.rpc_client.get_block(block_number).await?;
+        let block_results = self.rpc_client.get_block_results(block_number).await?;
+
+        Ok(self.handle_txs(block, block_results, parser, cursor_label))
+    }
+
+    #[instrument(err, skip(self, parser))]
+    #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
+    async fn get_logs_in_tx<T>(
```
tkporter, line 260: nice that the impl of this could just use other work we've already done

tkporter (approved):
nice!

PR #4518, opened 2024-09-18
feat: embed package version in deployed bytecode

yorhodes:
### Description

Embed NPM package version in all deployed contracts bytecode. Follows [similar pattern to published CLI version](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3010e36ec9ac981fb0e760841c2e9b6a4dde1691/typescript/cli/src/version.ts#L0-L1). 

Should not be conflated with `Mailbox.VERSION` used for message encoding.

### Drive-by Changes

Unfixes solidity NPM package version from typescript packages

### Backward compatibility

No, only applies to new deployments or upgrades

### Testing

Tested [changeset glob](https://changesets-docs.vercel.app/fixed-packages#using-glob-expressions)
```ts
console.log(micromatch(['@hyperlane-xyz/sdk', '@hyperlane-xyz/core', '@hyperlane-xyz/cli'], ['@hyperlane-xyz/!(core)|*']));
['@hyperlane-xyz/sdk', '@hyperlane-xyz/cli']
```

Added test that version function is on ABI


nambrot (commented):
Generally looks good, but would probably make two requests:

1. A CI test that requires all contracts to have this version accessible (presumably have a blacklist for ones that are excempt)
2. What's the migration story?

yorhodes:
> 1. A CI test that requires all contracts to have this version accessible (presumably have a blacklist for ones that are excempt)

added in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4518/commits/7346e4bde76a8de6f14f6682f171018d921d03cf

> 2. What's the migration story?

as specified in the design doc, I do not recommend that we upgrade and instead just assume a version ceiling for those contracts which do not have this onchain

solidity/contracts/AttributeCheckpointFraud.sol
```diff
@@ -26,7 +27,8 @@ struct Attribution {
  * @title AttributeCheckpointFraud
  * @dev The AttributeCheckpointFraud contract is used to attribute fraud to a specific ECDSA checkpoint signer.
  */
-contract AttributeCheckpointFraud is Ownable {
+
+contract AttributeCheckpointFraud is Ownable, PackageVersioned {
```
nambrot, line 31: Wdyt of checking in the contract with the version and then just have a git hook enforcing no diff? I know its pollutes the git diff more, but it honestly feels kinda dirty to modify smart contract source code at runtime

solidity/test/exhaustiveversion.test.ts
```diff
@@ -0,0 +1,63 @@
+import { CompilerOutputContract } from 'hardhat/types';
+import { readFile, readdir } from 'node:fs/promises';
+import { basename, join } from 'node:path';
+
+const EXCLUDE = [
```
nambrot, line 5: NIT: `TO_EXCLUDE` or `EXCLUDE_PATTERNS`

solidity/test/exhaustiveversion.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  '.dbg',
+  'test',
+  'Test',
+  'mock',
+  'Mock',
+  'interfaces',
+  'libs',
+  'Abstract',
+  'Base',
```
nambrot, line 14: This feels overly broad, can we be more specific?

solidity/test/exhaustiveversion.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  'Storage',
```
nambrot, line 15: Ditto

nambrot (approved):

solidity/contracts/AttributeCheckpointFraud.sol
yorhodes: wdym by modify at runtime
its still before compile time, the file is checked in with the version...

yorhodes: 530f985ed3cc0b2f2d15718cd73777063917dea9

solidity/test/exhaustiveversion.test.ts
yorhodes: good catch, was missing StorageGasOracle

PR #4522, opened 2024-09-18
fix: warp core sealevel native

yorhodes:
### Description

- **Modify native adapter collateral balance check**

### Drive-by changes

- None

### Related issues

- Opens https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4558

### Backward compatibility

Yes

### Testing

Manual in warp UI, released as a beta build  5.2.1-beta0


typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -516,6 +516,14 @@ export class SealevelHypNativeAdapter extends SealevelHypTokenAdapter {
   }
 
   override async getBalance(owner: Address): Promise<bigint> {
+    console.log({ owner });
+    if (eqAddress(owner, this.addresses.warpRouter)) {
+      const collateralAccount = this.deriveNativeTokenCollateralAccount();
+      console.log({ collateralAccount });
+      const balance = await this.getProvider().getBalance(collateralAccount);
```
tkporter, line 523: any reason to not do `this.wrappedNative.getBalance(collateralAccount)`? Just to play well with the existing abstractions

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -516,6 +516,14 @@ export class SealevelHypNativeAdapter extends SealevelHypTokenAdapter {
...
+      console.log({ balance });
+      return BigInt(balance.toString());
```
tkporter, line 525: like you called out, we could consider subtracting the required rent here to be 100% correct. Without it, we're accepting that we may allow someone to transferRemote more tokens than they'll be able to receive.

The way to do this would be:
1. get the account data length of `collateralAccount` via an RPC call (the data len is 0 for this collateral account)
2. use the Solana SDK to calculate the required rent (we can't hardcode this bc it's different on Solana vs Eclipse)

Example showing the rent via the CLI:
```
# Solana mainnet
$ solana rent 0 -u mainnet-beta
Rent per byte-year: 0.00000348 SOL
Rent per epoch: 0.000002439 SOL
Rent-exempt minimum: 0.00089088 SOL

# eclipse
$ solana rent 0 -u https://mainnetbeta-rpc.eclipse.xyz
Rent per byte-year: 0.000000034 SOL
Rent per epoch: 0.000000023 SOL
Rent-exempt minimum: 0.000008704 SOL
```

tkporter (approved):
lgtm, happy either way

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
yorhodes: I initially tried this but its a pubkey vs an address type

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
yorhodes: this is what I had written up locally
```ts
      const bytecodeLength = accountInfo!.data.length;
      const rent = await provider.getMinimumBalanceForRentExemption(bytecodeLength);
      console.log({ rent });
      const balance = await provider.getBalance(collateralAccount);
      console.log({ balance });
      const availableBalance = balance - rent;
      console.log({availableBalance})
      return BigInt(availableBalance.toString());
```

yorhodes: opened an issue

PR #4524, opened 2024-09-19
fix(contracts): add rebasing compatibility for `HypERC4626`

aroralanuk:
### Description

- Added overrides for transferFrom, totalSupply to reflect the internal share based accounting for the 4626 mirror asset

### Drive-by changes

- Overridden `_transfer` to update the Transfer event to display the asset being transfers as amount not the internal shares.

### Related issues

- fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/6

### Backward compatibility

Yes

### Testing

Fuzz testing


solidity/contracts/token/extensions/HypERC4626.sol
```diff
@@ -31,6 +33,67 @@ contract HypERC4626 is HypERC20 {
...
+    // ============ Public Functions ============
+
+    /// Override transfer to handle underlying amounts while using shares internally
+    /// @inheritdoc ERC20Upgradeable
+    function transfer(
+        address to,
+        uint256 amount
+    ) public virtual override returns (bool) {
+        address owner = _msgSender();
+        _transfer(owner, to, assetsToShares(amount));
```
yorhodes, line 45: nit: unnecessary intermediate variable
```suggestion
        _transfer(_msgSender(), to, assetsToShares(amount));
```

solidity/contracts/token/extensions/HypERC4626.sol
```diff
@@ -31,6 +33,67 @@ contract HypERC4626 is HypERC20 {
...
+        _spendAllowance(sender, spender, amount);
+        _transfer(sender, recipient, shares);
+        return true;
+    }
+
+    /// Override totalSupply to return the total assets instead of shares. This reflects the actual circulating supply in terms of assets, accounting for rebasing
+    /// @inheritdoc ERC20Upgradeable
+    function totalSupply() public view virtual override returns (uint256) {
+        return sharesToAssets(super.totalSupply());
+    }
```
yorhodes, line 67:
```suggestion
    function totalSupply() public view virtual override returns (uint256) {
        return sharesToAssets(totalShares());
    }
```

solidity/contracts/token/extensions/HypERC4626.sol
```diff
@@ -31,6 +33,67 @@ contract HypERC4626 is HypERC20 {
...
+
+    /// This returns the balance of the account in terms of assets, accounting for rebasing
+    /// @inheritdoc ERC20Upgradeable
+    function balanceOf(
+        address account
+    ) public view virtual override returns (uint256) {
+        return sharesToAssets(super.balanceOf(account));
+    }
```
yorhodes, line 75:
```suggestion
    function balanceOf(
        address account
    ) public view virtual override returns (uint256) {
        return sharesToAssets(shareBalanceOf(account));
    }
```

solidity/contracts/token/extensions/HypERC4626.sol
```diff
@@ -71,32 +136,15 @@ contract HypERC4626 is HypERC20 {
...
-    }
-
-    function assetsToShares(uint256 _amount) public view returns (uint256) {
-        return _amount.mulDiv(PRECISION, exchangeRate);
-    }
-
-    function sharesToAssets(uint256 _shares) public view returns (uint256) {
-        return _shares.mulDiv(exchangeRate, PRECISION);
+    ) internal virtual override {
+        super._transfer(sender, recipient, amount);
```
yorhodes, line 147: does this not already emit an event?

solidity/test/token/HypERC4626Test.t.sol
```diff
@@ -385,6 +492,32 @@
         );
     }
 
+    function testTotalSupply() public {
```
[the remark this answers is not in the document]
yorhodes: annoying that this is showing up here

yorhodes (approved):

solidity/contracts/token/extensions/HypERC4626.sol
aroralanuk: It does but the event will shares amount as transferAmount in the event which is not desirable

eunseong-theori: In `super._transfer()`, an event is still emitted showing the share amount. To avoid duplicate event calls, the `super._transfer()` call should be removed, and the internal logic needs to be implemented directly within the `_transfer()` override function.

aroralanuk: _balances is private in ERC20Upgradeable, so I cannot update it in the overriding function _transfer in HypERC4626, I can update the event signature to be TransferAssets() maybe?

minebuu: Yes, that is one approach, but the concern arises when handling the `Transfer` event in off-chain (e.g., on a CEX). They might call `transfer(from, to, 10)`, but the `Transfer()` event would emit the `share` value instead of 10. While it's unlikely that an ERC4626 token will be directly listed on a CEX, it would still be prudent to add a note about this in the documentation or function comments to prevent any unforeseen issues.

Another solution could be accessing `_balances` through the storage slot. https://programtheblockchain.com/posts/2018/03/09/understanding-ethereum-smart-contract-storage/

eunseong-theori: Yes, that is one approach, but the concern arises when handling the `Transfer` event off-chain (e.g., on a CEX). They might call `transfer(from, to, 10)`, but the `Transfer()` event would emit the `share` value instead of 10. While it's unlikely that an ERC4626 token will be directly listed on a CEX, it would still be prudent to add a note about this in the documentation or function comments to prevent any unforeseen issues.

Another solution could be accessing `_balances` through the storage slot.
https://programtheblockchain.com/posts/2018/03/09/understanding-ethereum-smart-contract-storage/

aroralanuk: After talking with the team, we decided that we prefer to just omit the event and document in the contract that the amount in the event is the share amount, not the asset amount.

PR #4526, opened 2024-09-19
fix: native value covers message value in HypNative

yorhodes:
### Description

`HypNative.transferRemote` overload checks that the native `msg.value` deposited covers the warp message `amount` in addition to checking the mailbox post-dispatch fee is covered.

### Backward compatibility

No, applies to net new `HypNative` and `HypNativeScaled` deployments

### Testing

Unit Tests


aroralanuk (approved):
lgtm

ltyu (approved):

Issue #3434, opened 2024-03-18
Use Jito for tx submission on Solana v2


tkporter:
## Problem

- seeing consistent issues with tx submission

## Solution

- let's try out Jito

tkporter:
done with https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3429

PR #4528, opened 2024-09-19
feat: v3 Jito submission (#3429)

daniel-savu:
Ports this v2 fix to v3 https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3434

originally authored by @tkporter


tkporter (approved):

typescript/infra/renzo-warp-route-config.yaml
```diff
@@ -0,0 +1,1430 @@
+arbitrum:
```
tkporter, line 1: was this intentional?

daniel-savu: nope, thanks

PR #4533, opened 2024-09-19
feat: monitor sealevel ATA payer balances in the key funder

tkporter:
### Description

- Tested and set up alerts
- We already have the relayer balances from the agents themselves, so just need the ATA payers (which are responsible for paying for the rent of token accounts that don't yet exist when bridging in)
- Putting it here because it's easy and this is what we did in v2 (see #3432, which this is based off)

### DIFF
--- a/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
+++ b/typescript/infra/scripts/funding/fund-keys-from-deployer.ts
@@ -887,18 +887,24 @@ class ContextFunder {
     const provider = new Connection(rpcUrls[0], 'confirmed');
 
     for (const { pubkey, walletName } of accounts) {
-      logger.info('Fetching Solana wallet balance', {
-        chain,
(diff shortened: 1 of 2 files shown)


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -812,6 +859,54 @@ class ContextFunder {
...
+      );
+      const balance = await provider.getBalance(pubkey);
+      logger.info(
+        {
+          balance,
+          chain,
+          pubkey: pubkey.toString(),
+          walletName,
+        },
+        'Got sealevel chain wallet balance',
```
paulbalaji, line 896:
```suggestion
        'Retrieved sealevel chain wallet balance',
```

paulbalaji (approved):
overall lgtm

Mo-Hussain (approved):

PR #4534, opened 2024-09-20
feat(contracts): add nonce for monotonically increasing delivery ordering for `HypERC4626`

aroralanuk:
### Description

- Added a `rateUpdateNonce` in `HypERC4626Collateral` and `previousNonce` in `HypERC4626` to ensure we only update the exchangeRate on the synthetic asset if the update was after the last recorded update. This is to make sure we don't update it to a stale exchange rate which may cause losses to users using the synthetic asset. 

### Drive-by changes

- `processInboundMessage` in MockMailbox` to simulate processing of messages out of order.

### Related issues

- fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/12

### Backward compatibility

Yes

### Testing

Unit test

### DIFF
--- a/.changeset/red-actors-shop.md
+++ b/.changeset/red-actors-shop.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/core': patch
+---
+
+Added nonce to HypERC4626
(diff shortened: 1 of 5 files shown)


yorhodes (approved):
hmm this is definitely a little scary
I think we should have an "ordered" middleware or something

Issue #3355, opened 2024-03-04
Successfully scrape Cosmos chains in e2e test


tkporter:
## Problem

We'll have mostly built out Scraper support for Cosmos at this point, but don't have a great way of testing it. We don't have much of a setup to roll out significant changes to the scraper without affecting prod, especially for the cosmos rollout - there's a testnet4 scraper, however we aren't on any Cosmos chains on testnet4, and arguably the testnet4 scraper is responsible for "prod" data in the explorer.

## Solution

Just as we've done for Ethereum chains in the Ethereum e2e test, we can get better test coverage and build confidence in rolling out new scraper changes by using the scraper in the cosmos e2e tests: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e737d998ff23b5aa6cf2be9634789ca5dc5a4275/rust/utils/run-locally/src/cosmos/mod.rs#L309

and termination invariants to the cosmos e2e test https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e737d998ff23b5aa6cf2be9634789ca5dc5a4275/rust/utils/run-locally/src/cosmos/mod.rs#L558 just as we have for the Ethereum one: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/utils/run-locally/src/invariants.rs#L84-L134

PR #4536, opened 2024-09-20
feat: E2E Test Cosmos Scraper Invariant

ameten:
### Description

- Add metrics based Scraper invariant into E2E Test for Cosmos
- Report error if transaction contains multiple contract execution messages
- Report error if fee are expressed in unsupported denomination

### Related issues

- Successfully scrape Cosmos chains in e2e test #3355

### Backward compatibility

Yes

### Testing

E2E Testing

### DIFF
--- a/rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
+++ b/rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
@@ -159,6 +159,10 @@ impl CosmosProvider {
         Ok(contract)
     }
 
+    /// Reports if transaction contains fees expressed in unsupported denominations
+    /// The only denomination we support at the moment is the one we express gas minimum price
+    /// in the configuration of a chain. If fees contain an entry in a different denomination,
+    /// we report it in the logs.
     fn report_unsupported_denominations(&self, tx: &Tx, tx_hash: &H256) {
         let supported_denomination = self.connection_conf.get_minimum_gas_price().denom;
         let unsupported_denominations = tx
@@ -173,6 +177,7 @@ impl CosmosProvider {
         if !unsupported_denominations.is_empty() {
             error!(
                 ?tx_hash,
+                ?supported_denomination,
                 ?unsupported_denominations,
                 "transaction contains fees in unsupported denominations, manual intervention is required");
         }
--- a/rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
+++ b/rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
@@ -9,7 +9,7 @@ use tendermint::hash::Algorithm;
 use tendermint::Hash;
 use tendermint_rpc::{client::CompatMode, Client, HttpClient};
 use time::OffsetDateTime;
-use tracing::error;
+use tracing::{error, warn};
 
 use hyperlane_core::{
     BlockInfo, ChainCommunicationError, ChainInfo, ChainResult, ContractLocator, HyperlaneChain,
@@ -22,7 +22,8 @@ use crate::libs::account::CosmosAccountId;
 use crate::providers::rpc::CosmosRpcClient;
 use crate::{ConnectionConf, CosmosAmount, HyperlaneCosmosError, Signer};
 
-static MICRO_TO_ATTO: Lazy<U256> = Lazy::new(|| U256::from(1_000_000_000_000u128));
+/// Exponent value for atto units (10^-18).
+static ATTO_EXPONENT: Lazy<u32> = Lazy::new(|| 18);
 
 /// Abstraction over a connection to a Cosmos chain
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


rust/main/utils/run-locally/src/cosmos/mod.rs
```diff
@@ -615,37 +616,38 @@ fn run_locally() {
 
 fn termination_invariants_met(
     relayer_metrics_port: u32,
+    scraper_metrics_port: u32,
     messages_expected: u32,
     starting_relayer_balance: f64,
 ) -> eyre::Result<bool> {
-    let gas_payments_scraped = fetch_metric(
+    let expected_gas_payments = messages_expected;
+    let gas_payments_event_count = fetch_metric(
```
daniel-savu, line 624: nice renaming, makes things less confusing

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
```diff
@@ -128,23 +132,66 @@ impl CosmosProvider {
...
+    fn convert_fee(&self, coin: &Coin) -> U256 {
+        let gas_price = self.connection_conf.get_minimum_gas_price();
+        if coin.denom.as_ref() != gas_price.denom {
+            return U256::zero();
+        }
+
+        let price = gas_price.amount.parse::<f64>().unwrap_or(1.0);
+        let amount = U256::from(coin.amount);
+        if price < 1.0 {
+            amount * *MICRO_TO_ATTO
```
daniel-savu, line 190: can you provide some more details about what this does?

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
```diff
@@ -128,23 +132,66 @@ impl CosmosProvider {
...
+            .amount
+            .iter()
+            .filter(|c| c.denom.as_ref() != supported_denomination)
+            .map(|c| c.denom.as_ref())
+            .fold("".to_string(), |acc, denom| acc + ", " + denom);
+
+        if !unsupported_denominations.is_empty() {
+            error!(
+                ?tx_hash,
+                ?unsupported_denominations,
```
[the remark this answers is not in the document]
ameten: I added `supported_denomination` into the logs. And I added a doc commit for the method as well.

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
ameten: I added a doc comment to the method.

daniel-savu: hm, so the goal is to scale the fee amount based on the fee denomination of chain we're reading from. It feels more natural to use the prefix in `gas_price.denom` for that, and check it against the [symbol](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/eb47aaee8627deab2e200bb48ee7df47d2fd33f4/rust/main/config/mainnet_config.json#L2064), assuming we only support native token fees for now. E.g. if the `gasPrice` denom is `untrn` and the symbol is `NTRN`, the `u` prefix should tell us by how much to scale. This would probably make this function easier to understand as well

ameten: Discussed in person:

1. The problem is that gas price for Neutron and Osmosis is below a single unit of the smallest denomination allowed for the corresponding currency: NTRN and OSMO.
2. Neutron and Osmosis have denomination 6, namely `untrn` and `uosmo` correspondingly for NTRN and OSMO.
3. If we keep gas price in `untrn` and `uosmo`, it will be stored as zero in database since we are using integer values for gas price in database and value after the decimal point will be rounded out.

So, the solution was to scale gas price to `10^-18` so that the gas price is bigger than unity.

Agreed:
1. We continue to scale gas prices for Neutron and Osmosis so that they are expressed in `10^-18` of its corresponding currency. It is inline with gas price in Ethereum which is expressed in wei.
2. We will use `nativeToken` to identify the way to scale instead of relying on the current heuristic: `price < 1.0`.

ameten: I pushed a commit.

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
```diff
@@ -22,7 +22,8 @@ use crate::libs::account::CosmosAccountId;
 use crate::providers::rpc::CosmosRpcClient;
 use crate::{ConnectionConf, CosmosAmount, HyperlaneCosmosError, Signer};
 
-static MICRO_TO_ATTO: Lazy<U256> = Lazy::new(|| U256::from(1_000_000_000_000u128));
+/// Exponent value for atto units (10^-18).
+static ATTO_EXPONENT: Lazy<u32> = Lazy::new(|| 18);
```
daniel-savu, line 26: nit: could make this a const `u32` now since there's no expression left to evaluate

daniel-savu (approved):

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
ameten: Fixed

PR #4545, opened 2024-09-22
feat(warpMonitor): Support collateral value monitoring

Mo-Hussain:
### Description
- Support continuous value monitoring by allowing promethues to scrape value metrics for native and collateral tokens
- Uses the `CoinGeckoTokenPriceGetter`
- Added `getTokenPriceByIds` helper method to the `CoinGeckoTokenPriceGetter`, to support fetching prices using the coin gecko id only
- Add a warp_route_id label to metrics

### Testing

Manual

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
@@ -10,6 +10,7 @@ data:
       type: collateral
       hypAddress: '0xdD313D475f8A9d81CBE2eA953a357f52e10BA357'
       tokenAddress: '0xd9343a049d5dbd89cd19dc6bca8c48fb3a0a42a7'
+      tokenCoingeckoId: 'orion-protocol' # ORN->LUMIA 1:1
       name: Lumia Token
       symbol: LUMIA
       decimals: 18
--- a/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
@@ -10,7 +10,7 @@ data:
       type: collateral
       hypAddress: '0xdD313D475f8A9d81CBE2eA953a357f52e10BA357'
       tokenAddress: '0xd9343a049d5dbd89cd19dc6bca8c48fb3a0a42a7'
-      tokenCoingeckoId: 'lumia'
+      tokenCoinGeckoId: lumia
       name: Lumia Token
       symbol: LUMIA
       decimals: 18
--- a/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
@@ -10,7 +10,7 @@ data:
       type: collateral
       hypAddress: '0xdD313D475f8A9d81CBE2eA953a357f52e10BA357'
       tokenAddress: '0xd9343a049d5dbd89cd19dc6bca8c48fb3a0a42a7'
-      tokenCoingeckoId: 'orion-protocol' # ORN->LUMIA 1:1
+      tokenCoingeckoId: 'lumia'
       name: Lumia Token
       symbol: LUMIA
       decimals: 18
--- a/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
@@ -8,6 +8,7 @@ data:
       type: collateral
       hypAddress: '0x9AD81058c6C3Bf552C9014CB30E824717A0ee21b'
       tokenAddress: '0x15700B564Ca08D9439C58cA5053166E8317aa138'
+      tokenCoinGheckoId: sei-fastusd
       name: fastUSD
       symbol: fastUSD
       decimals: 18
--- a/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
@@ -8,7 +8,7 @@ data:
       type: collateral
       hypAddress: '0x9AD81058c6C3Bf552C9014CB30E824717A0ee21b'
       tokenAddress: '0x15700B564Ca08D9439C58cA5053166E8317aa138'
-      tokenCoinGheckoId: sei-fastusd
+      tokenCoingeckoId: elixir-deusd # unique setup where we want deUSD to be deposited as collateral and we want fastUSD to be minted as a synthetic on sei
       name: fastUSD
       symbol: fastUSD
       decimals: 18
--- a/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
+++ b/typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
@@ -8,7 +8,7 @@ data:
       type: collateral
       hypAddress: '0x9AD81058c6C3Bf552C9014CB30E824717A0ee21b'
       tokenAddress: '0x15700B564Ca08D9439C58cA5053166E8317aa138'
-      tokenCoingeckoId: elixir-deusd # unique setup where we want deUSD to be deposited as collateral and we want fastUSD to be minted as a synthetic on sei
+      tokenCoinGeckoId: elixir-deusd # unique setup where we want deUSD to be deposited as collateral and we want fastUSD to be minted as a synthetic on sei
       name: fastUSD
       symbol: fastUSD
       decimals: 18
--- a/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
+++ b/typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
@@ -120,7 +120,6 @@ async function main(): Promise<boolean> {
 }
 
 interface warpRouteInfo {
-  warpRouteId: string;
   balance: number;
   value?: number;
 }
@@ -131,13 +130,9 @@ async function checkBalance(
   multiProtocolProvider: MultiProtocolProvider,
   tokenPriceGetter: CoinGeckoTokenPriceGetter,
 ): Promise<ChainMap<warpRouteInfo>> {
-  const chains = Object.keys(tokenConfig) as ChainName[];
-
   const output = objMap(
     tokenConfig,
     async (chain: ChainName, token: WarpRouteConfig[ChainName]) => {
-      const warpRouteId = createWarpRouteConfigId(token.symbol, chains);
-
       switch (token.type) {
         case TokenType.native: {
           switch (token.protocolType) {
@@ -160,7 +155,6 @@ async function checkBalance(
               });
 
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(nativeBalance, token.decimals),
                 ),
@@ -195,7 +189,6 @@ async function checkBalance(
                 value: nativeValue,
               });
               return {
-                warpRouteId,
                 balance: balanceFloat,
                 value: nativeValue,
               };
@@ -211,7 +204,6 @@ async function checkBalance(
               );
               const tokenBalance = await adapter.getBalance(token.hypAddress);
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(tokenBalance, token.decimals),
                 ),
@@ -257,7 +249,6 @@ async function checkBalance(
               });
 
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(collateralBalance, token.decimals),
                 ),
@@ -303,7 +294,6 @@ async function checkBalance(
                 value: collateralValue,
               });
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(collateralBalance, token.decimals),
                 ),
@@ -325,7 +315,6 @@ async function checkBalance(
                 await adapter.getBalance(token.hypAddress),
               );
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(collateralBalance, token.decimals),
                 ),
@@ -344,7 +333,6 @@ async function checkBalance(
               );
               const syntheticBalance = await tokenContract.totalSupply();
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(syntheticBalance, token.decimals),
                 ),
@@ -368,7 +356,6 @@ async function checkBalance(
                 await adapter.getTotalSupply(),
               );
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(syntheticBalance, token.decimals),
                 ),
@@ -376,7 +363,7 @@ async function checkBalance(
             }
             case ProtocolType.Cosmos:
               // TODO - cosmos synthetic
-              return { warpRouteId, balance: 0 };
+              return { balance: 0 };
           }
           break;
         }
@@ -393,7 +380,6 @@ async function checkBalance(
               const syntheticBalance = await xerc20.totalSupply();
 
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(syntheticBalance, token.decimals),
                 ),
@@ -428,7 +414,6 @@ async function checkBalance(
               );
 
               return {
-                warpRouteId,
                 balance: parseFloat(
                   ethers.utils.formatUnits(collateralBalance, token.decimals),
                 ),
@@ -441,7 +426,7 @@ async function checkBalance(
           }
         }
       }
-      return { warpRouteId, balance: 0 };
+      return { balance: 0 };
     },
   );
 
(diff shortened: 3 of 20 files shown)


typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -176,10 +225,35 @@ async function checkBalance(
...
                 token.hypAddress,
               );
-
-              return parseFloat(
+              const collateralBalanceFloat = parseFloat(
                 ethers.utils.formatUnits(collateralBalance, token.decimals),
               );
+              let collateralValue: number | undefined = undefined;
+              let collateralPrice: number | undefined = undefined;
+              if (token.tokenCoinGeckoId) {
```
Mo-Hussain, line 233: Should we make `tokenCoinGeckoId` this required?

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -142,9 +177,21 @@ async function checkBalance(
               const balance = ethers.BigNumber.from(
                 await adapter.getBalance(token.hypAddress),
               );
-              return parseFloat(
+              const price = await tokenPriceGetter.getTokenPrice(chain);
```
Mo-Hussain, line 180: Is is safe to assume that a native token type, the token will always be the gas token of the given chain?

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -142,9 +177,21 @@ async function checkBalance(
...
+              const balanceFloat = parseFloat(
                 ethers.utils.formatUnits(balance, token.decimals),
               );
+              const nativeValue = balanceFloat * price;
+              logger.debug('Native balance', {
```
Mo-Hussain, line 185: I will remove all these logs before merging

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -153,4 +155,30 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
...
+    const toQuery = ids.filter((id) => !this.cache.isFresh(id));
+    if (toQuery.length > 0) {
+      let response: any;
+      try {
+        response = await this.coinGecko.simple.price({
+          ids: toQuery,
+          vs_currencies: [currency],
+        });
+
+        if (response.success === true) {
```
Mo-Hussain, line 172: Sometimes we get rate limited here

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -468,27 +563,89 @@ const getXERC20Limit = async (
...
+    return undefined;
+  }
+}
+
+async function getNativeTokenValue(
+  chain: ChainName,
+  balanceFloat: number,
+  tokenPriceGetter: CoinGeckoTokenPriceGetter,
+): Promise<number | undefined> {
+  const price = await getTokenPriceByChain(chain, tokenPriceGetter);
```
Mo-Hussain, line 583: Is it safe to always assume that the token for the native warp variant will always match the gas token of the chain? `getTokenPriceByChain` will do this

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -337,12 +415,29 @@ export function updateTokenBalanceMetrics(
         token_name: token.name,
         wallet_address: token.hypAddress,
         token_type: token.type,
+        warp_route_id: createWarpRouteConfigId(
```
Mo-Hussain, line 418: @tkporter - we discussed the limitations of using this, mainly being that it can change for the renzo case and we would need to change to config infra every time and that it is very long and increasing in length

How about we have manual override in our infra config that could be used instead if set? We can set this manual override for renzo only for now?

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -153,4 +155,32 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
     // Update the cache with the newly fetched prices
     chains.map((chain, i) => this.cache.put(chain, prices[i]));
   }
+
+  public async getTokenPriceByIds(
```
Mo-Hussain, line 159: The try catch behavior could be pulled up into calling code instead of being here. The problem with this is that it can mask the reason for why the query failed. Sometimes we are rate limited and the cache is not updated as we don't have a price. We then have a cache miss which raises the exception that is bubbled up

typescript/sdk/src/gas/token-prices.ts
Mo-Hussain: I'll look at using the API as a follow up

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -337,12 +445,29 @@ export function updateTokenBalanceMetrics(
...
+          Object.keys(tokenConfig) as ChainName[],
+        ),
       })
-      .set(balances[chain]);
+      .set(balances[chain].balance);
+    if (balances[chain].value) {
+      warpRouteCollateralValue
+        .labels({
+          chain_name: chain,
+          token_address: token.tokenAddress ?? ethers.constants.AddressZero,
```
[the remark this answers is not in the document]
tkporter: yeah I think so

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -337,12 +415,29 @@ export function updateTokenBalanceMetrics(
         token_name: token.name,
```
tkporter, line 415: these labels seem to be nearly the same between `warpRouteTokenBalance` and `warpRouteCollateralValue`. any reason to not have them be exactly the same? can we also do all this once and put the labels in a variable that are reused?

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -108,11 +119,17 @@ async function main(): Promise<boolean> {
   return true;
 }
 
+interface warpRouteInfo {
```
tkporter, line 122: nit: `warpRouteInfo` -> `WarpRouteInfo`

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -108,11 +119,17 @@ async function main(): Promise<boolean> {
...
+  balance: number;
+  value?: number;
```
tkporter, line 124: I could see it not being super clear that value means the USD value, some comments or a rename of `valueUsd` could help

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -156,9 +195,11 @@ async function checkBalance(
                 { ibcDenom: token.ibcDenom },
               );
               const tokenBalance = await adapter.getBalance(token.hypAddress);
-              return parseFloat(
-                ethers.utils.formatUnits(tokenBalance, token.decimals),
-              );
+              return {
```
tkporter, line 198: why not also do it for Cosmos?

Also wonder if we could pull the whole 
```
const balanceFloat = parseFloat(
                ethers.utils.formatUnits(balance, token.decimals),
              );

              const nativeValue = await getNativeTokenValue(
                chain,
                balanceFloat,
                tokenPriceGetter,
              );

              return {
                balance: balanceFloat,
                value: nativeValue,
              };
```
part out into a function or something, we duplicate that

same goes for the collateral stuff

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -216,9 +276,11 @@ async function checkBalance(
               const collateralBalance = ethers.BigNumber.from(
                 await adapter.getBalance(token.hypAddress),
               );
-              return parseFloat(
-                ethers.utils.formatUnits(collateralBalance, token.decimals),
-              );
+              return {
```
tkporter, line 279: ditto we should do this for Cosmos

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -468,27 +563,89 @@ const getXERC20Limit = async (
...
+): Promise<number | undefined> {
+  try {
+    return tokenPriceGetter.getTokenPrice(chain);
+  } catch (e) {
+    logger.warn('Error getting token price', e);
+    return undefined;
+  }
+}
+
+async function getNativeTokenValue(
```
tkporter, line 578: I guess we have this so we don't need to specify the coingecko ID for native?

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
```diff
@@ -10,6 +10,7 @@ data:
       type: collateral
       hypAddress: '0xdD313D475f8A9d81CBE2eA953a357f52e10BA357'
       tokenAddress: '0xd9343a049d5dbd89cd19dc6bca8c48fb3a0a42a7'
+      tokenCoingeckoId: 'orion-protocol' # ORN->LUMIA 1:1
```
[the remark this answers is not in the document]
Mo-Hussain: I took this from registry @paulbalaji, was there a reason for picking orion-protocol?

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -468,27 +563,89 @@ const getXERC20Limit = async (
...
+  return balanceFloat * price;
+}
+
 async function checkWarpRouteMetrics(
   checkFrequency: number,
   tokenConfig: WarpRouteConfig,
   chainMetadata: ChainMap<ChainMetadata>,
 ) {
+  const tokenPriceGetter = new CoinGeckoTokenPriceGetter(
+    new CoinGecko(),
```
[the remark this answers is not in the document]
Mo-Hussain: Yes I have but it may be due to me using a much higher sampling window locally, the service is resilient to rate limits, we could use our api key to eliminate this issue

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
Mo-Hussain: For native token type we assume that the native has token will be the token used, is this a safe assumption?

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -468,27 +563,89 @@ const getXERC20Limit = async (
   };
 };
 
+async function getTokenPriceByChain(
+  chain: ChainName,
+  tokenPriceGetter: CoinGeckoTokenPriceGetter,
+): Promise<number | undefined> {
+  try {
+    return tokenPriceGetter.getTokenPrice(chain);
```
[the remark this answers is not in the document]
Mo-Hussain: Great catch!

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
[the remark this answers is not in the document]
paulbalaji: there was no lumia token at the time, and ORN->LUMIA should be a 1:1 swap

happy to switch over to `lumia` now though!

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
Mo-Hussain: @tkporter!

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
```diff
@@ -10,6 +10,7 @@ data:
       type: collateral
       hypAddress: '0xdD313D475f8A9d81CBE2eA953a357f52e10BA357'
       tokenAddress: '0xd9343a049d5dbd89cd19dc6bca8c48fb3a0a42a7'
```
Mo-Hussain, line 12: I'm getting a little confused with this setup, theres a Lumia ERC20 on Ethereum and also Lumia tokens on Lumia L2, does that make sense? Do they have a canonical bridge?

typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
```diff
@@ -8,6 +8,7 @@ data:
       type: collateral
       hypAddress: '0x9AD81058c6C3Bf552C9014CB30E824717A0ee21b'
       tokenAddress: '0x15700B564Ca08D9439C58cA5053166E8317aa138'
+      tokenCoinGheckoId: sei-fastusd
```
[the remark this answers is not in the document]
Mo-Hussain: Do we want to change the name and symbol on the collateral side to deUSD?

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
paulbalaji: would chat to @cmcewen

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -153,4 +155,32 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
     // Update the cache with the newly fetched prices
     chains.map((chain, i) => this.cache.put(chain, prices[i]));
```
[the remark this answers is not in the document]
Mo-Hussain: [(chain) => this.metadata[chain].gasCurrencyCoinGeckoId || chain,](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6dae7f9f5562bb5d2c140aeee5ef382fe9f501c9/typescript/sdk/src/gas/token-prices.ts#L151)

typescript/infra/config/environments/mainnet3/warp/sei-FASTUSD-deployments.yaml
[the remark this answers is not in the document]
tkporter: we probably should but I'm happy either way as long as it's consistent with what's in the registry

typescript/sdk/src/gas/token-prices.ts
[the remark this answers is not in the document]
tkporter: yeah exactly, like we should make sure the cache is used in a consistent way

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
tkporter: unfortunately I don't have any context here, agree Connor is best

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
tkporter: I think this is fine, extra context https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4545#discussion_r1809145827

typescript/infra/config/environments/mainnet3/warp/bsc-lumia-LUMIA-deployments.yaml
Mo-Hussain: Will look into this as a follow up

typescript/sdk/src/metadata/warpRouteConfig.ts
```diff
@@ -10,7 +10,7 @@ const TokenConfigSchema = z.object({
   type: z.nativeEnum(TokenType),
   hypAddress: z.string(), // HypERC20Collateral, HypERC20Synthetic, HypNativeToken address
   tokenAddress: z.string().optional(), // external token address needed for collateral type eg tokenAddress.balanceOf(hypAddress)
-  tokenCoinGeckoId: z.string().optional(), // CoinGecko id for token
+  tokenCoingeckoId: z.string().optional(), // CoinGecko id for token
```
tkporter, line 13: dumb nit but any reason not to keep it as `tokenCoinGeckoId`? We do `CoinGecko` for the registry's `gasCurrencyCoinGeckoId`, consistency probably makes sense

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -407,6 +404,9 @@ export function updateTokenBalanceMetrics(
         token.symbol,
         Object.keys(tokenConfig) as ChainName[],
       ),
+      related_chain_names: Object.keys(tokenConfig)
```
tkporter, line 407: we should make sure this ordering is deterministic, I don't think Object.keys gives any guarantees -- e.g. by sorting like what we do on the warp IDs. I want to avoid us ending up using different labels for the same warp route if this ordering were to change

also I take it you felt it was preferable to create a new label instead of using the existing `warp_route_id`? it's not a huge deal but we do duplicate this data

typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
```diff
@@ -590,26 +592,25 @@ async function getNativeTokenWarpInfo(
...
   balanceFloat: number,
   tokenPriceGetter: CoinGeckoTokenPriceGetter,
 ): Promise<number | undefined> {
-  logger.debug('tokenCoinGeckoId', tokenCoinGeckoId);
   const price = await getCollateralTokenPrice(
-    tokenCoinGeckoId,
+    tokenCoingeckoId,
     tokenPriceGetter,
   );
+  logger.debug(`${tokenCoingeckoId} token price ${price ? price : undefined}`);
```
tkporter, line 613: nit: could you just use `price` directly instead of `price ? price : undefined`?

tkporter (approved):

typescript/sdk/src/metadata/warpRouteConfig.ts
Mo-Hussain: Ah fair, will make the change!

PR #4549, opened 2024-09-23
chore: don't ask for api key if it already exists

nambrot:
### Description

CLI will ask for API keys even if your (local) registry already includes the API key. This PR fixes that. 

Can't say i love the current code setup, it seems to me that it should be more like a pass on the registry and filter chains that don't have api keys, prompt for the api key and then persist that in the registry if its writable. But since i have so little context, kept it mostly this way to accomplish the desired product experience.


ltyu (approved):
noice

PR #4552, opened 2024-09-23
fix(contracts): quote management for L2->L1 hooks 

aroralanuk:
### Description

- check for sufficient fees in `AbstractMessageIdAuthHook` and refund surplus
- add a child hook to OPL2ToL1Hook and ArbL2ToL1Hook to use the igp to pay for the destination gas fees

~~Note: LayerzeroL2Hook currently also refunds from msg.value, will make it into issue to be fixed later as we're using the layerzero hooks right now.~~

### Drive-by changes

- None

### Related issues

- fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/10

### Backward compatibility

No

### Testing

Fuzz


solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
```diff
@@ -82,11 +86,19 @@ abstract contract AbstractMessageIdAuthHook is
...
             id
         );
         _sendMessageId(metadata, payload);
+
+        uint256 _overpayment = msg.value - _quoteDispatch(metadata, message);
+        if (_overpayment > 0) {
+            address _refundAddress = metadata.refundAddress(msg.sender);
+            require(_refundAddress != address(0), "no refund address");
+            payable(_refundAddress).sendValue(_overpayment);
+        }
```
yorhodes, line 101: hm, I dont like the fact that `_sendMessageId` could pass all of the `msg.value` to the bridge and then this would revert
I also dont think defaulting to the `msg.sender` for the refund address is a good idea, wouldnt that often be the mailbox/another hook?

aroralanuk: A. I think the _sendMessageId should pass quoteDispatch not msg.value

B. What's an alternative here?

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -50,30 +52,41 @@ contract ArbL2ToL1Hook is AbstractMessageIdAuthHook {
...
+    /// @inheritdoc AbstractPostDispatchHook
     function _quoteDispatch(
-        bytes calldata,
-        bytes calldata
+        bytes calldata metadata,
+        bytes calldata message
     ) internal view override returns (uint256) {
-        return GAS_QUOTE;
+        return
+            metadata.msgValue(0) + childHook.quoteDispatch(metadata, message);
```
yorhodes, line 72: realizing that the IGP adds the default ISM overhead which this message would not incur?

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -50,30 +52,41 @@ contract ArbL2ToL1Hook is AbstractMessageIdAuthHook {
...
-        bytes memory payload
+        bytes calldata message
     ) internal override {
+        bytes memory payload = abi.encodeCall(
+            AbstractMessageIdAuthorizedIsm.verifyMessageId,
+            message.id()
+        );
+
+        childHook.postDispatch{
+            value: childHook.quoteDispatch(metadata, message)
```
yorhodes, line 88: hm, I dont love that we have to quote here

solidity/contracts/hooks/OPL2ToL1Hook.sol
```diff
@@ -58,27 +64,32 @@ contract OPL2ToL1Hook is AbstractMessageIdAuthHook {
...
-            msg.value >= metadata.msgValue(0) + GAS_QUOTE,
-            "OPL2ToL1Hook: insufficient msg.value"
+        bytes memory payload = abi.encodeCall(
+            AbstractMessageIdAuthorizedIsm.verifyMessageId,
+            message.id()
         );
+
+        childHook.postDispatch{
+            value: childHook.quoteDispatch(metadata, message)
+        }(metadata, message);
```
yorhodes, line 88: could this block be on `AbstractMessageIdAuthHook`?

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -35,13 +36,14 @@ import {ArbSys} from "@arbitrum/nitro-contracts/src/precompiles/ArbSys.sol";
...
+    using Message for bytes;
 
     // ============ Constants ============
 
     // precompile contract on L2 for sending messages to L1
     ArbSys public immutable arbSys;
-    // Immutable quote amount
-    uint256 public immutable GAS_QUOTE;
+    // child hook to call first
+    AbstractPostDispatchHook public immutable childHook;
```
yorhodes, line 46: using the `Abstract` as the interface is a bit weird
can we just use the `IPostDispatchHook`?

solidity/contracts/hooks/ArbL2ToL1Hook.sol
```diff
@@ -50,30 +52,41 @@ contract ArbL2ToL1Hook is AbstractMessageIdAuthHook {
         uint32 _destinationDomain,
         bytes32 _ism,
         address _arbSys,
-        uint256 _gasQuote
+        address _childHook
     ) AbstractMessageIdAuthHook(_mailbox, _destinationDomain, _ism) {
         arbSys = ArbSys(_arbSys);
-        GAS_QUOTE = _gasQuote;
+        childHook = AbstractPostDispatchHook(_childHook);
```
yorhodes, line 58: nit: if the type implies "hook" the name does not need to include this

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: Yes, we'll need to override that when specifying metadata to this hook. Maybe I can add ARB_ISM_GAS_USAGE as a constant to the hook which overrides the metadata? This is also an argument for specifying the childHook ie igp and not the abstract childHook.

solidity/contracts/hooks/OPL2ToL1Hook.sol
aroralanuk: Other hooks don't need child hooks.

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: Better to over-specify than under-specify no?

yorhodes: with naming I dont agree with this but dont feel strongly enough to debate

solidity/contracts/hooks/ArbL2ToL1Hook.sol
yorhodes: I dont think we can override the IGP gas limit

solidity/contracts/hooks/OPL2ToL1Hook.sol
```diff
@@ -30,13 +33,16 @@ import {ICrossDomainMessenger} from "../interfaces/optimism/ICrossDomainMessenge
...
+    using Message for bytes;
 
     // ============ Constants ============
 
     // precompile contract on L2 for sending messages to L1
     ICrossDomainMessenger public immutable l2Messenger;
-    // Immutable quote amount
-    uint32 public immutable GAS_QUOTE;
+    // child hook to call first
+    IPostDispatchHook public immutable childHook;
```
yorhodes, line 43: why not have this on the abstract hook?

solidity/contracts/hooks/OPL2ToL1Hook.sol
yorhodes: you have the same code on the arbitrum hook

solidity/contracts/hooks/layer-zero/LayerZeroV2Hook.sol
yorhodes, line 1: why did we change this at all?

solidity/test/hooks/layerzero/LayerZeroV2Hook.t.sol
```diff
@@ -165,25 +165,19 @@ contract LayerZeroV2HookTest is Test {
...
-        mailbox.dispatch{value: balance}(
-            HYPERLANE_DEST_DOMAIN,
-            address(crossChainCounterApp).addressToBytes32(),
-            "Hello World!",
-            metadata,
-            hook
-        );
-        assertEq(address(alice).balance, extraValue);
-    }
+    // TODO: fix double refund for LayerZeroV2Hook
```
yorhodes, line 168: I dont think we should merge this if we broke something and dont know why

solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
```diff
@@ -82,20 +86,27 @@ abstract contract AbstractMessageIdAuthHook is
...
+        _sendMessageId(metadata, message);
+
+        uint256 _overpayment = msg.value - _quoteDispatch(metadata, message);
+        if (_overpayment > 0) {
+            address _refundAddress = metadata.refundAddress(msg.sender);
+            require(
+                _refundAddress != address(0),
+                "AbstractPostDispatchHook: no refund address"
+            );
+            payable(_refundAddress).sendValue(_overpayment);
```
eunseong-theori, line 99: [In `LayerZeroV2Hook.sol`](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/9093369186993468571b5d411e712ddd9a30c98c/solidity/contracts/hooks/layer-zero/LayerZeroV2Hook.sol#L81), when calling `lZEndpoint.send{value: msg.value}(msgParams, refundAddress);`, all of `msg.value` is sent instead of the amount specified by `_quoteDispatch()`, which can cause a revert in overpayment scenarios

solidity/contracts/hooks/ArbL2ToL1Hook.sol
aroralanuk: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2cc51c1640db7757a08ecd16241a49efd17aff3c/solidity/contracts/hooks/igp/InterchainGasPaymaster.sol#L272

solidity/contracts/hooks/OPL2ToL1Hook.sol
aroralanuk: only arbL2ToL1Hook and OpL2ToL1Hook, remember the rule of 3?

solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
aroralanuk: Fix with `lZEndpoint.send{value: quote}(msgParams, refundAddress)`

solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
yorhodes: the hyperlane `message.sender()` like we do elsewhere

yorhodes (approved):

solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
aroralanuk: fixed here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4552/commits/e16958ea5e754c1541eb508e1beb53fdc49224b3

PR #4553, opened 2024-09-23
fix(contracts): add check for valid mailbox and relayer in `TrustedRelayerIsm`

aroralanuk:
### Description

- minor fix: add an isContract check for mailbox and non-zero check for relayer when instantiating TrustedRelayerIsm

### Drive-by changes

None

### Related issues

- partly fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/14

### Backward compatibility

Yes

### Testing

Unit

### DIFF
--- a/solidity/contracts/isms/TrustedRelayerIsm.sol
+++ b/solidity/contracts/isms/TrustedRelayerIsm.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.8.0;
 
 // ============ Internal Imports ============
 import {IInterchainSecurityModule} from "../interfaces/IInterchainSecurityModule.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 import {Message} from "../libs/Message.sol";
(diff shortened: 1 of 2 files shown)


yorhodes (approved):

ltyu (commented):
always feel like address(0) checks are not needed any more: https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222

you can technically set this as any address that you don't own and its considered "misconfigured"

yorhodes:
> always feel like address(0) checks are not needed any more: [forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222](https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222)
> 
> you can technically set this as any address that you don't own and its considered "misconfigured"

it is quite useful to catch these configuration issues at deploy time even if solidity will include the 0x0.call check at "runtime"

PR #4556, opened 2024-09-23
chore: root repo vscode settings for `rust-analyzer`

daniel-savu:
### Description

`rust-analyzer` only detects workspaces that are 1 directory deep in the source tree, otherwise it has to be configured with the path to workspaces. This PR checks in `.vscode` settings with paths to the rust workspaces. 

This also fixes the issue where someone checks out the monorepo to make TS changes but can't commit because our husky precommit setups assumes that rust workspaces have already been built (which `rust-analyzer` does implicitly).

Noting that the problem is only there if the repo isn't opened using the `mono.code-workspace` setting, which people I talked to don't use (other than @paulbalaji)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "rust-analyzer.linkedProjects": [
+        "./rust/main/Cargo.toml",
+        "./rust/sealevel/Cargo.toml",
+    ],
+}


jmrossy (approved):
Optional but now would be a good time mirror all the useful settings we've collected in mono.code-workspace, which are currently not benefiting most devs

daniel-savu:
> Optional but now would be a good time mirror all the useful settings we've collected in mono.code-workspace, which are currently not benefiting most devs

@jmrossy this particular PR fixes lost functionality so is more urgent, happy to have a follow-up PR for the nice-to-have ones

PR #4560, opened 2024-09-23
fix: don't reuse prefix key in relayer

daniel-savu:
### Description

This PR fixes a bug @tkporter flagged, which would cause agents to reindex every igp payment on restart (but luckily with no impact on message processing).

Unfortunately the same DB prefix key was used to store different items: `gas_payment_by_sequence` and `gas_payment_block_by_sequence`, with the latter overwriting the former. Sequence indexing relies on `gas_payment_by_sequence` to know whether a certain payment was already indexed - if not, it queries an RPC for it and stores it in the db. But because `gas_payment_by_sequence` would get overwritten and thus corrupted, checking whether a certain `gas_payment_by_sequence` is already in the db would fail with a decoding error, and because errors are treated as if the payment just wasn't indexed yet, the RPC call would be made again.

Fortunately no migration is required, since just re-running the relayer will cause it to overwrite the old corrupted entries with correct ones, and would also create new prefix entries for the igp blocks.


tkporter (approved):
🙏

PR #4562, opened 2024-09-23
feat: revamp testnet4 gasoracle/igp calculations

paulbalaji:
looking to resolve https://github.com/hyperlane-xyz/issues/issues/1374 more longer term

feat: revamp testnet4 gasoracle/igp calculations
- update print-gas-prices script to support testnets
- update print-token-prices script to support testnets (requires https://github.com/hyperlane-xyz/hyperlane-registry/pull/228)
- remove the common/rare/mythic nomenclature, use mainnet prices
- refactor igp utils, reuse across both mainnet3 + testnet4


tkporter (approved):
down for this as long as sending to Sepolia isn't egregious

PR #4565, opened 2024-09-24
fix(relayer): dry-run before sending txs

daniel-savu:
### Description

Fixes a bug in the individual submission logic. There is a race condition where a message can become undeliverable between when it was simulated in the `prepare` step and when it's actually submitted.

### Context
The bug was discovered as part of batched messages but affects chains without batching too.

- when delivery fails inside batch submission simulation, we fall back to individual submission
- individual submission doesn't simulate the tx outcome, and instead just submits directly
    - this is wrong, because the batch may cause messages to become undeliverable for long periods of time (such as when a mint cap is reached)
    - I believe simulation used to happen in the individual submission step, but we switched to using the already computed gas estimate from the IGP check in the `prepare` step, to save an RPC call
- this results in senders being charged from their IGP allowance unnecessarily, and messages becoming undeliverable due to insufficient gas

The correct flow requires three delivery simulations:
1. in the `prepare` step, to estimate gas and check the IGP policy
2. in the batch submission logic, to filter messages that are made invalid by earlier messages in the batch
3. right before submitting again, in case the batch caused the message to not be deliverable

### Solution
Dry-run right before submitting an operation, and if it fails send it back to the prep queue

### DIFF
--- a/rust/main/agents/relayer/src/msg/pending_message.rs
+++ b/rust/main/agents/relayer/src/msg/pending_message.rs
@@ -338,11 +338,12 @@ impl PendingOperation for PendingMessage {
 
         // To avoid spending gas on a tx that will revert, dry-run just before submitting.
         if let Some(metadata) = self.metadata.as_ref() {
-            if let Err(_) = self
+            if self
                 .ctx
                 .destination_mailbox
                 .process_estimate_costs(&self.message, metadata)
                 .await
+                .is_err()
             {
                 return self.on_reprepare::<String>(None, ReprepareReason::ErrorEstimatingGas);
             }
--- a/rust/main/agents/relayer/src/msg/pending_message.rs
+++ b/rust/main/agents/relayer/src/msg/pending_message.rs
@@ -325,10 +325,10 @@ impl PendingOperation for PendingMessage {
     }
 
     #[instrument]
-    async fn submit(&mut self) {
+    async fn submit(&mut self) -> PendingOperationResult {
         if self.submitted {
             // this message has already been submitted, possibly not by us
-            return;
+            return PendingOperationResult::Success;
         }
 
         let state = self
@@ -336,6 +336,7 @@ impl PendingOperation for PendingMessage {
             .clone()
             .expect("Pending message must be prepared before it can be submitted");
 
+        // To avoid spending gas on a tx that will revert, dry-run just before submitting.
         if let Some(metadata) = self.metadata.clone() {
             if let Err(_) = self
                 .ctx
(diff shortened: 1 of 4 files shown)


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -269,7 +273,7 @@ impl PendingOperation for PendingMessage {
             .process_estimate_costs(&self.message, &metadata)
             .await
         {
-            Ok(metadata) => metadata,
```
tkporter, line 272: lol

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -321,17 +325,29 @@ impl PendingOperation for PendingMessage {
...
 
+        // To avoid spending gas on a tx that will revert, dry-run just before submitting.
+        if let Some(metadata) = self.metadata.clone() {
+            if let Err(_) = self
+                .ctx
+                .destination_mailbox
+                .process_estimate_costs(&self.message, &metadata)
+                .await
+            {
+                return self.on_reprepare::<String>(None, ReprepareReason::RevertedOrReorged);
```
tkporter, line 347: should this be `ErrorEstimatingGas`? as the doc for RevertedOrReorged seems like it relates to the tx actually landing on chain at some point `/// Delivery transaction reverted or reorged`

rust/main/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -197,6 +200,9 @@ pub enum ReprepareReason {
     #[strum(to_string = "Error estimating costs for process call")]
     /// Error estimating costs for process call
     ErrorEstimatingGas,
+    #[strum(to_string = "Simulating message in batch failed")]
+    /// Error simulating message in batch
+    BatchSimulationFailure,
```
tkporter, line 205: nit: is this used?

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -321,17 +325,29 @@ impl PendingOperation for PendingMessage {
...
+            return PendingOperationResult::Success;
         }
 
         let state = self
             .submission_data
             .clone()
             .expect("Pending message must be prepared before it can be submitted");
 
+        // To avoid spending gas on a tx that will revert, dry-run just before submitting.
+        if let Some(metadata) = self.metadata.clone() {
```
tkporter, line 340: does this need to be cloned or could you do `as_ref()`? as a reference to metadata is all that's needed later

tkporter (approved):
lgtm, nice

PR #4571, opened 2024-09-25
fix: gas oracle overhead calculation

paulbalaji:
fix: gas oracle overhead calculation

as spotted by @tkporter  https://canary.discord.com/channels/935678348330434570/1288601439006097498/1288602404950114334


tkporter (approved):
thank you 🙏

PR #4573, opened 2024-09-25
feat: add warp route monitoring for new Eclipse deploys

tkporter:
### Description

- Updates the monitoring script to support native balance getting and synthetic total supply getting
- Drive-by to update `tokenProgramPubKey` -> `tokenMintPubKey`, which is more accurate (the mint isn't a program)


paulbalaji (approved):
changes lgtm but would appreciate @Mo-Hussain having a gander too

typescript/infra/config/environments/mainnet3/warp/eclipse-SOL-deployments.yaml
```diff
@@ -0,0 +1,23 @@
+# Configs and artifacts for the deployment of Hyperlane Warp Routes
```
Mo-Hussain, line 1: Did you craft these manually?

Mo-Hussain: Is there a way I can double check these configs? where did it come from?

Mo-Hussain (approved):

typescript/infra/config/environments/mainnet3/warp/eclipse-SOL-deployments.yaml
tkporter: yeah I did, I crafted them using values from the registry. I've also ran with the configs so feel good they're accurate

PR #4297, opened 2024-08-12
feat: add arbitrumsepolia, basesepolia, ecotestnet, optimismsepolia, polygonamoy to core chains

paulbalaji:
Deploy to arbitrumsepolia, basesepolia, ecotestnet, optimismsepolia, polygonamoy

PR #4574, opened 2024-09-25
feat: sep 24 testnet deployments

paulbalaji:
feat: add more testnets
- `berabartio`
- `camptestnet`
- `formtestnet`
- `soneiumtestnet`
- `suavetoliman`

if RPCs get sorted:
- `citreatestnet`
- `hyperliquidevmtestnet`


typescript/infra/config/environments/testnet4/igp.ts
```diff
@@ -22,8 +22,6 @@ export const storageGasOracleConfig: AllStorageGasOracleConfigs =
     gasPrices,
     (local, remote) =>
       getTokenExchangeRateFromValues(local, remote, tokenPrices),
-    (local) => parseFloat(tokenPrices[local]),
```
paulbalaji, line 25: introduced in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4562 but we don't really need to provide these functions on testnet

typescript/infra/config/environments/testnet4/chains.ts
```diff
@@ -18,11 +18,6 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
       gasPrice: 8 * 10 ** 9, // 8 gwei
     },
   },
-  scrollsepolia: {
```
paulbalaji, line 21: introduced in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4297, have to remove to get any txs through (base fee was higher than this forced price)

typescript/infra/scripts/print-token-prices.ts
```diff
@@ -50,15 +58,26 @@ async function main() {
 
   const prices = objMap(ids, (_, id) => {
```
paulbalaji, line 59: avoid exiting completely if a coin is not on gecko or has no USD price - common on longer tail or testnets

typescript/infra/src/config/gas-oracle.ts
```diff
@@ -54,7 +54,8 @@ function getLocalStorageGasOracleConfig(
   return remotes.reduce((agg, remote) => {
     let exchangeRate = getTokenExchangeRate(local, remote);
     if (!gasPrices[remote]) {
-      throw new Error(`No gas price found for chain ${remote}`);
+      // Will run into this case when adding new chains
```
paulbalaji, line 57: avoid throwing completely if we're running this for the first time with new chains added that don't have config values yet

typescript/infra/src/config/gas-oracle.ts
```diff
@@ -230,6 +231,10 @@ export function getTokenExchangeRateFromValues(
   remote: ChainName,
   tokenPrices: ChainMap<string>,
 ): BigNumber {
+  if (!tokenPrices[local] || !tokenPrices[remote]) {
```
paulbalaji, line 234: some dependency issue here

goal is to provide some default value to satisfy the config on load, knowing it will get overwritten when the script actually gets run

typescript/infra/config/environments/testnet4/agent.ts
```diff
@@ -45,51 +44,70 @@ export const hyperlaneContextAgentChainConfig: AgentChainConfig<
     alfajores: true,
     arbitrumsepolia: true,
     basesepolia: true,
+    berabartio: true,
     bsctestnet: true,
+    camptestnet: true,
+    citreatestnet: false,
```
paulbalaji, line 50: keeping citrea/hyperliquid disabled for now til we get a nice RPC to work with

typescript/infra/config/environments/testnet4/validators.ts
```diff
@@ -173,72 +173,156 @@ export const validatorChainConfig = (
         'sepolia',
       ),
     },
-    plumetestnet: {
+    // plumetestnet: {
```
tkporter, line 176: any reason to keep these 2 around?

typescript/infra/config/environments/testnet4/validators.ts
```diff
@@ -173,72 +173,156 @@ export const validatorChainConfig = (
...
+          [Contexts.Hyperlane]: ['0x541dd3cb282cf869d72883557badae245b63e1fd'],
           [Contexts.ReleaseCandidate]: [],
           [Contexts.Neutron]: [],
         },
-        'optimismsepolia',
+        'berabartio',
       ),
     },
-    polygonamoy: {
+    // citreatestnet: {
```
tkporter, line 268: is this intentionally commented out? i know it's disabled but i see that `hyperliquidevmtestnet` isn't commented out and is also disabled

typescript/infra/scripts/print-token-prices.ts
tkporter: curious if we'd ever run into a mainnet chain where we can't get the price? i guess so

typescript/infra/src/config/gas-oracle.ts
tkporter: why wasn't this an issue in the past? do you think we should warn here just in case

typescript/infra/src/config/gas-oracle.ts
tkporter: can you put the reason into a comment so this doesn't get forgotten

tkporter (approved):

typescript/infra/scripts/print-token-prices.ts
paulbalaji: ya we have a couple atm e.g. https://www.coingecko.com/en/coins/real-ether

typescript/infra/src/config/gas-oracle.ts
paulbalaji: I probably introduced something with the recent refactor, gonna look into it

PR #4579, opened 2024-09-26
chore: try depot

paulbalaji:
chore: try depot 
- https://depot.dev/docs/container-builds/overview


paulbalaji:
monorepo build taking 38s is insane!

image

PR #4581, opened 2024-09-27
fix(contracts): InterchainAccountRouter minor audit remediation

aroralanuk:
### Description

- disabled the ICARouter's ability to change hook given that the user doesn't expect the hook to change after they deploy their ICA account. Hook is not part of the derivation like ism on the destination chain and hence, cannot be configured custom by the user.

### Drive-by changes

- MailboxClient events for hook and ism setting
- ProtocolFee events for setting beneficiary and owner

### Related issues

- partly fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/14

### Backward compatibility

No

### Testing

Unit tests


solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes, line 1: imo it would be quite simplifying if we just treated this like a normal router
theres not really a big cost of just deploying your own if you cant enroll on an existing set

yorhodes: but probably not the place to make this change

yorhodes (approved):
my mental model for events is to surface data that is cumulative/not a constant time view call
dont feel strongly but this seems superfluous
I dont think we will ever build an indexer for these?

aroralanuk:
> my mental model for events is to surface data that is cumulative/not a constant time view call dont feel strongly but this seems superfluous I dont think we will ever build an indexer for these?

is hookset not something we can use in the validator if say the merkle hook gets updated in the future?

yorhodes (changes requested):
all lgtm except the only initializing change, please revert

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -160,6 +161,12 @@ contract InterchainAccountRouter is Router {
         }
     }
 
+    function setHook(
+        address _hook
+    ) public override onlyInitializing onlyContractOrNull(_hook) onlyOwner {
+        hook = IPostDispatchHook(_hook);
+    }
```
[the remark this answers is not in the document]
aroralanuk: fixed

yorhodes (approved):

PR #4583, opened 2024-09-27
chore: testnet fixes

paulbalaji:
chore: testnet fixes


typescript/infra/config/environments/testnet4/tokenPrices.json
```diff
@@ -1,24 +1,23 @@
 {
-  "alfajores": "0.672791",
-  "arbitrumsepolia": "2580.94",
-  "basesepolia": "2580.94",
+  "alfajores": "10",
```
tkporter, line 2: rip

tkporter (approved):

typescript/infra/config/environments/testnet4/tokenPrices.json
paulbalaji: f around find out

PR #4593, opened 2024-09-30
chore: check if HEAD is behind main before deploying agents

paulbalaji:
chore: check if HEAD is behind main before deploying agents

image

image

### DIFF
--- a/typescript/infra/scripts/agents/deploy-agents.ts
+++ b/typescript/infra/scripts/agents/deploy-agents.ts
@@ -1,8 +1,48 @@
+import { confirm } from '@inquirer/prompts';
+import chalk from 'chalk';
+import { execSync } from 'child_process';
+
 import { createAgentKeysIfNotExists } from '../../src/agents/key-utils.js';
 import { HelmCommand } from '../../src/utils/helm.js';
 import { getConfigsBasedOnArgs } from '../core-utils.js';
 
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/agents/deploy-agents.ts
```diff
@@ -1,9 +1,49 @@
...
 import { getConfigsBasedOnArgs } from '../core-utils.js';
 
 import { AgentCli } from './utils.js';
 
+async function fetchLatestMain() {
+  try {
+    console.log(
+      chalk.grey.italic('Fetching latest changes from origin/main...'),
+    );
+    execSync('git fetch origin main', { stdio: 'inherit' });
```
daniel-savu, line 16: if this is ran from outside the monorepo with `yarn --cwd`, git would be unable to find the project. If git commands fail, we may just want to show a `confirm` warning rather than shutting down the process

daniel-savu (approved):

typescript/infra/scripts/agents/deploy-agents.ts
paulbalaji: oh dang, that's a little annoying

TIL `yarn --cwd` is a thing

daniel-savu: oh that's neat!

Issue #4513, opened 2024-09-18
feat(cli): Better chain selection


ltyu:
## Problem
Currently, some of commands in the CLI (`hyperlane warp init`) have the user multi-select chains. The UX is quite unintuitive. For example, the list is getting long so users need to scroll through the list to find the chain they want. As our list continues to grow users may be unclear which chain they've selected without scrolling back up. 

## Solution
Design and develop a better chain selection. This is open ended, and there are some discussions such as https://discord.com/channels/935678348330434570/1283108503161929779/1283166946711441589

One nice feature is being able to search.

Issue #4508, opened 2024-09-17
Bug: chain selection for warp init doesn't work on windows 


avious00:
Jake using Bash in Windows, unable to use arrow keys to move up or down to select chains for warp init

PR #4596, opened 2024-09-30
feat: better chain selection concept

xeno097:
### Description

This PR implements an updated view of the multi-chain selection step that now allows searching for chains in the current list 

#### Before:

image

image

#### After:

image

image

image

### Drive-by changes

- Updated the `runMultiChainSelectionStep` function to take as param an object instead of a list of params because the list was growing larger

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4513

### Backward compatibility

- Yes

### Testing

- Manual
  - Manual testing has also been conducted on Windows to see if https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4508 was solved. In this case, the following were discovered:
    - The chain selection is unusable on `gitbash`.
    - The chain selection works perfectly fine using `powershell` and `cmd`. I assume the issue is linked to how `gitbash` handles inputs or simulates a UNIX environment on Windows. CLI users on windows should use either one of these options


typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,402 @@ export async function inputWithInfo({
...
+
+// the isDownKey function from the inquirer package is not used
+// because it detects j and n as custom keybindings that cause
+// the option selection to go down instead of writing the letters
+// in the search string
+function isDownKey(key: KeypressEvent): boolean {
+  return key.name === 'down';
+}
+
+export const searchableCheckBox = createPrompt(
```
ltyu, line 248: I can't really tell that i'm able to search, until i start typing. Is there a good way to let the user know that it's searchable?

ltyu (commented):
Great progress.  I love that you thought about other potential use cases. my feedback is generally, I would prefer to keep thing simpler, until we need them

for example, i noticed that there are multiple help modes, but 1) it's unclear how to use them, and 2) if we'll ever need them because we only use `always` in `runMultiChainSelectionStep`

typescript/cli/src/utils/input.ts
xeno097: Adding a hint to the help message should be okay I think. We already do the same for telling the user to use tabs to select and enter to complete the selection. Examples:

Before
image

After
image

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,402 @@ export async function inputWithInfo({
...
+
+      return `${prefix} ${message} ${answer}`;
+    }
+
+    let helpTipTop = '';
+    let helpTipBottom = '';
+    if (
+      theme.helpMode === 'always' ||
+      (theme.helpMode === 'auto' &&
+        showHelpTip &&
```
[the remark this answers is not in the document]
xeno097: yep no problem I wanted to keep "compatibility" with the package we use for prompt that allows to choose between help modes but I can remove it no problem!

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,402 @@ export async function inputWithInfo({
...
+      const answer = theme.style.answer(
+        theme.style.renderSelectedChoices(selection, optionState.options),
+      );
+
+      return `${prefix} ${message} ${answer}`;
+    }
+
+    let helpTipTop = '';
+    let helpTipBottom = '';
+    if (
```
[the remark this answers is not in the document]
xeno097: yep exactly! let me simplify it i agree that it is hard to read after days of not looking at this

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,402 @@ export async function inputWithInfo({
...
+        (instructions === undefined || instructions))
+    ) {
+      if (typeof instructions === 'string') {
+        helpTipTop = theme.style.helpTip(instructions);
+      } else {
+        const keys = [
+          `${theme.style.key('tab')} to select`,
+          `and ${theme.style.key('enter')} to proceed`,
+        ];
+        helpTipTop = theme.style.helpTip(`(Press ${keys.join(', ')})`);
```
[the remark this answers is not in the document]
xeno097: yep!

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,393 @@ export async function inputWithInfo({
...
+      name,
+      short: choice.short ?? name,
+      description: choice.description,
+      disabled: choice.disabled ?? false,
+      checked: choice.checked ?? false,
+    };
+  });
+}
+
+function sortNormalizedItems<Value>(
```
jmrossy, line 188: FYI, localeCompare may be of use here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#sort_an_array

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,393 @@ export async function inputWithInfo({
...
+      loop = true,
+      required,
+      validate = () => true,
+      selectableOptionsSeparator,
+    } = config;
+    const theme = makeTheme<SearchableCheckboxTheme>(
+      checkboxTheme,
+      config.theme,
+    );
+    const firstRender = useRef(true);
```
jmrossy, line 265: So strange to see these react hooks here. I wondered if inquirer was using react under the hood but actually they implemented these themselves! https://github.com/SBoudrias/Inquirer.js/blob/main/packages/core/src/lib/use-memo.ts

typescript/cli/src/utils/input.ts
```diff
@@ -53,3 +71,393 @@ export async function inputWithInfo({
...
+      error = `${theme.style.error(errorMsg)}`;
+    } else if (
+      optionState.options.length === 0 &&
+      searchTerm !== '' &&
+      status === 'idle'
+    ) {
+      error = theme.style.error('No results found');
+    }
+
+    return `${prefix} ${message}${helpTipTop} ${searchTerm}\n${page}${helpTipBottom}${choiceDescription}${error}${ansiEscapes.cursorHide}`;
```
jmrossy, line 461: This function is very long. would it be possible to break it up into more digestible chunks?

jmrossy:
Mostly looks good to me so far!

typescript/cli/src/utils/input.ts
xeno097: yep I'll try to split it!

ltyu (approved):
LGTM!

PR #4598, opened 2024-10-01
fix: reuse SDK transaction types in submitters

paulbalaji:
### Description

Updating the types in the submitters because they were using their own types instead of the existing ones for some reason. Also adding `chain` as an argument to the `Annotated` type, so that we can take advantage of this in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4599 and defend ourselves against futures where chainid/domainid are not the same for EVM

This also fixes the issue currently where there's a small type mismatch between submitters and modules

### Drive-by changes

- move `createTransferOwnershipTx` from the abstract hyperlane module to the evm module deployer, as it's EVM specific 

### Related issues

useful for https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4599, but also something that needs to be fixed itself

### Backward compatibility

Yes

### Testing

ci

### DIFF
--- a/typescript/utils/src/types.ts
+++ b/typescript/utils/src/types.ts
@@ -114,5 +114,6 @@ export type ParsedLegacyMultisigIsmMetadata = {
 };
 
 export type Annotated<T> = T & {
+  chain: string; // TODO: Change to ChainName after moving to SDK from Utils
   annotation?: string;
 };
--- a/typescript/cli/src/config/submit.ts
+++ b/typescript/cli/src/config/submit.ts
@@ -1,12 +1,9 @@
 import { stringify as yamlStringify } from 'yaml';
 
 import {
-  PopulatedTransactions,
-  PopulatedTransactionsSchema,
+  AnnotatedEV5Transaction,
   SubmissionStrategy,
 } from '@hyperlane-xyz/sdk';
-import { PopulatedTransaction } from '@hyperlane-xyz/sdk';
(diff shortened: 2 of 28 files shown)


typescript/utils/src/types.ts
```diff
@@ -113,5 +113,6 @@ export type ParsedLegacyMultisigIsmMetadata = {
 };
 
 export type Annotated<T> = T & {
+  chain: string; // TODO: Change to ChainName after moving to SDK from Utils
```
yorhodes, line 116: bit confused by this as the transactions chain ID should always map 1:1 to a chain name

paulbalaji: but if we have two chains with the same chainid, we no longer know which one we're submitting to from the chainId alone

paulbalaji: here's an example in the duplicate-chainids pr where having the chain directly is helpful https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4599/files#diff-d572648c3e0dd7a6d4eff6670862ae1bcb5f1c1f3feb8c0d70fb60a765ae7d48

xeno097 (approved):

typescript/utils/src/types.ts
yorhodes: oh I see, sorry I thought this was to different domain IDs

paulbalaji:
closing PR in favor of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4642

PR #4599, opened 2024-10-01
feat: support duplicate chainIds with unique domainIds

paulbalaji:
### Description

This PR serves as a spike, not intended (atm) to be merged.

The problem has surfaced twice now of multiple chains reusing the same chainid:
	- celo classic testnet & celo L2 testnet
	- molten & shape

This PR allows the set of chain metadata to include unique chains/domains that share the same chainId under the hood. This required changes in multiple places where we assumed that the domainId/chainId would be the same for EVM chains (no such assumption was made on other EVMs).

Touches:
- evm core/ism/hook/warp modules
- multiprovider + chain metadata manager
- safe service
- renaming `ChainNameOrId` to `ChainNameOrDomain` (where bulk of the diff comes from)

### Drive-by changes

- since we now provide the `chain` as part of an annotated tx, we can simplify the `sendTransaction` call to one argument

### Related issues

https://www.notion.so/hyperlanexyz/Shape-Molten-Chain-ID-Clash-1156d35200d6806db7d7dc62a95c8085

### Backward compatibility

no

### Testing

ci

### DIFF
--- a/.changeset/thin-bananas-explain.md
+++ b/.changeset/thin-bananas-explain.md
@@ -1,0 +1,7 @@
+---
+'@hyperlane-xyz/utils': major
+'@hyperlane-xyz/cli': major
+'@hyperlane-xyz/sdk': major
+---
+
+Support duplicate chainIds with unique domainIds
--- a/typescript/sdk/src/core/EvmCoreModule.ts
+++ b/typescript/sdk/src/core/EvmCoreModule.ts
@@ -17,6 +17,7 @@ import {
 } from '../contracts/types.js';
 import { DeployedCoreAddresses } from '../core/schemas.js';
 import { CoreConfig } from '../core/types.js';
+import { EvmModuleDeployer } from '../deploy/EvmModuleDeployer.js';
 import { HyperlaneProxyFactoryDeployer } from '../deploy/HyperlaneProxyFactoryDeployer.js';
 import {
   ProxyFactoryFactories,
@@ -137,6 +138,7 @@ export class EvmCoreModule extends HyperlaneModule<
         data: contractToUpdate.interface.encodeFunctionData('setDefaultIsm', [
           deployedIsm,
         ]),
+        chain: this.chainName,
       });
     }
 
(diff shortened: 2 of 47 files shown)


typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -124,6 +124,7 @@ export class EvmIsmModule extends HyperlaneModule<
 
     this.chain = this.multiProvider.getChainName(this.args.chain);
     this.domainId = this.multiProvider.getDomainId(this.chain);
+    this.chainId = this.multiProvider.getEvmChainId(this.chain);
```
xeno097, line 127: I think the `domainId` field can be removed because I don't see it being used anymore in this class.

xeno097 (commented):
Just leaving some small suggestions here

typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -47,9 +47,8 @@ export class EvmCoreModule extends HyperlaneModule<
   protected coreReader: EvmCoreReader;
   public readonly chainName: string;
 
-  // We use domainId here because MultiProvider.getDomainId() will always
-  // return a number, and EVM the domainId and chainId are the same.
   public readonly domainId: Domain;
+  public readonly chainId: Domain;
```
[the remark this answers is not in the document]
paulbalaji: nice catch yeah this is a typo from me

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: good spot

typescript/sdk/src/types.ts
```diff
@@ -1,12 +1,12 @@
...
-import type { ChainId, Domain } from '@hyperlane-xyz/utils';
+import type { Domain } from '@hyperlane-xyz/utils';
 
 // An alias for string to clarify type is a chain name
 export type ChainName = string;
 // A map of chain names to a value type
 export type ChainMap<Value> = Record<string, Value>;
 
-export type ChainNameOrId = ChainName | ChainId | Domain;
+export type ChainNameOrId = ChainName | Domain;
```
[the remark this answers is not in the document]
paulbalaji: I considered it but worried about making the diff too big. I'll put it into the xeno code review commit for easier reviewing

xeno097 (approved):

.changeset/thin-bananas-explain.md
```diff
@@ -0,0 +1,7 @@
+---
+'@hyperlane-xyz/utils': minor
+'@hyperlane-xyz/cli': minor
+'@hyperlane-xyz/sdk': minor
+---
+
+Support duplicate chainIds with unique domainIds
```
paulbalaji, line 7: these are probably major actually

PR #4605, opened 2024-10-02
fix(CLI): do not double info log approval requirement

yorhodes:
### Description

Got this log during `warp send`

```
Approval is required for transfer of deUSD
Approval required for transfer of deUSD
Approval is required for transfer of deUSD
Approval required for transfer of deUSD
```

### DIFF
--- a/typescript/sdk/src/warp/WarpCore.ts
+++ b/typescript/sdk/src/warp/WarpCore.ts
@@ -528,7 +528,7 @@ export class WarpCore {
       token.addressOrDenom,
       amount,
     );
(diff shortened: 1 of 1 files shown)


paulbalaji (approved):

paulbalaji (commented):
for other reviewers, you can find the other log https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/1cd4fe4792659cbb14c710d290d425059c82bb0d/typescript/sdk/src/warp/WarpCore.ts#L329

PR #4609, opened 2024-10-02
feat: HyperlaneDb trait

daniel-savu:
### Description

Defines a `HyperlaneDb` trait for use in the validator submitter, so a mock can be passed in to simulate indexed data from reorged state.

Although the `HyperlaneDb` trait can't be `Clone` to be object safe, as long as we use `Arc`s everywhere, all will be ok. Eventually the aim is to replace all direct usage of `HyperlaneRocksDb` with `Arc<dyn HyperlaneDb>`.

### DIFF
--- a/rust/main/config/testnet_config.json
+++ b/rust/main/config/testnet_config.json
@@ -1446,6 +1446,194 @@
       "index": {
         "from": 1921514
       }
+    },
+    "test1": {
+      "blockExplorers": [
+        {
+          "apiKey": "fakekey",
+          "apiUrl": "https://api.etherscan.io/api",
+          "family": "etherscan",
+          "name": "Etherscan",
+          "url": "https://etherscan.io"
+        }
+      ],
+      "blocks": {
+        "confirmations": 1,
+        "estimateBlockTime": 3,
+        "reorgPeriod": 0
+      },
+      "chainId": 9913371,
+      "displayName": "Test 1",
+      "domainId": 9913371,
+      "isTestnet": true,
+      "name": "test1",
+      "nativeToken": {
+        "decimals": 18,
+        "name": "Ether",
+        "symbol": "ETH"
+      },
+      "protocol": "ethereum",
+      "rpcUrls": [
+        {
+          "http": "http://127.0.0.1:8545"
+        }
+      ],
+      "aggregationHook": "0x7F54A0734c5B443E5B04cc26B54bb8ecE0455785",
+      "domainRoutingIsm": "0xb0279Db6a2F1E01fbC8483FCCef0Be2bC6299cC3",
+      "fallbackRoutingHook": "0x99bbA657f2BbC93c02D617f8bA121cB8Fc104Acf",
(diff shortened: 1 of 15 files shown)


ameten:
Is the aim eventually to make `HyperlaneSqlDb` to implement `HyperlaneDb` trait?

rust/main/hyperlane-base/src/db/error.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Error opening the database
+    OpeningError {
+        /// Rocksdb error during opening
+        #[source]
+        source: rocksdb::Error,
+        /// Raw database path provided
+        path: PathBuf,
+        /// Parsed path used
+        canonicalized: PathBuf,
+    },
```
ameten, line 21: I wonder if we should hide the error from a particular type of database and expose a generalised error to the user of `HyperlaneDb` trait?

ameten (approved):

rust/main/hyperlane-base/src/db/storage_types.rs
```diff
@@ -8,15 +8,18 @@ use hyperlane_core::{
...
+pub struct InterchainGasPaymentData {
+    /// The amount of tokens paid for the gas.
     pub payment: U256,
+    /// The amount of gas paid for.
     pub gas_amount: U256,
 }
 
 /// Subset of `InterchainGasExpenditure` excluding the message id which is
 /// stored in the key.
+#[allow(missing_docs)]
```
tkporter, line 20: just wanna make sure this is intentional as it seems like some docs were added

tkporter (approved):

rust/main/config/testnet_config.json
```diff
@@ -1446,6 +1446,194 @@
       "index": {
         "from": 1921514
       }
+    },
+    "test1": {
```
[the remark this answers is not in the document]
daniel-savu: they're ephemeral but deterministic, so thought it's reasonable to add them

rust/main/hyperlane-base/src/db/storage_types.rs
daniel-savu: I wasn't fully sure what the fields mean here tbh and I preferred under-documenting that mis-documenting

rust/main/hyperlane-base/src/db/error.rs
daniel-savu: `DbError` is generically named, and its members could be anything based on specific implementations - `DbError::RocksError` can only be created from a rocksdb error so it won't be misused. 

`DbError` _is_ made fully generalised when it's converted to `ChainCommunicationError`, and I considered defining the error as an associated type, but usage ends up looking _more_ concrete, e.g. `Arc<dyn HyperlaneDb<DbError = RocksDbError>>`

Due to that I'm not sure it's worth generalising further in the current state, could you give an example?

ameten: If we just converting into a different error, we don't need to generalise. If we would like to react on the error in a meaningful way, we may want to generalise it.

PR #4610, opened 2024-10-02
feat: include `id` field in `list_operations` endpoint

daniel-savu:
### Description

The `id` field was missing if the serialized `PendingOperation` was a `PendingMessage`, which made it hard to look up the message in relayer logs or the explorer. This PR takes advantage of `id()` being a required method on `PendingOperation` to derive and display the ID as part of endpoint logic.

### Testing

Unit Tests

### DIFF
--- a/rust/main/agents/relayer/src/server/list_messages.rs
+++ b/rust/main/agents/relayer/src/server/list_messages.rs
@@ -3,8 +3,8 @@ use axum::{
     routing, Router,
 };
 use derive_new::new;
-use hyperlane_core::QueueOperation;
-use serde::Deserialize;
+use hyperlane_core::{QueueOperation, H256};
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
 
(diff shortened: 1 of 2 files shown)


rust/main/agents/relayer/src/server/list_messages.rs
```diff
@@ -108,15 +123,40 @@ mod tests {
...
+                .with_id(H256::from_str(id_2).unwrap()),
+        ) as QueueOperation;
+        let expected_response = r#"[
+  {
+    "id": "0x1acbee9798118b11ebef0d94b0a2936eafd58e3bfab91b05da875825c4a1c39b",
+    "operation": {
+      "destination_domain": {
+        "Known": "Arbitrum"
+      },
+      "id": "0x1acbee9798118b11ebef0d94b0a2936eafd58e3bfab91b05da875825c4a1c39b",
```
daniel-savu, line 143: the reason there already is an `id` in here is because it's a field on `MockOperation` - that field is missing on `HyperlaneMessage` because it's derived, so now we'll have it explicit

daniel-savu:
For whoever ends up extending the endpoint: to add tracking for status transition history and IGP expenditure, they must be added as serializable fields on the `PendingMessage` struct and `typetag` will do its magic to reflect them in the serialized `PendingOperation`

rust/main/agents/relayer/src/server/list_messages.rs
tkporter: makes sense, a comment may be useful but nbd

tkporter (approved):
nice

PR #4613, opened 2024-10-03
feat: sept 30 mainnet batch

paulbalaji:
### Description

Deploy to immutablezkevm rari rootstock alephzeroevm chiliz lumia superposition

### Drive-by changes

- fix igp defaults
- helper for updating agent configs + running prettier

### Related issues

na

### Backward compatibility

ye

### Testing

manual


jmrossy (dismissed):

typescript/infra/config/environments/mainnet3/igp.ts
```diff
@@ -31,7 +31,7 @@ const storageGasOracleConfig: AllStorageGasOracleConfigs =
     (local, remote) =>
       getTokenExchangeRateFromValues(local, remote, tokenPrices),
     (local) => parseFloat(tokenPrices[local]),
-    getOverheadWithOverrides,
+    (local, remote) => getOverheadWithOverrides(local, remote),
```
daniel-savu, line 34: is this change needed?

typescript/infra/config/environments/mainnet3/validators.ts
```diff
@@ -896,5 +896,76 @@ export const validatorChainConfig = (
         'oortmainnet',
       ),
     },
+
+    immutablezkevm: {
+      interval: 5,
+      reorgPeriod: getReorgPeriod('immutablezkevm'),
+      validators: validatorsConfig(
+        {
+          [Contexts.Hyperlane]: ['0xa787c2952a4d22f776ee6e87e828e6f75de24330'],
```
daniel-savu, line 905: I assume onboarding other validators comes later?

typescript/infra/src/config/gas-oracle.ts
```diff
@@ -236,16 +236,13 @@ export function getTokenExchangeRateFromValues(
   // Workaround for chicken-egg dependency problem.
   // We need to provide some default value here to satisfy the config on initial load,
   // whilst knowing that it will get overwritten when a script actually gets run.
-  if (!tokenPrices[local] || !tokenPrices[remote]) {
-    return BigNumber.from(1);
```
daniel-savu, line 240: my read here is that if either price is undefined, we want to return an exchange rate of 1. The new logic seems to have a new case where e.g. `tokenPrices[local]` can be undefined and `tokenPrices[remote]` is defined, and we'd return `1 / tokenPrices[local]` with the added premium. Is this intended?

daniel-savu (approved):

typescript/infra/config/environments/mainnet3/validators.ts
paulbalaji: correct - our deployment process is split into two parts:
1. initial contracts + AW agents
2. enrolling validators + updating ISMs

ameten (approved):

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -129,6 +137,7 @@ export const hyperlaneContextAgentChainConfig: AgentChainConfig<
     bsc: true,
     celo: true,
     cheesechain: true,
+    chiliz: true,
```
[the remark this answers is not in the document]
paulbalaji: no we're going to keep it live to test some tx overrides

typescript/infra/config/environments/mainnet3/igp.ts
paulbalaji: not strictly needed, but keeping for clarity/consistency with other function args. I intended to make it this way in the original PR, but fixing it now with a drive-by

typescript/infra/src/config/gas-oracle.ts
paulbalaji: yup this is intended. without this change I was getting underflow/overflow errors on some exchange pairings.

It's a weird one because this case of tokenPrices[X] being undefined is only on the initial config loading, which is quickly overriden when `print-token-prices` is ran. The root problem here is a dependency issue between the config/jsons/scripts that I've not had the time to fix yet

daniel-savu: would be useful to add a comment or TODO to clarify

paulbalaji: there is just above the diff https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/449945d53d10cadb84fa65604507ce9cfbdcbeb3/typescript/infra/src/config/gas-oracle.ts#L236-L238

paulbalaji: I'll add a comment about the under/overflow too

PR #4618, opened 2024-10-03
feat: Generate Cosmos address for Injective

ameten:
### Description

Injective uses different logic compared to Neutron and Osmosis to generate account addresses from public key. It is implemented in this PR so that sender address is correctly generated from signer info of transaction.

Depending on the type of public key communicated in the transaction, we shall generate either Bitcoin-like or Ethereum-like address for Injective. 

We can also specify the type of account id for the signer. Bitcoin-like will be chosen by default, so, we don't need to change the configuration of Relayer.

### Drive-by changes

Made some errors to be warnings. We'll add metrics to measure how frequent the issues are and decide how to fix them:
1. transaction contains multiple contract execution messages
2. transaction contains fees in unsupported denominations

Made Scraper to refrain from recording transaction which it cannot parse properly.

### Backward compatibility

Yes

### Testing

Added unit tests
Manual testing with Injective blockchain


rust/main/hyperlane-base/src/settings/parser/mod.rs
```diff
@@ -332,9 +332,16 @@ fn parse_signer(signer: ValueParser) -> ConfigResult<SignerConf> {
                 .get_key("prefix")
                 .parse_string()
                 .unwrap_or_default();
+            let account_id_type = signer
+                .chain(&mut err)
+                .get_opt_key("account_id_type")
+                .parse_string()
+                .end()
+                .unwrap_or("BITCOIN");
```
tkporter, line 340: nit: to get feedback on the parsing here instead of later on, I think we could have the `account_id_type` on `SignerConf::CosmosKey` be a `AccountIdType` directly instead of a `String`, and then we can do
```
let account_id_type = signer
                .chain(&mut err)
                .get_opt_key("account_id_type")
                .parse_from_string()
                .unwrap_or(AccountIdType::Bitcoin);
```
to parse into the enum here.

Similar to how it works up above for the AWS region. I think it may require `FromStr`/`str::parse` on `AccountIdType`

rust/main/chains/hyperlane-cosmos/src/libs/account_id_type.rs
```diff
@@ -0,0 +1,40 @@
+use crate::HyperlaneCosmosError;
+
+/// Specifies the account id (address) type
+pub enum AccountIdType {
```
tkporter, line 4: nit: We use `strum`'s `EnumString` ([docs](https://docs.rs/strum/latest/strum/derive.EnumString.html#example-howto-use-enumstring)) elsewhere in the code that could be helpful here instead of the manual impls. Example here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/11e13e56f080e5b533e0fd53b8cddf153a9e85e7/rust/main/hyperlane-core/src/chain.rs#L159

Don't feel strongly though if you'd prefer not to bring that dep into this crate

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
```diff
@@ -178,12 +241,17 @@ impl CosmosProvider {
...
-            error!(
+            let msg = "transaction contains fees in unsupported denominations, manual intervention is required";
+            warn!(
                 ?tx_hash,
                 ?supported_denomination,
                 ?unsupported_denominations,
-                "transaction contains fees in unsupported denominations, manual intervention is required");
+                msg,
+            );
+            Err(ChainCommunicationError::CustomError(msg.to_owned()))?
```
tkporter, line 251: I'm now remembering more details of what I poorly documented in https://github.com/hyperlane-xyz/issues/issues/1369, flagging just in case we're not on the same page

Maybe this is your understanding as well but the issue is that if we propagate an Err here, it propagates very far back and prevents the indexed data (e.g. the messages indexed in a range), and any other data in that block range batch, from being written to the DB. If you trace back the call sites all the way to `store_logs` it'll be more clear

So i think this is fine as long as we accept that all data in batches where a tx has this problem won't get written, and we plan to revisit this

tkporter (approved):
🫡

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
```diff
@@ -146,10 +210,9 @@ impl CosmosProvider {
 
         let contract_execution_messages_len = contract_execution_messages.len();
         if contract_execution_messages_len > 1 {
-            error!(
-                ?tx_hash,
-                ?contract_execution_messages,
-                "transaction contains multiple contract execution messages, we are indexing the first entry only");
+            let msg = "transaction contains multiple contract execution messages, we are indexing the first entry only";
```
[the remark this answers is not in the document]
ameten: No, I have not seen this error.

rust/main/hyperlane-base/src/settings/parser/mod.rs
ameten: Fixed

rust/main/chains/hyperlane-cosmos/src/libs/account_id_type.rs
ameten: Fixed

rust/main/chains/hyperlane-cosmos/src/providers/cosmos/provider.rs
ameten: I agree that we can start with such implementation: refrain from storing logs from transaction we cannot parse properly. It will allows us to add the support for missing features and re-visit the transaction without a need for a manual backfill of data.

I created a few issues preliminary: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4632, https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4633, https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4634

PR #4627, opened 2024-10-04
wip: do not start up validator if fraud was reported

daniel-savu:


rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -400,7 +400,7 @@ impl BaseMetadataBuilder {
                     continue;
                 }
 
-                match config.build(None).await {
+                match config.build_and_validate(None).await {
```
daniel-savu, line 403: A really nice side-effect of panicking in `CheckpointSyncerConf::build_and_validate` when a reorg flag is detected is that the relayer's `submit` tokio task for that chain will shut down as well, preventing further processing until the reorg is resolved, which I thought would be much harder to achieve:

https://www.notion.so/hyperlanexyz/Agent-Halting-Via-Fraud-Flag-10e6d35200d680dfbe28f266cc0b5ec2?pvs=4#10e6d35200d680a79f47d17c2434145d

ameten (approved):

PR #4628, opened 2024-10-04
feat(warpChecker): Support Renzo PZETH route

Mo-Hussain:
### Description

- support Renzo PZETH warp route in checker

### Drive-by changes

- introduce `CheckerWarpRouteIds` enum for warp routes that are supported in the checker, eclipse and sol routes are not currently supported

### Testing

Manual

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/warpIds.ts
+++ b/typescript/infra/config/environments/mainnet3/warp/warpIds.ts
@@ -17,19 +17,3 @@ export enum WarpRouteIds {
   InevmInjectiveINJ = 'INJ/inevm-injective',
   MantapacificNeutronTIA = 'TIA/mantapacific-neutron',
 }
-
-// add new warp route ids here if they are supported by the checker tooling, add an entry to warpConfigGetterMap also
-export enum CheckerWarpRouteIds {
-  Ancient8EthereumUSDC = WarpRouteIds.Ancient8EthereumUSDC,
-  ArbitrumBaseBlastBscEthereumFraxtalLineaModeOptimismSeiTaikoZircuitEZETH = WarpRouteIds.ArbitrumBaseBlastBscEthereumFraxtalLineaModeOptimismSeiTaikoZircuitEZETH,
-  ArbitrumNeutronEclip = WarpRouteIds.ArbitrumNeutronEclip,
-  ArbitrumNeutronTIA = WarpRouteIds.ArbitrumNeutronTIA,
-  EthereumInevmUSDC = WarpRouteIds.EthereumInevmUSDC,
-  EthereumInevmUSDT = WarpRouteIds.EthereumInevmUSDT,
-  EthereumVictionETH = WarpRouteIds.EthereumVictionETH,
-  EthereumVictionUSDC = WarpRouteIds.EthereumVictionUSDC,
-  EthereumVictionUSDT = WarpRouteIds.EthereumVictionUSDT,
-  EthereumZircuitPZETH = WarpRouteIds.EthereumZircuitPZETH,
-  InevmInjectiveINJ = WarpRouteIds.InevmInjectiveINJ,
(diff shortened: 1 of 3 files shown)


typescript/infra/config/environments/mainnet3/warp/warpIds.ts
```diff
@@ -12,6 +12,23 @@ export enum WarpRouteIds {
   EthereumVictionETH = 'ETH/ethereum-viction',
   EthereumVictionUSDC = 'USDC/ethereum-viction',
   EthereumVictionUSDT = 'USDT/ethereum-viction',
+  EthereumZircuitPZETH = 'PZETH/ethereum-zircuit',
   InevmInjectiveINJ = 'INJ/inevm-injective',
   MantapacificNeutronTIA = 'TIA/mantapacific-neutron',
 }
+
+// add new warp route ids here if they are supported by the checker tooling, add an entry to warpConfigGetterMap also
+export enum CheckerWarpRouteIds {
```
Mo-Hussain, line 21: I split this out into a separate enum as not all warp routes are support by the checker yet

tkporter: do we need to have `WarpRouteIds` as well? I'd be happy to just have a single enum that covers the warp route IDs that we want to check

Or is the difference that we may want monitoring on some warp routes but not checking?

Just feeling like the web of warp IDs is already a bit tedious so it'd be nice to not need another place where we write all these long names lol

typescript/infra/scripts/check/check-warp-deploy.ts
```diff
@@ -25,7 +25,7 @@ async function main() {
   const failedWarpRoutesChecks: string[] = [];
 
   // TODO: consider retrying this if check throws an error
-  for (const warpRouteId of Object.values(WarpRouteIds)) {
+  for (const warpRouteId of Object.values(CheckerWarpRouteIds)) {
```
tkporter, line 28: extending my previous comment, if the only use of the new enum is for this change, we could do `Object.keys(warpConfigGetterMap)`

typescript/infra/config/environments/mainnet3/warp/warpIds.ts
```diff
@@ -1,6 +1,6 @@
 export enum WarpRouteIds {
   Ancient8EthereumUSDC = 'USDC/ancient8-ethereum',
-  ArbitrumBaseBlastBscEthereumFraxtalLineaModeOptimismZircuitEZETH = 'EZETH/arbitrum-base-blast-bsc-ethereum-fraxtal-linea-mode-optimism-zircuit',
+  ArbitrumBaseBlastBscEthereumFraxtalLineaModeOptimismSeiTaikoZircuitEZETH = 'EZETH/arbitrum-base-blast-bsc-ethereum-fraxtal-linea-mode-optimism-sei-taiko-zircuit',
```
tkporter, line 3: not sure if you saw Le's recent PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4695 that does this too

tkporter (approved):

typescript/infra/scripts/check/check-warp-deploy.ts
Mo-Hussain: I'll do this instead.

We access `WarpRouteIds` here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/389fd91ec6f00023ee4f1b71f85e0eb9ebd5c4d9/typescript/infra/config/environments/mainnet3/agent.ts#L358

And we don't support `EthereumEclipseUSDC` yet in the checker

typescript/infra/config/environments/mainnet3/warp/warpIds.ts
Mo-Hussain: I did this on the plane and hadn't seen his change lol

PR #4629, opened 2024-10-05
feat: Add block tag support in `reorgPeriod` parameter

apastushenka:
### Description

Allows setting an arbitrary string in `reorgPeriod`, which is used as a block tag to get the finalized block. Currently, only the Ethereum connector supports this feature with the following tags:
- finalized
- safe
- pending
- latest
- earliest

This is currently useful for EVM-compatible chains that don't have a fixed reorganization period, but instead rely on the block tag `Finalized` to indicate finality. For such chains, you should set `reorgPeriod` to `finalized` instead of a number of blocks.

### Backward compatibility

Yes


paulbalaji:
please run `yarn changeset add` to add a changeset :)

rust/main/hyperlane-core/src/chain.rs
```diff
@@ -549,4 +626,32 @@ mod tests {
...
+
+    #[test]
+    fn parse_reorg_period() {
+        assert_eq!(
+            serde_json::from_value::<ReorgPeriod>(0.into()).unwrap(),
+            ReorgPeriod::None
+        );
+
+        assert_eq!(
+            serde_json::from_value::<ReorgPeriod>("0".into()).unwrap(),
```
daniel-savu, line 638: also `ReorgPeriod::Block(0)`

rust/main/hyperlane-core/src/chain.rs
```diff
@@ -39,6 +42,80 @@ impl<'a> std::fmt::Display for ContractLocator<'a> {
...
+        }
+    }
+
+    pub fn is_none(&self) -> bool {
+        matches!(self, ReorgPeriod::None)
+    }
+}
+
+impl Serialize for ReorgPeriod {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
daniel-savu, line 74: IMO, `ReorgPeriod::Tag` should be defined as an enum. Also the serialization should look like the one below, which I believe you can configure `derive(Serialize)` to give you with the flattening flag

```
            ReorgPeriod::None => "None"
            ReorgPeriod::Blocks(blocks) => blocks.to_string(),
            ReorgPeriod::Tag(tag) => tag.to_string()
```

Then to deserialize first you try to find the string in the predefined `Tag` enum, then match it against the `None` string, or if it's a number it's parsed as blocks (probably harder to define via the `derive`)

rust/main/hyperlane-core/src/chain.rs
```diff
@@ -549,4 +626,32 @@ mod tests {
         );
         assert!("foo".parse::<KnownHyperlaneDomain>().is_err());
     }
+
+    #[test]
+    fn parse_reorg_period() {
+        assert_eq!(
+            serde_json::from_value::<ReorgPeriod>(0.into()).unwrap(),
+            ReorgPeriod::None
```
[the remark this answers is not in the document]
apastushenka: The reason why `0` is parsed to `ReorgPeriod::None` is to maintain compatibility with current configs, where `"reorgPeriod": 0` means no reorg period. Also, `ReorgPeriod::Block(0)` is supposed to be an invalid variant because it’s the same as `ReorgPeriod::None`.

rust/main/hyperlane-core/src/chain.rs
sapph1re: @daniel-savu it was @tkporter's [idea to have it as a string](https://discord.com/channels/935678348330434570/1285726511793704971/1288917019550289941):

> so I'd prefer
> 
> enum ReorgConfiguration {
>   Blocks(u64),
>   Tag(String),
> }
> 
> instead of
> 
> enum ReorgConfiguration {
>   Blocks(u64),
>   Tag(TagType),
> }
> 
> enum TagType {
>   Finalized,
>   Safe
> }

tkporter: Yeah as this type is meant to be protocol agnostic I think having it be as flexible as possible is in our favor. Otherwise I guess we'd need this to be a union of all protocol supported tags, which feels a bit awk

My original motivation though was hoping that ethers-rs would let us pass in arbitrary tags because I've seen some chains use new tags outside of the typical ones, but sadly it looks like ethers-rs has their own enum they force you to use. But maybe on other chains this won't always be the case

rust/main/hyperlane-core/src/chain.rs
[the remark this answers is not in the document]
tkporter: +1 would prefer 0 to be ReorgPeriod::None, specifically for the case of lagged calls where we only want to do the whole "get the block number and then subtract the reorg period" bit for non-zero reorg periods.

We could maybe change that logic to be if the reorg period blocks is > 0 instead of `is_none` if you think having `None` as an option is confusing though. But I think having None as a variant can be useful

typescript/sdk/src/metadata/chainMetadata.test.ts
```diff
@@ -62,6 +62,16 @@ describe('ChainMetadataSchema', () => {
         grpcUrls: [],
       }),
     ).to.eq(true);
+
+    expect(
+      isValidChainMetadata({
+        ...minimalSchema,
+        blocks: {
+          confirmations: 1,
+          reorgPeriod: 'finalized',
```
tkporter, line 71: 😎 nice

rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -219,20 +220,25 @@ where
...
-    if let Some(lag) = maybe_lag {
-        let fixed_block_number: BlockNumber = provider
-            .get_block_number()
-            .await
-            .map_err(ChainCommunicationError::from_other)?
-            .saturating_sub(lag.get().into())
-            .into();
-        Ok(call.block(fixed_block_number))
+    if !lag.is_none() {
+        let reorg_period = EthereumReorgPeriod::try_from(lag)?;
```
tkporter, line 230: it's a bit of a shame we need to do this conversion so often when most of the ethereum structs have a `reorg_period: EthereumReorgPeriod` readily available

I guess the main reason is that the chain agnostic `Mailbox` and `MerkleTreeHook` traits from `hyperlane-core` require the reorg period to be passed in for functions like `count` / `tree` / `latest_checkpoint` etc

Iirc we originally did this to underscore how important it was that those functions only use finalized state. @daniel-savu curious what you'd think of removing those params from the chain agnostic traits and requiring implementations to handle finality concerns internally?

I'm happy if we continue as it is like this, but something about this feels a bit weird

tkporter (approved):
Great job, really exciting to see such a meaningful contribution like this :)

I'll leave the final green light to @daniel-savu

rust/main/chains/hyperlane-cosmos/src/utils.rs
```diff
@@ -24,20 +24,25 @@ pub(crate) const CONTRACT_ADDRESS_ATTRIBUTE_KEY: &str = "_contract_address";
...
-        Some(lag) => {
+    let block_height = match reorg_period {
+        ReorgPeriod::Blocks(blocks) => {
             let tip = provider.latest_block_height().await?;
-            let block_height = tip - lag.get();
+            let block_height = tip - blocks.get() as u64;
             Some(block_height)
         }
-        None => None,
+        ReorgPeriod::None => None,
```
daniel-savu, line 40: should this return `provider.latest_block_height().await?`?

rust/main/chains/hyperlane-cosmos/src/utils.rs
```diff
@@ -24,20 +24,25 @@ pub(crate) const CONTRACT_ADDRESS_ATTRIBUTE_KEY: &str = "_contract_address";
...
+        ReorgPeriod::Tag(_) => {
+            return Err(ChainCommunicationError::CustomError(
+                "Cosmos does not support reorg period as a tag".into(),
```
daniel-savu, line 43: would be useful to add an entry to [HyperlaneCosmosError](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/02a5b92ba774c3ce8c706966c9b9f23272215fa5/rust/main/chains/hyperlane-cosmos/src/error.rs#L13) and have a string parameter inside, to include the `tag` in the error message for easier debugging. Then here you can return e.g. `HyperlaneCosmosError::ReorgPeriodValueNotSupported(tag).into()`

rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -219,20 +220,25 @@ where
...
+        let block = match reorg_period {
+            EthereumReorgPeriod::Blocks(lag) => provider
+                .get_block_number()
+                .await
+                .map_err(ChainCommunicationError::from_other)?
+                .saturating_sub(lag.into())
+                .into(),
+            EthereumReorgPeriod::Tag(tag) => tag,
```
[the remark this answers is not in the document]
daniel-savu: in that case I think the problem is in `get_finalized_block_number` - it shouldn't make the network call if `tag` is `BlockId::Number`, but I still think we can reuse `get_finalized_block_number` here

rust/main/hyperlane-core/src/chain.rs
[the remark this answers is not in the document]
daniel-savu: oke sounds good

rust/main/hyperlane-core/src/chain.rs
daniel-savu: you probably need to support conversions for each individual tag anyway (as happens now in `TryFrom<&ReorgPeriod> for EthereumReorgPeriod`), so you need to deal with all tags variants either way, it's just that the parsing happens further into execution. 

I'm ok with allowing arbitrary strings, but mostly because the `TryFrom`s live in chain-specific integration crates, rather than having an enum that uses types from all these chain SDKs in `hyperlane-core`/`hyperlane-base`

rust/main/chains/hyperlane-ethereum/src/tx.rs
daniel-savu: while I agree we could push more encapsulation in the chain-specific implementations, I think the specific change you're suggesting would be quite invasive and would need to be part of a bigger chain builder overhaul. To me it's really the chain builders that feel weird rather than these conversions specifically :P

rust/main/chains/hyperlane-cosmos/src/utils.rs
apastushenka: Good idea! Maybe it would be even better to add it to `ChainCommunicationError`, something like
```
#[error("Unsupported reorg period: {0:?}")]
UnsupportedReorgPeriod(ReorgPeriod),
```
and reuse it in `EthereumReorgPeriod` too?

rust/main/chains/hyperlane-ethereum/src/tx.rs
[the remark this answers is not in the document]
daniel-savu: sorry should've been more clear @apastushenka, I'd like you to add this optimization in `get_finalized_block_number` to avoid the extra network call while reusing its logic

daniel-savu:
@apastushenka we're making a new release today so I'll merge the PR and get the remaining [comment](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4629#discussion_r1811084234) fixed in a follow up PR.

Really appreciate you making this contribution, especially since our requirements for this weren't very well defined. Thank you!

daniel-savu (approved):

PR #4636, opened 2024-10-07
fix: recipient rate limited ISM

yorhodes:
### Description

Add recipient restriction to rate limited ISM. This prevents multiple recipients from sharing the same rate limit ISM and denial of service attacks.

### Backward compatibility

No

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/isms/warp-route/RateLimitedIsm.sol
+++ b/solidity/contracts/isms/warp-route/RateLimitedIsm.sol
@@ -42,7 +42,7 @@ contract RateLimitedIsm is
 
     /// @inheritdoc IInterchainSecurityModule
     function moduleType() external pure returns (uint8) {
-        return uint8(IInterchainSecurityModule.Types.UNUSED);
+        return uint8(IInterchainSecurityModule.Types.NULL);
     }
 
     /**
--- a/solidity/contracts/isms/warp-route/RateLimitedIsm.sol
+++ b/solidity/contracts/isms/warp-route/RateLimitedIsm.sol
@@ -16,6 +16,8 @@ contract RateLimitedIsm is
     using Message for bytes;
     using TokenMessage for bytes;
 
+    address public immutable recipient;
+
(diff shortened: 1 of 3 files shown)


solidity/test/isms/RateLimitedIsm.t.sol
```diff
@@ -62,6 +64,12 @@ contract RateLimitedIsmTest is Test {
         rateLimitedIsm.verify(bytes(""), encodedMessage);
     }
 
+    function test_verifyOnlyRecipient() external {
+        bytes memory _message = MessageUtils.build(ORIGIN);
```
nambrot, line 68: IIUC, this relies on the fact that message recipient checking is before amount checking right? I guess the expectRevert will surface future issues, but i dont love that the message isn't a "valid" message in the first place

nambrot (approved):

solidity/contracts/isms/warp-route/RateLimitedIsm.sol
```diff
@@ -42,7 +52,12 @@ contract RateLimitedIsm is
     function verify(
         bytes calldata,
         bytes calldata _message
-    ) external validateMessageOnce(_message) returns (bool) {
+    )
+        external
```
nambrot, line 56: This seems like a good time to the mermaid diagrams and how atomicity with process is enforced

nambrot (approved):

solidity/test/isms/RateLimitedIsm.t.sol
yorhodes: the ordering doesnt matter, a reverts a revert

solidity/contracts/isms/warp-route/RateLimitedIsm.sol
yorhodes: opened https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4637

Issue #4449, opened 2024-09-06
GCS: validator fails to understand folders


sapph1re:
When trying to run a validator specifying checkpointSyncer settings for GCS:
`--checkpointSyncer.type gcs   --checkpointSyncer.bucket an-s3-signatures-repository   --checkpointSyncer.folder atletaolympia`

You get an error, apparently because it tries to report metadata with bucket name as `bucket/folder`, but apparently the "/" character is not accepted there:
```
  2024-09-06T21:41:51.634228Z  INFO hyperlane_base::db::rocks: Opening existing db, path: /hyperlane_db
    at hyperlane-base/src/db/rocks/mod.rs:86
    in hyperlane_base::db::rocks::from_path with db_path: "/hyperlane_db"

Building provider with signer
Building provider without signer
Building provider with signer
Building provider without signer
  2024-09-06T21:41:57.367309Z  INFO validator::server::eigen_node: Serving the EigenNodeAPI routes...
    at agents/validator/src/server/eigen_node.rs:66

Building provider without signer
thread 'main' panicked at 'Failed to report agent metadata: error validating bucket name an-s3-signatures-repository/atletaolympia

Caused by:
    Character '/' @ 27 is not allowed

Location:
    hyperlane-base/src/types/gcs_storage.rs:181:9', agents/validator/src/validator.rs:178:14
stack backtrace:
   0:     0x58678357db81 - std::backtrace_rs::backtrace::libunwind::trace::h66dc1c6acf794faa
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x58678357db81 - std::backtrace_rs::backtrace::trace_unsynchronized::ha80d20099a67f790
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x58678357db81 - std::sys_common::backtrace::_print_fmt::h7b959d43f35f16d4
                               at /rustc/d5c2e9c342b358556da91d61ed4133f6f50fc0c3/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x58678357db81 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hdaa196410d9ee0b9
[...]

PR #4652, opened 2024-10-09
fix: GCS folder support

IvanPsurtcev:
### Description

This PR fixes the problem of this issue:
[https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4449](url)
The change was only in the gcs_storage.rs file. Slightly tweaked methods so that you can write files to a specified folder in the terminal.

When trying to run a validator specifying checkpointSyncer settings for GCS:
`--checkpointSyncer.type gcs   --checkpointSyncer.bucket an-s3-signatures-repository   --checkpointSyncer.folder atletaolympia`

```
Building provider without signer
thread 'main' panicked at 'Failed to report agent metadata: error validating bucket name an-s3-signatures-repository/atletaolympia

Caused by:
    Character '/' @ 27 is not allowed

Location:
    hyperlane-base/src/types/gcs_storage.rs:181:9', agents/validator/src/validator.rs:178:14
```

To solve this problem I added the `folder: Option<String>` field to the GcsStorageClient structure, I also added bucket and folder processing to the impl CheckpointSyncer, GcsStorageClientBuilder methods

### DIFF
--- a/rust/main/hyperlane-base/src/types/gcs_storage.rs
+++ b/rust/main/hyperlane-base/src/types/gcs_storage.rs
@@ -11,6 +11,7 @@ use ya_gcp::{
     },
     AuthFlow, ClientBuilder, ClientBuilderConfig,
 };
+use tracing::{info, error};
 
 const LATEST_INDEX_KEY: &str = "gcsLatestIndexKey";
 const METADATA_KEY: &str = "gcsMetadataKey";
@@ -80,6 +81,8 @@ pub struct GcsStorageClient {
     inner: StorageClient,
     // bucket name of this client's storage
     bucket: String,
+    // folder name of this client's storage
+    folder: Option<String>,
 }
 
 impl GcsStorageClientBuilder {
@@ -94,13 +97,12 @@ impl GcsStorageClientBuilder {
         let inner = ClientBuilder::new(ClientBuilderConfig::new().auth_flow(self.auth))
             .await?
             .build_storage_client();
-        let bucket = if let Some(folder) = folder {
-            format! {"{}/{}", bucket_name.into(), folder}
-        } else {
-            bucket_name.into()
-        };
 
-        Ok(GcsStorageClient { inner, bucket })
+        let bucket = bucket_name.into();
+        let folder = folder;
+
+        GcsStorageClient::validate_bucket_name(&bucket)?;
+        Ok(GcsStorageClient { inner, bucket, folder })
     }
 }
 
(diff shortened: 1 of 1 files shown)


sapph1re:
@tkporter @daniel-savu  👀

rust/main/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -109,6 +111,24 @@ impl GcsStorageClient {
     fn get_checkpoint_key(index: u32) -> String {
         format!("checkpoint_{index}_with_id.json")
     }
+
+    fn object_path(&self, object_name: &str) -> String {
+        if let Some(folder) = &self.folder {
+            format!("{}/{}", folder.trim_end_matches('/'), object_name)
```
daniel-savu, line 117: makes sense to `folder.trim_end_matches('/')`, but instead of doing this each time here should it live in `GcsStorageClientBuilder::build` instead? Also worth logging there if anything _is_ trimmed, to avoid any silent unexpected behavior

rust/main/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -94,13 +97,12 @@ impl GcsStorageClientBuilder {
...
             .build_storage_client();
-        let bucket = if let Some(folder) = folder {
-            format! {"{}/{}", bucket_name.into(), folder}
-        } else {
-            bucket_name.into()
-        };
 
-        Ok(GcsStorageClient { inner, bucket })
+        let bucket = bucket_name.into();
+        let folder = folder;
```
daniel-savu, line 102: this line can be removed
```suggestion
```

rust/main/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -117,11 +137,12 @@ impl GcsStorageClient {
     }
 }
 
-// required by `CheckpointSyncer`
+// Required by `CheckpointSyncer`
 impl fmt::Debug for GcsStorageClient {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("S3Storage")
+        f.debug_struct("GcsStorageClient")
```
daniel-savu, line 143: good catch

rust/main/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -195,28 +216,42 @@ impl CheckpointSyncer for GcsStorageClient {
...
+        match self.inner.insert_object(&self.bucket, &object_name, serialized_metadata.into_bytes()).await {
+            Ok(_) => {
+                info!("Successfully uploaded metadata to '{}'", object_name);
+                Ok(())
+            }
+            Err(e) => {
+                error!("Failed to upload metadata to '{}': {:?}", object_name, e);
+                Err(e.into())
+            }
+        }
```
daniel-savu, line 231: since there are 5 calls to `insert_object` in this file, it'd make sense to refactor this into a function that just logs `Successfully posted to GCS storage` / `Failed to [...]` without going into specifics, and then `#[instrument(skip(...)]`-ing the functions that call this so that they show up in the log `spans` (while skipping all function arguments to avoid polluting the logs). So basically in `write_latest_index`, `write_checkpoint`, `write_metadata`, `write_announcement`, `write_reorg_status`

daniel-savu (commented):
nice, so the TL;DR is that GCS has no notion of folders and you just prepend it to the object key?

sapph1re:
> nice, so the TL;DR is that GCS has no notion of folders and you just prepend it to the object key?

GCS does have folders, but the existing code for GCS doesn't work with them correctly. That's why @IvanPsurtcev has made this addition to support folders like they are for AWS.

This is actually quite essential, because running multiple validators on one bucket requires them using different folders.

rust/main/hyperlane-base/src/types/gcs_storage.rs
IvanPsurtcev: Fixed the above edits:
https://github.com/IvanPsurtcev/hyperlane-monorepo/commit/9eec9548afb406428016571fc383de593e42da8c

daniel-savu (approved):
thank you!

daniel-savu:
@IvanPsurtcev I think merging `main` will fix CI

IvanPsurtcev:
> @IvanPsurtcev I think merging `main` will fix CI

Done

PR #4654, opened 2024-10-09
chore: disable sealevel e2e when possible

daniel-savu:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -216,9 +216,9 @@ jobs:
       - name: Check for Rust file changes
         id: check-rust-changes
         run: |
-          if [[ -n "$(git diff HEAD ${{ github.event.pull_request.base.sha }} -- ./rust)" ]]; then
+          if [[ -n "$(git diff ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha || github.sha }} -- ./rust)" ]]; then
             echo "rust_changes=true" >> $GITHUB_OUTPUT
-            echo "$(git diff HEAD ${{ github.event.pull_request.base.sha }} -- ./rust)" 
+            echo "$(git diff ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha || github.sha }} -- ./rust)" 
           else
             echo "rust_changes=false" >> $GITHUB_OUTPUT
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -212,6 +212,15 @@ jobs:
         env:
           RUST_BACKTRACE: 'full'
 
+      - name: Check for Rust file changes
+        id: check-rust-changes
+        run: |
+          if [[ -n "$(git diff origin/${{ github.base_ref }} HEAD -- ./rust)" ]]; then
```
paulbalaji, line 218: this means even agent-config changes will lead to a full sealevel e2e, do we want that?

daniel-savu: it's hard not to have false positives but it's better than false negatives imo - cargo.toml and .lock files could change, and possibly other configs too

daniel-savu: since most agent config changes happen on new deploy batches, I think it's acceptable

paulbalaji: sgtm!

.github/workflows/test.yml
```diff
@@ -212,6 +212,15 @@ jobs:
         env:
           RUST_BACKTRACE: 'full'
 
+      - name: Check for Rust file changes
+        id: check-rust-changes
+        run: |
+          if [[ -n "$(git diff ${{ github.event.pull_request.base.sha }} HEAD -- ./rust)" ]]; then
```
paulbalaji, line 218:
```suggestion
          if [[ -n "$(git diff ${{ github.event.pull_request.head.sha || github.sha }} HEAD -- ./rust)" ]]; then
```

paulbalaji (approved):

.github/workflows/test.yml
```diff
@@ -212,6 +213,16 @@ jobs:
         env:
           RUST_BACKTRACE: 'full'
 
+      - name: Check for Rust file changes
+        id: check-rust-changes
+        run: |
+          if [[ -n "$(git diff ${{ github.event.pull_request.head.sha || github.sha }} ${{ github.event.pull_request.base.sha }} -- ./rust)" ]]; then
+            echo "rust_changes=true" >> $GITHUB_OUTPUT
+            echo "$(git diff ${{ github.event.pull_request.head.sha || github.sha }} ${{ github.event.pull_request.base.sha }} -- ./rust)"
```
paulbalaji, line 221:
```suggestion
          if [[ -n "$(git diff ${{ github.event.pull_request.head.sha || github.sha }} ${{ github.event.pull_request.base.sha }}~1 -- ./rust)" ]]; then
            echo "rust_changes=true" >> $GITHUB_OUTPUT
            echo "$(git diff ${{ github.event.pull_request.head.sha || github.sha }} ${{ github.event.pull_request.base.sha }}~1 -- ./rust)" 
```

PR #4659, opened 2024-10-10
feat: onchain checkpoint

### LINKED ISSUE #4586: Validator should be able to be configured submit checkpoints on-chain
Hyperlane agents have the checkpoint syncer abstraction to allow validators to publish their checkpoints and relayers to read them. Currently, the primary options for those are web2 cloud storage like S3 and GCS. This bounty suggests the addition of a web3 native one where checkpoints are just written on-chain.

The easiest way would be to write a simple smart contract that allows any account to publish a checkpoint. It would be useful to indicate the location of the checkpoint syncer similar to `s3://bucket-name/region` as `onchain://chainName/contractAddress`. Validators could then use the existing chain signer abstraction to submit these checkpoints on-chain.

NOOMA-42:
### Description

enable validator to submit checkpoints on-chain

### Drive-by changes

add onchain checkpoint interface

### Related issues

- Fixes #4586 

### Backward compatibility

Yes


solidity/contracts/libs/CheckpointLib.sol
```diff
@@ -12,6 +12,11 @@ struct Checkpoint {
     bytes32 messageId;
 }
 
+struct SignedCheckpoint {
+    Checkpoint checkpoint;
+    bytes signature; // 72 bytes
```
yorhodes, line 17: this should be 65 bytes

solidity/contracts/CheckpointStorage.sol
```diff
@@ -0,0 +1,106 @@
...
+    struct AgentMetadata {
+        string gitSha;
+    }
+
+    mapping(address validator => mapping(uint32 index => SignedCheckpoint signedCheckpoint))
+        public submittedCheckpoints;
+    mapping(address validator => SignedAnnouncement signedAnnouncement)
+        public validatorAnnouncements;
+    mapping(address validator => AgentMetadata agentMetadata)
+        public validatorMetadata;
```
yorhodes, line 30: this is redundant with the `ValidatorAnnounce` contract

solidity/contracts/CheckpointStorage.sol
```diff
@@ -0,0 +1,106 @@
...
+            CheckpointLib.isEmpty(
+                submittedCheckpoints[validator][
+                    signedCheckpoint.checkpoint.index
+                ].checkpoint
+            ),
+            "Checkpoint already submitted"
+        );
+        if (signedCheckpoint.checkpoint.index > validatorLatestIndex) {
+            validatorLatestIndex = signedCheckpoint.checkpoint.index;
+        }
```
yorhodes, line 61: the latest index cannot be global for all validators

NOOMA-42:
@yorhodes would love to have your review :)

PR #4660, opened 2024-10-10
feat: Store as much as possible from every range

ameten:
### Description

Currently, if Scraper cannot fetch and parse a block or a transaction, it will fail the whole chunk of logs. It means that a transaction which can be fetched and parsed, but happens to be in the same chunk as unparseable transaction, won't be inserted into database. 

This PR fixes it. Scraper will do its best to insert as many as possible blocks and transactions into database from the chunk.

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4632

### Backward compatibility

Yes

### Testing

Local run of E2E tests.
Local run of Scraper

### DIFF
--- a/rust/main/agents/scraper/src/chain_scraper/mod.rs
+++ b/rust/main/agents/scraper/src/chain_scraper/mod.rs
@@ -277,8 +277,7 @@ impl HyperlaneSqlDb {
 
         let ensured_blocks = blocks
             .into_iter()
-            .filter_map(|(hash, block_info)| block_info)
-            .filter(|info| info.id != -1);
+            .filter_map(|(hash, block_info)| block_info.filter(|b| b.id != -1));
 
         Ok(ensured_blocks)
     }
--- a/rust/main/agents/scraper/src/chain_scraper/mod.rs
+++ b/rust/main/agents/scraper/src/chain_scraper/mod.rs
@@ -173,7 +173,7 @@ impl HyperlaneSqlDb {
             let ids = self.db.get_txn_ids(hashes_to_insert.drain(..)).await?;
 
             for (hash, (txn_id, _block_id)) in chunk.iter_mut() {
-                ids.get(hash).map(|id| *txn_id = Some(*id));
+                *txn_id = ids.get(hash).copied();
             }
         }
 
@@ -269,13 +269,15 @@ impl HyperlaneSqlDb {
                 .collect::<HashMap<_, _>>();
 
             for (block_ref, _) in blocks_to_insert.drain(..) {
-                hashes.get(&block_ref.hash).map(|id| block_ref.id = *id);
+                if let Some(id) = hashes.get(&block_ref.hash) {
+                    block_ref.id = *id;
+                }
             }
         }
 
         let ensured_blocks = blocks
             .into_iter()
-            .filter_map(|(hash, block_info)| block_info.map(|info| info))
+            .filter_map(|(hash, block_info)| block_info)
             .filter(|info| info.id != -1);
 
         Ok(ensured_blocks)
--- a/rust/main/agents/scraper/src/chain_scraper/mod.rs
+++ b/rust/main/agents/scraper/src/chain_scraper/mod.rs
@@ -14,7 +14,7 @@ use hyperlane_core::{
     HyperlaneWatermarkedLogStore, Indexed, InterchainGasPayment, LogMeta, H256,
 };
 use itertools::Itertools;
-use tracing::trace;
+use tracing::{trace, warn};
 
 use crate::db::{
     BasicBlock, BlockCursor, ScraperDb, StorableDelivery, StorableMessage, StorablePayment,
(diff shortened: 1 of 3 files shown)


rust/main/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -249,28 +269,24 @@ impl HyperlaneSqlDb {
                 .collect::<HashMap<_, _>>();
 
             for (block_ref, _) in blocks_to_insert.drain(..) {
-                block_ref.id = hashes[&block_ref.hash];
+                hashes.get(&block_ref.hash).map(|id| block_ref.id = *id);
```
daniel-savu, line 272: similar here, the `.map` is a bit misleading to me - I thought we were going to return a mapped version of `hashes`. I'd prefer something like this for readability
```suggestion
                if let Some(id) = hashes.get(&block_ref.hash) {
                    block_ref.id = *id;
                }
```

rust/main/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -249,28 +269,24 @@ impl HyperlaneSqlDb {
...
-            let block = block.as_ref().unwrap();
-            assert_eq!(hash, &block.hash);
-            assert!(block.id > 0);
-        }
-
-        Ok(blocks
+        let ensured_blocks = blocks
             .into_iter()
-            .map(|(hash, block_info)| block_info.unwrap()))
+            .filter_map(|(hash, block_info)| block_info.map(|info| info))
```
daniel-savu, line 278: could we write as 
```suggestion
            .filter_map(|(hash, block_info)| block_info)
```

daniel-savu: Would also be ok with merging the `filter` logic below with the `filter_map` logic here, to avoid iterating twice

rust/main/agents/scraper/src/db/message.rs
```diff
@@ -178,20 +179,20 @@ impl ScraperDb {
...
-                OnConflict::columns([delivered_message::Column::MsgId])
-                    .update_columns([
-                        delivered_message::Column::TimeCreated,
-                        delivered_message::Column::DestinationTxId,
-                    ])
-                    .to_owned(),
-            )
-            .exec(&self.0)
-            .await?;
+        if !models.is_empty() {
```
daniel-savu, line 182: you can also just `return Ok(0)` if `models.is_empty()` to avoid the new indentation

rust/main/agents/scraper/src/db/message.rs
```diff
@@ -272,28 +273,28 @@ impl ScraperDb {
...
-                    message::Column::Sender,
-                    message::Column::Recipient,
-                    message::Column::MsgBody,
-                    message::Column::OriginTxId,
-                ])
-                .to_owned(),
-            )
-            .exec(&self.0)
-            .await?;
+        if !models.is_empty() {
```
daniel-savu, line 276: here too, can `return Ok(0)` if `models.is_empty()` to avoid the new indentation

rust/main/agents/scraper/src/db/payment.rs
```diff
@@ -42,26 +42,26 @@ impl ScraperDb {
...
-                .update_columns([
-                    gas_payment::Column::TimeCreated,
-                    gas_payment::Column::Payment,
-                    gas_payment::Column::GasAmount,
-                ])
-                .to_owned(),
-            )
-            .exec(&self.0)
-            .await?;
+        if !models.is_empty() {
```
daniel-savu, line 45: same here

rust/main/agents/scraper/src/chain_scraper/mod.rs
```diff
@@ -153,43 +149,55 @@ impl HyperlaneSqlDb {
...
+            if txns_to_insert.is_empty() {
+                continue;
+            }
+
             self.db.store_txns(txns_to_insert.drain(..)).await?;
             let ids = self.db.get_txn_ids(hashes_to_insert.drain(..)).await?;
 
             for (hash, (txn_id, _block_id)) in chunk.iter_mut() {
-                let _ = txn_id.insert(ids[hash]);
+                ids.get(hash).map(|id| *txn_id = Some(*id));
```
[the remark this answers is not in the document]
ameten: It will work in this case (not the other ones).

Below version leaves `txn_id` untouched if `ids` does not contain `hash`
```
                ids.get(hash).map(|id| *txn_id = Some(*id));
```

This version updates all `txn_id`, sometime with the same value of `None'.
```
                *txn_id = ids.get(hash).copied();
```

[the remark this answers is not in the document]
ameten: Fixed.

rust/main/agents/scraper/src/chain_scraper/mod.rs
ameten: Fixed.

rust/main/agents/scraper/src/chain_scraper/mod.rs
ameten: I suppose that `filter_map` and `filter` will be executed lazily when elements will be collected into a collection. So, as far as I understand, there will be only one pass but applying all the operations to each element.

I think you are proposing this:
```
            .filter_map(|(hash, block_info)| block_info.filter(|b| b.id != -1));
```

rust/main/agents/scraper/src/db/message.rs
ameten: Fixed

rust/main/agents/scraper/src/db/message.rs
ameten: Fixed.

ameten:
@daniel-savu, great catches! Thank you!

rust/main/agents/scraper/src/chain_scraper/mod.rs
ameten: Implemented.

PR #4672, opened 2024-10-11
fix(cli): Update warp apply to apply changes in single command

### LINKED ISSUE #4671: fix: warp apply requires multiple calls for multiple updates
## Problem
Currently, `warp apply` only updates in place _or_ extends. It cannot do both in a single call. This is problematic because if there are both kinds of updates, then multiple `warp apply` are needed. 

As shown in Renzo's deployment config, a config is added to extend **and** a new multisig ISM is added for each existing chain. 

Update `warp apply` such that only 1 call is needed for all config updates.

ltyu:
### Description
This PR fixes a limitation in `warp apply` such that it can only extend _or_ update an existing warp route. This means that for configs with both changes require `warp apply` to be called multiple times. An example is when Renzo deploys to new chain, and it needs to update the existing ISMs.

### Related issues
- Fixes #4671 

### Backward compatibility
Yes

### Testing
Manual/Unit Tests

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -1,4 +1,5 @@
 import { confirm } from '@inquirer/prompts';
+import { groupBy } from 'lodash-es';
 import { stringify as yamlStringify } from 'yaml';
 
 import { buildArtifact as coreBuildArtifact } from '@hyperlane-xyz/core/buildArtifact.js';
@@ -53,6 +54,7 @@ import {
   Address,
   ProtocolType,
   assert,
+  isObjEmpty,
   objFilter,
   objKeys,
   objMap,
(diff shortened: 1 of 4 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -457,94 +448,119 @@ export async function runWarpRouteApply(
...
+    apiKeys,
+  );
 
-    await enrollRemoteRouters(params, mergedRouters);
+  const mergedRouters = mergeAllRouters(
+    multiProvider,
+    existingConfigs,
+    newDeployedContracts,
+    warpCoreConfigByChain,
+  );
```
yorhodes, line 514: can we just join the newDeployedContracts with the existing deployed and pass that to the update step?

yorhodes (approved):
mostly lgtm although it would intuitively make more sense to me if we did
```
contracts = contracts.extend()
contracts.update()
```

xeno097 (approved):

typescript/cli/src/deploy/warp.ts
ltyu: i can try to do that in a next PR. wanted to keep the diffs to a minimal for this

PR #4674, opened 2024-10-12
feat: allow update destination gas

### LINKED ISSUE #4529: feat: Add setDestinationGas to warp apply

ltyu:
### Description
Updates destination gas on all chain by getting the value from warpConfig.gas. It then synchronizes all destination gas amounts across all chains.

### Related issues
- Fixes #4529 

### Backward compatibility

Yes

### Testing

Manual/e2e


paulbalaji:
I assume this some additional context for this fix: https://discord.com/channels/935678348330434570/1294369638150311978/1294403359071014985

ltyu:
> I assume this is the additional context for this fix: https://discord.com/channels/935678348330434570/1294369638150311978/1294403359071014985

Yup thats it

typescript/cli/src/deploy/warp.ts
```diff
@@ -685,6 +700,38 @@ async function enrollRemoteRouters(
   return transactions;
 }
 
+/**
+ * Populates the destination gas amounts for each chain using warpConfig.gas OR querying other router's destinationGas
+ */
```
yorhodes, line 705: I dont think we want to mix these 2 modes
`read` should query from onchain
`update` shouldnt rely on existing onchain values imo

typescript/sdk/src/router/schemas.ts
```diff
@@ -32,6 +32,13 @@ export const RouterConfigSchema = MailboxClientConfigSchema.merge(
...
 
+const DestinationGasDomain = z.string();
+const DestinationGasAmount = z.string(); // This must be a string type to match Ether's type
+export const DestinationGasSchema = z.record(
+  DestinationGasDomain,
+  DestinationGasAmount,
+);
 export const GasRouterConfigSchema = RouterConfigSchema.extend({
   gas: z.number().optional(),
+  destinationGas: DestinationGasSchema.optional(),
```
yorhodes, line 43: okay now im quite confused

yorhodes: the intent was for the local domain's `gas` config to be used on the other origins as destination gas

yorhodes: is the idea that this `destinationGas` is always derived?

xeno097, line 43: What would be the difference between `gas` and `destinationGas`? Would it be just a "cached" way of storing the remote gas configuration on the current chain?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -232,4 +234,28 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+     */
+    const domains = await warpRoute.domains();
+
+    return Object.fromEntries(
+      await Promise.all(
+        domains.map(async (domain) => {
+          return [domain, (await warpRoute.destinationGas(domain)).toString()];
+        }),
+      ),
+    );
```
xeno097, line 259: I feel like this would be an excellent use case for the multicall discussion we had the other day

typescript/sdk/src/router/schemas.ts
ltyu: yes `destinationGas` is aways derived.

typescript/cli/src/deploy/warp.ts
ltyu: in this case, it has to. 

the new warp chain needs to know the other chain's destinationGas. the only way to do this is by reading existing onchain values

xeno097 (approved):

typescript/sdk/src/router/schemas.ts
ltyu: `gas` is used to describe a chain's own `gas` value, but is set on the other chains. destinationChain is a derived mapping that lists the _other_ `gas` values.
```
{
  arbitrum: {
    gas: 10
  },
  base: {
    gas: 20
 }
}
```
will result in:
```
{
  arbitrum: {
    destinationGas: {
      [base]: 20
    }
  },
    base: {
    destinationGas: {
      [arbitrum]: 10
    }
  }
}
```

PR #4678, opened 2024-10-14
fix: update objMerge implementation

paulbalaji:
### Description

Updating the `objMerge` implementation

A bug in the original implementation meant that the `update-agent-config` script did _not_ overwrite the `blocks.reorgPeriod` if there was a change, this new version does.

### Drive-by changes

- gracefully handle missing startBlock data when generating agent config
- fix objMerge calls that had the wrong order

### Related issues

definitely want to fix this bug before attempting to update our agent configs with changes in https://github.com/hyperlane-xyz/hyperlane-registry/pull/276

### Backward compatibility

should be, yes

### Testing

ci, manual testing when generating agent config files

### DIFF
--- a/typescript/utils/src/objects.test.ts
+++ b/typescript/utils/src/objects.test.ts
@@ -35,6 +35,20 @@ describe('Object utilities', () => {
     expect(merged).to.eql({ a: 2, b: { c: ['arr2'] } });
   });
 
+  it('objMerge with array and no merge', () => {
+    const obj1 = { a: 1, b: { c: ['arr1'] } };
+    const obj2 = { a: 2, b: { c: ['arr2'] } };
+    const merged = objMerge(obj1, obj2, 10, false);
+    expect(merged).to.eql({ a: 2, b: { c: ['arr2'] } });
+  });
+
+  it('objMerge overwrites nested values', () => {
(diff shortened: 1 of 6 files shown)


typescript/utils/src/objects.test.ts
```diff
@@ -35,6 +35,20 @@ describe('Object utilities', () => {
     expect(merged).to.eql({ a: 2, b: { c: ['arr2'] } });
   });
 
+  it('objMerge with array and no merge', () => {
+    const obj1 = { a: 1, b: { c: ['arr1'] } };
+    const obj2 = { a: 2, b: { c: ['arr2'] } };
+    const merged = objMerge(obj1, obj2, 10, false);
+    expect(merged).to.eql({ a: 2, b: { c: ['arr2'] } });
+  });
```
jmrossy, line 43: Am I missing something or is this identical to the test case above it?

jmrossy (approved):

typescript/utils/src/objects.test.ts
```diff
@@ -35,6 +35,20 @@ describe('Object utilities', () => {
...
+
```
paulbalaji, line 44:
```suggestion
```

PR #4682, opened 2024-10-15
feat: implement reorg audit remediations

paulbalaji:
### Description

Bump up .registryrc and update agent configs with updated chain metadata including:
- reorg periods
- technical stack

Updates originally made in https://github.com/hyperlane-xyz/hyperlane-registry/pull/276

Relies on `objMerge` to be fixed https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4678

### Drive-by changes

- remove extraneous overrides + bring exiting relayer env vars into config
	- so it's in the image build directly, and we minimise break-glass-in-case-of-emergency overrides to the stateful set
- ensure txOverrides are only included if defined in infra or registry	
	- ensure we delete txo verrides from the read-in config, before merging with the generated config

### Related issues

na

### Backward compatibility

yes

### Testing

na


typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -19,13 +19,10 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
     },
   },
   polygon: {
-    blocks: {
```
paulbalaji, line 22: removed, as it's the same as registry

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -19,13 +19,10 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
...
-      confirmations: 3,
-    },
     transactionOverrides: {
       // A very high max fee per gas is used as Polygon is susceptible
       // to large swings in gas prices.
-      maxFeePerGas: 550 * 10 ** 9, // 550 gwei
+      maxFeePerGas: 800 * 10 ** 9, // 800 gwei
```
paulbalaji, line 25: 550->800, matching the sts override

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -34,23 +31,6 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
...
   },
-  ethereum: {
-    blocks: {
-      confirmations: 3,
-    },
-    transactionOverrides: {
-      maxFeePerGas: 150 * 10 ** 9, // gwei
-      maxPriorityFeePerGas: 5 * 10 ** 9, // gwei
-    },
-  },
```
paulbalaji, line 45: removed overrides as discussed, also removing blocks.confirmations because registry has it set lower

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -34,23 +31,6 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
...
-  scroll: {
-    transactionOverrides: {
-      // Scroll doesn't use EIP 1559 and the gas price that's returned is sometimes
-      // too low for the transaction to be included in a reasonable amount of time -
-      // this often leads to transaction underpriced issues.
-      gasPrice: 2 * 10 ** 9, // 2 gwei
-    },
-  },
```
paulbalaji, line 53: removing scroll gasPrice override entirely, will monitor the situation 

if this is unfruitful, we can add back at 0.2 gwei (which is still about 2-5x more than usual averages)

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -75,7 +55,6 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
   // chiliz: {
   //   transactionOverrides: {
   //     maxFeePerGas: 100000 * 10 ** 9, // 100,000 gwei
-  //     maxPriorityFeePerGas: 1 * 10 ** 9, // 1 gwei
```
paulbalaji, line 78: this is set in registry already

typescript/infra/scripts/agents/update-agent-config.ts
```diff
@@ -138,10 +144,18 @@ export async function writeAgentConfig(
...
+    const currentAgentConfig: AgentConfig = readJSONAtPath(filepath);
+    // Remove transactionOverrides from each chain in the agent config
+    // To ensure all overrides are configured in infra code or the registry, and not in JSON
+    for (const chainConfig of Object.values(currentAgentConfig.chains)) {
+      delete chainConfig.transactionOverrides;
+    }
+    writeJsonAtPath(filepath, objMerge(currentAgentConfig, agentConfig));
+  } else {
+    writeJsonAtPath(filepath, agentConfig);
+  }
```
paulbalaji, line 158: currently tx overrides never get removed - only added or updated. this change ensures that we correctly derive these values from chains.ts or from the registry

ameten (approved):

PR #4685, opened 2024-10-15
feat: Upgrade Scraper

ameten:
### Description

Upgrade Scraper so that it records transactions with non-native tokens.

### Backward compatibility

Yes

### Testing

Manual test of Scraper
Test of ability to remove transaction and related data from database and re-index it again


paulbalaji:
replaced by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4686

PR #4694, opened 2024-10-16
feat: enroll new chains + validators on default ISMs

paulbalaji:
### Description

feat: enroll new chains + validators on default ISMs

- migrate new chains to ICAs
- migrate some existing SAFE owned chains to ICAs
- ensure defaultHook on inevm is made hot again
- add J to safe signers

### Drive-by changes

- IGP gas/token price updates
- add tx overrides when deploying ICA accounts
- swap out A for J on safe signers
- add support for updating owners when doing a safe multisend
- add support for checking and surfacing defaulthook/requiredhook owner violations
- improve logging + skip bytecode mismatches for now
- parallelise call inference in base app governor
- parallelise deployment of ICAs in get-owner-ica.ts
- reuse existing configured gasPrices for sealevel chains

### Related issues

na

### Backward compatibility

yes

### Testing

will be testing as part of check-deploy


typescript/infra/scripts/print-gas-prices.ts
```diff
@@ -48,6 +56,7 @@ async function main() {
 async function getGasPrice(
   mpp: MultiProtocolProvider,
   chain: string,
+  gasPrice?: GasPriceConfig,
```
tkporter, line 59: nit: just to avoid conflicts with line 65, maybe `oldGasPrice` or `existingGasPrice`?

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
```diff
@@ -82,6 +83,10 @@ export abstract class HyperlaneAppChecker<
   }
 
   addViolation(violation: CheckerViolation): void {
+    if (violation.type === ViolationType.BytecodeMismatch) {
+      rootLogger.warn({ violation }, `Found bytecode mismatch. Ignoring...`);
```
tkporter, line 87: should we do something more here?

tkporter (approved):
🤝

typescript/sdk/src/deploy/HyperlaneAppChecker.ts
paulbalaji: i think we'll have to fix it properly with something like this  https://discord.com/channels/935678348330434570/1296158577765126145/1296165727996350618

PR #4695, opened 2024-10-16
chore: Update warp Ids for Renzo

ltyu:
### Description
Updates the warp Ids for latest renzo deployment to sei and taiko used by the checker


tkporter (approved):
ty

PR #4704, opened 2024-10-17
feat: script to print and execute pending safe txs

paulbalaji:
### Description

feat: script to print and optionally execute pending safe txs

### Drive-by changes

na

### Related issues

add optional chain selection array to `withChains` and `withChainsRequired`

### Backward compatibility

yes

### Testing

Example without execution:
image

Example with execution:
image

### DIFF
--- a/typescript/infra/scripts/safes/get-pending-txs.ts
+++ b/typescript/infra/scripts/safes/get-pending-txs.ts
@@ -1,0 +1,178 @@
+import { confirm } from '@inquirer/prompts';
+import chalk from 'chalk';
+import yargs from 'yargs';
+
+import { MultiProvider } from '@hyperlane-xyz/sdk';
+
+import { Contexts } from '../../config/contexts.js';
+import { safes } from '../../config/environments/mainnet3/owners.js';
+import { Role } from '../../src/roles.js';
+import { executeTx, getSafeAndService } from '../../src/utils/safe.js';
+import { withChains } from '../agent-utils.js';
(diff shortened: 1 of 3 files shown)


typescript/infra/scripts/safes/get-pending-txs.ts
```diff
@@ -0,0 +1,192 @@
...
+    chains.map(async (chain) => {
+      if (!safes[chain]) {
+        console.error(chalk.red.bold(`No safe found for ${chain}`));
+        return;
+      }
+
+      if (chain === 'endurance') {
+        console.info(
+          chalk.gray.italic(
+            `Skipping chain ${chain} as it does not have a functional safe API`,
```
tkporter, line 47: 🙃

typescript/infra/src/utils/safe.ts
```diff
@@ -41,6 +41,40 @@ export function createSafeTransactionData(call: CallData): MetaTransactionData {
...
+    safeAddress,
+  );
+  let safeTransaction;
+  try {
+    safeTransaction = await safeService.getTransaction(safeTxHash);
+    if (!safeTransaction) {
+      throw new Error(`Failed to fetch transaction details for ${safeTxHash}`);
+    }
+  } catch (error) {
+    console.error(chalk.red(`Error fetching transaction: ${error}`));
```
tkporter, line 62: curious if you think the catching here / on line 69 should be left to the caller? feels useful for the caller to know if it was successful or not

tkporter (approved):
very nice

PR #4705, opened 2024-10-17
feat: add lumia warp route to infra

yorhodes:
### Description

- Add lumia warp route to infra config


typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBscLumiaLUMIAWarpConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+    bsc,
+    lumia,
+  };
+
+  const merged = objMap(configMap, (chain, config) => ({
+    ...routerConfig[chain],
+    ...config,
+    owner,
+    ownerOverrides: {
+      proxyAdmin: owner,
```
Mo-Hussain, line 39: nit: this does not need to be explicitly defined and the check will fall back to using owner if it is not defined. I'm not too opinionated on whether you think it's better to be explicit from readability perspective

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBscLumiaLUMIAWarpConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Lumia Team
+const owner = '0x8bBA07Ddc72455b55530C17e6f6223EF6E156863';
+
+export const getEthereumBscLUMIAWarpConfig = async (
+  routerConfig: ChainMap<RouterConfig>,
+): Promise<ChainMap<TokenRouterConfig>> => {
+  const ethereum = {
+    type: TokenType.collateral,
+    token: '0xD9343a049D5DBd89CD19DC6BcA8c48fB3a0a42a7',
```
Mo-Hussain, line 17: I would define the token address here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d8a9ad7abf6f66557d916dad90068d985be52460/typescript/infra/src/config/warp.ts#L5

Mo-Hussain (approved):

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBscLumiaLUMIAWarpConfig.ts
yorhodes: I dont expect this to be reused

PR #4720, opened 2024-10-21
fix: deriving ISM is very slow 

ltyu:
### Description
This PR adds concurrency back to the modules and use retryAsync.

### Drive-by changes

Add `maxAliasCount` param to yaml reader to prevent `ReferenceError: Excessive alias count indicates a resource exhaustion attack`

### Backward compatibility
Yes

### Testing
Manual


typescript/cli/src/utils/files.ts
```diff
@@ -15,6 +15,7 @@ import { objMerge } from '@hyperlane-xyz/utils';
 import { log } from '../logger.js';
 
 export const MAX_READ_LINE_OUTPUT = 250;
+export const MAX_ALIAS_YAML = 100_000; // Used for yaml maxAliasCount. Ref: https://eemeli.org/yaml/#tojs-options
```
xeno097, line 18: just curiosity but `100_000` has been chosen arbitrarily right?

xeno097 (approved):

xeno097 (approved):

paulbalaji (approved):
🔥

typescript/cli/src/utils/files.ts
ltyu: Yup!

ltyu: Some number other than -1 (infinity) is needed to prevent infinite laughs attack

PR #4721, opened 2024-10-21
fix: use hyp_message as tracing field instead of message

tkporter:
### Description

https://docs.rs/tracing/latest/tracing/#shorthand-macros
> unlike other fields, `message`'s shorthand initialization is just the string itself.

When using tracing::info etc, passing in an explicit `message` field takes precedence over the message string that's supplied as the second positional param! So we end up in situations where we see `message: HyperlaneMessage { id: ....` instead of the log message we want! e.g. https://cloudlogging.app.goo.gl/Nr8Q5W5KviD6sT8W9

Open to other field names, like `msg` or something, but this felt explicit


daniel-savu (approved):
wah so this was it, nice find! Everywhere we log `%message` we use `msg=` but nbd

paulbalaji (approved):

tkporter:
Updating to just do hyp_message everywhere

PR #4724, opened 2024-10-22
Update client.rs

LyDe17TC:
I've modified the get_block_height function so that it converts the height variable directly to u32 after retrieving it. If this conversion fails, it returns an explicit error indicating that the height of the block exceeds the capacity of u32.


daniel-savu:
I'm happy with the current behaviour, if the block number exceeds u32 capacity agents should crash as this would stop indexing from working. Up to @tkporter

LyDe17TC:
Ok no probleme, good luck

PR #4725, opened 2024-10-22
feat: zksync support

mshojaei-txfusion:
### Description

This PR introduces a series of changes aimed at enhancing zkSync support within the codebase. Key updates include the addition of the zksolc compiler for zkSync, integration of contract artifacts, CLI automation for core deployment, and compatibility adjustments in tests for the zkSync environment. It also includes improvements in contract verification on zkSync explorer and handling gas limits for zkSync deployments.

### Drive-by changes

- Added support for the zksolc compiler.
- Decoupled Hardhat configuration for EVM from zkSync.
- Enabled the export of zkSync-related contract artifacts.
- Verified compatibility of different deployment components based on the chain’s technical stack.
- Integrated zkSync provider and signer.
- Added support for contract verification on the zkSync explorer.
- Blacklisted static contracts from deployment on zkSync, as they are not supported.

### Related issues

No related issue

### Backward compatibility

Yes

### Testing
Manual testing and some automated tests were performed, including end-to-end tests on zkSync for warp read and apply functionalities.


ltyu (commented):
first pass. mostly looks good

typescript/cli/src/deploy/utils.ts
```diff
@@ -186,3 +191,29 @@ function transformChainMetadataForDisplay(chainMetadata: ChainMetadata) {
...
+  if (
+    shouldSkipStaticDeployment(chainTechnicalStack) &&
+    typeof config.defaultIsm !== 'string' &&
+    config.defaultIsm.type === IsmType.AGGREGATION
+  ) {
+    logRed('⛔ Static contract deployment not available on ZKSync!');
+    return false;
+  }
+  return true;
+}
```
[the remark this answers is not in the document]
ljankovic-txfusion: Elegant solution! 💯

typescript/cli/src/utils/keys.ts
```diff
@@ -77,14 +78,14 @@ async function addressToImpersonatedSigner(
...
  * @returns a signer for the private key
  */
-function privateKeyToSigner(key: string): ethers.Wallet {
+function privateKeyToSigner(key: string): Wallet {
   if (!key) throw new Error('No private key provided');
 
   const formattedKey = key.trim().toLowerCase();
   if (ethers.utils.isHexString(ensure0x(formattedKey)))
-    return new ethers.Wallet(ensure0x(formattedKey));
+    return new Wallet(ensure0x(formattedKey));
```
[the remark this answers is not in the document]
ljankovic-txfusion: It has to be backwards compatible because its used in bridging transactions from L2 to L1.
+ So far, we have had no issues using zksync wallet to deploy/send transactions on EVM.

typescript/cli/src/tests/warp-apply-zksync.e2e-test.ts
```diff
@@ -0,0 +1,337 @@
...
+const EXAMPLES_PATH = './examples';
+const CORE_CONFIG_PATH = `${EXAMPLES_PATH}/core-config-zksync.yaml`;
+const WARP_CONFIG_PATH_EXAMPLE = `${EXAMPLES_PATH}/warp-route-deployment-zksync.yaml`;
+
+const TEMP_PATH = '/tmp'; //temp gets removed at the end of all-test.sh
+const WARP_CONFIG_PATH_2 = `${TEMP_PATH}/zksync/warp-route-deployment-zksync.yaml`;
+const WARP_CORE_CONFIG_PATH_2 = `${ZK_REGISTRY_PATH}/deployments/warp_routes/ETH/${CHAIN_NAME_ZK_2}-config.yaml`;
+
+const TEST_TIMEOUT = 180_000; // Long timeout since these tests can take a while
+describe('WarpApply zkSync e2e tests', async function () {
```
paulbalaji, line 42: think you'll have to skip the warp apply + warp read e2e tests for now, because:
1. the test zksync chain metadata does not match that of the anvil networks we spin up
2. anvil networks don't really test zksync functionality anyway

```suggestion
describe.skip('WarpApply zkSync e2e tests', async function () {
```

so i'd say we keep the tests in, but skip here + in `warp-read-zksync.e2e-test.ts`, until we have the zksync e2e tests use https://docs.zksync.io/build/test-and-debug/in-memory-node

typescript/cli/test-configs/zksync/chains/zksync1/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+# yaml-language-server: $schema=../schema.json
+chainId: 260
+displayName: Zksync1
```
paulbalaji, line 3: consistency with anvil metadata
```suggestion
displayName: zksync1
```

typescript/cli/test-configs/zksync/chains/zksync2/metadata.yaml
```diff
@@ -0,0 +1,13 @@
+# yaml-language-server: $schema=../schema.json
+chainId: 270
+displayName: Zksync2
```
paulbalaji, line 3: consistency with anvil metadata
```suggestion
displayName: zksync2
```

typescript/cli/src/deploy/utils.ts
```diff
@@ -186,3 +190,26 @@ function transformChainMetadataForDisplay(chainMetadata: ChainMetadata) {
...
 }
+
+/**
+ * Checks if the given chain technical stack is compatible with the core configuration.
+ *
+ * @param {ChainTechnicalStack | undefined} params.chainTechnicalStack - The technical stack of the chain.
+ * @param {CoreConfig} params.config - The core configuration to check.
+ * @returns {boolean} True if the configuration is compatible, false otherwise.
+ */
+export function isIsmCompatible({
```
paulbalaji, line 201: This function only checks the top-level config, but I think it should also make sure that there is no aggregation or static isms anywhere nested the config

solidity/test/isms/MultisigIsm.t.sol
paulbalaji, line 1: will also need to fix other test files as per

[`b9d0c22` (#4761)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4761/commits/b9d0c2223a4f86c833757d735702653eded6814e)

typescript/sdk/src/deploy/verify/ZKSyncContractVerifier.ts
```diff
@@ -0,0 +1,157 @@
...
+      'Sending request to explorer...',
+    );
+
+    let response: Response;
+
+    response = await fetch(url.toString(), {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(options),
+    });
```
paulbalaji, line 129: linter
```suggestion
    const response = await fetch(url.toString(), {
      method: 'POST',
      headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify(options),
    });
```

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -135,6 +139,8 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
     switch (ismType) {
       case IsmType.MESSAGE_ID_MULTISIG:
       case IsmType.MERKLE_ROOT_MULTISIG:
+      case IsmType.STORAGE_MESSAGE_ID_MULTISIG:
+      case IsmType.STORAGE_MERKLE_ROOT_MULTISIG:
```
paulbalaji, line 143: this case is defined further down already
```suggestion
```

typescript/sdk/src/deploy/verify/ZKSyncContractVerifier.ts
```diff
@@ -0,0 +1,333 @@
...
+   * @returns The response from the explorer API
+   */
+  private async submitForm(
+    chain: ChainName,
+    verificationLogger: Logger,
+    options?: Record<string, any>,
+  ): Promise<any> {
+    const { apiUrl, family } = this.multiProvider.getExplorerApi(chain);
+
+    const url = new URL(apiUrl);
```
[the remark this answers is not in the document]
mshojaei-txfusion: For bouckscout they said, not yet supporting the verification yet on zksync, but for zksync etherscan, we added support, you just need to put `family:etherscan`, it would be handled by normal contract verifier.

typescript/cli/src/deploy/utils.ts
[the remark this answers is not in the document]
mshojaei-txfusion: Good point, but suggested code is not based the expected logic, so we refactored a bit to be able follow suggested pattern.

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -417,35 +440,53 @@ export abstract class HyperlaneDeployer<
...
-    const verificationInput = getContractVerificationInput({
-      name: contractName,
-      contract,
-      bytecode: factory.bytecode,
-      expectedimplementation: implementationAddress,
-    });
+    let verificationInput: ContractVerificationInput;
+    if (isZKSyncChain) {
+      if (!artifact) {
+        // TODO: ARTIFACT NOT FOUND ERROR
```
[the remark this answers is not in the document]
mshojaei-txfusion: handled, thanks for mentioning.

typescript/sdk/src/deploy/verify/ZKSyncContractVerifier.ts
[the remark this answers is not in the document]
paulbalaji: great thanks, can you leave this as a comment in the code as well please? 🙏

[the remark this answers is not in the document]
mshojaei-txfusion: Sure

typescript/cli/src/deploy/core.ts
```diff
@@ -81,6 +82,21 @@ export async function runCoreDeploy(params: DeployParams) {
 
   const userAddress = await signer.getAddress();
 
+  const { technicalStack: chainTechnicalStack } =
+    context.multiProvider.getChainMetadata(chain);
+
+  const ismType =
+    typeof config.defaultIsm === 'string'
+      ? (config.defaultIsm as IsmType)
```
ljankovic-txfusion, line 90: @paulbalaji In case config.defaultIsm is of type string, what could the potential value of that type be that would fall outside IsmType enum?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -308,34 +324,58 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
+    } else {
+      const contractFactory = factory.connect(signer);
+      const deployTx = contractFactory.getDeployTransaction(...params);
+      const estimatedGas = await signer.estimateGas(deployTx);
+      contract = await contractFactory.deploy(...params, {
+        gasLimit: addBufferToGasLimit(estimatedGas), // 10% buffer
+        ...overrides,
+      });
 
+      this.logger.trace(
```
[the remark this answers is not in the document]
ljankovic-txfusion: Could you please clarify?

typescript/sdk/src/utils/zksync.ts
```diff
@@ -0,0 +1,35 @@
...
+      const lowerCaseContractName = contractName.toLowerCase();
+      const lowerCaseName = name.toLowerCase();
+
+      // Check if the contract name matches
+      if (lowerCaseContractName === lowerCaseName) {
+        return true;
+      }
+
+      // Check if the qualified name matches
+      const qualifiedName = `${sourceName}:${contractName}`;
```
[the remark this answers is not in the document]
ljankovic-txfusion: If feel like if someone hardcodes just the contractName, there is a room for error in character casing, and if the fully qualified name is provided, the odds are its coming from artifact or some sort of internal process, and not user.

typescript/cli/src/tests/warp-apply-zksync.e2e-test.ts
ljankovic-txfusion: For testing zksync e2e I had to use both in-memory node and a dockerized version of zksync, since the in-memory node doesn't allow for custom chainId to be set [default is chainId: 260 on all in-memory node instances], if the same node process was used it would lead to chainId/domainId clash error

typescript/cli/src/deploy/utils.ts
ljankovic-txfusion: We added a check in HyperlaneIsmFactory for tech stack and Ism type

https://github.com/txfusion/hyperlane-monorepo/blob/3755332f6451b731618eac05e8200eb0ebbd9a5b/typescript/sdk/src/ism/HyperlaneIsmFactory.ts#L154

solidity/test/isms/MultisigIsm.t.sol
ljankovic-txfusion: btw. in `mailboxes.ts` imports for Validator and LegacyMultisigIsm are red. Seems like they have been removed from "@hyperlane-xyz/utils" and "@hyperlane-xyz/core" respectively.

typescript/cli/src/tests/warp-apply-zksync.e2e-test.ts
ljankovic-txfusion: Yeah, that makes complete sense

typescript/sdk/src/providers/MultiProvider.ts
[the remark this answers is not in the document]
paulbalaji: i think @ltyu meant move the this.logger.trace call from line 372 til after line 378

see how on line 360 we log "contract deployed" after the handleTx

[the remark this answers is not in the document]
ljankovic-txfusion: Fixed!

typescript/sdk/src/deploy/EvmModuleDeployer.ts
```diff
@@ -285,19 +320,21 @@ export class EvmModuleDeployer<Factories extends HyperlaneFactories> {
...
 
-      // add gas buffer
+      // add 10% buffer
       const hash = await factory['deploy(address[],uint8)'](
         sortedValues,
         threshold,
         {
-          gasLimit: addBufferToGasLimit(estimatedGas),
           ...overrides,
+          gasLimit: addBufferToGasLimit(estimatedGas),
```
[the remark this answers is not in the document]
ljankovic-txfusion: makes sense, fixed!

typescript/sdk/src/providers/MultiProvider.ts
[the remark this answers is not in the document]
paulbalaji: would also update the text from `Deploying contract` to `Contract deployed at`

typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -22,6 +22,7 @@ export enum ExplorerFamily {
   Etherscan = 'etherscan',
   Blockscout = 'blockscout',
   Routescan = 'routescan',
+  zksync = 'zksync',
```
paulbalaji, line 25: nit: matching the tech stack enum
```suggestion
  ZKSync = 'zksync',
```

paulbalaji (commented):
99% good to me, just the dist/evm -> core-utils/evm + explorer family enum change and i'm happy!

paulbalaji (approved):
lgtm - will wait on ltyu's review before merging

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -116,6 +123,15 @@ export abstract class HyperlaneDeployer<
     return this.options.contractVerifier?.verifyContract(chain, input, logger);
   }
 
+  async verifyContractForZKSync(
+    chain: ChainName,
+    input: ContractVerificationInput,
+    logger = this.logger,
+  ): Promise<void> {
+    const verifier = new ZKSyncContractVerifier(this.multiProvider);
+    return verifier?.verifyContract(chain, input, logger);
```
yorhodes, line 132: why would this be undefined?

typescript/sdk/src/deploy/protocolDeploymentConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ChainTechnicalStack } from '../metadata/chainMetadataTypes.js';
+
+/**
+ * @notice An array of chain technical stacks that are not supported for static deployment.
+ */
+export const skipStaticDeployment: ChainTechnicalStack[] = [
+  ChainTechnicalStack.ZKSync,
+];
+
+export const isIsmStatic: Record<IsmType, boolean> = {
```
yorhodes, line 11: we already have a "mutable ISM" thing somewhere that is kind of the inverse of this

typescript/sdk/src/deploy/proxyFactoryUtils.ts
yorhodes, line 1: I find this a bit hacky
is this assuming some implicit behavior of the consumer when the address is 0?

typescript/sdk/src/deploy/verify/utils.ts
```diff
@@ -64,6 +68,45 @@ export function getContractVerificationInput({
...
+  );
+}
+
+export function encodeArguments(abi: any, constructorArgs: any[]): string {
+  const contractInterface = new Interface(abi);
+  let deployArgumentsEncoded;
+  try {
+    deployArgumentsEncoded = contractInterface
+      .encodeDeploy(constructorArgs)
+      .replace('0x', '');
```
yorhodes, line 102: I wonder if this is related to the other thing I was running into

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -927,9 +928,10 @@ export class EvmHookModule extends HyperlaneModule<
       // deploy fallback hook
       const fallbackHook = await this.deploy({ config: config.fallback });
       // deploy routing hook with fallback
-      routingHook = await this.deployer.deployContract(
+      routingHook = await this.deployer.deployContractWithName(
```
yorhodes, line 931: what was this for?

typescript/sdk/src/hook/types.ts
```diff
@@ -42,6 +42,21 @@ export enum HookType {
...
+  [HookType.MERKLE_TREE]: 'merkleTreeHook',
+  [HookType.INTERCHAIN_GAS_PAYMASTER]: 'interchainGasPaymaster',
+  [HookType.AGGREGATION]: 'staticAggregationHook',
+  [HookType.PROTOCOL_FEE]: 'protocolFee',
+  [HookType.OP_STACK]: 'opStackHook',
+  [HookType.ROUTING]: 'domainRoutingHook',
+  [HookType.FALLBACK_ROUTING]: 'fallbackDomainRoutingHook',
+  [HookType.PAUSABLE]: 'pausableHook',
+  [HookType.ARB_L2_TO_L1]: 'arbL2ToL1Hook',
+};
```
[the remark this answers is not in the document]
yorhodes: sorry I dont understand
isnt the enum already string valued?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -127,7 +147,7 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     this.providers[chainName] = provider;
     const signer = this.signers[chainName];
     if (signer && signer.provider) {
-      this.setSigner(chainName, signer.connect(provider));
+      this.setSigner(chainName, (signer as EthersSigner).connect(provider));
```
yorhodes, line 150: dont really love this

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -155,7 +175,7 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     if (signer.provider) return signer;
     // Auto-connect the signer for convenience
     const provider = this.tryGetProvider(chainName);
-    return provider ? signer.connect(provider) : signer;
+    return provider ? signer.connect(provider as any) : signer;
```
yorhodes, line 178: does this mean that every consumer of the multi provider has to do this type narrowing of `getProvider`?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -413,7 +454,10 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     }
     const txReq = await this.prepareTx(chainNameOrId, tx);
     const signer = this.getSigner(chainNameOrId);
-    const response = await signer.sendTransaction(txReq);
+
+    const response = await signer.sendTransaction({
+      ...txReq,
+    });
```
yorhodes, line 460: why is this necessary?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -422,7 +466,10 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
    * Creates a MultiProvider using the given signer for all test networks
    */
   static createTestMultiProvider(
-    params: { signer?: Signer; provider?: Provider } = {},
+    params: {
+      signer?: Signer;
+      provider?: Provider;
+    } = {},
```
yorhodes, line 472: what changed here?

yorhodes (commented):
unfortunately dont think we can merge this to main yet due to the solidity changes

package.json
```diff
@@ -12,7 +12,8 @@
     "husky": "^8.0.0",
     "lint-staged": "^12.4.3",
     "prettier": "^2.8.8",
-    "tsx": "^4.7.1"
+    "tsx": "^4.7.1",
+    "zksync-ethers": "^5"
```
jmrossy, line 16: only tooling shared across the monorepo should be in this root package.json. Move it to the individual pacakges and specify a more specific version number

typescript/cli/examples/warp-route-deployment-zksync.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Typically used with the 'hyperlane deploy warp' command
+#
+# Token Types:
+#   native
+#   collateral
+#   synthetic
+#
+# see comprehensive [list](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/config.ts#L8)
+---
+zksync1:
```
jmrossy, line 11: This doesn't look like a valid deployment config. You would need at least 2 chains. Regardless, if you're not doing anything unique in this example, there's no need to add a new one

typescript/cli/examples/submit/strategy/json-rpc-chain-strategy.yaml
```diff
@@ -4,3 +4,9 @@ anvil2:
 anvil3:
   submitter:
     type: jsonRpc
+zksync1:
```
jmrossy, line 7: Are these local dev chains?

typescript/cli/src/context/context.ts
```diff
@@ -66,12 +67,11 @@ export async function getContext({
 }: ContextSettings): Promise<CommandContext> {
   const registry = getRegistry(registryUri, registryOverrideUri, !disableProxy);
 
-  let signer: ethers.Wallet | undefined = undefined;
+  let signer: Signer | Wallet | undefined = undefined;
```
jmrossy, line 70: does the zksync ethers wallet not implement the ethers Signer interface? If so, there's no need for the union type here.

typescript/cli/src/utils/keys.ts
```diff
@@ -1,5 +1,6 @@
 import { input } from '@inquirer/prompts';
 import { ethers, providers } from 'ethers';
+import { Wallet } from 'zksync-ethers';
```
jmrossy, line 3: ditto as above, confused by why we need to use the zk-specific wallet for the entire CLI here

yarn.lock
jmrossy, line 1: There's much more churn here than I would have expected for adding just one new package. For example, why have some react lib versions changed? Please revert the changes, make the package.json fix I mentioned above, and then ensure you're using yarn version 4.0.2 before running `yarn install`

typescript/sdk/src/zksync/ZKSyncDeployer.ts
```diff
@@ -0,0 +1,198 @@
+import assert from 'assert';
```
jmrossy, line 1: wrong assert. I thought we had a lint rule to catch these but maybe it stopped working

typescript/sdk/src/utils/zksync.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  ZkSyncArtifact,
+  loadAllZkArtifacts,
+} from '@hyperlane-xyz/core/zksync-artifacts';
```
jmrossy, line 4: Please don't use import paths into the package like this. If what you need isn't exported from the package's index.ts, modify the package to have it exported

jmrossy (dismissed):
Overall nice work but needs some fixes

typescript/sdk/src/zksync/ZKSyncDeployer.ts
ljankovic-txfusion: Will fix. Its also being used here btw. [EvmIsmModule.hardhat-test.ts](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/db5875cc22747362cd3e0a275b4897f1c3b7a800/typescript/sdk/src/ism/EvmIsmModule.hardhat-test.ts#L2).

typescript/cli/examples/submit/strategy/json-rpc-chain-strategy.yaml
ljankovic-txfusion: Yes, both zksync1 (in-memory node) & zksync2 (docker)

package.json
ljankovic-txfusion: We pretty much use it across the monorepo (solidity/sdk/cli), will specify version number

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: Seems like formatting, we do use your .prettierrc from the monorepo btw

typescript/sdk/src/hook/types.ts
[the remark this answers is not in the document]
ljankovic-txfusion: It is string valued, but its not the name of the contract, for instance
`HookType.FALLBACK_ROUTING = 'fallbackRoutingHook';`
but the correct name of the contract is `'fallbackDomainRoutingHook' `therefore, when we deploy this hook to zksync, we need to get the correct contract name to get its artifact using: `HookTypeToContractNameMap[HookType.FALLBACK_ROUTING]`

typescript/sdk/src/deploy/verify/utils.ts
ljankovic-txfusion: Nahh, I doubt it, its used in `getContractVerificationInputForZKSync`. IIRC you had an issue with `quoteDispatch`

typescript/cli/src/utils/keys.ts
ljankovic-txfusion: Agree, its confusing and far cry from ideal solution. In short, its backwards compatible with ethers wallet. We had discussions about limitations of the current Signer related features and its compatibility with MultiProvider, which is especially important for the upcoming Starknet SDK support.

Currently in `getContext` the value of `signer` is passed to `getMultiProvider` which in turn sets shared signer across all chains in the registry. At the time of getting singer, we don't know the protocol or chain technical stack that we can use to set the correct type of Signer.

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: Leftovers from when we had to hardcode gasLimit for zksync local node, will remove

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: So the quickest solution to this is handle zksync provider separately, but it doesn't really scale: 

```
    if (signer instanceof ZKSyncWallet) {
      return signer.connect(provider as ZKSyncProvider);
    }
    return (signer as EthersSigner).connect(provider);
```

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: Yeah, here we force signer as EthersSigner because ZKSyncWallet requires ZKSyncProvider that has a couple of additional functions [mostly for bridging & reading some state from L1] we currently have no use for here, but the rest of funcs are pretty much same

solidity/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "rootDir": "./types"
+    "rootDir": "./core-utils"
```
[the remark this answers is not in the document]
ljankovic-txfusion: `types` folder is created as a part of hardhat typechain command, to encompass zksync specific changes, we have decited to put typechain factories under `typechain` folder instead of `types`, and zksync specific code and artifacts under `zksync` folder, so in future, when we add some other chain/protocol, we can simply create another folder in core-utils/

This is the structure used on this branch, there are no changes to the way evm Typechain or build artifacts are imported elsewhere (sdk or cli).

```
Solidity
├── core-utils (or dist)
│   ├── typechain
│   │   ├── contracts
│   │   ├── @openzeppelin
│   │   ├── ...
│   │   └── index.js
│   ├── zksync
│   │   ├── artifacts
│   │   │   ├── ContractName.json
│   │   │   └── ...
│   │   ├── index.ts (zksync specific func)
│   │   └── buildArtifact.js
│   ├── buildArtifact.js (evm)
│   └── index.js (exports)
```

yarn.lock
ljankovic-txfusion, line 1: @jmrossy I took yarn.lock from the main branch and ran the `yarn` command which created this current diff

solidity/tsconfig.json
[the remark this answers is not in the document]
yorhodes: the annoying part is switching between git branches that have artifacts in a different path
it breaks the build and its pretty hard to clean up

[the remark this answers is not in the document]
yorhodes: had to manually rm some stuff

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -245,6 +247,13 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
   }
 
   async estimateHandle(message: DispatchedMessage): Promise<string> {
+    // This estimation is not possible on zksync as it is overriding transaction.from
+    // transaction.from must be a signer on zksync
```
yorhodes, line 251: @ltyu @xeno097

typescript/sdk/src/deploy/verify/BaseContractVerifier.ts
```diff
@@ -0,0 +1,208 @@
...
+      verificationLogger.debug(`No explorer family set, skipping`);
+      return false;
+    }
+
+    if (explorerApi.family === ExplorerFamily.Other) {
+      verificationLogger.debug(`Unsupported explorer family, skipping`);
+      return false;
+    }
+
+    if (input.address === ethers.constants.AddressZero) return false;
```
jmrossy, line 98: see `isZeroishAddress` in the utils package

typescript/sdk/src/deploy/verify/utils.ts
```diff
@@ -1,11 +1,15 @@
...
 
+import { ZKSyncArtifact } from '@hyperlane-xyz/core';
 import { Address, eqAddress } from '@hyperlane-xyz/utils';
 
+import { ExplorerFamily } from '../../metadata/chainMetadataTypes.js';
 import { ChainMap, ChainName } from '../../types.js';
 
 import { ContractVerificationInput } from './types.js';
 
+const { Interface } = await import('@ethersproject/abi');
```
jmrossy, line 11: Why are you doing a dynamic import of this? Instead, import it directly from the `ethers` package like other imports above

typescript/sdk/src/providers/MultiProtocolProvider.ts
```diff
@@ -99,12 +100,12 @@ export class MultiProtocolProvider<
     const providers = objMap(
       this.providers,
       (_, typeToProviders) => typeToProviders[ProviderType.EthersV5]?.provider,
-    ) as ChainMap<EthersV5Provider['provider'] | undefined>;
+    ) as ChainMap<ZKSyncProvider['provider'] | undefined>;
```
jmrossy, line 103: Similar to the concern I expressed above around the Wallet class, I disagree that ZKSyncProviders should replace Ethers V5 providers as the default provider type here

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -155,7 +175,13 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
     if (signer.provider) return signer;
     // Auto-connect the signer for convenience
     const provider = this.tryGetProvider(chainName);
-    return provider ? signer.connect(provider) : signer;
+    if (!provider) return signer;
+
+    // Handle ZKSync provider separately
+    if (signer instanceof ZKSyncWallet) {
```
jmrossy, line 181: Using `instanceof` this way has pitfalls and is generally a code smell. Is there a tangible property on the signer you can check instead?

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -308,34 +334,55 @@ export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
...
+  async handleDeploy<F extends ZKSyncContractFactory | ContractFactory>(
     chainNameOrId: ChainNameOrId,
     factory: F,
     params: Parameters<F['deploy']>,
+    artifact?: ZKSyncArtifact,
   ): Promise<Awaited<ReturnType<F['deploy']>>> {
-    // setup contract factory
+    const metadata = this.tryGetChainMetadata(chainNameOrId);
+    if (!metadata) {
+      throw new Error('Chain metadata not found!');
```
jmrossy, line 345: Just use `this.getChainMetadata(chainNameOrId);` and it'll throw if if doesn't find it

typescript/sdk/src/providers/providerBuilders.ts
```diff
@@ -109,13 +111,30 @@ export function defaultCosmJsWasmProviderBuilder(
...
+
 // Kept for backwards compatibility
 export function defaultProviderBuilder(
   rpcUrls: RpcUrl[],
   _network: number | string,
 ): providers.Provider {
   return defaultEthersV5ProviderBuilder(rpcUrls, _network).provider;
 }
+// Kept for backwards compatibility
+export function defaultZKProviderBuilder(
```
jmrossy, line 132: Why would we need this for backwards compat when it doesn't currently exist? You can remove this

jmrossy (commented):
Lots of nice work here but I still have some concerns, please see comments :)

typescript/sdk/src/providers/MultiProvider.ts
```diff
@@ -43,7 +55,7 @@ export interface MultiProviderOptions {
 export class MultiProvider<MetaExt = {}> extends ChainMetadataManager<MetaExt> {
   readonly providers: ChainMap<Provider>;
   readonly providerBuilder: ProviderBuilderFn<Provider>;
-  signers: ChainMap<Signer>;
+  signers: ChainMap<ZKSyncWallet | Signer>;
```
jmrossy, line 58: So ZKSyncWallet doesn't implement the Signer interface? I'd expect that it would since an Ethers.js Wallet does implement it

typescript/sdk/src/providers/providerBuilders.ts
ljankovic-txfusion: we use it in MultiProvider, will remove this comment

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: For types we can remove both ZKSyncWallet and ZKSyncProvider... but we need to be mindful that MultiProvider is also used for deploying contracts (& gas estimation...), and for that we use ZKSyncDeployer class, which strictly requires that the type of signer is ZKSyncWallet [extends Wallet base]. Meaning that type casting to ZKSyncWallet will still be required down in the code. We need to keep this in mind.

typescript/sdk/src/providers/MultiProtocolProvider.ts
ljankovic-txfusion: You are right, this should not be here. Resolved

typescript/sdk/src/deploy/verify/utils.ts
ljankovic-txfusion: Resolved

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: Resolved

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: Your comment on removing ZKSyncWallet form ChainMap helped me resolve this. No longer need to check for instance of

typescript/sdk/src/deploy/verify/BaseContractVerifier.ts
ljankovic-txfusion: Done!

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: @jmrossy 's comment helped me resolve this. Now typecasting is no longer needed for `signer.connect()`

sachanayush47:
I was trying this feature and I caught this error while deploying warp contracts: `Error: Technical stack zksync is not compatible with staticAggregationIsm`

sachanayush47:
Hi, I was trying native to native token token bridging on zk chains and I encountered this error - `Error validating transfer {"amount":"Insufficient collateral on destination"}` . I had sufficient funds on both of the chains.

typescript/cli/src/utils/keys.ts
ljankovic-txfusion: @jmrossy We created a PR to address this issue https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4914/files

typescript/sdk/src/providers/MultiProvider.ts
ljankovic-txfusion: This is resolved, only Signer is used now

solidity/test/lib/mailboxes.ts
```diff
@@ -18,8 +18,8 @@ import {
   LegacyMultisigIsm,
   TestMailbox,
   TestMerkleTreeHook,
-} from '../../types';
-import { DispatchEvent } from '../../types/contracts/Mailbox';
+} from '../../core-utils/evm/types';
```
paulbalaji, line 21: a bunch of tests still reference `core-utils/evm` but it looks like it's been moved instead to `core-utils/typechain`

typescript/cli/scripts/run-e2e-test.sh
```diff
@@ -3,10 +3,16 @@
 function cleanup() {
   set +e
   pkill -f anvil
+  rm -rf /tmp/anvil2
+  rm -rf /tmp/anvil3
+  rm -rf /tmp/zksync1
+  rm -rf /tmp/zksync2
```
paulbalaji, line 9: not necessary because of the rm -rf tmp below

typescript/sdk/src/hook/EvmHookModule.ts
paulbalaji: by default deployContract just uses the enum value, but here the enum value doesn't match the actual contract name (hence the type->name mapping)

its also non trivial to just change the enum because it affects other places, i hit this once before with the ism factory

.changeset/ten-spiders-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
+---
+
+Added ZKSync signer support using zksync-ethers package
```
paulbalaji, line 5: need to update the changeset and change to major

typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -30,7 +31,7 @@ export enum ChainTechnicalStack {
   OpStack = 'opstack',
   PolygonCDK = 'polygoncdk',
   PolkadotSubstrate = 'polkadotsubstrate',
-  ZkSync = 'zksync',
+  ZKSync = 'zksync',
```
paulbalaji, line 34: nit: can we we keep this as `ZkSync` and update all the `ZKSync` -> `ZkSync` for naming elsewhere as well for consistency?

imo it should generally be `ZkXYZ` rather than `ZKXYZ`, @ltyu wdyt?

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -21,10 +21,16 @@ import type {
...
+import {
+  Contract as ZKSyncBaseContract,
+  Provider as ZKSyncBaseProvider,
+  types as zkSyncTypes,
+} from 'zksync-ethers';
 
 import { Annotated, ProtocolType } from '@hyperlane-xyz/utils';
 
 export enum ProviderType {
+  ZKSync = 'zksync',
```
paulbalaji, line 33: ditto depending on the outcome of the ZKSync -> ZkSync comment earlier

typescript/sdk/src/metadata/chainMetadataTypes.ts
ljankovic-txfusion: We sure can, but zksync did the rebranding and is actually the owner of "ZK" trademark, so at the end the value should be ZKsync. But at the end it really doesn't matter, whatever you guys prefer

.changeset/ten-spiders-trade.md
ljankovic-txfusion: This is related to this PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4914/files

paulbalaji: gotcha, would need a new changeset for this PR then

ljankovic-txfusion: Since there are a lot of changes on `sdk` and `core` I propose we add new 'major' changeset for these too? @paulbalaji

solidity/test/lib/mailboxes.ts
ljankovic-txfusion: Fixed!

.changeset/ten-spiders-trade.md
paulbalaji: yup I'd agree

solidity/test/merkle.test.ts
```diff
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import { utils } from 'ethers';
 
 import merkleTestCases from '../../vectors/merkle.json' assert { type: 'json' };
-import { TestMerkle, TestMerkle__factory } from '../types';
+import { TestMerkle, TestMerkle__factory } from '../core-utils/evm/types';
```
paulbalaji, line 5: pls update _all_ the references e.g. this

core-utils/evm -> core-utils/typechain

solidity/test/message.test.ts
```diff
@@ -8,7 +8,11 @@ import {
 } from '@hyperlane-xyz/utils';
 
 import testCases from '../../vectors/message.json' assert { type: 'json' };
-import { Mailbox__factory, TestMessage, TestMessage__factory } from '../types';
+import {
+  Mailbox__factory,
+  TestMessage,
+  TestMessage__factory,
+} from '../core-utils/evm/types';
```
paulbalaji, line 15: core-utils/evm -> core-utils/typechain

solidity/test/mockMailbox.test.ts
```diff
@@ -3,7 +3,10 @@ import { utils } from 'ethers';
 
 import { addressToBytes32 } from '@hyperlane-xyz/utils';
 
-import { MockMailbox__factory, TestRecipient__factory } from '../types';
+import {
+  MockMailbox__factory,
+  TestRecipient__factory,
+} from '../core-utils/evm/types';
```
paulbalaji, line 9: core-utils/evm -> core-utils/typechain

solidity/test/testrecipient.test.ts
```diff
@@ -3,7 +3,7 @@ import { utils } from 'ethers';
 
 import { addressToBytes32 } from '@hyperlane-xyz/utils';
 
-import { TestRecipient, TestRecipient__factory } from '../types';
+import { TestRecipient, TestRecipient__factory } from '../core-utils/evm/types';
```
paulbalaji, line 6: core-utils/evm -> core-utils/typechain

solidity/test/merkle.test.ts
ljankovic-txfusion: All should be resolved

typescript/sdk/src/deploy/protocolDeploymentConfig.ts
ltyu: This is also a bit hard to read imo because in certain cases, its used with `isStaticDeploymentSupported()`, which uses an object (of bool values) and an array to determine a bool. 

Suggest we use arrays similar to what Yorke described for `MUTABLE_ISM_TYPE`. Maybe we can reuse that
```
const staticIsm = [IsmType.AGGREGATION, IsmType.MERKLE_ROOT_MULTISIG]
```

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -136,6 +151,14 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
       }`,
     );
 
+    const { technicalStack } = this.multiProvider.getChainMetadata(destination);
+
+    // For static ISM types it checks whether the technical stack supports static contract deployment
+    assert(
+      !isIsmStatic[ismType] || isStaticDeploymentSupported(technicalStack),
```
ltyu, line 158: is there a difference between this cond and `isIsmCompatible()`?

typescript/sdk/src/metadata/chainMetadataTypes.ts
ltyu: no opinions here. if it doesn't impact anything, then i'm okay with the change.

ltyu (approved):

typescript/sdk/src/providers/ProviderType.ts
mshojaei-txfusion: fixed

typescript/sdk/src/metadata/chainMetadataTypes.ts
mshojaei-txfusion: fixed

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
mshojaei-txfusion: Nice catch,
fixed here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4725/commits/14533751f5bf23da07c861bfa52005206b0dd60b

.changeset/wicked-cherries-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
```
jmrossy, line 2:
```suggestion
'@hyperlane-xyz/sdk': major
```

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -21,10 +21,16 @@ import type {
...
+import {
+  Contract as ZKSyncBaseContract,
+  Provider as ZKSyncBaseProvider,
+  types as zkSyncTypes,
+} from 'zksync-ethers';
 
 import { Annotated, ProtocolType } from '@hyperlane-xyz/utils';
 
 export enum ProviderType {
+  ZkSync = 'zksync',
```
jmrossy, line 33: This is not a big deal but would you agree that when adding an entry to an enum with several entries already, it makes sense to add it at the end? Especially when it starts with 'Z'. I wonder why so many developer tend to shove new things at the top as though they're more important than the existing parts.

typescript/sdk/src/index.ts
```diff
@@ -639,3 +645,5 @@ export {
   WarpTypedTransaction,
 } from './warp/types.js';
 export { WarpCore, WarpCoreOptions } from './warp/WarpCore.js';
+
```
jmrossy, line 648: When you leave empty spaces here, it interferes with the VS-code and/or prettier auto formatting

xeno097 (approved):

typescript/sdk/src/providers/ProviderType.ts
ljankovic-txfusion: Hmmm interesting thought In my experience, new code is more visible when you put it at the top, especially when its code you write yourself, has nothing to do with importance. Will address this

solidity/zk-hardhat.config.cts
```diff
@@ -0,0 +1,46 @@
...
+    compilerSource: 'binary',
+    enableEraVMExtensions: true,
+  },
+  defaultNetwork: 'zkSyncNetwork',
+  networks: {
+    zkSyncNetwork: {
+      url: 'http://127.0.0.1:8011',
+      ethNetwork: '',
+      zksync: true,
+    },
```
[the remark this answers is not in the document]
ljankovic-txfusion: Would keep it there, removing the zksync network reference has been know to cause issues for some reason

.changeset/wicked-cherries-pump.md
ljankovic-txfusion: Updated

typescript/sdk/src/providers/ProviderType.ts
ljankovic-txfusion: Resolved. All zksync references on this file are not below the existing ones

jmrossy: I agree that new code is more visible when it's at the top, but why do you feel that new code needs to me more visible than the existing code?

ljankovic-txfusion: Just being devil's advocate. I agree, It doesn't belong at the top, have no good explanation why this particular file is ordered the way it was, other files are fine. @mshojaei-txfusion lets make a note of this

PR #4727, opened 2024-10-22
feat: Make HyperlaneProvider to request block by height, not by hash

ameten:
### Description

Make HyperlaneProvider to request block by height, not by hash.

Since it is hard to find a block by hash for Solana, we switch to use block height. We have to do it for all type of chains since we need to change signature of chain-agnostic method

### Drive-by changes

* Small method rename

### Related issues

### Backward compatibility

Yes

### Testing

Manual run of Scraper for Ethereum and Neutron chains.

### DIFF
--- a/rust/main/hyperlane-core/src/traits/provider.rs
+++ b/rust/main/hyperlane-core/src/traits/provider.rs
@@ -42,10 +42,10 @@ pub enum HyperlaneProviderError {
     #[error("Could not find transaction from provider with hash {0:?}")]
     CouldNotFindTransactionByHash(H256),
     /// Could not find a block by height
-    #[error("Could not find block from provider with hash {0:?}")]
+    #[error("Could not find block from provider with height {0:?}")]
     CouldNotFindBlockByHeight(u64),
     /// The requested block does not have its hash
-    #[error("Block does not contain its hash {0:?}")]
+    #[error("Block with height {0:?} does not contain its hash")]
     BlockWithoutHash(u64),
     /// Incorrect block is received
     #[error("Requested block with height {0:?}, received block with height {1:?}")]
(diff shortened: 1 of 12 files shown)


rust/main/hyperlane-core/src/traits/provider.rs
```diff
@@ -35,13 +35,19 @@ pub trait HyperlaneProvider: HyperlaneChain + Send + Sync + Debug {
...
     #[error("Provider did not return gas used")]
     NoGasUsed,
-    /// Could not find a transaction, block, or other object
-    #[error("Could not find object from provider with hash {0:?}")]
-    CouldNotFindObjectByHash(H256),
+    /// Could not find a transaction by hash
+    #[error("Could not find transaction from provider with hash {0:?}")]
+    CouldNotFindTransactionByHash(H256),
+    /// Could not find a block by height
+    #[error("Could not find block from provider with hash {0:?}")]
```
tkporter, line 45: hit
```suggestion
    #[error("Could not find block from provider with height {0:?}")]
```

rust/main/hyperlane-core/src/traits/provider.rs
```diff
@@ -35,13 +35,19 @@ pub trait HyperlaneProvider: HyperlaneChain + Send + Sync + Debug {
...
+    CouldNotFindBlockByHeight(u64),
+    /// The requested block does not have its hash
+    #[error("Block does not contain its hash {0:?}")]
```
tkporter, line 48:
```suggestion
    #[error("Block with height {0:?} does not contain its hash")]
```

tkporter (approved):
nice!

rust/main/hyperlane-core/src/traits/provider.rs
ameten: Thank you! Fixed.

rust/main/hyperlane-core/src/traits/provider.rs
ameten: Thank you! Fixed.

PR #4729, opened 2024-10-22
refactor: implement outbound and inbound warp route amount transforms

yorhodes:
### Description

- implements `_outbound` and `_inbound` internal amount transforms for use in scaling warp routes
- simplify `HypNativeScaled` implementation

### Backward compatibility

Yes

### Testing

Existing HypNative Scaled Unit Tests


solidity/contracts/token/libs/TokenRouter.sol
```diff
@@ -130,7 +132,19 @@ abstract contract TokenRouter is GasRouter {
             _hook
         );
 
-        emit SentTransferRemote(_destination, _recipient, _amountOrId);
+        emit SentTransferRemote(_destination, _recipient, outboundAmount);
+    }
+
+    function _outbound(
```
nambrot, line 138: Definitely would appreciate some documentation and comments here

nambrot: As well as indicating the desired use case (which i believe is only scaling, right?

yorhodes: [2c96dd2](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4729/commits/2c96dd2e2025c849c461326b499e6726e492ccad)

PR #4730, opened 2024-10-22
fix: revert concurrency in Modules

ltyu:
### Description
Reverts https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4720

Reasons:
- Many of the public RPCs don't allow concurrency reads. The ones that do don't tell us how much they allow so no point in defaulting to some assumed number (20)
- retryAsync doesn't do anything here as the SmartProvider knows what to do if a provider fails (retries with cooldown). SmartProvider has it's own retryAsync
- Concurrency is configurable in the metadata. so configure for each provider, as intended.
Screenshot 2024-10-22 at 4 11 02 PM


yorhodes:
> Reason: [discord.com/channels/935678348330434570/1297972601536385034/1298373790996107285](https://discord.com/channels/935678348330434570/1297972601536385034/1298373790996107285)

this link doesnt want to load for me for whatever reason
lets try to keep PR descriptions with markdown text as much as possible (to preserve git log etc)

yorhodes (approved):

PR #4734, opened 2024-10-23
chore: add sleep log in `fetch_logs_with_cursor`

daniel-savu:

### DIFF
--- a/rust/main/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/main/hyperlane-base/src/contract_sync/mod.rs
@@ -202,11 +202,13 @@ where
             },
             CursorAction::Sleep(duration) => duration,
         };
-        debug!(
-            cursor = ?cursor,
-            "Cursor can't make progress, sleeping",
-        );
-        sleep(sleep_duration).await
+        if sleep_duration > Duration::from_secs(0) {
+            debug!(
+                cursor = ?cursor,
+                "Cursor can't make progress, sleeping",
+            );
+            sleep(sleep_duration).await
+        }
     }
 
     async fn dedupe_and_store_logs(
--- a/rust/main/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/main/hyperlane-base/src/contract_sync/mod.rs
@@ -170,7 +170,7 @@ where
                     Ok(logs) => logs,
                     Err(err) => {
                         warn!(?err, ?range, "Error fetching logs in range");
-                        break SLEEP_DURATION;
+                        break Some(SLEEP_DURATION);
                     }
                 };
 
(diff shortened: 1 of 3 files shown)


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -202,6 +202,10 @@ where
             },
             CursorAction::Sleep(duration) => duration,
         };
+        debug!(
+            cursor = ?cursor,
+            "Cursor can't make progress, sleeping",
```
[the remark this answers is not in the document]
daniel-savu: very good point, changed so that we only call sleep if the duration is greater than zero

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -202,7 +202,13 @@ where
             },
             CursorAction::Sleep(duration) => duration,
         };
-        sleep(sleep_duration).await
+        if sleep_duration > Duration::from_secs(0) {
```
tkporter, line 205: wonder if we should make this an option?

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -202,7 +202,13 @@ where
...
+            debug!(
+                cursor = ?cursor,
+                "Cursor can't make progress, sleeping",
+            );
+            sleep(sleep_duration).await
```
tkporter, line 210: wanna flag 2 things that I'm curious to get your take on:
1. the `CursorAction::Sleep(duration)` case is valid non-error behavior that would result in this log, wanna make sure that you're comfortable with this log message given that
2. given our previous run-ins around starving tasks, I wonder if our `sleep(Duration::default).await` here effectively served as a yield_now? I assume this is safe bc of all the async operations that happen in this function anyways

tkporter (approved):

rust/main/hyperlane-base/src/contract_sync/mod.rs
daniel-savu: 1. I did consider this, which is why it's a `debug` and not a warn/error. Would a different phrasing make it seem less like an error to you?
2. since this function doesn't loop, we yield by ending the function (because this function is itself called with `.await`), so we don't need to call sleep for that

tkporter: sgtm on 1, for 2 the reason I ask is iirc task will only possibly give up control if one of the functions that's awaited returns a Poll::Pending, but if it immediately is ready then it keeps control. But agree that because of the RPC interactions that we expect to happen in this function we'll likely always hit a Poll:Pending

daniel-savu: I think with the db processor is a special situation because we make synchronous calls indeed. Since this code will always index using network calls it's safe to assume we won't yield immediately

PR #4743, opened 2024-10-24
fix: Update reorgPeriod to 15 for BSC Chain Configuration

Fazleiw:
### Description

Reorgperiod BSC changess

### Drive-by changes

error / bug related to invalid data type from string to numerical

### Related issues

-Fixes bug 

### Backward compatibility

yes
### Testing

manual npm test

### DIFF
--- a/rust/main/config/mainnet_config.json
+++ b/rust/main/config/mainnet_config.json
@@ -429,7 +429,7 @@
       "blocks": {
         "confirmations": 1,
         "estimateBlockTime": 3,
-        "reorgPeriod": "finalized"
+        "reorgPeriod": "15"
       },
       "chainId": 56,
       "deployer": {


paulbalaji:
this is not an issue, but a feature with our latest sdk. please use the latest versions of our cli and sdk if you're having problems

Fazleiw:
> this is not an issue, but a feature with our latest sdk. please use the latest versions of our cli and sdk if you're having problems

but lot of people still said there lot of error happening even tho itss already updated

paulbalaji:
> > this is not an issue, but a feature with our latest sdk. please use the latest versions of our cli and sdk if you're having problems
> 
> but lot of people still said there lot of error happening even tho itss already updated

thanks for pointing the problem to me on discord - issue was a problem with the latest cli release itself , patch incoming soon :)

PR #4747, opened 2024-10-24
chore: update Eclipse validator sets to include Stride, minor cleanup

tkporter:
### Description

- deletes all the mainnet2 artifacts that we don't need
- moves the sealevel helper scripts into more explicit directories, and ensures thresholds are appropriate
- updates the eclipse ISM


rust/sealevel/environments/mainnet3/multisig-ism-message-id/eclipsemainnet/hyperlane/multisig-config.json
```diff
@@ -1,12 +1,21 @@
...
+  },
   "ancient8": {
     "threshold": 2,
     "validators": [
       "0xbb5842ae0e05215b53df4787a29144efb7e67551",
       "0xa5a56e97fb46f0ac3a3d261e404acb998d9a6969",
       "0x95c7bf235837cb5a609fe6c95870410b9f68bcff"
     ],
-    "type": 3
+    "type": "messageIdMultisigIsm"
```
daniel-savu, line 18: thought having a script was breaking the scripts?

typescript/infra/scripts/sealevel-helpers/print-multisig-ism-config.ts
```diff
@@ -0,0 +1,45 @@
...
+  // https://explorer.eclipse.xyz/tx/3wcMvqZZjQon9o8nD49e3ci16AUJopZRLAfsAfs16ZrxgoNLoboNvrbV1hQHbnN3KXrWSqHmKnmM28mUvh5Un5Hd/inspect.
+  // At the time, the Solana threshold was 3. Taking the max tx size of 1232 and the tx's size 1121,
+  // we can find the number of additional signatures to be: floor((1232 - 1121)/65) = floor(1.707) = 1.
+  // So the total number of signatures is 3 + 1 = 4.
+
+  const MAX_THRESHOLD = 4;
+
+  for (const chain of Object.keys(config)) {
+    if (config[chain].threshold > MAX_THRESHOLD) {
+      config[chain].threshold = MAX_THRESHOLD;
```
daniel-savu, line 35: nit: I normally use this command as `print-multisig-ism-config.ts > file.txt`. If others do too, we could use stderr to log that we're not using the configured threshold

daniel-savu (approved):

typescript/infra/scripts/sealevel-helpers/print-multisig-ism-config.ts
tkporter: this is why I didn't add a log, but I didn't consider stderr, good call

PR #4756, opened 2024-10-25
fix: isObject utils fn should return only boolean value

tiendn:
### Description
When `isObject` pass with param `null/undefined` it returns `null/undefined`

We should strict return boolean value

### Drive-by changes
```diff
export function isObject(item: any): boolean {
- return item && typeof item === 'object' && !Array.isArray(item);
+ return !!item && typeof item === 'object' && !Array.isArray(item);
}
```

### Testing
Has updated with test `object.test.ts`


jmrossy (approved):
Nice improvement!

PR #4757, opened 2024-10-25
fix(contracts): access control to prevent DOS on `RateLimitHook`

aroralanuk:
### Description

- Added onlySender to `RateLimitHook` to only allow a specific sender like a warp route to send messages through the hook.

### Drive-by changes

None

### DIFF
--- a/solidity/contracts/hooks/warp-route/RateLimitedHook.sol
+++ b/solidity/contracts/hooks/warp-route/RateLimitedHook.sol
@@ -1,18 +1,47 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity >=0.8.0;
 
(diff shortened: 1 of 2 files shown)


solidity/contracts/hooks/warp-route/RateLimitedHook.sol
```diff
@@ -1,59 +1,96 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
yorhodes, line 1: why change this?

yorhodes (approved):
lgtm

solidity/contracts/hooks/warp-route/RateLimitedHook.sol
aroralanuk: Consistency with the rest of the codebase

PR #4753, opened 2024-10-25
fix: Add unit tests for math utility functions and fix median calculation

tiendn:
### Description
This pull request introduces the following changes to the math utility functions:
#### Fix Median Calculation:
Corrected the median function to handle even-length arrays properly by averaging the two middle numbers.

#### Add Unit Tests:
Implemented unit tests for the math utility functions using Chai to ensure their correctness and reliability.
The functions tested include:
- `median`: Tests for both odd and even-length arrays.
- `sum`: Verifies the sum of an array of numbers.
- `mean`: Checks the calculation of the mean.
- `stdDev`: Validates the standard deviation calculation, including cases with negative numbers.
- `randomInt`: Ensures the generated random integer falls within the specified range.

### Drive-by changes
- Code: Updated the median function in math.ts to correctly calculate the median for even-length arrays.
- Tests: Added a new test file math.test.ts with comprehensive test cases for each function.

### Related issues

Fixes [#[4754]](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4754)

### Backward compatibility
Yes

### Testing

- All tests have been executed and passed successfully, confirming the correctness of the functions.
- Special attention was given to edge cases, such as arrays with negative numbers and random integer generation.

PR #4758, opened 2024-10-25
test: Coverage more @hyperlane-xyz/utils test

tiendn:
### Description
Add more coverage `utils` package test

### Testing

More unittests function


tiendn:
Will be added more unitests later for prevent conflict code.

.changeset/curly-squids-pay.md
yorhodes, line 1: I think because this is just adding tests and dev dependencies no need to add a changeset

typescript/utils/package.json
```diff
@@ -17,6 +17,7 @@
     "chai": "4.5.0",
     "mocha": "^10.2.0",
     "prettier": "^2.8.8",
+    "sinon": "^19.0.2",
```
jmrossy, line 20: Please use the same version as in other packages (so 13.0.2 like in the SDK) or upgrade it for all packages across the whole monorepo. See the SDK's package.json for an example. You'll also want to add these:
```
    "@types/sinon": "^17.0.1",
    "@types/sinon-chai": "^3.2.12",
    ```

Comment (jmrossy) [reply typescript/utils/package.json:20]: And then a separate issue: you need to run `yarn` after changing a package.json so the yarn.lock file can be updated. That's why CI is failing

**Review by jmrossy — COMMENTED** [review]
This is great! Just need to fix the deps and like Yorke said, we don't need a changeset for this.

File: .changeset/curly-squids-pay.md
Comment (tiendn) [reply .changeset/curly-squids-pay.md:1]: Tks, i'll remove it

File: typescript/utils/package.json (hunk 1 above)
Comment (tiendn) [reply typescript/utils/package.json:20]: I don't know why, but in my MacOS, it cannot find patch for `typechain`
```
Error: typechain@patch:typechain@npm%3A8.3.2#~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch: 
 ENOENT: no such file or directory, open '~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch'
```

Comment (tiendn) [reply typescript/utils/package.json]: After use relative path for patch, I've updated `yarn.lock` successfully~

Comment (jmrossy) [reply typescript/utils/package.json]: @tiendn I think you're using an outdated version of yarn. That's also why the yarn.lock file has a huge amount of changes. It should have minimal changes. Please ensure you're using yarn 4.0.2 as is specified in the .yarnrc.yml file. If you're using any version of yarn 4, it _should_ just set it to 4.0.2 automatically when run in the monorepo

File: typescript/utils/src/checkpoints.test.ts [hunk 2]
```diff
@@ -0,0 +1,120 @@
+import { expect } from 'chai';
+
+import {
+  isCheckpoint,
+  isS3Checkpoint,
+  isS3CheckpointWithId,
+  isValidSignature,
+} from '../src/checkpoints.js';
+import { Checkpoint, S3Checkpoint, S3CheckpointWithId } from '../src/types.js';
```
jmrossy, line 9: These import paths don't need the `../src/` segments.

typescript/utils/package.json
tiendn: tks, I've fixed it

jmrossy:
Screenshot 2024-10-25 at 3 37 49 PM
@tiendn

tiendn:
> Screenshot 2024-10-25 at 3 37 49 PM @tiendn

hm, it needs https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4753 merge first.

jmrossy (approved):
Nice! Great to have this extra coverage

PR #4759, opened 2024-10-25
feat: Use H512 as transaction hash in HyperlaneProvider and Scraper

ameten:
### Description

Use H512 as transaction hash in HyperlaneProvider and Scraper.

### Related issues

- Contributes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4272

### Backward compatibility

Yes

### Testing

Manual run of Scraper for Ethereum and Neutron


daniel-savu (approved):
since so far the db has been storing these types as `H256`, what are the implications of starting to `store_logs` as `H512`?

ameten:
There is no implication in this PR. Transaction hash is 32 bytes in Ethereum and Cosmos chains. We pass transaction hash as H512 (64 bytes). When we store the hash to database, we remove leading zeros and store it as 32-byte-long binary string into `bytea` datatype of Postgres.

Once we add support for Solana which has 64-byte long transaction hashes, we'll store transaction hashes as 64-byte-long bynary strings into the same `bytea` datatype.

PR #4760, opened 2024-10-25
fix: ensure recipients on Sealevel warp transfers are 32 bytes

tkporter:
### Description

Ran into an issue trying to send from Sealevel -> Cosmos where the recipient was 20 bytes. This hasn't been an issue so far because:
- no previous Sealevel -> Cosmos transfers have been done in the UI
- Sealevel -> EVM transfers are successful bc `addressToBytes` actually pads the 20 byte EVM addresses to 32 bytes
- Sealevel -> Sealevel transfers are fine because Sealevel addys are always 32 bytes

### DIFF
--- a/.changeset/olive-maps-applaud.md
+++ b/.changeset/olive-maps-applaud.md
@@ -1,0 +1,6 @@
+---
+'@hyperlane-xyz/utils': patch
+'@hyperlane-xyz/sdk': patch
+---
+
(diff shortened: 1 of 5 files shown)


paulbalaji (approved):
changeset pls

paulbalaji (approved):

daniel-savu (approved):

PR #4764, opened 2024-10-25
test: add more object utils test + remove undefined value when use with invertKeysAndValues fnc

tiendn:
### Description
Add filter `undefined/null` value when use with utils **invertKeysAndValues** function
Add more test at `objects.test.ts`

### Drive-by changes
```diff
return Object.fromEntries(
    Object.entries(data)
+      .filter(([_, value]) => value !== undefined && value !== null) // Filter out undefined and null values
      .map(([key, value]) => [value, key]),
);
```

### Related issues
None

### Backward compatibility
Currently, nowhere call this **invertKeysAndValues** func

### Testing
Yes, more tests.

### DIFF
--- a/typescript/utils/src/objects.test.ts
+++ b/typescript/utils/src/objects.test.ts
@@ -1,12 +1,24 @@
 import { expect } from 'chai';
 
 import {
+  arrayToObject,
   deepCopy,
   deepEquals,
+  deepFind,
   diffObjMerge,
+  invertKeysAndValues,
+  isObjEmpty,
   isObject,
+  objFilter,
+  objKeys,
+  objLength,
+  objMap,
+  objMapEntries,
   objMerge,
   objOmit,
+  pick,
+  promiseObjAll,
+  stringifyObject,
 } from './objects.js';
 
 describe('Object utilities', () => {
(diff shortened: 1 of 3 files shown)


typescript/utils/src/objects.test.ts
```diff
@@ -83,6 +95,116 @@ describe('Object utilities', () => {
...
+    expect(result).to.eql({});
+  });
+
+  it('invertKeysAndValues should invert the keys and values', () => {
+    const obj = { a: '1', b: '2' };
+    const result = invertKeysAndValues(obj);
+    expect(result).to.eql({ '1': 'a', '2': 'b' });
+  });
+
+  it('invertKeysAndValues should return an empty object if the object is empty', () => {
```
jmrossy, line 176: This is optional but consider nesting another `describe` block under the outer `describe` block to organize the tests by their function. E.g. a `describe('invertKeysAndValues', () =>` block

jmrossy (approved):
One optional comment but otherwise looks good!

typescript/utils/src/objects.test.ts
tiendn: Thank you! Please review again

jmrossy (approved):

PR #4765, opened 2024-10-25
Add relaying message context to config derivation

yorhodes:

### DIFF
--- a/typescript/sdk/src/core/HyperlaneRelayer.ts
+++ b/typescript/sdk/src/core/HyperlaneRelayer.ts
@@ -75,12 +75,18 @@ export class HyperlaneRelayer {
   async getHookConfig(
     chain: ChainName,
     hook: Address,
+    messageContext?: DispatchedMessage,
   ): Promise<DerivedHookConfig> {
     let config: DerivedHookConfig | undefined;
     if (this.cache?.hook[chain]?.[hook]) {
       config = this.cache.hook[chain][hook] as DerivedHookConfig | undefined;
     } else {
-      const evmHookReader = new EvmHookReader(this.multiProvider, chain);
+      const evmHookReader = new EvmHookReader(
+        this.multiProvider,
+        chain,
+        undefined,
+        messageContext,
+      );
       config = await evmHookReader.deriveHookConfig(hook);
     }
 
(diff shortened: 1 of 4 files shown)


typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -66,6 +67,7 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
     protected readonly concurrency: number = multiProvider.tryGetRpcConcurrency(
       chain,
     ) ?? DEFAULT_CONTRACT_READ_CONCURRENCY,
+    protected readonly messageContext?: DispatchedMessage,
   ) {
```
xeno097, line 71: I think that it would more reusable to add the `messageContext` to the `deriveRoutingConfig` method signature instead of tying the class instance to a specific message. I don't have a use case in my mind right now but this way instead of having to create a new instance of the class we would just need to call `deriveRoutingConfig` with a different message context and be able to reuse the same instance

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -96,6 +97,7 @@ export class EvmHookReader extends HyperlaneReader implements HookReader {
     protected readonly concurrency: number = multiProvider.tryGetRpcConcurrency(
       chain,
     ) ?? DEFAULT_CONTRACT_READ_CONCURRENCY,
+    protected readonly messageContext?: DispatchedMessage,
```
xeno097, line 100: here too I would do the same as suggested in the `EvmIsmReader`

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -98,12 +104,18 @@ export class HyperlaneRelayer {
...
     let config: DerivedIsmConfig | undefined;
     if (this.cache?.ism[chain]?.[ism]) {
       config = this.cache.ism[chain][ism] as DerivedIsmConfig | undefined;
     } else {
-      const evmIsmReader = new EvmIsmReader(this.multiProvider, chain);
+      const evmIsmReader = new EvmIsmReader(
+        this.multiProvider,
+        chain,
+        undefined,
+        messageContext,
```
ltyu, line 117: why does it need to be initialized with this?

typescript/sdk/src/ism/EvmIsmReader.ts
yorhodes: the `messageContext` needs to be propagated across all of the recursive function calls so this was the "least invasive" way for me to make this change
would you prefer it to be optional on every function?

xeno097 (approved):

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4765#discussion_r1819179492

PR #4767, opened 2024-10-25
feat: core deploy apply admin proxy ownership fixes

### LINKED ISSUE #4728: feat(cli,sdk): core commands support changing proxy admin ownership
ref: https://www.notion.so/hyperlanexyz/Infra-CLI-migration-10c6d35200d68087bb86ce352c525b25#10e6d35200d680578cc5c5d00b7f1dba

xeno097:
### Description

This PR updates the `hyperlane core init`, `hyperlane core deploy` and `hyperlane core apply` commands to allow a user to change ownership of the mailbox ProxyAdmin contract by setting a value in the config. 

### Drive-by changes

- deduped `randomAddress` test util implementations across the `sdk`, 'infra' and `cli` package
- added `anvil1` to the `run-e2e-test.sh` script to test `hyperlane core` commands in isolation
- implemented the `proxyAdminOwnershipUpdateTxs` to deduplicate proxy admin ownership tx data generation

### Related issues

- Fixes #4728 

### Backward compatibility

- Yes

### Testing

- Manual
- e2e

### NOTE:

- Merge https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4726 first


typescript/sdk/src/core/EvmCoreReader.ts
```diff
@@ -77,4 +82,19 @@ export class EvmCoreReader implements CoreReader {
 
     return results as CoreConfig;
   }
+
+  private async _getProxyAdminConfig(
```
ltyu, line 86: not the biggest fan of `_` in front of private functions. i know we do it in solidity, but lets keep it without, unless you want to update the other private functions to keep consistency :)

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -288,38 +292,6 @@ export class EvmERC20WarpModule extends HyperlaneModule<
     );
   }
 
-  updateProxyAdminOwnershipTxs(
```
ltyu, line 291: nit: you can always keep this function and just call the static proxyAdminOwnershipUpdateTxs() to keep the same function signature of only having actualConfig and expectedConfig

typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -332,11 +344,27 @@ export class EvmCoreModule extends HyperlaneModule<
     const { merkleTreeHook, interchainGasPaymaster } =
       serializedContracts[chainName];
 
+    // Update the ProxyAdmin owner of the Mailbox if the config defines a different owner from the current signer
+    const currentProxyOwner = await proxyAdmin.owner();
+    if (
+      config?.proxyAdmin?.owner &&
+      config.proxyAdmin.owner !== currentProxyOwner
```
ltyu, line 351: can use eqAddress()

ltyu (approved):
e2e test failing in CI. it might be related to using anvil1, which is used by foundry-test.

typescript/sdk/src/deploy/proxy.ts
```diff
@@ -67,3 +72,36 @@ export async function isProxy(
...
+
+  // Return early because old config files did not have the
+  // proxyAdmin property
+  if (!expectedConfig.proxyAdmin?.address) {
+    return transactions;
+  }
+
+  const actualProxyAdmin = actualConfig.proxyAdmin!;
+  assert(
+    eqAddress(actualProxyAdmin.address!, expectedConfig.proxyAdmin.address),
```
ltyu, line 91: This prevents changing of an existing proxyAdmin to another proxyAdmin. i know this is functionally different than what is used here (change proxyadmin.owner to another owner). which are we expecting? cc @paulbalaji

typescript/sdk/src/core/EvmCoreReader.ts
xeno097: I might be willing to change all private functions to use `_` 🤣 but I'll change it to keep the project consistent. Old habits coming back ahah.

typescript/sdk/src/deploy/proxy.ts
```diff
@@ -67,3 +72,36 @@ export async function isProxy(
...
+  transactions.push(
+    // Internally the createTransferOwnershipTx method already checks if the
+    // two owner values are the same and produces an empty tx batch if they are
+    ...transferOwnershipTransactions(
+      chainId,
+      actualProxyAdmin.address!,
+      actualProxyAdmin,
+      expectedConfig.proxyAdmin,
+    ),
+  );
```
paulbalaji, line 104: do we also need to call `changeProxyAdmin`?

typescript/sdk/src/deploy/proxy.ts
paulbalaji: I feel like both could be useful - at the top level we detect if the proxyAdmin has changed. If it's the same address but we have a different config, then we hit this function

typescript/sdk/src/core/schemas.ts
```diff
@@ -3,12 +3,15 @@ import { z } from 'zod';
...
-import { OwnableSchema } from '../schemas.js';
+import { DeployedOwnableSchema, OwnableSchema } from '../schemas.js';
 
 export const CoreConfigSchema = OwnableSchema.extend({
   defaultIsm: IsmConfigSchema,
   defaultHook: HookConfigSchema,
   requiredHook: HookConfigSchema,
+  // This field is set as optional because the old core config
+  // did not have it and we want to maintain backward compatibility
+  proxyAdmin: DeployedOwnableSchema.optional(),
```
paulbalaji, line 14: define "old core config" - out of 140ish chains with a `mailbox` in registry, 134 have a `proxyAdmin`

typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -91,6 +96,11 @@ export class EvmCoreModule extends HyperlaneModule<
     transactions.push(
       ...(await this.createDefaultIsmUpdateTxs(actualConfig, expectedConfig)),
       ...this.createMailboxOwnerUpdateTxs(actualConfig, expectedConfig),
+      ...proxyAdminOwnershipUpdateTxs(
+        actualConfig,
+        expectedConfig,
+        this.domainId,
```
paulbalaji, line 102: another case of chainId/domainId conflation :D 

should use the chainId

typescript/cli/scripts/run-e2e-test.sh
```diff
@@ -3,16 +3,20 @@
...
   rm -f ./test-configs/anvil/chains/anvil2/addresses.yaml
   rm -f ./test-configs/anvil/chains/anvil3/addresses.yaml
   set -e
 }
 
 cleanup
 
-echo "Starting anvil2 and anvil3 chain for E2E tests"
+echo "Starting anvil1, anvil2 and anvil3 chain for E2E tests"
+# Anvil1 should be used only for core commands e2e testing to avoid interfierence with other e2e tests
```
paulbalaji: why are you unable to reuse anvil2/3?

also typo
```suggestion
# Anvil1 should be used only for core commands e2e testing to avoid interference with other e2e tests
```

typescript/sdk/src/core/EvmCoreModule.ts
paulbalaji: can you also have chainId be the first arg to be consistent with `transferOwnershipTransactions`

typescript/sdk/src/deploy/proxy.ts
xeno097: I agree happy to make this change!

typescript/sdk/src/core/schemas.ts
xeno097: For example, if someone uses an old config file where the `proxyAdmin` field was not defined because it was generated with an old version of the CLI, Example:
- Deploy core contracts with old cli
- Update the CLI to the new version that adds this field
- Try to read the core deployment config. If the fields is set as required this will fail

typescript/cli/src/tests/commands/core.ts
```diff
@@ -45,6 +45,7 @@ export async function hyperlaneCoreApply(
         --registry ${REGISTRY_PATH} \
         --config ${coreOutputPath} \
         --chain ${chain} \
+        --key ${ANVIL_KEY} \
```
paulbalaji, line 48: doh!

typescript/sdk/src/core/schemas.ts
paulbalaji: I see - wonder if it's ok to just make it a breaking change that we demand the `proxyAdmin` now and if an old version doesn't have it then that's on the user

wdyt @ltyu?

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -109,7 +109,11 @@ export class EvmERC20WarpModule extends HyperlaneModule<
       ...this.createRemoteRoutersUpdateTxs(actualConfig, expectedConfig),
       ...this.createSetDestinationGasUpdateTxs(actualConfig, expectedConfig),
       ...this.createOwnershipUpdateTxs(actualConfig, expectedConfig),
-      ...this.updateProxyAdminOwnershipTxs(actualConfig, expectedConfig),
+      ...proxyAdminOwnershipUpdateTxs(
+        actualConfig,
+        expectedConfig,
+        this.domainId,
```
paulbalaji:
```suggestion
        this.chainId,
```

typescript/sdk/src/core/EvmCoreModule.ts
paulbalaji:
```suggestion
        this.chainId,
```

typescript/sdk/src/core/schemas.ts
xeno097: Technically it shouldn't be a problem because worst case scenario the user can just `core read` to 'update' the config to the latest version but it is an inconvenience anyways because the first time the CLI will fail and the fix is not that obvious. deferring to @ltyu for his thoughts. Maybe we should consider versioning configs in the future to avoid breaking changes and allows conversion to an old config file to a new one

typescript/cli/src/tests/commands/core.ts
xeno097: yep that silly of a mistake it was

PR #4768, opened 2024-10-25
feat: cli command for verify warp route contracts

ltyu:
### Description
Uses warp artifacts to derive the verification artifacts from etherscan and RPC api.

usage: `hyperlane warp verify --symbol`

### Drive-by
- Write to registry when blockexplorer api keys are added

### Backward compatibility
Yes

### Testing
Manual


typescript/sdk/src/deploy/verify/utils.ts
```diff
@@ -84,3 +91,204 @@ export function shouldAddVerificationInput(
...
+}
+
+export async function getEtherscanConstructorArgs({
+  blockExplorerApiKey,
+  blockExplorerApiUrl,
+  bytecode,
+  chainName,
+  contractAddress,
+  multiProvider,
+}: {
```
xeno097, line 148: It seems that these 2 functions have the same  input `type`, maybe it would be a good idea to create a single type def and reuse it for both

typescript/cli/src/verify/warp.ts
```diff
@@ -0,0 +1,137 @@
...
+  MultiProvider,
+  PostDeploymentContractVerifier,
+  TokenType,
+  VerificationInput,
+  WarpCoreConfig,
+  hypERC20contracts,
+  hypERC20factories,
+  isProxy,
+  proxyImplementation,
+  verificationUtils, // verificationUtils,
```
xeno097, line 17: I think this comment slipped in the review ehehe

xeno097 (commented):
🔥

typescript/cli/src/verify/warp.ts
ltyu: fixed

typescript/sdk/src/deploy/verify/utils.ts
ltyu: Generally yes! 
but i've reduced the parameters here by removing the blockExplorer keys and using the registry now, so maybe not worth it to add a type. lmk if you feel like i should :)

xeno097: you can keep it like this np! Generally, I like to create "base types" for types that have common fields and share a common context but I'll leave it up to you is not a deal breaker

xeno097 (approved):

PR #4769, opened 2024-10-26
test: append-more-multisigmIsm-test

tiendn:
### Description
Append more test for `multisigIsm.ts`

### Testing
Add 2 test cases:
- has a valid number of validators for each threshold
- has valid EVM addresses for each validator

### DIFF
--- a/typescript/sdk/src/consts/multisigIsm.test.ts
+++ b/typescript/sdk/src/consts/multisigIsm.test.ts
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { isAddress } from 'viem';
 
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/consts/multisigIsm.test.ts
```diff
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { isAddress } from 'viem';
```
jmrossy, line 2: Use our own address utilities from the `@hyperlane-xyz/utils` package

tiendn: Done, thank you :D

jmrossy (approved):

PR #4771, opened 2024-10-28
feat: widget update for de-duplication

Xaroz:
### Description

Update widgets with new components from the hyperlane explorer and warp UI. These will be used to replace the current components from both mentioned repo

### Drive-by changes

None

### Backward compatibility

Not exactly, Spinner and WideChevron were renamed to SpinnerIcon and WideChevronIcon, these must be updated accordingly 

### Testing

Manual and Visual Testing with storybook


typescript/widgets/src/components/DatetimeField.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { ChangeEvent } from 'react';
+
+interface Props {
+  timestamp: number | null;
+  onChange: (t: number | null) => void;
+  name?: string;
+}
+
+export function DatetimeField({ timestamp, onChange, name }: Props) {
```
jmrossy, line 9: Type the props as extensions to the fields they'll be used on, and then spread the `{...props}` onto the input field. See the svg icon components for an example

typescript/widgets/src/icons/Arrow.tsx
```diff
@@ -1,12 +1,14 @@
...
 
-type Props = SVGProps<SVGSVGElement> & {
+import { DefaultIconProps } from './types.js';
+
+type Props = DefaultIconProps & {
   direction: 'n' | 'e' | 's' | 'w';
 };
 
-function _ArrowIcon({ fill, className, direction, ...rest }: Props) {
+function _ArrowIcon({ color, fill, className, direction, ...rest }: Props) {
```
jmrossy, line 11: replace fill instead of having both fill and color

typescript/widgets/src/icons/Help.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { memo } from 'react';
+
+import { ColorPalette } from '../color.js';
+import { IconButton } from '../components/IconButton.js';
+
+import { QuestionMarkIcon } from './QuestionMark.js';
+
+function _HelpIcon({
```
jmrossy, line 8: This is redundant with the existing Tooltip component in this lib. Let's keep that other one and cut this one

typescript/widgets/src/index.ts
```diff
@@ -8,34 +9,50 @@ export {
...
 export { CopyButton } from './components/CopyButton.js';
+export { DatetimeField } from './components/DatetimeField.js';
 export { IconButton } from './components/IconButton.js';
 export { LinkButton } from './components/LinkButton.js';
 export { SegmentedControl } from './components/SegmentedControl.js';
+export { SelectField, type SelectOption } from './components/SelectField.js';
 export { TextInput } from './components/TextInput.js';
 export { Tooltip } from './components/Tooltip.js';
-export * from './consts.js';
+export { HYPERLANE_EXPLORER_API_URL } from './consts.js';
```
jmrossy, line 21: 👍

typescript/widgets/src/layout/Card.tsx
```diff
@@ -0,0 +1,25 @@
...
+export type CardProps = PropsWithChildren<{
+  className?: string;
+  inverseMargin?: boolean;
+}>;
+
+export function Card({ className, inverseMargin, children }: CardProps) {
+  return (
+    <div
+      className={`${
+        inverseMargin ? cardStyles.inverseMargin : cardStyles.padding
```
jmrossy, line 12: When would you want the inverse margins applied here? They exist in the warp ui so the notification banner can use them and be flush with the card edges. 

More broadly, I think the exist shape, shadowing, padding, and roundedness of cards are kind of an app-specific thing. I recommend we remove this for now and let the apps define their own card components, which are quite simple anyway

jmrossy (commented):
Thank you, overall looks good! Just a few minor things to fix please

typescript/widgets/src/components/DatetimeField.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+  };
+
+  return (
+    <input
+      type="datetime-local"
+      value={toShortIsoString(timestamp)}
+      onChange={handleChange}
+      name={name}
+      className="htw--ml-px"
```
jmrossy, line 26: This px is an explorer-specific style

typescript/widgets/src/icons/Help.tsx
Xaroz: I removed this component but our `Tooltip` component has an anchor tag as the parent, not sure if that is something we should look into or it was intended to be that way

typescript/widgets/src/layout/Card.tsx
Xaroz: Yeah makes sense to me, I will remove this and each app can have their own Card component

jmrossy (approved):

PR #4773, opened 2024-10-28
WIP DAO ISM

tkporter:
### Description

- Moves to JSON for our ICAs
- Keeps around old ICAs for now so we're capable of submitting txs to transfer ownership
- WIP before I create all these ICAs


typescript/infra/config/environments/mainnet3/aw-icas.json
```diff
@@ -0,0 +1,34 @@
+{
```
Mo-Hussain, line 1: Are these all the current ICAs?

typescript/infra/scripts/check/check-owner-ica.ts
```diff
@@ -55,23 +49,26 @@ async function main() {
...
+      routerOverride: ownerChainInterchainAccountRouter,
+    });
+    if (!eqAddress(expected.ica, actualAccount)) {
       mismatchedResults[chain] = {
-        Expected: expectedAddress,
-        Actual: actualAccount,
+        Expected: expected,
+        Actual: {
+          ica: actualAccount,
+          ism: expected.ism,
```
Mo-Hussain, line 70: Does it make sense to use the expected ism here?

typescript/infra/scripts/get-owner-ica.ts
```diff
@@ -27,26 +62,59 @@ function getArgs() {
...
   } = await getArgs();
   const config = getEnvironmentConfig(environment);
   const multiProvider = await config.getMultiProvider();
 
+  let artifacts: ChainMap<IcaArtifact>;
+  try {
+    artifacts = await readAbacusWorksIcas(environment);
+  } catch (err) {
+    console.error('Error reading artifacts, defaulting to no artifacts:', err);
+    artifacts = {};
```
Mo-Hussain, line 89: Could this have unintended consequences if someone mistakenly moves the file or renames it?

typescript/infra/scripts/get-owner-ica.ts
```diff
@@ -60,58 +128,319 @@ async function main() {
...
+    owner: routingIsmOwner,
+    domains: {
+      [originChain]: {
+        type: IsmType.AGGREGATION,
+        modules: [
+          // This will always use the default ISM.
+          // We burn ownership and have no domains in the routing table.
+          {
+            type: IsmType.FALLBACK_ROUTING,
+            owner: '0xdead00000000000000000000000000000000dead',
```
Mo-Hussain, line 425: Is this a typical burn address on ethereum?

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -371,17 +371,37 @@ export abstract class HyperlaneAppGovernor<
...
+          Used ISM override: ${ismOverride}, recovered account: ${recoveredAccount}. Is the ICA's ISM override correct?
+          Defaulting to manual submission.`,
+        ),
+      );
+      return {
+        type: SubmissionType.MANUAL,
+        chain,
+        call,
+      };
+    }
```
paulbalaji, line 403: so once we have deployed the new icas and uncomment out the relevant code for them in the config, i'm not sure I see how we still submit to the old icas to transfer ownership to the new ones

typescript/infra/scripts/get-owner-ica.ts
paulbalaji: I think `0x000000000000000000000000000000000000dEaD` is more commonly used

typescript/infra/scripts/check/check-owner-ica.ts
tkporter: mm good point, annoyingly we can't easily recover the ISM address given an ICA address, so I'll just say it's unknown

typescript/infra/scripts/get-owner-ica.ts
tkporter: Probably - the reason this doesn't throw is if the file doesn't exist due to valid reasons, e.g. for a new environment. But in the spirit of being extra conservative, happy to throw and leave a note about what to do if it doesn't exist

typescript/infra/scripts/get-owner-ica.ts
tkporter: happy to move to that

typescript/infra/src/govern/HyperlaneAppGovernor.ts
tkporter: Maybe I'm misunderstanding but I thought transferring would involve:
- changing the owner to the new ICA
- we keep `_icaAddress` as the old ICA, so we're capable of sending a tx via the old ICA to transfer to the new owner

And then after that's executed, we:
- update `_icaAddress` to be the new ICA and `_icaIsmAddress` as needed, so we can submit new txs via this new ICA

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -173,7 +180,12 @@ export const ethereumChainOwners: ChainMap<OwnableConfig> = Object.fromEntries(
           // So we need to keep the Safe and ICA addresses somewhere in the config
           // to be able to track down which addresses are SAFEs, ICAs, or standard SIGNERS.
           ...(safes[local] && { _safeAddress: safes[local] }),
-          ...(icas[local] && { _icaAddress: icas[local] }),
+          // After transitioning to the new ICAs, remove the following:
+          ...(oldIcas[local] && { _icaAddress: oldIcas[local] }),
+
+          // After transitioning to the new ICAs, uncomment the following:
+          // ...(icaArtifact && { _icaAddress: icaArtifact.ica }),
+          // ...(icaArtifact?.ism && { _icaIsmAddress: icaArtifact.ism }),
```
paulbalaji, line 188: should we throw if icaArtifact.ism isn't set?

typescript/infra/scripts/get-owner-ica.ts
Mo-Hussain: If we are deploying to a new env, we'd need to do a code change right? as in read from a different file maybe?

paulbalaji:
1. get-owner-ica
2. check-owner-ica
3. check-owner-ica --legacy
image

PR #4775, opened 2024-10-29
chore: disable arcadiatestnet until redeployment

paulbalaji:
### Description

chore: disable arcadiatestnet until redeployment
- testnet had to be reset

### Drive-by changes

- take down arcadiatestnet validator
- redeploy scraper, relayer, keyfunder without offending chain

### DIFF
--- a/typescript/infra/config/environments/testnet4/funding.ts
+++ b/typescript/infra/config/environments/testnet4/funding.ts
@@ -28,7 +28,7 @@ export const keyFunderConfig: KeyFunderConfig<
   desiredBalancePerChain: {
     alfajores: '5',
     arbitrumsepolia: '0.1',
-    arcadiatestnet: '0.1',
+    // arcadiatestnet: '0.1',
     basesepolia: '0.1',
     berabartio: '0.1',
     bsctestnet: '5',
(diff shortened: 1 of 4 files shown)


typescript/infra/config/environments/testnet4/funding.ts
```diff
@@ -28,7 +28,7 @@ export const keyFunderConfig: KeyFunderConfig<
   desiredBalancePerChain: {
     alfajores: '5',
     arbitrumsepolia: '0.1',
-    arcadiatestnet: '0.1',
+    // arcadiatestnet: '0.1',
```
tkporter, line 31: I don't totally recall but I think some infra scripts may create a provider for all supported chains and still fail even if we don't use the provider, but if this works thats fine

tkporter (approved):

typescript/infra/config/environments/testnet4/funding.ts
paulbalaji: `arcadiatestnet` been removed from the supported chains too - but I left the other references commented out for when we network stabilises again to add the deployment back

PR #4780, opened 2024-10-29
feat: add taint & toleration to mainnet3 relayer node pool

tkporter:
### Description

See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

- In GKE, for the larger relayer node pool, added a taint: `NoSchedule`, label `component=relayer`
- In our relayer workloads, they now have a toleration that matches that taint
- As part of rollout, first did a NoExecute (evicting everything on the big nodes), then moved to NoSchedule (so that some other pods like daemonsets would still get onto these bigger nodes). I made sure the hyperlane and neutron context relayers have these tolerations

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1309

### DIFF
--- a/typescript/infra/src/config/agent/agent.ts
+++ b/typescript/infra/src/config/agent/agent.ts
@@ -36,6 +36,7 @@ export interface HelmRootAgentValues {
   image: HelmImageValues;
   hyperlane: HelmHyperlaneValues;
   nameOverride?: string;
+  tolerations?: KubernetesToleration[];
 }
 
 // See rust/main/helm/values.yaml for the full list of options and their defaults.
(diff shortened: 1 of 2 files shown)


typescript/infra/src/config/agent/agent.ts
```diff
@@ -36,6 +36,7 @@ export interface HelmRootAgentValues {
   image: HelmImageValues;
   hyperlane: HelmHyperlaneValues;
   nameOverride?: string;
+  tolerations?: KubernetesToleration[];
```
tkporter, line 39: the helm charts already honor this

paulbalaji (approved):
lgtm - i referred to the k8s docs to understand more about how this works too https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

PR #4782, opened 2024-10-30
feat: add a script for getting typical remote gas amounts

tkporter:
### Description

- Sometimes Jake requests an update on these for his own models, this makes it easier for us to get it for him

### DIFF
--- a/typescript/infra/scripts/get-typical-remote-gas-amounts.ts
+++ b/typescript/infra/scripts/get-typical-remote-gas-amounts.ts
@@ -1,0 +1,48 @@
+import { ChainMap } from '@hyperlane-xyz/sdk';
+
+import { getOverheadWithOverrides } from '../config/environments/mainnet3/igp.js';
+import { getTypicalRemoteGasAmount } from '../src/config/gas-oracle.js';
+
+import { getArgs } from './agent-utils.js';
+import { getEnvironmentConfig } from './core-utils.js';
+
+// This script exists to print the typical local -> remote gas amounts for a given environment.
+// This is useful for Jake to use in his own models for assessing message costs.
+
+async function main() {
+  const args = await getArgs().argv;
+
+  if (args.environment !== 'mainnet3') {
+    throw new Error('This script only supports the mainnet3 environment');
+  }
+
+  const environmentConfig = getEnvironmentConfig(args.environment);
+
+  // Local -> Remote -> Amount of gas.
+  // Local is important because depending on the validator threshold, the cost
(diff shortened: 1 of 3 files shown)


typescript/infra/src/config/gas-oracle.ts
```diff
@@ -112,6 +116,15 @@ function getLocalStorageGasOracleConfigOverride(
   }, {});
 }
 
+export function getTypicalRemoteGasAmount(
+  local: ChainName,
+  remote: ChainName,
+  getOverhead: (local: ChainName, remote: ChainName) => number,
+): number {
+  const handleGasUsage = 50_000;
```
paulbalaji, line 124: worth making this a public const somewhere if we reuse 50k somewhere else?
```suggestion
  const handleGasUsage = 50_000;
```

typescript/infra/scripts/get-typical-remote-gas-amounts.ts
```diff
@@ -0,0 +1,48 @@
...
+      amounts[local] = amounts[local] || {};
+      amounts[local][remote] = getTypicalRemoteGasAmount(
+        local,
+        remote,
+        getOverheadWithOverrides,
+      );
+    }
+  }
+
+  console.log(JSON.stringify(amounts, null, 2));
```
paulbalaji, line 42: couls use `stringifyObject(amounts, 'json', 2)`

paulbalaji (approved):

typescript/infra/scripts/get-typical-remote-gas-amounts.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getArgs } from './agent-utils.js';
+import { getEnvironmentConfig } from './core-utils.js';
+
+// This script exists to print the typical local -> remote gas amounts for a given environment.
+// This is useful for Jake to use in his own models for assessing message costs.
+
+async function main() {
+  const args = await getArgs().argv;
+
+  if (args.environment !== 'mainnet3') {
```
Mo-Hussain, line 16: nit: do we not have an enum for this?

Mo-Hussain (approved):

typescript/infra/scripts/get-typical-remote-gas-amounts.ts
tkporter: I think this is something you called out a couple months ago as well - we don't and the reason was that it's sorta tough to do without a gnarly cyclic dependency. Happy to have someone add it at some point but I spent > 1hr previously to no avail

PR #4785, opened 2024-10-30
feat: oct 30 enrollments

paulbalaji:
### Description

- update validators for op, mode, lisk as part of superlane
- add merkly/mitosis validators for apechain, arbitrumnova, b3, fantom, gravity, harmony, kaia, morph, orderly, snaxchain, zeronetwork, zksync
- add blockhunters to default set on redstone, cheesechain

### Drive-by changes

- igp updates
- introduce `chainsToSkip` list to reuse in deploy.ts and check-deploy.ts
- filter out `chainsToSkip` in deploy.ts and check-deploy.ts
- add retry logic when doing safe api operations in app governor 
- update scripts for checking/getting ICAs

### Related issues

na

### Backward compatibility

yes

### Testing

manual


typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -128,16 +128,29 @@ export const icas: Partial<
...
+  // rootstock: '0x69350aeA98c5195F2c3cC6E6A065d0d8B12F659A',
+  // alephzeroevm: '0x004a4C2e4Cd4F5Bd564fe0A6Ab2Da56258aE576f',
+  // chiliz: '0xb52D281aD2BA9761c16f400d755837493e2baDB7',
+  // lumia: '0x418E10Ac9e0b84022d0636228d05bc74172e0e41',
+  // superposition: '0x34b57ff8fBA8da0cFdA795CC0F874FfaB14B1DE9',
+  // flow: '0xf48377f8A3ddA7AAD7C2460C81d939434c829b45',
+  // metall2: '0x2f1b1B0Fb7652E621316460f6c3b019F61d8dC9a',
+  // polynomial: '0xC20eFa1e5A378af9233e9b24515eb3408d43f900',
+
+  // // Oct 30, 2024 batch
```
jmrossy, line 142: Did you meant to leave this commented?

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -568,25 +569,58 @@ export abstract class HyperlaneAppGovernor<
         );
         this.canPropose[chain].set(safeAddress, canPropose);
       } catch (error) {
+        if (!(error instanceof Error)) {
```
jmrossy, line 572: When would this happen?

typescript/infra/config/environments/mainnet3/owners.ts
paulbalaji: yup, intentional for now - will complete the governance process in a separate step

typescript/infra/src/govern/HyperlaneAppGovernor.ts
paulbalaji: I'm not sure, I mainly did it to keep the typings further down - should I instead just assert?

```ts
} catch (error) {
	assert(error instanceof Error, 'Error is not an instance of Error');
```

jmrossy: Yeah that could work. does `catch (error: Error)` work? Otherwise an `as Error` seems also fine here

paulbalaji: `error: Error` won't work because it expects `any` or `unknown` - i've changed it to `as Error` when reading the error message

typescript/infra/scripts/check/check-owner-ica.ts
```diff
@@ -29,17 +32,32 @@ async function main() {
...
 
+  const checkOwnerIcaChains = (
+    chains?.length ? chains : Object.keys(icas)
+  ).filter(isEthereumProtocolChain);
+
   const ownerConfig: AccountConfig = {
     origin: ownerChain,
     owner: owner,
   };
+  const ownerChainInterchainAccountRouter =
```
tkporter, line 43: should there also be a zero check here? just in case

typescript/infra/scripts/get-owner-ica.ts
```diff
@@ -52,6 +52,8 @@ async function main() {
     origin: ownerChain,
     owner: originOwner,
   };
+  const ownerChainInterchainAccountRouter =
```
tkporter, line 55: ditto

tkporter (approved):

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -70,6 +70,11 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
   //     confirmations: 3,
   //   },
   // },
+  // taiko: {
```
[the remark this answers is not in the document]
paulbalaji: ah it wasn't needed in the end, will remove

typescript/infra/scripts/check/check-owner-ica.ts
paulbalaji: hopefully never hit it, but yes i'll add it

PR #4787, opened 2024-10-30
feat(warpMonitor): Use coingecko api key for value monitoring

Mo-Hussain:
### Description

- Fetching coingecko api from GCP secrets and suing with CoinGecko Client
- Using coingecko-api-v3 package to support passing API keys 

### Drive-By Changes

- Rename eclipse warp route config files and add `tokenCoinGeckoId`

### Testing

Manual


typescript/sdk/package.json
```diff
@@ -14,10 +14,9 @@
     "@safe-global/safe-deployments": "1.37.8",
     "@solana/spl-token": "^0.3.8",
     "@solana/web3.js": "^1.78.0",
-    "@types/coingecko-api": "^1.0.10",
     "@wagmi/chains": "^1.8.0",
     "bignumber.js": "^9.1.1",
-    "coingecko-api": "^1.0.10",
+    "coingecko-api-v3": "^0.0.29",
```
jmrossy, line 19: It's strange that they put out a new package instead of just making a new major version for the package they already had

jmrossy (approved):

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -85,10 +84,11 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
 
   static withDefaultCoinGecko(
     chainMetadata: ChainMap<ChainMetadata>,
+    apiKey?: string,
     expirySeconds?: number,
     sleepMsBetweenRequests = 5000,
   ): CoinGeckoTokenPriceGetter {
-    const coinGecko = new CoinGecko();
+    const coinGecko = new CoinGeckoClient({}, apiKey);
```
paulbalaji, line 91: does it still behave ok if we don't provide an api key (e.g. when doing igp config in cli)

typescript/sdk/src/gas/token-prices.ts
```diff
@@ -97,6 +97,10 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
     );
   }
 
+  static getSecretName(environment: string): string {
```
paulbalaji, line 100: this feels very infra specific that should probably be there instead of the sdk

Mo-Hussain: Good call out, I'll move this out

typescript/sdk/src/gas/token-prices.ts
Mo-Hussain: Yes this is optional, do we use this class for doing igp config in cli?

paulbalaji: i'm mistaken we do not - we hit the rest api directly

typescript/sdk/package.json
tkporter: interesting, seems like there's no official JS/TS sdk for coingecko, so these are different authors

tkporter (approved):

PR #4795, opened 2024-10-31
feat(widgets): ChainSearchMenu improvements

### LINKED ISSUE #4779: Chain picker UX improvements
Mostly based on feedback from [this discord thread](https://discord.com/channels/935678348330434570/1008821009731829810/1293671472900083785)

1. Include option to hide 2nd column by setting `customListItemField` prop to `null` 
2. Instead of auto-sorting by `customListItemField` when it's set, let that be customizable when a `defaultSortField` prop
3. Have search input be auto-focused when menu is shown
4. When the search results contain only 1 chain, have it be selected by pressing the `enter` key in the search form
5. Show a small `x`button on the right side of search input field when it has text inputted. Clicking x clears the input
6. Restyle the search input: Replace background color with a border. Darken border on focus

Rough example:
Screenshot 2024-10-29 at 12 33 51 PM

Xaroz:
### Description

- Updated ChainSearchMenu with improvements mentioned in #4779 
- Update ChainSearchMenu story with stories related to the new changes

### Drive-by changes

No

### Related issues

Fixes #4779

### Backward compatibility

Yes

### Testing

Manual testing and visual testing with storybook

### DIFF
--- a/.changeset/plenty-pens-peel.md
+++ b/.changeset/plenty-pens-peel.md
@@ -9,3 +9,4 @@ Update widgets with components from explorer and warp ui
 - Add components: DatetimeField and SelectField
 - New stories: IconList and Fade
 - Add "Icon" suffix for icons that did not have it
+- Update ChainSearchMenu with improvements
--- a/.changeset/plenty-pens-peel.md
+++ b/.changeset/plenty-pens-peel.md
@@ -9,4 +9,3 @@ Update widgets with components from explorer and warp ui
 - Add components: DatetimeField and SelectField
 - New stories: IconList and Fade
 - Add "Icon" suffix for icons that did not have it
-- Update ChainSearchMenu with improvements
--- a/typescript/widgets/src/chains/ChainSearchMenu.tsx
+++ b/typescript/widgets/src/chains/ChainSearchMenu.tsx
@@ -316,7 +316,7 @@ function useCustomizedListItems(
     [defaultSortField],
   ) as ChainSortByOption[];
 
-  // Sort by defaultSortField initially, if value is "custom", sort by custom field by default
+  // Sort by defaultSortField initially, if value is "custom", sort using custom field by default
   const defaultSortState = useMemo(
     () =>
       defaultSortField
--- a/typescript/widgets/src/chains/ChainSearchMenu.tsx
+++ b/typescript/widgets/src/chains/ChainSearchMenu.tsx
@@ -62,7 +62,7 @@ export interface ChainSearchMenuProps {
   showAddChainMenu?: boolean;
   // Include add button above list
   showAddChainButton?: boolean;
-  // Use this field to sort by default
+  // Field by which data will be sorted by default
   defaultSortField?: DefaultSortField;
 }
 
--- a/typescript/widgets/src/chains/ChainSearchMenu.tsx
+++ b/typescript/widgets/src/chains/ChainSearchMenu.tsx
@@ -19,7 +19,7 @@ import { ChainAddMenu } from './ChainAddMenu.js';
 import { ChainDetailsMenu } from './ChainDetailsMenu.js';
 import { ChainLogo } from './ChainLogo.js';
 
-enum ChainSortByOption {
+export enum ChainSortByOption {
   Name = 'name',
   ChainId = 'chain id',
   Protocol = 'protocol',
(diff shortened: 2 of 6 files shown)


typescript/widgets/src/chains/ChainSearchMenu.tsx
```diff
@@ -303,19 +313,22 @@ function useCustomizedListItems(customListItemField) {
       ...(customListItemField ? [customListItemField.header] : []),
       ...Object.values(ChainSortByOption),
     ],
-    [customListItemField],
+    [defaultSortField],
```
jmrossy, line 316: This looks incorrect. `customListItemField` is used in the memo fn but `customListItemField` isn't. I would have expected the linter to catch this

typescript/widgets/src/components/SearchMenu.tsx
```diff
@@ -91,12 +99,28 @@ export function SearchMenu<
     [data, searchQuery, sortState, filterState, searchFn],
   );
 
+  const handleKeyDown = useCallback(
```
jmrossy, line 102: This approach is okay but what I expected to see was that the input field get wrapped by a `form`. That way, the enter key would automatically trigger the form submit, and then you'd execute your callback here in that submit fn. WDYT of that approach? Feels a bit more robust to me than watching for a specific key name maybe?

typescript/widgets/src/components/SearchMenu.tsx
```diff
@@ -157,21 +181,36 @@ export function SearchMenu<
...
         className="htw-absolute htw-left-4 htw-top-1/2 -htw-translate-y-1/2 htw-opacity-50"
       />
+
       <TextInput
+        onChange={onChange}
+        value={value}
+        ref={ref}
         {...props}
-        className="htw-w-full htw-rounded-lg htw-px-11 htw-py-3"
+        className="htw-bg-transparent focus:htw-bg-transparent htw-border htw-border-gray-200 focus:htw-border-gray-400 htw-w-full htw-rounded-lg htw-px-11 htw-py-3"
```
jmrossy, line 201: Is the bg-transparent class necessary here? I don't think it's necessary?

.changeset/plenty-pens-peel.md
```diff
@@ -9,3 +9,4 @@ Update widgets with components from explorer and warp ui
 - Add components: DatetimeField and SelectField
 - New stories: IconList and Fade
 - Add "Icon" suffix for icons that did not have it
+- Update ChainSearchMenu with improvements
```
jmrossy, line 12: This is fine but FYI you don't have to manually manage your changeset files like this. You can just make a new one for each PR and the tooling with stitch them together when we release

jmrossy (commented):
Mostly looks good, just some small things to fix please

typescript/widgets/src/components/SearchMenu.tsx
Xaroz: Wrapped the `input` around a `form` and change `handleKeyDown` to `handleSubmit`

typescript/widgets/src/chains/ChainSearchMenu.tsx
Xaroz: Fixed dependecy

typescript/widgets/src/components/SearchMenu.tsx
Xaroz: I have to put `bg-transparent` or `bg-inherit` because the `input` by default comes with a background color, for now I changed to `bg-inherit` so that the color comes directly from the parent.

Lmk if you would rather have `bg-transparent` or if I should change the `input` bg color directly (although this would affect where it is used in other places

typescript/widgets/src/chains/ChainSearchMenu.tsx
```diff
@@ -53,13 +55,15 @@ export interface ChainSearchMenuProps {
...
   // Replace the default 2nd column (deployer) with custom data
-  customListItemField?: CustomListItemField;
+  customListItemField?: CustomListItemField | null;
   // Auto-navigate to a chain details menu
   showChainDetails?: ChainName;
   // Auto-navigate to a chain add menu
   showAddChainMenu?: boolean;
   // Include add button above list
   showAddChainButton?: boolean;
+  // Use this field to sort by default
```
jmrossy, line 65:
```suggestion
  // Field by which data will be sorted by default
```

jmrossy (approved):
Nice!

PR #4798, opened 2024-11-01
feat: detangle chainId / domainId EVM assumptions

paulbalaji:
### Description

Detangle assumption that chainId == domainId for EVM chains.
- required to support new domain IDs for sept26 chain deploy batch https://github.com/hyperlane-xyz/hyperlane-registry/pull/352

Domain IDs and Chain Names are still unique, but chainId is no longer guaranteed to be a unique identifier.
- required to support shadow lumia (current deployment) + prod lumia (new deployment with different domain id)

Touches:
- evm core/ism/hook/warp modules
- multiprovider + chain metadata manager
- safe service
- renaming `ChainNameOrId` to `ChainNameOrDomain` (where bulk of the diff comes from)

### Drive-by changes

- ~~add name of a `chain` to `Annotated` types, since chain ID on a populated transaction is a looser tie to a chain than `domainId` or `chain`~~
- ~~simplify multiprovider.sendtransaction, since `chain` is provided with the annotated transaction~~

- ensure domainId is not optional in the chain metadata schema
	- requires registry fix https://github.com/hyperlane-xyz/hyperlane-registry/pull/357
- update ethers v5 json rpc submitter to require a `chain` as well

### Related issues

- mostly redoes work from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4599
- will have to be rechecked after zksync changes are also in

### Backward compatibility

no

### Testing

ci, manual


typescript/sdk/src/providers/transactions/transformer/ethersV5/EV5InterchainAccountTxTransformer.ts
```diff
@@ -37,18 +36,16 @@ export class EV5InterchainAccountTxTransformer
...
+        { to, data, chain }: AnnotatedEV5Transaction,
       ) => {
-        assert(chainId, 'Invalid PopulatedTransaction: chainId is required');
-        assert(to, 'Invalid PopulatedTransaction: to is required');
-        assert(data, 'Invalid PopulatedTransaction: data is required');
-        const txChain = this.multiProvider.getChainName(chainId);
-        txChainToInnerCalls[txChain] ||= [];
-        txChainToInnerCalls[txChain].push({ to, data });
+        assert(to, 'Invalid AnnotatedEV5Transaction: "to" is required');
+        assert(data, 'Invalid AnnotatedEV5Transaction: "data" is required');
```
tkporter, line 46: does this need an assert on `chain`?

typescript/infra/scripts/funding/fund-deterministic-key-from-deployer.ts
```diff
@@ -79,7 +79,8 @@ async function main() {
         logger.info(
           `Funding ${address} on chain '${chain}' with ${value} native tokens`,
         );
-        await multiProvider.sendTransaction(chain, {
+        await multiProvider.sendTransaction({
```
tkporter, line 82: I did kinda like the old interface fwiw but will defer to ppl closer to the SDK

tkporter (commented):
lgtm but deferring to others

typescript/infra/scripts/funding/fund-deterministic-key-from-deployer.ts
paulbalaji: fair, I just felt it was cleaner to cut the arg out since it's redundant to mention `chain` twice

typescript/sdk/src/metadata/ChainMetadataManager.ts
```diff
@@ -56,59 +62,56 @@ export class ChainMetadataManager<MetaExt = {}> {
...
   addChain(metadata: ChainMetadata<MetaExt>): void {
     ChainMetadataSchema.parse(metadata);
-    // Ensure no two chains have overlapping names/domainIds/chainIds
+    // Ensure no two chains have overlapping names/domainIds
     for (const chainMetadata of Object.values(this.metadata)) {
-      const { name, chainId, domainId } = chainMetadata;
+      const { name, domainId } = chainMetadata;
       if (name == metadata.name)
         throw new Error(`Duplicate chain name: ${name}`);
-      // Chain and Domain Ids should be globally unique
```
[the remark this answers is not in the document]
paulbalaji: CLI still defaults to domainId being the chainId, so we'll still be able to catch duplicates in the general case - we can also add stricter tests in the registry itself rather than trying to shape it here

> don't create new deploys on chains that already have mailboxes

the cli currently checks if we've got a mailbox for the selected `chain`

typescript/sdk/src/providers/transactions/transformer/ethersV5/EV5InterchainAccountTxTransformer.ts
paulbalaji: We don't *need* one because `AnnotatedEV5Transaction` extends from an ethers `PopulatedTransaction` where all fields are optional - so we need the assert to ensure values from that type are properly set.

```ts
import type { PopulatedTransaction as EV5Transaction } from 'ethers';

export type AnnotatedEV5Transaction = Annotated<EV5Transaction>;
```
image


Whereas `chain` is a required property of the `Annotated<>` wrapper we're using


```ts
export type Annotated<T> = T & {
  chain: string;
  annotation?: string;
};
```

typescript/infra/scripts/funding/fund-deterministic-key-from-deployer.ts
xeno097: I prefer the cleaner approach in my case

typescript/sdk/src/providers/transactions/transformer/ethersV5/EV5InterchainAccountTxTransformer.ts
tkporter: makes sense!

xeno097 (approved):
Good effort 🔥

typescript/sdk/src/providers/transactions/transformer/ethersV5/EV5InterchainAccountTxTransformer.ts
yorhodes: `Pick` and `Required` might be useful here
https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype

yorhodes (commented):
submitting stale review

typescript/cli/examples/submit/strategy/json-rpc-chain-strategy.yaml
```diff
@@ -1,6 +1,8 @@
 anvil2:
   submitter:
+    chain: anvil2
     type: jsonRpc
```
yorhodes, line 4: just flagging that the top level key is redundant with the inner key

typescript/cli/src/deploy/warp.ts
```diff
@@ -885,12 +885,20 @@ async function submitWarpApplyTransactions(
   params: WarpApplyParams,
   chainTransactions: Record<string, AnnotatedEV5Transaction[]>,
```
yorhodes, line 886: the key of this is a chainId as a string?

typescript/sdk/src/types.ts
```diff
@@ -1,12 +1,12 @@
...
+import type { Domain } from '@hyperlane-xyz/utils';
 
 // An alias for string to clarify type is a chain name
 export type ChainName = string;
 // A map of chain names to a value type
-export type ChainMap<Value> = Record<string, Value>;
+export type ChainMap<Value> = Record<ChainName, Value>;
 
-export type ChainNameOrId = ChainName | ChainId | Domain;
+export type ChainNameOrDomain = ChainName | Domain;
```
yorhodes, line 10: I personally think changing this types name is not really necessary and is the main reason this diff is so large

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -818,8 +817,8 @@ class ContextFunder {
   ) {
     const l1Chain = L2ToL1[l2Chain];
     const crossChainMessenger = new CrossChainMessenger({
-      l1ChainId: this.multiProvider.getDomainId(l1Chain),
-      l2ChainId: this.multiProvider.getDomainId(l2Chain),
+      l1ChainId: this.multiProvider.getEvmChainId(l1Chain),
```
yorhodes, line 820: isnt the multiprovider already EVM specific?

typescript/widgets/src/chains/ChainAddMenu.tsx
```diff
@@ -155,14 +155,7 @@ function tryParseMetadataInput(
     return failure('name is already in use by another chain');
   }
 
-  if (multiProvider.tryGetChainMetadata(newMetadata.chainId)) {
-    return failure('chainId is already in use by another chain');
-  }
```
yorhodes, line 160: honestly im not sure we should remove this constraint from the multiprovider as a signer can think they are signing for chain X but their transaction can be replayed on chain Y

typescript/widgets/src/messages/useMessageStage.ts
```diff
@@ -172,17 +172,15 @@ async function fetchMessageState(
...
-) {
-  const metadata = await multiProvider.getChainMetadata(domainId);
-  if (metadata?.blocks?.confirmations) return metadata.blocks.confirmations;
-  else return DEFAULT_FINALITY_BLOCKS;
+function getFinalityBlocks(domainId: number, multiProvider: MultiProvider) {
+  const metadata = multiProvider.getChainMetadata(domainId);
+  return metadata?.blocks?.confirmations !== undefined
+    ? metadata.blocks.confirmations
+    : DEFAULT_FINALITY_BLOCKS;
 }
```
yorhodes, line 180: seems like an unnecessary diff

yorhodes (approved):
mostly lgtm
I dont think this needed to such a breaking change

typescript/infra/scripts/funding/fund-keys-from-deployer.ts
paulbalaji: the multiprovider is evm specific but the function itself is from the chain metadata manager which is not chain specific

it's effectively a shorthand for ensuring we get a number in EVM land

typescript/cli/src/deploy/warp.ts
paulbalaji: that's right - but it's because the lodash `groupBy` helper returns a Dictionary whose keys are always strings

typescript/cli/examples/submit/strategy/json-rpc-chain-strategy.yaml
paulbalaji: aye, just copying the pattern from the other submitter strategies here

typescript/widgets/src/chains/ChainAddMenu.tsx
paulbalaji: I think this is probably fine for us because my understanding is with our model we want to make sure that if we have a chain X and chain Y with the same chain ID, they are both actually chain X but just with different domain X and domain Y

typescript/cli/src/deploy/warp.ts
yorhodes: I did not realize we added lodash
@jmrossy would have a conniption

paulbalaji: aha yeah idk when it got added but it's been there for a little bit

yorhodes: https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/1474865ae14d014a86893d79ac0a36eea43e4a70 I guess he was onboard :)

paulbalaji: I forgot I was the one who added it 🤣

typescript/sdk/src/providers/transactions/submitter/TxSubmitterInterface.ts
```diff
@@ -22,7 +22,7 @@ export interface TxSubmitterInterface<TProtocol extends ProtocolType> {
    * @param txs The array of transactions to execute
    */
   submit(
-    ...txs: ProtocolTypedTransaction<TProtocol>['transaction'][]
+    ...txs: Annotated<ProtocolTypedTransaction<TProtocol>['transaction']>[]
```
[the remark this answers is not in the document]
paulbalaji: added to scratch notes for future

typescript/utils/src/types.ts
```diff
@@ -17,6 +17,7 @@ export const ProtocolSmallestUnit = {
 
 /********* BASIC TYPES *********/
 export type Domain = number;
+export type EvmChainId = number;
 export type ChainId = string | number;
```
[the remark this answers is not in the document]
paulbalaji: reduces need to parseInt in places by making the expected type explicit in the evm-specific classes

typescript/cli/src/deploy/warp.ts
jmrossy: Just be sure to use `lodash-es`!

PR #4801, opened 2024-11-01
feat: Scraper populates transaction fields for Solana

ameten:
### Description

Scraper populates transaction fields for Solana

### Related issues

### Backward compatibility

Yes (configuration is updated)

### Testing

Manual test of Scraper
Run E2E tests for Ethereum and Solana

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/provider.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/provider.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use solana_sdk::signature::Signature;
-use solana_transaction_status::EncodedTransaction;
+use solana_transaction_status::{EncodedTransaction, UiMessage};
 
 use hyperlane_core::{
     BlockInfo, ChainCommunicationError, ChainInfo, ChainResult, HyperlaneChain, HyperlaneDomain,
@@ -76,16 +76,32 @@ impl HyperlaneProvider for SealevelProvider {
     /// for all chains, not only Ethereum-like chains.
     async fn get_txn_by_hash(&self, hash: &H512) -> ChainResult<TxnInfo> {
         let signature = Signature::new(hash.as_bytes());
-        let transaction = self.rpc_client.get_transaction(&signature).await?;
 
-        let ui_transaction = match transaction.transaction.transaction {
+        let txn_confirmed = self.rpc_client.get_transaction(&signature).await?;
+        let txn_with_meta = txn_confirmed.transaction;
+
+        let txn = match txn_with_meta.transaction {
             EncodedTransaction::Json(t) => t,
             t => Err(Into::<ChainCommunicationError>::into(
                 HyperlaneSealevelError::UnsupportedTransactionEncoding(t),
             ))?,
         };
 
-        let received_signature = ui_transaction
+        let message = match txn.message {
+            UiMessage::Parsed(m) => m,
+            m => Err(Into::<ChainCommunicationError>::into(
+                HyperlaneSealevelError::UnsupportedMessageEncoding(m),
+            ))?,
+        };
+
+        let signer = message
+            .account_keys
+            .first()
+            .ok_or(HyperlaneSealevelError::UnsignedTransaction(*hash))?;
+        let pubkey = decode_pubkey(&signer.pubkey)?;
+        let sender = H256::from_slice(&pubkey.to_bytes());
+
+        let received_signature = txn
             .signatures
             .first()
             .ok_or(HyperlaneSealevelError::UnsignedTransaction(*hash))?;
@@ -113,7 +129,7 @@ impl HyperlaneProvider for SealevelProvider {
             max_fee_per_gas: None,
             gas_price: None,
             nonce: 0,
-            sender: Default::default(),
+            sender,
             recipient: None,
             receipt: Some(receipt),
             raw_input_data: None,
--- a/rust/main/chains/hyperlane-sealevel/src/provider.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/provider.rs
@@ -4,7 +4,7 @@ use async_trait::async_trait;
 use solana_sdk::signature::Signature;
 use solana_transaction_status::{
     option_serializer::OptionSerializer, EncodedTransaction, EncodedTransactionWithStatusMeta,
-    UiMessage, UiTransaction,
+    UiMessage, UiTransaction, UiTransactionStatusMeta,
 };
 
 use hyperlane_core::{
@@ -65,7 +65,7 @@ impl SealevelProvider {
         Ok(())
     }
 
-    fn extract_sender(hash: &H512, txn: &UiTransaction) -> ChainResult<H256> {
+    fn sender(hash: &H512, txn: &UiTransaction) -> ChainResult<H256> {
         let message = match &txn.message {
             UiMessage::Parsed(m) => m,
             m => Err(Into::<ChainCommunicationError>::into(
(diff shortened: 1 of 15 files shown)


rust/main/config/mainnet_config.json
```diff
@@ -2821,6 +2822,7 @@
       "name": "solanamainnet",
       "nativeToken": {
         "decimals": 9,
+        "denom": "lamport",
```
paulbalaji, line 2825: if denom is not provided, it defaults to "" - what happens in this case?

ie is there now an expectation that all sealevel nativeToken metadata now includes `denom` too?

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -2,37 +2,110 @@ use std::sync::Arc;
...
+    HyperlaneProvider, HyperlaneProviderError, NativeToken, TxnInfo, TxnReceiptInfo, H256, H512,
+    U256,
 };
 
 use crate::error::HyperlaneSealevelError;
 use crate::utils::{decode_h256, decode_h512, decode_pubkey};
 use crate::{ConnectionConf, SealevelRpcClient};
 
+/// Exponent value for atto units (10^-18).
+const ATTO_EXPONENT: u32 = 18;
```
daniel-savu, line 21: noting that this is also defined in `hyperlane-cosmos` in `provider.rs`

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -2,37 +2,110 @@ use std::sync::Arc;
...
+        Ok(U256::from(gas))
+    }
+
+    fn fee(&self, meta: &UiTransactionStatusMeta) -> ChainResult<U256> {
+        let exponent = ATTO_EXPONENT - self.native_token.decimals;
+        let coefficient = U256::from(10u128.pow(exponent));
+
+        let amount_in_native_denom = U256::from(meta.fee);
+
+        Ok(amount_in_native_denom * coefficient)
```
daniel-savu, line 99: should we use `checked_mul` to make sure any overflows don't go undetected?

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -76,44 +150,38 @@ impl HyperlaneProvider for SealevelProvider {
...
-                    Box::new(received_hash),
-                ),
-            ))?;
-        }
+        Self::validate_transaction(hash, txn)?;
+        let sender = Self::sender(hash, txn)?;
+        let meta = Self::meta(txn_with_meta)?;
+        let gas_used = Self::gas(meta)?;
+        let fee = self.fee(meta)?;
+        let gas_price = Some(fee / gas_used);
```
daniel-savu, line 169: should we use `checked_div`?

rust/main/config/mainnet_config.json
daniel-savu: afaict this field isn't used anywhere in agent logic, but is here just to be stored in the scraper DB along with the other data. So if it's empty and the chain isn't scraped there is no implication, otherwise the DB column will just be the empty string - but this assumption already existed before this PR

daniel-savu (approved):

rust/main/config/mainnet_config.json
tkporter: yeah if possible I think we should try to keep the most succinct config surface area - it seems like it was provided here for completeness but isn't necessary? Imo just requiring `denom` for Cosmos is most natural and having it for Sealevel chains too could create some confusion as to what exactly needs to be configured when adding a Sealevel chain. So if possible I'd probably prefer to just keep denom as a Cosmos thing, possibly even making it an Option<String> for the new `NativeToken` that isn't Cosmos-specific

ameten: Field `nativeToken.denom` is not used in Scraper logic for Sealevel chain, so, we can default it to any value. 

For Sealevel chains, though, the denomination depends on the chain itself: Solana - `lamport`, Eclipse - `gwei`.

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -2,37 +2,110 @@ use std::sync::Arc;
...
+            ))?;
+        }
+        Ok(())
+    }
+
+    fn sender(hash: &H512, txn: &UiTransaction) -> ChainResult<H256> {
+        let message = match &txn.message {
+            UiMessage::Parsed(m) => m,
+            m => Err(Into::<ChainCommunicationError>::into(
+                HyperlaneSealevelError::UnsupportedMessageEncoding(m.clone()),
```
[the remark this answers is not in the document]
ameten: It is correct. We don't expect this error and error below in this method to be thrown. If it happens, something is wrong with RPC and we'll be able to see it in logs.

rust/main/config/mainnet_config.json
ameten: I added defaults so that we don't need to add `nativeToken` to the configuration.

rust/main/chains/hyperlane-sealevel/src/provider.rs
ameten: Added

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -2,37 +2,110 @@ use std::sync::Arc;
...
+
+    fn gas(meta: &UiTransactionStatusMeta) -> ChainResult<U256> {
+        let OptionSerializer::Some(gas) = meta.compute_units_consumed else {
+            Err(HyperlaneSealevelError::EmptyComputeUnitsConsumed)?
+        };
+
+        Ok(U256::from(gas))
+    }
+
+    fn fee(&self, meta: &UiTransactionStatusMeta) -> ChainResult<U256> {
```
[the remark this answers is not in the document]
ameten: We need it since the compute unit (gas unit) costs less than 1 lamport on Solana and we need to express the gas price in an integer number. So, we convert all fees into units of 10^-18 (same for all the chains) so that gas price is meaningful.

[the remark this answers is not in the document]
ameten: I added a comment into the source code.

rust/main/config/mainnet_config.json
```diff
@@ -793,6 +793,7 @@
       "name": "eclipsemainnet",
       "nativeToken": {
         "decimals": 9,
```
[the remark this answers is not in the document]
ameten: Added the default which is protocol dependent.

rust/main/hyperlane-base/src/settings/parser/connection_parser.rs
```diff
@@ -174,6 +157,41 @@ pub fn build_cosmos_connection_conf(
...
+    }
+}
+
+fn parse_native_token(chain: &ValueParser, err: &mut ConfigParsingError) -> NativeToken {
+    let native_token_decimals = chain
+        .chain(err)
+        .get_key("nativeToken")
+        .get_key("decimals")
+        .parse_u32()
+        .unwrap_or(18);
```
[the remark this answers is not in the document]
ameten: Done

rust/main/chains/hyperlane-sealevel/src/provider.rs
ameten: I added a check which reports a warning as it is done for Cosmos. Otherwise, gas price will be zero which is the edge case we accepted.

rust/main/chains/hyperlane-sealevel/src/provider.rs
ameten: Refactored.

tkporter (approved):

PR #4804, opened 2024-11-01
fix(cli): re-enable space key for multiselect prompt

xeno097:
### Description

Re-enables space key press for multi-select prompt

### Drive-by changes

- None

### Backward compatibility

- YES

### Testing

- Manual


yorhodes (approved):
love it
add a changeset?

PR #4810, opened 2024-11-04
fix: warp route checker

tkporter:
### Description

- Updates the .registryrc, which we will need to update when we add new warp route configs to infra, otherwise we'll get
```
Checking warp route WBTC/eclipsemainnet-ethereum...
Error checking warp route USDT/eclipsemainnet-ethereum-solanamainnet: Error: Warp route config for USDT/eclipsemainnet-ethereum-solanamainnet not found in registry
Error checking warp route WBTC/eclipsemainnet-ethereum: Error: Warp route config for WBTC/eclipsemainnet-ethereum not found in registry
Failed to check warp routes: USDT/eclipsemainnet-ethereum-solanamainnet, WBTC/eclipsemainnet-ethereum
```
- Uses the `Always` imagePullPolicy (https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy) to ensure we always use latest main
- A consequence of removing Lumia addresses from the registry in https://github.com/hyperlane-xyz/hyperlane-registry/pull/348 is that this broke checking of the Lumia warp route. I manually specify the required addresses here now.
- Also needed to explicitly specify the proxyAdmin ownership

Definitely some opportunities for improvement with the warp route deploy flow


typescript/infra/helm/check-warp-deploy/templates/cron-job.yaml
```diff
@@ -17,7 +17,7 @@ spec:
           containers:
           - name: check-warp-deploy
             image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
-            imagePullPolicy: IfNotPresent
+            imagePullPolicy: Always
```
Mo-Hussain, line 20: Not needed for this PR, but I wonder if it would be better to use the `latest`, I think it would always default `imagePullPolicy: Always` in that case. I think I tried latest and has issues

Mo-Hussain (approved):

typescript/infra/helm/check-warp-deploy/templates/cron-job.yaml
tkporter: I think the problem with latest is that images from branch CI will be available there

paulbalaji (approved):

PR #4811, opened 2024-11-04
feat: alert instead of crashing relayer when building cursors

daniel-savu:
### Description

- we should make sure there are no `unwrap`s outside a `tokio::spawn`, otherwise they'll take down the entire relayer. The offenders have so far been the indexing cursors, so this PR just logs an error and sets a new metric flag instead of panicking: `critical_error`, an `IntGauge` type because prometheus doesn't have boolean gauges. For now this is only used for origin chains, but I envision this as being set when a critical error occurs in the submitter too, in the future
- before setting the flag, cursor instantiation is now retried 10 times (to avoid short lived RPC hiccups)

### Drive-by changes
Removes usage of `Builder` derive macro in metric types, since they're not complex enough to warrant it

### Backward compatibility

Yes

### Testing

the existing e2e

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -419,10 +419,13 @@ impl Relayer {
     ) -> Instrumented<JoinHandle<()>> {
         let index_settings = self.as_ref().settings.chains[origin.name()].index_settings();
         let contract_sync = self.message_syncs.get(origin).unwrap().clone();
-        let cursor = contract_sync
-            .cursor(index_settings)
-            .await
-            .unwrap_or_else(|err| panic!("Error getting cursor for origin {origin}: {err}"));
+        let cursor = match contract_sync.cursor(index_settings).await {
+            Ok(cursor) => cursor,
+            Err(err) => {
+                error!(?err, origin=?origin, "Error building cursor for origin");
+                return tokio::spawn(async {}).instrument(info_span!("MessageSync"));
+            }
+        };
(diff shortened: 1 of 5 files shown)


rust/main/agents/relayer/src/relayer.rs
```diff
@@ -419,10 +419,13 @@ impl Relayer {
...
         let index_settings = self.as_ref().settings.chains[origin.name()].index_settings();
         let contract_sync = self.message_syncs.get(origin).unwrap().clone();
-        let cursor = contract_sync
-            .cursor(index_settings)
-            .await
-            .unwrap_or_else(|err| panic!("Error getting cursor for origin {origin}: {err}"));
+        let cursor = match contract_sync.cursor(index_settings).await {
+            Ok(cursor) => cursor,
+            Err(err) => {
+                error!(?err, origin=?origin, "Error building cursor for origin");
```
ameten, line 425: I wonder if we need a high urgency alert once such error is logged?

daniel-savu: added a new `critical_error` metric per chain that we'll alert on

ameten (approved):

tkporter (approved):
I imagine we'll wanna document this in the public docs at some point

daniel-savu:
> I imagine we'll wanna document this in the public docs at some point
@tkporter good point, I'll add a quick entry to the relayer monitoring & alerting guide (https://docs.hyperlane.xyz/docs/operate/relayer/monitoring-alerting)

PR #4812, opened 2024-11-04
Implement multiple message relay

yorhodes:

### DIFF
--- a/typescript/cli/src/status/message.ts
+++ b/typescript/cli/src/status/message.ts
@@ -2,12 +2,10 @@ import type { TransactionReceipt } from '@ethersproject/providers';
 import { input } from '@inquirer/prompts';
 
 import { ChainName, HyperlaneCore, HyperlaneRelayer } from '@hyperlane-xyz/sdk';
-import { assert, parseWarpRouteMessage } from '@hyperlane-xyz/utils';
 
 import { WriteCommandContext } from '../context/types.js';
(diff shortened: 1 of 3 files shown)


.changeset/lovely-planes-end.md
```diff
@@ -0,0 +1,6 @@
+---
+'@hyperlane-xyz/cli': minor
+'@hyperlane-xyz/sdk': minor
+---
+
+Allow self-relaying of all messages for a given dispatch transaction.
```
paulbalaji, line 6:
```suggestion
Allow self-relaying of all messages if there are multiple in a given dispatch transaction.
```

yorhodes:
would be cool to use this for https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4722

typescript/cli/src/commands/status.ts
```diff
@@ -19,12 +19,11 @@ export const statusCommand: CommandModuleWithWriteContext<
       description: 'Dispatch transaction hash',
     },
   },
-  handler: async ({ context, origin, destination, id, relay, dispatchTx }) => {
```
xeno097, line 22: Considering that the destination var is unused maybe we should remove it from the command options too so that users won't try to set it and then see unexpected results because the value is not used

xeno097 (approved):

PR #4813, opened 2024-11-05
feat: Scraper report mailbox as recipient of transactions for Sealevel

ameten:
### Description

Scraper reports mailbox as recipient of transactions for Sealevel

### Related issues

### Backward compatibility

Yes

### Testing

Manual test of Scraper
E2E tests for Ethereum and Sealevel

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/provider.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/provider.rs
@@ -24,19 +24,21 @@ pub struct SealevelProvider {
     domain: HyperlaneDomain,
     rpc_client: Arc<SealevelRpcClient>,
     native_token: NativeToken,
+    contract_address: H256,
 }
 
 impl SealevelProvider {
     /// Create a new Sealevel provider.
-    pub fn new(domain: HyperlaneDomain, conf: &ConnectionConf) -> Self {
+    pub fn new(domain: HyperlaneDomain, conf: &ConnectionConf, contract_address: H256) -> Self {
         // Set the `processed` commitment at rpc level
         let rpc_client = Arc::new(SealevelRpcClient::new(conf.url.to_string()));
         let native_token = conf.native_token.clone();
 
-        SealevelProvider {
+        Self {
             domain,
             rpc_client,
             native_token,
+            contract_address,
         }
     }
 
@@ -120,6 +122,7 @@ impl HyperlaneChain for SealevelProvider {
             domain: self.domain.clone(),
             rpc_client: self.rpc_client.clone(),
             native_token: self.native_token.clone(),
+            contract_address: self.contract_address,
         })
     }
 }
@@ -189,7 +192,7 @@ impl HyperlaneProvider for SealevelProvider {
             gas_price,
             nonce: 0,
             sender,
-            recipient: None,
+            recipient: Some(self.contract_address),
             receipt: Some(receipt),
             raw_input_data: None,
         })
--- a/rust/main/hyperlane-base/src/settings/chains.rs
+++ b/rust/main/hyperlane-base/src/settings/chains.rs
@@ -183,6 +183,7 @@ impl ChainConf {
             ChainConnectionConf::Sealevel(conf) => Ok(Box::new(h_sealevel::SealevelProvider::new(
                 locator.domain.clone(),
                 conf,
+                self.addresses.mailbox,
             )) as Box<dyn HyperlaneProvider>),
             ChainConnectionConf::Cosmos(conf) => {
                 let provider = CosmosProvider::new(
--- a/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
@@ -39,7 +39,11 @@ impl SealevelInterchainGasPaymaster {
         conf: &ConnectionConf,
         igp_account_locator: &ContractLocator<'_>,
     ) -> ChainResult<Self> {
-        let provider = SealevelProvider::new(igp_account_locator.domain.clone(), conf);
+        let provider = SealevelProvider::new(
+            igp_account_locator.domain.clone(),
+            conf,
+            igp_account_locator.address,
+        );
         let program_id =
             Self::determine_igp_program_id(provider.rpc(), &igp_account_locator.address).await?;
         let (data_pda_pubkey, _) =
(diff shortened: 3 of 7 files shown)


rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -189,7 +192,7 @@ impl HyperlaneProvider for SealevelProvider {
             gas_price,
             nonce: 0,
             sender,
-            recipient: None,
+            recipient: Some(self.contract_address),
```
tkporter, line 195: for my own understanding - bc there can be multiple instructions in Sealevel, you didn't want to grab the recipient of one of the instructions in tx (because it's not clear which instruction we care about), and instead would prefer to always have the Mailbox address?

what about txs that aren't made to the Mailbox, e.g. for message dispatches or gas payments?

Fwiw this field isn't really used so I don't think the correctness of it matters much, but I could see this being a bit confusing having every recipient be the Mailbox even when it's not the case

rust/main/hyperlane-base/src/settings/chains.rs
```diff
@@ -183,6 +183,7 @@ impl ChainConf {
             ChainConnectionConf::Sealevel(conf) => Ok(Box::new(h_sealevel::SealevelProvider::new(
                 locator.domain.clone(),
                 conf,
+                self.addresses.mailbox,
```
tkporter, line 186: worth documenting why we do this imo, otherwise this could be easily forgotten

rust/main/chains/hyperlane-sealevel/src/provider.rs
ameten: There can be multiple instructions in a Sealevel transaction on the highest level and they can be Native Programs or any other programs. To reliably identify a program which eventually invokes the Mailbox (so, the program is the proper recipient of the transaction), we need to simulate the execution of this program, I think.

Message dispatches are made by Mailbox program, as far as I can tell. Gas payments are done by a separate program, so, it is lost in this implementation.

We can discuss if there are easier ways to identify the recipient of the transaction.

rust/main/hyperlane-base/src/settings/chains.rs
ameten: I added comments.

rust/main/chains/hyperlane-sealevel/src/provider.rs
daniel-savu: How about checking if `transaction.message.account_keys.contains(self.contract_address)` to gain more confidence in setting `recipient` to `Some`, and otherwise set it to `None`? Because even if we use this provider method in a superfluous way, it could theoretically be used to fetch any tx whatsoever by hash

daniel-savu (approved):

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -24,19 +24,24 @@ pub struct SealevelProvider {
     domain: HyperlaneDomain,
     rpc_client: Arc<SealevelRpcClient>,
     native_token: NativeToken,
+    contract_address: H256,
```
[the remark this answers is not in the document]
daniel-savu: actually I think `tx_with_meta.meta.loaded_addresses.writable.contains(...)` would be even more robust

rust/main/chains/hyperlane-sealevel/src/provider.rs
daniel-savu: actually I think `tx_with_meta.meta.loaded_addresses.writable.contains(...)` would be even more robust

ameten: Discussed in person. We extract program from the top-level instructions.

rust/main/chains/hyperlane-sealevel/src/provider.rs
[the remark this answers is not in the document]
ameten: Discussed in person. We extract program from the top-level instructions.

rust/main/chains/hyperlane-sealevel/src/error.rs
```diff
@@ -42,6 +42,12 @@ pub enum HyperlaneSealevelError {
     /// Empty compute units consumed
     #[error("received empty compute units consumed in transaction")]
     EmptyComputeUnitsConsumed,
+    /// Too many non-native programs
+    #[error("transaction contains too many non-native programs, hash: {0:?}")]
+    TooManyNonNativePrograms(H512),
+    /// No non-native programs
+    #[error("transaction contains no non-native programs, hash: {0:?}")]
+    NoNativePrograms(H512),
```
tkporter, line 50: should this be `NoNonNativePrograms`?

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -80,6 +91,39 @@ impl SealevelProvider {
...
+            .filter(|program_id| !NATIVE_PROGRAMS.contains(*program_id))
+            .collect::<Vec<&String>>();
+
+        if programs.len() > 1 {
+            Err(HyperlaneSealevelError::TooManyNonNativePrograms(*hash))?;
+        }
+
+        let program_id = programs
+            .first()
+            .ok_or(HyperlaneSealevelError::NoNativePrograms(*hash))?;
```
tkporter, line 120: any thoughts on how likely this is to occur? if you think there's a possibility, wondering if we should just put a dummy value in this case or something

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -80,6 +91,39 @@ impl SealevelProvider {
...
+            })
+            .map(|ii| match ii {
+                UiParsedInstruction::Parsed(iii) => &iii.program_id,
+                UiParsedInstruction::PartiallyDecoded(iii) => &iii.program_id,
+            })
+            .filter(|program_id| !NATIVE_PROGRAMS.contains(*program_id))
+            .collect::<Vec<&String>>();
+
+        if programs.len() > 1 {
+            Err(HyperlaneSealevelError::TooManyNonNativePrograms(*hash))?;
```
tkporter, line 115: should we err here? I think it's decently likely that there could be multiple - I think it'd make sense to just grab the first one instead of erroring?

tkporter (approved):

rust/main/chains/hyperlane-sealevel/src/provider.rs
ameten: We have considered it during adding Cosmos and settled on erroring. Getting an error here should give us a better idea how we should handle it.

rust/main/chains/hyperlane-sealevel/src/error.rs
ameten: Great catch! Fixed.

PR #4820, opened 2024-11-05
feat: ica router management in `core apply`

### LINKED ISSUE #4819: (CLI migration) ICA enrollment/unenrollment in `core apply`

xeno097:
### Description

This PR adds the required logic to enroll/unenroll ICAs between them when running `core apply`

### Drive-by changes

- Implements the `EvmIcaReader`
- Updates the `EvmCoreModule` to read the current chain ICA config
-  Adds tests for the `EvmIcaRouterReader`

### Related issues

- Fixes #4819 

### Backward compatibility

- Yes

### Testing

- Manual
- E2E

### Note

- merge #4767 first


typescript/sdk/src/core/EvmIcaModule.ts
```diff
@@ -15,33 +29,182 @@ import {
...
+        expectedConfig[domainId].interchainSecurityModule
+          ? addressToBytes32(expectedConfig[domainId].interchainSecurityModule!)
+          : ethers.utils.hexZeroPad('0x', 32),
+      );
+    }
+
+    const remoteTransactions: AnnotatedEV5Transaction[] = domainsToEnroll.map(
+      (domainId) => ({
+        annotation: `Enrolling InterchainAccountRouter on domain ${this.domainId} on InterchainAccountRouter at ${expectedConfig[domainId].address} on domain ${domainId}`,
+        chainId: Number(this.multiProvider.getChainId(domainId)),
```
ltyu, line 139: think this is no longer true due to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798/files

typescript/sdk/src/core/EvmIcaModule.ts
```diff
@@ -15,33 +29,182 @@ import {
...
+      data: InterchainAccountRouter__factory.createInterface().encodeFunctionData(
+        'unenrollRemoteRouters(uint32[])',
+        [routesToUnenroll],
+      ),
+    });
+
+    const remoteTransactions: AnnotatedEV5Transaction[] = routesToUnenroll.map(
+      (domainId) => ({
+        annotation: `Removing InterchainAccountRouter on domain ${this.domainId} from InterchainAccountRouter at ${actualConfig[domainId].address} on domain ${domainId}`,
+        chainId: Number(this.multiProvider.getChainId(domainId)),
```
ltyu, line 196: ditto

typescript/sdk/src/core/EvmIcaModule.ts
xeno097: Good catch! Should probably use the new `getEvmChainId` method here

typescript/sdk/src/core/EvmIcaModule.ts
```diff
@@ -15,33 +29,182 @@ import {
...
+        data: InterchainAccountRouter__factory.createInterface().encodeFunctionData(
+          'enrollRemoteRouter(uint32,bytes32)',
+          [
+            this.domainId,
+            addressToBytes32(this.args.addresses.interchainAccountRouter),
+          ],
+        ),
+      }),
+    );
+
```
paulbalaji, line 150: imo these remote transactions in the enroll/unenroll methods should move to their own functions so as not to clutter this method. Generally prefer to have this class only really deal with creating transactions for itself

imagine we're deploying to N chains and call core deploy for each in parallel - if I understand it right here we're going to end up with multiple txs for enrolling the same router?

typescript/sdk/src/core/EvmIcaModule.ts
```diff
@@ -15,33 +29,182 @@ import {
...
+        data: InterchainAccountRouter__factory.createInterface().encodeFunctionData(
+          'unenrollRemoteRouter(uint32)',
+          [this.domainId],
+        ),
+      }),
+    );
+
+    transactions.push(...remoteTransactions);
+
+    return transactions;
```
paulbalaji, line 207: same comment as above

typescript/cli/src/deploy/core.ts
```diff
@@ -127,7 +127,11 @@ export async function runCoreApply(params: ApplyParams) {
   if (transactions.length) {
     logGray('Updating deployed core contracts');
     for (const transaction of transactions) {
-      await multiProvider.sendTransaction(chain, transaction);
+      await multiProvider.sendTransaction(
+        // Using the provided chain id because there might be remote chain transactions included in the batch
+        transaction.chainId ?? chain,
+        transaction,
```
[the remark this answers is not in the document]
xeno097: We use the provided chain id in the transaction because using the `chain` variable by default assumed that all txs were done on the same chain which is not the case anymore now that we also send txs on remote chains to register the current router.

typescript/cli/src/read/core.ts
```diff
@@ -23,9 +25,18 @@ export async function executeCoreRead({
...
   }
 
+  if (!interchainAccountRouter) {
+    interchainAccountRouter = addresses?.interchainAccountRouter;
+
+    assert(
+      interchainAccountRouter,
+      `${chain} interchainAccountRouter not provided and none found in registry.`,
+    );
+  }
```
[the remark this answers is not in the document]
xeno097: Thanks I have to update this then

typescript/cli/src/tests/core.e2e-test.ts
```diff
@@ -18,28 +19,38 @@ import {
...
 const EXAMPLES_PATH = './examples';
 const CORE_CONFIG_PATH = `${EXAMPLES_PATH}/core-config.yaml`;
 
 const TEMP_PATH = '/tmp'; // /temp gets removed at the end of all-test.sh
-const CORE_READ_CONFIG_PATH = `${TEMP_PATH}/anvil2/core-config-read.yaml`;
+const CORE_READ_CHAIN_2_CONFIG_PATH = `${TEMP_PATH}/${CHAIN_NAME_2}/core-config-read.yaml`;
+const CORE_READ_CHAIN_3_CONFIG_PATH = `${TEMP_PATH}/${CHAIN_NAME_3}/core-config-read.yaml`;
 
-const TEST_TIMEOUT = 100_000; // Long timeout since these tests can take a while
+const TEST_TIMEOUT = 200_000; // Long timeout since these tests can take a while
```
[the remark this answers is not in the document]
xeno097: I agree

typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -116,7 +143,7 @@ export class EvmCoreModule extends HyperlaneModule<
    * @returns Transaction that need to be executed to update the ISM configuration.
    */
   async createDefaultIsmUpdateTxs(
-    actualConfig: CoreConfig,
+    actualConfig: DerivedCoreConfig,
```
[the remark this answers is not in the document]
xeno097: The main difference is to make fields that we know are defined once the contracts are deployed on-chain also required on the type instead of having to check each time for if the field is undefined or not because we were using the "old" config schema where almost all the fields are optional

typescript/sdk/src/core/EvmIcaModule.ts
xeno097: yep we are going to end with at least one tx for each chain where we need to register the new router + the tx on the current chain to register the remote routers with the current one. 

When you say
>  imo these remote transactions in the enroll/unenroll methods should move to their own functions so as not to clutter this method. Generally prefer to have this class only really deal with creating transactions for itself

You prefer that this class only generates txs for the current configured chain?

typescript/cli/src/deploy/core.ts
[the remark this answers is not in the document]
paulbalaji: but then inside the multiprovider it would get the chain for that chainid, thinking it is the domainid

which affects chains such as rootstock/immutable/flow where we have new domainids that are different to the chainids. I think so far it hasn't broken anything because it's basically the same chain metadata with different domain IDs - but not to say it doesn't break in the future 

also when we get the update txs here, we're assuming that the current signer is the owner - which might not be the case for SAFE/ICA owned chains. I think the update txs should go through a tx submitter instead of the raw multiprovider?

typescript/sdk/src/core/EvmIcaModule.ts
paulbalaji: > You prefer that this class only generates txs for the current configured chain?
Yup, iirc this is how the warp module works atm too?

typescript/cli/src/deploy/core.ts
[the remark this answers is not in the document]
xeno097: > which affects chains such as rootstock/immutable/flow where we have new domainids that are different to the chainids. I think so far it hasn't broken anything because it's basically the same chain metadata with different domain IDs - but not to say it doesn't break in the future

Ok so probably the chain id field on the annotated tx should be updated to `domainId` or something similar so to remove the confusion there too no?  

> also when we get the update txs here, we're assuming that the current signer is the owner - which might not be the case for SAFE/ICA owned chains. I think the update txs should go through a tx submitter instead of the raw multiprovider?

yep at the moment support for the submitter is only implemented on `warp apply` but will be added to `core apply` too. I can do that in a follow up pr when I add ICA support to `warp apply` after #4930 is merged

typescript/sdk/src/core/EvmIcaModule.ts
xeno097: > Yup, iirc this is how the warp module works atm too?

Yup, but there we have all the related chains defined in the same config file so for each warp token on a chain a new warp module is created to send the update txs. Here we can't do the same because each core definition belongs to one chain. And probably extracting those remote txs to another function might be error prone when someone wants to reuse the core update logic but the remote ica enrollment/unenrollment is somewhere else and might be forgotten

typescript/cli/src/deploy/core.ts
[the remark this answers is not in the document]
paulbalaji: > Ok so probably the chain id field on the annotated tx should be updated to domainId or something similar so to remove the confusion there too no?

this is why I wanted `chain` on that type back in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798 but then the diff would be quite big so we opted to minimise the diff

typescript/sdk/src/core/EvmIcaModule.ts
paulbalaji: I guess if there's absolutely no way else to do it then I will begrudgingly accept this if we do have a path to make this per-chain again in the future

xeno097: What we might do in the future is to maybe group txs per chain outside of the module, as we do for [`warp apply` txs](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/deploy/warp.ts#L485)

typescript/sdk/src/ica/schemas.ts
```diff
@@ -0,0 +1,43 @@
+import { z } from 'zod';
+
+import { ZHash } from '../metadata/customZodTypes.js';
+import { DerivedOwnableSchema } from '../schemas.js';
+
+export const RemoteIcaRouterConfigSchema = z.record(
```
ltyu, line 6: Is this the same as https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/455a897fb50f88812ca3842ac0adb41cce817eb4/typescript/sdk/src/router/types.ts#L80-L83

Can we reuse?

typescript/cli/src/read/core.ts
```diff
@@ -23,9 +25,13 @@ export async function executeCoreRead({
     );
   }
 
+  if (!interchainAccountRouter) {
```
ltyu, line 28: for visibility, we discussed removing this and just rely on the chain addresses. in the future, we will have an ica command to update other icas

typescript/sdk/src/core/EvmIcaModule.ts
ltyu: agree with paul. the modules should handle the configs in a per chain basis. 

happy to approve this PR as is but lets keep that convention

typescript/sdk/src/ica/schemas.ts
xeno097: solved in dms

typescript/sdk/src/core/EvmIcaModule.ts
xeno097: yep i agree we can look for a better way to do this later

ltyu (approved):
🔥

PR #4822, opened 2024-11-05
fix: add collateralChainName to warp read

ltyu:
### Description
Fixes "warp read does not return collateralChainName for rebase vault type"  and "warp check fails when checking the config because collateralChainName is missing" of CLI 5.6 Bug Bash

### Backward compatibility
Yes

### Testing
Manual/Unit Tests

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -230,7 +230,7 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
       if (chainMetadata.nativeToken) {
         const { name, symbol, decimals } = chainMetadata.nativeToken;
         return {
-          type: TokenType.native,
+          type,
           name,
           symbol,
           decimals,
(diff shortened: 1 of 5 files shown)


xeno097 (approved):
Good job! Maybe add tests for this specific warp config just to be sure that in the future we don't make a change that introduces the bug again

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -65,7 +65,7 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
     // Derive the config type
     const type = await this.deriveTokenType(warpRouteAddress);
     const baseMetadata = await this.fetchMailboxClientConfig(warpRouteAddress);
-    const tokenMetadata = await this.fetchTokenMetadata(type, warpRouteAddress);
+    const tokenMetadata = await this.fetchTokenConfig(type, warpRouteAddress);
```
paulbalaji, line 68: rename var to `tokenConfig` too?

paulbalaji (commented):
need changeset too

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: done

paulbalaji (approved):

PR #4824, opened 2024-11-05
feat: better chain selection for single chain

### LINKED ISSUE #4823: Better chain selection for single chain selection

xeno097:
### Description

This PR updates the single chain selection prompt to use a search prompt instead of a select one.

### Drive-by changes

- Adds missing `:` where the `runSingleChainSelectionStep` is used.
- Installs the `@inquirer/search` package

### Related issues

- Fixes #4823 

### Backward compatibility

- Yes

### Testing

- Manual


typescript/cli/package.json
```diff
@@ -11,6 +11,7 @@
     "@inquirer/core": "9.0.10",
     "@inquirer/figures": "1.0.5",
     "@inquirer/prompts": "^3.0.0",
+    "@inquirer/search": "^3.0.1",
```
paulbalaji, line 14: not for this PR but could you create a separate issue to investigate and bump up the inquirer packages across the board to more recent versions? 

i ask because prompts/figures don't look upto date, and v3 of search seems to be wanting v10+ of core (we only specify v9 here)

xeno097: will do sir! thanks for pointing it out

ltyu (approved):
LTGTM

typescript/cli/src/config/hooks.ts
yorhodes, line 1: alternatively you could append/assert the colon in the `runSingleChainSelectionStep` function

xeno097: great suggestion 🔥

PR #4826, opened 2024-11-06
feat: add weETHs warp route, some drive-by cleanup

tkporter:
### Description

- Deploys weETHs between Ethereum and Eclipsemainnet
- Removes all the warp route generation scripts in favor of a single one
- Renames some things throughout infra to be closer to the canonical warp IDs (e.g. add all chain names, sort them, etc)
- Removes a bunch of now unnecessary ts-ignores
- Moves toward using router configs for non-EVM chains in the warp configs

Waiting to merge https://github.com/hyperlane-xyz/hyperlane-registry/pull/362 before updating the registryrc in here so the warp config checker works


rust/sealevel/client/src/warp_route.rs
```diff
@@ -192,19 +192,25 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
         app_config: &TokenConfig,
         program_id: Pubkey,
     ) {
-        if let Some(ata_payer_funding_amount) = self.ata_payer_funding_amount {
-            if matches!(
-                app_config.token_type,
-                TokenType::Collateral(_) | TokenType::Synthetic(_)
-            ) {
-                fund_ata_payer_up_to(ctx, client, program_id, ata_payer_funding_amount);
+        let try_fund_ata_payer = |ctx: &mut Context, client: &RpcClient| {
```
tkporter, line 195: Moved this because otherwise it would try to fund the ATA payer before it exists via initialization, which is fine if the amount is high enough, but I ran with a low ata payment amount which was insufficient to cover rent

daniel-savu (approved):

PR #4827, opened 2024-11-06
feat: deploy to arcadiatestnet, alephzeroevmtestnet, inksepolia

paulbalaji:
### Description

feat: deploy to arcadiatestnet, alephzeroevmtestnet, inksepolia

### Drive-by changes

- update agent configs

### Testing

manual


typescript/infra/config/environments/testnet4/chains.ts
```diff
@@ -18,6 +18,11 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
       gasPrice: 8 * 10 ** 9, // 8 gwei
     },
   },
+  scrollsepolia: {
+    transactionOverrides: {
+      gasPrice: 0.5 * 10 ** 9, // 0.5 gwei
```
tkporter, line 23: keep or no? wondering if we want this in the agents or not, coming back to the convo in https://www.notion.so/hyperlanexyz/Fixing-Relayer-Gas-Price-Issues-1356d35200d6807baeadf10fd0451d90?d=1356d35200d68012b1f2001ce65e3c71#1356d35200d6804298a1f42603071de2

tkporter (approved):

typescript/infra/config/environments/testnet4/chains.ts
paulbalaji: hm I'll leave it as a deploy-only override for now

PR #4828, opened 2024-11-06
chore: fix path

drewstone:
### Description
Fix the path in StaticAggregationIsm.sol. I can't copy the test files into my own solidity project due to this path issue.

Pls issue new npm package if this is merged 🫶


solidity/contracts/isms/aggregation/StaticAggregationIsm.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity >=0.8.0;
 import {AbstractAggregationIsm} from "./AbstractAggregationIsm.sol";
 import {AggregationIsmMetadata} from "../../isms/libs/AggregationIsmMetadata.sol";
 import {MetaProxy} from "../../libs/MetaProxy.sol";
-import {PackageVersioned} from "contracts/PackageVersioned.sol";
+import {PackageVersioned} from "../../contracts/PackageVersioned.sol";
```
aroralanuk, line 8:
```suggestion
import {PackageVersioned} from "../../PackageVersioned.sol";
```

aroralanuk (commented):
Good catch, can you also add this fix to other contracts like `AbstractMessageIdAuthorizedIsm`?

PR #4831, opened 2024-11-07
feat: persistent ts relayer

yorhodes:
### Description

- Adjust long running TS relayer with retry queue instead of crashing.
- Adds a "whitelist" for relaying to specific message senders/recipients.
- Adds a "symbol" flag borrowed from warp commands for filtering on a specific warp route.

### Drive-by changes

None

### Related issues

- Enables warp route deployer to run the CLI relayer in the background and test/share the warp UI.

### Backward compatibility

Yes

### Testing

CLI e2e tests


typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -206,44 +276,92 @@ export class HyperlaneRelayer {
...
+        lastAttempt: Date.now(),
+        message: message.message,
+        dispatchTx: event.transactionHash,
+      });
+    }, this.whitelistChains());
 
     this.stopRelayingHandler = removeHandler;
+
+    // start flushing backlog
+    void this.flushBacklog();
```
yorhodes, line 355: this might be an antipattern

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -31,19 +36,64 @@ const DerivedHookConfigWithAddressSchema =
...
 
+type MessageWhitelist = ChainMap<Set<Address>>;
+
+// message must have origin and destination chains in the whitelist
+// if whitelist has non-empty address set for chain, message must have sender and recipient in the set
+export function messageMatchesWhitelist(
+  whitelist: MessageWhitelist,
+  message: ParsedMessage,
+): boolean {
+  const originAddresses = whitelist[message.originChain ?? message.origin];
```
ltyu, line 64: iiuc, this is the chainId, and they are no longer unique. 

if there are chains with the same originId, but only 1 is whitelisted, does this mean they are all whitelisted?

ltyu (commented):
love this upgrade!

typescript/cli/src/commands/relayer.ts
```diff
@@ -1,35 +1,82 @@
...
 
-import { agentTargetsCommandOption } from './options.js';
+import {
+  agentTargetsCommandOption,
+  overrideRegistryUriCommandOption,
+  symbolCommandOption,
+} from './options.js';
 import { MessageOptionsArgTypes } from './send.js';
 
+const DEFAULT_RELAYER_CACHE = `${overrideRegistryUriCommandOption.default}/relayer-cache.json`;
```
jmrossy, line 20: Could this maybe lead to someone accidentally committing this cache file (in certain cases)?

typescript/sdk/src/core/HyperlaneRelayer.ts
```diff
@@ -206,44 +276,92 @@ export class HyperlaneRelayer {
     );
   }
 
-  start(chains = this.core.chains()): void {
-    assert(!this.stopRelayingHandler, 'Relayer already started');
+  protected async flushBacklog(): Promise<void> {
+    while (this.stopRelayingHandler) {
```
jmrossy, line 280: This infinite loop approach is okay when the only consumer of this is the CLI but if were to ever use this in UI, we would want a design that relays when consumers call in. Otherwise, this could block the UI thread. In general, this loop-and-watch behavior _feels_ like it belongs in the CLI IMO, whereas an SDK class should be more flexible as to how it's used

typescript/sdk/src/core/HyperlaneRelayer.ts
jmrossy: it sure is! Did you do this to swallow the promise error? At least `.catch` it in that case. But more generally speaking, I think the design of this class should be different (see above)

jmrossy (commented):
A very useful addition but I have some concerns about the Relayer class design. I'm further away from the CLI though so I'll leave it to you and @ltyu to make any final decisions

typescript/cli/src/commands/relayer.ts
yorhodes: definitely

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: this isnt for swallowing the error, its for marking the promise explicitly as hanging (this is what the linter suggests btw)

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: > we would want a design that relays when consumers call in

unless I misunderstand what you mean by this, the existing `relayMessage` function does what you are describing

>  In general, this loop-and-watch behavior feels like it belongs in the CLI IMO

interesting, im not really sure how that would look
it seems like it would require a lot of "internal" knowledge for clients to implement the behavior you are describing

I quite like the current interface of
```js
relayer.start()
process.once('SIGINT', () => relayer.stop())
```

typescript/infra/scripts/relay.ts
```diff
@@ -10,11 +10,10 @@ const CACHE_PATH = process.env.RELAYER_CACHE ?? './relayer-cache.json';
 async function main() {
   const { environment } = await getArgs().argv;
   const { core } = await getHyperlaneCore(environment);
-  const relayer = new HyperlaneRelayer({ core });
 
-  // target subset of chains
-  // const chains = ['ethereum', 'polygon', 'bsc']
-  const chains = undefined;
+  // target subset of chains and senders/recipients
+  const whitelist = undefined;
```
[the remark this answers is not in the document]
yorhodes: yeah this is just meant for documentation purposes so you can use from infra

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: this is the domain ID as encoded by the mailbox onchain

xeno097 (approved):

typescript/sdk/src/core/HyperlaneRelayer.ts
yorhodes: if you were referring to the `originChain` thats the human readable identifier which should be unique

typescript/cli/src/tests/relay.e2e-test.ts
```diff
@@ -0,0 +1,70 @@
...
+const SYMBOL = 'ETH';
+
+const WARP_DEPLOY_OUTPUT = `${REGISTRY_PATH}/deployments/warp_routes/${SYMBOL}/${CHAIN_NAME_1}-${CHAIN_NAME_2}-config.yaml`;
+
+const EXAMPLES_PATH = './examples';
+const CORE_CONFIG_PATH = `${EXAMPLES_PATH}/core-config.yaml`;
+const WARP_CONFIG_PATH = `${EXAMPLES_PATH}/warp-route-deployment.yaml`;
+
+const TEST_TIMEOUT = 100_000; // Long timeout since these tests can take a while
+describe.only('hyperlane relayer e2e tests', async function () {
```
ltyu: can remove .only

ltyu (approved):

PR #4835, opened 2024-11-07
feat: refactor warp monitor to use WarpCore config

tkporter:
### Description

- Removes the warp monitor specific config format and all the handcrafted yamls from infra
- Massive refactor to warp monitor to accommodate this -- tried to make it a bit easier to work work by splitting it into a few files. The changes are generally backward compatible with some small exceptions, like now the metrics use the `TokenStandard` instead of `TokenType` (the former is present in the warp core config, the latter isn't, and there's no way atm to map from the former to the latter)
- The warp monitor now takes in a warp ID and gets its config from the registry (recall this is the registry version defined in `.registryrc` if running in the monorepo image)
- Deploying the warp monitor now lets you do many at a time and select the ones to deploy interactively. This makes it a bit easier to work with the warp IDs
- The names of the warp deploys are based on the warp IDs. Because warp IDs can change as new chains are added, this may cause issues in the future - we'll just need to be cautious of removing the old release to avoid accidentally having 2 of them. For now I think this is fine. The helm release names are also capped at 53 chars so they needed to be truncated
- Some changes to the SDK to get the necessary data here. Happy to pull some of these into the warp monitor if people feel they're inappropriate for the SDK (like the `getBridgedSupply`)
- Added a test to make sure any warp route IDs in infra are found in the registry

- Corresponding PR to the registry here https://github.com/hyperlane-xyz/hyperlane-registry/pull/370


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -24,20 +24,20 @@ import {
...
-import inevmEthereumUsdcAddresses from './warp/inevm-USDC-addresses.json';
-import inevmEthereumUsdtAddresses from './warp/inevm-USDT-addresses.json';
-import injectiveInevmInjAddresses from './warp/injective-inevm-addresses.json';
-import mantaTIAAddresses from './warp/manta-TIA-addresses.json';
-import merklyEthAddresses from './warp/merkly-eth-addresses.json';
-import renzoEzEthAddressesV3 from './warp/renzo-ezETH-addresses-v3.json';
-import victionEthereumEthAddresses from './warp/viction-ETH-addresses.json';
-import victionEthereumUsdcAddresses from './warp/viction-USDC-addresses.json';
-import victionEthereumUsdtAddresses from './warp/viction-USDT-addresses.json';
+import ancient8EthereumUsdcAddresses from './warp/artifacts/ancient8-USDC-addresses.json';
```
tkporter, line 27: coming for these next 🔪

typescript/infra/helm/warp-routes/templates/env-var-external-secret.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       data:
         GCP_SECRET_OVERRIDES_ENABLED: "true"
         GCP_SECRET_OVERRIDE_{{ $.Values.hyperlane.runEnv | upper }}_COINGECKO_API_KEY: {{ printf "'{{ .%s_coingecko_api_key | toString }}'" .Values.hyperlane.runEnv }}
+        GCP_SECRET_OVERRIDE_HYPERLANE_{{ .Values.hyperlane.runEnv | upper }}_KEY_DEPLOYER: {{ print "'{{ .deployer_key | toString }}'" }}
```
tkporter, line 25: required because behind the scenes some of the getters in the monitor try to fetch this, despite it not being used. Not ideal, but not concerned imo

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
```diff
@@ -0,0 +1,239 @@
...
+      .describe('checkFrequency', 'frequency to check balances in ms')
+      .demandOption('checkFrequency')
+      .alias('v', 'checkFrequency') // v as in Greek letter nu
+      .number('checkFrequency')
+      .parse();
+
+  startMetricsServer(metricsRegister);
+
+  const envConfig = getEnvironmentConfig(environment);
+  const registry = await envConfig.getRegistry();
```
Mo-Hussain, line 47: I think it could be useful to have an interactive step that warns us if the .registry commit is behind main on registry, we could enable ignoring the warning a proceeding anyway in special cases where want to use an old version of the registry.

I think it would be quite easy to forget to update the commit

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
```diff
@@ -0,0 +1,239 @@
...
+  const chainMetadata = await registry.getMetadata();
+
+  // The Sealevel warp adapters require the Mailbox address, so we
+  // get router configs (that include the Mailbox address) for all chains
+  // and merge them with the chain metadata.
+  const routerConfig = await getRouterConfigsForAllVms(
```
Mo-Hussain, line 53: Nice!

typescript/infra/test/warpIds.test.ts
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'chai';
+
+import { WarpRouteIds } from '../config/environments/mainnet3/warp/warpIds.js';
+import { getRegistry } from '../config/registry.js';
+
+describe('Warp IDs', () => {
```
Mo-Hussain, line 6: Nice!

typescript/infra/config/environments/mainnet3/agent.ts
Mo-Hussain: As in read from registry instead?

Mo-Hussain:
lgtm on the infra front, will defer on SDK owners on inlcuding `getBridgedSupply` in the SDK, the adapter changes look good to however

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
tkporter: Because registry commits are so frequent I wonder if there's another heuristic we can go off of? Otherwise I'd expect this to just always complain

We could maybe have the deploy command make sure that the commit in the registryrc contains all the warp routes we want?

Mo-Hussain: The only downside to this is that the it may not pick up changes to configs, which would be very rare. If did happen that one time when we redeployed those eclipse routes. This may not be too much of a concern  

I think we would throw an error if we cant find the warp route id in registry anyway no?

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
```diff
@@ -0,0 +1,239 @@
...
+    envConfig,
+    await envConfig.getMultiProvider(),
+  );
+  const multiProtocolProvider = new MultiProtocolProvider(
+    objMerge(chainMetadata, routerConfig),
```
jmrossy, line 58: I wouldn't expect a router config to get merged into a chain metadata object. Maybe routerConfig is an imperfect name?

typescript/infra/scripts/warp-routes/monitor/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { rootLogger } from '@hyperlane-xyz/utils';
+
+export const logger = rootLogger.child({ module: 'warp-balance-monitor' });
+
+export async function gracefullyHandleError(
```
jmrossy, line 5: This seems like an odd pattern to me. IMO `gracefullyHandleError(...)` isn't more clear than a try/catch around the call. personally, I would opt for making the calls themselves swallow the error if that's the desired behavior. But if you do want a general 'catcher' util, it could be moved up and called just `try()`

typescript/sdk/src/token/Token.ts
```diff
@@ -353,10 +397,18 @@ export class Token implements IToken {
     return Object.values(PROTOCOL_TO_NATIVE_STANDARD).includes(this.standard);
   }
 
+  isCollateralized(): boolean {
```
jmrossy, line 400: 👍

typescript/sdk/src/token/Token.ts
```diff
@@ -341,6 +342,49 @@ export class Token implements IToken {
     return new TokenAmount(balance, this);
   }
 
+  /**
+   * Gets the amount of tokens bridged, if possible.
+   * @param multiProvider A multiProvider that contains the chain for this token.
+   * @returns The amount of tokens bridged, or undefined if not supported.
+   */
+  async getBridgedSupply(
```
jmrossy, line 350: This is fine for now but IMO this _feels_ like something that we'd add to `IHypTokenAdapter` and then implement in the adapters, rather than a big function here branching based on standard. That would also remove the need for casting.

jmrossy (commented):
This is great 🔥

typescript/infra/scripts/warp-routes/monitor/utils.ts
tkporter: yeah agree it's a little awkward - the main motivation was wanting to do some things concurrently in `updateTokenMetrics` and not wanting a lot of nested try/catches. I'll rename to `tryFn` for now

typescript/sdk/src/token/Token.ts
tkporter: agreed - done!

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
tkporter: will isolate the mailboxes to make it more clear

typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
tkporter: > I think we would throw an error if we cant find the warp route id in registry anyway no?

it would throw but wouldn't exit - fixed now!
I think this should help. Kinda inclined to kick the can on getting better registry integrations in this case, wdyt? We'd get alerted on this crashing and someone will fix it if there's an issue

jmrossy (approved):
LGTM 👍

typescript/infra/helm/warp-routes/templates/_helpers.tpl
```diff
@@ -72,11 +72,11 @@ The warp-routes container
      value: pretty
   command:
   - ./node_modules/.bin/tsx
-  - ./typescript/infra/scripts/warp-routes/monitor-warp-routes-balances.ts
+  - ./typescript/infra/scripts/warp-routes/monitor/monitor-warp-routes-balances.ts
```
paulbalaji, line 75: this is the most minor of nits ever but should the script instead be 
```diff
-monitor-warp-routes-balances.ts
+monitor-warp-route-balances.ts
```

paulbalaji (approved):
lgtm overall

typescript/infra/helm/warp-routes/templates/_helpers.tpl
tkporter: lol yeah ill rename

PR #4837, opened 2024-11-07
fix: Remove coingecko-api-v3 library and de-dupe fetching utils

jmrossy:
### Description

- Remove coingecko-api-v3 lib which is unnecessary and breaks bundling
- Remove `getCoingeckoTokenPrices` function which is redundant with `CoinGeckoTokenPriceGetter`
- Simplify `CoinGeckoTokenPriceGetter` and remove need for Mock class

Related: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4787

See discussion here: https://discord.com/channels/935678348330434570/1304125818817220653

### Drive-by changes

- Fix bundling issue with Storybook in widgets lib

### Backward compatibility

No

### Testing

Rewrote unit tests and tested Storybook


typescript/sdk/src/gas/token-prices.ts
```diff
@@ -153,19 +146,35 @@ export class CoinGeckoTokenPriceGetter implements TokenPriceGetter {
...
+
+    const resp = await fetch(url);
+    const idPrices = await resp.json();
+
+    return ids.map((id) => {
+      const price = idPrices[id]?.[currency];
+      if (!price) throw new Error(`No price found for ${id}`);
+      return Number(price);
+    });
+  }
```
jmrossy, line 179: This is all that was needed to replace what the lib was doing!

Mo-Hussain (approved):

paulbalaji (approved):

PR #4841, opened 2024-11-08
feat: nov 8 enrollments

paulbalaji:
### Description

- enroll new validators for alephzeroevmmainnet, chilizmainnet, flowmainnet, immutablezkevmmainnet, metal, polynomialfi, rarichain, rootstockmainnet, superpositionmainnet, flame, prom
- update validators for inevm
- update ownership for recent two deploy batches
	- ICAs re-enabled
	- zeronetwork safe configured

### Drive-by changes

- igp updates
- fix some infra helper scripts
	- skip all chains to skip for ICA
	- only skip lumia for validator announce check

### Testing

manual + infra helper scripts


typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -130,18 +131,32 @@ export const icas: Partial<
   // ----------------------------------------------------------
   // lumia: '0x418E10Ac9e0b84022d0636228d05bc74172e0e41',
```
tkporter, line 132: does having this commented out result in us trying to transfer ownership to the deployer key?

typescript/infra/scripts/check/check-validator-announce.ts
```diff
@@ -25,7 +25,7 @@ async function main() {
 
   const targetNetworks = (
     chains && chains.length > 0 ? chains : config.supportedChainNames
-  ).filter(isEthereumProtocolChain);
+  ).filter((chain) => isEthereumProtocolChain(chain) && chain !== 'lumia');
```
tkporter, line 28: why not lumia?

typescript/infra/scripts/check/check-owner-ica.ts
```diff
@@ -39,7 +33,9 @@ async function main() {
 
   const checkOwnerIcaChains = (
     chains?.length ? chains : Object.keys(icas)
-  ).filter(isEthereumProtocolChain);
+  ).filter(
+    (chain) => isEthereumProtocolChain(chain) && !chainsToSkip.includes(chain),
```
tkporter, line 37: what's the purpose of `chainsToSkip`? iiuc we want to skip the chains for some things and not for other things, a comment of how it's intended to be used where it's defined would be nice. Or maybe to have it in the name

typescript/infra/config/environments/mainnet3/gasPrices.json
```diff
@@ -96,23 +96,23 @@
     "decimals": 9
   },
   "ethereum": {
-    "amount": "13.154482028",
+    "amount": "10.988801871",
```
tkporter, line 99: wonder if we should put this up to like 15, there have been some spikes recently?

tkporter (approved):

typescript/infra/scripts/check/check-validator-announce.ts
paulbalaji: I was thinking of working around and adding the lumia addresses to the metadata manually so the script still works, but thought it easier to skip lumia instead

can update the script if we want

typescript/infra/config/environments/mainnet3/gasPrices.json
paulbalaji: yeah that sgtm

typescript/infra/scripts/check/check-owner-ica.ts
paulbalaji: added a comment in chain.ts it's defined

typescript/infra/scripts/check/check-validator-announce.ts
paulbalaji: decided to just leave a comment for why not lumia

PR #4845, opened 2024-11-11
feat: relayer app contexts use warp route artifacts from the registry

tkporter:
### Description

This PR aims to solve 2 problems: forgetting to add warp routes as app contexts in the relayer, and deduplicating artifacts that live in infra and the registry

- Removes a bunch of the duplicated warp route artifacts in infra, in favor of getting them from the registry
- Now all WarpRouteIds defined in infra will be auto-populated into the mainnet3 relayers as app contexts, with the label as the warp route ID
- This significantly increased the size of the metric app context configuration, which was previously supplied as an env var to the relayers. This caused the env var size limit to be hit. As a workaround, larger configuration strings are now moved into a configmap that will be read in as a JSON by the relayer. I intentionally kept some of the configs as env vars, e.g. the relayChains, because this is much more ergonomic to change on the fly and many of us have operational muscle memory around this that I didn't want to break.
- Note one consequence of this is that merkly's ETH app is prone to running out of ETH and causing prep queues to rise. Because of this, I'll ignore merkly eth from our high urgency alert

### Drive-by changes

- There are a few places we were whitelisting warp routes in the gas config that I've now removed. From what I can tell none of these are necessary anymore as they were intended to be temporary workarounds.
- Updated the addresses for merkly from here https://minter.merkly.com/hyperlane/docs


typescript/infra/src/config/agent/relayer.ts
```diff
@@ -45,17 +49,31 @@ export interface BaseRelayerConfig {
...
 
 // Full relayer-specific agent config for a single chain
 export type RelayerConfig = Omit<RelayerAgentConfig, keyof AgentConfig>;
+// Config intended to be set as configMap values, these are usually really long
+// and are intended to derisk hitting max env var length limits.
+export type RelayerConfigMapConfig = Pick<
+  RelayerConfig,
+  'addressBlacklist' | 'gasPaymentEnforcement' | 'metricAppContexts'
+>;
+// The rest of the config is intended to be set as env vars.
```
yorhodes, line 63: do we enforce this? might be nice to use zod on process.env

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -344,101 +334,49 @@ const gasPaymentEnforcement: GasPaymentEnforcement[] = [
...
-  //   name: 'merkly_nft',
-  //   matchingList: routerMatchingList(merklyErc20Addresses),
-  // },
-];
+// Gets metric app contexts, including:
+// - helloworld
+// - all warp routes defined in WarpRouteIds, using addresses from the registry
+// - misc important applications not defined in the registry, e.g. merkly
+const metricAppContextsGetter = (): MetricAppContext[] => {
+  const warpContexts = Object.values(WarpRouteIds).map((warpRouteId) => {
```
yorhodes, line 351: whats the motivation behind a subset of the registry, so that our metrics dont get polluted?

yorhodes (approved):
lgtm
my takeaway was basically make the metric config builders async to pull the warp route addresses from the registry

typescript/infra/config/environments/mainnet3/agent.ts
tkporter: yeah exactly - atm we have high urgency alerts for these if the prep queue goes up, for now I'd like us to be stringent about which apps become eligible for a high urgency alert. Def open to changing this in the future though, ideally when it's easier to configure alerts via iaac or something

typescript/infra/src/config/agent/relayer.ts
tkporter: not enforced, this is just a practice I'm having us do internally. Because these are env vars piped into the agent image i think zod will be hard / overkill

typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -344,101 +334,49 @@ const gasPaymentEnforcement: GasPaymentEnforcement[] = [
...
+  });
+
+  return [
+    ...warpContexts,
+    {
+      name: 'helloworld',
+      matchingList: routerMatchingList(
+        helloWorld[Contexts.Hyperlane].addresses,
+      ),
+    },
```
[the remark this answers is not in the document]
tkporter: ha yeah, it was actually pretty useful one time recently where moonbeam made this gas metering change and we have no organic traffic to moonbeam

PR #4850, opened 2024-11-12
feat: Add scraping of Sealevel transactions into E2E tests

ameten:
### Description

Add scraping of Sealevel transactions into E2E tests

### Drive-by changes

- Add Stride as known Hyperlane domain
- Calculate log index for Sealevel chains from transaction itself

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4273

### Backward compatibility

Yes

### Testing

Manual run of Scraper
E2E tests for Ethereum and Sealevel

### DIFF
--- a/rust/main/agents/scraper/migration/src/m20230309_000001_create_table_domain.rs
+++ b/rust/main/agents/scraper/migration/src/m20230309_000001_create_table_domain.rs
@@ -462,6 +462,22 @@ const DOMAINS: &[RawDomain] = &[
         is_test_net: true,
         is_deprecated: false,
     },
+    RawDomain {
+        name: "sealeveltest1",
+        token: "SOL",
+        domain: 13375,
+        chain_id: 0,
+        is_test_net: true,
+        is_deprecated: false,
+    },
+    RawDomain {
+        name: "sealeveltest2",
+        token: "SOL",
+        domain: 13376,
+        chain_id: 0,
+        is_test_net: true,
+        is_deprecated: false,
+    },
 ];
 
 #[derive(DeriveMigrationName)]
--- a/rust/main/chains/hyperlane-sealevel/src/log_meta_composer.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/log_meta_composer.rs
@@ -36,7 +36,6 @@ impl LogMetaComposer {
     pub fn log_meta(
         &self,
         block: UiConfirmedBlock,
-        log_index: U256,
         pda_pubkey: &Pubkey,
         pda_slot: &Slot,
     ) -> Result<LogMeta, HyperlaneSealevelError> {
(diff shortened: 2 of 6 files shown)


rust/main/chains/hyperlane-sealevel/src/log_meta_composer.rs
```diff
@@ -64,15 +63,17 @@ impl LogMetaComposer {
...
-                .ok_or(HyperlaneSealevelError::NoTransactions(format!(
+        let (transaction_index, transaction_hash, program_index) = transaction_hashes
+            .into_iter()
+            .next()
+            .ok_or(HyperlaneSealevelError::NoTransactions(format!(
                 "block which should contain {} transaction does not contain any after filtering",
                 self.transaction_description,
             )))?;
 
+        // Construct log index which will be increasing relative to block
```
tkporter, line 74: fwiw I view us as being able to define this however we want per VM, and don't think we necessarily need to replicate the EVM behavior here, but happy as it is

rust/main/chains/hyperlane-sealevel/src/log_meta_composer.rs
```diff
@@ -64,15 +63,17 @@ impl LogMetaComposer {
...
+        let log_index = U256::from((transaction_index << 8) + (program_index as usize));
```
tkporter, line 75: Noting that because the log index is used to determine "uniqueness" of an IGP payment, changing this may have some consequences. If we re-index a payment, we'll consider them as two separate payments because the log index is different:
- scraper: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/agents/scraper/src/db/payment.rs#L58
- other agents: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/hyperlane-base/src/db/rocks/hyperlane_db.rs#L169-L181

What I'm not sure about is if we'd expect to actually double-process any gas payments?
On the relayer/validator side I'm pretty sure this won't happen, but unsure about the scraper.

rust/main/utils/run-locally/src/main.rs
```diff
@@ -448,19 +448,19 @@ fn main() -> ExitCode {
```
tkporter, line 448: should we add some invariants to the e2e test to make sure that these sealevel events / txs are being indexed? similar to what we do for the EVM things

rust/main/agents/scraper/migration/src/m20230309_000001_create_table_domain.rs
```diff
@@ -462,6 +462,22 @@ const DOMAINS: &[RawDomain] = &[
         is_test_net: true,
         is_deprecated: false,
     },
```
tkporter, line 464: Something I'm wondering - is the scraper indexing the SVM IGP payments and deliveries in a sequence aware fashion, or using the rate limited / watermark approach?

In the relayer it indexes SVM IGP payments in a sequence aware manner, but other ones in a rate limited / watermark approach. We'll want to do the same in the scraper otherwise I'd expect bugs to occur

For reference, this is how we do it in the relayer:

We create ContractSyncs here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/5db46bd315de02f1c7514aeee24aec349b6b8f80/rust/main/agents/relayer/src/relayer.rs#L161, which calls https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/5db46bd315de02f1c7514aeee24aec349b6b8f80/rust/main/hyperlane-base/src/settings/base.rs#L205-L250 that creates a `sequenced_contract_sync` if the CursorType is SequenceAware (the case for SVM IGP payments), and a `watermark_contract_sync` otherwise

The db used by the sequence aware cursors also requires it to impl `HyperlaneSequenceAwareIndexerStoreReader` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/5db46bd315de02f1c7514aeee24aec349b6b8f80/rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs#L31 which iiuc is only implemented for the Scraper's SQL db for messages atm https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/5db46bd315de02f1c7514aeee24aec349b6b8f80/rust/main/agents/scraper/src/chain_scraper/mod.rs#L393

PR #4851, opened 2024-11-12
chore: fix documentation of inaccurate account list for SVM synthetic transfer_remote

tkporter:
### Description

Removes an account that is not actually expected


tkporter:
closing in favor of #4854, where I include this as a driveby

PR #4860, opened 2024-11-15
feat: LRU Cache Implementation, ISM and PendingOperation::prepare caching

### LINKED ISSUE #3746: Pre-TGE: Caching ISM configurations in the relayer
## Problem

ISMs tend to be pretty static, we can afford to cache ISM configurations for a given address

## Solution

To have fewer RPCs when building metadata, cache ISM configs

Mantas-M:
### Description

Implementation of a LRU cache using the moka crate. 
Implementation of MeteredCache exposing cache hit and miss count

Caching ISM functions.
Caching PendingOperation::prepare functions.

Things to note:
1. Storing values which are expensive to clone, if there is a need for it down the line, 
we would need to go by the [guidelines](https://docs.rs/moka/latest/moka/future/struct.Cache.html#avoiding-to-clone-the-value-at-get):

```
If you want to store values that will be expensive to clone, wrap them by std::sync::Arc before storing in a cache. Arc is a thread-safe reference-counted pointer and its clone() method is cheap.
```

2. We can choose the hash function which will be used for cache key generation.
   The default is `std::hash::RandomState`, but we can choose a different one by providing a custom hasher.

3. We can change the ttl to be an `Option<ExpirationType>` in all the set functions,
   and resort to the default ttl if the value is `None`. But it may be less readable in the code
   outside the cache module.

### Drive-by changes

No

### Related issues

Closes #3746

### Backward compatibility

Yes

### Testing

Unit Tests - For Base Cache functionality
Manual Tests - For Metered Cache prometheus metrics


rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl Expiration {
+    /// Get the duration until the entry expires
+    pub fn as_duration(&self) -> Option<Duration> {
+        match self.variant {
+            ExpirationType::AfterDuration(duration) => Some(duration),
+            ExpirationType::AfterTimestamp(timestamp) => {
+                let target_time = UNIX_EPOCH + Duration::from_secs(timestamp);
+                target_time.duration_since(SystemTime::now()).ok()
+            }
```
ameten, line 49: I wonder whether an entry will be expired if this method is called after the moment in time specified by `AfterTimestamp(timestamp)`?

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
```diff
@@ -0,0 +1,104 @@
...
+use derive_new::new;
+use hyperlane_core::H256;
+use maplit::hashmap;
+use prometheus::IntCounterVec;
+use serde::{de::DeserializeOwned, Serialize};
+use std::fmt::Debug;
+
+use crate::cache::HyperlaneCache;
+
+use super::{CacheResult, Expiration};
```
ameten, line 12: Let's organize imports as follows (separated by empty line):
1. `std` sorted alphabetically
2. external crates (not Hyperlane) sorted alphabetically
3. hyperlane crates sorted alphabetically
4. internal `crate` imports
5. internal `super` imports

```suggestion
use std::fmt::Debug;

use async_trait::async_trait;
use derive_builder::Builder;
use derive_new::new;
use maplit::hashmap;
use prometheus::IntCounterVec;
use serde::{de::DeserializeOwned, Serialize};

use hyperlane_core::H256;

use crate::cache::HyperlaneCache;

use super::{CacheResult, Expiration};
```

rust/main/hyperlane-base/src/cache/error.rs
```diff
@@ -0,0 +1,13 @@
+/// Error type for the cache module.
+#[derive(thiserror::Error, Debug)]
+pub enum CacheError {
+    /// Error when key or value serialization fails.
+    #[error("Failed to serialize input: {0}")]
+    FailedToSerializeInuput(#[source] serde_json::Error),
```
ameten, line 6:
```suggestion
    FailedToSerializeInput(#[source] serde_json::Error),
```

rust/main/hyperlane-base/src/cache/error.rs
```diff
@@ -0,0 +1,13 @@
...
+
+    /// Error when enitity fetched from cache is deserialized incorrectly.
```
ameten, line 8:
```suggestion
    /// Error when entity fetched from cache is deserialized incorrectly.
```

rust/main/hyperlane-base/src/cache/error.rs
```diff
@@ -0,0 +1,13 @@
...
+    /// Most of the time this can be caused due the missmatch of the type
```
ameten, line 9:
```suggestion
    /// Most of the time this can be caused due the mismatch of the type
```

rust/main/hyperlane-base/src/cache/cache_types.rs
ameten, line 1: I wonder if this type should be closer where it is used, like in relayer metadata or even in hyperlane-ethereum?

rust/main/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
```diff
@@ -36,19 +36,48 @@ impl MultisigIsmMetadataBuilder for MerkleRootMultisigMetadataBuilder {
...
+                let index: u32 = unwrap_or_none_result!(
+                    self.highest_known_leaf_index().await,
+                    debug!("Couldn't get highest known leaf index")
+                );
+
+                self.cache_call_result(None, fn_name, &NoParams, &index)
+                    .await;
+                index
+            }
+        };
```
ameten, line 56: I would like to see this expansions of making a call directly to RPC with support for cached results to be extracted into an individual method. It contains a hidden contract on value of `fn_name` to be the same as RPC method name which can be explained in the method doc comment.

rust/main/hyperlane-base/src/cache/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+pub use error::*;
+pub use moka::*;
+
+use async_trait::async_trait;
+use hyperlane_core::H256;
+use serde::{de::DeserializeOwned, Serialize};
+
+/// Hyperlane Cache Interface
+#[async_trait]
+pub trait HyperlaneCache: Send + Sync {
```
ameten, line 15: I wonder if descriptive name will work better here like `ContractMethodCallCache`?

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::fmt::Debug;
+
+/// Cache for storing function calls with serializable results
+#[derive(Debug, Clone)]
+pub struct HyperlaneMokaCache(BaseCache);
+
+/// Used when the function has no parameters
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NoParams;
```
ameten, line 14: I wonder if `NoParams` is actually part of interface, not part of concrete implementation, and should be next to `HyperlaneCache`?

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+        fn_params: &impl Serialize,
+    ) -> CacheResult<Option<T>> {
+        let key = (contract_address, method, fn_params);
+        let value = self.get::<T>(&key).await?;
+
+        match value {
+            Some((value, _)) => Ok(Some(value)),
+            None => Ok(None),
+        }
+    }
```
ameten, line 66: It looks like all these methods are used only once in the implementation of trait `HyperlaneCache`. I wonder why we need to extract them as separate methods?

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Basic cache information.
+#[derive(Debug, Clone)]
+pub struct MeteredCacheConfig {
+    /// The name of the cache set on creation.
+    pub cache_name: &'static str,
```
ameten, line 18: I wonder why we need to make it `'static`?

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
ameten, line 1: It looks like that `MeteredCache` can work with any implementation of `HyperlaneCache`. I wonder why `MeteredCache` is placed inside `moka` module?

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
```diff
@@ -0,0 +1,95 @@
+use super::{BaseCache, CacheResult, Expiration, ExpirationType};
+use crate::cache::{CacheError, HyperlaneCache};
+use async_trait::async_trait;
+use hyperlane_core::H256;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::fmt::Debug;
+
+/// Cache for storing function calls with serializable results
+#[derive(Debug, Clone)]
+pub struct HyperlaneMokaCache(BaseCache);
```
ameten, line 10: If we rename `HyperlaneCache` into (for example) `ContractMethodCallCache`, would it make sense to hide the implementation detail about this cache (specifically `Moka` part) and call it `LocalContractMethodCallCache` (as example)?

rust/main/hyperlane-base/src/cache/moka/mod.rs
```diff
@@ -0,0 +1,322 @@
+/// Moka expiry trait implementation for dynamic lifetimes
+mod dynamic_expiry;
+mod hyperlane_cache;
+mod metered_cache;
+
+pub use dynamic_expiry::*;
+pub use hyperlane_cache::*;
+pub use metered_cache::*;
```
ameten, line 8: Let's export only symbols we are going to use outside.

ameten:
@Mantas-M, thank you for the contribution. Great start on LRU cache. I am happy to discuss the feedback on a call if it helps.

rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
Mantas-M: If the timestamp which is called with is in the past, the entity will auto expire, we can add a protection to allow only timestamps in the future

Mantas-M:
> @Mantas-M, thank you for the contribution. Great start on LRU cache. I am happy to discuss the feedback on a call if it helps.

Thank you for the review @ameten , could you please give me your telegram handle so we can have that quick overview call you suggested?

Mantas-M:
> @Mantas-M, thank you for the contribution. Great start on LRU cache. I am happy to discuss the feedback on a call if it helps.

Thank you for the review, I've contacted you to telegram for booking the quick call.

rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
Mantas-M: I investigated further into this and you were correct, returning `None` if the target time is earlier than `now` would [cause the entry to never expire](https://docs.rs/moka/latest/moka/policy/trait.Expiry.html#method.expire_after_create).

A fix was implemented that if this were to happen, we would return a zero Duration, leading to instant expiry.
Also a test case was added for this case, as suggested.

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/c87b460e78a8f7f056348664a0578bf14ce6309f)

Good catch

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
Mantas-M: Updated for all new cache files. Though not sure if the `mod.rs` files are entirely correct, might be good to double check.

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/5902bacd220ba8c8328cb79d5b8e518defbb12a3)

rust/main/hyperlane-base/src/cache/error.rs
Mantas-M: [Fixed](https://github.com/Mantas-M/hyperlane-monorepo/commit/5a1387390c87aaa923e0eb22fb43608e777db1d9)

rust/main/hyperlane-base/src/cache/error.rs
Mantas-M: [Fixed](https://github.com/Mantas-M/hyperlane-monorepo/commit/5a1387390c87aaa923e0eb22fb43608e777db1d9)

rust/main/hyperlane-base/src/cache/cache_types.rs
Mantas-M: Moved to isolated `ccip_read` module where it's used, to avoid bloating the cache_types in the future.

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/1b6b6adb45b06670c1c9d7780c52fd2c1aae6f13)

rust/main/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
Mantas-M: As discussed, moved the calls to their own functions to aid readability.
As a bonus, moved all the other cached ism calls as well. 

[Feat](https://github.com/Mantas-M/hyperlane-monorepo/commit/d7767694d9314029e47a55edcec261f8ca5ef12e)

rust/main/hyperlane-base/src/cache/mod.rs
Mantas-M: As discussed, should be more usage/feature specific naming, so I renamed it to `FunctionCallCache` since not all calls which are made are contract calls.

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/3a0c126000a14ec47acdf1bdc4752e551362ecba)

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
Mantas-M: As discussed, even though the interface does not rely on `NoParams`, we should still move it there so it would be a suggested parameter to use when storing/retrieving cached data.

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/f9dc61b63813d69a017ca628d592b2d2b73aab67)

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
Mantas-M: As discussed, good point 🫡 

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/36365e7834e41010c414193f89568fe40b461904)

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
Mantas-M: As discussed, changed `cache_name` to String to avoid needing to to be `'static`

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/5ddbe070c107f7c51f27f645cc6dbe7b3db389a8)

rust/main/hyperlane-base/src/cache/moka/metered_cache.rs
Mantas-M: Moved out of `moka` into `cache` since it can be implemented on other caches as well 🫡 

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/6bf5ede410186f98c9985a0581b1e28b1b91a77c)

rust/main/hyperlane-base/src/cache/moka/hyperlane_cache.rs
Mantas-M: As discussed, renamed to `LocalCache` to be more functionality specific 

[Fix](https://github.com/Mantas-M/hyperlane-monorepo/commit/fcdb62c2f7fcd22417178905a6341303e9d2fd7f)

rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
ameten: Nice.

rust/main/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
ameten: I added a doc comment to one of such methods. It will be great if similar comments are added to all the methods.

ameten:
I think we need to resolve a single conversation and this PR will be ready for a merge.

Mantas-M:
Seems like the Non-Cosmos E2E tests fail due to a timeout.

Example 1
tg_image_3583231269
Example 2
tg_image_1351088567

It is a possibility that since the cache stores a bunch of data and the runners are already slow and weak, it does not have enough resources to complete the run in time.

Since I cannot run the E2E tests due to GCP dependencies, I cannot test and debug it locally to see if that is actually the case.

Mantas-M:
@ameten Ran the E2E locally, and it seems like the runner issue theory is debunked and is not what is causing the problem, needs some further investigation into how the cache effects the E2E next year 🎄

Mantas-M:
Just a note on this issue, the caching works well, just needs some tweaking on the default values since the env in the EVM e2e tests is not optimized for caching and the tests time out before the cache expires. 

When you guys have some free time and want to merge this PR in, let me know and we can tweak the default values to work with the e2e tests.

Mantas-M:
Thanks for the response @daniel-savu 

That's quite unfortunate, since I was hoping to not need to maintain this PR. Will keep my fingers crossed that it's indeed a month and that the new changes don't introduce more work, which could be avoided by merging now.

rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
```diff
@@ -0,0 +1,87 @@
...
+                    .ok()
+                    .or(Some(Duration::ZERO))
+            }
+            ExpirationType::Never => None,
+            ExpirationType::Default => Some(Duration::from_secs(60 * 5)),
+        }
+    }
+
+    /// Calculate the time to live for the entry
+    /// Returns None if the entry should never expire or if the expiration time is in the past
```
tkporter, line 60: kinda feels like Duration::ZERO should be returned instead of None if the expiration time is in the past?

rust/main/hyperlane-base/src/cache/moka/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        let ttl = Expiration::from(ttl);
+        self.cache.insert(key, (value, ttl.clone())).await;
+        Ok(ttl)
+    }
+
+    /// Get the number of entries in the cache
+    /// This will run any pending tasks before returning the entry count
+    /// which ensures that the count is accurate.
+    /// Used only in test code.
```
tkporter, line 77: wonder if instead of naming these `_xxx` they should just live in a block like
```
#[cfg(test)]
impl BaseCache {
  // put them here
}
```

rust/main/hyperlane-base/src/cache/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    async fn cache_call_result(
+        &self,
+        contract_address: Option<H256>,
+        method: &str,
+        fn_params: &(impl Serialize + Send + Sync),
+        result: &(impl Serialize + Send + Sync),
+    ) -> CacheResult<Expiration>;
+
+    /// Get a cached call result with the given parameters
+    async fn get_cached_call_result<T>(
```
tkporter, line 34: after reading the PR I think the right abstraction may be:
```
async fn get_cached_call_result(
  &self,
  // e.g. "ethereum", etc
  // particularly useful for using with metrics
  chain_name: &str,
  // e.g. "is_contract" - ideally this is static and does not include any dynamic templating
  // bc then the metrics are more useful
  method: &str,
  // This would include the contract address if applicable, then params, etc
  method_params: &(impl Serialize + Send + Sync)
)
```

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -9,11 +9,16 @@ use std::{
 use async_trait::async_trait;
 use derive_new::new;
 use eyre::Result;
+
```
tkporter, line 12: noting that we'd also want to have this cached but feel free to not do this and we can do this later https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/aa68bafdb08fc4d4010246383ec1d86360ab8a9f/rust/main/agents/relayer/src/msg/metadata/base_builder.rs#L159

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -613,6 +599,151 @@ impl PendingMessage {
...
+    }
+
+    async fn store_to_cache(
+        &self,
+        contract_address: Option<H256>,
+        method: &str,
+        fn_params: &(impl Serialize + Send + Sync),
+        result: &(impl Serialize + Send + Sync),
+    ) {
+        if let Err(err) = self
```
tkporter, line 737: curious if you'd expect this to be a common thing or not? I assume this exists just to not error out of the logic when it's just a cache store issue and we want to preserve functionality

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -90,6 +90,42 @@ pub trait MultisigIsmMetadataBuilder: AsRef<MessageMetadataBuilder> + Send + Syn
...
+    /// Implicit contract in this method: function name `validators_and_threshold` matches
+    /// the name of the method `validators_and_threshold`.
+    async fn call_validators_and_threshold(
+        &self,
+        multisig_ism: &dyn MultisigIsm,
+        message: &HyperlaneMessage,
+        context: &'static str,
+    ) -> Result<(Vec<H256>, u8)> {
+        let contract_address = Some(multisig_ism.address());
+        let ism_domain = multisig_ism.domain().id();
```
tkporter, line 107: just noting if you go with my recommendation to move to using the human friendly chain name instead of id, you can do `multisig_ism.domain().name()`

tkporter (commented):
Very top tier work Mantas, thank you very much! Very impressed by the cleanliness of the code & the whole design of things 🔥🔥🔥

Left some recommendations around:
- things we don't want to cache
- some tweaks to the interface / metrics
- making sure there's a limit on cache capacity because we will be caching untrusted things and wanna make sure they don't grow forever
- some nits

rust/main/hyperlane-base/src/cache/moka/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+    cache: Cache<String, (String, Expiration), RandomState>,
+}
+
+/// The result type for cache operations, which can return a `CacheError`
+pub type CacheResult<T> = std::result::Result<T, CacheError>;
+
+impl BaseCache {
+    /// Create a new cache with the given name
+    pub fn new(name: &str) -> Self {
+        let cache = Cache::builder()
```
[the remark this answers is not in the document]
Mantas-M: Not sure regarding the default capacity, but we can add a size limit just in case.

However I don't quite understand your justification of creating one cache per chain as with the current implementation it would create one per agent which is shared.

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -113,6 +115,39 @@ impl AggregationIsmMetadataBuilder {
...
+    /// Implicit contract in this method: function name `modules_and_threshold` matches
+    /// the name of the method `modules_and_threshold`.
+    async fn call_modules_and_threshold(
+        &self,
+        ism: &dyn AggregationIsm,
+        message: &HyperlaneMessage,
+    ) -> eyre::Result<(Vec<H256>, u8)> {
+        let contract_address = Some(ism.address());
+        let ism_domain = ism.domain().id();
+        let fn_key = format!("modules_and_threshold_{}", ism_domain);
```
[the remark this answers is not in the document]
Mantas-M: Technically with some functions we would not really need it since due to the input data, the keys cannot clash. 
If we want to enforce encoding the domain everywhere then we can clean it up like you mentioned.

rust/main/hyperlane-base/src/cache/moka/mod.rs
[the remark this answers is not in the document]
tkporter: Ah my bad I missed that. Yes in that case let's pick an arbitrary number that should be good for all chains, maybe 50 MB or something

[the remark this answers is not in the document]
Mantas-M: Sounds good, will update to 50mb

rust/main/agents/relayer/src/msg/pending_message.rs
Mantas-M: Perhaps the way it could happen if you pass some data, which fails to get serialized, stored, it runs out of memory, etc. Generally edge cases and would not happen I believe.

And yeah it's just added as a `warn` that something's wrong with the cache

rust/main/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
```diff
@@ -15,6 +15,74 @@ use super::base::{MetadataToken, MultisigIsmMetadataBuilder, MultisigMetadata};
...
+        }
+    }
+
+    /// Returns the merkle leaf id by message id.
+    /// This method will attempt to get the value from cache first. If it is a cache miss,
+    /// it will request it from merkle tree prover. The result will be cached for future use.
+    ///
+    /// Implicit contract in this method: function name `get_merkle_leaf_id_by_message_id` matches
+    /// the name of the method `get_merkle_leaf_id_by_message_id`.
+    async fn call_get_merkle_leaf_id_by_message_id(
```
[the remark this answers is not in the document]
Mantas-M: Removed on conflict resolution
Removed on conflict resolution
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/26db9ecd55830813ab948a890be981295f947ae2

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -208,16 +215,9 @@ impl PendingOperation for PendingMessage {
         // If the message has already been processed, e.g. due to another relayer having
         // already processed, then mark it as already-processed, and move on to
         // the next tick.
-        let is_already_delivered = match self
-            .ctx
-            .destination_mailbox
-            .delivered(self.message.id())
-            .await
-        {
+        let is_already_delivered = match self.is_message_delivered().await {
```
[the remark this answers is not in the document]
Mantas-M: Removed on conflict resolution
https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/26db9ecd55830813ab948a890be981295f947ae2

rust/main/agents/relayer/src/msg/metadata/multisig/message_id_multisig.rs
```diff
@@ -16,6 +16,45 @@ use super::base::{MetadataToken, MultisigIsmMetadataBuilder, MultisigMetadata};
...
 pub struct MessageIdMultisigMetadataBuilder(MessageMetadataBuilder);
 
+impl MessageIdMultisigMetadataBuilder {
+    /// Returns the merkle leaf id by message id.
+    /// This method will attempt to get the value from cache first. If it is a cache miss,
+    /// it will request it from merkle tree prover. The result will be cached for future use.
+    ///
+    /// Implicit contract in this method: function name `get_merkle_leaf_id_by_message_id` matches
+    /// the name of the method `get_merkle_leaf_id_by_message_id`.
+    async fn call_get_merkle_leaf_id_by_message_id(
```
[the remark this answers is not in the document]
Mantas-M: Removed on conflict resolution
https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/26db9ecd55830813ab948a890be981295f947ae2

rust/main/agents/relayer/src/msg/metadata/multisig/merkle_root_multisig.rs
```diff
@@ -15,6 +15,74 @@ use super::base::{MetadataToken, MultisigIsmMetadataBuilder, MultisigMetadata};
...
 pub struct MerkleRootMultisigMetadataBuilder(MessageMetadataBuilder);
+
+impl MerkleRootMultisigMetadataBuilder {
+    /// Returns highest known leaf index.
+    /// This method will attempt to get the value from cache first. If it is a cache miss,
+    /// it will request it from merkle tree prover. The result will be cached for future use.
+    ///
+    /// Implicit contract in this method: function name `highest_known_leaf_index` matches
+    /// the name of the method `highest_known_leaf_index`.
+    async fn call_highest_known_leaf_index(&self) -> Result<Option<u32>> {
```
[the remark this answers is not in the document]
Mantas-M: Removed on conflict resolution
https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/26db9ecd55830813ab948a890be981295f947ae2

rust/main/hyperlane-base/src/cache/moka/mod.rs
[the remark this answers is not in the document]
Mantas-M: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/9e61465d8f4fed2d180bfa333d53db9efd362714

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -613,6 +599,151 @@ impl PendingMessage {
...
+        self.store_to_cache(contract_address, &fn_key, &fn_params, &ism_address)
+            .await;
+
+        Ok(ism_address)
+    }
+
+    async fn get_from_cache<U: DeserializeOwned>(
+        &self,
+        contract_address: Option<H256>,
+        method: &str,
```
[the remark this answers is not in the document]
Mantas-M: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/6129ecc939bf1e58b3adc535fba62d1f32429422

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -613,6 +599,151 @@ impl PendingMessage {
...
+        fn_params: &(impl Serialize + Send + Sync),
+        result: &(impl Serialize + Send + Sync),
+    ) {
+        if let Err(err) = self
+            .ctx
+            .cache
+            .cache_call_result(contract_address, method, fn_params, result)
+            .await
+        {
+            warn!(error=?err, "Error caching {:?} result", method);
```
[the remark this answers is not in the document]
Mantas-M: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/670bc03fe1063b5073ef04827bb47cb0db149ae9

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -613,6 +599,151 @@ impl PendingMessage {
...
+        contract_address: Option<H256>,
+        method: &str,
+        fn_params: &(impl Serialize + Send + Sync),
+    ) -> Option<U> {
+        self.ctx
+            .cache
+            .get_cached_call_result::<U>(contract_address, method, fn_params)
+            .await
+            .map_err(|err| {
+                warn!(error=?err, "Error checking cache stored {:?} result", method);
```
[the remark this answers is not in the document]
Mantas-M: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/670bc03fe1063b5073ef04827bb47cb0db149ae9

rust/main/hyperlane-base/src/cache/moka/mod.rs
Mantas-M: Yeah, IIRC I wanted to keep them as valid functions one can call on the cache if needed, but since they were only used in tests it was showing an `unused` warning.

I guess can be moved back to the main impl when they're needed
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/7184297b6b0e3d75203c2ae87653f1f6f904cc52

rust/main/agents/relayer/src/msg/pending_message.rs
Mantas-M: [Feeling free in that case](https://youtu.be/f4Mc-NYPHaQ?si=bLlhV3IdmyAfGuIS&t=45) 🤣

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -613,6 +599,151 @@ impl PendingMessage {
...
+        Ok(is_delivered)
+    }
+
+    /// Checks if the recipient is a contract.
+    /// This method will attempt to get the value from cache first. If it is a cache miss,
+    /// it will request it from the provider. The result will be cached for future use.
+    ///
+    /// Implicit contract in this method: function name `is_contract` matches
+    /// the name of the method `is_contract`.
+    async fn is_recipient_contract(&mut self) -> Result<bool, PendingOperationResult> {
```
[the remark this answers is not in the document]
Mantas-M: I honestly tried to implement it this way when I just started out and was facing a lot of issues. Would be nice for sure, but would need to dump quite a few more hours to get it working and then refactor 😓

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
Mantas-M: [Did just that](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/b4d2a222bac86996e35053a365eb05dd138092ae)

rust/main/hyperlane-base/src/cache/mod.rs
Mantas-M: [As thy wish](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/b4d2a222bac86996e35053a365eb05dd138092ae)

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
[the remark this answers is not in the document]
Mantas-M: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/b4d2a222bac86996e35053a365eb05dd138092ae

rust/main/hyperlane-base/src/cache/metered_cache.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// - `method`: the call stored in the cache.
+    /// - `status`: the status of the call.
+    #[builder(setter(into, strip_option), default)]
+    pub miss_count: Option<IntCounterVec>,
+}
+
+/// Expected label names for the metric.
+pub const HIT_COUNT_HELP: &str = "Number of cache hits";
+/// Help string for the metric.
+pub const HIT_COUNT_LABELS: &[&str] = &["cache_name", "method", "status"];
```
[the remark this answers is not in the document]
Mantas-M: [🫡🫡🫡🫡🫡🫡🫡🫡🫡🫡🫡🫡🫡🫡 ](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860/commits/65d1a0a787b21744c0e1f2483a5235166bd4aed9)

rust/main/hyperlane-base/src/cache/moka/dynamic_expiry.rs
tkporter: I guess my point is why is `None` used to cover both those very different use cases? How can you disambiguate the two? If we are already past the time where the Expiration should've expired, shouldn't we say that there are 0 more seconds to live?

Iiuc this is just used for a test so it doesn't totally matter though

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -90,6 +91,55 @@ pub trait MultisigIsmMetadataBuilder: AsRef<MessageMetadataBuilder> + Send + Syn
...
+    /// Implicit contract in this method: function name `validators_and_threshold` matches
+    /// the name of the method `validators_and_threshold`.
+    async fn call_validators_and_threshold(
+        &self,
+        multisig_ism: &dyn MultisigIsm,
+        message: &HyperlaneMessage,
+    ) -> Result<(Vec<H256>, u8), MetadataBuildError> {
+        let ism_domain = multisig_ism.domain().name();
+        let fn_key = "validators_and_threshold";
+        let call_params = (multisig_ism.address(), message);
```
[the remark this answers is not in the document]
Mantas-M: Haha, yeah it depends how you want it to look. 

The contract address was not needed only when caching the db calls, which are now removed, so it could be thrown back into the signature as a param. 

Could also make a new struct or a tuple type that can be used.

Really depends how you guys will use it in the future, this current implementation, the "Function call cache" is made to cache smart contract calls. If it extends beyond that you might have params that are not used and become options, or another implementation on top of the BaseCache.

For now it's just a matter of taste. (If I correctly understood what you mean with this comment)

Mantas-M:
Cant directly reply to the message for some reason
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860#discussion_r2033668690

The reason being why it's different is that it uses the `as_duration` function, which returns an Option, and that function is used to manage the TTLs within the cache. (And as I pointed out before, in that function the None and Duration::ZERO is understood differently)

But you are correct that it's only used in the test.

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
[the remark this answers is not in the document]
tkporter: Just meant it as an observation :)

tkporter (approved):
Thank you sir

PR #4862, opened 2024-11-15
fix: Use block slot instead of block height for Sealevel chains

ameten:
### Description

Use block slot instead of block height for Sealevel chains

### Backward compatibility

Yes

### Testing

Manual run of validator against Solana mainnet

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
@@ -257,7 +257,10 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
     #[instrument(level = "debug", err, ret, skip(self))]
     #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
     async fn get_finalized_block_number(&self) -> ChainResult<u32> {
-        self.rpc_client.get_block_height().await
+        // we should not report block height since SequenceAwareIndexer uses block slot in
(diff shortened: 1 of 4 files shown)


rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -257,7 +257,10 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
     #[instrument(level = "debug", err, ret, skip(self))]
     #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
     async fn get_finalized_block_number(&self) -> ChainResult<u32> {
-        self.rpc_client.get_block_height().await
+        // we should not report block height since SequenceAwareIndexer uses block slot in
+        // `latest_sequence_count_and_tip` and we should not report block slot here
+        // since block slot cannot be used as watermark
+        unimplemented!()
```
tkporter, line 263: I guess you'd prefer to just panic if the indexing logic ever changes such that we end up hitting this? Wondering if we should just have this return the slot instead so that any future logic could leverage this if it wanted
But anyways I don't have a strong opinion as e2e tests will fail anyways

tkporter (approved):

rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
ameten: Yes, the idea is to prevent to use rate-limited cursors for Sealevel since we should not use them.

PR #4868, opened 2024-11-18
feat: add Flow <> Ethereum cbBTC warp route to infra, mild refactor to warp config getters

tkporter:
### Description

- Includes the Flow cbBTC warp route getter (https://github.com/hyperlane-xyz/hyperlane-registry/pull/380)
- Includes the ownership transfer for the weETHs warp route
- Forces config getters to be more explicit about ownership by having the router config no longer implicitly set owner configs. See https://discord.com/channels/935678348330434570/1306554055895683113 for some context. Now you need to explicitly opt into transferring ownership to our colder setup

### Drive-by changes

- Removes a bunch of ts-ignores that aren't needed in the config getters


typescript/infra/config/warp.ts
```diff
@@ -77,6 +76,19 @@ export async function getWarpConfig(
...
+    return configWithoutOwner;
+  });
+  // Isolate the owners from the router config
+  const abacusWorksEnvOwnerConfig = objMap(routerConfig, (_chain, config) => {
+    const { owner, ownerOverrides } = config;
+    return {
+      owner,
+      ownerOverrides,
+    };
+  });
```
paulbalaji, line 91: nit: could do a single reduce over routerConfig to avoid the double loop

paulbalaji (approved):

ameten (approved):

typescript/infra/src/config/environment.ts
```diff
@@ -104,3 +104,10 @@ export async function getRouterConfigsForAllVms(
   // Merge, giving evmRouterConfig precedence
   return objMerge(allRouterConfigs, evmRouterConfig);
 }
+
+export function getOwnerConfigForAddress(owner: string): OwnableConfig {
+  return {
+    owner,
+    ownerOverrides: {},
```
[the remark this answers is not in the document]
tkporter: I'll add a comment, it's a sanity check to make sure any other overrides aren't present

typescript/infra/config/warp.ts
tkporter: gonna keep as is just for readability / easier types

PR #4870, opened 2024-11-18
feat: Add Zerion config generator

ltyu:
### Description
Adds Zero config generators for USDB, WBTC, BNB, Pol

### Drive-by changes
- Conditional for address zero check

### Backward compatibility
Yes

### Testing
Manual


typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+  const blast: TokenRouterConfig = {
+    ...routerConfig.blast,
+    type: TokenType.collateral,
+    token: USDB_ADDRESS,
+    interchainSecurityModule: {
+      owner: '0x723e7694dc346e5a15fB6F6A0144479aC624C66F',
+      type: IsmType.FALLBACK_ROUTING,
+      domains: {},
+    },
+    owner: safes.blast,
```
tkporter, line 25: we should remove all explicit `owner` setting and any use of the `safes` object directly - if you console.log `routerConfig.blast`, you'll see that the owner is already set as part of `routerConfig.blast`

The reason is that we don't always want Safe ownership, sometimes we want ICA ownership. E.g. on scroll, we actually want our ICA on there to own this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/owners.ts#L118

The infra checker's logic says "if the config asks for an owner that is equal to the environment's intended owner address (which is sometimes a Safe, sometimes an ICA), then make sure that the proxy admin that's used is the same one for the rest of the environment (i.e. the one from addresses.yaml. Otherwise, if a different owner is desired, keep the existing proxy admin and transfer ownership of that proxy admin to your different desired owner"

This is why you're seeing the different behavior on scroll - the desired owner for the mainnet environment on scroll is the ICA, but you're setting it to the Safe

typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+const USDB_ADDRESS = '0x4300000000000000000000000000000000000003';
+
+export const getBlastZeroNetworkUSDBWarpConfig = async (
+  routerConfig: ChainMap<RouterConfig>,
+): Promise<ChainMap<TokenRouterConfig>> => {
+  const blast: TokenRouterConfig = {
+    ...routerConfig.blast,
+    type: TokenType.collateral,
+    token: USDB_ADDRESS,
+    interchainSecurityModule: {
```
tkporter, line 20: is this not able to be set to `ethers.constants.AddressZero`?
imo we should much prefer this over the fallback routing hook

typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+      owner: '0x723e7694dc346e5a15fB6F6A0144479aC624C66F',
```
tkporter, line 21: if we do need to keep it as a fallback routing hook, can we make sure that ownership of this is either set to the dead address, or have it be set to `routerConfig.<chain>.owner`?

typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
ltyu: Will leave this as is until we fix https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4871

typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
ltyu: updated to address(0)

typescript/infra/config/environments/mainnet3/warp/configGetters/getBlastZeroNetworkUSDBWarpConfig.ts
tkporter: could we actually just do this now?

I think it's fine for this to be inconsistent with what the CLI does - I doubt we'll wanna update this later because it requires multisig interactions so imo there's no downside to moving over to the zero address now

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumModeScrollZeronetworkWBTCConfig.ts
```diff
@@ -0,0 +1,70 @@
...
+      owner: ethers.constants.AddressZero,
+      type: IsmType.FALLBACK_ROUTING,
+      domains: {},
+    },
+  };
+
+  const scroll: TokenRouterConfig = {
+    ...routerConfig.scroll,
+    type: TokenType.collateral,
+    token: collateralAddresses.token,
```
tkporter, line 46:
```suggestion
    token: collateralAddresses.scroll,
```

tkporter (approved):

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumModeScrollZeronetworkWBTCConfig.ts
ltyu: good catch

tkporter (approved):

PR #4874, opened 2024-11-19
feat: more app contexts

tkporter:
### Description

- Adds the endurance USDC warp deploy so it's a relayer app context
- Adds Velodrome contracts, contracts were found from their private fork / in tg with them https://github.com/velodrome-finance/superchain-contracts-private/tree/main/deployment-addresses
- Adds Aave contracts, contracts were found by poking around https://app.aave.com/governance/v3/proposal/?proposalId=92, which led to https://github.com/bgd-labs/aave-delivery-infrastructure?tab=readme-ov-file#deployed-addresses and seeing which contracts were sending/receiving messages in our explorer. They don't have a typical router setup -- some messaging is unidirectional. The sending address is more static than the receiving address, so I've set the matching list to be based off the sender to make configuration simpler. https://github.com/bgd-labs/aave-delivery-infrastructure?tab=readme-ov-file#deployed-addresses is the source of truth for these addresses

### Drive-by changes

- Forces a warp ID to be picked in the warp monitor flow


paulbalaji (approved):
lgtm, just need to run `yarn agent-configs` and commit to make ci happy

jmrossy (approved):

PR #4875, opened 2024-11-19
fix: Make main and sealevel workspaces compilable with our forked dependencies

ameten:
### Description

- Make main and sealevel workspaces compilable with our forked dependencies
- Make Sealevel client be able to parse block with new kind of instruction errors

### Backward compatibility

Yes

### Testing

E2E Ethereum and Sealevel tests

### DIFF
--- a/rust/main/Cargo.toml
+++ b/rust/main/Cargo.toml
@@ -124,7 +124,7 @@ serde_derive = "1.0"
 serde_json = "1.0"
 sha2 = { version = "0.10.6", default-features = false }
 sha256 = "1.1.4"
- sha3 = "0.10"
+sha3 = "0.10"
 solana-account-decoder = "=1.14.13"
 solana-client = "=1.14.13"
 solana-sdk = "=1.14.13"
--- a/rust/main/Cargo.toml
+++ b/rust/main/Cargo.toml
@@ -236,42 +236,42 @@ version = "=0.5.2"
 
 [patch.crates-io.solana-account-decoder]
 git = "https://github.com/hyperlane-xyz/solana.git"
-branch = "ameten/hyperlane-1.14.13-compilable"
+tag = "hyperlane-1.14.13-2024-11-20"
 version = "=1.14.13"
 
 [patch.crates-io.solana-clap-utils]
 git = "https://github.com/hyperlane-xyz/solana.git"
-branch = "ameten/hyperlane-1.14.13-compilable"
+tag = "hyperlane-1.14.13-2024-11-20"
 version = "=1.14.13"
 
 [patch.crates-io.solana-cli-config]
 git = "https://github.com/hyperlane-xyz/solana.git"
-branch = "ameten/hyperlane-1.14.13-compilable"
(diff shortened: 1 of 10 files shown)


rust/main/Cargo.toml
```diff
@@ -124,26 +124,11 @@ serde_derive = "1.0"
 serde_json = "1.0"
 sha2 = { version = "0.10.6", default-features = false }
 sha256 = "1.1.4"
-sha3 = "0.10"
+ sha3 = "0.10"
```
paulbalaji, line 127:
```suggestion
sha3 = "0.10"
```

ameten: Thank you! Fixed

rust/main/Cargo.toml
```diff
@@ -126,24 +126,9 @@ sha2 = { version = "0.10.6", default-features = false }
 sha256 = "1.1.4"
 sha3 = "0.10"
 solana-account-decoder = "=1.14.13"
-solana-banks-client = "=1.14.13"
```
tkporter, line 129: were you able to remove all these because they're specified down below in patch?

rust/main/chains/hyperlane-sealevel/src/rpc/client/tests.rs
```diff
@@ -0,0 +1,14 @@
+use crate::SealevelRpcClient;
+
+//#[tokio::test]
+async fn _test_get_block() {
```
tkporter, line 4: just wanna confirm this test is intentionally kept?

rust/main/Cargo.toml
```diff
@@ -256,42 +236,42 @@ version = "=0.5.2"
 
 [patch.crates-io.solana-account-decoder]
 git = "https://github.com/hyperlane-xyz/solana.git"
-tag = "hyperlane-1.14.13-2023-07-04"
+branch = "ameten/hyperlane-1.14.13-compilable"
```
tkporter, line 239: I assume after https://github.com/hyperlane-xyz/solana/pull/2 is merged you'll create a tag there instead of relying on the branch?

tkporter (approved):

rust/main/Cargo.toml
tkporter: ah yes I see your discord message about this

rust/main/Cargo.toml
ameten: `solana-banks-client` is not used in `main` workspace and it is not specified in `patch` section of this file.

rust/main/Cargo.toml
ameten: It is correct.

rust/main/chains/hyperlane-sealevel/src/rpc/client/tests.rs
ameten: Yes, I thought to keep some test code even if we don't run it. Future engineer will be able to use it with another block/slot.

PR #4878, opened 2024-11-19
feat: add GcpValidator using GcpStorageWrapper

paulbalaji:
### Description

Introduce GcpValidator for retrieving announcements, checkpoints and metadata for a Validator posting to a GCP bucket. Uses GcpStorageWrapper for bucket operations.

### Drive-by changes

update `check-validator-version` to read gcp buckets too

### Backward compatibility

yes

### Testing

manual

before:
```
Error getting metadata for validator 0x14ADB9e3598c395Fe3290f3ba706C3816Aa78F59 on chain flowmainnet: Error: Unable to parse location gs://ff-hyperlane-validator-1/gcsAnnouncementKey

│ 18      │ 'flowmainnet'  │ '0x14ADB9e3598c395Fe3290f3ba706C3816Aa78F59' │ 'flow foundation' │ 'UNKNOWN'                                  │
```

after:
```
│ 22      │ 'flowmainnet'           │ '0x14ADB9e3598c395Fe3290f3ba706C3816Aa78F59' │ 'flow foundation' │ 'nov-7-batch'              │
```

### DIFF
--- a/typescript/infra/scripts/check/check-validator-version.ts
+++ b/typescript/infra/scripts/check/check-validator-version.ts
@@ -3,6 +3,7 @@ import { execSync } from 'child_process';
 import { ValidatorAnnounce__factory } from '@hyperlane-xyz/core';
 import {
   ChainName,
+  GcpValidator,
   S3Validator,
   defaultMultisigConfigs,
 } from '@hyperlane-xyz/sdk';
@@ -142,9 +143,14 @@ async function main() {
 
         // Get metadata from each storage location
         try {
-          const validatorInstance = await S3Validator.fromStorageLocation(
-            location,
-          );
+          let validatorInstance;
+          if (location.startsWith('gs://')) {
+            validatorInstance = await GcpValidator.fromStorageLocation(
+              location,
+            );
+          } else {
+            validatorInstance = await S3Validator.fromStorageLocation(location);
+          }
 
           const metadata = await validatorInstance.getMetadata();
           const gitSha = metadata?.git_sha;
--- a/typescript/infra/scripts/check/check-validator-version.ts
+++ b/typescript/infra/scripts/check/check-validator-version.ts
@@ -3,9 +3,8 @@ import { execSync } from 'child_process';
 import { ValidatorAnnounce__factory } from '@hyperlane-xyz/core';
 import {
   ChainName,
-  GcpValidator,
-  S3Validator,
   defaultMultisigConfigs,
+  getValidatorFromStorageLocation,
 } from '@hyperlane-xyz/sdk';
 import { Address } from '@hyperlane-xyz/utils';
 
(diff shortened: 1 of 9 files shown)


typescript/infra/scripts/check/check-validator-version.ts
```diff
@@ -146,8 +147,16 @@ async function main() {
 
         // Get metadata from each storage location
         try {
-          const s3Validator = await S3Validator.fromStorageLocation(location);
-          const metadata = await s3Validator.getMetadata();
+          let validatorInstance;
+          if (location.startsWith('gs://')) {
```
tkporter, line 151: might be useful to have this logic somewhere for other consumers to use eventually

typescript/sdk/src/gcp/storage.ts
```diff
@@ -0,0 +1,89 @@
...
+    } catch (e: any) {
+      if (e.code === 404) {
+        return undefined;
+      }
+      throw e;
+    }
+  }
+
+  url(key: string): string {
+    const Key = this.formatKey(key);
```
tkporter, line 86: nit: is this intentionally capitalized? if we don't wanna conflict with the existing key i'd pick a different camelCase name

typescript/sdk/src/gcp/storage.ts
```diff
@@ -0,0 +1,89 @@
...
+        return undefined;
+      }
+
+      const [metadata] = await file.getMetadata();
+      const [contents] = await file.download();
+      const body = contents.toString('utf-8');
+
+      const result = {
+        data: JSON.parse(body),
+        modified: new Date(metadata.updated ?? ''),
```
tkporter, line 70: Is the `?? ''` as a constructor param intentional?
```
> new Date('')
Invalid Date
```

typescript/sdk/src/gcp/validator.ts
```diff
@@ -0,0 +1,130 @@
...
+        storageLocation = storageLocation.slice(0, -1);
+      }
+    }
+    if (storageLocation.startsWith(LOCATION_PREFIX)) {
+      const bucketName = storageLocation.slice(LOCATION_PREFIX.length);
+      const pieces = bucketName.split('/');
+      if (pieces.length >= 1) {
+        const storageConfig = {
+          bucket: pieces[0],
+          folder: pieces.slice(1).join('/'),
```
tkporter, line 51: iiuc this will set `folder` to `''` instead of `undefined` if there's no folder, is that intentional?

tkporter (approved):

typescript/sdk/src/gcp/storage.ts
paulbalaji: intentional but changed to unix start date instead as the backup if `updated` is not provided

Issue #4271, opened 2024-08-07
Epic: Solana scraper support


tkporter:
Note: sequencing-wise, it may make sense to coordinate with https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3329 for many things including shipping this

```[tasklist]
### Tasks
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3331
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4272
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4301
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4273
- [ ] https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4274
```

PR #4883, opened 2024-11-21
fix: Avoid Relayer to use GetBlock for Sealevel

ameten:
### Description

Avoid Relayer to use GetBlock for Sealevel.

It is a temporary solution until we migrate Sealevel integration to the latest Solana libraries

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4271

### Backward compatibility

Yes

### Testing

E2E Ethereum and Sealevel test

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/mailbox.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/mailbox.rs
@@ -648,10 +648,15 @@ pub struct SealevelMailboxIndexer {
     program_id: Pubkey,
     dispatch_message_log_meta_composer: LogMetaComposer,
     delivery_message_log_meta_composer: LogMetaComposer,
+    advance_log_index: bool,
 }
 
 impl SealevelMailboxIndexer {
-    pub fn new(conf: &ConnectionConf, locator: ContractLocator) -> ChainResult<Self> {
+    pub fn new(
+        conf: &ConnectionConf,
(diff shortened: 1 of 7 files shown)


rust/main/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -648,10 +648,15 @@ pub struct SealevelMailboxIndexer {
...
     delivery_message_log_meta_composer: LogMetaComposer,
+    advance_log_index: bool,
 }
 
 impl SealevelMailboxIndexer {
-    pub fn new(conf: &ConnectionConf, locator: ContractLocator) -> ChainResult<Self> {
+    pub fn new(
+        conf: &ConnectionConf,
+        locator: ContractLocator,
+        advance_log_index: bool,
```
tkporter, line 658: I see some are `advance_log_meta` and some are `advance_log_index`, should we consolidate around one?
also wondering if it should be `advanced_..`

tkporter (approved):

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
ameten: Yes, just a misprint. 

Regarding `advance...`. The idea is that relayer relies on a `basic` log meta while scraper relies on `advanced` one.

PR #4884, opened 2024-11-21
feat: ensure consistent decimals in Token checker

tkporter:
### Description

- ensures actual (i.e. looking at onchain info, or chain metadata for native warp routes) decimals are consistent and nonzero
- ensures any configured decimals are also consistent (but doesn't require them to be present)
- cobbled together pretty quickly - not sure if there's a more elegant way of having non-chain-specific checks

### DIFF
--- a/typescript/sdk/src/token/checker.ts
+++ b/typescript/sdk/src/token/checker.ts
@@ -69,7 +69,7 @@ export class HypERC20Checker extends ProxiedRouterChecker<
     const hypToken = this.app.router(this.app.getContracts(chain));
 
     // Check all actual decimals are consistent
-    const actualChainDecimals = await this.getAllActualDecimals();
+    const actualChainDecimals = await this.getEvmActualDecimals();
     this.checkDecimalConsistency(
       chain,
       hypToken,
@@ -128,14 +128,10 @@ export class HypERC20Checker extends ProxiedRouterChecker<
     }
   }
 
-  getChainDecimals(): Record<ChainName, number | undefined> {
-    return objMap(this.configMap, (_chain, config) => config.decimals);
-  }
-
   private cachedAllActualDecimals: Record<ChainName, number> | undefined =
     undefined;
 
-  async getAllActualDecimals(): Promise<Record<ChainName, number>> {
+  async getEvmActualDecimals(): Promise<Record<ChainName, number>> {
     if (this.cachedAllActualDecimals) {
       return this.cachedAllActualDecimals;
     }
--- a/typescript/sdk/src/token/checker.ts
+++ b/typescript/sdk/src/token/checker.ts
@@ -5,8 +5,9 @@ import {
   ERC20__factory,
   HypERC20Collateral,
   IXERC20Lockbox__factory,
+  TokenRouter,
 } from '@hyperlane-xyz/core';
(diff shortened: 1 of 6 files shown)


typescript/sdk/src/token/checker.ts
```diff
@@ -86,37 +111,119 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
-        this.addViolation(violation);
-      }
+    }
+    if (!collateralToken) {
+      throw new Error('Collateral token not found');
+    }
+    return collateralToken;
+  }
+
+  async checkDecimalConsistency(
```
paulbalaji, line 203: not an async function - will also fix the lint/prettier errors you're getting
```suggestion
  checkDecimalConsistency(
```

typescript/sdk/src/token/checker.ts
```diff
@@ -86,37 +111,119 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
+    const entries = await Promise.all(
+      this.getEvmChains().map(async (chain) => {
+        const token = this.app.router(this.app.getContracts(chain));
+        return [chain, await this.getActualDecimals(chain, token)];
+      }),
+    );
+
+    this.cachedAllActualDecimals = Object.fromEntries(entries);
+
+    return Object.fromEntries(entries);
```
jmrossy, line 147:
```suggestion
    return this.cachedAllActualDecimals;
```

typescript/sdk/src/token/checker.ts
```diff
@@ -86,37 +111,119 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
+    hypToken: TokenRouter,
+  ): Promise<number> {
+    const expectedConfig = this.configMap[chain];
+    let decimals: number | undefined = undefined;
+
+    if (isNativeConfig(expectedConfig)) {
+      decimals =
+        this.multiProvider.getChainMetadata(chain).nativeToken?.decimals;
+    } else if (isSyntheticConfig(expectedConfig)) {
+      decimals = await (hypToken as unknown as ERC20).decimals();
```
jmrossy, line 161: Could this cast be avoided if `isSyntheticConfig` had a `: x is ERC20` type clause on it?

typescript/sdk/src/token/checker.ts
```diff
@@ -86,37 +111,119 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
+    } else if (isCollateralConfig(expectedConfig)) {
+      const collateralToken = await this.getCollateralToken(chain);
+      decimals = await collateralToken.decimals();
+    }
+
+    if (!decimals) {
```
jmrossy, line 167: This will throw if decimals === 0, just confirming that's what you want?

jmrossy (approved):

typescript/sdk/src/token/checker.ts
tkporter: think this would require the expectedConfig coming from hypToken or passing hypToken into `isSyntheticConfig` - for now I'll keep as is (this was what we were doing elsewhere before this PR too)

typescript/sdk/src/token/checker.ts
tkporter: fixed ty :)

PR #4891, opened 2024-11-22
feat: Add hook update into WarpModule.update()

ltyu:
### Description
This is a multi-part PR.

This adds `createHookUpdateTxs()` to `WarpModule.update()` such that it updates a warp route _without_ a hook, and one with an existing hook.

### Backward compatibility
Yes

### Testing
Unit Tests


typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -42,12 +46,14 @@ import { EvmERC20WarpRouteReader } from './EvmERC20WarpRouteReader.js';
 import { HypERC20Deployer } from './deploy.js';
 import { TokenRouterConfig, TokenRouterConfigSchema } from './schemas.js';
 
+``;
```
jmrossy, line 49: typo?

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -343,6 +383,123 @@ export class EvmERC20WarpModule extends HyperlaneModule<
     return { deployedIsm, updateTransactions };
   }
 
+  /**
+   * Updates or deploys the hook using the provided configuration.
+   *
+   * @returns Object with deployedHook address, and update Transactions
+   */
+  public async deployOrUpdateHook(
```
jmrossy, line 391: Nit, public keyword is unnecessary

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -343,6 +383,123 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+      staticMessageIdWeightedMultisigIsmFactory,
+    } = this.args.addresses;
+
+    assert(actualConfig.proxyAdmin?.address, 'ProxyAdmin address is undefined');
+
+    const hookModule = new EvmHookModule(
+      this.multiProvider,
+      {
+        chain: this.args.chain,
+        config: actualConfig.hook as HookConfig,
```
jmrossy, line 477: Can the casts here and below be avoided?

jmrossy (approved):
I don't know this code well but the changes look fine to me :)

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: the cast here can be as i will use an `assert`

The case below for `deployedHook` is trickier because we add `address` to the derived hook config. The type `TokenRouterConfig.hook` doesn't have `address`, but it actually does. 

This is tech debt and would be better to create a `DerivedTokenRouteConfig`, but we need to add additional logic to remove `address` when we do the config diff.

ltyu: Created an issue for this: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4932

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: Removed

typescript/sdk/src/token/EvmERC20WarpModule.ts
ltyu: removed

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -280,6 +280,47 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+    const actualDeployedHook = (actualConfig.hook as DerivedHookConfig)
+      ?.address;
+
+    // Try to deploy or update Hook with the expected config
+    const {
+      deployedHook: expectedDeployedHook,
+      updateTransactions: hookUpdateTransactions,
+    } = await this.deployOrUpdateHook(actualConfig, expectedConfig);
+
+    // If an Hook is updated in-place, push the update txs
```
paulbalaji, line 302:
```suggestion
    // If a Hook is updated in-place, push the update txs
```

paulbalaji (approved):

xeno097 (approved):

PR #4892, opened 2024-11-24
feat: cursor metrics

### LINKED ISSUE #4715: Expose metrics for sequence indexing cursor progress
# Context 
Sequenced events like `HyperlaneMessage` and `MerkleTreeInsertion` are indexed using a forward-backward cursor which is initialized to the sequence number at the blockchain tip, when the agent is started. 

For instance, if the last `HyperlaneMessage` had nonce 123, the forward-backward cursor would instantiate:
- a forward cursor that looks for messages in new blocks, with nonces 124 and up
- a backward cursor that iterates blocks from the tip back to genesis (or to whatever is configured in `index.from`)

# Problem
When an RPC becomes healthy after an outage, it's hard to know if agents are making any progress indexing the missed events over that period. If they're running but not making any progress, the RPC may still be having issues. The only way to know this is by checking raw logs, which requires knowledge of what to look for.

# Suggested Solution
A simple way to provide better observability here is to have some new Prometheus metrics for indexed events:
- for all event types (regardless if sequenced or not): the block number the cursor(s) are currently at
- for sequenced events: the sequence number that the backward and forward cursors are on, and the max known sequence

To continue with the `HyperlaneMessage` example, the metrics at a random point for a given chain would look something like:
- `cursor_current_block{eventType: 'HyperlaneMessage', cursorType: 'ForwardSequenced'}: 118000`
- `cursor_current_sequence{eventType: 'HyperlaneMessage', cursorType: 'ForwardSequenced'}: 240`
- `cursor_current_block{eventType: 'HyperlaneMessage', cursorType: 'BackwardSequenced'}: 91021`
- `cursor_current_sequence{eventType: 'HyperlaneMessage', cursorType: 'BackwardSequenced'}: 82`
- `cursor_max_sequence{eventType: 'HyperlaneMessage'}: 345`

Based on this, it will be much easier to have dashboard showing the indexing progress agents make after each RPC rotation and restart, per indexed event and chain.

Mantas-M:
### Description

Adds metrics to ForwardBackward cursor in order to be able to monitor sync progress more effectively.

Introduced metrics
- `cursor_current_block`: Which block is the forward and backward cursor indexing currently for each event
Cursor Current Block BSC
_Example above shows all event average_

- `cursor_current_sequence`: Which sequence is the cursor currently on for a specific event
Cursor Current Sequence
_As in the graph for Base above, we can see that the Forward cursor only begins showing data if there are new events to index since agent launch_

- `cursor_max_sequence`: The max sequence the cursor can reach
image

### Related issues

- Fixes #4715

### Backward compatibility

Yes

### Testing

Manual


Mantas-M:
Read the issue again and I thought to refactor the metrics to be cursor generic due to this requirement set by @daniel-savu 
`for all event types (regardless if sequenced or not): the block number the cursor(s) are currently at`

Initially seems like it would be not necessary due to the `contract_sync_block_height` metric doing mostly the same thing:

image

But this makes it cleaner in my opinion and when/if the internal Grafana dashboards would be migrated to the new metric, `contract_sync_block_height` could be removed entirely. 

Now with the `CursorMetrics` being passed in to both the `RateLimitedContractSyncCursor` and the `ForwardBackwardSequenceAwareSyncCursor` we get a nice group of metrics from both sequence aware and non sequence aware event indexing cursor block heights:

image

Additionally, since the implementation of `name` on the `Indexable` trait, we could technically remove the label strings passed into the `sync` function making it a bit nicer and keeping all the hardcoded strings on the `Indexable` trait implementation.

image

rust/main/hyperlane-base/src/contract_sync/cursors/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// - `chain`: Chain the cursor is collecting data from.
+    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited` or `backward_rate_limited`.
+    pub cursor_current_block: IntGaugeVec,
+
+    /// Current sequence of the cursor.
+    /// Only used by sequence aware cursors.
+    /// Labels:
+    /// - `event_type`: the event type the cursor is indexing. Could be anything implementing `Indexable`.
+    /// - `chain`: Chain the cursor is collecting data from.
+    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited` or `backward_rate_limited`.
```
ameten, line 21: I wonder if values `forward_rate_limited` and `backward_rate_limited` are valid for this metric?

rust/main/hyperlane-base/src/contract_sync/cursors/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Struct encapsulating prometheus metrics used by SequenceAware and RateLimited cursors.
+#[derive(Debug, Clone)]
+pub struct CursorMetrics {
+    /// Current block of the cursor.
+    /// Used by both sequence aware and rate limited cursors.
+    /// Labels:
+    /// - `event_type`: the event type the cursor is indexing. Could be anything implementing `Indexable`.
+    /// - `chain`: Chain the cursor is collecting data from.
+    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited` or `backward_rate_limited`.
+    pub cursor_current_block: IntGaugeVec,
```
ameten, line 14: I doubt that we have `backward_rate_limited` cursors.

rust/main/hyperlane-base/src/contract_sync/cursors/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Struct encapsulating prometheus metrics used by SequenceAware and RateLimited cursors.
+#[derive(Debug, Clone)]
+pub struct CursorMetrics {
+    /// Current block of the cursor.
+    /// Used by both sequence aware and rate limited cursors.
+    /// Labels:
+    /// - `event_type`: the event type the cursor is indexing. Could be anything implementing `Indexable`.
+    /// - `chain`: Chain the cursor is collecting data from.
+    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited` or `backward_rate_limited`.
```
Mantas-M, line 13:
```suggestion
    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited`.
```

rust/main/hyperlane-base/src/contract_sync/cursors/metrics.rs
Mantas-M, line 21:
```suggestion
    /// - `cursor_type`: The type of cursor. E.g. `forward_sequenced`, `backward_sequenced`, `forward_rate_limited`.
```

rust/main/hyperlane-base/src/contract_sync/cursors/rate_limited.rs
```diff
@@ -155,12 +165,24 @@ impl<T> RateLimitedContractSyncCursor<T> {
...
+            .with_label_values(label_values)
+            .set(latest_block as i64);
+    }
 }
 
 #[async_trait]
 impl<T> ContractSyncCursor<T> for RateLimitedContractSyncCursor<T>
 where
-    T: Send + Sync + Debug + 'static,
+    T: Send + Sync + Debug + 'static + Indexable,
```
ameten, line 185:
```suggestion
    T: Indexable + Send + Sync + Debug + 'static,
```

rust/main/hyperlane-base/src/contract_sync/cursors/metrics.rs
Mantas-M: `forward_rate_limited` is indeed valid as it's the only mode for the RateLimitedCursor, as seen [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4892#issuecomment-2504574528)

`backward_rate_limited` removed in both occurrences.

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
```diff
@@ -67,11 +70,17 @@ pub(crate) struct ForwardBackwardSequenceAwareSyncCursor<T> {
     forward: ForwardSequenceAwareSyncCursor<T>,
     backward: BackwardSequenceAwareSyncCursor<T>,
     last_direction: SyncDirection,
+    metrics: Arc<CursorMetrics>,
```
daniel-savu, line 73: curious what motivated storing the metrics in`ForwardBackwardSequenceAwareSyncCursor` instead of inside `ForwardSequenceAwareSyncCursor` and `BackwardSequenceAwareSyncCursor`. I think placing them inside the individual `Forward` and `Backward` cursors would be better for composability (say, a future cursor type that only uses the `ForwardSequenceAwareSyncCursor` - with the current setup you'd need to duplicate the metrics updating logic for it)

daniel-savu (commented):
mostly lgtm but will leave the final green light to @ameten!

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/mod.rs
ameten: It is a very good point. We have PR which makes the indexing direction configurable by chain. https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4990

Mantas-M: Ah, was not aware of this type of functionality. Will have it updated. Thanks for the feedback

ameten (approved):

Mantas-M:
Just a note about the failing E2E test @ameten 

Seems like a dependency package called `home` of some crate used by sealevel [just got a fresh update](https://crates.io/crates/home/versions) requiring a higher rustc version.
image

Not related to this PR, but should be pinned down to the previous version to avoid future issues.

ameten:
The dependency issue is fixed in main. @Mantas-M

Mantas-M:
@ameten Up to date with `main` and ready to merge ✅

PR #4893, opened 2024-11-25
chore: transfer ownership for cbBTC warp deploy on Flow <> Ethereum

tkporter:
### Description

- Transfers ownership of the cbBTC warp route between flow and ethereum
- A fix to ensure that proxy admin ownership is set

### Drive-by changes

- a driveby to prevent rebuilding the multiprovider tons of times in the check warp deploy script

### DIFF
--- a/typescript/infra/src/config/environment.ts
+++ b/typescript/infra/src/config/environment.ts
@@ -109,6 +109,8 @@ export function getOwnerConfigForAddress(owner: string): OwnableConfig {
   return {
     owner,
     // To ensure that any other overrides aren't applied
-    ownerOverrides: {},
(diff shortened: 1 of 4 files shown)


paulbalaji (approved):

typescript/infra/src/config/environment.ts
```diff
@@ -109,6 +109,8 @@ export function getOwnerConfigForAddress(owner: string): OwnableConfig {
   return {
```
Mo-Hussain, line 109: Thoughts on adding a second optional proxyAdmin argument for this function/

Mo-Hussain (approved):

PR #4774, opened 2024-10-28
feat: script to check versions of default validators

paulbalaji:
### Description

feat: script to check versions of default validators

also prints age of the commit for mismatched validators

### Drive-by changes

update s3 validator to return `metadata_latest.json`

### Related issues

na

### Backward compatibility

ye

### Testing

manual

PR #4896, opened 2024-11-25
feat: add validator operator aliases to default multisig config

paulbalaji:
### Description

feat: add validator operator aliases to default multisig config

tldr:
updated this type and anywhere it's been used
image

- updates `MultisigConfig.validators` type from Address[] to ValidatorConfig[]
- `ValidatorConfig` includes both an address + an alias, to be able to track the operator aliases in-code
- update any usage of `defaultMultisigConfigs` and `MultisigConfig` type
- add `multisigConfigToIsmConfig` helper method to easily convert to `MultisigIsmConfig` format (ie. no aliases)

this will simplify the logic in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4774, and enable us to auto-generate the default ISMs docs page after an SDK update

### Testing

- added extra tests to ensure address/aliases are somewhat sound
- `yarn tsx scripts/check/check-validator-announce.ts -e mainnet3` still looks good


.changeset/four-meals-cheat.md
```diff
@@ -0,0 +1,6 @@
+---
+'@hyperlane-xyz/cli': major
+'@hyperlane-xyz/sdk': major
```
paulbalaji, line 3:
```suggestion
'@hyperlane-xyz/cli': minor
'@hyperlane-xyz/sdk': minor
```

tkporter (approved):

PR #4904, opened 2024-11-27
feat: retry program deployments in SVM tooling with higher fees

tkporter:
### Description

- A common failure case on Solana has been having trouble getting program deployment txs to land. Parts of the program bytecode are written to a buffer in hundreds of different txs, and some of these fail. Now we will:
  - specify a buffer account
  - automatically retry with the same buffer account (building upon any successful writes to it) and with a higher compute unit price

What will probably follow at some point:
- programmatically decide what fees to start using

### Drive-by changes

- Always use existing keys if they are present, removing some code complexity here


ameten (approved):

rust/sealevel/client/src/cmd_utils.rs
```diff
@@ -44,62 +47,91 @@ pub(crate) fn account_exists(client: &RpcClient, account: &Pubkey) -> Result<boo
...
+            println!("Sleeping for 2 seconds to fully allow program to be deployed");
+            sleep(Duration::from_secs(2));
+
+            return Ok(program_id);
+        }
+
+        println!(
+            "Failed to deploy program with compute unit price {}",
+            compute_unit_price
         );
```
daniel-savu, line 119: I think this logic can still live in a separate function (like `deploy_program` before) for readability, even if that function isn't called anywhere else

rust/sealevel/client/src/cmd_utils.rs
```diff
@@ -44,62 +47,91 @@ pub(crate) fn account_exists(client: &RpcClient, account: &Pubkey) -> Result<boo
...
-        println!("Program {} already deployed", program_keypair.pubkey());
-    }
 
-    Ok(())
-}
+        compute_unit_price = if compute_unit_price > 0 {
+            compute_unit_price * 11 / 10
+        } else {
+            1000
+        };
```
daniel-savu, line 125: would move this logic elsewhere and replace magic numbers with some consts or add comments

daniel-savu (approved):
🔥 this will also result in much fewer `program close` commands needing to be run to get back the rent balance and keep retrying

rust/sealevel/client/src/cmd_utils.rs
tkporter: added a comment

rust/sealevel/client/src/cmd_utils.rs
tkporter: good call

PR #4905, opened 2024-11-27
fix: on SVM collateral warp routes, specify the mint when transferring out as readonly

tkporter:
### Description

- This was surfaced a while back but we never made the tweak -- on the collateral warp routes, the account infos exposed to the relayer informing it how to process a message ask for the token mint to be writeable. This isn't necessary - it can be read-only because nothing is changing in the data of the token mint. A consequence of this is that we need to pay higher fees for accounts that have very high contention (like USDC)
- We have strong test coverage here - if this change wasn't possible, our Sealevel tests would fail

### DIFF
--- a/rust/sealevel/programs/hyperlane-sealevel-token-collateral/src/plugin.rs
+++ b/rust/sealevel/programs/hyperlane-sealevel-token-collateral/src/plugin.rs
@@ -444,7 +444,7 @@ impl HyperlaneSealevelTokenPlugin for CollateralPlugin {
             vec![
                 AccountMeta::new_readonly(token.plugin_data.spl_token_program, false).into(),
                 AccountMeta::new_readonly(spl_associated_token_account::id(), false).into(),
-                AccountMeta::new(token.plugin_data.mint, false).into(),
+                AccountMeta::new_readonly(token.plugin_data.mint, false).into(),
                 AccountMeta::new(recipient_associated_token_account, false).into(),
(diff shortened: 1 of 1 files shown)


ameten (approved):

daniel-savu (approved):
amazing

rust/sealevel/programs/hyperlane-sealevel-token-collateral/src/plugin.rs
```diff
@@ -444,7 +444,7 @@ impl HyperlaneSealevelTokenPlugin for CollateralPlugin {
             vec![
                 AccountMeta::new_readonly(token.plugin_data.spl_token_program, false).into(),
                 AccountMeta::new_readonly(spl_associated_token_account::id(), false).into(),
-                AccountMeta::new(token.plugin_data.mint, false).into(),
+                AccountMeta::new_readonly(token.plugin_data.mint, false).into(),
```
daniel-savu, line 447: doesn this apply to synthetic routes as well? If so, worth changing there too even if those haven't caused issues

tkporter: Because a transfer out for a synthetic involves minting the token and increasing the total supply, the mint needs to be writeable :(

https://github.com/solana-labs/solana-program-library/blob/5380b0e7e2f4a479a450c1b7091690ddb312ba73/token/program-2022/src/instruction.rs#L1722-L1736

PR #4913, opened 2024-11-28
feat: track ATA payer balances in warp monitors instead of key funder

tkporter:
### Description

- We need to track balances of "ATA payer" accounts, which are accounts that have balances that are used to pay for rent for recipients that have never received a particular token before. The balance is in the native token as that's what the SVM uses to charge rent. Previously, we were tracking these in the key funder in a very manual and annoying process. Now, this is tracked automatically when a warp route monitor is set up.
- Removed the old logic in the key funder
- Consolidated some code to ensure the same gauge is used to track wallet balances throughout TS

What I'll do after this is merged:
- deploy an updated key funder to no longer track ATA payers
- deploy new warp route monitors with this

Example of the metric now included in the warp monitor:
```
# HELP hyperlane_wallet_balance Current balance of a wallet for a token
# TYPE hyperlane_wallet_balance gauge
hyperlane_wallet_balance{chain="eclipsemainnet",wallet_address="CijxTbPs9JZxTUfo8Hmz2imxzHtKnDFD3kZP3RPy34uJ",wallet_name="SOL/eclipsemainnet-solanamainnet/ata-payer",token_symbol="Native",token_name="Native"} 0.039941128

```

### Drive-by changes

- Some drive-bys to allow you to omit a warp route ID as a CLI arg to some scripts to get an interactive prompt. Found this useful as it's annoying to remember and type out the warp route ID structure

### DIFF
--- a/typescript/infra/scripts/check/check-utils.ts
+++ b/typescript/infra/scripts/check/check-utils.ts
@@ -38,6 +38,7 @@ import { logViolationDetails } from '../../src/utils/violation.js';
 import {
   Modules,
   getArgs as getRootArgs,
+  getWarpRouteIdInteractive,
   withAsDeployer,
   withChains,
   withContext,
@@ -192,7 +193,7 @@ export async function getGovernor(
     governor = new ProxiedRouterGovernor(checker);
   } else if (module === Modules.WARP) {
     if (!warpRouteId) {
-      throw new Error('Warp route id required for warp module');
+      warpRouteId = await getWarpRouteIdInteractive();
     }
     const config = await getWarpConfig(multiProvider, envConfig, warpRouteId);
     const warpAddresses = getWarpAddresses(warpRouteId);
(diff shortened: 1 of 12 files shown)


typescript/infra/scripts/check/check-utils.ts
```diff
@@ -192,7 +193,7 @@ export async function getGovernor(
     governor = new ProxiedRouterGovernor(checker);
   } else if (module === Modules.WARP) {
     if (!warpRouteId) {
-      throw new Error('Warp route id required for warp module');
+      warpRouteId = await getWarpRouteIdInteractive();
```
Mo-Hussain, line 196: nice!

typescript/infra/scripts/warp-routes/monitor/types.ts
```diff
@@ -9,3 +9,10 @@ export interface WarpRouteBalance {
   balance: number;
   valueUSD?: number;
 }
+
+export interface NativeWalletBalance {
+  chain: string;
+  walletAddress: string;
+  walletName: string;
+  balance: number;
```
paulbalaji, line 17:
```suggestion
  chain: ChainName;
  walletAddress: Address;
  walletName: string;
  balance: number;
```

paulbalaji (approved):

PR #4915, opened 2024-11-28
fix: function write_checkpoint did not write checkpoint_{index}_with_id.json files to the folder

IvanPsurtcev:
### Description

Added a mechanism for writing checkpoint files to the checkpoint folder in the write_chekpoint function. Without this change we wrote a lot of files to the root folder, which was not a correct behavior.


rust/main/hyperlane-base/src/types/gcs_storage.rs
```diff
@@ -241,8 +241,9 @@ impl CheckpointSyncer for GcsStorageClient {
         signed_checkpoint: &SignedCheckpointWithMessageId,
     ) -> Result<()> {
         let object_name = Self::get_checkpoint_key(signed_checkpoint.value.index);
+        let full_path = self.object_path(&object_name);
```
daniel-savu, line 244: maybe rename `object_name` (on line 242) to `object_key` and `full_path` to `object_name` to avoid confusion, since the methods below use `object_name` with a different purpose than here

daniel-savu (approved):

daniel-savu:
Thank you!

PR #4917, opened 2024-11-28
feat(warpMonitor): Add validator_names label to warp balance monitor metric

Mo-Hussain:
### Description
- Add validator_names, number_of_validators, validator_threshold labels to warp balance monitor metric
- The intention is to supplement the Value at Risk dashboard with this validator set information
- It may also be feasible to create alerts that consider the value at risk, validator set info and thresholds to warn us if the validators set is not secure enough

### Backward compatibility
- This will create new time series for the `hyperlane_warp_route_token_balance` metric as we are adding new labels which will mean that All warp route diffs, the warp routes and xERC20 related dashboards will switch to using a different time series. Whilst not ideal as the graphs will look weird, the downside is not significant

### DIFF
--- a/typescript/infra/scripts/warp-routes/monitor/metrics.ts
+++ b/typescript/infra/scripts/warp-routes/monitor/metrics.ts
@@ -88,6 +88,10 @@ export function updateTokenBalanceMetrics(
       .filter((chainName) => chainName !== token.chainName)
       .sort()
       .join(','),
+    // we are assuming that all routes are using the defaultISM
+    // we could do a check to verify this however this would exclude routes like Renzo that include the defaultISM as part of an aggregation
+    // the goal of these metrics is to inform us of the Value at risk for routes that rely (or partially rely) on the defaultISM
+    // we are technically overestimating the total value at risk using this approach as we will account for balances that are not secured by the defaultISM
     validator_names: defaultMultisigConfigs[token.chainName].validators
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/warp-routes/monitor/metrics.ts
```diff
@@ -76,6 +88,13 @@ export function updateTokenBalanceMetrics(
       .filter((chainName) => chainName !== token.chainName)
       .sort()
       .join(','),
+    validator_names: defaultMultisigConfigs[token.chainName].validators
```
Mo-Hussain, line 91: We are making the assumption that all warp routes are using the defaultISM

Mo-Hussain: Considered filtering out warp routes that do not use the defaultISM using the deploy config or reading on-chain state. The problem with this is that it would exclude routes like Renzo which has a lot of value that we want to account for as its ISM is an aggregation that includes the defaultISM. Discussed this with @nambrot a while back

nambrot: I primarily worry about value that uses the default ISM exclusively, so i would be ok with excluding warp routes like renzo for now. Obviously ideally, I would like a dashboard that segments these appropriately

PR #4918, opened 2024-11-28
feat: overrides in the relayer to force collateral warp route mints to be read-only

tkporter:
### Description

Some context, and the fix for future warp routes: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4905

This will force messages to certain collateral warp route recipients to have mint accounts as readonly. This is a first step toward requiring lower fees for some txs. 

An example of Helius fees for a USDC process that includes the USDC mint as writeable:
```
{"jsonrpc":"2.0","result":{"priorityFeeLevels":{"min":0.0,"low":292041.0,"medium":2207571.0,"high":10405557.0,"veryHigh":22996019.0,"unsafeMax":2000000000.0}},"id":"1"}
```

vs when it's not writeable:
```
{"jsonrpc":"2.0","result":{"priorityFeeLevels":{"min":0.0,"low":0.0,"medium":1.0,"high":150000.0,"veryHigh":10000000.0,"unsafeMax":2000000000.0}},"id":"1"}
```

The alternative to having this PR would be to:
- keep marking mints as writeable, forcing us to overpay
- or do a program upgrade of the warp routes (unlikely to happen)

### Drive-by changes

- Removed `#[allow(warnings)]` from mailbox.rs and made required changes


rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -96,9 +96,11 @@ mod tests {
         let (addr, mut rx) = setup_test_server();
 
         let client = reqwest::Client::new();
-        let mut message = HyperlaneMessage::default();
-        // Use a random destination domain
-        message.destination = 42;
+        let message = HyperlaneMessage {
+            // Use a random destination domain
+            destination: 42,
+            ..Default::default()
```
paulbalaji, line 102: why does adding `default()` after not override the destination

tkporter: it's dumb rust syntax, i agree intuitively you'd expect it to apply overrides top to bottom but the only way you can do the `..Default::default()` is if it's at the end

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -257,14 +283,27 @@ impl SealevelMailbox {
...
+            .await?;
+
+        if let Some(forced_readonly_account) =
+            RECIPIENT_FORCED_READONLY_ACCOUNTS.get(&recipient_program_id)
+        {
+            for account_meta in account_metas.iter_mut() {
+                if account_meta.pubkey == *forced_readonly_account {
+                    account_meta.is_writable = false;
+                }
+            }
```
daniel-savu, line 282: nit
```suggestion
            account_metas
                .iter_mut()
                .filter(|account_meta| account_meta.pubkey == *forced_readonly_account)
                .for_each(|account_meta| account_meta.is_writable = false);
```

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -95,6 +102,25 @@ const PROCESS_COMPUTE_UNIT_PRICE_MICRO_LAMPORTS: u64 =
...
+    static ref RECIPIENT_FORCED_READONLY_ACCOUNTS: HashMap<Pubkey, Pubkey> = HashMap::from([
+        // EZSOL
+        (pubkey!("b5pMgizA9vrGRt3hVqnU7vUVGBQUnLpwPzcJhG1ucyQ"), pubkey!("ezSoL6fY1PVdJcJsUpe5CM3xkfmy3zoVCABybm5WtiC")),
+        // ORCA
+        (pubkey!("8acihSm2QTGswniKgdgr4JBvJihZ1cakfvbqWCPBLoSp"), pubkey!("orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE")),
+        // USDC
+        (pubkey!("3EpVCPUgyjq2MfGeCttyey6bs5zya5wjYZ2BE6yDg6bm"), pubkey!("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v")),
+        // USDT
+        (pubkey!("Bk79wMjvpPCh5iQcCEjPWFcG1V2TfgdwaBsWBEYFYSNU"), pubkey!("Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB")),
+        // USDT
```
daniel-savu, line 95: nit
```suggestion
        // WIF
```

daniel-savu (approved):

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
tkporter: 🦀 nice

PR #4922, opened 2024-11-29
feat: Multi EVM Chain Signers

mshojaei-txfusion:
### Description
This PR modifies the warp init command to read the multiProtocolSigner from the signer middleware instead of creating a new instance. 

### Related issues

-  https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4910
-  Fixes issue with duplicate signer initialization in warp route configuration
- Improves consistency with other commands that use signer middleware

### Backward compatibility

Yes - This change is backward compatible as it maintains the same functionality while making the multiProtocolSigner parameter optional.

### Testing

Manual testing


typescript/cli/src/context/context.ts
```diff
@@ -66,19 +109,24 @@ export async function getContext({
 }: ContextSettings): Promise<CommandContext> {
   const registry = getRegistry(registryUri, registryOverrideUri, !disableProxy);
 
-  let signer: ethers.Wallet | undefined = undefined;
-  if (key || requiresKey) {
+  //Just for backward compability
```
paulbalaji, line 112:
```suggestion
  //Just for backward compatibility
```

typescript/cli/src/context/types.ts
```diff
@@ -24,7 +26,10 @@ export interface CommandContext {
   multiProvider: MultiProvider;
   skipConfirmation: boolean;
   key?: string;
-  signer?: ethers.Signer;
+  // just for evm chains backward compability
```
paulbalaji, line 29:
```suggestion
  // just for evm chains backward compatibility
```

paulbalaji (commented):
quick one to make codespell happy

mshojaei-txfusion:
> quick one to make codespell happy

fixed, thanks

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -0,0 +1,249 @@
...
+        chain: argv.chain,
+        config,
+        addresses: coreAddresses,
+      });
+
+      const transactions = await evmCoreModule.update(config);
+
+      return Array.from(new Set(transactions.map((tx) => tx.chainId))).map(
+        (chainId) => argv.context.multiProvider.getChainName(chainId),
+      );
```
xeno097, line 214: You can simplify this by directly using the keys of the `interchainAccountRouter.remoteIcaRouters` field if defined in the config instead of having to run the `core apply` logic here too, because depending on the on-chain config the parsing can take a while. The keys of the `remoteIcaRouters` field are domainIds that can be provided as input to the `getChainName` method of the multiprovider.

mshojaei-txfusion: This wont work, because there are cases that you want apply a change to have no key on `remoteIcaRouters `, but there are keys on chain.
long story short this test case will fail: 
`should unenroll a remote ICA Router and update the config on all involved chains`

xeno097 (approved):

PR #4923, opened 2024-11-29
feat: Sealevel IGP quoting in WarpCore / token adapters

tkporter:
### Description

- Implements SVM IGP quoting. Unfortunately the warp routes themselves don't have direct getters for this, so we first need to get the expected destination gas amount, and then do a call to the IGP
- An unfortunate consequence is that performing a "view call" on Sealevel (i.e. a simulateTransaction call) requires you to specify a payer that is capable of paying for the tx as if it were real (i.e. must have funds, must be an EOA). I couldn't find any workaround other than just requiring the payer to be passed in :(

### Drive-by changes

- Some further support for non-overhead IGP usage / some minor renaming to more closely match the program jargon

### Testing

I tested this with an infra script that was calling the functions against live warp routes. Not sure what the unit test practices we have here are though, seems like we don't have any


typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts
```diff
@@ -1,40 +1,123 @@
...
+      // ignore the recent blockhash we pass in, and have the node use its latest one
+      replaceRecentBlockhash: true,
+      // ignore signature verification
+      sigVerify: false,
+    });
 
-    const accountInfo = await connection.getAccountInfo(new PublicKey(address));
-    if (!accountInfo) throw new Error(`No account info found for ${address}}`);
+    const base64Data = simulationResponse.value.returnData?.data?.[0];
+    if (base64Data === undefined) {
```
jmrossy, line 107:
```suggestion
   assert(base64Data, 'No return data when quoting gas payment, may happen if the payer has insufficient funds')
```

typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts
```diff
@@ -56,3 +139,62 @@ export class SealevelOverheadIgpAdapter extends BaseSealevelAdapter {
...
+
+    this.overheadIgp = new PublicKey(addresses.overheadIgp);
+  }
+
+  async getAccountInfo(): Promise<SealevelOverheadIgpData> {
+    const address = this.addresses.overheadIgp;
+    const connection = this.getProvider();
+
+    const accountInfo = await connection.getAccountInfo(new PublicKey(address));
+    if (!accountInfo) throw new Error(`No account info found for ${address}}`);
```
jmrossy, line 182: ditto on assert, though I'm def nitpicking

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -283,11 +288,34 @@ export abstract class SealevelHypTokenAdapter
...
   async quoteTransferRemoteGas(
-    _destination: Domain,
+    destination: Domain,
+    sender?: Address,
   ): Promise<InterchainGasQuote> {
-    // TODO Solana support
-    return { amount: 0n };
+    const tokenData = await this.getTokenAccountData();
+    const destinationGas = tokenData.destination_gas?.get(destination);
+    if (destinationGas === undefined) {
```
jmrossy, line 299: consider `isNullish` or just `!`. Sorry to keep nitpicking on these but it's good for JS code to be defensive against more falsy values than just undefined

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -283,11 +288,34 @@ export abstract class SealevelHypTokenAdapter
...
+      return { amount: 0n };
+    }
+
+    const igp = this.getIgpAdapter(tokenData);
+    if (!igp) {
+      return { amount: 0n };
+    }
+
+    if (sender === undefined) {
```
jmrossy, line 308: assert

jmrossy (approved):
Nice work here. Good to see this mostly fitting neatly into the existing abstractions. RE testing, consider adding a case to WarpCore.test.ts but I don't consider that blocking.

Also look like the linter found a few nitpicks

PR #4925, opened 2024-11-30
chore: update seaorm to latest

### LINKED ISSUE #4793: chore: Upgrade SeaORM to the latest version
## Problem

Currently we are using obsolete version of SeaORM, namely `0.11.1` and we generate entities using version `0.12.3`. It produces different source code for the entities which a developer has to remove when the developer changes the migration scripts.

## Solution

Upgrade SeaORM to the latest version everywhere:
1. To be used for migrations
2. To be used for generating entities
3. To be used by Scraper to interact with Postgres DB

### Tasks
1. Upgrade SeaORM in `main/Cargo.toml` (`sea-orm` and `sea-orm-migration`).
2. Upgrade SeaORM in `generate_entities.rs` and regenerate entities. Make sure that new datatypes can handle existing data without issues.
4. Make the code compilable again.
5. Remove SeaORM from `sealevel/Cargo.toml`.

## Testing

1. Run Scraper from mainstream branch with current SeaORM version and populate database with data from some blockchain
6. Switch to the feature branch with the latest SeaORM version.
7. Run Scraper from feature brach and confirm that all the data is populated in the same format as before

Mantas-M:
### Description

Updated `sea-orm` and `sea-orm-migration` to `1.1.1` (latest as of opening this PR) and resolved any issues that occurred due to the version bump. 

### Drive-by changes

- Removed `sea-orm` from `rust/sealevel`
- Updated `sea-orm-migration` data types
- Bumped up the `rust-toolchain` to `1.81.0` due to `sea-orm-cli` issue: https://github.com/servo/rust-url/issues/992 

### Related issues

- Resolves #4793

### Backward compatibility

Yes

### Testing

Manual, as described in #4793


rust/main/agents/scraper/src/db/generated/prelude.rs
```diff
@@ -1,4 +1,4 @@
-//! `SeaORM` Entity. Generated by sea-orm-codegen 0.12.3
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.1
 #[allow(unused_imports)]
 pub use super::{
     block::Entity as Block, cursor::Entity as Cursor,
```
ameten, line 4: I suppose we can commit prelude which was generated by `generate_entities`.

rust/main/agents/scraper/migration/bin/init_db.rs
```diff
@@ -4,7 +4,7 @@ use common::*;
 
 mod common;
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), DbErr> {
```
ameten, line 8: I wonder if it produces the same database schema as before (i.e. old version vs new version)?

ameten:
And it will be great to check if newly generated entities can work with old schema without issue. I can see that we have replaced `BigDecimal` with `Decimal` in some cases.

rust/main/agents/scraper/migration/bin/init_db.rs
Mantas-M: It appears so, I did the test which you mentioned in the issue of the description, which is to generate a schema on the old version, scrape a bit, then switch to the new version and continue scraping, everything worked fine

rust/main/agents/scraper/src/conversions.rs
```diff
@@ -1,10 +1,13 @@
...
 
 use hyperlane_core::U256;
 
-pub fn u256_to_decimal(v: U256) -> BigDecimal {
+pub fn u256_to_decimal(v: U256) -> Decimal {
     let mut buf = [0u8; 32];
     v.to_little_endian(&mut buf);
-    BigDecimal::from(BigInt::from_bytes_le(Sign::Plus, &buf as &[u8]))
+    let big_dec = &BigDecimal::from(BigInt::from_bytes_le(Sign::Plus, &buf as &[u8])).to_string();
+    Decimal::from_str(big_dec).unwrap()
```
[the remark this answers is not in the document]
Mantas-M: Decimal and BigDecimal are both mapped to the same type on PostgreSQL, I was wondering why the change in that too but if you look at the [definition of `Wei` (which uses BigDecimal)](https://github.com/Mantas-M/hyperlane-monorepo/blob/chore/update-seaorm/rust/main/agents/scraper/migration/src/l20230309_types.rs) that has not changed from the old version.

rust/main/agents/scraper/src/db/generated/prelude.rs
Mantas-M: I am not sure what you mean by `commit` could you please specify?

Mantas-M:
> And it will be great to check if newly generated entities can work with old schema without issue. I can see that we have replaced `BigDecimal` with `Decimal` in some cases.

I checked it during the manual tests, scraping with the old/new versions, it was working fine ✅

Mantas-M:
Implementation details of Decimal vs BigDecimal
https://github.com/SeaQL/sea-orm/blob/master/src/driver/sqlx_postgres.rs#L516

rust/main/agents/scraper/src/db/generated/prelude.rs
ameten: If you run `generate_entities`, it will generate entities files. They will be the same as currently committed in this PR with the exception of `prelude.rs` file. I think that `prelude.rs` can be also added to this PR. Current one and the generate one are equivalent.

Mantas-M: Ahhh, I get it now, the new prelude is added in this PR. It is just that the generated one has a line by line format, not a block format, I changed it back to the block format so we can use the `allow_unused_imports` on the whole block as before. You can see the update I did here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4925/commits/ed6d556bf56f700fb536bcc4045f7cc1fd168176

But you are right, the generated one is the same as the old one so everything is fine in that regard

rust/main/agents/scraper/migration/bin/init_db.rs
ameten: It is good that it works. It is not the test for this comment though (but a good test for [that comment](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4925#issuecomment-2517970903))

The test is mostly manual for this:
1. Start Postgres on one port.
2. Init schema there (`init_db.rs`) from `main`.
3. Start Postgres on a different port.
4. Init schema from this branch.
5. Compare the tables and field types.

rust/main/agents/scraper/src/conversions.rs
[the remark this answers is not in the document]
Mantas-M: Another note on this, the BigDecimal -> Decimal was a change which was made by the entity generation of the updated `sea-orm` package, and has nothing do with any specific changes made to the schema.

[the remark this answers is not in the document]
ameten: I suppose that the definition of column type `Wei` did not change since it is specified in the native Postgres type: `decimal`, which is equivalent of `numeric` ("[The types decimal and numeric are equivalent](https://www.postgresql.org/docs/current/datatype-numeric.html)"). So, `Wei` defines the column type, not the type which is going to be used in Rust struct which is mapped to the record in the table.

I don't believe that Rust `Decimal` can contain all values possible in `U256`. Type `Decimal` supports only 96-bit integers. In contrast, `BigDecimal` internally contains a vector of digits (see `BigUint` type).

I wonder if it is possible to enable feature `with-bigdecimal` on Postgres driver?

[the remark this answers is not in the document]
Mantas-M: Not from what I found though going through code and docs.

Their discord is pretty dead, but it's better than nothing so I've tried to [ask there](https://discord.com/channels/873880840487206962/1315689749276790844/1315689749276790844).

In general, I doubt you would be storing gas values that overflow the decimal type in rust, perhaps one thing we could do is swap it to bigint 🤔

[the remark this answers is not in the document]
Mantas-M: Seems like we're not going to have an answer to this any time soon, so I am guessing this PR is on pause until I either burn more time trying to get it back to `BigDecimal` or we can agree on a different direction to take.

rust/main/agents/scraper/migration/bin/init_db.rs
ameten: Great! So, database schema is the same.

rust/main/agents/scraper/src/db/generated/prelude.rs
ameten: I see it now. We need to add `allow_unused_imports` anyway and it gets removed by generation script. So, this file will change no matter what. Block imports are fine then.

rust/main/agents/scraper/src/conversions.rs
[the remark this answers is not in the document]
ameten: It looks like `BigDecimal` is not supported by SeaORM at the moment. I found this issue: https://github.com/SeaQL/sea-orm/issues/1530.

We have two options to proceed:
1. Contribute code into SeaORM so that `BigDecimal` is fully supported by generation script.
2. Manually add `BigDecimal` into the generated code with adding required dependencies with an elaborated comment why we did this.

I think 2. should be fairly feasible and acceptable result.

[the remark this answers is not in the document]
Mantas-M: Nice find, I've gone ahead and implemented the second option, and ran through the tests to make sure it's working as expected.

ameten (approved):

Mantas-M:
🎉🎉🎉

ameten:
Hey @Mantas-M, I had to revert this change in this PR - https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5030.

Scraper started to report that many queries are running for a long time. I'll check if the upgrade of SeaORM is responsible and if not, will bring it back.

PR #4936, opened 2024-12-02
feat: require Sealevel native transfers to cover the rent of the recipient

tkporter:
### Description

- Happened a few times where someone transferred SOL from Eclipse -> SOL on Solana, and the recipient did not yet exist (so had no SOL balance to cover its rent). The amount on these transfers would be insufficient to cover the rent of the recipient, so the transfer could never be delivered. See https://discord.com/channels/935678348330434570/1311371339889639486/1311381212828532858 for context on the first time this happened. There were then a few other instances the day after
- This introduces `ITokenAdapter.getMinimumTransferAmount`, where impls generally return a minimum of 0, but native Sealevel and native Sealevel warp routes have some logic that is aware of required rent exemption balances
- Atm, in WarpCore this only checks getMinimumTransferAmount on the destination side and not the origin -- my reasoning here is that:
  a. for the specific issue this is addressing, the destination side is the only one where issues can arise. On the origin, the account that escrows the SOL is always rent exempt
  b. checking on the origin for completeness could be done, but would require some extra logic to be aware of where tokens are escrowed on the origin side (i.e. the recipient should be the account that escrows the SOL), which felt unnecessary / confusing

### DIFF
--- a/typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
+++ b/typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
@@ -636,6 +636,10 @@ export class SealevelHypNativeAdapter extends SealevelHypTokenAdapter {
     return this.wrappedNative.getMetadata();
   }
 
+  override async getMinimumTransferAmount(recipient: Address): Promise<bigint> {
+    return this.wrappedNative.getMinimumTransferAmount(recipient);
+  }
+
   override async getMedianPriorityFee(): Promise<number | undefined> {
     // Native tokens don't have a collateral address, so we don't fetch
     // prioritization fee history
--- a/typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
+++ b/typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
@@ -173,7 +173,7 @@ export class SealevelTokenAdapter
     return { decimals: 9, symbol: 'SPL', name: 'SPL Token', totalSupply: '' };
   }
 
-  async getMinimumTransferAmount(recipient: Address): Promise<bigint> {
+  async getMinimumTransferAmount(_recipient: Address): Promise<bigint> {
     return 0n;
   }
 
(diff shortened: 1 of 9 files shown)


typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -643,13 +647,54 @@ export class WarpCore {
...
+
+    const destinationName = this.multiProvider.getChainName(destination);
+    const destinationToken =
+      originToken.getConnectionForChain(destinationName)?.token;
+    assert(destinationToken, `No connection found for ${destinationName}`);
+    const destinationAdapter = destinationToken.getAdapter(this.multiProvider);
+
+    // Convert the originTokenAmount to a destination amount
+    const destinationTokenAmount = destinationToken.amount(
+      convertDecimals(
```
jmrossy, line 672: It's nbd but If you convert `minDestinationTransferAmount` to originToken's decimals and _before_ comparing it the origin amount, I think you can avoid the need for 2 `convertDecimals` calls. They're not expensive or anything but it's kinda verbose.

jmrossy (approved):
Nice fix!

typescript/sdk/src/warp/WarpCore.ts
tkporter: clever, fixed!

typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts
```diff
@@ -90,6 +90,24 @@ export class SealevelNativeTokenAdapter
...
+    const recipientLamports = recipientAccount?.lamports ?? 0;
+
+    const minRequiredLamports =
+      await provider.getMinimumBalanceForRentExemption(recipientDataLength);
+
+    if (recipientLamports < minRequiredLamports) {
+      return BigInt(minRequiredLamports - recipientLamports);
+    }
+
+    return 0n;
```
yorhodes, line 108: should this be `minRequiredLamports`?

PR #4942, opened 2024-12-04
feat: Retrieve gas payments by sequence for scraping Sealevel 

ameten:
### Description

- Retrieve gas payment by origin, interchain_gas_paymaster and sequence
- Add Sealevel to Scraper in E2E Ethereum and Sealevel tests

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4271

### Backward compatibility

Yes (after backfill of origin, destination, interchain_gas_paymaster)

### Testing

E2E Ethereum and Sealevel tests (including scraping Sealevel)

### DIFF
--- a/rust/main/agents/scraper/migration/src/m20230309_000004_create_table_gas_payment.rs
+++ b/rust/main/agents/scraper/migration/src/m20230309_000004_create_table_gas_payment.rs
@@ -1,5 +1,3 @@
-use std::borrow::BorrowMut as _;
-
 use sea_orm::ConnectionTrait;
 use sea_orm_migration::prelude::*;
 
@@ -41,11 +39,15 @@ impl MigrationTrait for Migration {
                             .big_unsigned()
                             .not_null(),
                     )
-                    .col(ColumnDef::new(GasPayment::Origin).unsigned())
-                    .col(ColumnDef::new(GasPayment::Destination).unsigned())
+                    .col(ColumnDef::new(GasPayment::Origin).unsigned().not_null())
+                    .col(
+                        ColumnDef::new(GasPayment::Destination)
+                            .unsigned()
+                            .not_null(),
+                    )
                     .col(
                         ColumnDef::new_with_type(GasPayment::InterchainGasPaymaster, Address)
-                            .borrow_mut(),
+                            .not_null(),
                     )
                     .col(ColumnDef::new(GasPayment::Sequence).big_integer())
                     .foreign_key(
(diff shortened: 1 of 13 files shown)


rust/main/utils/run-locally/src/invariants/termination_invariants.rs
```diff
@@ -126,7 +126,7 @@ pub fn termination_invariants_met(
         "Verbose logs not expected at the log level set in e2e"
     );
 
-    let gas_payment_sealevel_events_count = fetch_metric(
+    let _gas_payment_sealevel_events_count = fetch_metric(
```
daniel-savu, line 129: you should be able to remove this altogether now, since it's not used

rust/main/agents/scraper/src/db/generated/gas_payment.rs
```diff
@@ -90,10 +87,6 @@ impl ColumnTrait for Column {
 impl RelationTrait for Relation {
     fn def(&self) -> RelationDef {
         match self {
-            Self::Destination => Entity::belongs_to(super::domain::Entity)
-                .from(Column::Destination)
-                .to(super::domain::Column::Id)
-                .into(),
             Self::Domain => Entity::belongs_to(super::domain::Entity)
                 .from(Column::Domain)
                 .to(super::domain::Column::Id)
```
daniel-savu, line 92: I know we were planning on deprecating `Domain` but I'm not sure why `Destination` is being removed from here

rust/main/agents/scraper/migration/src/m20230309_000004_create_table_gas_payment.rs
```diff
@@ -90,6 +87,19 @@ impl MigrationTrait for Migration {
...
             .await?;
 
+        manager
+            .create_index(
+                Index::create()
+                    .table(GasPayment::Table)
+                    .name("gas_payment_origin_interchain_gas_paymaster_sequence_idx")
+                    .col(GasPayment::Origin)
+                    .col(GasPayment::InterchainGasPaymaster)
+                    .col(GasPayment::Sequence)
```
daniel-savu, line 97: if the message has no `Sequence` would the index still be effective? Maybe there's some other column we could use if not

daniel-savu (approved):

rust/main/agents/scraper/src/db/generated/gas_payment.rs
ameten: It looks like we are getting gas payment which pay for messages in chain we don't support. Such chains are not mentioned in `domain` table. So, we cannot apply a foreign key constrained to `destination` field.

rust/main/agents/scraper/migration/src/m20230309_000004_create_table_gas_payment.rs
ameten: I will use the index to find gas payments by sequence for sequence-aware indexing. If a record does not have a sequence, it means that watermark-aware indexing is used and we don't need to use this index.

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
ameten: Fixed

PR #4945, opened 2024-12-04
fix: relay message from tx containing multiple dispatches

paulbalaji:
### Description

fix: relay message from tx containing multiple dispatches

without this change, it would automatically just assume you wanted to relay the 1st message (event index 0) from the transaction

### DIFF
--- a/typescript/cli/src/status/message.ts
+++ b/typescript/cli/src/status/message.ts
@@ -94,7 +94,11 @@ export async function checkMessageStatus({
     const merkleAddress = chainAddresses[origin].merkleTreeHook;
     stubMerkleTreeConfig(relayer, origin, hookAddress, merkleAddress);
 
-    deliveredTx = await relayer.relayMessage(dispatchedReceipt);
+    deliveredTx = await relayer.relayMessage(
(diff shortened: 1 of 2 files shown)


typescript/cli/src/status/message.ts
```diff
@@ -94,7 +94,11 @@ export async function checkMessageStatus({
     const merkleAddress = chainAddresses[origin].merkleTreeHook;
     stubMerkleTreeConfig(relayer, origin, hookAddress, merkleAddress);
 
-    deliveredTx = await relayer.relayMessage(dispatchedReceipt);
+    deliveredTx = await relayer.relayMessage(
+      dispatchedReceipt,
+      undefined,
```
daniel-savu, line 99: I'm surprised setting this to `undefined` works, given the arg has type `number`

xeno097 (approved):

typescript/cli/src/status/message.ts
paulbalaji: haha same, but i'm just copying what we did in a different part of the code

PR #4957, opened 2024-12-05
feat: unenroll warp routes

ltyu:
### Description
Add feature to unenroll warp routes using `warp apply`

### Drive-by changes
Update enroll warp route to use `difference()` similar to ICA enrollment

### Backward compatibility
Yes

### Testing
Manual/Unit Tests


typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -148,46 +152,84 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+          routesToEnroll.map((k) => Number(k)),
+          routesToEnroll.map((a) =>
+            addressToBytes32(expectedRemoteRouters[a].address),
+          ),
+        ],
+      ),
+    });
+
+    return updateTransactions;
+  }
```
xeno097, line 190: Shouldn't we also add txs to remove/add the destination gas (in another method)?

ltyu: that should be triggered by update to config.destinationGas. functionally, unenrolling will also "remove" the destinationGas from warp read.

xeno097 (approved):

PR #4966, opened 2024-12-07
chore: Combine schema and types files

jmrossy:
### Description

- Combine remaining schema and types files
- De-dupe some Ownable types
- Export more schemas from SDK

### Backward compatibility

Yes


xeno097 (approved):
🔥

PR #4975, opened 2024-12-09
feat: support safe tx parsing for zksync chains

paulbalaji:
### Description

- correctly generate core config for zksync chains
	- copied from `pb/zksync` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/pb/zksync/typescript/infra/config/environments/mainnet3/core.ts
	- note the storage aggregation ism was not around when doing the first zksync chain deploys, and we haven't gone back and updated the config generation yet to include this
- uncomment swell ICA address so tx parser knows it's an ICA
- update ISM reader to return storagemultisigism types on zksync
	- since static ISMs are not supported on zksync


paulbalaji:
replaced by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5042

PR #4976, opened 2024-12-09
fix: suppress help on CLI failures

yorhodes:
### Description

No longer outputs help dialog on any error

### DIFF
--- a/typescript/cli/cli.ts
+++ b/typescript/cli/cli.ts
@@ -32,9 +32,6 @@ import { configureLogger, errorRed } from './src/logger.js';
 import { checkVersion } from './src/utils/version-check.js';
 import { VERSION } from './src/version.js';
 
-// From yargs code:
-const MISSING_PARAMS_ERROR = 'Not enough non-option arguments';
-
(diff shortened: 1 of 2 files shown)


typescript/cli/cli.ts
```diff
@@ -73,14 +70,7 @@ try {
     .demandCommand()
     .strict()
     .help()
-    .fail((msg, err, yargs) => {
-      if (msg && !msg.includes(MISSING_PARAMS_ERROR)) errorRed('Error: ' + msg);
-      console.log('');
-      yargs.showHelp();
-      console.log('');
-      if (err) errorRed(err.toString());
```
jmrossy, line 81: Do we definitely want to cut this line? Can a command fail without throwing an error?

yorhodes: https://yargs.js.org/docs/#api-reference-failfn-boolean
https://yargs.js.org/docs/#api-reference-showhelponfailenable-message
customizing the fail handler seems unnecessary and I dont really understand the conditions in here

yorhodes: atm the help string is logged whether or not the error is related to a missing argument

jmrossy (approved):

PR #4980, opened 2024-12-10
feat: ica submission strategy support in the cli

xeno097:
### Description

This PR updates the implementation for sending ICA transactions using the CLI when running `warp apply` and adds a `relay` flag so that ICA transactions can be self-relayed when using the JSON RPC strategy as the internal submission strategy.

### Drive-by changes

 - Removes the transformer logic from the submitter getter

### Related issues

- #4979 

### Backward compatibility

- No, the way to define how to send ICA transactions has changed compared to previous CLI versions

### Testing

- Manual
- E2E


typescript/sdk/src/middleware/account/InterchainAccount.ts
```diff
@@ -222,16 +225,38 @@ export class InterchainAccount extends RouterApp<InterchainAccountFactories> {
...
+  const knownDomains = await currentIca.domains();
+
+  const remoteIcaAddresses = await Promise.all(
+    knownDomains.map((domainId) =>
+      currentIca
+        .routers(domainId)
+        .then((address) => [
+          multiProvider.getChainName(domainId),
+          { interchainAccountRouter: bytes32ToAddress(address) },
+        ]),
```
yorhodes, line 251: nit: avoid callback syntax and use async/await

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
```diff
@@ -39,3 +40,43 @@ export const EV5ImpersonatedAccountTxSubmitterPropsSchema =
...
+      .object({
+        type: z.literal(TxSubmitterType.IMPERSONATED_ACCOUNT),
+      })
+      .merge(
+        EV5ImpersonatedAccountTxSubmitterPropsSchema.omit({ chain: true }),
+      ),
+  ])
+  .default({
+    type: TxSubmitterType.JSON_RPC,
+  });
```
yorhodes, line 72: instead of duplicating the union here, can we just make this a "recursive" definition and have internalSubmitter be the top level submitter schema?

yorhodes: otherwise this will be come stale when new submitters are added

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,203 @@
...
+
+type SubmitterFactoryMapping<
+  E extends TxSubmitterType,
+  TConfig extends { type: E },
+  TResult,
+> = {
+  [K in E]: (
+    config: Extract<TConfig, { type: K }>,
+  ) => Promise<TResult> | TResult;
+};
```
yorhodes, line 48: im a bit confused by the need for all of these

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
yorhodes, line 1: I would expect to see a similar negative diff for removing the ICA transformer?

typescript/sdk/src/providers/transactions/transformer/ethersV5/EV5InterchainAccountTxTransformer.ts
yorhodes, line 1: shouldnt this transformer directory be removed entirely?

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,203 @@
...
+  ProtocolTypedReceipt,
+} from '../../ProviderType.js';
+import { CallData } from '../types.js';
+
+import { TxSubmitterInterface } from './TxSubmitterInterface.js';
+import { TxSubmitterType } from './TxSubmitterTypes.js';
+import { EV5GnosisSafeTxBuilder } from './ethersV5/EV5GnosisSafeTxBuilder.js';
+import { EV5GnosisSafeTxSubmitter } from './ethersV5/EV5GnosisSafeTxSubmitter.js';
+import { EV5ImpersonatedAccountTxSubmitter } from './ethersV5/EV5ImpersonatedAccountTxSubmitter.js';
+import { EV5JsonRpcTxSubmitter } from './ethersV5/EV5JsonRpcTxSubmitter.js';
```
yorhodes, line 22: I really think the ICA submitter should not need to know about these other submitters

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,203 @@
...
+  const internalSubmitterFactory = internalSubmitterMap[config.type];
+  // Sanity check
+  if (!internalSubmitterFactory) {
+    throw new Error(
+      `Internal submitter factory not found for type: ${config.type}`,
+    );
+  }
+
+  return internalSubmitterFactory(config as any);
+}
```
yorhodes, line 95: ditto here
this doesnt seem like it should be necessary for the ICA submitter to implement

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,203 @@
...
+
+export class EvmIcaTxSubmitter
+  implements TxSubmitterInterface<ProtocolType.Ethereum>
+{
+  readonly txSubmitterType: TxSubmitterType =
+    TxSubmitterType.INTERCHAIN_ACCOUNT;
+
+  private constructor(
```
yorhodes, line 103: why make this private?

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,203 @@
...
+    multiProvider: MultiProvider,
+  ): Promise<EvmIcaTxSubmitter> {
+    const internalSubmitter = await getInternalSubmitter(
+      config.chain,
+      multiProvider,
+      config.internalSubmitter,
+    );
+
+    const owner =
+      config.owner ?? (await multiProvider.getSignerAddress(config.chain));
```
yorhodes, line 121: this seems a little dangerous
I would be in favor of making the origin chain and owner address required fields to avoid potential mixups like the one we had

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
```diff
@@ -39,3 +40,43 @@ export const EV5ImpersonatedAccountTxSubmitterPropsSchema =
...
+
+export const EvmIcaTxSubmitterPropsSchema = z.object({
+  chain: ZChainName,
+  owner: ZHash.optional(),
+  destinationChain: ZChainName,
```
yorhodes, line 77: I am actually a bit confused by the relationship between the chains/owner and internal submitter
if the internal submitter is a safe submitter then the owner address would have to match? should we remove the chain and owner from this schema entirely?

yorhodes (commented):
Just reviewed the SDK changes so far. I have come around to eliminating the transformer so lets do it wholesale rather than piecemeal.

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
xeno097: oky I'll make it required. I agree that it makes it less error-prone and clearer at the same time.

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
xeno097: The chain field specifies which chain the ICA transactions will be sent from, while the owner address is the owner of the ICA so yes in case the internal submitter is a safe, the addresses should match. I think I can enforce this requirement in the zod validator.

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
xeno097: Oky, should we then add a limit on the nesting it can be done when creating the strategy config? I created the subtype to make it more difficult to create an invalid/overly complex config

typescript/sdk/src/middleware/account/InterchainAccount.ts
```diff
@@ -227,16 +230,38 @@ export class InterchainAccount extends RouterApp<InterchainAccountFactories> {
...
   }
+
+  const currentIca = InterchainAccountRouter__factory.connect(
+    config.localRouter,
+    multiProvider.getSigner(chain),
+  );
+
+  const knownDomains = await currentIca.domains();
+
+  const remoteIcaAddresses = await Promise.all(
```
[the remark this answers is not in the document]
xeno097: because i was young and inexperienced in the jedi ways of hyperlane, i'll revist this PR and update accordingly

typescript/sdk/src/providers/transactions/submitter/builder/TxSubmitterBuilder.ts
```diff
@@ -38,10 +35,7 @@ export class TxSubmitterBuilder<TProtocol extends ProtocolType>
     module: 'submitter-builder',
   });
 
-  constructor(
-    private currentSubmitter: TxSubmitterInterface<TProtocol>,
-    private currentTransformers: TxTransformerInterface<TProtocol>[] = [],
```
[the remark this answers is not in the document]
ltyu: nvm, i see the reason in the Design Docs

typescript/sdk/src/providers/transactions/submitter/IcaTxSubmitter.ts
```diff
@@ -0,0 +1,139 @@
...
+    | ProtocolTypedReceipt<ProtocolType.Ethereum>['receipt'][]
+  > {
+    if (txs.length === 0) {
+      return [];
+    }
+
+    const transactionChains = new Set(txs.map((tx) => tx.chainId));
+    if (transactionChains.size !== 1) {
+      throw new Error(
+        'ICA transactions should have all the same destination chain',
```
ltyu, line 93: isnt this the origin chain according to the example

xeno097: Noup, because they transactions without the ica submitter would have been sent from the chain initially specified in each one. Imagine this:
when you need to perform an update to a warp route token on sepolia all the update transaction will have as destination chain sepolia, but if you send those transactions with an ica from those transactions will be sent from another chain (basesepolia for examaple), but the intended destination chain to receive those transactions will always be sepolia

xeno097: If you see below, after asserting that all the transactions are directed to the same chain we compare the destination chain name of the txs with the destination chain name configured on this ica submitter

typescript/sdk/src/providers/transactions/submitter/submitterBuilderGetter.ts
```diff
@@ -0,0 +1,86 @@
...
+  return new TxSubmitterBuilder<TProtocol>(submitter);
+}
+
+export async function getSubmitter<TProtocol extends ProtocolType>(
+  multiProvider: MultiProvider,
+  submitterMetadata: SubmitterMetadata,
+  registry: IRegistry,
+): Promise<TxSubmitterInterface<TProtocol>> {
+  let interchainAccountRouterAddress: Address | undefined;
+  if (submitterMetadata.type === TxSubmitterType.INTERCHAIN_ACCOUNT) {
```
ltyu, line 43: can this be part of the switch, or better yet, EvmIcaTxSubmitter.fromConfig

typescript/cli/examples/submit/strategy/gnosis-tx-builder-ica-strategy.yaml
```diff
@@ -0,0 +1,14 @@
+polygonamoy:
+  submitter:
+    type: interchainAccount
+    # The origin chain for the ICA transactions
+    chain: basesepolia
+    # The destination chain for the transactions
+    destinationChain: polygonamoy
+    # The owner of the remote ICA account
+    owner: '0x32e02dDB233958834fcB5E93b11E254F8D1EAd1b'
+    internalSubmitter:
```
ltyu, line 10: first time reading this and with admittedly weak ICA experience, this submitter + internalSubmitter relationship is confusing to me. 

i would add a comment here on it's intention

ltyu (approved):

PR #4987, opened 2024-12-11
fix(cli): fix signer init strategy

xeno097:
### Description

Probably fixes an issue due to a recent pr merge that changes how signers are initialized based on the command to be executed

### Drive-by changes

- Updates the key info message log level from info to debug to avoid the following
image

### Backward compatibility

- Yes

### Testing

- Manual

### DIFF
--- a/typescript/cli/src/context/strategies/chain/ChainResolverFactory.ts
+++ b/typescript/cli/src/context/strategies/chain/ChainResolverFactory.ts
@@ -1,7 +1,6 @@
 import { CommandType } from '../../../commands/signCommands.js';
 
 import { MultiChainResolver } from './MultiChainResolver.js';
-import { SingleChainResolver } from './SingleChainResolver.js';
 import { ChainResolver } from './types.js';
 
 /**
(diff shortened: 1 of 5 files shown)


typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -58,9 +59,9 @@ export class MultiChainResolver implements ChainResolver {
         return this.resolveRelayerChains(argv);
       case ChainSelectionMode.CORE_APPLY:
         return this.resolveCoreApplyChains(argv);
-      case ChainSelectionMode.ORIGIN_DESTINATION:
+      case ChainSelectionMode.DEFAULT:
       default:
-        return this.resolveOriginDestinationChains(argv);
+        return this.resolveRelayerChains(argv);
```
paulbalaji, line 64: do we still need the RELAYER enum if the default is that behaviour?

xeno097: makes sense going to remove it

typescript/cli/src/context/strategies/chain/ChainResolverFactory.ts
```diff
@@ -11,13 +10,16 @@ import { ChainResolver } from './types.js';
 export class ChainResolverFactory {
   private static strategyMap: Map<CommandType, () => ChainResolver> = new Map([
     [CommandType.WARP_DEPLOY, () => MultiChainResolver.forWarpRouteConfig()],
-    [CommandType.WARP_SEND, () => MultiChainResolver.forOriginDestination()],
+    // Using the forRelayer resolver because warp send allows the user to self relay the tx
+    [CommandType.WARP_SEND, () => MultiChainResolver.forRelayer()],
```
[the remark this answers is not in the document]
xeno097: I think in general I would like to nuke the whole duplicity between signer commands array  commands enum but I want to talk with tx-fusion to see if they added the enum for some specific reason and if no then nuke the whole enum and go back to the old default behaviour where all evm chains were initialized with same value provided in the `HYP_KEY` env

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -151,7 +127,29 @@ export class MultiChainResolver implements ChainResolver {
...
+
+    if (argv.destination) {
+      chains.push(argv.destination);
+    }
+
+    if (!argv.chains) {
+      return Array.from(
+        new Set([...chains, ...this.getEvmChains(multiProvider)]),
+      );
+    }
```
paulbalaji, line 145: thank you!!! 🙏

paulbalaji (approved):

PR #4989, opened 2024-12-12
fix(contracts): make `releaseValueToRecipient` internal

aroralanuk:
### Description

- partially fixes https://github.com/chainlight-io/2024-08-hyperlane/issues/1


.changeset/nine-eyes-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/core': patch
```
yorhodes, line 2: I think this would technically be a `minor` or `major` change

yorhodes (approved):

PR #4995, opened 2024-12-12
feat: ignore unprocessable messages in metrics, add some new warp monitoring

tkporter:
### Description

- Ignores known unprocessable ezETH and pzETH messages from metrics, to avoid alerts there
- Added some logic to remove stale warp route monitors if a warp ID changes such that the previous warp monitor is stale
- Deployed new warp monitors for ezETH and pzETH to include swell
- Pulled in the treasure warp route ID that wasn't put into a PR yet

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/warpIds.ts
+++ b/typescript/infra/config/environments/mainnet3/warp/warpIds.ts
@@ -5,6 +5,7 @@ export enum WarpRouteIds {
   ArbitrumEthereumZircuitAMPHRETH = 'AMPHRETH/arbitrum-ethereum-zircuit',
   ArbitrumNeutronEclip = 'ECLIP/arbitrum-neutron',
   ArbitrumNeutronTIA = 'TIA/arbitrum-neutron',
+  ArbitrumTreasureMAGIC = 'MAGIC/arbitrum-treasure',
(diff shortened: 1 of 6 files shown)


Mo-Hussain (approved):

typescript/infra/config/environments/mainnet3/warp/warpIds.ts
```diff
@@ -5,6 +5,7 @@ export enum WarpRouteIds {
   ArbitrumEthereumZircuitAMPHRETH = 'AMPHRETH/arbitrum-ethereum-zircuit',
   ArbitrumNeutronEclip = 'ECLIP/arbitrum-neutron',
   ArbitrumNeutronTIA = 'TIA/arbitrum-neutron',
+  ArbitrumTreasureMAGIC = 'MAGIC/arbitrum-treasure',
```
Mo-Hussain, line 8: Do we need to update .registryrc?

ameten (approved):

paulbalaji (approved):

typescript/infra/config/environments/mainnet3/warp/warpIds.ts
paulbalaji: isn't this already in main without a registry update https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4998

Issue #5001, opened 2024-12-13
Operation status metric is incorrect for the prepare queue


daniel-savu:
`operation_status` labels are incorrect in the prepare queue. For instance, check:
- this [visualization](https://abacusworks.grafana.net/goto/FVARQmINR?orgId=1) of the prepare queue. Some messages have a status of `ReadyToSubmit`
- the source [code](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ea75978f8cf84d4dcc19e8c354b76ee29fa7e201/rust/main/agents/relayer/src/msg/op_submitter.rs#L281), based on which a message can only have a `ReadyToSubmit` when pushed to the submit queue. `queue.push(op, None)` does result in an operation being pushed to a queue without having its status changed, but whenever an operation fails at the submit step, it's send back to the prepare queue with a `Retry` status ([source](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ea75978f8cf84d4dcc19e8c354b76ee29fa7e201/rust/main/agents/relayer/src/msg/op_submitter.rs#L352))
- the logs for one of the messages in the queue at the time, which confirm that when `pop`ped it only had status `ReadyToSubmit` when in the submit queue, and `Retry(ErrorSubmitting)` otherwise: https://cloudlogging.app.goo.gl/5AnNqYQaPhyanWu56

I think the operation_status metric mislabeling is related to the issue above, might help find the root problem. At least for the `operation_status` mislabelling, the bug is likely in this function: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ea75978f8cf84d4dcc19e8c354b76ee29fa7e201/rust/main/hyperlane-core/src/traits/pending_operation.rs#L94

PR #5002, opened 2024-12-13
fix: operation_status metric inaccuracy

daniel-savu:
### Description

The `new_metric` passed into `set_status_and_update_metrics` was actually using the old `status` of the queue operation, causing prepare queue messages to be reflected in metrics as having the status they had in the submit queue. See https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5001 for more details.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5001

### Backward compatibility

Yes

### Testing

Will check on RC before releasing


tkporter (commented):
nice find!

PR #5003, opened 2024-12-13
chore: treasure fix relayer images

daniel-savu:


typescript/infra/config/environments/mainnet3/agent.ts
paulbalaji, line 1: testnet4 as well?

daniel-savu: fixed, good catch!

paulbalaji (approved):

PR #5004, opened 2024-12-13
feat(infra): warp monitor for pumpBTC warp route b/w ethereum<>sei

aroralanuk:


.registryrc
```diff
@@ -1 +1 @@
-01f2271592da4a06962e31ed384201f32cb070ab
+3aeee159382e1bbbf6539627aee52c9059d6bce6
```
paulbalaji, line 1: pull main before merging - because this is an older registry version than we have in main now

aroralanuk: done

paulbalaji (commented):
lgtm but would appreciate @Mo-Hussain giving the green light too

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumSeiPumpBTCWarpConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+const ethereumOwner = '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba';
+const seiOwner = '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba';
+
+export const getEthereumSeiPumpBTCWarpConfig = async (
+  routerConfig: ChainMap<RouterConfigWithoutOwner>,
+  _abacusWorksEnvOwnerConfig: ChainMap<OwnableConfig>,
+): Promise<ChainMap<HypTokenRouterConfig>> => {
+  const ethereum: HypTokenRouterConfig = {
+    ...routerConfig.ethereum,
+    ...getOwnerConfigForAddress(ethereumOwner),
```
nambrot, line 25: Is this necessary? My understanding is that if we just set `owner: ethereumOwner` is sufficient and preferred @aroralanuk @tkporter

aroralanuk: aren't we transferring ownership to PumpBTC before handing it over?

tkporter: this is the practice I was encouraging to ensure that proxyAdmin ownership is also explicitly set

It's a little redundant now that the `routerConfig` omits the `owner` and `ownerOverrides` now, but there was one scenario where we previously set `owner: ethereumOwner` and unintentionally had some `ownerOverrides` in this object that ended up ignoring the `owner` param. This is when I accidentally had transferred ownership of a proxy admin for a warp route

Happy to change the practice but probably makes sense to keep things consistent atm and use this

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumSeiPumpBTCWarpConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+  TokenType,
+} from '@hyperlane-xyz/sdk';
+
+import { getOwnerConfigForAddress } from '../../../../../src/config/environment.js';
+import {
+  RouterConfigWithoutOwner,
+  tokens,
+} from '../../../../../src/config/warp.js';
+
+const ethereumOwner = '0x77A0545Dc1Dc6bAee8d9c1d436c6688a75Ae5777';
```
tkporter, line 16: can you add comment on who owns these here

tkporter (approved):

PR #5007, opened 2024-12-13
feat: use zks_estimateFee for getting EIP 1559 price params on zkSync chains

tkporter:
### Description

Uses `zks_estimateFee` (https://docs.zksync.io/zksync-protocol/api/zks-rpc#zks_estimatefee), following the pattern that is used in the JS/TS zkSync tooling:

> sendTransaction calls populateFeeData https://github.com/zksync-sdk/zksync-ethers/blob/64763688d1bb5cee4a4c220c3841b803c74b0d05/src/signer.ts#L650

> populateFeeData calls estimateFee https://github.com/zksync-sdk/zksync-ethers/blob/64763688d1bb5cee4a4c220c3841b803c74b0d05/src/signer.ts#L702

> estimateFee uses that json rpc https://github.com/zksync-sdk/zksync-ethers/blob/main/src/provider.ts#L311


rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -202,15 +206,88 @@ where
...
+
+    let result = provider
+        .provider()
+        .request("zks_estimateFee", [tx.clone()])
+        .await?;
+    tracing::debug!(?result, ?tx, "Successfully fetched zkSync fee estimate");
+    Ok(result)
+}
+
+// From
```
daniel-savu, line 269: I believe this should've had a link which is missing but will merge as is

daniel-savu (approved):

PR #5012, opened 2024-12-16
fix: balance check skip confirmation

mkykadir:
### Description

Balance checking will skip confirmation when CLI flag is defined. When balances are insufficient, CLI asks for confirmation to continue with deployment, if `skip-confirmations` is provided through CLI args this should be bypassed.

### Backward compatibility

No, compatible with backwards

### Testing

Manual


jmrossy (approved):
Good idea!

PR #5017, opened 2024-12-16
fix: Hide outline in some widget components

jmrossy:
### Description

- Hide ouline in button and text input
- Rename clip path id in discord logo

### Backward compatibility

Yes

### DIFF
--- a/typescript/widgets/src/components/Button.tsx
+++ b/typescript/widgets/src/components/Button.tsx
@@ -7,7 +7,7 @@ export function Button(props: Props) {
   const { className, children, ...rest } = props;
 
   const base =
-    'htw-flex htw-items-center htw-justify-center htw-rounded-sm htw-transition-all outline-none focus:outline-none';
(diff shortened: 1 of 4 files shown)


typescript/widgets/src/components/Button.tsx
```diff
@@ -7,7 +7,7 @@ export function Button(props: Props) {
   const { className, children, ...rest } = props;
 
   const base =
-    'htw-flex htw-items-center htw-justify-center htw-rounded-sm htw-transition-all';
+    'htw-flex htw-items-center htw-justify-center htw-rounded-sm htw-transition-all outline-none focus:outline-none';
```
Xaroz, line 10: missing `htw` prefix

Xaroz (approved):

typescript/widgets/src/components/Button.tsx
jmrossy: Good catch!!!

PR #5019, opened 2024-12-16
feat: Add Zerion memecoins to getter

ltyu:

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBaseJackieChainWarpConfig.ts
+++ b/typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBaseJackieChainWarpConfig.ts
@@ -1,0 +1,40 @@
+import { ethers } from 'ethers';
+
+import {
+  ChainMap,
+  HypTokenRouterConfig,
+  OwnableConfig,
+  TokenType,
+} from '@hyperlane-xyz/sdk';
+
+import {
+  RouterConfigWithoutOwner,
+  tokens,
+} from '../../../../../src/config/warp.js';
+
+const ISM_CONFIG = ethers.constants.AddressZero; // Default ISM
+
+export const getBasezeronetworkRe7LRTWarpConfig = async (
+  routerConfig: ChainMap<RouterConfigWithoutOwner>,
+  abacusWorksEnvOwnerConfig: ChainMap<OwnableConfig>,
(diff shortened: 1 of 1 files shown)


typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBaseBRETTWarpConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+export const getBaseZeronetworkBrettConfig = async (
+  routerConfig: ChainMap<RouterConfigWithoutOwner>,
+  abacusWorksEnvOwnerConfig: ChainMap<OwnableConfig>,
+): Promise<ChainMap<HypTokenRouterConfig>> => {
+  const base: HypTokenRouterConfig = {
+    ...routerConfig.base,
+    ...abacusWorksEnvOwnerConfig.base,
+    type: TokenType.collateral,
+    token: tokens.base.brett,
+    interchainSecurityModule: ISM_CONFIG,
```
nambrot, line 26: Should we remove this? I.e. omit ism config if its 0x0?

typescript/infra/src/config/warp.ts
```diff
@@ -19,6 +19,9 @@ export const tokens: ChainMap<Record<string, Address>> = {
   base: {
     cbBTC: '0xcbB7C0000aB88B473b1f5aFd9ef808440eed33Bf',
     USDC: '0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913',
+    jc: '0x93aC8f81e799c3b76112A649715DD86CEa655BDF',
```
nambrot, line 22: This is a bigger issue beyond this PR, so feel free to resolve this, but wondering if we should use the token list standard for this

nambrot (approved):

typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumBaseBRETTWarpConfig.ts
ltyu: the other configs do this. i'm not sure of the implications of removing it, but also okay with being explicit with this.

typescript/infra/src/config/warp.ts
ltyu: For future ref: https://tokenlists.org/

nambrot (approved):

PR #5023, opened 2024-12-17
fix(warpChecker): Check for token mismatch before decimals check

Mo-Hussain:
### Description

- reorder checks to so that a token mismatch is caught before decimal checks are done. This will avoid us calling `.decimals` on a contract that is not a token contract i.e if the config has been misconfigured 

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/token/checker.ts
+++ b/typescript/sdk/src/token/checker.ts
@@ -68,6 +68,7 @@ export class HypERC20Checker extends ProxiedRouterChecker<
     const expectedConfig = this.configMap[chain];
     const hypToken = this.app.router(this.app.getContracts(chain));
 
+    // Check if configured token type matches actual token type
     if (isNativeTokenConfig(expectedConfig)) {
       try {
         await this.multiProvider.estimateGas(chain, {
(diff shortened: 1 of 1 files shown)


paulbalaji (approved):
lgtm, add a `patch` changeset

typescript/sdk/src/token/checker.ts
```diff
@@ -126,6 +103,29 @@ export class HypERC20Checker extends ProxiedRouterChecker<
         this.addViolation(violation);
       }
     }
+
+    // Check all actual decimals are consistent
```
tkporter, line 107: if this needs to live here as opposed to other places, a comment would be useful

typescript/sdk/src/token/checker.ts
```diff
@@ -126,6 +103,29 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
+    const actualChainDecimals = await this.getEvmActualDecimals();
```
tkporter, line 108: >  This will avoid us calling .decimals on a contract that is not a token contract i.e if the config has been misconfigured

Is this because one of the prior calls will throw or something? As I don't see an early return

Mo-Hussain: For a collateral token, we will call .decimal() on it when we go down this code path. If the config in code was setup incorrectly such that the collateral ER20 is not in fact and ERC20, we throw and unhandled error which is not so easy to understand.  

We could instead wrap the decimal call in a try catch and then throw with a better error message but reordering these checks ensures the the collateral ERC20 in the config in code matched on chain state before calling decimals

PR #5031, opened 2024-12-17
fix: Stop advancing cursor when not all logs were stored

ameten:
### Description

Stop advancing cursor when not all logs were stored.

### Backward compatibility

Yes

### Testing

E2E Ethereum and Sealevel test (for no regressions)

### DIFF
--- a/rust/main/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/main/hyperlane-base/src/contract_sync/mod.rs
@@ -250,8 +250,21 @@ where
 
             accumulated_stored_logs_count += stored_logs_count as u64;
             if accumulated_stored_logs_count == deduped_logs_count {
+                info!(
+                    ?accumulated_stored_logs_count,
+                    ?stored_logs_count,
+                    ?deduped_logs_count,
+                    "Successfully stored all logs",
+                );
                 break;
             }
+
+            warn!(
+                ?accumulated_stored_logs_count,
+                ?stored_logs_count,
+                ?deduped_logs_count,
+                "Repeat storing logs since not all logs were stored yet",
+            );
         }
 
         logs


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -221,7 +230,7 @@ where
         &self,
         logs: Vec<(Indexed<T>, LogMeta)>,
         stored_logs_metric: &GenericCounter<AtomicU64>,
-    ) -> Vec<(Indexed<T>, LogMeta)> {
+    ) -> Result<Vec<(Indexed<T>, LogMeta)>, Vec<(Indexed<T>, LogMeta)>> {
```
daniel-savu, line 233: sounds like what you really want to return here is a tuple of `(Vec<(Indexed<T>, LogMeta)>, bool)` where the bool indicates the success of the `store_logs` call. I think that'd be cleaner than this overloaded `Result` type

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -179,7 +181,12 @@ where
                     }
                 };
 
-                let logs = self.dedupe_and_store_logs(logs, stored_logs_metric).await;
+                let logs_result = self.dedupe_and_store_logs(logs, stored_logs_metric).await;
+                let (logs, error) = match logs_result {
+                    Ok(logs) => (logs, false),
+                    Err(logs) => (logs, true),
+                };
```
daniel-savu, line 188: this usage pattern also indicates a `bool` tuple item would be cleaner than a `Result`

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -243,7 +252,12 @@ where
         }
         // Report amount of deliveries stored into db
         stored_logs_metric.inc_by(stored as u64);
-        logs
+
+        if (stored as usize) < logs.len() {
+            return Err(logs);
```
daniel-savu, line 257: worth logging a warning if we stored insufficient logs? Probably good to include the number of deduped logs vs actually stored logs as structured log fields

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -199,10 +206,12 @@ where
...
-                if let Err(err) = cursor.update(logs, range).await {
-                    warn!(?err, "Error updating cursor");
-                    break Some(SLEEP_DURATION);
-                };
+                if !error {
+                    if let Err(err) = cursor.update(logs, range).await {
+                        warn!(?err, "Error updating cursor");
+                        break Some(SLEEP_DURATION);
+                    };
+                }
```
daniel-savu, line 214: if you actually get an error, could add a `warn` that we're not even gonna update the cursor to retry, including the range. Also can you pls add `range` to the `warn` above as a drive-by (the `Error updating cursor` one)

daniel-savu (commented):
Great idea that we shouldn't skip over if inserting logs into the db fails. However I wonder if it makes sense to request the logs again with the cursor, versus re-attempting to store the logs we deduped into the db? Presumably the logs won't change even if we fetch them again. As you suggested the problem is likely at the db insertion step, so maybe we could propagate whatever error occurs in the db that prevents all logs from being stored.

tl;dr This change adds reliability at the cost of extra network calls, but I have a feeling we wouldn't even need to make the extra network call - maybe just sleep for a bit and retry inserting to the db. But would need to see the exact db logs / errors first

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: Fixed

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: Fixed

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: Added a warning

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: Added

ameten:
Currently Scraper aims to store as many logs as possible from a range. If it cannot get a block or a transaction from RPC, it will fail to store the logs. So, the problem is in unreliable RPC, not DB error.

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -221,29 +230,39 @@ where
...
                 "Stored logs in db",
             );
         }
         // Report amount of deliveries stored into db
-        stored_logs_metric.inc_by(stored as u64);
-        logs
+        stored_logs_metric.inc_by(stored_logs_count as u64);
+
+        let deduped_logs_len = logs.len();
+        if (stored_logs_count as usize) < deduped_logs_len {
```
daniel-savu, line 257: what would happen in this scenario:
- we get 5 logs from the RPC
- we only manage to insert 2
- we skip updating the cursor because 2 < 5
- we get 5 logs from the RPC again
- we manage to insert the other 3
- we still skip updating the cursor because 3 < 5
- we keep looping?

ameten: As it is implemented in this PR, yes.

daniel-savu: then this sounds like a bug? The indexer would become permanently stuck in this infinite loop whenever one message isn't inserted successfully

ameten: I see what you mean. `store_logs` is not idempotent in sense how many logs it inserts.

ameten: I implemented your suggestion @daniel-savu with retrying `store_logs`.

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -230,39 +222,39 @@ where
...
-        // Store deliveries
-        let stored_logs_count = match self.store.store_logs(&logs).await {
-            Ok(stored) => stored,
-            Err(err) => {
-                warn!(?err, "Error storing logs in db");
-                Default::default()
+        let deduped_logs_count = logs.len() as u64;
+        let mut accumulated_stored_logs_count: u64 = 0;
+
+        loop {
```
daniel-savu, line 232: nit: `loop` to me indicates a long-running process, but since this is only expected to do a couple of iterations until full success I'd just use a `while accumulated_stored_logs_count < deduped_logs_count`

daniel-savu (approved):

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -225,24 +226,34 @@ where
...
+                    count = stored_logs_count,
+                    sequences = ?logs.iter().map(|(log, _)| log.sequence).collect::<Vec<_>>(),
+                    "Stored logs in db",
+                );
+            }
+            // Report amount of deliveries stored into db
+            stored_logs_metric.inc_by(stored_logs_count as u64);
+
+            accumulated_stored_logs_count += stored_logs_count as u64;
+            if accumulated_stored_logs_count == deduped_logs_count {
```
daniel-savu, line 252: but still maybe worth a warning if we didn't store all the logs in one go

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -225,24 +226,47 @@ where
...
-            debug!(
-                domain = self.domain.as_ref(),
-                count = stored,
-                sequences = ?logs.iter().map(|(log, _)| log.sequence).collect::<Vec<_>>(),
-                "Stored logs in db",
+            // Report amount of deliveries stored into db
+            stored_logs_metric.inc_by(stored_logs_count as u64);
+
+            accumulated_stored_logs_count += stored_logs_count as u64;
+            if accumulated_stored_logs_count == deduped_logs_count {
```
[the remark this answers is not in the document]
tkporter: accommodating this re-indexing behavior is probably something we should try to unit test for tbh

PR #5042, opened 2024-12-18
feat: support safe tx parsing for zksync chains

paulbalaji:
### Description

1. correctly generate core config for zksync chains
	- copied from `pb/zksync` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/pb/zksync/typescript/infra/config/environments/mainnet3/core.ts
	- note the storage aggregation ism was not around when doing the first zksync chain deploys, and we haven't gone back and updated the config generation yet to include this
	- required so we infer the correct config for zksync changes and don't assume it's the same as non-zksync chains

2. update ISM reader to return storagemultisigism types on zksync
	- since static ISMs are not supported on zksync
	- the moduleType is the same, so the reader has to determine if it's static/storage
	- note: at the moment on non-zksync we assume it's a static multisig/aggregation ISM
	- required so we correctly compare the config vs onchain config

### Testing

- these changes were used to verify SAFE txs for the last couple of batches
- the config changes have also been used in production on zksync chains since October 23rd 2024 [`26d198a` (#4761)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4761/commits/26d198a18fe0dce0ebf28ab75c32788e7104e2fc#diff-d4db62438f3fd9acf24be52093e81d126292859f9f48c81aa1704d41fe8ddf1a)


cmcewen (changes requested):
Tests not all passing

typescript/infra/config/environments/mainnet3/core.ts
```diff
@@ -52,17 +74,30 @@ export const core: ChainMap<CoreConfig> = objMap(
       ...owner,
     };
 
+    // No static aggregation ISM support on zkSync
+    const defaultZkSyncIsm: RoutingIsmConfig = {
```
tkporter, line 78: nit: do we need to build this when isZkSyncChain is false? not a big deal if it's clunky to only do it conditionally

typescript/infra/config/environments/testnet4/core.ts
```diff
@@ -36,11 +38,26 @@ export const core: ChainMap<CoreConfig> = objMap(
         .map((origin) => [origin, defaultMultisigConfigs[origin]]),
     );
```
tkporter, line 40: is this the exact same logic as mainnet3? wonder if there's an easy opportunity for consolidation

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -208,9 +216,14 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
       async (module) => this.deriveIsmConfig(module),
     );
 
+    // If it's a zkSync chain, it must be a StorageAggregationIsm
+    const ismType = this.isZkSyncChain
+      ? IsmType.STORAGE_AGGREGATION
```
tkporter, line 221: if we have storage aggregations now, do we still need to have the weird setup for avoiding aggregation hooks / ISMs?

tkporter (approved):
you can dismiss requested changes

typescript/sdk/src/ism/EvmIsmReader.ts
paulbalaji: I believe we have aggregation isms but not aggregation hooks, so at least that can be updated. But I would prefer to do that in a separate PR at the time we actually enact those changes onchain, otherwise we'll get the checker complaining

typescript/infra/config/environments/testnet4/core.ts
paulbalaji: yup it is - it's almost identical except for the protocol fee. mainnet3 set to 0 wei, testnet4 is 1 wei

Issue #4059, opened 2024-06-26
Relayer retry API: add the # of messages prioritized in the queue to the http response


tkporter:
## Problem

Paul and I recently tried retrying messages for a particular domain shortly after starting up - because it takes some time to get to a point where the messages are actually in the prep queue, when we retried no messages were actually prioritized. We had to look at logs to realize it was bc of kinda slow startup.

## Solution

Adding the # of messages in the queue that were prioritized would be useful to indicate this was the case

## Nice to Have

**Describe non-essential extensions to the solution**
Additional features which should be implemented if they are easy to accommodate but otherwise can be skipped

kamiyaa:
How is the retry being done?

kamiyaa:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4779a19358e119e17e7889a3534fe862e760c948/rust/main/agents/relayer/src/msg/op_queue.rs#L71

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4779a19358e119e17e7889a3534fe862e760c948/rust/main/agents/relayer/src/msg/op_submitter.rs#L251

It looks like there is an internal queue mechanism for processing these retry message (along with potential other messages?). With this current design, it's difficult to return the results of the message retry, because we will somehow need to wait for the operation to be processed and somehow get its results.

I'm thinking we just remove the queue mechanism (for message retry) and just do the message retry logic in the http handler.
What do you think? @tkporter @daniel-savu

PR #5043, opened 2024-12-18
feat: add the # of messages prioritized in the queue to the http resp…

kamiyaa:
### Description

Add a mpsc channel between relayer and message retry handler, so message retry handler knows the retry results from the relayer.

`POST /message_retry` endpoint now returns a JSON with details on the retry request:

```rust
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct MessageRetryResponse {
    /// ID of the retry request
    pub uuid: String,
    /// how many pending operations were processed
    pub processed: usize,
    /// how many of the pending operations matched the retry request pattern
    pub matched: u64,
}
```

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4059

### Testing

Updated unit tests to return a response so the handler is not blocked


rust/main/Cargo.toml
```diff
@@ -32,7 +32,7 @@ anyhow = "1.0"
 async-trait = "0.1"
 async-rwlock = "1.3"
 auto_impl = "1.0"
-axum = "0.6.1"
+axum = { version = "0.6.1", features = ["macros"] }
```
kamiyaa, line 35: Can remove this feature if we deem it unnecessary and makes the production binary too big.
I just found it useful during development to have more clear error messages.
See https://docs.rs/axum/latest/axum/attr.debug_handler.html

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,86 @@
...
+use tokio::sync::{broadcast::Sender, mpsc, Mutex};
 
 const MESSAGE_RETRY_API_BASE: &str = "/message_retry";
 
-#[derive(new, Clone)]
+#[derive(new)]
 pub struct MessageRetryApi {
-    tx: Sender<MatchingList>,
+    tx: Sender<MessageRetryRequest>,
+    rx: Arc<Mutex<mpsc::Receiver<MessageRetryResponse>>>,
```
kamiyaa, line 16: This doesn't necessarily have to be an `Arc<Mutex<>>` but mostly because we need to satisfy
```rust
    pub fn router(&self) -> Router;
    pub fn get_route(&self) -> (&'static str, Router);
```

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
-    match tx.send(retry_req_payload) {
-        Ok(_) => "Moved message(s) to the front of the queue".to_string(),
-        // Technically it's bad practice to print the error message to the user, but
-        // this endpoint is for debugging purposes only.
-        Err(err) => format!("Failed to send retry request to the queue: {}", err),
+) -> Result<Json<MessageRetryResponse>, String> {
+    let uuid = uuid::Uuid::new_v4();
+    let uuid_string = uuid.to_string();
+
+    tracing::debug!("Sending message retry request: {uuid_string}");
```
tkporter, line 48: nit: for the tracing logs we intentionally want to avoid templated strings and favor structured logs, they're nicer to work with when they're parsed e.g. in GCP logging

so something like this, and always have the message be a static string
```
tracing::debug!(uuid=uuid_string, "Sending message retry request");
```

goes here and throughout the pr

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
+        .map_err(|err| {
+            // Technically it's bad practice to print the error message to the user, but
+            // this endpoint is for debugging purposes only.
+            format!("Failed to send retry request to the queue: {}", err)
+        })?;
+
+    let mut rx = state.rx.lock().await;
+
+    // Wait for response from message retry.
+    // Warning: this potentially blocks other retry requests
```
tkporter, line 65: I wonder if we should use the lock that we acquire above as a way to prevent this? i.e. by acquiring the lock before we do the `tx.send`.

It's a bit of a weird and clunky hammer, but imo the usage of this is infrequent and only by the relayer operator, so I'm not really concerned about the poor UX / practice not allowing concurrent retry requests. I don't have a strong intuition for how quick we expect to get an ack back here, I assume we can end up blocking for many seconds if prep queues are large

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
+    // from properly returning because their response might be consumed by
+    // another request
+    loop {
```
tkporter, line 68: especially as we have a lock acquired, can we have a timeout? It can be long, like 60s or more, but want to avoid waiting forever if something goes wrong

rust/main/agents/relayer/src/server/message_retry.rs
tkporter: Actually a maybe better idea if we wanna avoid holding locks for this -- couldn't we have `MessageRetryRequest` include a channel that's used for an ack back to this specific request?

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
+        tracing::debug!("Waiting for response from relayer: {uuid_string}");
+        if let Some(resp) = rx.recv().await {
```
tkporter, line 70: note that when we send a retry request, we're doing a broadcast to *all* submitters. This way we can have a single retry request that is capable of retrying messages destined for multiple different chains (e.g. imagine validators on chain A went down for a bit, halting messages originating from chian A, and they're back up now -- we wanna make sure that all messages originating from A for all destinations are retried)

So as this is written I believe we'd just get the response from the first submitter that's handled the request, but not all the submitters. So the response to the user should probably be the summed responses from each submitter

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +75,55 @@ impl OpQueue {
         // The other consideration is whether to put the channel receiver in the OpQueue or in a dedicated task
         // that also holds an Arc to the Mutex. For simplicity, we'll put it in the OpQueue for now.
         let mut message_retry_requests = vec![];
+
         while let Ok(message_id) = self.retry_rx.lock().await.try_recv() {
```
tkporter, line 79: nit: this should've been renamed before but can you rename `message_id` to something more accurate, as it's actually a `MessageRetryRequest`

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
+    pub uuid: String,
+    pub pattern: MatchingList,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct MessageRetryResponse {
+    /// ID of the retry request
+    pub uuid: String,
+    /// how many pending operations were processed
+    pub processed: usize,
```
tkporter, line 36: nit: would probably also avoid `processed` as we use this jargon elsewhere in lots of places, bound to get confusing

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +75,55 @@ impl OpQueue {
...
                         operation = %op,
                         queue_label = %self.queue_metrics_label,
                         "Retrying OpQueue operation"
                     );
-                    op.reset_attempts()
+                    op.reset_attempts();
+                }
+                // update retry metrics
+                for (retry_req, req_metric) in message_retry_requests.iter_mut() {
+                    if retry_req.pattern.op_matches(&op) {
```
tkporter, line 114: can we avoid doing this here just for incrementing `.matched`, as we have very similar logic a few lines above? ideally shouldn't need to call `op_matches` more than once unnecessarily

tkporter (commented):
good stuff!

excited that this task is fairly meaty, lots of interesting concurrency and touching a lot of systemic things in the relayer :)

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: You mean like a oneshot channel inside `MessageRetryRequest`?

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: Ah, that makes sense

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: What would you propose as a better name?

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: Yep :+1:

rust/main/agents/relayer/src/server/message_retry.rs
tkporter: yeah exactly

oneshot may or may not be the right channel bc of the whole multiple submitters needing to send a response back thing

But I think this is probably a better design? wdyt

rust/main/agents/relayer/src/server/message_retry.rs
tkporter: idk maybe `evaluated` or something?

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: Yeah, sounds good to me. Just generate a mpsc channel for every retry request. And that channel will be solely used for that specific retry request.

rust/main/Cargo.toml
daniel-savu: nicee it was actually quite frustrating to debug axum problems but i'd keep this feature as a dev dependency if possible pls - it doesn't help the release build once it passes e2e

rust/main/agents/relayer/src/server/message_retry.rs
daniel-savu: not sure if memory usage assumptions were considered here but I think we should be ok given the very low volume of retry requests

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,57 @@ impl OpQueue {
...
         let mut queue = self.queue.lock().await;
+        let queue_length = queue.len();
+
         let mut reprioritized_queue: BinaryHeap<_> = queue
             .drain()
             .map(|Reverse(mut op)| {
-                if message_retry_requests.iter().any(|r| r.op_matches(&op)) {
+                let matches = message_retry_requests
+                    .iter_mut()
+                    .map(|(retry_req, req_metric)| {
```
daniel-savu, line 102: nit: rename `req_metric` to `retry_response`

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,57 @@ impl OpQueue {
...
+                        let match_res = retry_req.pattern.op_matches(&op);
+                        // update retry metrics
+                        if match_res {
+                            req_metric.matched += 1
+                        }
+                        match_res
+                    })
+                    .any(|x| x);
```
daniel-savu, line 110: I'm not super confident this works as expected since `any` will stop evaluating after the first `true` result and you're not collecting before it. You probably want to `filter` instead of `map`, `collect`, and then take the `len` of the fitered requests to see if there were any matches

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,57 @@ impl OpQueue {
...
+                if matches {
                     info!(
```
daniel-savu, line 112: realizing it'd be useful to include the retry request in this log

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,57 @@ impl OpQueue {
...
                     );
-                    op.reset_attempts()
+                    op.reset_attempts();
                 }
                 Reverse(op)
             })
             .collect();
+
+        tracing::debug!("Sending retry request metrics back");
+        for (retry_req, mut req_metric) in message_retry_requests {
```
daniel-savu, line 124: how about including this debug log inside the for loop and including each retry_req and response in there for later reference?

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,87 @@
...
-        // Technically it's bad practice to print the error message to the user, but
-        // this endpoint is for debugging purposes only.
-        Err(err) => format!("Failed to send retry request to the queue: {}", err),
+) -> Result<Json<MessageRetryResponse>, String> {
+    let uuid = uuid::Uuid::new_v4();
+    let uuid_string = uuid.to_string();
+
+    tracing::debug!(uuid = uuid_string, "Sending message retry request");
+
+    let (transmitter, mut receiver) = mpsc::channel(ENDPOINT_MESSAGES_QUEUE_SIZE);
```
daniel-savu, line 47: note this should be as big as the total number of chains the relayer operates. `ENDPOINT_MESSAGES_QUEUE_SIZE` is 100 which is already less than we have in prod I think. You can pass in the length of the configured `relayChains` when the server is constructed, and use that length as the channel size

daniel-savu: parsed here: https://github.com/kamiyaa/hyperlane-monorepo/blob/491f98110ed68cf3bd6838049ab3451fee775205/rust/main/agents/relayer/src/settings/mod.rs#L108

daniel-savu: actually much easier, pass the length as the max between `self.origin_chains` and `self.destination_chains` https://github.com/kamiyaa/hyperlane-monorepo/blob/491f98110ed68cf3bd6838049ab3451fee775205/rust/main/agents/relayer/src/relayer.rs#L66

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,87 @@
...
+        })?;
+
+    let mut resp = MessageRetryResponse {
+        uuid: uuid_string,
+        evaluated: 0,
+        matched: 0,
+    };
+
+    // Wait for responses from relayer
+    tracing::debug!(uuid = resp.uuid, "Waiting for response from relayer");
```
daniel-savu, line 68: probably worth including the entire request instead of just the uuid in this log. Or if not here in another log, so it's easy to find the request in logs by searching by uuid

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,87 @@
...
+    while let Some(relayer_resp) = receiver.recv().await {
+        tracing::debug!(uuid = resp.uuid, "Relayer response");
+        resp.evaluated += relayer_resp.evaluated;
+        resp.matched += relayer_resp.matched;
     }
+
+    Ok(Json(resp))
```
daniel-savu, line 75: I think it'd make things easier to debug if we returned a map of responses by uuid in the json response

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,87 @@
...
+    let mut resp = MessageRetryResponse {
+        uuid: uuid_string,
+        evaluated: 0,
+        matched: 0,
+    };
+
+    // Wait for responses from relayer
+    tracing::debug!(uuid = resp.uuid, "Waiting for response from relayer");
+    while let Some(relayer_resp) = receiver.recv().await {
+        tracing::debug!(uuid = resp.uuid, "Relayer response");
```
daniel-savu, line 70: nit, and would make sense to me to also include the response
```suggestion
        tracing::debug!(uuid = resp.uuid, "Submitter response to retry request");
```

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -58,12 +120,22 @@ mod tests {
...
-        (addr, broadcast_tx.subscribe())
+        let retry_req_rx = broadcast_tx.subscribe();
+
+        TestServerSetup {
+            socket_address: addr,
+            retry_req_rx,
+        }
     }
 
+    #[tracing_test::traced_test]
```
daniel-savu, line 131: this is so useful, always wondered why we couldn't see the logs 🔥 great addition

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -137,25 +263,52 @@ mod tests {
...
+                    .pattern
+                    .op_matches(&(Box::new(pending_operation.clone()) as QueueOperation)));
+                let resp = MessageRetryResponse {
+                    uuid: req.uuid,
+                    evaluated: 10,
+                    matched: 2,
+                };
+                req.transmitter.send(resp).await.unwrap();
+            }
+        };
```
daniel-savu, line 280: sweet, I think you can refactor this into a function that's reused across unit tests

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -137,25 +263,52 @@ mod tests {
...
-        assert!(list.op_matches(&(Box::new(pending_operation) as QueueOperation)));
+        let resp_body = response
+            .text()
+            .await
+            .expect("Failed to parse response body");
+
+        let resp_json: MessageRetryResponse =
+            serde_json::from_str(&resp_body).expect("Failed to deserialize response body");
+        assert_eq!(resp_json.evaluated, 10);
+        assert_eq!(resp_json.matched, 2);
```
daniel-savu, line 302: similar here, the logic is mostly duplicated so you can refactor

daniel-savu (commented):
😎 awesome work, hitting the ground running from the first PR!!

rust/main/agents/relayer/Cargo.toml
```diff
@@ -42,8 +42,10 @@ tokio = { workspace = true, features = [
 ] }
 tokio-metrics.workspace = true
 tracing-futures.workspace = true
+tracing-test.workspace = true
```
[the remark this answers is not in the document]
kamiyaa: Yes, thanks!

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: okay :+1:

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: we can't really log the specific retry request because we only check if some retry request matches the operation.
Unless we want to log each retry request that matches in the `.map()`

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: Good idea :+1:

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: What do you mean by this? I think currently the prepare, submit and confirm queue all return the same UUID :thinking: 

Or are you thinking of something like:

```json
{
    "prepare_queue": {
        "evaluated": ...,
        "matched": ...,
   },
    "submit_queue": {
        "evaluated": ...,
        "matched": ...,
   },
    "confirm_queue": {
        "evaluated": ...,
        "matched": ...,
   }
}
```

rust/main/Cargo.toml
kamiyaa: Updated to dev-dependency

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: Why is it as big as total number of chains?
I thought there were just 3 queues? Or is it 3 queues per chain?

daniel-savu: hm, actually this just sets the max number of messages in the channel, for some reason I was thinking about the max number of consumers. The max consumers is `num_chains * num_queues`, where `num_chains` is 100+ and `num_queues` is 3 as you point out. 

Actually since: 1) the tokio `mpsc` applies backpressure by making senders wait; and 2) we expect very low message volumes since it's only relayer operators manually triggering this; then we can set the size to be `1` here 

just trying to be a bit memory conscious since the relayer started using up increasingly more recently

kamiyaa: Okay, I can update it to 1 :+1:

rust/main/agents/relayer/src/msg/op_queue.rs
daniel-savu: you're right, though we could take advanced of the `filter` result and accumulate it into a vec of maps, so we include all request IDs that matched in the log. It'd also be cleaner to refactor everything inside the `.map(|Reverse(mut op)| {` into its own function, since it's getting quite crowded

rust/main/agents/relayer/src/server/message_retry.rs
daniel-savu: oh right, every message has its own channel actually and you're just adding up responses from each sender, I missed that

Things are a bit more complicated:
- we should use `try_recv` instead of `recv` everywhere, to avoid blocking on an empty channel. However because different senders may take varying amounts of time to respond to the request and the channel may be spuriously empty before we received all responses:
- we need to wait on a response from every sender (`num_chains * num_queues`). So we should keep a counter and only return once we got responses from all `receiver.sender_strong_count()`

daniel-savu: your logic to add up the response fields makes sense now

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,69 @@ impl OpQueue {
...
+
+        for (retry_req, mut retry_response) in message_retry_requests {
+            retry_response.evaluated = queue_length;
+            tracing::debug!(
+                uuid = retry_response.uuid,
+                evaluated = retry_response.evaluated,
+                matched = retry_response.matched,
+                "Sending relayer retry response back"
+            );
+            let _ = retry_req.transmitter.send(retry_response).await;
```
daniel-savu, line 138: we should log an error if this `send` fails

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -44,7 +44,7 @@ async fn retry_message(
 
     tracing::debug!(uuid = uuid_string, "Sending message retry request");
 
-    let (transmitter, mut receiver) = mpsc::channel(ENDPOINT_MESSAGES_QUEUE_SIZE);
+    let (transmitter, mut receiver) = mpsc::channel(1);
```
daniel-savu, line 47: nit: add a comment mentioning that this channel is only created to service this single retry request so we're expecting a single response from each transmitter end, hence why we're using a size of 1

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -128,12 +134,41 @@ mod tests {
         }
     }
 
+    async fn send_retry_responses_task(
```
daniel-savu, line 137: nit: this is technically a future rather than a task, even after the `join` it's still a future that executes the joined futures concurrently

daniel-savu (approved):

daniel-savu:
🙏 go @kamiyaa! You need to fix clippy for CI to pass btw: `cargo clippy -- -D warnings`

kamiyaa:
> cargo clippy -- -D warnings

Ah, let me do that before merging!

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,71 @@ impl OpQueue {
...
+                let matches = !matched_requests.is_empty();
+                if matches {
                     info!(
                         operation = %op,
                         queue_label = %self.queue_metrics_label,
                         "Retrying OpQueue operation"
                     );
-                    op.reset_attempts()
+                    op.reset_attempts();
+                    for matched_req in matched_requests {
```
tkporter, line 122: nit: does this need to be looped after? or can we just log this on line 106/107 in the filter_map when we find a match

imo doesnt totally matter if we log before or after we actually reset the op

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,71 @@ impl OpQueue {
...
+        let queue_length = queue.len();
+
         let mut reprioritized_queue: BinaryHeap<_> = queue
             .drain()
             .map(|Reverse(mut op)| {
-                if message_retry_requests.iter().any(|r| r.op_matches(&op)) {
+                let matched_requests: Vec<_> = message_retry_requests
+                    .iter_mut()
+                    .filter_map(|(retry_req, retry_response)| {
+                        let match_res = retry_req.pattern.op_matches(&op);
```
tkporter, line 103: nit: really a dumb stylistic thing but `if retry_req.pattern.op_matches(&op) {` feels easier to read than creating a temp variable

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,71 @@ impl OpQueue {
...
+                        if match_res {
+                            retry_response.matched += 1;
+                            Some(retry_req.uuid.clone())
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
+
+                let matches = !matched_requests.is_empty();
```
tkporter, line 114: nit: similar thing here if it's a simple call to get the boolean imo no need for an extra var

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,71 @@ impl OpQueue {
...
+        for (retry_req, mut retry_response) in message_retry_requests {
+            retry_response.evaluated = queue_length;
+            tracing::debug!(
+                uuid = retry_response.uuid,
+                evaluated = retry_response.evaluated,
+                matched = retry_response.matched,
+                "Sending relayer retry response back"
+            );
+            if let Err(err) = retry_req.transmitter.send(retry_response).await {
+                tracing::error!(err = err.to_string(), "Failed to send retry response");
```
tkporter, line 139: nit: does this work? to use the Debug impl instead of to_string, which is the pattern we generally go with 
```
tracing::error!(?err, "Failed to send retry response");
```

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -74,27 +74,71 @@ impl OpQueue {
...
+                if matches {
                     info!(
                         operation = %op,
                         queue_label = %self.queue_metrics_label,
                         "Retrying OpQueue operation"
                     );
-                    op.reset_attempts()
+                    op.reset_attempts();
+                    for matched_req in matched_requests {
+                        info!(uuid = matched_req, "Matched request");
```
tkporter, line 123: nit: should this be a debug or maybe trace? feels pretty noisy if e.g. we wanna retry all messages to a destination, which could match thousands of messages

Also was the intent to put the op in here?

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,95 @@
...
 
-#[derive(new, Clone)]
+#[derive(new)]
 pub struct MessageRetryApi {
-    tx: Sender<MatchingList>,
+    retry_request_transmitter: Sender<MessageRetryRequest>,
+}
+
+#[derive(Clone, Debug)]
+pub struct MessageRetryApiState {
```
tkporter, line 17: what's the difference between this and `MessageRetryApi` now? can we consolidate into one type?

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,95 @@
...
-        // this endpoint is for debugging purposes only.
-        Err(err) => format!("Failed to send retry request to the queue: {}", err),
+) -> Result<Json<MessageRetryResponse>, String> {
+    let uuid = uuid::Uuid::new_v4();
+    let uuid_string = uuid.to_string();
+
+    tracing::debug!(uuid = uuid_string, "Sending message retry request");
+
+    // This channel is only created to service this single
+    // retry request so we're expecting a single response
```
tkporter, line 48: as we expect 1 response from each transmitter end, if there are N transmitter ends (i.e. N submission tasks), wouldn't we expect N responses sent via the channel? So this would cause all sends other than the first one to block?

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,95 @@
...
+
+    let mut resp = MessageRetryResponse {
+        uuid: uuid_string,
+        evaluated: 0,
+        matched: 0,
+    };
+
+    // Wait for responses from relayer
+    tracing::debug!(uuid = resp.uuid, "Waiting for response from relayer");
+    while let Some(relayer_resp) = receiver.recv().await {
```
tkporter, line 72: this doesn't guarantee that it waits for all responses from all the submitters right?
e.g. if like 2 out of 5 responses come in and we receive those 2 but still have 3 responses that haven't come in yet, we never get those 3 responses

tkporter: Wonder if we could use the transmitter's https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Sender.html#method.receiver_count to figure out how many responses we'd expect to get back? and only move on after we've gotten `receiver_count` responses

rust/main/utils/run-locally/src/server.rs
```diff
@@ -0,0 +1,61 @@
+use std::io;
+
+use reqwest::Url;
+use serde::{Deserialize, Serialize};
+
+use crate::RELAYER_METRICS_PORT;
+
+/// Copied from agents/relayer/src/server/message_retry.rs
```
tkporter, line 8: can we not import this?

rust/main/utils/run-locally/src/server.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct MessageRetryResponse {
+    /// ID of the retry request
+    pub uuid: String,
+    /// how many pending operations were evaluated
+    pub evaluated: usize,
+    /// how many of the pending operations matched the retry request pattern
+    pub matched: u64,
+}
+
+pub fn run_retry_request() -> io::Result<MessageRetryResponse> {
```
tkporter, line 19: can you add some comments here / call_retry_request so to just say what the intended behavior / use is here (e.g. mentioning that it retries all messages)

rust/main/utils/run-locally/src/main.rs
```diff
@@ -483,6 +484,15 @@ fn main() -> ExitCode {
     // give things a chance to fully start.
     sleep(Duration::from_secs(10));
 
+    // test retry request
+    let retry_req = std::thread::spawn(server::run_retry_request);
```
tkporter, line 488: do you need to `std::thread::spawn` / join? afaict run_retry_request isn't async and does some tokio orchestration so it could be called directly from a sync place like here

tkporter (commented):
mostly stylistic nits but I think the reading responses is prone to a race condition

rust/main/agents/relayer/src/server/message_retry.rs
daniel-savu: It'd have to be moved to `hyperlane-base` to be imported in both usage sites, but this is a relayer-specific data structure so may not fit that well in `-base`. I think we have similar duplication between agent code and e2e elsewhere too, cleanest would be to define a new crate where these types are defined but I personally was ok with this small duplication (CI would fail if the structs diverged)

rust/main/agents/relayer/src/server/message_retry.rs
daniel-savu: chatted with @kamiyaa about this, elsewhere in the codebase we use `try_recv` which behaves as you describe, but [recv](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv) only returns `None` when the channel is closed according to the docs (i.e. when all transmitters are dropped)

daniel-savu: funnily enough we chatted about that exact approach of counting submitters and responses but just using `recv` is simpler

rust/main/utils/run-locally/src/server.rs
daniel-savu: It'd have to be moved to `hyperlane-base` to be imported in both usage sites, but this is a relayer-specific data structure so may not fit that well in `-base`. I think we have similar duplication between agent code and e2e elsewhere too, cleanest would be to define a new crate where these types are defined but I personally was ok with this small duplication (CI would fail if the structs diverged)

rust/main/agents/relayer/src/server/message_retry.rs
daniel-savu: my bad, I had thought that this channel was used to notify recipients of the retry request for some reason (so single producer, multiple consumer), and suggested to jeff to set this to 1: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5043/files/a32c8dd0507db672fb1165ffcd5127a56f1bf17a#r1899741152

agree that with this being the channel used by submitters to send messages to the server it should have the length of the configured `relayChains`

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: good call, I think it was a remnant of previous changes

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: It's so we can create the router.

```rust
impl MessageRetryApi {
    pub fn router(&self) -> Router {
        Router::new()
            .route("/", routing::post(retry_message))
            .with_state(MessageRetryApiState {
                retry_request_transmitter: self.retry_request_transmitter.clone(),
            })
    }
```

But yeah, we can totally consolidate it

rust/main/utils/run-locally/src/main.rs
kamiyaa: Good call, I originally had this because I was running into errors that I thought were tokio runtime errors.
I just ran it again without the thread spawn and it worked so I will remove it

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: Yeah I can change it to debug. I don't think we planned to put the op in the log? Because it doesn't implement Display

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: Yes :+1: Thanks!

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: sounds good, I updated it to create a channel same size as `originChains.len()`

rust/main/agents/relayer/src/server/message_retry.rs
kamiyaa: > e.g. if like 2 out of 5 responses come in and we receive those 2 but still have 3 responses that haven't come in yet, we never get those 3 responses

I think in this case, the function will block until those 3 responses come in? Is that not the behavior we want?

rust/main/utils/run-locally/src/server.rs
tkporter: remind me why we can't import from the relayer in run-locally?

rust/main/agents/relayer/src/server/message_retry.rs
tkporter: ahh nice I didn't read the docs, good call this is way better

tkporter (approved):

rust/main/utils/run-locally/src/server.rs
daniel-savu: it looks like we already are actually, cc @kamiyaa 
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/99c91d8da34e417af536100cb4b726eae08b14c4/rust/main/utils/run-locally/Cargo.toml#L36

kamiyaa: Added! Good call!

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,27 +72,67 @@ impl OpQueue {
...
             .map(|Reverse(mut op)| {
-                if message_retry_requests.iter().any(|r| r.op_matches(&op)) {
+                let matched_requests: Vec<_> = message_retry_requests
+                    .iter_mut()
+                    .filter_map(|(retry_req, retry_response)| {
+                        // update retry metrics
+                        if retry_req.pattern.op_matches(&op) {
+                            debug!(uuid = retry_req.uuid, "Matched request");
+                            retry_response.matched += 1;
+                            Some(retry_req.uuid.clone())
```
tkporter, line 105: seems like this isn't used anymore? think we can go back to using `.any` instead of `filter_map` and avoid this allocation

rust/main/agents/relayer/src/server/message_retry.rs
```diff
@@ -1,32 +1,89 @@
...
-        Err(err) => format!("Failed to send retry request to the queue: {}", err),
+) -> Result<Json<MessageRetryResponse>, String> {
+    let uuid = uuid::Uuid::new_v4();
+    let uuid_string = uuid.to_string();
+
+    tracing::debug!(uuid = uuid_string, "Sending message retry request");
+
+    // This channel is only created to service this single
+    // retry request so we're expecting a single response
+    // from each transmitter end, hence we are using a channel of size 1
```
tkporter, line 45: nit: comment is stale

PR #5045, opened 2024-12-18
feat(warpMonitor): Always use collateral token symbol for building wa…

Mo-Hussain:
…rp routes id

### DIFF
--- a/typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
+++ b/typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
@@ -89,12 +89,18 @@ async function pollAndUpdateWarpRouteMetrics(
     chainMetadata,
     apiKey: await getCoinGeckoApiKey(),
   });
+  const collateralTokenSymbol = getWarpRouteCollateralTokenSymbol(warpCore);
 
   while (true) {
     await tryFn(async () => {
       await Promise.all(
         warpCore.tokens.map((token) =>
-          updateTokenMetrics(warpCore, token, tokenPriceGetter),
+          updateTokenMetrics(
+            warpCore,
+            token,
+            tokenPriceGetter,
+            collateralTokenSymbol,
+          ),
         ),
       );
     }, 'Updating warp route metrics');
@@ -107,6 +113,7 @@ async function updateTokenMetrics(
   warpCore: WarpCore,
   token: Token,
   tokenPriceGetter: CoinGeckoTokenPriceGetter,
+  collateralTokenSymbol: string,
 ) {
   const promises = [
     tryFn(async () => {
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
lgtm

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -310,6 +322,18 @@ async function getCoinGeckoApiKey(): Promise<string | undefined> {
   return apiKey;
 }
 
+function getWarpRouteCollateralTokenSymbol(warpCore: WarpCore): string {
+  const collateralToken = warpCore.tokens.find(
+    (token) =>
+      token.isCollateralized() ||
+      token.standard === TokenStandard.EvmHypXERC20Lockbox,
+  );
+  if (!collateralToken) {
```
tkporter, line 331: eventually we may have warp routes that are all synthetic, e.g. if someone issues their token natively as a warp route

could we maybe have the logic be "if all token symbols are consistent, use that, otherwise pick the collateral one. If there are multiple collateral symbols and they differ, arbitrarily use the symbol that is first alphabetically"

tkporter: also a comment describing why we need to have special logic would probably be good

Makes me wonder if the registry every gets in weird scenarios here with differring token symbols too

cmcewen (changes requested):
Tests not all passing

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -310,6 +322,48 @@ async function getCoinGeckoApiKey(): Promise<string | undefined> {
   return apiKey;
 }
 
+function getWarpRouteCollateralTokenSymbol(warpCore: WarpCore): string {
+  // We need to have a deterministic way to determine the symbol of the warp route
+  // as its used to identify the warp route in metrics. This method should support routes where:
+  // - All tokens have the same symbol, token standards can be all collateral, all synthetic or a mix)
+  // - All tokens have different symbol, but there is a collateral token to break the tie, where there are multiple collateral tokens, alphabetically first is chosen
+  // doesn't support routes where:
+  // - All tokens have different symbols and there is no collateral token to break the tie
```
tkporter, line 331: any thoughts on just arbitrarily picking the first alphabetical one?

I don't feel strongly though, this feels unlikely to be a case we hit and I'm happy to deal with it later if needed

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -310,6 +322,48 @@ async function getCoinGeckoApiKey(): Promise<string | undefined> {
...
+        (${Array.from(uniqueSymbols).join(
+          ', ',
+        )}) and no collateral token was found to break the tie`,
+    );
+  }
+
+  const collateralSymbols = collateralTokens.map((token) => token.symbol);
+  const uniqueCollateralSymbols = [...new Set(collateralSymbols)];
+
+  if (uniqueCollateralSymbols.length === 1) {
```
tkporter, line 360: probably easier to read if we just always do
```
return uniqueCollateralSymbols.sort()[0];
```

tkporter (approved):

PR #5047, opened 2024-12-18
feat: add pending transactions to strategy

yorhodes:
- **Fix strategy flag propagation**
- **Error on strategy flag with nonexistent file**
- **Add pending fetch to safe submitter**
- **Implement strategy pending command**
- **Use context strategy?**


cmcewen (changes requested):
Tests not all passing

PR #5051, opened 2024-12-19
feat(funding): Update desiredBalancePerChain for mainnet chains

Mo-Hussain:
### Description

- update desiredBalancePerChain inline with adjusted high urgency alerts that account for at least 2 days worth of transactions at current usage


typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -46,7 +46,7 @@ export const keyFunderConfig: KeyFunderConfig<
     boba: '0.05',
     bsc: '5',
     bsquared: '0.002',
-    celo: '3',
+    celo: '3.5',
```
paulbalaji, line 49: any reason not for a round 4?

typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -119,7 +119,7 @@ export const keyFunderConfig: KeyFunderConfig<
     taiko: '0.2',
     tangle: '2',
     telos: '100',
-    treasure: '800',
+    treasure: '900',
```
paulbalaji, line 122: why the increase?

Mo-Hussain: So I hadn't realised that key funder will only fund up to the desired balance when the balance is less than half the desired balance (400), this takes us close the alert threshold

typescript/infra/config/environments/mainnet3/funding.ts
Mo-Hussain: Can do

paulbalaji (approved):

PR #5062, opened 2024-12-20
fix(hyperlane-ethereum): only apply gas overrides when submitting

daniel-savu:
### Description

We want gas overrides (and the gas buffer) to be added to a tx just before submitting, to increase its chances of landing. However we want these to be excluded in the `prepare` step when just checking that the sender's gas payment is enough to approximately cover the estimated `gas_limit` returned by the destination chain RPC

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5109

### Backward compatibility

Yes

### Testing

will rely on e2e and will first release this to the RC context


rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -315,6 +315,7 @@ where
         &self,
         message: &HyperlaneMessage,
         metadata: &[u8],
+        apply_gas_overrides: bool,
```
kamiyaa, line 318: nit: maybe we can have a `ContractCallOptions` struct that holds different parameters. I think this makes the code a bit easier to read compared to a boolean function parameter.
But I also see some places that don't have `tx_gas_estimate`, so not sure.

```rust
#[derive(Clone, Debug)]
pub struct ContractCallOptions {
    pub apply_gas_overrides: bool,
    pub tx_gas_estimate: Option<U256>,
}

async fn process_contract_call(
    &self,
    message: &HyperlaneMessage,
    metadata: &[u8],
    options: &ContractCallOptions,
) -> ChainResult<ContractCall<M, ()>> {
    ...
}
```

cmcewen (dismissed):
Tests not all passing

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -355,7 +355,7 @@ impl PendingOperation for PendingMessage {
             if self
                 .ctx
                 .destination_mailbox
-                .process_estimate_costs(&self.message, metadata)
+                .process_estimate_costs(&self.message, metadata, true)
```
tkporter, line 358: any reason why this needs to be true?

iiuc this being true or false doesn't change the behavior we care about, which is just that an eth_estimateGas is successful. We don't care about the buffer being applied or anything here

I think if it doesn't matter, then we can keep the `process_estimate_costs` interface the same as before, and just have `process_contract_call` be the one with the interface change, and have `process_estimate_costs` always pass false as this param to `process_contract_call`. I think this would be kinda nice because it'd keep the Mailbox trait impl more self explanatory / simple

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -547,8 +554,11 @@ where
...
     ) -> ChainResult<TxCostEstimate> {
-        let contract_call = self.process_contract_call(message, metadata, None).await?;
+        let contract_call = self
+            .process_contract_call(message, metadata, apply_gas_overrides, None)
```
tkporter, line 560: I think if `apply_gas_overrides` is false, when we call `process_contract_call` we end up never calling eth_estimateGas, and we'd hit the `.ok_or(HyperlaneProtocolError::ProcessGasLimitRequired)?;` - I assume this is why e2e is failing

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: > iiuc this being true or false doesn't change the behavior we care about, which is just that an eth_estimateGas is successful. We don't care about the buffer being applied or anything here

in this specific call site we don't care about the gas limit value returned so it doesn't matter. In the other call site where we call this to check IGP policy, it does matter. `pending_message.rs` is VM-agnostic code so we can't have it be aware of which ethereum-specific method to call (like `process_contract_call`): https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/02c74ee103556e539eee1861511f8ab9cd9d0d56/rust/main/agents/relayer/src/msg/pending_message.rs#L287

To achieve cleaner interfaces we may want to add a `meets_igp_policy` method on the `PendingOperation` trait, but the implementation sounds every more complicated to me.

daniel-savu: I expect this particular call to change as a result of the submitter refactor anyway, so only see it as relatively short lived

tkporter: I mean that we can have the VM-agnostic interface of `process_estimate_costs` be "give me the cost estimate that I'm going to use when evaluating if the gas payment policy was met, and give an error if the tx will fail", and each VM implementation knows that. 

That way we don't need this `apply_gas_overrides: bool` in the VM-agnostic interface, because it doesn't need to be a param at all. We just have the ethereum `process_estimate_costs` always pass false here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5062/files#diff-5f49b3c719e488330745b8f07595760a4c875fc78e39621fdefb97bc0c8e6b63R560. No VM-agnostic logic would ever require it to be true, so why modify the interface to include it?

imo keeping it this way is more confusing in an already confusing setup of how we fill transaction params. Also imo more likely to cause confusion when it comes to future VM impls in here

The reason I'm pushing back harder is I think any change to the VM-agnostic interfaces should be heavily scrutinized, so even if you imagine it needing to change later there's also a possibility it won't

daniel-savu: ok i see the light now sorry. I thought this was the only entry point for setting the gas limit when building a tx, hence why I thought it's not possible to do it in a vm-agnostic way. Other places use `fill_tx_gas_params` or `process_contract_call` to set the gas limit. I think it's a shortcoming of the current design to have so many entrypoints that achieve the same thing, will try to improve this in the revamp

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
daniel-savu: fully agree this is better, i'm going for a less clean implementation here because this will certainly change after the submitter revamp

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: this test is a good example of the abstraction being overloaded to actually include buffers in some places: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/bef103b9068882df25a781c44164c55d93211d50/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs#L712

PR #4959, opened 2024-12-06
feat: SVM priority fee oracles and transaction submitters, more consistent tx submission

tkporter:
### Description

The fruition of https://www.notion.so/hyperlanexyz/Landing-Solana-Transactions-15a6d35200d680cbaf5eea5f9dbec1b1

Major highlights:
- Instead of using the max CUs, estimate CU usage and apply a 10% buffer
- Allow for priority fees to be configured. If not specified, defaults to a constant priority fee of 0. Supported options are:
  - a constant fee
  - the Helius priority fee API (https://docs.helius.dev/solana-apis/priority-fee-api)
- Allow for transaction submission to be configured. If not specified, defaults to the configured RPC url for the chain. Supported options are:
  - An arbitrary RPC URL, which will use priority fee auctions for fees
  - Jito, which will use a tip for fees
- In our own infra, we have logic to plumb through the Helius URL found at https://console.cloud.google.com/security/secret-manager/secret/mainnet3-rpc-endpoint-helius-solanamainnet/versions?project=abacus-labs-dev
- Signs txs with a finalized blockhash, as suggested by [How to Land Transactions on Solana](https://www.helius.dev/blog/how-to-land-transactions-on-solana)
- Partially implements `process_estimate_costs` to test that messages are processable. Does not yet actually enforce gas limits, because this will take some rollout effort. This is part of https://www.notion.so/hyperlanexyz/Alerting-on-critical-processing-issues-for-any-and-all-messages-15b6d35200d680a79d40fabc2e4e49da

PR #5063, opened 2024-12-20
feat: bump relayer memory usage to 20GB, start using new SVM submission in relayer

tkporter:
### Description

- deploys #4959
- Sets a small priority fee for Eclipse following some recent spikes there, https://discord.com/channels/935678348330434570/1320851852157255701 for context
- Makes the relayer request 20GB of memory, forcing relayers to have their own node in our bigger node pool (capacity 29 GB)
- Driveby to use G (Gigabytes) instead of Gi (Gibibytes) to match what GCP uses in their UI, also a bit more intuitive
- Added a `--concurrency <num>` flag that can be used when deploying agents. Especially useful when releasing to all validators, which we have tons of now.
- Concurrently refetching remote repositories on helm poses issues, so I made this not happen by default anymore. It's not configurable at the CLI level to enable it, but we have the ability to easily add it later. Because our agent deploy doesn't actually require any remote helm repos, it's not necessary 
- already rolled out


cmcewen (changes requested):
Tests not passing

daniel-savu (approved):

paulbalaji (approved):

cmcewen (approved):

PR #5064, opened 2024-12-20
fix: correct instructions in Rust readme for building docker image

anson-caldera:
### Description

Fix incorrect instructions in Rust readme for building the docker image. The build does not work as the readme dictates. Being in the rust folder gives a `ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref y1ezs2da6re8bcvxgo9psu1xk::hydxdsbeaxqx3rvjjzako1134: "/rust/main/config": not found` error. Being at the top level causes a successful build.

Also have to edit the build script to target the correct directory dockerfile.


tkporter (approved):
Thanks! lgtm

PR #5065, opened 2024-12-21
feat: Enable sending messages from a chain to itself

anson-caldera:
### Description

Currently, sending messages from a chain to itself is not permitted. This enables it.


rust/build.sh
```diff
@@ -19,4 +19,4 @@ if [[ -z $TAG ]]; then
   fi
 fi
 
-DOCKER_BUILDKIT=1 docker build $PLATFORM -t gcr.io/abacus-labs-dev/hyperlane-agent:$TAG .
+DOCKER_BUILDKIT=1 docker build -f rust/Dockerfile $PLATFORM -t gcr.io/abacus-labs-dev/hyperlane-agent:$TAG .
```
paulbalaji, line 22: instead of changing this by hardcoding `-f rust/Dockerfile`, add the following after the shebang at the top
```sh
# Change directory to where script is
cd "$(dirname "$0")"
```

so you'll be able to run rust/build.sh from whatever path you're at

.changeset/hot-baboons-yell.md
```diff
@@ -0,0 +1,5 @@
+---
```
tkporter, line 1: can you actually remove this file? `@hyperlane-xyz/core` is meant to be just for our Solidity contracts - we don't have a changeset setup for the Rust side atm

tkporter (approved):
thanks 🙏

.changeset/hot-baboons-yell.md
anson-caldera: done

paulbalaji (approved):

PR #5071, opened 2024-12-24
fix: fix SVM scraper dispatch and delivery bugs

tkporter:
### Description

Context here https://discord.com/channels/935678348330434570/1320897742779977849

Fixes two bugs, adding tests for them:
- Txs that had reverted with an error are no longer considered valid in the scraper's log meta population codepath. We ran into an instance where two attempted process txs landed in the same block. We'd get two potentially matching txs in the same block when indexing the process, which causes issues
- Versioned txs were not supported. We ran into a versioned dispatch tx for the first time (https://eclipsescan.xyz/tx/Ku7gtQzEztksncHMHYkXq8PNst5rZsznrXMDmVqoQjc7SR4PsXz33qjNMckNcHriCxuhZJGZequo8moNsRP9GZK). The list of accounts wasn't populated correctly because some were dynamic (from an address lookup table). This caused us to think that the Mailbox wasn't being used in a tx, when it was. The solution here is to ensure the list is ordered correctly and has all the accounts, including the dynamic ones.


rust/main/chains/hyperlane-sealevel/src/log_meta_composer.rs
```diff
@@ -238,9 +243,30 @@ fn filter_by_validity(
...
     };
 
+    // Orders the account keys in line with the behavior of compiled instructions.
+    // If the meta specifies loaded (i.e. dynamic) addresses,
+    let account_keys = match &meta.loaded_addresses {
+        OptionSerializer::Some(addresses) => {
+            // If there are loaded addresses, we have a versioned transaction
+            // that includes dynamically loaded addresses (e.g. from a lookup table).
+            // The order of these is [static, dynamic writeable, dynamic readonly] and
+            // follows the iter ordering of https://docs.rs/solana-sdk/latest/solana_sdk/message/struct.AccountKeys.html.
```
kamiyaa, line 253: Nice link to docs :+1:

kamiyaa (approved):
Looks good :+1:

rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
```diff
@@ -49,6 +69,27 @@ pub fn test_search_delivered_message_transaction() {
     assert!(!transaction_hashes.is_empty());
 }
 
+#[test]
+pub fn test_search_delivered_message_reverted_transaction() {
```
daniel-savu, line 73: nice, would be good to also test the multi-tx per block case, where 1 was successful and 1+ failed

daniel-savu (approved):

rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
tkporter: added

PR #5074, opened 2024-12-26
feat(relayer): relayer exposes latest tree index as metric

aroralanuk:
### Description

- Added `latest_tree_insertion_count` to CoreMetrics which can then be imported to Grafana. 
- In the `MerkleTreeProcessor`, we set this metric via `max_leaf_index_gauge` to reflect the latest leaf index processed and inserted locally to the tree.

### Drive-by changes

none

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1390

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/rust/main/agents/relayer/src/merkle_tree/processor.rs
+++ b/rust/main/agents/relayer/src/merkle_tree/processor.rs
@@ -74,6 +74,7 @@ impl MerkleTreeProcessor {
             .retrieve_merkle_tree_insertion_by_leaf_index(&self.leaf_index)?
         {
             // Update the metrics
+            // we assume that leaves are inserted in order so this will be monotonically increasing
             self.metrics
                 .max_leaf_index_gauge
                 .set(insertion.index() as i64);
--- a/rust/main/agents/relayer/src/merkle_tree/processor.rs
+++ b/rust/main/agents/relayer/src/merkle_tree/processor.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp::max,
     fmt::{Debug, Formatter},
     sync::Arc,
     time::Duration,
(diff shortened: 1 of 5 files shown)


rust/main/hyperlane-base/src/metrics/core.rs
```diff
@@ -112,6 +113,16 @@ impl CoreMetrics {
             registry
         )?;
 
+        let latest_tree_insertion_count = register_int_gauge_vec_with_registry!(
+            opts!(
+                namespaced!("latest_tree_insertion_count"),
```
tkporter, line 118: as mentioned irl, could see this being interpreted as "how many leaves were inserted in the tree since the relayer started up" so would prefer something like `max_merkle_tree_index` or something

aroralanuk: ya, renamed it to `latest_leaf_index` which is closer to the associated gauge `max_leaf_index_gauge` in MerkleTreeProcessorMetrics here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5074/commits/ef8afa0beccbc96426506752a9616b1d00be72d1

rust/main/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -89,13 +93,11 @@ pub struct MerkleTreeProcessorMetrics {
...
 impl MerkleTreeProcessorMetrics {
-    pub fn new() -> Self {
+    pub fn new(metrics: &CoreMetrics, origin: &HyperlaneDomain) -> Self {
         Self {
-            max_leaf_index_gauge: IntGauge::new(
-                "max_leaf_index_gauge",
-                "The max merkle tree leaf index",
-            )
-            .unwrap(),
+            max_leaf_index_gauge: metrics
```
tkporter, line 98: nit: don't feel too strongly but flagging that `max_leaf_index_gauge` may be nice to rename to `latest_leaf_index` as well for consistency

tkporter (approved):

PR #5075, opened 2024-12-26
fix(agents/sealevel): handle non configured igp account payments

aroralanuk:
### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1392

### Backward compatibility

Yes

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
@@ -251,7 +251,17 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
                 if igp_account_filter == sealevel_payment.igp_account_pubkey {
                     payments.push((sealevel_payment.payment, sealevel_payment.log_meta));
                 } else {
-                    tracing::debug!(sealevel_payment=?sealevel_payment, igp_account_filter=?igp_account_filter, "Found interchain gas payment for a different IGP account, skipping");
+                    tracing::debug!(sealevel_payment=?sealevel_payment, igp_account_filter=?igp_account_filter, "Found interchain gas payment for a different IGP account, neutralizing payment");
+
+                    let non_matching_payment = InterchainGasPayment {
+                        gas_amount: U256::from(0),
+                        payment: U256::from(0),
+                        ..*sealevel_payment.payment.inner()
+                    };
+                    payments.push((
+                        Indexed::new(non_matching_payment).with_sequence(nonce.into()),
+                        sealevel_payment.log_meta,
+                    ));
                 }
             }
         }


rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -251,7 +251,17 @@ impl Indexer<InterchainGasPayment> for SealevelInterchainGasPaymasterIndexer {
                 if igp_account_filter == sealevel_payment.igp_account_pubkey {
                     payments.push((sealevel_payment.payment, sealevel_payment.log_meta));
                 } else {
-                    tracing::debug!(sealevel_payment=?sealevel_payment, igp_account_filter=?igp_account_filter, "Found interchain gas payment for a different IGP account, skipping");
+                    tracing::debug!(sealevel_payment=?sealevel_payment, igp_account_filter=?igp_account_filter, "Found interchain gas payment for a different IGP account, neutralizing payment");
+
+                    let non_matching_payment = InterchainGasPayment {
```
tkporter, line 256: getting a little hard to read, could we change the logic around here starting with line 249 to be something like:
```
if let Ok(mut sealevel_payment) = self.get_payment_with_sequence(nonce.into()).await {
  // Comment explaining why we do this
  if sealevel_payment.igp_account_pubkey != self.igp.igp_account {
    sealevel_payment.gas_amount = U256::zero();
    sealevel_payment.payment = U256::zero();
  }
  payments.push((sealevel_payment.payment, sealevel_payment.log_meta));
}
```

tkporter (commented):
thought about it, I'm down to do this for now but let's add a comment clearly flagging why we did this / indicating we may want to change in the future if we want to index multiple IGPs at a time

rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
aroralanuk: done

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
```diff
@@ -128,11 +136,11 @@ pub fn termination_invariants_met(
 
     // TestSendReceiver randomly breaks gas payments up into
     // two. So we expect at least as many gas payments as messages.
-    if gas_payment_events_count < total_messages_expected {
+    if gas_payment_events_count < total_messages_expected + sol_messages_with_non_matching_igp {
```
ameten, line 139: I wonder if we can include `sol_messages_with_non_matching_igp` into `total_messages_expected`? Or is it not possible due to some other invariants?

ameten (approved):

rust/main/utils/run-locally/src/solana.rs
```diff
@@ -341,6 +381,56 @@ pub fn initiate_solana_hyperlane_transfer(
     message_id
 }
 
+#[apply(as_task)]
+pub fn initiate_solana_non_igp_paying_transfer(
```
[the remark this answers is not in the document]
aroralanuk: No, this should be initiate_solana_non_matching_igp_paying_transfer. Fixed, good catch!

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
aroralanuk: `total_messages_expected` is checked against at multiple places, like `delivered_messages_scraped`. I have added a comment "this is total messages expected to be delivered" to avoid any confusion.

PR #5077, opened 2024-12-26
test: cli e2e warp init command tests

xeno097:
### Description

Adds e2e tests for the `warp init` command

### Drive-by changes

- 

### Related issues

- 

### Backward compatibility

- YES

### Testing

- E2E

### DIFF
--- a/typescript/cli/src/tests/commands/helpers.ts
+++ b/typescript/cli/src/tests/commands/helpers.ts
@@ -1,5 +1,5 @@
 import { ethers } from 'ethers';
-import { $ } from 'zx';
+import { $, ProcessPromise } from 'zx';
 
 import { ERC20Test__factory, ERC4626Test__factory } from '@hyperlane-xyz/core';
 import { ChainAddresses } from '@hyperlane-xyz/registry';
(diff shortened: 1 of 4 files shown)


typescript/cli/src/tests/commands/helpers.ts
```diff
@@ -44,6 +44,39 @@ export const WARP_CORE_CONFIG_PATH_2 = `${REGISTRY_PATH}/deployments/warp_routes
...
+  await sleep(500);
+}
+
+export async function selectAnvil2AndAnvil3(
+  stream: ProcessPromise,
+): Promise<void> {
+  // Scroll down through the mainnet chains list and select anvil2
+  await asyncStreamInputWrite(
+    stream.stdin,
+    `${KeyBoardKeys.ARROW_DOWN.repeat(3)}${KeyBoardKeys.TAB}`,
```
ltyu, line 69: will this change as we add more chains?

ltyu (approved):
lgtm!

typescript/cli/src/tests/commands/helpers.ts
xeno097: it shouldn't because we override the registry for the tests and we only have `anvil1`, `anvil2` and `anvil3` in the custom registry for the tests

PR #5090, opened 2024-12-30
test(cli): core commands e2e tests

xeno097:
### Description

Adds more E2E tests to the core commands in the CLI 

### Drive-by changes

- Removes the unused `--skip-confirmation` from the `core deploy` command
- Updates the `warp init` tests to use the `handlePrompts` function 
- Moves the core e2e tests into the `test/core` folder

### Related issues

- #5080 

### Backward compatibility

- YES
### Testing

- Manual
- E2E

### DIFF
--- a/typescript/cli/src/tests/commands/core.ts
+++ b/typescript/cli/src/tests/commands/core.ts
@@ -14,7 +14,16 @@ export function hyperlaneCoreDeployRaw(
   coreInputPath: string,
   privateKey?: string,
   skip?: boolean,
+  hyp_key?: string,
 ): ProcessPromise {
+  if (hyp_key) {
+    return $`HYP_KEY=${hyp_key} yarn workspace @hyperlane-xyz/cli run hyperlane core deploy \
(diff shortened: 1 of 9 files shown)


typescript/cli/src/tests/commands/core.ts
```diff
@@ -1,11 +1,45 @@
...
 
 import { ANVIL_KEY, REGISTRY_PATH } from './helpers.js';
 
+/**
+ * Deploys the Hyperlane core contracts to the specified chain using the provided config.
+ */
+export function hyperlaneCoreDeployRaw(
+  coreInputPath: string,
+  privateKey?: string,
+  skip?: boolean,
```
jmrossy, line 16: Maybe give this a more clear name? In the context of a test, skip typically means to skip the test

jmrossy (approved):
😎

typescript/cli/src/tests/commands/core.ts
xeno097: will do! makes sense 🔥

PR #5100, opened 2025-01-02
fix: use middleware submission tweaks

daniel-savu:
### Description

Uses the short-term fixes described in this [design doc](https://www.notion.so/hyperlanexyz/State-of-EVM-provider-middleware-16c6d35200d680d8a6b5f0d32cd8c66b?pvs=4) and implemented in https://github.com/hyperlane-xyz/ethers-rs/pull/23

These fixes are expected to make the relayer self-recover from nonce gaps. 

An expected drawback is that tx hashes are dropped more often due to the nonce manager resyncing to nonces that end up being too low by 1, which can slow down submissions. After 12h of testing, any submit queue spikes were only 1 min long and the biggest one had 7 messages - so short enough not to trigger the "rising submit queue" alert which stays pending for 6min: https://abacusworks.grafana.net/alerting/de6hwplr7eakgc

### Drive-by changes
- Removes verbosity from a bunch of logs (we had `HyperlaneMessage` in a single log occur 6-7 times, but wasn't noticeable in `span`s because it was originating from args named `message` which have a clash with inner workings of the `tracing` lib)
- rewords some logs because they were misleading

### Related issues
Fixes: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5107

### Backward compatibility
Yes

### Testing
E2E with a gas escalator that randomly dropped every other tx - the nonce manager was able to recover but was slower to submit. Then evaluated on RC for 12h as described above.


rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -189,7 +189,7 @@ impl Deref for MessageMetadataBuilder {
 
 #[async_trait]
 impl MetadataBuilder for MessageMetadataBuilder {
-    #[instrument(err, skip(self), fields(destination_domain=self.destination_domain().name()))]
+    #[instrument(err, skip(self, message), fields(destination_domain=self.destination_domain().name()))]
```
tkporter, line 192: i assume we'd generally still have the message somewhere in the spans?

tkporter (approved):

rust/main/agents/relayer/src/msg/metadata/base.rs
daniel-savu: we do yes, confirmed with e2e and RC logs that the lifecycle is still trackable via the message ID: https://cloudlogging.app.goo.gl/Koyyzv992JQh6cp8A

PR #5103, opened 2025-01-03
feat: bump overhead when sending to zkSync chains

tkporter:
### Description

Context: https://discord.com/channels/935678348330434570/1319115367167033399/1319442102454849639

zkSync chains have non standard EVM gas metering, and their eth_estimateGas RPC returns abnormally high estimates compared to the actual usage. As a semi-temporary workaround, we bump the overhead to result in more gas paid for. The impact of this is actually not meaningful when it comes to cost borne by users - because the zkSync chains are cheap, our min fee logic means that applying this still results in a ~20c cost to users

### DIFF
--- a/typescript/infra/config/environments/mainnet3/igp.ts
+++ b/typescript/infra/config/environments/mainnet3/igp.ts
@@ -1,6 +1,7 @@
 import {
   ChainMap,
   ChainName,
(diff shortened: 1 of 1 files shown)


yorhodes (approved):
clean

paulbalaji (approved):

PR #5110, opened 2025-01-06
feat: deploy SOON and warp routes

tkporter:
### Description

- core and SOL / Bonk warp routes on SOON


typescript/infra/config/environments/mainnet3/gasPrices.json
```diff
@@ -16,23 +16,23 @@
     "decimals": 9
```
tkporter, line 16: I didn't actually apply any changes onchain here or in tokenPrices yet, happy to undo these changes and just keep the SOON diff if that makes sense

rust/sealevel/environments/mainnet3/solanamainnet/gas-oracle-configs-solanamainnet.json
```diff
@@ -7,5 +7,14 @@
             "gasPrice": "2",
             "tokenDecimals": 9
         }
+    },
+    {
+        "domain": 50075007,
```
tkporter, line 12: used the same data as eclipse, given they are both ETH gas token

rust/sealevel/environments/mainnet3/warp-routes/SOL-solanamainnet-soon/token-config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "solanamainnet": {
+    "type": "native",
+    "decimals": 9,
+    "interchainGasPaymaster": "AkeHBbE5JkwVppujCQQ6WuxsVsJtruBAjUo6fDCFp6fF",
+    "interchainSecurityModule": "HPNQHcc7wRcwsPG1aVBpPgpYuV7h2MJiVgKnAgTdDRBM"
```
tkporter, line 6: will be moving away from this to the default ISM eventually

typescript/infra/config/environments/mainnet3/gasPrices.json
paulbalaji: yeah would be good to keep just the SOON diff here - avoids confusion when looking back in history since i'll be updating this in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5116 too

rust/sealevel/environments/mainnet3/chain-config.json
```diff
@@ -37,7 +37,7 @@
     ],
     "technicalStack": "opstack"
   },
-  "alephzeroevm": {
+  "alephzeroevmmainnet": {
```
paulbalaji, line 40: damn this is an old change finally being propagated

paulbalaji (commented):
lgtm once gas/token price diff is minimised

paulbalaji (approved):

Issue #4887, opened 2024-11-21
Relayer can still crash upon startup due to bad RPCs


tkporter:
## Problem

Despite most relayer crashes being fixed (the ones to do with contract sync building), it's still possible to fail e.g. here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/58425a2de3578bbe01a1ff81441fdfa3caa023b4/rust/main/agents/relayer/src/relayer.rs#L138-L143

## Solution

We probably don't want to take down everything if one chain is having issues

kamiyaa:
Talked with @daniel-savu about this

Takeaways:
1. Ensure no `unwrap()`s in tasks the relayer spawns
2. replace `try_join_all` with `join_all` perhaps. Such that 1 failed future does not affect all futures
 - potentially have some sort of iterator pattern that yields `Ok/Err` for these futures
 - maybe https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html
3. Unittesting panic!() s in all tasks spawned by relayer to ensure liveliness is not lost when a single task dies

PR #5115, opened 2025-01-07
feat: don't crash relayer due to flaky RPCs

kamiyaa:
### Description
 - instead of crashing relayer if any RPC calls fail during `build_mailboxes` and `build_validator_announces`, we just log the error and metric
 - relayer will continue running without those chains

### Related issues
Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4887

### Backward compatibility

Are these changes backward compatible? Are there any infrastructure implications, e.g. changes that would prohibit deploying older commits using this infra tooling?

Yes

### Testing

Added unittests to test ensure functions run correctly and log errors in metrics

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -18,8 +18,8 @@ use hyperlane_base::{
 };
 use hyperlane_core::{
     rpc_clients::call_and_retry_n_times, ChainCommunicationError, ContractSyncCursor,
-    HyperlaneDomain, HyperlaneMessage, InterchainGasPayment, MerkleTreeInsertion, QueueOperation,
-    H512, U256,
+    HyperlaneDomain, HyperlaneMessage, InterchainGasPayment, Mailbox, MerkleTreeInsertion,
+    QueueOperation, ValidatorAnnounce, H512, U256,
 };
 use tokio::{
     sync::{
@@ -135,12 +135,10 @@ impl BaseAgent for Relayer {
             .map(|origin| (origin.clone(), HyperlaneRocksDB::new(origin, db.clone())))
             .collect::<HashMap<_, _>>();
 
-        let mailboxes = settings
-            .build_mailboxes(settings.destination_chains.iter(), &core_metrics)
-            .await?;
-        let validator_announces = settings
-            .build_validator_announces(settings.origin_chains.iter(), &core_metrics)
-            .await?;
+        let mailboxes = Self::build_mailboxes(&settings, &core_metrics, &chain_metrics).await;
+
+        let validator_announces =
+            Self::build_validator_announces(&settings, &core_metrics, &chain_metrics).await;
 
         let contract_sync_metrics = Arc::new(ContractSyncMetrics::new(&core_metrics));
 
@@ -236,7 +234,9 @@ impl BaseAgent for Relayer {
 
         let mut msg_ctxs = HashMap::new();
         let mut destination_chains = HashMap::new();
-        for destination in &settings.destination_chains {
+
+        // only iterate through destination chains that were successfully instantiated
+        for (destination, dest_mailbox) in mailboxes.iter() {
             let destination_chain_setup = core.settings.chain_setup(destination).unwrap().clone();
             destination_chains.insert(destination.clone(), destination_chain_setup.clone());
             let transaction_gas_limit: Option<U256> =
@@ -246,18 +246,19 @@ impl BaseAgent for Relayer {
                     transaction_gas_limit
                 };
 
-            for origin in &settings.origin_chains {
+            // only iterate through origin chains that were successfully instantiated
+            for (origin, validator_annonce) in validator_announces.iter() {
                 let db = dbs.get(origin).unwrap().clone();
                 let metadata_builder = BaseMetadataBuilder::new(
                     origin.clone(),
                     destination_chain_setup.clone(),
                     prover_syncs[origin].clone(),
-                    validator_announces[origin].clone(),
+                    validator_annonce.clone(),
                     settings.allow_local_checkpoint_syncers,
                     core.metrics.clone(),
                     db,
                     IsmAwareAppContextClassifier::new(
-                        mailboxes[destination].clone(),
+                        dest_mailbox.clone(),
                         settings.metric_app_contexts.clone(),
                     ),
                 );
@@ -268,7 +269,7 @@ impl BaseAgent for Relayer {
                         destination: destination.id(),
                     },
                     Arc::new(MessageContext {
-                        destination_mailbox: mailboxes[destination].clone(),
+                        destination_mailbox: dest_mailbox.clone(),
                         origin_db: dbs.get(origin).unwrap().clone(),
                         metadata_builder: Arc::new(metadata_builder),
                         origin_gas_payment_enforcer: gas_payment_enforcers[origin].clone(),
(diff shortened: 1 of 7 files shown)


rust/main/agents/relayer/src/relayer.rs
```diff
@@ -246,18 +246,19 @@ impl BaseAgent for Relayer {
...
-            for origin in &settings.origin_chains {
+            // only iterate through origin chains that were successfully instantiated
+            for (origin, validator_annonce) in validator_announces.iter() {
                 let db = dbs.get(origin).unwrap().clone();
                 let metadata_builder = BaseMetadataBuilder::new(
                     origin.clone(),
                     destination_chain_setup.clone(),
                     prover_syncs[origin].clone(),
-                    validator_announces[origin].clone(),
+                    validator_annonce.clone(),
```
kamiyaa, line 256: Since we can no longer assume all `mailboxes / validator_announce` were successfully instantiated, we should loop through the successfully instantiated ones instead of blindly going through all `origin_chains / destination_chains`

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
         }))
         .instrument(span)
     }
+
+    /// Helper function to build and return a hashmap of mailboxes.
+    /// Any chains that fail to build mailbox will not be included
+    /// in the hashmap. Errors will be logged and chain metrics
+    /// will be updated for chains that fail to build mailbox.
+    pub async fn build_mailboxes(
```
kamiyaa, line 629: Move logic into functions so we can test them in isolation

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
...
+            transaction_gas_limit: None,
+            skip_transaction_gas_limit_for: HashSet::new(),
+            allow_local_checkpoint_syncers: true,
+            metric_app_contexts: Vec::new(),
+        }
+    }
+
+    #[tokio::test]
+    #[tracing_test::traced_test]
+    async fn test_failed_build_mailboxes() {
```
kamiyaa, line 803: Turns out we don't need to do too many fancy things to cause an error for `build_mailboxes` or `build_validator_announces`.
All we need to do is just not have a `ChainConf` associated with said chain and it will cause an error in `build_mailbox / build_validator_announce`

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
...
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Helper function to build and return a hashmap of vaidator announces.
+    /// Any chains that fail to build vaidator announce will not be included
+    /// in the hashmap. Errors will be logged and chain metrics
+    /// will be updated for chains that fail to build vaidator announce.
```
ameten, line 653:
```suggestion
    /// Helper function to build and return a hashmap of validator announces.
    /// Any chains that fail to build validator announce will not be included
    /// in the hashmap. Errors will be logged and chain metrics
    /// will be updated for chains that fail to build validator announce.
```

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
...
+    ) -> HashMap<HyperlaneDomain, Arc<dyn Mailbox>> {
+        settings
+            .build_mailboxes(settings.destination_chains.iter(), &core_metrics)
+            .await
+            .into_iter()
+            .filter_map(|(origin, mailbox_res)| match mailbox_res {
+                Ok(mailbox) => Some((origin, mailbox)),
+                Err(err) => {
+                    let message = err.to_string();
+                    error!(?err, origin=?origin, "{message}");
```
ameten, line 642: Since we are going to have error serialised into a field in logs, would it make sense to give a more high level error message here, for example, an error message which will be independent from the internal error so that we can easily search by it in the logs?

ameten (approved):

rust/main/agents/relayer/src/relayer.rs
tkporter: agreed, can you have the log message be a literal string like "Critical error when building mailbox" or something? here and in the validator announce one. Having `?err` is enough for us

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -135,12 +135,10 @@ impl BaseAgent for Relayer {
             .map(|origin| (origin.clone(), HyperlaneRocksDB::new(origin, db.clone())))
             .collect::<HashMap<_, _>>();
 
-        let mailboxes = settings
-            .build_mailboxes(settings.destination_chains.iter(), &core_metrics)
```
tkporter, line 139: are these used anymore? https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/hyperlane-base/src/settings/base.rs#L148-L152

if not just delete imo

tkporter (approved):

rust/main/agents/relayer/src/relayer.rs
kamiyaa: sounds good!

rust/main/agents/relayer/src/relayer.rs
kamiyaa: Still being used :+1:

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
...
+    ) -> HashMap<HyperlaneDomain, Arc<dyn ValidatorAnnounce>> {
+        settings
+            .build_validator_announces(settings.origin_chains.iter(), &core_metrics)
+            .await
+            .into_iter()
+            .filter_map(|(origin, mailbox_res)| match mailbox_res {
+                Ok(mailbox) => Some((origin, mailbox)),
+                Err(err) => {
+                    let message = err.to_string();
+                    error!(?err, origin=?origin, "{message}");
```
tkporter, line 667: can you do this one too

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -620,7 +621,277 @@ impl Relayer {
...
+        chain_metrics: &ChainMetrics,
+    ) -> HashMap<HyperlaneDomain, Arc<dyn Mailbox>> {
+        settings
+            .build_mailboxes(settings.destination_chains.iter(), &core_metrics)
+            .await
+            .into_iter()
+            .filter_map(|(origin, mailbox_res)| match mailbox_res {
+                Ok(mailbox) => Some((origin, mailbox)),
+                Err(err) => {
+                    let message = err.to_string();
```
tkporter, line 641: and rm this now that it's unused

kamiyaa: yep, thanks!

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -246,18 +246,19 @@ impl BaseAgent for Relayer {
                     transaction_gas_limit
                 };
 
-            for origin in &settings.origin_chains {
+            // only iterate through origin chains that were successfully instantiated
+            for (origin, validator_annonce) in validator_announces.iter() {
```
tkporter, line 250:
```suggestion
            for (origin, validator_announce) in validator_announces.iter() {
```

rust/main/agents/relayer/src/relayer.rs
tkporter, line 256:
```suggestion
                    validator_announce.clone(),
```

tkporter (approved):
typo nit

PR #5118, opened 2025-01-07
feat: hyperlane warp send all chains

ltyu:
### Description
Update `hyperlane warp send` to send a round trip transfer to all chains in WarpCoreConfig, if `--origin` and/or `--destination` is not provided.

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -287,7 +287,7 @@ const send: CommandModuleWithWriteContext<
 
     let chains: ChainName[] = warpCoreConfig.tokens.map((t) => t.chainName);
     if (roundTrip) {
-      // Appends the reverse of the array to roundtrip, excluding the 1st
+      // Appends the reverse of the array to roundtrip, excluding the 1st (e.g. [1,2,3] becomes [1,2,3,2,1])
       // We make a copy because .reverse() is mutating
       const reversed = [...chains].reverse().slice(1, chains.length + 1);
       chains.push(...chains, ...reversed);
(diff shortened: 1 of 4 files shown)


typescript/cli/src/send/transfer.ts
```diff
@@ -40,51 +38,41 @@ export async function sendTestTransfer({
...
+  chains: ChainName[];
   amount: string;
   recipient?: string;
   timeoutSec: number;
   skipWaitForDelivery: boolean;
   selfRelay?: boolean;
 }) {
-  const { chainMetadata } = context;
-
-  if (!origin) {
```
ltyu, line 53: Deleted because this does not really work when an origin/destination is provided, but not part of the WarpCoreConfig. Also, can just use `--origin` and `--destination`.

typescript/cli/src/commands/warp.ts
```diff
@@ -275,24 +277,55 @@ const send: CommandModuleWithWriteContext<
...
+    let chains: ChainName[] = warpCoreConfig.tokens.map((t) => t.chainName);
+    if (roundTrip) {
+      // Appends the reverse of the array, excluding the 1st (e.g. [1,2,3] becomes [1,2,3,2,1])
+      const reversed = [...chains].reverse().slice(1, chains.length + 1); // We make a copy because .reverse() is mutating
+      chains.push(...reversed);
+    } else {
+      // Assume we want to use use `--origin` and `--destination` params, prompt as needed.
+      const chainMetadata = objFilter(
+        context.chainMetadata,
+        (key, _metadata): _metadata is any => chains.includes(key),
```
xeno097, line 297: Maybe a small optimization but I think it would be better to create a set and use the `.has` method to be a bit more efficient here

xeno097 (approved):

PR #5120, opened 2025-01-07
Manually tested infra secrets + CLI http registry

yorhodes:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "undici": "^5.11",
     "@trivago/prettier-plugin-sort-imports/@babel/parser": "^7.22.7",
     "@typechain/ethers-v5": "11.1.2",
-    "typechain@npm:^8.0.0": "patch:typechain@npm%3A8.3.2#~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch"
+    "typechain@npm:^8.0.0": "patch:typechain@npm%3A8.3.2#~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch",
+    "@hyperlane-xyz/registry": "portal:/Users/yorhodes/hyperlane/registry"
   }
 }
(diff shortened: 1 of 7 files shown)


typescript/cli/src/utils/files.ts
```diff
@@ -92,7 +92,7 @@ export function mergeJson<T extends Record<string, any>>(
 }
 
 export function readYaml<T>(filepath: string): T {
-  return yamlParse(readFileAtPath(filepath)) as T;
+  return yamlParse(readFileAtPath(filepath), { maxAliasCount: -1 }) as T;
```
yorhodes, line 95: will pull this into an atomic PR

package.json
```diff
@@ -52,6 +52,7 @@
     "undici": "^5.11",
     "@trivago/prettier-plugin-sort-imports/@babel/parser": "^7.22.7",
     "@typechain/ethers-v5": "11.1.2",
-    "typechain@npm:^8.0.0": "patch:typechain@npm%3A8.3.2#~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch"
+    "typechain@npm:^8.0.0": "patch:typechain@npm%3A8.3.2#~/.yarn/patches/typechain-npm-8.3.2-b02e27439e.patch",
+    "@hyperlane-xyz/registry": "portal:/Users/yorhodes/hyperlane/registry"
```
yorhodes, line 56: will need to fix this
actually `yarn link` worked pretty well for me here

typescript/infra/scripts/http-registry.ts
```diff
@@ -0,0 +1,25 @@
...
+async function main() {
+  const { environment } = await getArgs().argv;
+
+  const registry =
+    environment === 'mainnet3'
+      ? await getMainnet3Registry()
+      : await getTestnet4Registry();
+
+  const server = new HttpServer(registry);
+  await server.start();
```
paulbalaji, line 17: oooooh I see, I like this a lot

typescript/cli/src/utils/files.ts
Mo-Hussain: Why is the context for this change? Going to remove it from this PR for now

PR #5122, opened 2025-01-08
feat: Add Superseed warp route config

nambrot:
### Description

Includes the warp route config for SuperSeed FiatToken warp routes

### Drive-by changes

- Add dummy soon owner
- Indicate errors during warp config parsing


typescript/infra/src/config/warp.ts
```diff
@@ -1,8 +1,7 @@
-import { ChainMap, OwnableConfig, RouterConfig } from '@hyperlane-xyz/sdk';
-import { Address } from '@hyperlane-xyz/utils';
+import { OwnableConfig, RouterConfig } from '@hyperlane-xyz/sdk';
 
 // Common collateral tokens to be used by warp route deployments.
-export const tokens: ChainMap<Record<string, Address>> = {
+export const tokens = {
```
cmcewen, line 4: what was the error you were getting there?

cmcewen (approved):

typescript/infra/src/config/warp.ts
nambrot: None, it just seems superflous

typescript/infra/scripts/warp-routes/generate-warp-config.ts
```diff
@@ -24,7 +24,9 @@ async function main() {
   const parsed = WarpRouteDeployConfigSchema.safeParse(warpConfig);
 
   if (!parsed.success) {
-    rootLogger.error(parsed.error.format());
+    rootLogger.error('Error parsing warp config:');
+    rootLogger.dir(warpConfig, { depth: null });
+    rootLogger.dir(parsed.error.format(), { depth: null });
```
paulbalaji, line 29: i think you need to use console.dir here

PR #5132, opened 2025-01-09
feat: return # of messages prioritized in the queue when calling message retry endpoint

kamiyaa:
### Description

Add a mpsc channel between relayer and message retry handler, so message retry handler knows the retry results from the relayer.

`POST /message_retry` endpoint now returns a JSON with details on the retry request:

```rust
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct MessageRetryResponse {
    /// ID of the retry request
    pub uuid: String,
    /// how many pending operations were processed
    pub processed: usize,
    /// how many of the pending operations matched the retry request pattern
    pub matched: u64,
}
```

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4059

### Testing

Updated unit tests to return a response so the handler is not blocked


rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,27 +72,66 @@ impl OpQueue {
...
+                message_retry_requests
+                    .iter_mut()
+                    .for_each(|(retry_req, retry_response)| {
+                        if !retry_req.pattern.op_matches(&op) {
+                            return;
+                        }
+                        debug!(
+                            uuid = retry_req.uuid,
+                            ?op,
+                            pattern = ?retry_req.pattern, "Matched retry request");
```
kamiyaa, line 110: Added more informative debug messages so we know which requests and patterns are being matched.
I hope this isn't the reason why the endpoint is slow now

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,27 +72,66 @@ impl OpQueue {
         // The other consideration is whether to put the channel receiver in the OpQueue or in a dedicated task
         // that also holds an Arc to the Mutex. For simplicity, we'll put it in the OpQueue for now.
         let mut message_retry_requests = vec![];
-        while let Ok(message_id) = self.retry_rx.lock().await.try_recv() {
-            message_retry_requests.push(message_id);
+
+        // we only need to lock self.retry_receiver once
+        {
+            let mut retry_receiver = self.retry_receiver.lock().await;
```
kamiyaa, line 78: not sure if this really matters but I think this way, we only lock once instead of on each iteration

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -447,4 +503,116 @@ pub mod test {
         assert_eq!(popped[3], op_ids[0]);
         assert_eq!(popped[4], op_ids[1]);
     }
+
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn test_process_retry_requests_by_id() {
```
kamiyaa, line 509: tests to make sure we aren't retrying all messages

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -105,7 +107,7 @@ impl SerialSubmitter {
     pub fn new(
         domain: HyperlaneDomain,
         rx: mpsc::UnboundedReceiver<QueueOperation>,
-        retry_op_transmitter: Sender<MatchingList>,
+        retry_op_transmitter: &Sender<Arc<MessageRetryRequest>>,
```
daniel-savu, line 110: `Sender` holds an `Arc` internally as well btw https://docs.rs/tokio/latest/src/tokio/sync/broadcast.rs.html#165

don't think a reference to an Arc would save that much memory - unless you think it's more ergonomic or clearer, would keep this arg without a ref

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -32,11 +32,13 @@ use hyperlane_core::{
 };
 
 use crate::msg::pending_message::CONFIRM_DELAY;
-use crate::settings::matching_list::MatchingList;
+use crate::server::MessageRetryRequest;
 
 use super::op_queue::OpQueue;
 use super::op_queue::OperationPriorityQueue;
 
+pub const SUBMITTER_QUEUE_COUNT: usize = 3;
```
tkporter, line 40: can we add a comment as to why this is important? is there any way we can make sure that e.g. tests fail if this ever becomes stale?

tkporter (commented):
haven't looked at any logic just a driveby

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -16,7 +16,7 @@ pub type OperationPriorityQueue = Arc<Mutex<BinaryHeap<Reverse<QueueOperation>>>
 pub struct OpQueue {
     metrics: IntGaugeVec,
     queue_metrics_label: String,
-    retry_rx: Arc<Mutex<Receiver<MatchingList>>>,
+    retry_receiver: Arc<Mutex<Receiver<Arc<MessageRetryRequest>>>>,
```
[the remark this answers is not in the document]
kamiyaa: Okay, I will change it back :+1:

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,28 +72,56 @@ impl OpQueue {
...
+                            matched = true;
+                        });
+                    if matched {
+                        op.reset_attempts();
+                    }
+                    Reverse(op)
+                })
+                .collect();
+            queue.append(&mut reprioritized_queue);
+        }
```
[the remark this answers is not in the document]
kamiyaa: good call :+1:

rust/main/agents/relayer/src/msg/op_submitter.rs
kamiyaa: Yep, can add comment. Technically there is a way to test this with capacity() and permits
https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.capacity

But I think that would require us to refactor OpSubmitter in a way that is easily testable via unittests

rust/main/utils/run-locally/src/main.rs
```diff
@@ -539,6 +540,9 @@ fn main() -> ExitCode {
 
         sleep(Duration::from_secs(5));
     }
+    // test retry request
+    let resp = server::run_retry_request().expect("Failed to process retry request");
```
tkporter, line 544: if `SUBMITTER_QUEUE_COUNT` was ever inaccurate, would we expect this to fail?

happy to ship either way

tkporter (approved):

rust/main/utils/run-locally/src/main.rs
kamiyaa: no, I think it would just be a bit slow to receive a response

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,28 +72,54 @@ impl OpQueue {
...
+                            // update retry metrics
+                            retry_response.matched += 1;
+                            matched = true;
+                        });
+                    if matched {
+                        op.reset_attempts();
+                    }
+                    Reverse(op)
+                })
+                .collect();
```
daniel-savu, line 106: given the bug where requests were matching every item in the opqueue, i'd really like to improve unit testing of this logic section. The area I highlighted can be refactored into its own function: `reprioritize_matching(queue: &mut OpQueue, retry_requests: &[MessageRetryRequest]) -> Vec<MessageRetryQueueResponse>`

with this you can then unit test all edge cases you can think of. Some examples: 
- empty queue
- same/different origin/sender/recipient queued items
- retry request with every field set to a wildcard
- retry request with multiple [ListElement](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f74aca42426b197d8a6948eff1ee92e9f250f6ea/rust/main/agents/relayer/src/settings/matching_list.rs#L227) fields specified
- retry request with an array of ListElements
- many identical retry requests (response should have the same `evaluated` and `matched`)

another benefit is that we'll no longer need to create the vec tuple with empty responses in here which makes the logic harder to understand

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -447,4 +489,114 @@ pub mod test {
...
+
+        let retry_response = receiver.recv().await.unwrap();
+
+        assert_eq!(retry_response.evaluated, 5);
+        assert_eq!(retry_response.matched, 1);
+    }
+
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn test_process_retry_requests_by_domain() {
```
daniel-savu, line 546: this test gives similar coverage to `test_destination_domain` with an additional check on the response content, so I think they can be merged

rust/main/utils/run-locally/src/main.rs
daniel-savu: > if SUBMITTER_QUEUE_COUNT was ever inaccurate, would we expect this to fail?

this check relies on there being unprocessable messages in queues (`ZERO_MERKLE_INSERTION_KATHY_MESSAGES` + `SOL_MESSAGES_WITH_NON_MATCHING_IGP`)

so the assumption is:
- we'll keep having these unprocessable messages
- they'll either be in the prepare or submit queues
----
I would however like this to fail if things get too slow (for whatever reason, including more queues), so in `run-locally/src/server.rs` it'd be great to add a `.timeout(100ms)` when building `client.post(url).json(&body).send()`

daniel-savu (changes requested):
hitting the red button to invalidate @tkporter's approval :P

rust/main/agents/relayer/src/msg/op_queue.rs
kamiyaa: Added test cases and moved logic into its own function

kamiyaa:
Addressed comments :+1:

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -72,28 +72,56 @@ impl OpQueue {
...
-            message_retry_requests.push(message_id);
+        {
+            let mut retry_receiver = self.retry_receiver.lock().await;
+            while let Ok(retry_request) = retry_receiver.try_recv() {
+                message_retry_requests.push((
+                    retry_request,
+                    MessageRetryQueueResponse {
+                        evaluated: 0,
+                        matched: 0,
+                    },
```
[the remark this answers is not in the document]
kamiyaa: Updated to use `default()` :+1:

daniel-savu (approved):

PR #5151, opened 2025-01-11
feat(relayer): add core metric for the highest seen index in tree 

aroralanuk:
### Description

- Instead of using the latest leaf inserted into the tree which is slower, we retrieve the highest index of leaf currently stored in the hyperlane_db while we're syncing.
- The syncer task is in a JoinHandle with an index_updater task, which interrupts the syncer every 5 seconds to update if the syncer saw a new higher leaf index. This is because the syncer is a continuously running process and this is the easiest way to check progress in between. 
- Renamed latest_leaf_index to latest_tree_insertion for clarity.

### Drive-by changes

None

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/1396

### Backward compatibility

Yes

### Testing

invariant in e2e and manually testing

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -2,7 +2,6 @@ use std::{
     collections::{HashMap, HashSet},
     fmt::{Debug, Formatter},
     sync::Arc,
-    time::Duration,
 };
 
 use async_trait::async_trait;
@@ -11,7 +10,7 @@ use eyre::Result;
 use futures_util::future::try_join_all;
 use hyperlane_base::{
     broadcast::BroadcastMpscSender,
-    db::{HyperlaneDb, HyperlaneRocksDB, DB},
+    db::{HyperlaneRocksDB, DB},
     metrics::{AgentMetrics, MetricsUpdater},
     settings::{ChainConf, IndexSettings},
     AgentMetadata, BaseAgent, ChainMetrics, ContractSyncMetrics, ContractSyncer, CoreMetrics,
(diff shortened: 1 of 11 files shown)


rust/main/agents/relayer/src/relayer.rs
```diff
@@ -526,16 +527,38 @@ impl Relayer {
...
                     SyncOptions::new(Some(cursor), tx_id_receiver),
                 )
-                .await
+                .await;
         }))
-        .instrument(info_span!("MerkleTreeHookSync"))
+        .instrument(info_span!("MerkleTreeHookSync"));
+
+        let index_updater_handle =
+            tokio::spawn(TaskMonitor::instrument(&task_monitor, async move {
```
tkporter, line 547: imo creating a new task just to update this metric & using the DB as an intermediary for this info is unnatural

I was thinking instead of having this be super merkle-tree specific, this would be more of a generic sequence number that's tracked when sequence-aware indexing

And in poking around where I'd imagine this would live, it turns out that Mantas actually added exactly this metric just a couple weeks ago 😵‍💫 https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs#L456-L459

https://abacusworks.grafana.net/goto/t9kdDYDNg?orgId=1

So given we already have this metric, let's just use the one that was already added `hyperlane_cursor_max_sequence` for our alert. Can you have this PR just be to rename `latest_leaf_index_gauge` -> `latest_tree_insertion_gauge`?

Sorry for sending you down this path - I wasn't aware of this change

tkporter: Separately, I think our Grafana may be maxing out on its cardinality, because `hyperlane_cursor_max_sequence` seems to only exist for some chains but not others. Will create a thread in backend eng about this

aroralanuk: for some reason, the sealevel changes are showing up here so I'm closing this issue and making the changes here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5178

PR #5167, opened 2025-01-14
feat: Add logging to agents

ameten:
### Description

Add logging to agents
* When an agent start
* When cursor indexing task is exiting
* When cursor sync method is exiting

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5098

### Backward compatibility

Yes

### Testing

E2E tests

### DIFF
--- a/rust/main/agents/relayer/src/main.rs
+++ b/rust/main/agents/relayer/src/main.rs
@@ -18,7 +18,9 @@ mod memory_profiler;
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 20)]
 async fn main() -> Result<()> {
+    // Logging is not initialised at this point, so, using `println!`
     println!("Relayer starting up...");
+
     let agent_main_fut = agent_main::<Relayer>();
 
     #[cfg(feature = "memory-profiling")]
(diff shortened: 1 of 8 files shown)


rust/main/agents/relayer/src/main.rs
```diff
@@ -18,6 +18,7 @@ mod memory_profiler;
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 20)]
 async fn main() -> Result<()> {
+    println!("Relayer starting up...");
```
tkporter, line 21: is this before tracing is available? same goes for all the println usage throughout this PR

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -108,6 +109,11 @@ where
                     .await;
             }
         }
+
+        // Although the above loop should never end (unless by panicking),
+        // we put log here to make sure that we see when this method returns normally.
```
tkporter, line 114: so this is if we ever change this function such that the loop ever exits, then we have this log?

tkporter (approved):

rust/main/agents/relayer/src/main.rs
ameten: It is correct. Tracing is not initialised at this point. So, using `println!`. I added a comment.

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: This log is added out of desperation why task stopped making any progress. Presumably, compiler should optimise it out, but just in case there is a weird exit from the loop (which is not a panic), we may see this log.

PR #5168, opened 2025-01-14
fix: allow CI to clone any registry commit

paulbalaji:
### Description

try cloning registry from commit that's already merged


.registryrc
```diff
@@ -1 +1 @@
-9d7868fa55b97d610c9c1916d22d71d3e67b5aa8
+da7e33fa395e497c154c6904fa79c91b7b963954
```
```suggestion
9d7868fa55b97d610c9c1916d22d71d3e67b5aa8
```

PR #5181, opened 2025-01-15
feat: Add contract sync liveness metrics

ameten:
### Description

Add contract sync liveness metrics

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5098

### Backward compatibility

Yes

### Testing

E2E testing in CI

### DIFF
--- a/rust/main/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/main/hyperlane-base/src/contract_sync/mod.rs
@@ -98,8 +98,13 @@ where
             .metrics
             .stored_events
             .with_label_values(&[label, chain_name]);
+        let iteration_metric = self
+            .metrics
+            .iterations
+            .with_label_values(&[label, chain_name]);
 
         loop {
+            iteration_metric.inc();
             if let Some(rx) = opts.tx_id_receiver.as_mut() {
                 self.fetch_logs_from_receiver(rx, &stored_logs_metric).await;
             }
--- a/rust/main/hyperlane-base/src/contract_sync/mod.rs
+++ b/rust/main/hyperlane-base/src/contract_sync/mod.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashSet, fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc, time::Duration,
+    time::UNIX_EPOCH,
 };
 
 use axum::async_trait;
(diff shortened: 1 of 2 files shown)


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -98,8 +98,13 @@ where
             .metrics
             .stored_events
             .with_label_values(&[label, chain_name]);
+        let iteration_metric = self
+            .metrics
+            .iterations
+            .with_label_values(&[label, chain_name]);
 
         loop {
+            iteration_metric.inc();
```
daniel-savu, line 107: over a long enough time window, this metric will overflow right? Since there are way more loops than new blocks or messages. If we just want to make sure the loop is alive we can just alternative between two values (like `true` and `false`) and liveness would be lost when no more state alternations are happening

ameten: I suppose this metric is incremented every time when there are messages in internal channel (receiver) and there are new blocks to index (cursor). Otherwise, there is a 5 seconds sleep period which should prevent this metric to grow very quickly.

If the frequency of this metric to overflow, we should be fine.

Regarding the alternating metric, presumably, we can end up in a situation when this metrics will seemly stay constant if Prometheus is not scraping metrics frequently enough. Effectively, this metric will alternate between two values, but Prometheus will scrape only rare entries of it.

daniel-savu: > I suppose this metric is incremented every time when there are messages in internal channel (receiver) and there are new blocks to index (cursor). Otherwise, there is a 5 seconds sleep period which should prevent this metric to grow very quickly.
If the frequency of this metric to overflow, we should be fine.

fair, could be useful to mention some of this in a comment

> Regarding the alternating metric, presumably, we can end up in a situation when this metrics will seemly stay constant if Prometheus is not scraping metrics frequently enough. Effectively, this metric will alternate between two values, but Prometheus will scrape only rare entries of it.

that's also true, I guess another solution would be to set the unix timestamp as the metric to follow (`last_tick_timestamp`). Happy to go with whatever you think is easier to understand

ameten: Timestamp makes sense here. Great point, Daniel!

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -120,6 +126,15 @@ where
         info!(chain = chain_name, label, "contract sync loop exit");
     }
 
+    fn update_liveness_metric(liveness_metric: &GenericGauge<AtomicI64>) {
+        liveness_metric.set(
+            UNIX_EPOCH
+                .elapsed()
+                .map(|d| d.as_millis() as i64)
```
daniel-savu, line 133: since this metric is only expected to be updated every secs, it's probably safer to use `d.as_secs()` instead (millis returns u256, whereas secs returns a [u64](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs))

ameten: Ok.

daniel-savu (approved):

Issue #5060, opened 2024-12-20
bug: Operation status labels are reset across relayer restarts


daniel-savu:
## Problem

**Is your feature request related to a problem? Please describe.**

The relayer has an enum of possible statuses a pending operation (i.e. pending message) can have [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/12b0e28b6f5528c13cd3b8d26c1f6df595fb0af3/rust/main/hyperlane-core/src/traits/pending_operation.rs#L169). New statuses are assigned based on where a message is in its submission lifecycle. For instance, all messages are created with a `FirstPrepareAttempt` status, then if metadata is fetched successfully (e.g. validator signatures for the multisig ISM) the message's status is updated to `ReadyToSubmit` [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/143b46e4705387799e791bf7849486f78cfba169/rust/main/agents/relayer/src/msg/op_submitter.rs#L282), then if the transaction on the destination chain is successfully dispatched the status is updated to `Confirm(SubmittedBySelf)` [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/143b46e4705387799e791bf7849486f78cfba169/rust/main/agents/relayer/src/msg/op_submitter.rs#L608), and after the message is finalized on the destination it is removed from in-memory queues.

This status is also persisted to the local db when it is updated so upon restarting all old messages should be loaded into the prep queue with their old statuses (such as `Retry(GasPaymentNotFound)`).

The problem is that relayer restarts causes all old messages to be loaded into the prep queue with a `FirstPrepareAttempt` status, as can be observed below for `celo`. The red colour chunk is always reset upon restart (you may even need to zoom in to see it), although all of those messages showing up as green have been attempted before.

<img width="1126" alt="Screenshot 2024-12-20 at 13 40 21" src="https://github.com/user-attachments/assets/514a3b31-dc84-40c7-9109-8914ebe88c98" />

https://abacusworks.grafana.net/goto/8cJ2XfIHg?orgId=1

## Solution

Statuses should stay consistent across restarts by storing and retrieving them from the local db. Understand and fix why this is not the case

kamiyaa:
Discussed this with @daniel-savu 

Takeaways:
 - potentially add method to `State` to return a list of Validator tasks to be dropped
 - need to find a way to drop relayer task without dropping its tmpdir
[...]

PR #5182, opened 2025-01-15
feat: attempt to fetch operation status in database upon relayer restart

kamiyaa:
### Description

 - Add logic to retrieve message status from db before defaulting to `PendingOperationStatus::FirstPrepareAttempt`.
 - Add e2e tests to restart relayer and check message statuses were successfully read from db

### Drive-by changes

 - add `impl Default` to MerkleTreeBuilder to satisfy `clippy`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5060

### Backward compatibility

Yes

### Testing

 - Add e2e tests to restart relayer and check message statuses were successfully read from db


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -510,13 +510,20 @@ impl PendingMessage {
...
-            message,
-            ctx,
-            // Since we don't persist the message status for now, assume it's the first attempt
-            PendingOperationStatus::FirstPrepareAttempt,
-            app_context,
-        );
+        // Attempt to fetch status about message from database
+        let mut pm = match ctx.origin_db.retrieve_status_by_message_id(&message.id()) {
+            Ok(Some(status)) => {
+                tracing::debug!(?status, "Message status retrieved from db");
```
ameten, line 516: Let's have a or corresponding message for the case when we could not retrieve status.

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -510,13 +510,20 @@ impl PendingMessage {
...
     ) -> Self {
-        let mut pm = Self::new(
-            message,
-            ctx,
-            // Since we don't persist the message status for now, assume it's the first attempt
-            PendingOperationStatus::FirstPrepareAttempt,
-            app_context,
-        );
+        // Attempt to fetch status about message from database
+        let mut pm = match ctx.origin_db.retrieve_status_by_message_id(&message.id()) {
```
ameten, line 514: I wonder if it worth to refactor this method: retrieve all the required bits of data from database and construct the instance of `PendingMessage` at the end?

rust/main/utils/run-locally/src/main.rs
```diff
@@ -539,10 +476,115 @@ fn main() -> ExitCode {
...
+    let relayer_logfile = File::open(log_file_path).unwrap();
+
+    const RETRIEVED_MESSGE_LOG: &str = "Message status retrieved from db";
+    let invariant_logs = &[RETRIEVED_MESSGE_LOG];
+
+    log!("Checking message statuses were retrieved from logs...");
+    let log_counts = get_matching_lines(&relayer_logfile, invariant_logs);
+    assert_eq!(
+        log_counts.get(RETRIEVED_MESSGE_LOG),
+        Some(&ZERO_MERKLE_INSERTION_KATHY_MESSAGES)
```
daniel-savu, line 582: if the message status is retrieved but it's not the expected one, this check would still pass. We should specifically makes sure that the found logs are `CouldNotFetchMetadata` ones

rust/main/utils/run-locally/src/main.rs
```diff
@@ -536,13 +474,151 @@ fn main() -> ExitCode {
...
+    state.push_agent(relayer_env.spawn("RLY", Some(&AGENT_LOGGING_DIR)));
+}
+
+fn message_retrieval_invariants_met() -> eyre::Result<bool> {
+    let log_file_path = AGENT_LOGGING_DIR.join("RLY-output.log");
+    let relayer_logfile = File::open(log_file_path).unwrap();
+
+    log!("Checking message statuses were retrieved from logs...");
+
+    const RETRIEVED_MESSGE_LOG: &str = "Message status retrieved from db";
```
tkporter, line 596: if we use this in e2e can you put this const in the relayer and we import from it? that way we're less likely to accidentally break e2e if we change a log line. iirc we can import from the relayer crate in here

tkporter: also nit typo `RETRIEVED_MESSGE_LOG` -> `RETRIEVED_MESSAGE_LOG`

rust/main/utils/run-locally/src/main.rs
```diff
@@ -536,13 +474,151 @@ fn main() -> ExitCode {
...
+            // string parse for id and remove duplicates by collecting
+            // into a hashset
+            s.rsplit_once("id\u{1b}[0m\u{1b}[2m=\u{1b}[0m\"")
+                .and_then(|(_, s2)| s2.split_once("\""))
+                .map(|(id, _)| id)
+        })
+        .collect();
+
+    let no_metadata_message_count = no_metadata_message_ids.len();
+    if no_metadata_message_count < ZERO_MERKLE_INSERTION_KATHY_MESSAGES as usize {
```
tkporter, line 612: it wasn't obvious to me why we expected to have `CouldNotFetchMetadata` messages in this test even after `termination_invariants_met` until I read the code and realized it's because of these messages that we don't insert into the merkle tree. Could you make sure we have a comment somewhere (preferably where `message_retrieval_invariants_met` is called) that describes what's going on in this test and why it works?

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -510,27 +510,39 @@ impl PendingMessage {
...
-            // Since we don't persist the message status for now, assume it's the first attempt
-            PendingOperationStatus::FirstPrepareAttempt,
-            app_context,
-        );
-        match pm
-            .ctx
+        // Attempt to fetch status about message from database
+        let message_status = match ctx.origin_db.retrieve_status_by_message_id(&message.id()) {
+            Ok(Some(status)) => {
+                tracing::debug!(
```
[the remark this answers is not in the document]
tkporter: I understand this is used by e2e though - so I guess either make e2e use trace logging instead of debug, or maybe we can have the log severity feature gated or something. But definitely would prefer to not have this logged upon startup with typical `debug` operations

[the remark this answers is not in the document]
kamiyaa: I think what we can do is use `EnvFilter`
https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

Then in production we have `RUST_LOG=debug` and then for e2e, we can have `RUST_LOG=trace`
and then change this log message to `trace!()`.

What do you think?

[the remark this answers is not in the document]
daniel-savu: I think easiest is to gate the `debug` with the `test-utils` feature, it was created for this specific reason. I'm concerned about the size of logs getting out of hand if we switch to `trace` level in e2e, given we already have >10k logs being generated each time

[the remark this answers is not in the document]
kamiyaa: If we don't want our agents to be too noisy (in e2e or prod) we can have custom filters set up for dependencies.
This way, we only show errors from dependencies, but show debug/trace for our own crates/packages.

```rust
    let trace_filter = filter::Targets::new()
        .with_target("alloy_rpc_client", Level::ERROR)
        .with_target("alloy_transport_http", Level::ERROR)
        .with_target("ethers_providers", Level::ERROR)
        .with_target("hyper", Level::ERROR)
        .with_target("reqwest", Level::ERROR)
        .with_target("rustls", Level::ERROR)
        .with_target("sqlx", Level::ERROR)
        .with_default(Level::DEBUG);

    tracing_subscriber::registry()
        .with(trace_filter);
```

rust/main/utils/run-locally/src/utils.rs
```diff
@@ -134,3 +134,15 @@ pub fn get_matching_lines(file: &File, search_strings: &[&str]) -> HashMap<Strin
...
+    let reader = io::BufReader::new(file);
+    let mut lines = reader.lines();
+    let mut matched_lines = Vec::new();
+    while let Some(Ok(line)) = lines.next() {
+        if line.contains(search_string) {
+            matched_lines.push(line);
+        }
+    }
+    matched_lines
+}
```
daniel-savu, line 148: see my other comment first, since I think we don't need this function

the opening of the file, iteration and matching of a predicate is common to both this function and `get_matching_lines`, so we should reuse it as much as possible. You can refactor such that the generic logic works like a `fold`, and the arguments it takes are the initial value and the accumulation closure (exactly like `fold` does).

rust/main/utils/run-locally/src/main.rs
```diff
@@ -536,13 +474,151 @@ fn main() -> ExitCode {
...
+        .iter()
+        .filter(|s| s.contains("CouldNotFetchMetadata"))
+        .filter_map(|s| {
+            // string parse for id and remove duplicates by collecting
+            // into a hashset
+            s.rsplit_once("id\u{1b}[0m\u{1b}[2m=\u{1b}[0m\"")
+                .and_then(|(_, s2)| s2.split_once("\""))
+                .map(|(id, _)| id)
+        })
+        .collect();
```
daniel-savu, line 609: I think the only thing needed from here is the log count, and for that we can use a refactored version of `get_matching_lines` that takes a Vec<Vec<String>> which requires lines to match 1 or more strings at the same time. Given that, we can likely remove `get_matched_lines`. That would make the utility reusable for many other future checks as well

rust/main/utils/run-locally/src/utils.rs
kamiyaa: yeah, that makes sense :+1:

rust/main/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
daniel-savu: tracing already doesn't include dependencies - we have a special verbosity filter called `dependecyTrace` for that. Let's go with the feature gating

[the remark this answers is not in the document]
kamiyaa: okay :+1:

rust/main/utils/run-locally/src/utils.rs
```diff
@@ -119,30 +118,24 @@ pub fn stop_child(child: &mut Child) {
     };
 }
 
-pub fn get_matching_lines(file: &File, search_strings: &[&str]) -> HashMap<String, u32> {
+pub fn get_matching_lines(file: &File, search_strings: &[Vec<String>]) -> Vec<u32> {
```
daniel-savu, line 121: hmm I think if you use `Vec<&str>` you can actually use it as a hashmap key, which would avoid the hardcoding array indices. The indices approach can lead to hard to debug errors later and isn't very easy to reason about at first glance
```suggestion
pub fn get_matching_lines<'a>(
    file: &File,
    search_strings: Vec<Vec<&'a str>>,
) -> HashMap<Vec<&'a str>, u32> {
```

and then in other places like `relayer_restart_invariants_met` you can call as e.g.
```rust
let line_filters = vec![RETRIEVED_MESSAGE_LOG, "CouldNotFetchMetadata"];
...
let matched_logs = get_matching_lines(&relayer_logfile, vec![line_filters.clone()]);

let no_metadata_message_count = matched_logs.get(&line_filters).unwrap().clone();
```

rust/main/utils/run-locally/src/main.rs
```diff
@@ -435,7 +436,7 @@ fn main() -> ExitCode {
     let mut failure_occurred = false;
     let starting_relayer_balance: f64 = agent_balance_sum(9092).unwrap();
 
-    while !SHUTDOWN.load(Ordering::Relaxed) {
+    'main_loop: while !SHUTDOWN.load(Ordering::Relaxed) {
```
daniel-savu, line 439: if you need to use labels, it's a strong indication the code reached a complexity where it needs to be refactored quite a bit

you can use that fact that once `termination_invariants_met` is true, the relayer can be restarted before killing the other processes

hoping lines 439-505 can be simplified quite a lot

rust/main/utils/run-locally/src/utils.rs
kamiyaa: ah good idea!

rust/main/utils/run-locally/src/main.rs
kamiyaa: Simplified lines 439-505 a bit :+1:

rust/main/utils/run-locally/src/main.rs
```diff
@@ -520,29 +459,178 @@ fn main() -> ExitCode {
...
+            if relayer_restart_invariants_met().unwrap_or(false) {
+                // end condition reached successfully
+                break;
+            } else if (Instant::now() - loop_start).as_secs() > config.ci_mode_timeout {
+                // we ran out of time
+                log!("CI timeout reached before relayer restart invariants were met");
+                failure_occurred = true;
+                break;
             }
         }
```
daniel-savu, line 489: this while loop follows a very similar pattern to the one above and could be refactored

kamiyaa: Refactored into function `wait_for_condition`

tkporter (commented):
enjoying the cleanup around run-locally :)

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -510,27 +512,42 @@ impl PendingMessage {
...
-            PendingOperationStatus::FirstPrepareAttempt,
-            app_context,
-        );
-        match pm
-            .ctx
+        // Attempt to fetch status about message from database
+        let message_status = match ctx.origin_db.retrieve_status_by_message_id(&message.id()) {
+            Ok(Some(status)) => {
+                #[cfg(feature = "test-utils")]
+                tracing::debug!(
```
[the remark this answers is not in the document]
kamiyaa: yeah sure, we can add trace.

rust/main/utils/run-locally/src/main.rs
```diff
@@ -491,68 +429,220 @@ fn main() -> ExitCode {
...
-                break;
-            } else if (Instant::now() - loop_start).as_secs() > config.ci_mode_timeout {
-                // we ran out of time
-                log!("CI timeout reached before queues emptied");
-                failure_occurred = true;
-                break;
-            }
-        }
 
-        // verify long-running tasks are still running
```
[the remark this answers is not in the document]
kamiyaa: Added as a param

rust/main/utils/run-locally/src/main.rs
```diff
@@ -510,49 +450,217 @@ fn main() -> ExitCode {
...
-                    log!(
-                        "Child process {} exited unexpectedly, with code {}. Shutting down",
-                        name,
-                        status.code().unwrap()
-                    );
-                    failure_occurred = true;
-                    SHUTDOWN.store(true, Ordering::Relaxed);
-                    break;
+        },
+        || {
```
tkporter, line 454: should we also have this for the restart invariant condition too?

rust/main/utils/run-locally/src/main.rs
```diff
@@ -510,49 +450,224 @@ fn main() -> ExitCode {
...
+            // end condition reached successfully
+            break;
+        }
+        if check_ci_timed_out(config.ci_mode_timeout, start_time) {
+            // we ran out of time
+            log!("CI timeout reached before invariants were met");
+            return false;
+        }
+        if shutdown_criteria_fn() {
+            return false;
```
tkporter, line 637: can you still do the `SHUTDOWN.store(true, Ordering::relaxed)` here? there may be some things running concurrently that would need to be indicated to shut down as well

rust/main/utils/run-locally/src/main.rs
```diff
@@ -510,49 +450,224 @@ fn main() -> ExitCode {
...
+    true
+}
+
+/// check if CI has timed out based on config
+fn check_ci_timed_out(timeout_secs: u64, start_time: Instant) -> bool {
+    (Instant::now() - start_time).as_secs() > timeout_secs
+}
+
+/// verify long-running tasks are still running
+fn long_running_processes_live_check(state: &mut State) -> bool {
```
tkporter, line 649: nit: imo just reading this signature I'd think that returning true means that things are live and are ok, but it's the opposite, can you rename accordingly?

tkporter (approved):

rust/main/utils/run-locally/src/main.rs
kamiyaa: Would be kind of hard because both `wait_for_condition` depends on this `SHUTDOWN`.
If we did the SHUTDOWN here, then we need to reset it for the second `wait_for_condition`.
So you would need to do `Ctrl+C` twice to cancel the e2e locally.

rust/main/utils/run-locally/src/main.rs
kamiyaa: Yep :+1:

rust/main/utils/run-locally/src/main.rs
tkporter: Imo a `shutdown_criteria_fn` being true should mean that the test essentially skips everything else, but does what it needs to gracefully shut down things. I think just adding a `SHUTDOWN.store(true, Ordering::relaxed)` would do this-- if we hit this for the first time we call `wait_for_condition`, we get back `test_passed` as false, and we end the test early with this
```
if !test_passed {
  log!("Failure occurred during E2E");
  return report_test_result(test_passed);
}
```

Which is the behavior we want imo

kamiyaa: ah I see

tkporter (approved):

PR #5183, opened 2025-01-15
feat: infra safe tx parser auto-infers txs to parse

paulbalaji:
### Description

- feat: infra safe tx parser infers txs to parse
- users are no longer required to provide the exact txs to parse
- now supports multiple txs per chain
- still allows someone to provide `--chains`/`-c` arg if they wish to specify exactly which chains to check
- support timelock txs
- write the yaml output to file instead of requiring users to `| tee file.txt`
- add catch-all case for ownership transfers on ownables

### Drive-by changes

- rootLogger instead of console logs (make eslint happy in editor)
- remove `withTxHashes` helper as it's no longer used anywhere
- fix dependency issue by moving getPendingTxs helper to utils
- fixes

### Related issues

no longer need https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5160

### DIFF
--- a/typescript/infra/src/tx/govern-transaction-reader.ts
+++ b/typescript/infra/src/tx/govern-transaction-reader.ts
@@ -195,7 +195,7 @@ export class GovernTransactionReader {
       value: tx.value,
     });
 
-    let insight = '';
+    let insight;
     if (
       decoded.functionFragment.name ===
       timelockControllerInterface.functions[
@@ -224,10 +224,15 @@ export class GovernTransactionReader {
       insight = `Cancel scheduled transaction ${id}`;
     }
 
+    const args = formatFunctionFragmentArgs(
+      decoded.args,
+      decoded.functionFragment,
+    );
+
     return {
       chain,
-      to: `Timelock Controller (${chain} ${timelocks[chain]})`,
-      insight,
+      to: `Timelock Controller (${chain} ${tx.to})`,
+      ...(insight ? { insight } : { args }),
     };
   }
 
@@ -258,7 +263,7 @@ export class GovernTransactionReader {
       value: tx.value,
     });
 
-    let insight = '';
+    let insight;
     if (
       decoded.functionFragment.name ===
       tokenRouterInterface.functions['setHook(address)'].name
@@ -327,12 +332,9 @@ export class GovernTransactionReader {
       insight = `Unenroll remote routers for ${insights.join(', ')}`;
     }
 
-    if (
-      decoded.functionFragment.name ===
-      tokenRouterInterface.functions['transferOwnership(address)'].name
-    ) {
-      const [newOwner] = decoded.args;
-      insight = `Transfer ownership to ${newOwner}`;
+    let ownableTx = {};
+    if (!insight) {
+      ownableTx = await this.readOwnableTransaction(chain, tx);
     }
 
     assert(tx.to, 'Warp Module transaction must have a to address');
@@ -340,6 +342,7 @@ export class GovernTransactionReader {
     const token = this.warpRouteIndex[chain][tokenAddress];
 
     return {
+      ...ownableTx,
       chain,
       to: `${token.symbol} (${token.name}, ${token.standard}, ${tokenAddress})`,
       insight,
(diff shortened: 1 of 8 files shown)


typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -153,6 +169,68 @@ export class GovernTransactionReader {
...
+      decoded.functionFragment.name ===
+      timelockControllerInterface.functions['cancel(bytes32)'].name
+    ) {
+      const [id] = decoded.args;
+      insight = `Cancel scheduled transaction ${id}`;
+    }
+
+    return {
+      chain,
+      to: `Timelock Controller (${chain} ${timelocks[chain]})`,
```
tkporter, line 229: nit: should we maybe just use `tx.to` instead of `timelocks[chain]`? to make this function potentially usable for future situations where we change the criteria for `isTimelockControllerTransaction` -- e.g. if we ever want it to work for a timelock that's different from the ones we have canonically defined

typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -629,6 +709,40 @@ export class GovernTransactionReader {
     };
   }
 
+  private async readOwnableTransaction(
+    chain: ChainName,
+    tx: AnnotatedEV5Transaction,
+  ): Promise<GovernTransaction> {
+    if (!tx.data) {
+      throw new Error('⚠️ No data in proxyAdmin transaction');
```
tkporter, line 717: nit: proxyAdmin not accurate

typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -255,16 +327,24 @@ export class GovernTransactionReader {
       insight = `Unenroll remote routers for ${insights.join(', ')}`;
     }
 
+    if (
+      decoded.functionFragment.name ===
```
tkporter, line 331: feels a little weird to have this explicitly handled when we also have the general purpose ownable logic

could this just call into the general purpose logic? as we always know that a warp route is ownable

typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -629,6 +709,40 @@ export class GovernTransactionReader {
...
+      ownableInterface.functions['transferOwnership(address)'].name
+    ) {
+      const [newOwner] = decoded.args;
+      insight = `Transfer ownership to ${newOwner}`;
+    }
+
+    return {
+      chain,
+      insight,
```
tkporter, line 742: would this be weird if it's `''`?

typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -138,6 +149,11 @@ export class GovernTransactionReader {
       return this.readWarpModuleTransaction(chain, tx);
     }
 
+    // If it's an Ownable transaction
+    if (await this.isOwnableTransaction(chain, tx)) {
```
tkporter, line 153: this one's interesting because it's not quite mutually exclusive with the others, but we return early if we get a match for the others. Maybe just a sign that we should rethink the setup of this class at some point

typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -43,6 +43,15 @@ const safeDeploymentsVersions = {
   },
 };
 
+// Override for chains that haven't yet been published in the safe-deployments package.
+// Temporary until PR to safe-deployments package is merged and SDK dependency is updated.
+const chainOverrides = {
```
tkporter, line 48: surprised we only hit this just now

typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -43,6 +43,15 @@ const safeDeploymentsVersions = {
...
+  543210: {
```
tkporter, line 49: can you add a comment for what chain this is?

tkporter (approved):
noting that the govern tx reader is getting a bit unweildy but it was never meant to be super elegant so happy to ship

typescript/infra/src/tx/govern-transaction-reader.ts
paulbalaji: haha yeah. i actually added this catch-all for Ownables because we don't keep track of every ProxyAdmin in an artifact somewhere, so i noticed some warp tx parsing fail because it would not be able to determine what type of contract the ProxyAdmin was

maybe a sign to have better ProxyAdmin detection too

yorhodes:
> noting that the govern tx reader is getting a bit unweildy but it was never meant to be super elegant so happy to ship

very much agree with this

typescript/infra/src/tx/govern-transaction-reader.ts
yorhodes: just ran into this

PR #5191, opened 2025-01-16
feat: improve scraper startup error handling

kamiyaa:
### Description

- handle errors and sets set chain metrics if scraper fails for chain during startup

### Drive-by changes
 - implement Clone for `ScraperDb`
   - this is because DbConn doesn't implement Clone when `mock` feature is enabled

### Related issues
Fixes https://github.com/hyperlane-xyz/issues/issues/1408

### Backward compatibility

Yes

### Testing

Added unittest to test function. But don't have tests in place for testing the logic in `Scraper::run()`

### DIFF
--- a/rust/main/agents/scraper/src/agent.rs
+++ b/rust/main/agents/scraper/src/agent.rs
@@ -229,7 +229,7 @@ impl Scraper {
         let mut scrapers: HashMap<u32, ChainScraper> = HashMap::new();
 
         for domain in settings.chains_to_scrape.iter() {
-            match Self::build_chain_scraper(domain, &settings, metrics.clone(), scraper_db.clone())
+            match Self::build_chain_scraper(domain, settings, metrics.clone(), scraper_db.clone())
                 .await
             {
                 Ok(scraper) => {
--- a/rust/main/agents/scraper/src/agent.rs
+++ b/rust/main/agents/scraper/src/agent.rs
@@ -98,7 +98,7 @@ impl BaseAgent for Scraper {
                 }
             };
 
-            match MetricsUpdater::new(
+            let metrics_updater = match MetricsUpdater::new(
                 chain_conf,
                 self.core_metrics.clone(),
                 self.agent_metrics.clone(),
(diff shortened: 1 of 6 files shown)


rust/main/agents/scraper/src/agent.rs
```diff
@@ -137,8 +139,11 @@ impl BaseAgent for Scraper {
 impl Scraper {
     /// Sync contract data and other blockchain with the current chain state.
     /// This will spawn long-running contract sync tasks
-    async fn scrape(&self, domain_id: u32) -> Instrumented<JoinHandle<()>> {
-        let scraper = self.scrapers.get(&domain_id).unwrap();
+    async fn scrape(&self, domain_id: u32) -> eyre::Result<Instrumented<JoinHandle<()>>> {
+        let scraper = self.scrapers.get(&domain_id).ok_or_else(|| {
+            tracing::error!(domain_id, "Failed to get scraper for domain id");
+            eyre::eyre!("Failed to get scraper for domain id")
+        })?;
```
ameten, line 146: This method is invoked from iteration over `scrapers` hashmap. I wonder if it makes sense to pass `ChainScraper` as an argument for this method instead of re-retrieving it from the map and handling the error?

rust/main/agents/scraper/src/agent.rs
```diff
@@ -114,19 +88,47 @@ impl BaseAgent for Scraper {
...
+                }
+                Err(err) => {
+                    tracing::error!(?err, ?scraper.domain, "Failed to build metrics updater");
+                    self.chain_metrics
+                        .set_critical_error(scraper.domain.name(), true);
+                    continue;
+                }
+            }
         }
         if let Err(err) = try_join_all(tasks).await {
```
ameten, line 133: I have a concern regarding possible situation when we have scraper tasks running for a domain while getting `ChainConf` and `MetricsUpdater` fails. We shall have a scraper running without proper observability.

rust/main/agents/scraper/src/agent.rs
kamiyaa: Yeah good idea :+1:

rust/main/agents/scraper/src/agent.rs
kamiyaa: So would you recommend rearranging the order like so?
```
build chain_conf
build metrics_updater
build scraper task
```

ameten: Yes, it can be a solution. I would also build all tasks first and spawn them only if we can build all of them. To be specific, I propose the following:
1. build `chain_conf`
2. build `metric_updater` but don't spawn it
3. build scraper task (and all internal tasks, effectively it is just invoke `scrape` method)
4. spawn `metric_updater` task

ameten (approved):

PR #5200, opened 2025-01-17
Implement network deploy script with fork test

yorhodes:


solidity/script/avs/DeployNetwork.s.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity >=0.8.0;
+
+import "forge-std/Script.sol";
+
+import {ProxyAdmin} from "../../contracts/upgrade/ProxyAdmin.sol";
+import {TransparentUpgradeableProxy} from "../../contracts/upgrade/TransparentUpgradeableProxy.sol";
+import {TimelockControllerUpgradeable} from "@openzeppelin/contracts-upgradeable/governance/TimelockControllerUpgradeable.sol";
+
+contract TimelockControllerInitializer is TimelockControllerUpgradeable {
```
nambrot, line 10: Imo an odd name for a contract, maybe just call it Network or HyperlaneOmnibusNetwork or something like that??

solidity/script/avs/DeployNetwork.s.sol
```diff
@@ -0,0 +1,116 @@
...
+        __TimelockController_init(minDelay, proposers, executors, admin);
+    }
+}
+
+contract DeployNetwork is Script {
+    uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+
+    address networkRegistry = 0xC773b1011461e7314CF05f97d95aa8e92C1Fd8aA;
+
+    address proxyAdmin = 0x75EE15Ee1B4A75Fa3e2fDF5DF3253c25599cc659;
```
nambrot, line 26: Very nitty optional: Create a script that pulls these from config and add as ENV var here

solidity/foundry.toml
```diff
@@ -7,7 +7,7 @@ test = 'test'
 cache_path = 'forge-cache'
 allow_paths = ["../node_modules"]
 solc_version = '0.8.22'
-evm_version= 'paris'
+evm_version= 'shanghai'
```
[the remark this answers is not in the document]
nambrot: Does this imply a bytecode change for our contracts? If so, I would oppose this

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -8,6 +8,7 @@ import { supportedChainNames } from './supportedChainNames.js';
 
 export const upgradeTimelocks: ChainMap<Address | undefined> = {
   arbitrum: '0xAC98b0cD1B64EA4fe133C6D2EDaf842cE5cF4b01',
+  ethereum: '0x59cf937Ea9FA9D7398223E3aA33d92F7f5f986A2', // symbiotic network timelock
```
paulbalaji, line 11: this also moves mailbox/core proxyadmin to this, is that intentional?

typescript/infra/scripts/safes/propose.ts
```diff
@@ -0,0 +1,198 @@
...
+      args: [targets, values, payloads, ZERO_BYTES32, ZERO_BYTES32],
+    }),
+    // description: `Execute batch:\n ${description}`,
+  };
+
+  console.log(executeTx);
+
+  await multiProvider.sendTransaction(chain, executeTx);
+  return;
+  await multisend.sendTransactions([scheduleTx]);
```
paulbalaji, line 192: we're never sending the scheduleTx?

PR #5205, opened 2025-01-20
feat: zksync support for core solidity package

ljankovic-txfusion:
### Description

ZKSync PR 1

This PR introduces ZKSync support for Solidity contracts and restructures build files directory structure.

### Drive-by changes

- Added ZKSync compilation support using zksolc via `@matterlabs/hardhat-zksync-solc`
- Restructured typechain directory location to `core-utils/typechain`
- Decoupled Hardhat configuration for ZKSync from EVM.
- Updated build process to handle both standard and ZKSync contract artifacts

### Related issues

None

### Backward compatibility

Yes

### Testing

Testing was previously in [feat: zksync support](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4725) PR


solidity/exportBuildArtifact.sh
ljankovic-txfusion, line 1: I understand there is a code duplication for zksync artifacts, but it makes it easier to debug in the future

solidity/generate-artifact-exports.mjs
```diff
@@ -0,0 +1,35 @@
...
+ * @dev Reads each artifact file and writes it to srcOutputDir
+ */
+zksyncArtifacts.forEach((file) => {
+  const fileContent = readFileSync(file, 'utf-8');
+  let fileName = `${basename(file, '.json')}`;
+
+  const outputFile = join(srcOutputDir, `${fileName}.json`);
+
+  writeFileSync(outputFile, fileContent);
+});
```
paulbalaji, line 35: can we promise.all this?

ljankovic-txfusion: Refactored the whole file to be non-blocking

.changeset/chatty-starfishes-double.md
```diff
@@ -0,0 +1,11 @@
+---
+'@hyperlane-xyz/core': major
```
yorhodes, line 2: your PR description says backwards compatible
whats breaking here?

solidity/package.json
```diff
@@ -64,13 +67,14 @@
   ],
   "license": "Apache-2.0",
   "scripts": {
-    "build": "yarn version:update && yarn hardhat-esm compile && tsc && ./exportBuildArtifact.sh",
+    "build": "yarn version:update && yarn hardhat-esm compile && yarn hardhat-zk compile && ts-node generate-artifact-exports.mjs && tsc && ./exportBuildArtifact.sh && yarn copy-artifacts",
     "lint": "solhint contracts/**/*.sol",
-    "clean": "yarn hardhat-esm clean && rm -rf ./dist ./cache ./types ./coverage ./out ./forge-cache ./fixtures",
+    "clean": "yarn hardhat-esm clean && yarn hardhat-zk clean && rm -rf ./dist ./cache ./cache-zk ./types ./coverage ./out ./forge-cache ./fixtures",
```
yorhodes, line 72: it would be nice if there is a way to build without zksync as compilation is quite slow

yorhodes (approved):
:shipit:

solidity/package.json
ljankovic-txfusion: Hmmm there could be a flag to include zksync in build, we just need to make sure its used when publishing packages

solidity/generate-artifact-exports.mjs
```diff
@@ -0,0 +1,41 @@
+import { promises as fsPromises } from 'fs';
```
nambrot, line 1: I'm not familiar with the .mjs extension, what is it for?

solidity/package.json
nambrot: Agree, by default to not build zksync, but in the publishing script use the full build

nambrot (approved):

.changeset/chatty-starfishes-double.md
ljankovic-txfusion: updated

solidity/generate-artifact-exports.mjs
ljankovic-txfusion: its just to make sure we use ES6 file since we are using imports as opposed to cjs that "requires"

solidity/generate-artifact-exports.mjs
```diff
@@ -0,0 +1,60 @@
...
+  const fileExists = await fsPromises
+    .access(outputFile)
+    .then(() => true)
+    .catch(() => false);
+
+  if (fileExists) {
+    // File already exists, skipping...
+    // NOTE: Hardhat compiler produces duplicate artifacts when
+    // shared interfaces/libraries are used across different contracts
+    // This is expected behavior and we only need one copy of each artifact
```
ljankovic-txfusion, line 41: This is a fix to race condition caused by using async when processing files with the same name (deduplicates Interface contracts)

solidity/package.json
```diff
@@ -65,12 +68,14 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "yarn version:update && yarn hardhat-esm compile && tsc && ./exportBuildArtifact.sh",
+    "build:zk": "yarn hardhat-zk compile && ts-node generate-artifact-exports.mjs && ZKSYNC=true ./exportBuildArtifact.sh && yarn copy-artifacts",
```
ljankovic-txfusion, line 71: The least amount of changes before creating custom child process `build.js` script that checks zksync flag passed in command. 

The full build command is `yarn build && yarn build:zk`. 

Now `exportBuildArtifact.sh `exports zksync build artifacts only when ZKSYNC param is passed.

nambrot: Don't we need some changes to call the full build command then? so that our package publishing pipeline actually calls the full thing? IIRC, npm has a pre publish hook or something?

ljankovic-txfusion: I think that should be the "prepublishOnly" hook, if we dont want to change anything on the ci/ci pipeline, this  hook makes the most sense

ljankovic-txfusion: Updated package.json with "prepublishOnly" hook. Just tested it with `prepack` hook to create a local tar file using `yarn pack --dry-run` command and it builds and packs zk artifacts. Now when `publish` is called it should build zk stuff

PR #5207, opened 2025-01-20
feat: ZKSync provider types

ljankovic-txfusion:
### Description

This PR introduces specific ZKSync Provider types for SDK.

[PR to feat/core-zksync](https://github.com/txfusion/hyperlane-monorepo/pull/20)  

### Drive-by changes

- Added ZKSync provider types and interfaces
- Created builder functions for ZKSync provider configuration

### Related issues

https://linear.app/hyperlane-xyz/issue/ENG-1149

### Backward compatibility

Yes

### Testing

Build passing


.changeset/mighty-terms-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': patch
```
paulbalaji, line 2:
```suggestion
'@hyperlane-xyz/sdk': minor
```

paulbalaji (commented):
LGTM once 1 is merged

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -216,6 +234,12 @@ export interface CosmJsWasmTransaction
   transaction: ExecuteInstruction;
 }
 
+export interface ZKSyncTransaction
+  extends TypedTransactionBase<zkSyncTypes.TransactionRequest> {
+  type: ProviderType.ZkSync;
+  transaction: zkSyncTypes.TransactionRequest;
+}
+
```
nambrot, line 242: Doesn't it have to ba added to the enum below?

paulbalaji: i think it should?

typescript/sdk/package.json
paulbalaji, line 1: will need to check in the updated lockfile too

paulbalaji (approved):

typescript/sdk/src/providers/ProviderType.ts
nambrot: How did this not cause an issue in a later PR?

paulbalaji: ethers/zksync-ethers provider are similar enough in where we use them that it would still be fine with the original ethers provider type

would open up something breaking in the future should we do something that is in the zk-ethers but not in regular ethers

PR #5219, opened 2025-01-20
Implement distinct fn for new domain on RoutingISM

yorhodes:

### DIFF
--- a/solidity/contracts/isms/routing/DomainRoutingIsm.sol
+++ b/solidity/contracts/isms/routing/DomainRoutingIsm.sol
@@ -75,6 +75,21 @@ contract DomainRoutingIsm is
     }
 
     /**
+     * @notice Adds the specified origin domain
+     * @dev Reverts if the domain already exists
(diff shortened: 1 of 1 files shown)


paulbalaji (approved):

solidity/contracts/isms/routing/DomainRoutingIsm.sol
```diff
@@ -74,6 +70,37 @@ contract DomainRoutingIsm is
...
+     * @param _module The ISM to use to verify messages
+     */
+    function add(
+        uint32 _domain,
+        IInterchainSecurityModule _module
+    ) external onlyOwner {
+        _add(_domain, address(_module));
+    }
+
+    function add(
```
larryob, line 94: Do we want to use structs or arrays in configuration functions like this? You recommended structs elsewhere: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6625#discussion_r2180819285

PR #5228, opened 2025-01-21
feat: config starknet chain

mshojaei-txfusion:
### Description

This PR adds Starknet protocol support to the chain configuration CLI, including:
- New Starknet protocol type in chain metadata
- Domain Id convert for Starknet networks
- CLI prompts for Starknet chain configuration

[Difference with last PR](https://github.com/txfusion/hyperlane-monorepo/compare/feat/address-checker-starknet...txfusion:hyperlane-monorepo:feat/config-starknet-chain)

### Related issues

N/A

### Backward compatibility

Yes - Only adds new functionality without modifying existing behavior

### Testing

N/A


typescript/widgets/src/logos/protocols.ts
```diff
@@ -7,7 +7,7 @@ import { EthereumLogo } from './Ethereum.js';
 import { SolanaLogo } from './Solana.js';
 
 export const PROTOCOL_TO_LOGO: Record<
-  ProtocolType,
+  Exclude<ProtocolType, 'starknet'>,
```
paulbalaji, line 10: reuse the `NonStarknetProtocol` type defined elsewhere?

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
```diff
@@ -40,11 +40,14 @@ const logger = widgetLogger.child({
   module: 'walletIntegrations/multiProtocol',
 });
 
+type NonStarknetProtocol = Exclude<ProtocolType, 'starknet'>;
+type ProtocolRecord<T> = Record<NonStarknetProtocol, T>;
```
paulbalaji, line 44: `ProtocolMap` might be a better word given the existing `ChainMap` nomenclature we use

typescript/utils/src/types.ts
```diff
@@ -5,6 +5,7 @@ export enum ProtocolType {
   Ethereum = 'ethereum',
   Sealevel = 'sealevel',
   Cosmos = 'cosmos',
+  Starknet = 'starknet',
```
paulbalaji, line 8: since we're adding a new protocol type - automatically i would class this PR as a major change and not minor or patch

typescript/cli/src/config/chain.ts
```diff
@@ -284,11 +314,68 @@ async function addNativeTokenConfig(metadata: ChainMetadata): Promise<void> {
...
+        ).getNetwork();
+        return network.chainId.toString();
+      }
+      case ProtocolType.Starknet:
+        return new StarknetProvider({ nodeUrl: rpcUrl }).getChainId();
+    }
+  };
+}
+
+function formatChainIdBasedOnProtocol(chainId: string, protocol: ProtocolType) {
```
paulbalaji, line 367: nit: formatChainIdForProtocol

paulbalaji: and should that be `SupportedProtocol` as well?

typescript/cli/src/config/chain.ts
```diff
@@ -284,11 +314,68 @@ async function addNativeTokenConfig(metadata: ChainMetadata): Promise<void> {
...
       message: "Enter the native token's decimals:",
     });
+
+    if (isStarknet) {
+      denom = await input({
+        message: "Enter the native token's address:",
+        validate: (value) => isAddressStarknet(value),
+      });
+    }
   }
```
paulbalaji, line 324: is there any way to fetch this for the user without requiring input?

ljankovic-txfusion: On starknet mainnet there are two native tokens for fees - STRK and ETH, and AFAIK Paradex chain has an native token ETH address completely different from starknet mainnet  @mshojaei-txfusion  correct me if I am wrong

mshojaei-txfusion: We did not find any address on `starknet.js` to import, but we may be able to add a default native token address for starknet, and ask user to change it in case it was different.

typescript/cli/src/config/chain.ts
mshojaei-txfusion: You are right better to be `SupportedProtocol `

typescript/widgets/src/logos/protocols.ts
mshojaei-txfusion: wanted to do this but as this was temporary, what should I call it, where should I put this? on sdk?

PR #5232, opened 2025-01-21
feat: support for intents

fernandomg:
### Description

A first iteration towards the integration of Intents in the SDK.

This is still a WIP; CLI build is failing; and some types still need fixing as well as standardization of the `IntentCore` and `WarpCore`.

### Backward compatibility

Yes

### Testing

Manual

/cc @nambrot

### DIFF
--- a/typescript/infra/config/environments/testnet4/funding.ts
+++ b/typescript/infra/config/environments/testnet4/funding.ts
@@ -92,6 +92,8 @@ export const keyFunderConfig: KeyFunderConfig<
   igpClaimThresholdPerChain: {
     alfajores: '1',
     arbitrumsepolia: '0.05',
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/Token.ts
```diff
@@ -161,6 +166,28 @@ export class Token implements IToken {
         sourceChannel: 'channel-0',
         type: TokenConnectionType.Ibc,
       });
+    } else if (standard === TokenStandard.EvmIntent) {
+      assert(
+        this.intentRouterAddressOrDenom,
```
nambrot, line 171: Is this still required? I thought we said we would overload addressOrDenom for the router and collateralAddressorDenom for the token?

PR #5233, opened 2025-01-21
feat: scaled warp routes

yorhodes:
### Description

- Adopt scale across warp route contracts
- Remove native scaled and support scaled everywhere
- Update SDK deployer code to use new constructors

### Related issues

- opens https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5296

### Backward compatibility

SDK/CLI yes, core no

### Testing

Unit Tests


solidity/contracts/token/HypERC721.sol
```diff
@@ -7,12 +7,16 @@ import {IERC721Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC7
...
 
+uint256 constant ERC721_SCALING_FACTOR = 1;
+
 /**
  * @title Hyperlane ERC721 Token Router that extends ERC721 with remote transfer functionality.
  * @author Abacus Works
  */
 contract HypERC721 is ERC721EnumerableUpgradeable, TokenRouter {
-    constructor(address _mailbox) TokenRouter(_mailbox) {}
+    constructor(
```
nambrot, line 17: Just wondering if you thought about having a FungibleTokenRouter that has SCALE in the constructor to avoid this. I know the inheritance tree is already super deep, but figure we talk about it

typescript/sdk/src/token/deploy.ts
```diff
@@ -58,18 +58,21 @@ abstract class TokenDeployer<
   }
 
   async constructorArgs(_: ChainName, config: TokenRouterConfig): Promise<any> {
+    // TODO: derive from decimal ratios?
```
nambrot, line 61: Can you turn this into a Github issue and link it? I would also agree that we should do this based on decimal ratios, and not add another config option #configandartifacts

nambrot: But it sounds like we already had this?

nambrot (approved):

solidity/contracts/token/HypERC721.sol
yorhodes: to avoid the diff? then I would need to add more subclasses no?

typescript/sdk/src/token/config.ts
```diff
@@ -12,6 +12,7 @@ export enum TokenType {
   fastCollateral = 'fastCollateral',
   collateralUri = 'collateralUri',
   native = 'native',
+  // backwards compatible alias to native
```
jmrossy, line 15: Should we up the sdk major version and just drop this instead?

yorhodes: 🤷 I figured it was nice that this didnt need to break SDK and CLI users with existing config

solidity/contracts/token/HypERC721.sol
yorhodes: addressed

typescript/sdk/src/token/config.ts
paulbalaji: agree it's nice not to break existing config users, but we'll end up with a major SDK release for this anyway right because of the major changeset in its `core` dependency?

jmrossy: I'm fine either way, up to you guys

yorhodes: https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md we decoupled the core package from the rest of the typescript packages as an action item of https://www.notion.so/hyperlanexyz/Contracts-Release-Process-6b516e90b8db4e519a9f4881d53f7822?pvs=4

PR #5251, opened 2025-01-22
feat(widgets): add refresh icon and className to Tooltip component

Xaroz:
### Description

- Export new Refresh Icon
- Add tooltipClassName as a prop to Tooltip and added the default value to be the one used in this [PR](https://github.com/hyperlane-xyz/hyperlane-explorer/pull/147)

### Drive-by changes

No
### Related issues

Refresh button will be used to add refresh functionality to explorer

### Backward compatibility

Yes
### Testing

Storybook

### DIFF
--- a/typescript/widgets/src/components/Tooltip.tsx
+++ b/typescript/widgets/src/components/Tooltip.tsx
@@ -9,6 +9,7 @@ type Props = AnchorHTMLAttributes<HTMLAnchorElement> & {
   content: string;
   size?: number;
   placement?: PlacesType;
+  tooltipClassName?: string;
 };
 
 export function Tooltip({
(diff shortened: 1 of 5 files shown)


typescript/widgets/src/components/Tooltip.tsx
```diff
@@ -17,6 +18,7 @@ export function Tooltip({
   size = 16,
   className,
   placement = 'top-start',
+  tooltipClassName = 'max-w-[calc(100%-10px)] sm:max-w-[526px]',
```
jmrossy, line 21: This is probably okay but generally it's better to have prop classNames append into the default set instead replace it when provided. See clsx

jmrossy (approved):

PR #5256, opened 2025-01-22
chore(agents): fix high-severity dependabot alerts

daniel-savu:
### Description

Fixes high-severity dependabot which apply to agents - sealevel workspace errors are not very relevant since we only use the CLI from there.

https://github.com/hyperlane-xyz/hyperlane-monorepo/security/dependabot?q=is%3Aopen+ecosystem%3ARust

### Backward compatibility

Yes

### Testing

compilation and e2e must pass, which they did


aroralanuk (approved):
Nice!

tkporter:
Does this end up taking effect in any way? I would've thought there'd be some Cargo.lock change

I assume because we're explicitly adding these deps for the first time that these are dependencies of some of our dependencies?

PR #5270, opened 2025-01-23
feat(balances): Support automatic desired balances and alert thresholds based on daily burn approximations

Mo-Hussain:
### Description

- calculates daily burn based on recent relayer usage, min dollar value targets and a minimum number of tx, check this into monorepo as `dailyBurn.json`
```
yarn tsx ./scripts/funding/calculate-relayer-daily-burn.ts
```

- Support reading alert thresholds for `lowUrgencyKeyFunderBalance`, `highUrgencyRelayerBalance` and writing them to checked in config
```
yarn tsx ./scripts/funding/read-alert-thresholds.ts -e mainnet3 --alertType highUrgencyRelayerBalance --write
```

- Support updating checked in config based on the dailyBurn and multipliers
image

- Support writing alert queries to grafana
image

Note, the configs are not in use at the moment and will be used as we roll this out

### Testing

Manual


typescript/infra/config/environments/mainnet3/balances/dailyBurn.json
```diff
@@ -0,0 +1,119 @@
+{
```
Mo-Hussain, line 1: config checked in to make it easy to inspect and review changes

Mo-Hussain:
I have scripts that uses the config to write the alerts into grafana using the API pretty much ready. I wanted to use the opportunity to see how much of a lift it would be to use terraform provisioned alerts. These checked in configs create a nice interface for the grafana API approach vs terraform

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const alertWalletNameMapping: Record<AlertTypeValue, WalletName> = {
+  [AlertType.LowUrgencyKeyFunderBalance]: WalletName.KeyFunder,
+  [AlertType.LowUrgencyEngKeyFunderBalance]: WalletName.KeyFunder,
+  [AlertType.HighUrgencyRelayerBalance]: WalletName.Relayer,
+};
+
+const alertIdMapping: Record<AlertTypeValue, string> = {
+  [AlertType.LowUrgencyKeyFunderBalance]: 'KiJNg6p4k',
+  [AlertType.LowUrgencyEngKeyFunderBalance]: 'temp',
```
Mo-Hussain, line 37: Haven't created this alert yet so using a dummy id for now, we wont be able to read from it atm

typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -2,9 +2,12 @@ import { KeyFunderConfig } from '../../../src/config/funding.js';
 import { Role } from '../../../src/roles.js';
 import { Contexts } from '../../contexts.js';
 
+import rawDesiredBalancePerChain from './balances/desiredBalances.json';
 import { environment } from './chains.js';
 import { mainnet3SupportedChainNames } from './supportedChainNames.js';
 
+const desiredBalancePerChain = rawDesiredBalancePerChain;
```
paulbalaji, line 9: why is this line necessary?

typescript/infra/scripts/agent-utils.ts
```diff
@@ -90,6 +93,12 @@ export function getArgs() {
     .alias('e', 'environment');
 }
 
+export function withBalanceThresholdConfig<T>(args: Argv<T>) {
```
paulbalaji, line 96: starting to think these withXYZ helpers should be in their own `*-utils.ts` file, since these have nothing to do with agents

typescript/infra/scripts/funding/utils/utils.ts
```diff
@@ -0,0 +1,5 @@
+export function formatDailyBurn(dailyBurn: number) {
```
paulbalaji, line 1: this feels like a handy to thing to have in utils? like more generically "format to X sf"

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export type AlertTypeValue = `${AlertType}`;
+
+export enum WalletName {
+  KeyFunder = 'keyFunder',
+  Relayer = 'relayer',
+}
+
+export type WalletNameValue = `${WalletName}`;
```
paulbalaji, line 27: i don't see this used anywhere

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export const logger = rootLogger.child({ module: 'grafana' });
+
+export enum AlertType {
+  LowUrgencyKeyFunderBalance = 'lowUrgencyKeyFunderBalance',
+  LowUrgencyEngKeyFunderBalance = 'lowUrgencyEngKeyFunderBalance',
+  HighUrgencyRelayerBalance = 'highUrgencyRelayerBalance',
+}
+
+export type AlertTypeValue = `${AlertType}`;
```
paulbalaji, line 20: why is this necessary? just use `AlertType`

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+    relativeTimeRange: {
+      from: number;
+      to: number;
+    };
+    datasourceUid: string;
+    model: {
+      expr: string;
+    };
+  }>;
+}
```
paulbalaji, line 66: if possible add a docs reference for where this type comes from

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+  alertType: AlertTypeValue,
+): Promise<ChainMap<number>> {
+  const saToken = await fetchServiceAccountToken();
+  const alert = await getGrafanaAlert(alertType, saToken);
+  const alertQuery = alert.queries[0];
+  const walletName = alertWalletNameMapping[alertType];
+  return parsePromQLQuery(alertQuery, walletName);
+}
+
+export async function fetchServiceAccountToken(): Promise<string> {
```
paulbalaji, line 150: does it need to be exported?

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,105 @@
...
+      currentThresholds = readJSONAtPath(
+        `${THRESHOLD_CONFIG_PATH}/${configFileNameMapping[config]}`,
+      );
+    } catch (e) {
+      rootLogger.error(`Error reading ${config} config: ${e}`);
+    }
+
+    // Update the threshold for each chain, if it doesn't exist, create a new one
+    for (const chain in dailyBurn) {
+      if (!(chain in currentThresholds)) {
```
paulbalaji, line 75: nit: i generally prefer this
```suggestion
      if (!currentThresholds[chain]) {
```

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -0,0 +1,144 @@
...
+    dailyBurn,
+  }));
+
+  console.table(burnArray);
+
+  const chainsMissingInTokenPrices = mainnet3SupportedChainNames.filter(
+    (chain) => !(chain in tokenPrices),
+  );
+
+  if (chainsMissingInTokenPrices.length > 0) {
```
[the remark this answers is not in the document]
paulbalaji: yeah this should process.exit(1) after the error

typescript/infra/scripts/funding/utils/grafana.ts
paulbalaji: ditto for a few other exported fns in this file

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,105 @@
...
+    const newThresholds: ChainMap<number> = {};
+    try {
+      currentThresholds = readJSONAtPath(
+        `${THRESHOLD_CONFIG_PATH}/${configFileNameMapping[config]}`,
+      );
+    } catch (e) {
+      rootLogger.error(`Error reading ${config} config: ${e}`);
+    }
+
+    // Update the threshold for each chain, if it doesn't exist, create a new one
```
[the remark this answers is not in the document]
paulbalaji: what's the exact flow you would expect someone to go through here? 

1. print-token-prices
2. calculate-relayer-daily-burn
3. update-balance-threshold-config
4. deploy keyfunder
5. write-alerts (future PR?)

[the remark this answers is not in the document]
paulbalaji: also how can we account for chains where the token doesn't have a price yet so we assume 1 USD -> 25 token desired min balance, but we might only get like 1 token from the dev team for deployment? 

is that something that we'd manually override in the json afterwards?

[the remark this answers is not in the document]
Mo-Hussain: I was thinking 

1. print-token-prices
2. calculate-relayer-daily-burn
3. update-balance-threshold-config
4. PR, Review, Merge
5. deploy keyfunder
6. write-alerts

will 5 could happen before 4 if that's your usual flow

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+export type AlertTypeValue = `${AlertType}`;
+
+export enum WalletName {
+  KeyFunder = 'keyFunder',
+  Relayer = 'relayer',
+}
+
+export type WalletNameValue = `${WalletName}`;
+
+const alertWalletNameMapping: Record<AlertTypeValue, WalletName> = {
```
tkporter, line 29: would also suggest unifying these 3 maps into a single config

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,105 @@
...
+  configureRootLogger(LogFormat.Pretty, LogLevel.Info);
+  const { balanceThresholdConfig } = await withBalanceThresholdConfig(
+    yargs(process.argv.slice(2)),
+  ).argv;
+
+  const configToUpdate: BalanceThresholdConfig[] = balanceThresholdConfig
+    ? [balanceThresholdConfig]
+    : [];
+
+  if (configToUpdate.length === 0) {
```
tkporter, line 47: is there a downside to just always requiring that all configs are updated? can't think of a scenario where we wouldn't wanna update all of them just in case

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+  let saToken: string | undefined;
+
+  try {
+    saToken = (await fetchGCPSecret(
+      'grafana-balance-alert-thresholds-token',
+      false,
+    )) as string;
+  } catch (error) {
+    logger.error(
+      'Error fetching grafa service account token from GCP secrets:',
```
tkporter, line 160: nit: typo grafa

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const queries = data.data.map((d) => d.model.expr);
+
+    return {
+      title: data.title,
+      queries,
+      rawData: data,
+    };
+  } catch (error) {
+    if (error instanceof Error) {
```
tkporter, line 99: maybe just a stylistic difference but imo a try/catch here is unnecessary just to log and throw again, this should be the consumer's problem. Generally I think too many try catches for scripts that don't need to be super tolerant of errors make for hard to read code

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export enum AlertType {
+  LowUrgencyKeyFunderBalance = 'lowUrgencyKeyFunderBalance',
+  LowUrgencyEngKeyFunderBalance = 'lowUrgencyEngKeyFunderBalance',
+  HighUrgencyRelayerBalance = 'highUrgencyRelayerBalance',
+}
+
+export type AlertTypeValue = `${AlertType}`;
+
+export enum WalletName {
```
tkporter, line 22: I assume future plans would include ATA payers?

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,105 @@
...
+import { formatDailyBurn } from './utils/utils.js';
+
+const dailyBurn: ChainMap<number> = rawDailyBurn;
+
+export const configChoiceLabelMapping: Record<BalanceThresholdConfig, string> =
+  {
+    [BalanceThresholdConfig.RelayerBalance]: 'Desired Relayer Balance',
+    [BalanceThresholdConfig.LowUrgencyKeyFunderBalance]:
+      'Low Urgency Key Funder Balance Alert Threshold',
+    [BalanceThresholdConfig.LowUrgencyEngKeyFunderBalance]:
```
tkporter, line 31: any reason why this config isn't checked in?

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,105 @@
...
+} from './utils/constants.js';
+import { THRESHOLD_CONFIG_PATH } from './utils/grafana.js';
+import { formatDailyBurn } from './utils/utils.js';
+
+const dailyBurn: ChainMap<number> = rawDailyBurn;
+
+export const configChoiceLabelMapping: Record<BalanceThresholdConfig, string> =
+  {
+    [BalanceThresholdConfig.RelayerBalance]: 'Desired Relayer Balance',
+    [BalanceThresholdConfig.LowUrgencyKeyFunderBalance]:
```
tkporter, line 29: this one is supposed to consider the desired kathy balances right? or remind me where we settled with this

tkporter (commented):
As many of the values seem wrong, I think we should have some mechanisms to sanity check them when they're generated, e.g. > 0, ensure that dailyBurn < highUrgencyRelayerBalances < etc

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: The alert for this does not exist yet

typescript/infra/scripts/funding/utils/grafana.ts
Mo-Hussain: Yes

typescript/infra/config/environments/mainnet3/balances/highUrgencyRelayerBalance.json
```diff
@@ -0,0 +1,108 @@
...
+  "bitlayer": 0.000069,
+  "blast": 0.0062,
+  "bob": 0.0015,
+  "boba": 0.000011,
+  "bsc": 0.47,
+  "bsquared": 0.000026,
+  "celo": 1.2,
+  "cheesechain": 0.024,
+  "chilizmainnet": 12.32,
+  "conflux": 0.00084,
```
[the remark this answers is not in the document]
Mo-Hussain: I've clarified in the PR description, the current checked in version of `lowUrgencyKeyFunderBalance` and `highUrgencyRelayerBalance` are reading current thresholds not the proposed ones. The thinking was that we'd propose the new ones when we had the ability to write them through the API. Happy to check in the proposed ones now for review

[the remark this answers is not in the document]
Mo-Hussain: We want to be careful with how we make the changes to these alert/balance thresholds i.e we we want to stage them

1. LowUrgencyKeyFunder: we would want to share the new balances with Jake before we update this alert to over load him with alerts
2. LowUrgencyEngKeyFunder: set this once Jake has updated the key funder balances
3. desiredRelayerBalances: once the key funder has sufficient balance, can happen at the same time as 2
4. HighUrgencyRelayerBalance: once keyfunder has run successfully after 3  

With this in mind I was initially hesitant to check them in now

typescript/infra/scripts/funding/update-balance-threshold-config.ts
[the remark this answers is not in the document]
Mo-Hussain: If a token does not have a price, how do we handle things now? We don't set a token price? Do we typically run key funder for those chains?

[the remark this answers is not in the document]
paulbalaji: if there's no price we default to 1 here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/47ae33c6a4c58c350b6bd724a177289635d4bdd2/typescript/infra/scripts/print-token-prices.ts#L62-L79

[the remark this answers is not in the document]
paulbalaji: We still run key funder yes - but keyfunder doesn't care about price atm, just the native token balances

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: We settled in kathy balances being negligible but will confirm that this makes sense

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -0,0 +1,144 @@
...
+  configureRootLogger(LogFormat.Pretty, LogLevel.Info);
+
+  const sql = await getReadOnlyScraperDb();
+  let burnData: ChainMap<number>;
+  try {
+    burnData = await getRelayerDailyBurn(sql);
+  } catch (err) {
+    rootLogger.error('Error fetching daily burn data:', err);
+    process.exit(1);
+  } finally {
```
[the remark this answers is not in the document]
Mo-Hussain: We need to close the sql connection otherwise the script can hang, happy to get rid of the try catch and just close the connection after querying

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    // some chains may have had no messages in the last 10 days so we set daily burn based in the minimum dollar balance
+    const proposedDailyBurn =
+      row === undefined
+        ? minNativeBalance
+        : Math.max(row.daily_burn, minNativeBalance);
+
+    // only update the daily burn if the proposed daily burn is greater than the current daily burn
+    // add the chain to the daily burn if it doesn't exist
+    const newDailyBurn =
```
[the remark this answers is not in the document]
Mo-Hussain: Happy to do some but this should be obvious when comparing the diffs before commiting?

typescript/infra/config/environments/mainnet3/balances/highUrgencyRelayerBalance.json
```diff
@@ -0,0 +1,108 @@
...
+  "real": 0.00022,
+  "redstone": 0.0028,
+  "rivalz": 7.7e-7,
+  "rootstockmainnet": 0.00006,
+  "sanko": 0.0015,
+  "scroll": 0.022,
+  "sei": 3.59,
+  "shibarium": 0.03,
+  "snaxchain": 0.00036,
+  "solana": 0.75,
```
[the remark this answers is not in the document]
Mo-Hussain: Config has been read from the existing query 

```
  # Nautilus (v2)
  last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="nautilus"}[1d]) - 1500 or
  # TODO: we can re-add mainnet3 after the upcoming Solana deploy
  last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="solana", hyperlane_deployment!="mainnet3"}[1d]) - 0.75 or
  last_over_time(hyperlane_wallet_balance{wallet_name="nautilus-zbc-rent", chain="solana"}[1d]) - 0.75 or
```

Going to remove this assuming nautilus is no longer running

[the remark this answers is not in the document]
Mo-Hussain: This is reading the existing alert

```
  # Nautilus (v2)
  last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="nautilus"}[1d]) - 1500 or
  # TODO: we can re-add mainnet3 after the upcoming Solana deploy
  last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="solana", hyperlane_deployment!="mainnet3"}[1d]) - 0.75 or
  last_over_time(hyperlane_wallet_balance{wallet_name="nautilus-zbc-rent", chain="solana"}[1d]) - 0.75 or
```

Going to remove these?

typescript/infra/scripts/funding/update-balance-threshold-config.ts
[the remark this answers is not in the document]
Mo-Hussain: I would suggest in these situation, we manually override what is outputted in the `dailyRelayerBurn.json` that was you control the other thresholds

typescript/infra/scripts/agent-utils.ts
```diff
@@ -155,6 +164,13 @@ export function withChain<T>(args: Argv<T>) {
     .alias('c', 'chain');
 }
 
+export function withWrite<T>(args: Argv<T>) {
```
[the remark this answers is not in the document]
Mo-Hussain: `withOutputFile` expects a string argument for the output file, this seems redundant as we want to save to the same file everytime

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: I'll do this as a quick follow up

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -0,0 +1,119 @@
...
+  "boba": 0.0035,
+  "bsc": 0.161,
+  "bsquared": 0.00361,
+  "celo": 5.37,
+  "cheesechain": 3980,
+  "chilizmainnet": 103,
+  "conflux": 22.9,
+  "conwai": 844,
+  "coredao": 3.43,
+  "corn": 0.0000336,
```
Mo-Hussain, line 29: Seems low but we've barely any usage

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -0,0 +1,119 @@
...
+  "arbitrumnova": 0.00099,
+  "artela": 3.13,
+  "arthera": 48.2,
+  "astar": 66.6,
+  "astarzkevm": 0.00476,
+  "aurora": 0.00099,
+  "flame": 2.71,
+  "avalanche": 0.0908,
+  "b3": 0.00411,
+  "base": 0.0566,
```
Mo-Hussain, line 16: This seems a little low, burn in the last 24 hours was 0.09

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -0,0 +1,119 @@
...
+  "mode": 0.0136,
+  "molten": 4.56,
+  "moonbeam": 25.9,
+  "morph": 0.262,
+  "nero": 3.13,
+  "neutron": 10,
+  "oortmainnet": 24,
+  "optimism": 0.0314,
+  "orderly": 0.0047,
+  "osmosis": 7.36,
```
Mo-Hussain, line 76: My dashboard suggests we have a zero balance on the relayer, what is the deal with Osmosis @paulbalaji

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -0,0 +1,119 @@
...
+  "redstone": 0.00369,
+  "rivalz": 0.00099,
+  "rootstockmainnet": 0.00241,
+  "sanko": 0.111,
+  "scroll": 0.0168,
+  "sei": 8.8,
+  "shibarium": 7.29,
+  "snaxchain": 0.00417,
+  "solanamainnet": 5.3,
+  "soneium": 0.00628,
```
Mo-Hussain, line 93: Seems low

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -2,121 +2,118 @@
...
-  "lisk": "0.05",
+  "lisk": "0.0635",
   "lukso": "20",
-  "lumia": "1",
-  "lumiaprism": "1",
+  "lumia": "20.5",
+  "lumiaprism": "24.3",
   "mantapacific": "0.2",
-  "mantle": "20",
-  "matchain": "0.05",
```
Mo-Hussain: Why was this removed?

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -2,121 +2,118 @@
...
   "mode": "0.2",
-  "molten": "3",
-  "moonbeam": "100",
-  "morph": "0.1",
-  "nero": "5",
+  "molten": "36.5",
+  "moonbeam": "207",
+  "morph": "2.1",
+  "nero": "25",
+  "neutron": "80",
```
Mo-Hussain, line 72: This will result in an issue with key funder

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -2,121 +2,118 @@
...
   "oortmainnet": "2000",
   "optimism": "0.5",
   "orderly": "0.05",
-  "polygon": "40",
+  "osmosis": "58.9",
```
Mo-Hussain, line 76: We don't want this here @paulbalaji?

paulbalaji: I'm not sure what's going on with osmosis tbh - I think we are able to relay there but don't think we're actively relaying anything?

Mo-Hussain: For now I'll make sure the desired balance is set to zero

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -0,0 +1,370 @@
...
+  const queries = data.data.map((d) => d.model.expr);
+
+  return {
+    title: data.title,
+    queries,
+    rawData: data,
+  };
+}
+
+export async function exportGrafanaAlert(
```
Mo-Hussain, line 209: Export function was helpful in confirming that the alert config pre and post query update, only changed the query and not any other alert config

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -1,63 +1,60 @@
...
   );
 
-  if (chainsMissingInTokenPrices.length > 0) {
-    rootLogger.error(
-      `Token prices missing for chains: ${chainsMissingInTokenPrices.join(
-        ', ',
-      )} consider adding them to tokenPrices.json and running the script again.`,
-    );
-  }
+  console.table(burnArray);
```
paulbalaji, line 57: what happens if you just console.table(burnData)?

typescript/infra/scripts/funding/read-alert-thresholds.ts
```diff
@@ -1,30 +1,26 @@
...
   const { alertType, write } = await withWrite(
-    withAlertType(yargs(process.argv.slice(2))),
+    withAlertTypeRequired(yargs(process.argv.slice(2))),
   ).argv;
 
   const alertThresholds = await getAlertThresholds(alertType);
+  const orderedThresholds = orderThresholds(alertThresholds);
 
-  const alertThresholdArray = Object.entries(alertThresholds).map(
+  const alertThresholdArray = Object.entries(orderedThresholds).map(
```
paulbalaji, line 23: similar q here what happens if you just console.table(orderedThresholds)

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -2,93 +2,96 @@ import { checkbox } from '@inquirer/prompts';
...
-    configToUpdate.push(...selectedConfig);
-  }
+    : await checkbox({
+        message: 'Select the balance threshold config to update',
+        choices: Object.values(BalanceThresholdType).map((config) => ({
+          name: balanceThresholdConfigMapping[config].choiceLabel,
+          value: config,
+          checked: true, // default to all checked
+        })),
+      });
```
paulbalaji, line 39: can i also get a cli option that skips the prompt entirely and goes straight to the updates? something like `--yes` or `--all`

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -157,11 +278,93 @@ export async function fetchServiceAccountToken(): Promise<string> {
...
+  );
+
+  return `${config.queryTemplate.header}
+    ${queryFragments.join('\n    ')}
+
+${config.queryTemplate.footer}
+)`;
+}
+
+export function orderThresholds(
```
paulbalaji, line 360: nit: `sortThresholds`

typescript/infra/scripts/funding/utils/grafana.ts
```diff
@@ -17,108 +28,218 @@ export enum AlertType {
...
+}
 
-export const alertThresholdFileMapping: Record<AlertTypeValue, string> = {
-  [AlertType.LowUrgencyKeyFunderBalance]:
-    configFileNameMapping[BalanceThresholdConfig.LowUrgencyKeyFunderBalance],
-  [AlertType.LowUrgencyEngKeyFunderBalance]:
-    configFileNameMapping[BalanceThresholdConfig.LowUrgencyEngKeyFunderBalance],
-  [AlertType.HighUrgencyRelayerBalance]:
-    configFileNameMapping[BalanceThresholdConfig.HighUrgencyRelayerBalance],
+export const alertConfigMapping: Record<AlertType, AlertConfig> = {
```
paulbalaji, line 54: does it still make sense to have this here or should we have a separate file in infra/config for this? only asking because it's like 50 lines of config that gets reused across a couple files

Mo-Hussain: Happy to move to a separate config file

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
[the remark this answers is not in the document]
Mo-Hussain: Will do

typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -2,7 +2,6 @@ import { KeyFunderConfig } from '../../../src/config/funding.js';
 import { Role } from '../../../src/roles.js';
 import { Contexts } from '../../contexts.js';
 
-import desiredBalancePerChain from './balances/desiredRelayerBalances.json';
```
Mo-Hussain: Do not use the new desired relayer balances yet

typescript/infra/config/grafanaAlerts.ts
```diff
@@ -0,0 +1,164 @@
...
+    walletName: WalletName.KeyFunder,
+    grafanaAlertId: 'ae9z3blz6fj0gb',
+    configFileName:
+      balanceThresholdConfigMapping[
+        BalanceThresholdType.LowUrgencyKeyFunderBalance
+      ].configFileName,
+    choiceLabel:
+      balanceThresholdConfigMapping[
+        BalanceThresholdType.LowUrgencyKeyFunderBalance
+      ].choiceLabel,
```
paulbalaji, line 52: nit: any harm in doing this? think we end up passing along `dailyRelayerBurnMultiplier` as well but not a big deal right, if it makes this a little cleaner
```suggestion
    ...balanceThresholdConfigMapping[
      BalanceThresholdType.LowUrgencyKeyFunderBalance
    ],
```

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  if (chainsMissingInTokenPrices.length > 0) {
+    rootLogger.error(
+      `Token prices missing for chains: ${chainsMissingInTokenPrices.join(
+        ', ',
+      )} consider adding them to tokenPrices.json and running the script again.`,
+
+      process.exit(1),
+    );
```
paulbalaji, line 36: think you meant to exit after the log
```suggestion
    );
    process.exit(1);
```

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -0,0 +1,139 @@
...
+            ) / POWER(10, 18)
+          ) / COUNT(*) AS avg_tx_cost_native,
+          COUNT(*) / ${LOOK_BACK_DAYS} AS avg_daily_messages
+        FROM
+          message_view mv
+          LEFT JOIN DOMAIN dest_domain ON mv.destination_domain_id = dest_domain.id
+        WHERE
+          mv.send_occurred_at >= CURRENT_TIMESTAMP - (INTERVAL '1 day' * ${LOOK_BACK_DAYS})
+          AND dest_domain.is_test_net IS FALSE
+          AND mv.destination_domain_id not in (1408864445, 1399811149) -- ignore sealevel chains as scrapper does not capture all costs
```
paulbalaji, line 87:
```suggestion
          AND mv.destination_domain_id not in (1408864445, 1399811149) -- ignore sealevel chains as scraper does not capture all costs
```

typescript/infra/scripts/funding/utils/constants.ts
paulbalaji, line 1: this could probably go in config too?

typescript/infra/scripts/funding/utils/grafana.ts
paulbalaji, line 1: could move to src/funding/grafana.ts

paulbalaji: i like to think infra/scripts is where we have main() fns that we would call directly

and infra/src is where we can have utils like this. i know we're p bad at following through with this with a bunch of existing scripts (looking at agent-utils) but since it's a bunch of new code it's probably good to keep that in mind

paulbalaji (commented):
mostly lgtm with some nits / file moving

typescript/infra/config/grafanaAlerts.ts
Mo-Hussain: Agree that it looks cleaner but I think its best to avoid mixing configs where possible specially given that we may add more fields to `balanceThresholdConfigMapping` in future as this evolves

paulbalaji: sgtm

paulbalaji (approved):

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -0,0 +1,124 @@
...
+  "chilizmainnet": "800",
+  "conflux": "172",
+  "conwai": "7110",
+  "coredao": "36.5",
+  "corn": "0.001",
+  "cyber": "0.05",
+  "degenchain": "3850",
+  "dogechain": "100",
+  "duckchain": "5.14",
+  "eclipsemainnet": "0",
```
Mo-Hussain, line 34: Chains that are not supported by key-funder are set to zero, alert threshold will still be set appropriately. The update script will not update desired relayer balances if the current value is zero

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -0,0 +1,124 @@
+{
+  "abstract": "0",
```
Mo-Hussain, line 2: Discussed with @paulbalaji that where we want to add a new chain and the desired relayer balance should be 0, manually deletion of the alert thresholds is needed. We did discuss the possibility to add a script to support this but he felt it was ok to have it as manual

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,111 @@
...
+} from '../../src/funding/grafana.js';
+import { readJSONAtPath, writeJsonAtPath } from '../../src/utils/utils.js';
+import {
+  withBalanceThresholdConfig,
+  withConfirmAllChoices,
+} from '../agent-utils.js';
+
+const dailyBurn: ChainMap<number> = rawDailyBurn;
+
+const exclusionList = ['osmosis'];
```
tkporter, line 25: what chains would go here?

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,111 @@
...
+          rootLogger.info(`Skipping ${chain} as it is in the exclusion list`);
+          continue;
+        }
+
+        newThresholds[chain] = formatDailyRelayerBurn(
+          dailyBurn[chain] *
+            balanceThresholdConfigMapping[config].dailyRelayerBurnMultiplier,
+        ).toString();
+      } else {
+        // This will ensure that chains where the desired threshold is 0 will be unchanged
```
tkporter, line 72: why do we want this? can you add a comment

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (
+          config === BalanceThresholdType.RelayerBalance &&
+          parseFloat(currentThresholds[chain]) === 0
+        ) {
+          newThresholds[chain] = currentThresholds[chain];
+          continue;
+        }
+
+        newThresholds[chain] = Math.max(
+          formatDailyRelayerBurn(
```
tkporter, line 82: nit: as we use this function in places where it's not just the daily relayer burn, maybe there should be a more generic name? `formatBalanceThreshold` or something

typescript/infra/src/funding/grafana.ts
```diff
@@ -0,0 +1,207 @@
...
+  );
+
+  return `${config.queryTemplate.header}
+    ${queryFragments.join('\n    ')}
+
+${config.queryTemplate.footer}
+)`;
+}
+
+export function sortThresholds(
```
tkporter, line 197: whoa I had no idea Javascript objects had a stable ordering of keys following the insertion order, neat. Maybe this is better in `utils/objects.ts` as a generic `sortObject` or something? with a comment why it works, imo it's not obvious

typescript/infra/src/config/funding/alert-query-templates.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const LOW_URGENCY_KEY_FUNDER_FOOTER = `    # Mainnets that don't use key-funder and all funds are stored in the relayer's wallet
+    # Eclipse
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", hyperlane_context="hyperlane", chain=~"eclipsemainnet"}[1d]) - 1 or
+    # Any ATA payer on Eclipse
+    last_over_time(hyperlane_wallet_balance{wallet_name=~".*/ata-payer", chain=~"eclipsemainnet"}[1d]) - 0.01 or
+    # SOL/eclipsemainnet-solanamainnet
+    last_over_time(hyperlane_wallet_balance{wallet_name=~"SOL/eclipsemainnet-solanamainnet/ata-payer | USDC/eclipsemainnet-ethereum-solanamainnet/ata-payer", chain=~"eclipsemainnet"}[1d]) - 0.1 or
+
+    # Solana
```
tkporter, line 16: I know it's not already there but can you add SOON in here? was an oversight of mine

Also noting to myself that I'll need to add Sonic here once the deploy is finished

typescript/infra/src/config/funding/balances.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export const THRESHOLD_CONFIG_PATH = './config/environments/mainnet3/balances';
+
+const RELAYER_BALANCE_TARGET_DAYS = 8;
+const RELAYER_MIN_DOLLAR_BALANCE_TARGET = 25;
+export const RELAYER_MIN_DOLLAR_BALANCE_PER_DAY =
+  RELAYER_MIN_DOLLAR_BALANCE_TARGET / RELAYER_BALANCE_TARGET_DAYS;
+
+export const balanceThresholdConfigMapping: Record<
```
tkporter, line 21: very nice, much easier to read

typescript/infra/src/config/funding/balances.ts
```diff
@@ -0,0 +1,45 @@
...
+  BalanceThresholdType,
+  BalanceThresholdConfig
+> = {
+  [BalanceThresholdType.RelayerBalance]: {
+    configFileName: 'desiredRelayerBalances.json',
```
tkporter, line 26: any reason to have this be different from the rest? wonder if we should just have this type be called `BalanceThresholdType.DesiredRelayerBalance`

typescript/infra/src/config/funding/alert-query-templates.ts
```diff
@@ -0,0 +1,56 @@
...
+    last_over_time(hyperlane_wallet_balance{wallet_name=~".*/ata-payer", chain=~"solanamainnet"}[1d]) - 0.1 or
+
+    # SOON
+    # Any ATA payer on SOON
+    last_over_time(hyperlane_wallet_balance{wallet_name=~".*/ata-payer", chain=~"soon"}[1d]) - 0.005 or
+
+    # Neutron context lines stay with neutron context
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="mantapacific", hyperlane_context="neutron"}[1d]) - 0.02 or
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="arbitrum", hyperlane_context="neutron"}[1d]) - 0.02 or
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="neutron", hyperlane_context="neutron"}[1d]) - 0.7
```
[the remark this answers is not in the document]
Mo-Hussain: Will handle this as a follow up

tkporter (commented):
nice, looking really good - no super huge comments but more trying to understand what to expect after this is merged, like what is the new process for updating an alert atm

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: We've strongly types the `KeyFunder` config so that we need to include every mainnet3 chains in `desiredRelayerBalance`. IIUC, we don't run a relayer for osmosis so we don't want to set alerts for the chain. I'm not sure if any other chains will fall into this bracket in future

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: The intention here is to not update the desired relayer balance of chains that have an explicit zero value (new chains). 

It was also supposed to not update non-evm chains, which also currently have a zero values. Although having looked at key-funder code, I can see that we explicitly ignore non-evm chains, so there is no harm in updating these values.

I agree that this is confusing at the moment and the logic is not clear. I'm going to redesign this and use an override config to handle any special cases and make this logic simple

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -0,0 +1,124 @@
...
+  "chilizmainnet": 100,
+  "conflux": 21.5,
+  "conwai": 889,
+  "coredao": 4.56,
+  "corn": 0.0000316,
+  "cyber": 0.00443,
+  "degenchain": 481,
+  "dogechain": 10.5,
+  "duckchain": 0.643,
+  "eclipsemainnet": 0.025,
```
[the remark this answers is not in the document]
Mo-Hussain: At the moment yes, I am looking into how we can automate this, I will investigate a promQL query

typescript/infra/scripts/funding/read-alert-thresholds.ts
```diff
@@ -0,0 +1,41 @@
+import yargs from 'yargs';
```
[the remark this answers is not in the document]
Mo-Hussain: This was initially used by me to read the current thresholds from the alerts, I don't expect it to be used in the process. Happy to remove it.

typescript/infra/src/config/funding/alert-query-templates.ts
```diff
@@ -0,0 +1,56 @@
+// alert queries currently need to support special cases i.e cross VM (sealevel and cosmos chains) and ata payer (sealevel). These special cases are hard coded for now. We aim to add cross VM support to the key and will be able to remove special casing in the future
+export const LOW_URGENCY_KEY_FUNDER_HEADER = `# Note: use last_over_time(hyperlane_wallet_balance{}[1d]) to be resilient to gaps in the 'hyperlane_wallet_balance'
+# that occur due to key funder only running every hour or so.
+
+min by (chain, wallet_address, wallet_name) (
+    # Mainnets`;
+
+export const LOW_URGENCY_KEY_FUNDER_FOOTER = `    # Mainnets that don't use key-funder and all funds are stored in the relayer's wallet
+    # Eclipse
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", hyperlane_context="hyperlane", chain=~"eclipsemainnet"}[1d]) - 1 or
```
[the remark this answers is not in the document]
Mo-Hussain: In the low urgency key funder alert, we will auto add lines that will attempt to read key-funder balances for the non-evm chains which will have not data until key-funder supports those chains, so there is no harm in adding them. We still want to include these relayer balances line in the footer in the mean time before we add keyfunder support for these chains

typescript/infra/scripts/funding/write-alert.ts
```diff
@@ -0,0 +1,72 @@
+import { checkbox } from '@inquirer/prompts';
```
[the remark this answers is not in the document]
Mo-Hussain: I should have done a better job of laying this out in the comments, I wanted to build up the changes in smaller chunks to reduce the risk as we rolled out these changes. Your general understanding is correct, I'd also add

typescript/infra/scripts/funding/update-balance-threshold-config.ts
tkporter: ah that's interesting.. funnily we do technically have the relayer run for osmosis but we just don't expect any use because it's in a half-finished state, lol. So I'd generally expect no other chains to qualify here. Can you add a comment why osmosis is a special case? wanna make sure if we do close out the osmosis relayer that we know to remove this

typescript/infra/scripts/funding/update-balance-threshold-config.ts
tkporter: gotcha, yeah I think minimizing the # of special cases will make things more clear. So down to have the balances in the JSONs be script-generated so that we only need to modify key funder to connect the pieces together

typescript/infra/src/config/funding/alert-query-templates.ts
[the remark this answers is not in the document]
tkporter: ahh yes this makes sense, I forgot that we wouldn't have auto-added relayer low urgency alerts

typescript/infra/scripts/funding/read-alert-thresholds.ts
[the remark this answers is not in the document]
tkporter: happy either way, if you think it'd still be useful for the future then I'm good with a comment and keeping it, up to you

Issue #5267, opened 2025-01-22
Unify cli --registry and --overrides flags


nambrot:
Very similar to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5224

1. Remove override
2. Convert registry to array structure
3. Include default values in the registry array (GitHub and local registry)

![Image](https://github.com/user-attachments/assets/fc3a8c38-1849-47c2-8991-3b33379c7032)

![Image](https://github.com/user-attachments/assets/edf19469-85fd-4f42-a406-d2b2b303f349)

![Image](https://github.com/user-attachments/assets/aae3936c-3263-4a7f-9adc-8c2263912427)

![Image](https://github.com/user-attachments/assets/b81c0353-a0ad-450d-9b65-af6172b49009)

![Image](https://github.com/user-attachments/assets/66fadb9f-d05f-4294-bd45-95ec7d1fc63f)

PR #5271, opened 2025-01-23
refactor: simplify registry override

mshojaei-txfusion:
### Description

This PR includes two main improvements to the CLI's registry and parameter handling:

1. Removes the unused `--overrides` parameter from all warp command functions, which was previously only used for local testing purposes but is no longer needed.

2. Enhances the registry handling by:
   - Converting the registry parameter from a single URI to an array structure
   - Adding support for multiple registry sources, later registry takes priority over previous
   - Including default values in the registry array (GitHub and local registry)

These changes simplify the codebase and make the registry configuration more flexible and explicit.

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5267

### Backward compatibility

Yes - These changes are backward compatible. The removal of the unused `--overrides` parameter doesn't affect functionality since it was not being actively used. The registry array structure maintains compatibility with existing registry configurations while adding support for multiple sources.

### Testing

Manual testing


typescript/cli/cli.ts
```diff
@@ -43,8 +42,7 @@ try {
     .scriptName('hyperlane')
     .option('log', logFormatCommandOption)
     .option('verbosity', logLevelCommandOption)
-    .option('registry', registryUriCommandOption)
-    .option('overrides', overrideRegistryUriCommandOption)
```
nambrot, line 47: Are we sure that this is not used at all? I believe it actually is, so I would like to explicitly bump major versions for this and call this out in the changelog

nambrot (approved):
Approving pending changeset addition

typescript/cli/cli.ts
mkykadir: yeah that's a breaking change, it should be a major change

mkykadir (approved):

typescript/cli/cli.ts
nambrot: Actually this makes me think, can we just keep it open for backwards compatability? I.e. have `--registry` support multple, but `--overrides` just adds it to the end?

mshojaei-txfusion: makes sense, added.

nambrot (approved):

Issue #5255, opened 2025-01-22
[infra] list all announced validator addresses for a given chain


daniel-savu:
- to make debugging validator liveness and checkpoints easier
- if we convert the announced `s3`/`gcs` path strings to URLs, we can also fetch and print `/checkpoint_latest_index.json` for each validator, to help us debug whether a validator is really behind the others (since the relayer metric can have false positives)

PR #5286, opened 2025-01-24
feat(infra): script to print latest checkpoints with urls

paulbalaji:
### Description

- improve script for printing latest validator checkpoints
- group misc validator scripts into a new `validators` folder in `infra/scripts`
- iterates over ALL announced validators when ran with `--all`/`-a` flag

### Related issues

resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5255

### Testing

manual

image
image


typescript/infra/scripts/validators/print-latest-checkpoints.ts
```diff
@@ -0,0 +1,124 @@
...
+function getHttpsUrl(storageLocation: string): string {
+  if (storageLocation.startsWith('s3://')) {
+    // Convert s3:///bucket-name/region to https://bucket-name.s3.region.amazonaws.com
+    const [, , bucket, region] = storageLocation.split('/');
+    return `https://${bucket}.s3.${region}.amazonaws.com`;
+  } else if (storageLocation.startsWith('gs://')) {
+    // Convert gs://bucket-name to https://storage.googleapis.com/bucket-name
+    const bucket = storageLocation.replace('gs://', '');
+    return `https://storage.googleapis.com/${bucket}`;
+  }
```
daniel-savu, line 28: iirc we do already have this logic somewhere in the TS codebase, at least for aws

daniel-savu (approved):
💙

typescript/infra/scripts/validators/print-latest-checkpoints.ts
paulbalaji: indeed, this function is a relic before i switched to calling `getLatestCheckpointUrl`

Issue #5277, opened 2025-01-23
Script/CI job that ensures that browser-unfriendly dependencies are not required for sdk


nambrot:
![](https://cdn.discordapp.com/attachments/935678739663192184/1332038993641996331/Screenshot_2025-01-23_at_1.24.55_PM.png?ex=6793cd7c&is=67927bfc&hm=b0bc3b980705e1ae8cb662ef7ad89a14d2c62ce55ab05f45ce4a9175e748e2be&)

The SDK is used in the browser and when packages like fs or path are imported, they break the sdk consumer. We should have processes in place that prevent these dependencies from being required.

PR #5288, opened 2025-01-24
fix: browser compatibility checks

mshojaei-txfusion:
### Description

This PR enhances the ESLint configuration to restrict the usage of Node.js built-in modules across the codebase. This is particularly important for the SDK package which needs to maintain browser compatibility for frontend integrations.

Key changes:
- Added comprehensive ESLint rules to prevent usage of Node.js specific modules (fs, path, crypto, etc.)
- Updated error messages to guide developers towards environment-agnostic alternatives. (alternatives can be mentioned later on)
- Created a separate ESLint config for the Solidity package
- Excluded test files and AWS-related code from these restrictions

### Drive-by changes

- Added ESLint dependency to solidity/package.json
- Removed redundant eslint-disable comments from test files
- Updated lint script in solidity/package.json to use the new config

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5277

### Backward compatibility

Yes. These changes only affect development-time linting and do not impact runtime behavior.

### Testing

Manual testing:
- Verified ESLint rules correctly flag Node.js built-in module imports
- Confirmed test files in excluded directories are not affected
- Validated lint commands work as expected in both root and solidity packages


typescript/sdk/src/ism/metadata/aggregation.test.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-restricted-imports */
```
nambrot, line 1: Can we have the rule not apply to test files (which are not bundled I have to assume?)

mshojaei-txfusion: Yes I already did this, that is why I removed eslint disable from test files

solidity/eslint.config.mjs
```diff
@@ -0,0 +1,13 @@
+import MonorepoDefaults from '../eslint.config.mjs';
+
+export default [
+  ...MonorepoDefaults,
+  {
+    ignores: [
+      '**/test/**/*',
```
ljankovic-txfusion, line 7: Why are you ignoring "**/test/..." why not './test/**/*'?

ljankovic-txfusion (commented):
I'd rather put these changes in a 1# ZKSync PR, because there we introduce .ts in solidity

solidity/eslint.config.mjs
mshojaei-txfusion: fixed, thanks

typescript/sdk/src/ism/metadata/aggregation.test.ts
nambrot: Ah yes ofc sorry read it wrong

nambrot (approved):

PR #5299, opened 2025-01-27
feat: deploy Sonic SVM

tkporter:
### Description

- Deploys Sonic SVM core
- Adds SOL/solanamainnet-sonicsvm warp route id
- Deploys agents with Sonic
- Adds a script to help with program building


.registryrc
```diff
@@ -1 +1 @@
-6535b3fd17b752834c4a258f7ff9fed0d32590ef
+308dea9125e680f26ae2f994a243bdc0320c48cb
```
paulbalaji, line 1: yarn test will resolve itself once this gets updated post-merge

rust/sealevel/programs/build-programs.sh
```diff
@@ -0,0 +1,93 @@
...
+CORE_PROGRAM_PATHS=("mailbox" "ism/multisig-ism-message-id" "validator-announce" "hyperlane-sealevel-igp")
+TOKEN_PROGRAM_PATHS=("hyperlane-sealevel-token" "hyperlane-sealevel-token-collateral" "hyperlane-sealevel-token-native")
+
+build_program () {
+    PROGRAM_PATH=$1
+    PREV_DIR=$(pwd)
+    log "Building $PROGRAM_PATH"
+    cd $PROGRAM_PATH
+    cargo build-sbf
+    cd $PREV_DIR
```
paulbalaji, line 24: i like to use pushd/popd for this type of thing - should be available on `bash` but not on the basic `sh`
```suggestion
    PROGRAM_PATH=$1
    log "Building $PROGRAM_PATH"
    pushd $PROGRAM_PATH
    cargo build-sbf
    popd
```

rust/sealevel/programs/build-programs.sh
```diff
@@ -0,0 +1,93 @@
...
+    NEW_VERSION=$1
+
+    sh -c "$(curl -sSfL https://release.solana.com/v$NEW_VERSION/install)"
+}
+
+log () {
+    echo "#### $@"
+}
+
+main () {
```
paulbalaji, line 73: maybe worth adding a cleanup helper and set a trap to ensure the cli version gets sack back to the original value at the end

```bash
cleanup () {
    # Only reset if we changed the version in the first place
    if [ "$SOLANA_CLI_VERSION_AT_START" != "$SOLANA_CLI_VERSION_FOR_BUILDING_PROGRAMS" ] && \
       [ ! -z "$SOLANA_CLI_VERSION_AT_START" ]; then
        log "Resetting Solana CLI version back to $SOLANA_CLI_VERSION_AT_START..."
        set_solana_cli_version $SOLANA_CLI_VERSION_AT_START
    fi
}
```

and before line 78 we have
```bash
trap cleanup EXIT
```

tkporter: good call, done

paulbalaji (approved):

PR #5300, opened 2025-01-27
feat: Improve usability of Token.FromChainMetadataNativeToken

jmrossy:
### Description

Improve Token.FromChainMetadataNativeToken by having it fallback to default `nativeToken` metadata if it hasn't been set in the `chainMetadata`. This is useful because `nativeToken` is optional and it's relatively safe because pretty much only the `decimals` value is important in cases where this is used.

### DIFF
--- a/typescript/sdk/src/token/nativeTokenMetadata.ts
+++ b/typescript/sdk/src/token/nativeTokenMetadata.ts
@@ -1,0 +1,25 @@
+import { ProtocolType } from '@hyperlane-xyz/utils';
+
+import { NativeToken } from '../metadata/chainMetadataTypes.js';
+
+export const PROTOCOL_TO_DEFAULT_NATIVE_TOKEN: Record<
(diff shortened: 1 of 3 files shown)


typescript/sdk/src/token/nativeTokenMetadata.ts
```diff
@@ -0,0 +1,25 @@
+import { ProtocolType } from '@hyperlane-xyz/utils';
+
+import { NativeToken } from '../metadata/chainMetadataTypes.js';
+
+export const PROTOCOL_TO_DEFAULT_NATIVE_TOKEN: Record<
+  ProtocolType,
+  NativeToken
+> = {
+  [ProtocolType.Ethereum]: {
```
jmrossy, line 9: Note: I considered using the data from the registry here instead of duping the values for ethereum, solana, cosmoshub but that would introduce a package import cycle sdk<->registry

PR #5305, opened 2025-01-27
fix: Update Renzo to use hardcoded tokenPrices

ltyu:
### Description
Update Renzo to use hardcoded tokenPrices because tokenPrices.json get updated which cause the hook configs to be different.

### Backward compatibility
Yes

### DIFF
--- a/typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConfig.ts
+++ b/typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConfig.ts
@@ -311,6 +311,10 @@ export const getRenzoEZETHWarpConfig = async (): Promise<
     new Set(chainsToDeploy),
     new Set(Object.keys(xERC20)),
   );
+  const tokenPriceDiff = symmetricDifference(
+    new Set(chainsToDeploy),
+    new Set(Object.keys(tokenPrices)),
+  );
   if (validatorDiff.size > 0) {
     throw new Error(
       `chainsToDeploy !== validatorConfig, diff is ${Array.from(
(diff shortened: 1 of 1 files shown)


typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConfig.ts
```diff
@@ -36,6 +34,21 @@ const chainsToDeploy = [
 ];
 export const MAX_PROTOCOL_FEE = parseEther('100').toString(); // Changing this will redeploy the PROTOCOL_FEE hook
 
+const tokenPrices: ChainMap<string> = {
```
paulbalaji, line 37: worth having a quick test to ensure every chain in chainsToDeploy has a tokenPrice?

typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConfig.ts
```diff
@@ -36,6 +34,21 @@ const chainsToDeploy = [
...
+  bsc: '673.59',
+  mode: '3157.26',
+  linea: '3157.26',
+  ethereum: '3157.26',
+  fraxtal: '3168.75',
+  zircuit: '3157.26',
+  taiko: '3157.26',
+  sei: '0.354988',
+  swell: '3157.26',
+};
```
Mo-Hussain, line 51: Are the protocol fees updated? If so, how?

ltyu: yes, iiuc tokenPrices.json get updated. Not sure exactly when @paulbalaji

paulbalaji (approved):

typescript/infra/config/environments/mainnet3/warp/configGetters/getRenzoEZETHWarpConfig.ts
ltyu: Yeah they will, which will require us to change this. i think we're okay with this considering we've had a similar issue with decent updating their owners and us having to adjust https://discord.com/channels/935678348330434570/1333540574850187264/1333869976141955143

Mo-Hussain (approved):

PR #5306, opened 2025-01-28
feat: fix exposed private key in logs

kamiyaa:
### Description

Initially I was thinking of just reimplementing Debug for `SealevelInterchainSecurityModule`, but I saw that keypair is used in many different places.
So I think this solution is a bit more permanent. Having a wrapper around Solana's Keypair with a custom Debug function.

### Backward compatibility

yes

### Testing

Added unittest to test debug functionality

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/keypair.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/keypair.rs
@@ -1,0 +1,41 @@
+use solana_sdk::{signature::Keypair, signer::Signer};
+
+/// Wrapper around solana_sdk's Keypair.
+/// This implments a custom Debug so the private keys are
+/// not exposed.
+pub struct SealevelKeypair(pub Keypair);
+
+impl SealevelKeypair {
+    /// create new SealevelKeypair
+    pub fn new(keypair: Keypair) -> Self {
+        Self(keypair)
+    }
+    /// Return the underlying keypair
+    pub fn keypair(&self) -> &Keypair {
+        &self.0
+    }
+}
+
+impl std::fmt::Debug for SealevelKeypair {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.pubkey().to_string())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use solana_sdk::signature::Keypair;
+
+    use super::SealevelKeypair;
+
+    #[test]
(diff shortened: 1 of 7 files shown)


rust/main/chains/hyperlane-sealevel/src/keypair.rs
```diff
@@ -0,0 +1,41 @@
+use solana_sdk::{signature::Keypair, signer::Signer};
+
+/// Wrapper around solana_sdk's Keypair.
+/// This implements a custom Debug so the private keys are
+/// not exposed.
+pub struct SealevelKeypair(pub Keypair);
```
tkporter, line 6: curious how you feel about Deref, iirc it's kinda controversial but it could maybe allow for a more ergonomic use of this type if we do a
```
impl Deref for SealevelKeypair {
  type Target = Keypair;
  
  fn deref(&self) -> &Self::Target {
    &self.keypair
  }
}
```

tkporter (approved):
ty sir

don't feel strongly about my suggestion, whatever you prefer

rust/main/chains/hyperlane-sealevel/src/keypair.rs
kamiyaa: Ah yeah, that works 👍

rust/main/chains/hyperlane-sealevel/src/rpc/client.rs
```diff
@@ -513,7 +513,7 @@ impl SealevelRpcClient {
             Transaction::new_signed_with_payer(
                 &instructions,
                 Some(&payer.pubkey()),
-                &[payer],
+                &[payer.keypair()],
```
tkporter, line 516: wonder if you need to do this .keypair() or not

tkporter (approved):

rust/main/chains/hyperlane-sealevel/src/rpc/client.rs
kamiyaa: I tried without it, but it was giving me errors. So not exactly sure what was happening :eyes:

daniel-savu (approved):
lgtm! only nit could be to alias `Keypair = SealevelKeypair` if you want to minimize the diff

kamiyaa:
> lgtm! only nit could be to alias `Keypair = SealevelKeypair` if you want to minimize the diff

I think keeping it `SealevelKeypair` would be a bit clearer and less confusing when searching through the codebase :+1:

Issue #5258, opened 2025-01-22
Remove mailbox as required property of warp deploy config


nambrot:
Right now, warp deployment config (https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/26fbec8f6be06bb0712b3c8731bb4aa050e4c74c/typescript/sdk/src/token/types.ts#L99) is descending from https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/26fbec8f6be06bb0712b3c8731bb4aa050e4c74c/typescript/sdk/src/router/types.ts#L67 which requires the `mailbox` property. However, when consuming or reviewing warp deploy config, the mailbox property is usually useless, it should generally be fetched from the configured registry.

Thus this ticket proposes to have a dedicated type for `Omit<HypTokenRouterConfig, 'mailbox'>` that is the generally persisted schema of warp deploy config and when passed to things like deployers, the tooling hydrates the mailbox from the registry.

PR #5312, opened 2025-01-28
refactor: remove mailbox from warp config

mshojaei-txfusion:
### Description
Refactors Warp Route configuration to make mailbox addresses optional during initial config creation. This change separates the mailbox configuration from the base token configuration, introducing new types `WarpRouteDeployConfigWithoutMailbox` and `HypTokenRouterConfigWithoutMailbox`. This improves the configuration workflow by allowing more flexible mailbox address handling.

### Drive-by changes
- Extracted common refinement and transformation logic into separate functions for better code organization
- Improved type safety in configuration validation functions
- Removed unused `isAddress` import

### Related issues
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5237
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5258

### Backward compatibility
Yes - This change maintains backward compatibility while providing more flexibility in configuration

### DIFF
--- a/typescript/sdk/src/token/types.ts
+++ b/typescript/sdk/src/token/types.ts
@@ -1,7 +1,5 @@
 import { z } from 'zod';
 
-import { objMap } from '@hyperlane-xyz/utils';
-
 import { GasRouterConfigSchema } from '../router/types.js';
 import { isCompliant } from '../utils/schemas.js';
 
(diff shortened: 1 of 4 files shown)


typescript/sdk/src/token/types.ts
```diff
@@ -101,58 +101,86 @@ export const HypTokenRouterConfigSchema = HypTokenConfigSchema.and(
...
+    message: WarpRouteDeployConfigSchemaErrors.ONLY_SYNTHETIC_REBASE,
   });
+
+  return z.NEVER; // Causes schema validation to throw with above issue
+}
+
+export const WarpRouteDeployConfigSchema = z
+  .record(HypTokenRouterConfigSchema)
+  .refine(refineTokens, WarpRouteDeployConfigSchemaErrors.NO_SYNTHETIC_ONLY)
+  .transform(transformRebaseConfig);
```
yorhodes, line 177: generally prefer not to include drive-bys like this except because it makes what would otherwise be a very easy to review PR much harder to review
IIUC we have better test coverage on this now so relying on that exercising this

typescript/sdk/src/token/types.ts
```diff
@@ -101,58 +101,86 @@ export const HypTokenRouterConfigSchema = HypTokenConfigSchema.and(
...
-          config.collateralChainName = collateralChainName;
-        return config;
-      }) as Record<string, HypTokenRouterConfig>;
-    }
-
-    ctx.addIssue({
-      code: z.ZodIssueCode.custom,
-      message: WarpRouteDeployConfigSchemaErrors.ONLY_SYNTHETIC_REBASE,
+export const HypTokenRouterConfigWithoutMailboxSchema =
+  HypTokenConfigSchema.and(GasRouterConfigSchema.omit({ mailbox: true }));
```
yorhodes, line 105: I think zod will not error on extraneous fields so wondering why not just remove the mailbox field from the existing schema?

yorhodes (approved):

typescript/sdk/src/token/types.ts
mshojaei-txfusion: I refactored this section since both WarpRouteDeployConfigSchema and WarpRouteDeployConfigSchemaWithoutMailbox needed to share the same refine and transform functions.

typescript/sdk/src/token/types.ts
mshojaei-txfusion: I am not sure I understood correctly but it is not possible do remove mailbox from `MailBoxClienSchema` as it is required for other parts of code, also omit is not supported for `HypTokenRouterConfigMailboxSchema`

nambrot: Maybe this is on me for not having specified it better, but I want mailbox to be optional. I dont think it should be removed, just made optional.

typescript/sdk/src/token/types.ts
nambrot: I had somewhat the same reaction to @yorhodes. Maybe related to my other comment, but I was expecting/hoping that we would just modify WarpRouteDeployConfigSchema to make `mailbox` optional and not introduce another type. But it sounds like @mshojaei-txfusion that breaks somewhere else?

typescript/sdk/src/token/types.ts
mshojaei-txfusion: But on ticket you mentioned remove from deploy config, so you meant make it optional on deployment config and required for the other config?

typescript/sdk/src/token/types.ts
mshojaei-txfusion: This make sense, let me try it.

mshojaei-txfusion: The problem with this way is that these types are nested and at the end we would again need one with required mailbox, and the other one optional(or without) mailbox.
As it is required for deployment at the end we should have it as required parameter on `TokenDeployer`.

Anyway you can see incomplete version of this idea [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/compare/refactor/optional-mailbox-on-warp-config?expand=1) that hardcoded mailbox to just show how it would be in that case.

typescript/sdk/src/token/types.ts
nambrot: What I was trying to say is that it should be removed from the actual files. But on the type, I want it to be an optional field so that people can still specify it if they want to

typescript/sdk/src/token/types.ts
nambrot: IIUC, `HypTokenRouterConfig` is the config that does require `mailbox` and `WarpRouteDeployConfig` could have an optional mailbox, no?

mshojaei-txfusion: So our conclusion is having optional mailbox, and having two types for WarpRouteDeployConfig that one of them makes mailbox required

mshojaei-txfusion:
Closed this as https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5500 would resolve same issue with requested changes.

PR #5334, opened 2025-01-29
feat: Add new Sonic tokens, some SVM tooling drive-bys

tkporter:
### Description

Drive-bys for the SVM tooling:
- make sure that metadata for an SVM synthetic matches some basic expectations, e.g.
  - points to a legit machine readable URL (we had an issue here once)
  - image is a valid URL that's queryable
  - symbol / name match the token config (we had an issue here once)
- No longer require explicitly specifying if a collateral token is SPL Token or SPL Token 2022. We imply this instead by looking at the owner of the collateral mint. Removes this from all previous configs

And adds the following tokens between Solana and Sonic: sSOL, USDStar, USDC, USDT, SONIC


paulbalaji (approved):
lgtm but would also appreciate a second review from @daniel-savu on the rust improvments

rust/sealevel/client/src/warp_route.rs
```diff
@@ -39,6 +42,45 @@ use crate::{
...
+        assert!(
+            self.description.is_some(),
+            "Description must be provided for token with name: {}",
+            self.name
+        );
+        assert!(
+            self.image.is_some(),
+            "Image must be provided for token with name: {}",
+            self.name
+        );
```
daniel-savu, line 73: if these are mandatory, should their types have the `Option`s removed?

rust/sealevel/client/src/warp_route.rs
```diff
@@ -408,6 +437,35 @@ impl RouterDeployer<TokenConfig> for WarpRouteDeployer {
...
+                let metadata_uri = synthetic.uri.as_ref().expect("URI not provided");
+                println!("Validating metadata URI: {}", metadata_uri);
+                let metadata_response = reqwest::blocking::get(metadata_uri).unwrap();
+                let metadata_contents: SplTokenOffchainMetadata = metadata_response
+                    .json()
+                    .expect("Failed to parse metadata JSON");
+                metadata_contents.validate();
+
+                // Ensure that the metadata contents match the provided token config.
+                assert_eq!(metadata_contents.name, synthetic.name, "Name mismatch");
```
daniel-savu, line 460: amazing

daniel-savu (approved):

.registryrc
```diff
@@ -1 +1 @@
-26e513e8fe567428cb58e7634e0ed8c659430707
+3b7433b52d2f520e6b104f98d3da04f26b310a1f
```
paulbalaji, line 1: shouldn't we merge https://github.com/hyperlane-xyz/hyperlane-registry/pull/593 before merging this pr

PR #5335, opened 2025-01-29
feat(infra): Add script to check warp route pod status

### LINKED ISSUE #5247: Script for monitor status and grafana dashboard

Xaroz:
### Description

- Add script to check warp route monitor status by getting the pod using a given `warpRouteId`
- The script will also return logs equal to the amount determined by the `LOG_AMOUNT` constant 
- Logs a link to the grafana dashboard with the given `warpRouteId` 
- Updated `withWarpRouteId` function to enforce choices for `WarpRouteIds`

### Drive-by changes

No

### Related issues

fixes #5247 

### Backward compatibility

Yes
### Testing

Manual

image


typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,83 @@
...
+} from '@hyperlane-xyz/utils';
+
+import { WarpRouteIds } from '../../../config/environments/mainnet3/warp/warpIds.js';
+import { getArgs, withWarpRouteIdRequired } from '../../agent-utils.js';
+
+const orange = chalk.hex('#FFA500');
+const grafanaLink =
+  'https://abacusworks.grafana.net/d/ddz6ma94rnzswc/warp-routes?orgId=1&var-warp_route_id=';
+const logAmount = 5;
+const podPrefix = 'hyperlane-warp-route';
```
ltyu, line 18: Prefer these to be ALL_CAP_UNDERSCORE to designate consts, e.g. PROD_PREFIX

.changeset/small-oranges-dance.md
paulbalaji, line 1: don't need changesets for infra :D

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,83 @@
...
+
+async function main() {
+  configureRootLogger(LogFormat.Pretty, LogLevel.Info);
+  const { environment, warpRouteId } = await withWarpRouteIdRequired(
+    getArgs(),
+  ).parse();
+
```
paulbalaji, line 25: you'll also want to assert we're in the correct k8s context for the given env

```ts
const config = getEnvironmentConfig(environment);
await assertCorrectKubeContext(config);
```

paulbalaji (commented):
mostly lgtm, always a fan of `chalk` usage 👍

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,83 @@
...
+  configureRootLogger,
+  rootLogger,
+} from '@hyperlane-xyz/utils';
+
+import { WarpRouteIds } from '../../../config/environments/mainnet3/warp/warpIds.js';
+import { getArgs, withWarpRouteIdRequired } from '../../agent-utils.js';
+
+const orange = chalk.hex('#FFA500');
+const grafanaLink =
+  'https://abacusworks.grafana.net/d/ddz6ma94rnzswc/warp-routes?orgId=1&var-warp_route_id=';
```
ltyu, line 16: is it always the same url for other environments too (testnet, etc?)

ltyu (commented):
LGTM

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,83 @@
...
+    rootLogger.info(chalk.green(pod));
+
+    rootLogger.info(chalk.gray.italic(`Fetching latest logs...`));
+    const latestLogs = runKubernetesWarpRouteCommand(
+      'logs',
+      podWarpRouteId,
+      environment,
+      [`--tail=${logAmount}`],
+    );
+    rootLogger.info(latestLogs);
```
paulbalaji, line 50: something that would nice to have here is the parsing of the structured output from `kubectl logs` and then reprinting the output in a nicer format here

typescript/infra/scripts/warp-routes/monitor/status.ts
Xaroz: afaik it is!

typescript/infra/scripts/warp-routes/monitor/status.ts
Xaroz: updated the way logs print, please check the attached screenshot

ltyu (approved):

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,130 @@
...
+  configureRootLogger(LogFormat.Pretty, LogLevel.Info);
+  const { environment, warpRouteId } = await withWarpRouteIdRequired(
+    getArgs(),
+  ).parse();
+
+  const config = getEnvironmentConfig(environment);
+  await assertCorrectKubeContext(config);
+
+  try {
+    if (!Object.values(WarpRouteIds).includes(warpRouteId as WarpRouteIds)) {
```
Mo-Hussain, line 35: I'm surprised that we hadn't done this initially but I'd suggest enforcing the choices like this when defining `withWarpRouteId` 

```
export function withWarpRouteId<T>(args: Argv<T>) {
  return args
    .describe('warpRouteId', 'warp route id')
    .choices('warpRouteId', Object.values(WarpRouteIds));
}
```

That way we don't need this validation

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,130 @@
...
+      throw new Error(
+        `Invalid warpRouteId: ${warpRouteId}. Must be one of the defined WarpRouteIds.`,
+      );
+    }
+
+    const podWarpRouteId = getPodWarpRouteId(warpRouteId);
```
Mo-Hussain, line 41: Can we use this static method please https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0bae20d07c4a4e19fc10a5469424b4637c71dade/typescript/infra/src/warp/helm.ts#L52

Mo-Hussain, line 41: Would suggest we reuse this static method https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0bae20d07c4a4e19fc10a5469424b4637c71dade/typescript/infra/src/warp/helm.ts#L52

Xaroz: Updated to use this function and append the "-0" at the end, not sure if we want a monitor number as an argument?

typescript/infra/scripts/warp-routes/monitor/status.ts
```diff
@@ -0,0 +1,121 @@
...
+    rootLogger.info(
+      orange.bold(`Grafana Dashboard Link: ${GRAFANA_LINK}${warpRouteId}`),
+    );
+  } catch (error) {
+    rootLogger.error(error);
+    process.exit(1);
+  }
+}
+
+function runKubernetesWarpRouteCommand(
```
Mo-Hussain, line 64: We should consider adding this as a static method to the `HelmManager` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0bae20d07c4a4e19fc10a5469424b4637c71dade/typescript/infra/src/utils/helm.ts#L118

Mo-Hussain (approved):

paulbalaji (approved):

PR #5340, opened 2025-01-30
fix: filter undefined factory addresses

ljankovic-txfusion:
### Description

When deploying Hyperlane contracts using `warp deploy`, the core addresses obtained using older version of `core deploy` may not include addresses of newer factories (`staticMessageIdWeightedMultisigIsmFactory` and `staticMessageIdWeightedMultisigIsmFactory`). These `undefined` addresses would cause warp deploy to fail, even though they aren't absolutely necessary for warping. 

This fix handles undefined factory addresses that are filtered out during `warp deploy`. This way, warp deploy can proceed the weighted factories, and users won't have to pay gas fees for core deployment once again to.

Attached LogX chain core deployment lacking weighted factory addresses & the resulting error message from `warp deploy`

### Drive-by changes

- Added `undefined` address filtering in the `filterAddressesMap` function

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/contracts/contracts.ts
+++ b/typescript/sdk/src/contracts/contracts.ts
@@ -58,9 +58,12 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
   factories: F,
 ): HyperlaneAddressesMap<F> {
   const factoryKeys = Object.keys(factories);
-  // Filter out addresses that we do not have factories for
+  // Filter out addresses that we do not have factories for and remove undefined values
   const pickedAddressesMap = objMap(addressesMap, (_, addresses) =>
-    pick(addresses, factoryKeys),
+    objFilter(
+      pick(addresses, factoryKeys),
+      (_, value): value is Address => value !== undefined,
+    ),
   );
   // Filter out chains for which we do not have a complete set of addresses
   return objFilter(
--- a/.changeset/late-fishes-end.md
+++ b/.changeset/late-fishes-end.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': patch
+---
+
+Fix contract address filtering to remove undefined factory addresses from the addresses map


paulbalaji (approved):

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -58,9 +58,12 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
   factories: F,
 ): HyperlaneAddressesMap<F> {
   const factoryKeys = Object.keys(factories);
-  // Filter out addresses that we do not have factories for
+  // Filter out addresses that we do not have factories for and remove undefined values
   const pickedAddressesMap = objMap(addressesMap, (_, addresses) =>
-    pick(addresses, factoryKeys),
+    objFilter(
+      pick(addresses, factoryKeys),
+      (_, value): value is Address => value !== undefined,
```
ltyu, line 65: I think this addition is okay but wondering if the change makes more sense in attachContracts(), like it should never attempt to attach if address is undefined.

ltyu: nvm, attachContracts is an objMap so you'd end up with some undefined keys

ltyu (approved):

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -58,9 +58,12 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
   factories: F,
 ): HyperlaneAddressesMap<F> {
   const factoryKeys = Object.keys(factories);
-  // Filter out addresses that we do not have factories for
+  // Filter out addresses that we do not have factories for and remove undefined values
   const pickedAddressesMap = objMap(addressesMap, (_, addresses) =>
-    pick(addresses, factoryKeys),
+    objFilter(
```
nambrot, line 63: This seems like a bandaid solution to me. If folks want to deploy a warp route leveraging those ISMs, they will not realize this until at runtime referencing the factory will not work. Also IIUC, the type will now be incorrect? as the factories do not actually include the value?

It seems to me that we should either make the factory type account for optional, and then have cosumers of the factories do an explicit null check. Or provide the factory addresses as `0x0`?

typescript/sdk/src/contracts/contracts.ts
ljankovic-txfusion: I mean, we already got that error with LogX chain, it was throwing an error because 'ENS or address' were of value `undefined` when attaching to a factory. This solution is technically correct but the proper solution would be to educate the user on how they can resolve this themselves

typescript/sdk/src/contracts/contracts.ts
ljankovic-txfusion: I think that the current state makes `warp deploy` backwards incompatible with `core deploy`, and we should think how to make users' already deployed core contracts compatible with the current warp deploy? Just image a user updating hyperlane cli (to minor ver), only to find they can’t deploy new warp route because their core contracts do not contain some factories they may not ever use.   

@paulbalaji The complete solution would be to implement something like `hyperlane core deploy —fix` to deploy missing contracts without redeploying other contracts (especially the mailbox).  
In `warp deploy` that would entail changes to: 
	a. Check if warp deploy REQUIRES a particular core contract -> error and educate the user to run `core deploy —fix` to deploy missing contracts
	b. Check If warp deploy DOES NOT require a particular core contract -> warn without throwing and error and educate the user run `core deploy —fix`   
	
Perhaps this whole solution would be broken down into several tasks.
	
Within the scope of this particular PR, I would just:
  1. warn the user that they are missing a factory (‘and optionally tell them to do core deploy again’) instead of throwing an error
  2. and instead of filtering out undefined addresses, I would assign them address 0x0, to make types compatible

paulbalaji: @ljankovic-txfusion let's do that instead then 👍

nambrot: Created https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5349

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -62,9 +63,33 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
...
+  let chainWithMissingAddresses = '';
+  const filledAddressesMap = objMap(
+    pickedAddressesMap,
+    (chainName, addresses) =>
+      objMap(addresses, (key, value) => {
+        if (value === undefined) {
+          rootLogger.warn(
+            `Missing address for contract "${key}" on chain ${chainName}`,
+          );
+          chainWithMissingAddresses = chainName;
```
paulbalaji, line 76:
```suggestion
          chainsWithMissingAddresses.push(chainName);
```

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -62,9 +63,33 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
...
+          return constants.AddressZero;
+        }
+        return value;
+      }),
+  );
+  // Add summary warning if any addresses were missing
+  if (chainWithMissingAddresses) {
+    rootLogger.warn(
+      `\nWarning: Core deployment incomplete for chain: ${chainWithMissingAddresses}. ` +
+        `Please run 'core deploy' again for this chain to fix the deployment.\n`,
```
paulbalaji, line 86: I think we remove the `\n` at start and end

also replace `chainWithMissingAddresses` with a joined comma separated list of `chainsWithMissingAddresses`

typescript/sdk/src/contracts/contracts.ts
```diff
@@ -62,9 +63,33 @@ export function filterAddressesMap<F extends HyperlaneFactories>(
   const pickedAddressesMap = objMap(addressesMap, (_, addresses) =>
     pick(addresses, factoryKeys),
   );
+
+  let chainWithMissingAddresses = '';
```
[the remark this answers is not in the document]
ljankovic-txfusion: Used `Set`, but when I was testing I noticed that only one chain is filtered at the time, so I made it string

paulbalaji (approved):

Issue #5317, opened 2025-01-28
Warp Route extensions should be idempotent


nambrot:
Warp route extensions are meant to be done idempotently. That means that if you run the same `warp apply` command with the same deploy config and artifact references (WarpCore config), the outcome should be the same, whether or not whether the user actually submitted the transactions of the first `apply` run. Unfortunately, that is currently no true.

Specifically, only upon the first invocation, a `warp apply` command makes it past this if statement

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/228f7c3d16098f217a8cd9663df27265a2ea8561/typescript/cli/src/deploy/warp.ts#L568-L574

After that, [executeDeploy](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/228f7c3d16098f217a8cd9663df27265a2ea8561/typescript/cli/src/deploy/warp.ts#L584) will actually deploy the contracts on the new chains, which will make `extendedChains` to be an empty array.

Only on the first run, we do (rather oddly) a [mutation of the remoteRouters table](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/228f7c3d16098f217a8cd9663df27265a2ea8561/typescript/cli/src/deploy/warp.ts#L803). On the second run, we hit the code path only in [updateExistingWarpRoute](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/228f7c3d16098f217a8cd9663df27265a2ea8561/typescript/cli/src/deploy/warp.ts#L612). Since the original config does not include the router table, it will not create the enrollment transactions in 

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/26fbec8f6be06bb0712b3c8731bb4aa050e4c74c/typescript/sdk/src/token/EvmERC20WarpModule.ts#L146-L148

The objective of this ticket is to streamline the `warp apply` router enrollment handling to 
a) Allow `warp apply` to be idempontent. I.e. no matter what pending transactions are, when `warp apply` is run
b) Explicitly test these scenarios of (all without having to specify the router table manually in the warp deploy config
  i. extending a warp route in one shot
  ii. extending a warp route later
  iii. extending a warp route after a router was manually unenrolled
c) standardize the remoteRouters and destinationGas mapping behavior to:
  i. if the passed in config to `warp apply` contains the remoteRouters and destinationGas mapping use that
[...]

PR #5358, opened 2025-02-02
refactor: simplify warp route deployment and update logic

mshojaei-txfusion:
### Description

Fixes idempotency issues with warp route extensions and standardizes router enrollment behavior:

- Ensures warp apply commands produce consistent outcomes regardless of transaction status from previous runs
- Consolidates router enrollment logic into a single place instead of having multiple inference points
- Standardizes remoteRouters and destinationGas mapping behavior:
  - Uses config-provided mappings if present
  - Otherwise defaults to fully connected routers with chain-specific gas values
- Removes redundant enrollRemoteRouters function to reduce complexity

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5317
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5387

- `EvmERC20WarpModule.createEnrollRemoteRoutersUpdateTxs` was not able to detect remoteRouters address change
- destination gas was 0 for new routes when extending an warp route 

### Backward compatibility

Yes - Changes maintain backward compatibility with existing warp route configurations

### Testing

Added E2E tests covering key scenarios:
- One-shot warp route extension
- Recovery after manual router un-enrollment
- Recovery after in-complete extension
- All scenarios tested without manual router table specification


typescript/cli/src/deploy/warp.ts
```diff
@@ -627,17 +703,36 @@ async function updateExistingWarpRoute(
   );
   const transactions: AnnotatedEV5Transaction[] = [];
 
+  // Get all deployed router addresses
```
nambrot, line 706: I generally view these kind of transforms as "config expansion", i.e. really `WarpDeployConfig` is kind of a subset of `HypTokenRouterConfig` and so things like expanding remoteRouters and destinationGas i would just put into a separate function (like `populateDestinationGas` already is. i.e. ideally something like

```typescript
const expandedConfig = crossProduct({ remoteRouters: await getRouters(), destinationGas: await destinationGas()})

File: typescript/cli/src/tests/warp/warp-apply.e2e-test.ts [hunk 2]
```diff
@@ -371,4 +371,114 @@ describe('hyperlane warp apply e2e tests', async function () {
     expect(Object.keys(destinationGas_3)).to.include(chain2Id);
     expect(destinationGas_3[chain2Id]).to.equal('7777');
   });
+
+  it('should recover from manual router unenrollment', async () => {
```
nambrot, line 375: Can you add a bit more context here on what you believe the test is doing? I'm actually not sure I agree with the premise of the test 😅

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
```diff
@@ -371,4 +371,114 @@ describe('hyperlane warp apply e2e tests', async function () {
     expect(Object.keys(destinationGas_3)).to.include(chain2Id);
     expect(destinationGas_3[chain2Id]).to.equal('7777');
   });
+
```
nambrot, line 374: A test I very much would like to add is the one from the issue. I.e. a warp route was extended, but the update txs were not applied, and then running warp apply with the same deploy config and updated warp core config results in the expected outcome

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: make sense

typescript/cli/src/deploy/warp.ts
```diff
@@ -530,32 +529,73 @@ export async function runWarpRouteApply(
...
 
-async function extendWarpRoute(
+async function deployNewWarpRoutes(
   params: WarpApplyParams,
   apiKeys: ChainMap<string>,
   warpDeployConfig: WarpRouteDeployConfig,
   warpCoreConfigByChain: ChainMap<WarpCoreConfig['tokens'][number]>,
-) {
+): Promise<
+  | {
```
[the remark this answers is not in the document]
mshojaei-txfusion: removed return values

typescript/cli/src/deploy/warp.ts
```diff
@@ -530,32 +529,73 @@ export async function runWarpRouteApply(
...
+  );
+
   if (transactions.length == 0)
     return logGreen(`Warp config is the same as target. No updates needed.`);
 
   await submitWarpApplyTransactions(params, groupBy(transactions, 'chainId'));
 }
 
-async function extendWarpRoute(
+async function deployNewWarpRoutes(
```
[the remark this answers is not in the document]
mshojaei-txfusion: as I changed the functionality I renamed it, now that reverted that change no need to rename so reverted

typescript/cli/src/deploy/warp.ts
```diff
@@ -666,6 +761,17 @@ async function updateExistingWarpRoute(
...
         );
+
+        const { remoteRouters, chainDestinationGas } = configureRemoteRouters(
+          multiProvider,
+          chain,
+          deployedRoutersAddresses,
+          destinationGas,
+        );
+
+        config.remoteRouters = remoteRouters;
```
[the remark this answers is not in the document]
mshojaei-txfusion: In that case if you do `warp read` before your `warp apply`, apply wont fix unenrolled routes as read wrote the current state on warp-deployment file.
Is this ok?
so if you want to fix your broken routes you should not have any `remoteRoutes` and `destinationGas` properties for your route.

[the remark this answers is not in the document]
mshojaei-txfusion: or maybe we can define `--full-enrollment` or something like this later on separated PR.

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
mshojaei-txfusion: Yes this makes sense, I already tested it manually, will add e2e test for this. this will require to export extendWarpModule function

typescript/cli/src/deploy/warp.ts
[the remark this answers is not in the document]
nambrot: Clarified this on the ticket now as well, but basically I want this behavior:

c) standardize the remoteRouters and destinationGas mapping behavior to:
  i. if the passed in config to `warp apply` contains the remoteRouters and destinationGas mapping use that
  ii. it eh passed in config does not contain those mappings, assume fully connected routers and use the gas value of each chain to construct the destinationGas mapping
  
  Does that help clarify the intended behavior?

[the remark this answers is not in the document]
nambrot: Also would like to see test cases for this behavior as well

typescript/cli/src/deploy/warp.ts
```diff
@@ -556,57 +556,128 @@ async function extendWarpRoute(
...
     newDeployedContracts,
     warpCoreConfigByChain,
   );
 
+  // Get the updated core config that includes both existing and new contracts
   const { warpCoreConfig: updatedWarpCoreConfig, addWarpRouteOptions } =
     await getWarpCoreConfig(params, mergedRouters);
   WarpCoreConfigSchema.parse(updatedWarpCoreConfig);
 
+  // Update warpDeployConfig with newly deployed contract addresses
```
nambrot, line 610: I dont understand this part?

typescript/cli/src/deploy/warp.ts
```diff
@@ -556,57 +556,128 @@ async function extendWarpRoute(
...
+      chain
+    ] as HyperlaneContracts<HypERC20Factories>;
+    const router = getRouter(contracts);
+
+    if (isCollateralTokenConfig(config)) {
+      config.token = router.address;
+    }
+  }
+
+  // Update warpCoreConfigByChain with new addresses
```
nambrot, line 623: Can we avoid mutation somehow?

typescript/cli/src/deploy/warp.ts
```diff
@@ -627,17 +698,21 @@ async function updateExistingWarpRoute(
...
+  );
+
   await promiseObjAll(
     objMap(warpDeployConfig, async (chain, config) => {
       await retryAsync(async () => {
-        logGray(`Update existing warp route for chain ${chain}`);
-        const deployedConfig = warpCoreConfigByChain[chain];
-        if (!deployedConfig)
+        const deployedTokenRoute = deployedRoutersAddresses[chain];
+        if (!deployedTokenRoute)
```
nambrot, line 711: Should we error out here?

typescript/cli/src/deploy/warp.ts
```diff
@@ -530,20 +529,21 @@ export async function runWarpRouteApply(
...
-      apiKeys,
-      warpDeployConfig,
-      warpCoreConfigByChain,
-    )),
-  ];
+  await extendWarpRoute(
+    params,
+    apiKeys,
+    warpDeployConfig,
+    warpCoreConfigByChain,
```
nambrot, line 536: Let's be super explicit that this is getting mutated in `extendWarpRoute`

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: Other solution was keeping these outside of extend function and having a pure function for extend.
But then again we would need to iterate over them somewhere!

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: what is not clearing we are putting the new deployed routes addresses on  `newDeployedContracts ` and `warpCoreConfigByChain ` so they would including them on write and enrollment time.

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: sorry did not get your question

typescript/cli/src/deploy/warp.ts
nambrot: In general, this is a pretty long function, it might be useful to split it up and have more description function names/comments to make it more digestable?

typescript/cli/src/deploy/warp.ts
nambrot: Can we just create a copy `warpCoreConfigByChain` in here and then return that to the caller, and then the caller can update its reference of `warpCoreConfigByChain` explicitly? That imo makes the control flow of the mutation much more obvious (by making it part of the explicit API of the function)

typescript/cli/src/deploy/warp.ts
nambrot: If somehow deployedTokenRoute is unfined, it sounds like we should loudly error and exit, and not just log, but continue

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
mshojaei-txfusion: remove this one based on last clarifications you made, this was not suppose to be the case

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: Agree, I also dont like how it is RN, let me try to refactor it.

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: Generally agree that this mutation behavior is a bad practice for typescript. will refactor it.

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: Yes wont happen in new implementation, but yes you are right, and good catch.
in case happened should fail. fixed!

ltyu (approved):

typescript/cli/src/deploy/warp.ts
```diff
@@ -568,63 +568,67 @@ function splitWarpApplyConfig(
...
-  const updatedWarpDeployConfig = {
+  return {
+    newDeployedContracts,
+    updatedWarpCoreConfig,
+    addWarpRouteOptions,
+  };
+}
+
+/**
+ * Updates the Warp route configuration by incorporating newly deployed contract addresses.
```
nambrot, line 623: Let's be specific in which configuration, i.e. in this case `Warp Deploy Config`

typescript/cli/src/deploy/warp.ts
```diff
@@ -589,26 +599,180 @@ async function extendWarpRoute(
...
+        const config = { ...warpDeployConfig[chain] };
+        if (!isCollateralTokenConfig(config)) {
+          return [chain, config];
+        }
+
+        const router = getRouter(
+          contracts as HyperlaneContracts<HypERC20Factories>,
+        );
+
+        return [chain, { ...config, token: router.address }];
```
nambrot, line 645: I don't think this is intended. The `token` field in the `WarpDeployConfig` is very different, I think you might be mixing this up with the `WarpCoreConfig`? I would like you add a test case, because it sounds like this is in right now, but not breaking a test (when it should)

typescript/cli/src/deploy/warp.ts
```diff
@@ -589,26 +599,180 @@ async function extendWarpRoute(
...
+  const { newDeployedContracts, updatedWarpCoreConfig, addWarpRouteOptions } =
+    await deployWarpExtensionContracts(
+      params,
+      apiKeys,
+      existingConfigs,
+      initialExtendedConfigs,
+      warpCoreConfigByChain,
+    );
+
+  // Create new warpDeployConfig with updated addresses
```
nambrot, line 698: Why is this necessary?

typescript/cli/src/deploy/warp.ts
```diff
@@ -48,6 +47,7 @@ import {
   WarpRouteDeployConfigSchema,
   attachContractsMap,
   connectContractsMap,
+  gasOverhead,
```
nambrot, line 50: I know this was a large file before your changes, but I'm wondering if we can split it up?

nambrot: Maybe even move some of this into the SDK?

typescript/cli/src/deploy/warp.ts
nambrot: As in i want a test that checks that warp deploy config is unchanged

mshojaei-txfusion: Yes good catch and this was not needed at all, I mixed it while removing `enrollRemoteRouters` function.

typescript/cli/src/deploy/warp.ts
mshojaei-txfusion: it is not at all, my bad

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
```diff
@@ -802,4 +802,39 @@ describe('hyperlane warp apply e2e tests', async function () {
...
+      mailbox: chain2Addresses!.mailbox,
+      name: 'Ether',
+      owner: new Wallet(ANVIL_KEY).address,
+      symbol: 'ETH',
+      totalSupply: 0,
+      type: TokenType.native,
+    };
+
+    warpDeployConfig[CHAIN_NAME_3] = extendedConfig;
+    // Remove remoteRouters and destinationGas as they are written in readWarpConfig
```
nambrot, line 827: Could we not just get the warpDeployConfig without reading it? Would much prefer that to simulate the actual usage and therefore test

typescript/cli/src/deploy/warp.ts
```diff
@@ -530,55 +522,69 @@ export async function runWarpRouteApply(
...
-
   await submitWarpApplyTransactions(params, groupBy(transactions, 'chainId'));
 }
 
-async function extendWarpRoute(
+/**
+ * Splits warp deploy config into existing and extended configurations based on warp core chains
+ * for the warp apply process.
+ */
+function splitWarpApplyConfig(
```
nambrot, line 549: Should this be moved into configUtils as well?

typescript/cli/src/deploy/warp.ts
```diff
@@ -627,17 +687,22 @@ async function updateExistingWarpRoute(
...
-        if (!deployedConfig)
-          return logGray(
-            `Missing artifacts for ${chain}. Probably new deployment. Skipping update...`,
-          );
-
-        const deployedTokenRoute = deployedConfig.addressOrDenom!;
+        const deployedTokenRoute = deployedRoutersAddresses[chain];
+        assert(deployedTokenRoute, `Missing artifacts for ${chain}.`);
+
         const {
```
nambrot, line 706: This is outside the scope for this PR, but we really should be moving this into a constant IMO

nambrot: Created https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5543 and Assigned you @mshojaei-txfusion

typescript/cli/src/tests/commands/helpers.ts
```diff
@@ -230,12 +232,96 @@ export async function extendWarpConfig(params: {
     warpDeployPath,
   );
   warpDeployConfig[chainToExtend] = extendedConfig;
+  // Remove remoteRouters and destinationGas as they are written in readWarpConfig
```
nambrot, line 235: Like mentioned in another comment, I would like to avoid reading the config if we can. If that has to happen in another PR, so be it

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -153,12 +153,16 @@ export class EvmERC20WarpModule extends HyperlaneModule<
     const { remoteRouters: actualRemoteRouters } = actualConfig;
     const { remoteRouters: expectedRemoteRouters } = expectedConfig;
 
-    const routesToEnroll = Array.from(
-      difference(
-        new Set(Object.keys(expectedRemoteRouters)),
-        new Set(Object.keys(actualRemoteRouters)),
-      ),
-    );
+    const routesToEnroll = Object.entries(expectedRemoteRouters)
```
nambrot, line 156: Curious why you had to change this logic?

nambrot (commented):
Love the progress!!! I think this should be the last round

ltyu (approved):

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
```diff
@@ -120,4 +135,60 @@ describe('hyperlane warp apply owner update tests', async function () {
       normalizeConfig(warpDeployConfig[CHAIN_NAME_2].hook),
     );
   });
+
```
xeno097, line 138: Can you add a test that verifies that when changing a remote router to a new address, the route is successfully updated?

typescript/cli/src/deploy/warp.ts
```diff
@@ -544,55 +499,46 @@ export async function runWarpRouteApply(
...
+    warpCoreConfig,
+  );
+
+  // Then create and submit update transactions
+  const transactions: AnnotatedEV5Transaction[] = await updateExistingWarpRoute(
+    params,
+    apiKeys,
+    warpDeployConfig,
+    updatedWarpCoreConfig,
+  );
```
xeno097, line 515: It seems that when enrolling new routers, the addresses are added correctly to the routers but, the destination gas is set to 0 instead of the default value for the type of warp route

```yaml
anvil420:
  decimals: 18
  destinationGas:
    "42069": "0"
    "69": "0"
  ...
```

typescript/cli/src/tests/warp/warp-apply.e2e-test.ts
mshojaei-txfusion: yes good idea as it was not working as expected before

xeno097 (approved):

PR #5361, opened 2025-02-03
feat(balances): Adjust balance and alert threshold automation process

Mo-Hussain:
### Description

- Suggest changing the minimum number of daily txs that need to be supported to 100 from 200. Main motivation for this change is that it reduces the monetary cost of maintaining high dollar value relayer balance for BTC and ETH gas token chains. Table below shows the chains where the 200 tx calculations results in a significant dollar increase in relayer balance needed. We will also need to ~1.5x these amounts in Key funder so it does add up

```
┌────────────────┬─────────────────┬──────────────────────────────┬─────────────────────────────┐
│     chain      │ proposedDailyBurn│ proposedRelayerBallanceDollars│ currentRelayerBallanceDollars│
├────────────────┼─────────────────┼──────────────────────────────┼─────────────────────────────┤
│ bitlayer       │     0.00205     │            1620              │            198              │
│ bsquared       │     0.00358     │            2830              │            198              │
│ ethereum       │     0.269       │            6590              │           1530              │
│ morph          │     0.276       │            6750              │            306              │
│ rootstockmainnet│     0.00259    │            2040              │            197              │
└────────────────┴─────────────────┴──────────────────────────────┴─────────────────────────────┘
```

**I haven't throughly checked all balances to see if they are kosher, I will do this once we have agreement on the 100 to 200 tx change**

- Add `lowUrgencyEngKeyFunderBalance.json` config and add its alert query template

**calculate-relayer-daily-burn**

- use promQL query for sealevel daily burn

**update-balance-threshold-config**

- Support `desiredRelayerBalanceOverrides.json`, allow us to override this threshold for new chains, chains were we need to maintain a temporary low balance and special cases like osmosis. Alert thresholds are also adjusted based on this override. The overrides also allows us to respond to cases where the current thresholds are significantly larger than the proposed values and we want to reset

- Inform user if the proposed thresholds or daily burn estimation is 50% less than the current values, prompt user for manual review
image

If we want to reduce the thresholds, we can add an entry into the `desiredRelayerBalanceOverrides.json` and call `update-balance-threshold-config.ts` the reset all thresholds.

**I haven't done any manual threshold reductions yet, I think this would be good to do in a quick follow up do the changes are more obvious** 

- Add validation which checks if all the thresholds makes sense `lowUrgencyKeyFunderBalance[chain] > desiredRelayerBalances[chain] > lowUrgencyEngKeyFunderBalance[chain] > highUrgencyRelayerBalance[chain]`
image

**write-alert.ts**
- run validation for all configs before pushing changes to grafana

**Next Steps**
- Manually reduce thresholds based on them being more than 50% below and throughly review all thresholds 
- handling ata-payer balances and alerts
- automating IGP payments
- Roll out

### Testing

Manual


typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -1,124 +1,124 @@
...
-  "swell": 0.00454,
-  "taiko": 0.00726,
+  "superseed": 0.0126,
+  "superpositionmainnet": 0.00113,
+  "swell": 0.00238,
+  "taiko": 0.0048,
   "tangle": 3.13,
   "telos": 22.7,
   "torus": 5.82,
-  "treasure": 789,
```
Mo-Hussain, line 108: Not sure why treasure was so high initially, this is the current burn in the 10 look back window

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
```diff
@@ -0,0 +1,15 @@
+{
```
Mo-Hussain, line 1: overrides includes should include new chains, chains with temporarily low until we're able to fund more and special cases likes osmosis

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -1,24 +1,30 @@
...
 const tokenPrices: ChainMap<string> = rawTokenPrices;
 const currentDailyRelayerBurn: ChainMap<number> = rawDailyRelayerBurn;
+const desiredRelayerBalances: ChainMap<string> = rawDesiredRelayerBalances;
 
 const DAILY_BURN_PATH =
   './config/environments/mainnet3/balances/dailyRelayerBurn.json';
 
 const LOOK_BACK_DAYS = 10; // the number of days to look back for average destination tx costs
-const MIN_NUMBER_OF_TXS = 200; // the minimum number of txs to consider for daily burn
+const MIN_NUMBER_OF_TXS = 100; // the minimum number of txs to consider for daily burn
```
Mo-Hussain, line 27: suggested change to 100

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -125,9 +174,44 @@ async function getDailyRelayerBurn(sql: Sql) {
...
+    ) {
+      lowProposedDailyBurn.push({
+        chain,
+        proposedDailyBurn: proposedDailyRelayerBurn,
+        currentDailyBurn: currentDailyRelayerBurn[chain],
+      });
+    }
+
+    // push to burnInfoTable
+    burnInfoTable.push({
```
Mo-Hussain, line 190: show user more info about daily burn, specifically dollar values

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -1,124 +1,124 @@
 {
```
Mo-Hussain, line 1: Majority of the reductions are due to 1) 200 min tx => 100 min tx and 2) desiredRelayerBalanceOverrides where we are temporarily keeping the balance low

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -1,124 +1,124 @@
...
-  "abstract": 0.00291,
```
Mo-Hussain, line 2: Will do a through review of all thresholds and manual reduction based on relayer burn after 200 min tx => 100 min tx changes is agrees upon

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -1,104 +1,160 @@
...
 const dailyBurn: ChainMap<number> = rawDailyBurn;
 
-const exclusionList = ['osmosis'];
-
 async function main() {
-  const { balanceThresholdConfig, all } = await withConfirmAllChoices(
-    withBalanceThresholdConfig(yargs(process.argv.slice(2))),
-  ).argv;
-
-  const configToUpdate: BalanceThresholdType[] = all
```
Mo-Hussain, line 32: Made a change to always update all thresholds at once, this enables us to validate and avoid easily falling in different configs being out of whack. It does sacrifice some flexibility, which has been helpful during development, but it is not expected to be required for the production process

typescript/infra/src/funding/grafana.ts
```diff
@@ -197,11 +197,9 @@ ${config.queryTemplate.footer}
...
   newThresholds: ChainMap<string>,
 ): ChainMap<string> {
-  const orderedThresholds: ChainMap<string> = {};
-  Object.keys(newThresholds)
-    .sort()
-    .forEach((key) => {
-      orderedThresholds[key] = newThresholds[key];
-    });
-  return orderedThresholds;
+  return Object.fromEntries(
```
Mo-Hussain, line 200: Decided to revert to something I consider more reliable lol

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -1,104 +1,168 @@
...
-  withBalanceThresholdConfig,
-  withConfirmAllChoices,
-} from '../agent-utils.js';
 
 const dailyBurn: ChainMap<number> = rawDailyBurn;
 
-const exclusionList = ['osmosis'];
-
 async function main() {
-  const { balanceThresholdConfig, all } = await withConfirmAllChoices(
```
Mo-Hussain, line 28: Will remove these utils, just realised

typescript/infra/src/config/funding/balances.ts
```diff
@@ -22,24 +33,41 @@ export const balanceThresholdConfigMapping: Record<
...
   },
 };
+
+export interface ManualReview {
+  chain: string;
+  proposedThreshold: number;
+  currentThreshold: number;
+}
+
+interface ThresholdConfig {
```
Mo-Hussain, line 68: Could rename this to ThresholdData

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -125,10 +227,74 @@ async function getDailyRelayerBurn(sql: Sql) {
         ? Math.max(proposedDailyRelayerBurn, currentDailyRelayerBurn[chain])
```
Mo-Hussain, line 227: Considering to only update the daily burn if its 10% greater than the current burn, this is to avoid lots of small changes that will make the reviewing more cumbersome. Curious what you think @paulbalaji @tkporter

paulbalaji: 10% sgtm - good to have it as a const somewhere if we want to tweak it

typescript/infra/scripts/funding/write-alert.ts
```diff
@@ -1,42 +1,46 @@
...
+        `${THRESHOLD_CONFIG_PATH}/${balanceThresholdConfigMapping[balanceThresholdType].configFileName}`,
+      ) as ChainMap<string>;
+
+      return {
+        ...acc,
+        [balanceThresholdType]: {
+          thresholds,
+        },
+      };
+    }, {} as ThresholdConfigs);
```
paulbalaji, line 39: since ThresholdConfigs's keys are just the BalanceThresholdType, maybe instead of the reduce logic I see in a few places, we can just do a map instead

easier to follow that we're just mapping from a threshold type to its corresponding values

```
const balanceThresholdConfigs: ThresholdConfigs = Object.fromEntries(
  balanceThresholdTypes.map((type) => [
    type,
    {
      thresholds: readJSONAtPath(
        `${THRESHOLD_CONFIG_PATH}/${balanceThresholdConfigMapping[type].configFileName}`,
      ) as ChainMap<string>,
    },
  ])
);
```

typescript/infra/scripts/funding/update-balance-threshold-config.ts
```diff
@@ -1,104 +1,168 @@
...
+function handleManualReviews(newConfigs: ThresholdConfigs) {
+  for (const configKey of Object.keys(newConfigs) as BalanceThresholdType[]) {
+    const { manualReview } = newConfigs[configKey];
+    if (manualReview && manualReview.length > 0) {
+      rootLogger.info(
+        `The ${configKey} config contains ${manualReview.length} chain(s) where the proposed threshold is less than 50% of the current threshold. Please review the following items:`,
+      );
+      console.table(manualReview);
     }
   }
```
paulbalaji, line 167: should we also wait for a "yes" prompt for the user before continuing? because in the main logic above we log this and then just continue to write

or is it enough to let the user know to review the output files directly?

paulbalaji: from my pov as a user of this, i'd rather say y/n at the script so i don't have to filter through the output itself

paulbalaji (commented):
seems there's a mix of new stuff + refactoring here which makes the diff a little hard to read - would appreciate if there's a way to split up the changes a bit more

typescript/infra/src/funding/balances.ts
```diff
@@ -0,0 +1,91 @@
...
+ * This function logs warnings for chains present in lower-priority configs but missing in higher-priority configs.
+ * It also logs errors if the thresholds are not sized correctly.
+ */
+export function validateThresholds(newConfigs: ThresholdConfigs): void {
+  // Accumulate error and warning messages.
+  const errorMessages: string[] = [];
+  const warningMessages: string[] = [];
+
+  // Sort the config types by their weight (ascending order).
+  // Lower weight means a higher-priority threshold.
```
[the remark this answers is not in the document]
Mo-Hussain: I'm not super opinionated on this, can change the order

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -125,10 +227,74 @@ async function getDailyRelayerBurn(sql: Sql) {
...
+        metadata.protocol == ProtocolType.Sealevel && !metadata.isTestnet,
+    )
+    .reduce((acc: { [key: string]: string }, metadata) => {
+      acc[metadata.name] = metadata.domainId.toString();
+      return acc;
+    }, {});
+  return sealevelDomainIds;
+}
+
+function writeBurnDataToFile(burnData: ChainMap<number>) {
```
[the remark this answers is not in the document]
Mo-Hussain: Just refactored to make the main method easier to read

Mo-Hussain:
> seems there's a mix of new stuff + refactoring here which makes the diff a little hard to read - would appreciate if there's a way to split up the changes a bit more

This is fair, I tried to split up the commits. I did not want to include everything in this PR but decided to add more as we were busy with other things and I was waiting for reviews

typescript/infra/src/funding/balances.ts
```diff
@@ -0,0 +1,91 @@
...
+  const configTypesSorted = Object.keys(newConfigs)
+    .map((key) => key as BalanceThresholdType)
+    .sort(
+      (a, b) =>
+        balanceThresholdConfigMapping[a].weight -
+        balanceThresholdConfigMapping[b].weight,
+    );
+
+  for (let i = 0; i < configTypesSorted.length - 1; i++) {
```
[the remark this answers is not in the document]
Mo-Hussain: Yes

typescript/infra/scripts/funding/update-balance-threshold-config.ts
Mo-Hussain: We are only updating thresholds if the proposed thresholds are greater than the existing ones. These manual reviews are for cases where the proposed balance is 50% less than the current. 

Continuing does nothing to these thresholds, they are left unchanged. We could have an interactive process by which you can choose to reduces the thresholds by y/n

The alternative is that you look at the output, decide which chains you want reduce and add the new desired relayer balance to `desiredRelayerBalanceOverrides` and run the update script again. This will update all the related thresholds at once

tkporter:
Generally think this PR would benefit from reducing the complexity of the edge cases of things - there are a few places where there are meaningful if statements or fallbacks if something is undefined etc that weren't super obvious to me. Making things so there's generally 1 path of how things work and then small conditionals to change inputs to that path would help a lot

typescript/infra/config/environments/mainnet3/balances/dailyRelayerBurn.json
```diff
@@ -1,124 +1,125 @@
...
-  "zircuit": 0.00379,
+  "zeronetwork": 0.00127,
+  "zetachain": 10.1,
+  "zircuit": 0.00424,
   "zklink": 0.00102,
-  "zksync": 0.00284,
-  "zoramainnet": 0.00832
+  "zksync": 0.00159,
+  "zoramainnet": 0.00876,
+  "solanamainnet": 2.95,
```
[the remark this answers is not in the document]
Mo-Hussain: Ah the dailyRelayerBurn was not sorted, will do

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -297,6 +320,56 @@ function writeBurnDataToFile(burnData: ChainMap<number>) {
   }
 }
 
+async function handleLowProposedDailyBurn(
```
Mo-Hussain, line 323: Prompt user to review 
image

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,356 @@
...
+  rootLogger.info('All chain thresholds have been updated successfully.');
+}
+
+async function processChainThresholds(
+  chain: string,
+  chainDailyBurn: number,
+  currentThresholds: Record<BalanceThresholdType, number>,
+  thresholdOverride?: number,
+): Promise<Record<BalanceThresholdType, number>> {
+  const proposedThresholds = buildProposedThresholds(
```
Mo-Hussain, line 81: Opted to split the actions into functions to make it easier to understand, this does mean that we are looping over the thresholds on each function call, however imo this inefficiency to acceptable for the improved readability

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,356 @@
...
+    );
+  }
+
+  finalThresholds = await validateThresholdsInteractive(chain, finalThresholds);
+
+  return finalThresholds;
+}
+
+// threshold building functions
+function buildProposedThresholds(
```
Mo-Hussain, line 113: Happy to add more comments to each of these functions

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,356 @@
...
+        proposed[thresholdType],
+      );
+    }
+  }
+
+  return finalThresholds;
+}
+
+// review and validation functions
+async function handleUserReviews(
```
Mo-Hussain, line 190: interactive review 
image

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,356 @@
...
+    };
+    if (proposedThresholds) {
+      data.Proposed = proposedThresholds[thresholdType as BalanceThresholdType];
+    }
+    return data;
+  });
+  console.table(reviewTable);
+}
+
+async function interactiveThresholdUpdate(
```
Mo-Hussain, line 274: can interactively update all thresholds 
image

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,355 @@
+import { input, select } from '@inquirer/prompts';
```
Mo-Hussain, line 1: Rewrote this

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  writeConfigsToFile(updatedThresholds);
+  rootLogger.info('All chain thresholds have been updated successfully.');
+}
+
+async function processChainThresholds(
+  chain: string,
+  chainDailyBurn: number,
+  currentThresholds: Record<BalanceThresholdType, number>,
+  thresholdOverride?: number,
```
tkporter, line 79: nit: maybe `desiredRelayerBalanceOverride`?

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  finalThresholds = await validateThresholdsInteractive(chain, finalThresholds);
+
+  return finalThresholds;
+}
+
+// threshold building functions
+function buildProposedThresholds(
+  chainDailyBurn: number,
+  thresholdOverride?: number,
```
tkporter, line 115: ditto would suggest `desiredRelayerBalanceOverride `

typescript/infra/scripts/funding/update-balance-thresholds.ts
```diff
@@ -0,0 +1,358 @@
...
+      },
+    ],
+  });
+
+  switch (selectedChoice) {
+    case UserReview.AcceptAllProposed:
+      return proposedThresholds;
+    case UserReview.KeepAllCurrent:
+      rootLogger.warn(
+        `Consider add ${chain} to the desiredRelayerBalanceOverrides.json file, so that you are not prompted to review the thresholds in the future.`,
```
tkporter, line 220:
```suggestion
        `Consider adding ${chain} to the desiredRelayerBalanceOverrides.json file, so that you are not prompted to review the thresholds in the future.`,
```
nit

typescript/infra/src/config/funding/alert-query-templates.ts
```diff
@@ -32,7 +40,40 @@ export const LOW_URGENCY_KEY_FUNDER_FOOTER = `    # Mainnets that don't use key-
     (last_over_time(hyperlane_wallet_balance{wallet_name="relayer", chain="stride"}[1d])) - 10
 `;
 
-// TODO: add footer for LOW_URGENCY_ENG_KEY_FUNDER
+export const LOW_URGENCY_ENG_KEY_FUNDER_FOOTER = `    # Mainnets that don't use key-funder and all funds are stored in the relayer's wallet
+    # Eclipse
+    last_over_time(hyperlane_wallet_balance{wallet_name="relayer", hyperlane_context="hyperlane", chain=~"eclipsemainnet"}[1d]) - 0.5 or
```
tkporter, line 45: we have the ability to have these relayer balance alerts be dynamic using the `dailyRelayerBurn.json` / `desiredRelayerBalances.json` right? anyways I assume this can / should come later

typescript/infra/src/funding/balances.ts
```diff
@@ -0,0 +1,126 @@
...
+    const chainMap = readJSONAtPath(thresholdsFile) as ChainMap<number>;
+
+    result[thresholdType] = chainMap;
+  }
+
+  return result;
+}
+
+export function sortThresholdTypes(
+  thesholdTypes: BalanceThresholdType[],
```
tkporter, line 118: nit: typo `thesholdTypes`

tkporter (commented):
code generally all lgtm, thanks for the changes! much easier to follow now

I ran some of the scripts locally and get the impression that the values maybe need updating? I'd consider the final part now to be just confirming that all the checked in values look right, which I haven't looked at yet

In case it's important, I get the impression the values need to be updated by a script run bc:
- some `dailyRelayerBurn.json` values in the current commit aren't sorted in the way that running `calculate-relayer-daily-burn.ts` will apply them (e.g. ancient8 should be after alephzeromainnet)
- update-balance-thresholds.ts prompted me with like 20 chains as involving a 50% reduction but didn't have significant changes to the dailyRelayerBurn

Mo-Hussain:
> code generally all lgtm, thanks for the changes! much easier to follow now
> 
> I ran some of the scripts locally and get the impression that the values maybe need updating? I'd consider the final part now to be just confirming that all the checked in values look right, which I haven't looked at yet
> 
> In case it's important, I get the impression the values need to be updated by a script run bc:
> 
> * some `dailyRelayerBurn.json` values in the current commit aren't sorted in the way that running `calculate-relayer-daily-burn.ts` will apply them (e.g. ancient8 should be after alephzeromainnet)
> * update-balance-thresholds.ts prompted me with like 20 chains as involving a 50% reduction but didn't have significant changes to the dailyRelayerBurn

@tkporter - you are right about this, I'm going to add another commit that makes those reductions for those ~20 chains and get a review from you and @paulbalaji if he has time to review. For any chains where we don't want to make the reduction, we can add those to the override so we are not prompted to review again in future

tkporter:
I'm a little wary of using the override hammer too much - with it resulting in us not being prompted again it feels like an easy path to us forgetting that we overrode something and then we end up never realizing we need higher balances in the future?

Mo-Hussain:
That is certainly a risk worth considering, we can a take the posture of only adding chain to the override if we are repeatedly being prompted to manually review.

We could also add a warning for cases where the override is leading to us under funding the relayer by X percent?

typescript/infra/scripts/funding/calculate-relayer-daily-burn.ts
```diff
@@ -125,10 +227,74 @@ async function getDailyRelayerBurn(sql: Sql) {
...
+async function getSealeveDomainIds(): Promise<ChainMap<string>> {
+  const registry = await getRegistry();
+  const allMetadata = await registry.getMetadata();
+  const sealevelDomainIds: ChainMap<string> = Object.values(allMetadata)
+    .filter(
+      (metadata) =>
+        metadata.protocol == ProtocolType.Sealevel && !metadata.isTestnet,
+    )
+    .reduce((acc: { [key: string]: string }, metadata) => {
+      acc[metadata.name] = metadata.domainId.toString();
```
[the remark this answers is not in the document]
Mo-Hussain: I had it initially because the keyfunder config expects it to be a string, I've changed it to numbers now

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -1,98 +1,98 @@
 {
   "abstract": 0.0342,
   "alephzeroevmmainnet": 135,
-  "ancient8": 0.5,
+  "ancient8": 0.0151,
```
Mo-Hussain, line 4: Relayer burn is less

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -1,98 +1,98 @@
 {
```
Mo-Hussain, line 1: For all of these chains, I checked the relayer burn using our dashboard and ensured the the threshold was greater than burn * multiplier.

https://abacusworks.grafana.net/d/ce73gakt6rlkwf/relayer-and-key-funder-balances?orgId=1&from=now-7d&to=now&timezone=browser&var-chain=solanamainnet

the dashboard show us 2 burn values, the 24 hour burn and the 10 day average 24 houes burn. I used the max of those values.

For many of these chains the threshold is dictated by either the $25 or 100 tx requirement and the burn is significantly less

Mo-Hussain: I've reverted this and only reduced thresholds for 5 chains

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
```diff
@@ -1,15 +1,4 @@
...
-  "xai": 20,
-  "xlayer": 0.5,
-  "xpla": 100,
-  "zeronetwork": 0.05,
-  "zetachain": 20,
-  "zircuit": 0.02,
-  "zklink": 0.05,
-  "zksync": 0.05,
-  "zoramainnet": 0.2
+  "vana": 0.001
```
Mo-Hussain, line 3: Only including vana as I believe this is the only chain where we are keeping a temporary low balance

tkporter (approved):

PR #5366, opened 2025-02-04
Boilerplate code for app context operation verifier

ameten:


rust/sealevel/programs/hyperlane-sealevel-token-collateral/Cargo.toml
```diff
@@ -30,6 +30,7 @@ hyperlane-sealevel-igp = { path = "../hyperlane-sealevel-igp", features = [
 ] }
 hyperlane-sealevel-message-recipient-interface = { path = "../../libraries/message-recipient-interface" }
 hyperlane-sealevel-token-lib = { path = "../../libraries/hyperlane-sealevel-token" }
+hyperlane-warp-route = { path = "../../../main/applications/hyperlane-warp-route" }
```
daniel-savu, line 33: i think we want to completely avoid depending on the `main` workspace so we can eventually decouple the two workspaces and fix dependency issues

you could create a new crate that has no dependencies on the two workspaces, but is depended on by both and defines `TokenMessage`

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -291,7 +295,10 @@ impl PendingOperation for PendingMessage {
         {
             Ok(tx_cost_estimate) => tx_cost_estimate,
             Err(err) => {
-                return self.on_reprepare(Some(err), ReprepareReason::ErrorEstimatingGas);
+                let reason = self
+                    .clarify_reason(ReprepareReason::ErrorEstimatingGas)
+                    .await;
+                return self.on_reprepare(Some(err), reason);
```
daniel-savu, line 301: I think any time we call `on_reprepare` on this file we should consider parsing with the op-verifier, including in the submit step

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -664,6 +674,26 @@ impl PendingMessage {
...
+            .verify(&self.app_context, &self.message)
+            .await
+        {
+            Ok(()) => reason,
+            Err(e) => match e {
+                InsufficientAmountError | MalformedMessageError(_) => {
+                    ApplicationOperationVerificationFailed
+                }
+                ChainCommunicationError(_) | UnknownApplicationError(_) => reason,
+            },
```
daniel-savu, line 694: to me this indicates that `ApplicationOperationVerifierError` should be defined as a member of `ChainCommunicationError` rather than the other way around. I think it's valuable to keep those error reasons as status labels (e.g. `Retry(AppLevelStatus(InsufficientAmount))`, or `Retry(AppLevelStatus(MalformedMessage))`)

Then this `match` could become something like 
```suggestion
                ChainCommunicationError(AppVerificationError(app_err)) => AppVerification(app_err.to_status()),
                _ => reason
            },
```

rust/main/chains/hyperlane-ethereum/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use async_trait::async_trait;
+use derive_new::new;
+use tracing::debug;
+
+use hyperlane_application::{ApplicationOperationVerifier, ApplicationOperationVerifierError};
+use hyperlane_core::{Decode, HyperlaneMessage, H256};
+use hyperlane_warp_route::TokenMessage;
+
+const WARP_ROUTE_MARKER: &str = "/";
```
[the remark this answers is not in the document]
ameten: It is in the description of the task, but I'll check with product team as well.

rust/main/hyperlane-base/src/settings/base.rs
```diff
@@ -145,11 +147,12 @@ macro_rules! build_contract_fns {
 type SequenceIndexer<T> = Arc<dyn SequenceAwareIndexer<T>>;
 
 impl Settings {
+    build_contract_fns!(build_application_operation_verifier, build_application_operation_verifiers -> dyn ApplicationOperationVerifier);
```
daniel-savu, line 150: I think creating operation verifiers here with `build_contract_fns` indicates that they interact with smart contracts. Wdyt about having a constructor that's more similar to `contract_syncs` instead? I'd ideally want to remove this `build_contract_fns` macro at some point

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -664,6 +674,26 @@ impl PendingMessage {
...
+        match self
+            .ctx
+            .application_operation_verifier
+            .verify(&self.app_context, &self.message)
+            .await
+        {
+            Ok(()) => reason,
+            Err(e) => match e {
+                InsufficientAmountError | MalformedMessageError(_) => {
+                    ApplicationOperationVerificationFailed
```
daniel-savu, line 691: I think it'd be useful to include the error type in the label btw, such as `AppVerificationFailed(InsufficientAmount)` or `AppVerificationFailed(MalformedMessageError)`

rust/main/applications/hyperlane-application/src/operation_verifier.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// Trait to verify if operation is permitted for application context
+#[async_trait]
+pub trait ApplicationOperationVerifier: Send + Sync {
+    /// Checks if message is permitted for application context
+    async fn verify(
+        &self,
+        app_context: &Option<String>,
+        message: &HyperlaneMessage,
+    ) -> Result<(), ApplicationOperationVerifierError>;
```
daniel-savu, line 30: imo this is more of a labeling function that a verifying one. Even though at the application level we may treat these labels as errors, idiomatic would be to not rely on the returned Error type unless something bad happened in this function. If the AppVerifier isn't able to provide a label that's not an error, and neither is labelling that e.g. this message transfers zero value on cosmos

So rather than use `Result<(), Error>` and always rely on the error, I think we should use `Option<AppSpecificInfo>` or similar. Being able to apply a label is not an error as far as this method is concerned, and being unable to apply a label is fine and we retun `None`

rust/main/chains/hyperlane-cosmos/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        // Starting from this point we assume that we are in a warp route context
+
+        let mut reader = Cursor::new(message.body.as_slice());
+        let token_message = TokenMessage::read_from(&mut reader)
+            .map_err(|_| MalformedMessageError(message.clone()))?;
+
+        if token_message.amount() > U256::zero() {
+            return Err(InsufficientAmountError);
+        }
```
daniel-savu, line 51: I think we want to apply a label here only when the amount is zero. If the transferred warp route amount is >0 the message is ok

rust/main/chains/hyperlane-ethereum/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,63 @@
...
+        if Self::check_leading_zeros(&recipient) {
+            return Err(MalformedMessageError(message.clone()));
+        }
+
+        Ok(())
+    }
+}
+
+impl EthereumApplicationOperationVerifier {
+    fn check_leading_zeros(address: &H256) -> bool {
```
daniel-savu, line 58: nit: by just reading the function name it's unclear to me what the returned boolean means (is true good or bad?). Maybe `has_enough_leading_zeroes(...)`?

rust/main/chains/hyperlane-ethereum/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        // Starting from this point we assume that we are in a warp route context
+
+        let mut reader = Cursor::new(message.body.as_slice());
+        let token_message = TokenMessage::read_from(&mut reader)
+            .map_err(|_| MalformedMessageError(message.clone()))?;
+
+        let recipient = token_message.recipient();
+        if Self::check_leading_zeros(&recipient) {
+            return Err(MalformedMessageError(message.clone()));
```
daniel-savu, line 50: does this mean we'll consider this message malformed when it has sufficient (`ETHEREUM_ADDRESS_LEADING_ZEROS_COUNT`) leading zeroes? My understanding was that we should alert when there are fewer than `ETHEREUM_ADDRESS_LEADING_ZEROS_COUNT` leading zeroes

ameten: Fix

rust/main/chains/hyperlane-ethereum/src/application/operation_verifier.rs
ameten: Fixed

rust/sealevel/programs/hyperlane-sealevel-token-collateral/Cargo.toml
ameten: Discussed in person. We won't separate it at this point.

rust/main/agents/relayer/src/msg/pending_message.rs
ameten: It will create circular dependency, so, we won't do it.

rust/main/agents/relayer/src/msg/pending_message.rs
ameten: All errors are incapsulated inside `verify`

rust/main/applications/hyperlane-application/src/operation_verifier.rs
ameten: Implemented

rust/main/hyperlane-base/src/settings/base.rs
ameten: `build_contract_fns` is already misused to instantiate `HyperlaneProvider` which does not depend on contract. We can plan to remove this macro as separate task.

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,72 @@
...
+            .rpc()
+            // We assume that account will contain no data
+            .get_minimum_balance_for_rent_exemption(0)
+            .await
+        {
+            Ok(m) => m.into(),
+            Err(_) => return None,
+        };
+
+        if token_message.amount() < minimum {
```
daniel-savu, line 66: i'm not sure this check handles everything described in https://github.com/hyperlane-xyz/issues/issues/1414
> If the recipient is a native warp route and the end recipient (i.e. the one encoded in the first 32 bytes of the message body) does not exist yet (i.e. no balance at all), then the native token amount being warped to that recipient must be enough to cover the minimum rent for that account. See the implementation in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4936 for inspo

- we need to add a check that the HyperlaneMessage recipient is a native warp route
- we need to add a check that the Warp Transfer recipient does not exist yet

I think that currently, this check can return `AmountBelowMinimum` regardless of the token being transferred which is incorrect, because synthetic tokens have their rent paid by the ATA payer

ameten: Good point. Fixed

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,144 @@
...
+        let context = match app_context {
+            Some(c) => c,
+            None => return None,
+        };
+
+        if !context.starts_with(WARP_ROUTE_PREFIX) {
+            return None;
+        }
+
+        Some(())
```
daniel-savu, line 114: can probably be even simpler but

```suggestion
        app_context
            .as_ref()
            .map(|context| context.starts_with(WARP_ROUTE_PREFIX))?;
        Some(())
```

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,144 @@
...
+            ?message,
+            "Sealevel application operation verifier",
+        );
+
+        // Separate dependency on self and network invocations for ease of unit testing
+        let check_account_does_not_exist_and_get_minimum = |account: H256| async move {
+            let Ok(false) = self.account_exists(account).await else {
+                return None;
+            };
+
```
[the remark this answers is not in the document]
ameten: It is not a refactor change.

Original one returns `None` if account exists as well as if we have communication error with the chain.
Updated one returns `None` only if account exists and continues if we have communication error with the chain.

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
ameten: You are right. It can be done simple:

```suggestion
    app_context
            .as_ref()
            .map(|context| context.starts_with(NATIVE_WARP_ROUTE_PREFIX))?
            .then_some(())
```

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -668,6 +682,30 @@ impl Relayer {
...
+        settings
+            .build_application_operation_verifiers(settings.origin_chains.iter(), core_metrics)
+            .await
+            .into_iter()
+            .filter_map(|(origin, app_context_verifier_res)| match app_context_verifier_res {
+                Ok(app_context_verifier) => Some((origin, app_context_verifier)),
+                Err(err) => {
+                    error!(?err, origin=?origin, "Critical error when building application operation verifier");
+                    chain_metrics.set_critical_error(origin.name(), true);
+                    None
```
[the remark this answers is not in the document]
ameten: Discussed in person. We decided to keep critical error here.

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -245,6 +251,13 @@ impl BaseAgent for Relayer {
                 } else {
                     transaction_gas_limit
                 };
+            let application_operation_verifier =
+                match application_operation_verifiers.get(destination) {
+                    Some(v) => v,
+                    // if mailbox was successfully built, but application operation verifier was not,
+                    // we continue. Metrics should contain a critical error at this point.
+                    None => continue,
```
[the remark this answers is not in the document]
ameten: I made application operation verifier optional for a chain.

rust/main/chains/hyperlane-cosmos/src/application.rs
```diff
@@ -0,0 +1,3 @@
+pub use operation_verifier::CosmosApplicationOperationVerifier;
+
+mod operation_verifier;
```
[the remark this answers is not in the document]
ameten: Discussed in person. We keep it as it is in the change.

rust/main/chains/hyperlane-cosmos/src/application/operation_verifier/tests.rs
```diff
@@ -0,0 +1,88 @@
+use hyperlane_core::{Encode, HyperlaneMessage, H256, U256};
```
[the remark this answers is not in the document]
ameten: Discussed in person. We keep it as it is in the change.

tkporter (approved):
very nice 🤝

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -301,7 +305,11 @@ impl PendingOperation for PendingMessage {
         {
             Ok(tx_cost_estimate) => tx_cost_estimate,
             Err(err) => {
-                return self.on_reprepare(Some(err), ReprepareReason::ErrorEstimatingGas);
+                let reason = self
+                    .clarify_reason(ReprepareReason::ErrorEstimatingGas)
+                    .await
+                    .unwrap_or(ReprepareReason::ErrorEstimatingGas);
```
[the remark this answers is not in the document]
ameten: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5366#discussion_r1954312334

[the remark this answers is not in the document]
daniel-savu: yep i also meant to not pass the repreparereason as an arg, my snippet was misleading

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,139 @@
...
+        message: &HyperlaneMessage,
+    ) -> Option<ApplicationOperationVerifierReport> {
+        debug!(
+            ?app_context,
+            ?message,
+            "Sealevel application operation verifier",
+        );
+
+        // Separate dependency on self and network invocations for ease of unit testing
+        let check_account_does_not_exist_and_get_minimum = |account: H256| async move {
```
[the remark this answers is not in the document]
ameten: I fix the `verify_method` param name. I would like to keep less coupling with internals like assigning special meaning to 0 amount.

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,139 @@
...
+};
+use hyperlane_warp_route::TokenMessage;
+
+use crate::SealevelProvider;
+
+const NATIVE_WARP_ROUTE_PREFIX: &str = "SOL/";
+// Native SOL warp routers
+// from https://github.com/hyperlane-xyz/hyperlane-registry/tree/359bd01a99cf3c9f2f938c4bc200c5261e2f5bc1/deployments/warp_routes/SOL
+lazy_static! {
+    static ref NATIVE_WARP_ROUTES: HashSet<H256> = {
```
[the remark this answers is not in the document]
ameten: Yes, it is excessive, but it makes it simpler to reason about what is happening: first focus on native warp routes, next focus on the native side of a native warp route.

rust/main/chains/hyperlane-cosmos/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct CosmosApplicationOperationVerifier {}
+
+#[async_trait]
+impl ApplicationOperationVerifier for CosmosApplicationOperationVerifier {
+    async fn verify(
+        &self,
+        app_context: &Option<String>,
+        message: &HyperlaneMessage,
+    ) -> Option<ApplicationOperationVerifierReport> {
+        debug!(
```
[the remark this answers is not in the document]
ameten: Fixed

rust/main/applications/hyperlane-operation-verifier/src/operation_verifier/report.rs
```diff
@@ -0,0 +1,31 @@
+use hyperlane_application::ApplicationReport;
+use hyperlane_core::{HyperlaneMessage, U256};
+
+/// Application operation verifier report
+#[derive(Debug, Eq, PartialEq)]
+pub enum ApplicationOperationVerifierReport {
+    /// Amount below minimum (minimum, actual)
+    AmountBelowMinimum(U256, U256),
```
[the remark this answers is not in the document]
ameten: Fixed

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -674,6 +686,24 @@ impl PendingMessage {
...
+
+        match self
+            .ctx
+            .application_operation_verifier
+            .as_ref()?
+            .verify(&self.app_context, &self.message)
+            .await
+        {
+            Some(r) => {
+                info!(original = ?reason, report = ?r, app = ?self.app_context, message = ?self.message, "Clarifying reprepare reason with application report");
```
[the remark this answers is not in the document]
ameten: Fixed

rust/main/chains/hyperlane-sealevel/src/application/operation_verifier.rs
```diff
@@ -0,0 +1,139 @@
...
+    ) -> Option<ApplicationOperationVerifierReport>
+    where
+        F: FnOnce(H256) -> Fut,
+        Fut: Future<Output = Option<U256>>,
+    {
+        use ApplicationOperationVerifierReport::{AmountBelowMinimum, MalformedMessage};
+
+        Self::verify_context(app_context)?;
+
+        // Starting from this point we assume that we are in a warp route context
```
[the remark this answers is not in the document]
ameten: Fixed

PR #5367, opened 2025-02-04
feat: deploy missing factories in core deploy

ljankovic-txfusion:
### Description

Adds a new `--fix-factories` flag to the `hyperlane core deploy` command that allows deploying only missing factory contracts while keeping existing core contracts.

### Drive-by changes

- Added `--fix-factories` flag to core deploy command
- Added factory deployment planning logic to identify missing factories

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/core/EvmCoreModule.ts
+++ b/typescript/sdk/src/core/EvmCoreModule.ts
@@ -4,7 +4,6 @@ import {
   Mailbox,
   Mailbox__factory,
   Ownable__factory,
-  ProxyAdmin__factory,
 } from '@hyperlane-xyz/core';
 import {
   Address,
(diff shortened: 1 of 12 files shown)


typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -418,13 +418,24 @@ export class EvmCoreModule extends HyperlaneModule<
...
+      multiProvider,
+      contractVerifier,
+      factoryDeploymentPlan,
+    } = params;
 
     const proxyFactoryDeployer = new HyperlaneProxyFactoryDeployer(
       multiProvider,
       contractVerifier,
+      false,
+      factoryDeploymentPlan,
```
ljankovic-txfusion, line 435: Had to expand constructor params to include deployment plan since `await proxyFactoryDeployer.deploy({` is calling child `HyperlaneProxyFactoryDeployer.deployContract`

PR #5370, opened 2025-02-04
feat(agents): Cosmos SDK Module

yjamin:
### Description  

This PR adds Agent support for the upcoming **Hyperlane Cosmos SDK Module**, called **CosmosNative**, to differentiate it from the existing CosmWasm (CW) implementation.  

### Drive-by Fixes  

None

### Related Issues  

- None reported.  

### Backward Compatibility  

- No breaking changes—this PR extends functionality without affecting existing implementations.  

### Testing  

The CosmosNative implementation has been tested through:  

- **Automated E2E tests** – Following the same approach as the CW implementation.  
- **Deployed preview** – Currently live on the **alpha** network for additional validation.  

#### E2E Testing Details  

- They pretty much follow the same structure as the CW tests.  
- They spin up a lightweight Cosmos SDK chain (**simapp**) with the Hyperlane Cosmos module.  

> **Note:** The simapp includes a **preview** version of the Hyperlane module since it's not officially released yet.  

To run the tests locally, use:  

```sh
cargo test -r -p run-locally --features cosmosnative -- cosmosnative::test --nocapture
```


tkporter (commented):
Only spent about 20 mins the other day poking around some of the first files, hopefully will have time to leave more feedback but posting for now

The biggest thing that stands out to me is just how much duplication there is with `hyperlane-cosmos`. I could see there being some arguments for keeping this separate but I feel like consolidating as much as possible is probably the better move. Curious what @daniel-savu @ameten think

rust/main/chains/hyperlane-cosmos-native/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+tendermint-rpc = { workspace = true }
+time = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true }
+reqwest = { workspace = true, features = ["json"] }
+tonic = { workspace = true, features = ["transport", "tls", "tls-roots"] }
+tracing = { workspace = true }
+tracing-futures = { workspace = true }
+url = { workspace = true }
+prost = "0.13.4"
```
ameten, line 43: I wonder if we can reference it in workspace?

rust/main/chains/hyperlane-cosmos-native/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+base64 = { workspace = true }
+bech32 = { workspace = true }
+cosmrs = { workspace = true, features = ["tokio", "grpc", "rpc"] }
+crypto = { path = "../../utils/crypto" }
+derive-new = { workspace = true }
+futures = { workspace = true }
+hex = { workspace = true }
+http = { workspace = true }
+hyperlane-core = { path = "../../hyperlane-core", features = ["async"] }
+hyperlane-cosmwasm-interface.workspace = true
```
ameten, line 22:
```suggestion
hyperlane-cosmwasm-interface = { workspace = true }
```

rust/main/Cargo.toml
ameten, line 1: I wonder why some dependencies are remove. We don't use them anymore?

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/forward.rs
```diff
@@ -110,7 +110,7 @@ impl<T: Debug + Clone + Sync + Send + Indexable + 'static> ForwardSequenceAwareS
     /// If there are no logs to index, returns `None`.
     /// If there are logs to index, returns the range of logs, either by sequence or block number
     /// depending on the mode.
-    #[instrument(ret)]
+    // #[instrument(ret)]
```
ameten, line 113: I wonder why this annotation was removed?

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -123,10 +123,15 @@ impl<T: Debug + Clone + Sync + Send + Indexable + 'static> BackwardSequenceAware
...
-        // Query the block range ending at the current_indexing_snapshot's at_block.
-        current_indexing_snapshot
+        let start = current_indexing_snapshot
             .at_block
-            .saturating_sub(self.chunk_size)..=current_indexing_snapshot.at_block
+            .saturating_sub(self.chunk_size);
+        // limit this to 1
+        if start < 1 {
+            return 1..=current_indexing_snapshot.at_block;
+        }
```
ameten, line 132: I wonder if it will work for all chains. Presumably, Ethereum genesis block has height 0.

rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
```diff
@@ -16,7 +16,8 @@ pub trait CheckpointSyncer: Debug + Send + Sync {
     /// Update the latest index of this syncer if necessary
     async fn update_latest_index(&self, index: u32) -> Result<()> {
         let curr = self.latest_index().await?.unwrap_or(0);
-        if index > curr {
+        // always write the 0th index
+        if index > curr || index == 0 {
```
ameten, line 20: I wonder what is the issue which requires this change?

rust/main/config/testnet_config.json
```diff
@@ -1515,7 +1515,7 @@
       "protocol": "ethereum",
       "rpcUrls": [
         {
-          "http": "http://127.0.0.1:8545"
+          "http": "http://0.0.0.0:8545"
```
ameten, line 1518: I wonder if this change was intentional?

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
ameten, line 1: I wonder if it will make sense to have the functionality shared by CosmosWasm and CosmosNative implementations to be extracted into separate module and re-used?

Example: `sender_and_nonce` method in `CosmosProvider` and `CosmosNativeProvider`. There can be more examples like this.

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
```diff
@@ -0,0 +1,463 @@
...
+use serde::{de::DeserializeOwned, Deserialize};
+use tendermint::{hash::Algorithm, Hash};
+use tendermint_rpc::{
+    client::CompatMode,
+    endpoint::{block, block_results, tx},
+    Client, HttpClient,
+};
+use time::OffsetDateTime;
+use tonic::async_trait;
+use tracing::{debug, trace, warn};
```
ameten, line 43: It is worth to do everywhere.

It will be great to format imports as follows.
1. std (and empty line)
2. external crates (and empty line)
3. internal crates (like `hyperlane_core`) (and empty line)
4. `crate::` imports (and empty line)
5. `super::` imports (and empty line)

```suggestion
use std::{io::Cursor, ops::Deref};

use cosmrs::{
    crypto::PublicKey,
    proto::{
        cosmos::{
            auth::v1beta1::QueryAccountRequest,
            bank::v1beta1::{QueryBalanceRequest, QueryBalanceResponse},
            base::abci::v1beta1::TxResponse,
        },
        tendermint::types::Block,
    },
    tx::{SequenceNumber, SignerInfo, SignerPublicKey},
    AccountId, Any, Coin, Tx,
};
use derive_new::new;
use itertools::Itertools;
use prost::Message;
use reqwest::Error;
use serde::{de::DeserializeOwned, Deserialize};
use tendermint::{hash::Algorithm, Hash};
use tendermint_rpc::{
    client::CompatMode,
    endpoint::{block, block_results, tx},
    Client, HttpClient,
};
use time::OffsetDateTime;
use tonic::async_trait;
use tracing::{debug, trace, warn};

use hyperlane_core::{
    h512_to_bytes,
    rpc_clients::{BlockNumberGetter, FallbackProvider},
    utils::{self, to_atto},
    AccountAddressType, BlockInfo, ChainCommunicationError, ChainInfo, ChainResult,
    ContractLocator, HyperlaneChain, HyperlaneDomain, HyperlaneMessage, HyperlaneProvider,
    HyperlaneProviderError, LogMeta, ModuleType, RawHyperlaneMessage, TxnInfo, TxnReceiptInfo,
    H256, H512, U256,
};

use crate::{
    ConnectionConf, CosmosAccountId, CosmosAddress, CosmosAmount, HyperlaneCosmosError, Signer,
};

use super::{rest::RestProvider, RpcProvider};
```

ameten:
I would agree with @tkporter in [this comment](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5370#pullrequestreview-2604321532) that reducing the duplication of the code will be great since it will make the review of the change much easier.

It will be ideal to have the refactoring of the re-used parts as a separate PR and, if it too hard, at least as a separate commit so that it can be reviewed without other CosmosNative changes.

rust/main/chains/hyperlane-cosmos-native/src/indexers/dispatch.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[async_trait]
+impl SequenceAwareIndexer<HyperlaneMessage> for CosmosNativeDispatchIndexer {
+    #[instrument(err, skip(self), ret)]
+    #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
+    async fn latest_sequence_count_and_tip(&self) -> ChainResult<(Option<u32>, u32)> {
+        let tip = self.get_finalized_block_number().await?;
+        let sequence = self
+            .provider
+            .rest()
```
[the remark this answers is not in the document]
yjamin: In Cosmos, gRPC, REST, and RPC all communicate with the chain through ABCI. They serve as wrappers around raw ABCI queries, meaning the main differences is in convenience and probably some performance overhead.

We chose to remove gRPC since it is not fully supported on our public KYVE endpoints, but this shouldn't matter for the agents implementation. We used REST as the query client implementation, because of its simplicity and human readability.

If we want to stick with gRPC and RPC, we would need to manually implement the gRPC Query Client for the Cosmos module. This includes handling protobuf serialization for each struct, i think it would make sense to move this into a separate repository (similar to the CW queries) and add that to the monorepo.

rust/main/Cargo.toml
yjamin: Not too sure. I only used the cli for managing dependencies. Can re-add them if we want to be sure.

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
yjamin: Yes, good point, that was just annoying when testing with a local chain.

This will probably never be the case in a real world deployment, however i think it would make sense to add an optional additional property like `genesis_height` or `deployment_height` in [`IndexSettings`](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/hyperlane-base/src/settings/chains.rs#L187) that defines on what height the mailbox has been deployed and this should be the lower bound of each indexer, as there can't be any events prior to that height.

The default value of that property can then be set by the protocol type.

rust/main/hyperlane-base/src/settings/chains.rs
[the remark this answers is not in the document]
yjamin: Correct, they are not supported by the Cosmos Module implementation yet. Will add a comment for clarity 👍

rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
yjamin: Can't seem to replicate the issue anymore, nor did i fully understand it at that time.
When i first started the implementation, the syncer wouldn't write the 0th index. That was probably due to a faulty implementation of `latest_checkpoint` in our `merkle_tree_hook`. A lot has changed since then, i would just remove it again.

rust/main/config/testnet_config.json
yjamin: Nope, not intentional, will remove. Thanks :+1:

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
yjamin: Good to know, will adapt this convention :+1:

rust/main/Cargo.toml
```diff
@@ -122,10 +119,7 @@ sea-orm-migration = { version = "0.11.1", features = [
   "sqlx-postgres",
   "runtime-tokio-native-tls",
 ] }
-semver = "1.0"
 serde = { version = "1.0", features = ["derive"] }
-serde_bytes = "0.11"
-serde_derive = "1.0"
```
daniel-savu, line 128: were deps in this file removed to speed up development? I assume they are used in non-cosmos envs

rust/main/Cargo.toml
```diff
@@ -168,15 +162,9 @@ ya-gcp = { version = "0.11.3", features = ["storage"] }
 ## TODO: remove this
 cosmwasm-schema = "2.2.0"
 
-[profile.release.package.access-control]
-overflow-checks = true
-
```
daniel-savu, line 173: same for these checks

rust/main/chains/hyperlane-cosmos-native/src/providers/rpc.rs
```diff
@@ -0,0 +1,378 @@
...
+        Ok(RpcProvider {
+            provider,
+            conf,
+            signer,
+            gas_price,
+        })
+    }
+
+    /// Get the transaction by hash
+    pub async fn get_tx(&self, hash: &H512) -> ChainResult<TxResponse> {
```
daniel-savu, line 106: could you please merge in the changes from `hyperlane-cosmos/src/providers/rpc.rs` so we get some metrics from this provider

yjamin: Yes, for sure will add them. Just merged main into so we don't have any conflicts anymore. Will add the metrics once E2E tests pass.

rust/main/Cargo.toml
yjamin: Still not sure why they were removed, I'll re-add them to be sure. :+1:

rust/main/utils/run-locally/src/cosmosnative/mod.rs
```diff
@@ -0,0 +1,505 @@
...
+fn install_sim_app() -> PathBuf {
+    let target = make_target();
+
+    let dir_path = tempdir().unwrap().into_path();
+    let dir_path = dir_path.to_str().unwrap();
+
+    let release_name = format!("{BINARY_NAME}_{target}");
+    log!("Downloading Sim App {}", release_name);
+    // TODO: point to official releases on github, right now we only have this private preview binary of a simulation app
+    // once there are actual releases on the github, we can use them
```
daniel-savu, line 213: is there anything we can do to publish this as an official release or are we blocked on someone?

daniel-savu (commented):
great work! Left mostly nitpicks, the PR is pretty hard to review due to the logic duplication. Hopefully we'll find any remaining issues once the logic is deduplicated with CW

rust/main/utils/run-locally/src/cosmosnative/mod.rs
mbreithecker: Think we can do a first initial beta release tomorrow.

rust/main/chains/hyperlane-cosmos-native/src/indexers/indexer.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    /// Get the inner event
+    pub fn inner(self) -> T {
+        self.event
+    }
+}
+
+#[async_trait]
+/// todo
+pub trait EventIndexer<T: PartialEq + Send + Sync + 'static>: Indexer<T>
```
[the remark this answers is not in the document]
yjamin: Hey, yeah so my idea was to make a generic implementation for Indexer, so that structs only have to implement `EventIndexer<T>` and implement `Indexer<T>` automatically. Something like this:

```rust
impl <T: EventIndexer> Indexer for T {
    ...
}
```

However, I didn't get it to work and decided to just delegate the calls for now.

rust/main/chains/hyperlane-cosmos-native/src/indexers/delivery.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use crate::{ConnectionConf, CosmosNativeProvider, HyperlaneCosmosError, RpcProvider};
+
+use super::{EventIndexer, ParsedEvent};
+
+/// delivery indexer to check if a message was delivered
+#[derive(Debug, Clone)]
+pub struct CosmosNativeDeliveryIndexer {
+    provider: CosmosNativeProvider,
+    address: H256,
```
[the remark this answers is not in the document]
yjamin: The address provided here has nothing to do with a cosmos bech32 address. In contrast to the CW - where everything is a contract and each contract has a bech32 address - the cosmos module has instances of code, exactly like you'd write code in any language. We therefore had to build our own system to map instances to an address and we chose to go with 32byte hex addresses. Hope that cleans things up

rust/main/chains/hyperlane-cosmos-native/src/indexers/gas_paymaster.rs
```diff
@@ -0,0 +1,180 @@
+use std::ops::RangeInclusive;
```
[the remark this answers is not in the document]
yjamin: Good point, will change that :+1:

rust/main/hyperlane-base/src/contract_sync/cursors/mod.rs
```diff
@@ -88,6 +91,7 @@ impl Indexable for Delivery {
             HyperlaneDomainProtocol::Fuel => todo!(),
             HyperlaneDomainProtocol::Sealevel => CursorType::SequenceAware,
             HyperlaneDomainProtocol::Cosmos => CursorType::RateLimited,
+            HyperlaneDomainProtocol::CosmosNative => CursorType::RateLimited,
```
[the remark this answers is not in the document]
yjamin: How would the relayer know the difference between `CW` & `CosmosNative`? They both use different logic to index & submit message

rust/main/utils/run-locally/src/cosmosnative/mod.rs
```diff
@@ -0,0 +1,505 @@
...
+        .run()
+        .join();
+
+    concat_path(dir_path, BINARY_NAME)
+}
+
+#[allow(dead_code)]
+fn run_locally() {
+    // TODO: store all the created processes directly to not have lost children on crash
+    let hypd = install_sim_app().as_path().to_str().unwrap().to_string();
```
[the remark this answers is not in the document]
yjamin: Oh yeah for sure, it was just helpful for debug purposes. I'll add a debug flag similar to the Sealevel tests and only print logs then :+1:

rust/main/chains/hyperlane-cosmos-native/src/indexers/gas_paymaster.rs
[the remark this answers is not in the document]
yjamin: About the `merkle_tree_hook.rs` name for the `tree_insertion.rs` i don't think this is a fitting name as this struct only indexes tree insertions and does not give any information about the actual hook state so i would leave it at this if thats fine

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
```diff
@@ -0,0 +1,442 @@
...
+        &self.grpc
+    }
+
+    /// todo
+    pub async fn reorg_to_height(&self, reorg: &ReorgPeriod) -> ChainResult<u32> {
+        let height = self.rpc.get_block_number().await? as u32;
+        match reorg {
+            ReorgPeriod::None => return Ok(height),
+            ReorgPeriod::Blocks(blocks) if blocks.get() >= height => Ok(1), // height has to be at least 1 -> block 0 does not exist in cosmos
+            ReorgPeriod::Blocks(blocks) => Ok(height - blocks.get()),
```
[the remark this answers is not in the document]
yjamin: Yeah definitely, was not aware of the `checked_sub` method. Thanks!

rust/main/hyperlane-base/src/contract_sync/cursors/mod.rs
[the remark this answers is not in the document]
daniel-savu: Had a deeper look and chatted to @tkporter. Even if from an agents' point of view they have many similarities, it makes sense to have the new `HyperlaneDomainProtocol` since contracts and how we interact with them are different.

what I originally meant is `HyperlaneDomainProtocol::Cosmos` and `HyperlaneDomainProtocol::CosmosNative` have many things in common
- they use the same indexing cursor types (as configured in this file)
- there are several instances of special-casing the Cosmos protocol type which wouldn't have broken at compile time and were easy to miss, but likely need to be updated to include `CosmosNative` now
    - in the `confirm` step of the OpSubmitter (renamed MessageProcessor in the redesign doc), there is special casing for `Cosmos`, which should probably also apply to `CosmosNative` - [here](https://github.com/bcp-innovations/hyperlane-monorepo/blob/9b0bd440144d5a57a38c44eaaf5cadc362f88084/rust/main/agents/relayer/src/msg/op_submitter.rs#L431)
    - default decimals - [here](https://github.com/bcp-innovations/hyperlane-monorepo/blob/9b0bd440144d5a57a38c44eaaf5cadc362f88084/rust/main/hyperlane-core/src/metrics/agent.rs#L26)

however:
- `HyperlaneDomainProtocol` is used to decide which `ConnectionConf` to create out of the settings, so having separate `Cosmos` and `CosmosNative` is required
- for `Cosmos` and `CosmosNative`, `ConnectionConf` structs are the same except for one additional `gas_multiplier` field which could be `None` for CosmWasm. But `ConnectionConf` is used differently for the two, to decide which provider and indexing types to use. 

my suggested actions:
- check all places where `HyperlaneDomainProtocol::Cosmos` and `ConnectionConf::Cosmos` are used, and make sure the CosmosNative case is also accounted for correctly in that logic
- for a follow up PR: The two `ConnectionConf` cosmos variants should be merged and have an enum inside of the unifying `ConnectionConf::Cosmos(...)` determine what type of cosmos provider gets built

rust/main/chains/hyperlane-cosmos-native/src/indexers/gas_paymaster.rs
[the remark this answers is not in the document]
daniel-savu: that's the only event type emitted by the merkle tree hook contract - would still rather rename for consistency with the other implementations

rust/main/chains/hyperlane-cosmos-native/src/indexers/delivery.rs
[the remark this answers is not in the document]
daniel-savu: oh I see, makes sense

rust/main/chains/hyperlane-cosmos-native/src/indexers/indexer.rs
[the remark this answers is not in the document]
yjamin: I think there might be a misunderstanding. Right now the `EventIndexer` is exclusively used for `CosmosNative` and implements some useful defaults that generalize over the different Indexer specific to Cosmos. I think it'd be useful to keep that trait as these functions handle a lot of boilerplate (like parsing Blocks, Transactions and extracting events).

Also this logic can later be shared between CW & CosmosModule to have a single trait that is responsible for parsing raw block data & transactions and pipe the relevant events into the protocol specific parsing logic.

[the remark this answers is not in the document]
daniel-savu: might be better to rename to `CosmosEventIndexer` then but sgtm

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
```diff
@@ -0,0 +1,442 @@
...
+            ReorgPeriod::None => Ok(height),
+            // height has to be at least 1 -> block 0 does not exist in cosmos
+            ReorgPeriod::Blocks(blocks) => Ok(height.checked_sub(blocks.get()).unwrap_or(1)),
+            ReorgPeriod::Tag(_) => Err(ChainCommunicationError::InvalidReorgPeriod(reorg.clone())),
+        }
+    }
+
+    fn check_msg_process(tx: &Tx) -> ChainResult<Option<H256>> {
+        // check for all transfer messages
+        let remote_transfers: Vec<Any> = tx
```
daniel-savu, line 133: fyi "remote transfer" is only used in the context of warp route (token) transfers elsewhere in the codebase. I assume this just refers to outgoing raw messages here

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
```diff
@@ -0,0 +1,442 @@
...
+        let height = self.rpc.get_block_number().await? as u32;
+        match reorg {
+            ReorgPeriod::None => Ok(height),
+            // height has to be at least 1 -> block 0 does not exist in cosmos
+            ReorgPeriod::Blocks(blocks) => Ok(height.checked_sub(blocks.get()).unwrap_or(1)),
+            ReorgPeriod::Tag(_) => Err(ChainCommunicationError::InvalidReorgPeriod(reorg.clone())),
+        }
+    }
+
+    fn check_msg_process(tx: &Tx) -> ChainResult<Option<H256>> {
```
daniel-savu, line 131: naming here could be more suggestive, maybe `parse_message_recipient_from_tx(...)`?

rust/main/utils/run-locally/src/cosmosnative/types.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl AgentConfig {
+    pub fn new(node: &Deployment) -> Self {
+        AgentConfig {
+            name: node.name.clone(),
+            domain_id: node.domain,
+            metrics_port: node.metrics,
+            mailbox: node.contracts.mailbox.clone(),
+            interchain_gas_paymaster: node.contracts.igp.clone(),
+            validator_announce: node.contracts.mailbox.clone(),
```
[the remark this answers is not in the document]
yjamin: This was a design decision of the module. As the announce functionality was very minimal and we are not limited to the same extent as contracts we included the logic in the mailbox. The validator always announces storage locations bound to a mailbox, this was our intention behind this

rust/main/utils/run-locally/src/cosmosnative/mod.rs
```diff
@@ -0,0 +1,503 @@
...
+        &node1.contracts.tokens[0],
+        &node2.domain.to_string(),
+        account,
+        1000000u32,
+    );
+    node2.chain.remote_transfer(
+        KEY_CHAIN_VALIDATOR.0,
+        &node1.contracts.tokens[0],
+        &node1.domain.to_string(),
+        account,
```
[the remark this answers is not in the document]
yjamin: This is a consequence of the deterministic address generation in the module. Because both nodes are initialized in the same way, both of them share the same addresses for their contracts. 

Meaning in this case both `node1.contracts.tokens[0]` equal `node2.contracts.tokens[0]`.

I didn't think much of it when writing the code, because that was totally clear to me, but this for sure is confusing for someone that is not aware of that fact. 😅 

I'll change that, good catch 👍

.github/workflows/test.yml
```diff
@@ -306,6 +306,13 @@ jobs:
         env:
           RUST_BACKTRACE: 'full'
 
+      - name: agent tests (CosmosNative)
+        run: cargo test --release --package run-locally --bin run-locally --features cosmosnative -- cosmosnative::test --nocapture
+        if: ${{ matrix.e2e-type == 'cosmosnative' && steps.check-rust-changes.outputs.rust_changes == 'true' }}
+        working-directory: ./rust/main
+        env:
+          RUST_BACKTRACE: 'full'
```
paulbalaji, line 314: not specific to this PR at all but i think you can remove setting RUST_BACKTRACE from each of these e2e invocations, since we're already setting it once for the entire workflow at the top https://github.com/bcp-innovations/hyperlane-monorepo/blob/6311a5493c39fdd8c3fb7b7719c219918e09be58/.github/workflows/test.yml#L22

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
yjamin: Correct, that comment wasn't updated. Will change

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
yjamin: Yes, will change that to `parse_tx_message_recipient` and refactor it into smaller functions to make it more readable.

.github/workflows/test.yml
yjamin: I'd keep it as it is for this PR to not introduce any potential breaking changes

Issue #5272, opened 2025-01-23
Infra's checker should use registry-checked-in warp config


nambrot:

nambrot:
Re-opening this to have check-deploy specify a registry commit SHA as a command line arg

PR #5373, opened 2025-02-04
feat: Update infra checker use MergedRegistry

ltyu:
### Description
This PR updates the infra warp getter to use a MergeRegistry such that it uses the `DEFAULT_REGISTRY_URI` as the FileSystemRegistry and then overrides with `REGISTRY_URI`. Also adds the option of adding a [branch](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5373/files#diff-f749a128b49dc1373226397e4b16f5a195e337733c2a2d3638940aeaa9e40a65R95) to read from the GithubRegistry

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5272

### Backward compatibility
Yes, the intention of this is to have the same functionality as `getRegistry()` of infra registry.ts but with the option of using the GithubRegistry

### Testing
Manual


typescript/infra/config/registry.ts
```diff
@@ -64,13 +70,43 @@ export function getRegistry(): FileSystemRegistry {
...
+  logger.info(`Using ${overrideRegistryUri} as override registry`);
+  const registries = [primaryRegistryUri, overrideRegistryUri]
+    .map((uri) => uri.trim())
+    .filter((uri) => !!uri)
+    .map((uri) => {
+      if (isHttpsUrl(uri)) {
+        return new GithubRegistry({
+          uri,
+          logger: rootLogger.child({ module: 'infra-github-registry' }),
+          branch: process.env.REGISTRY_BRANCH || 'main',
```
ltyu, line 95: not really sure if we need this. may remove...

typescript/infra/config/registry.ts
```diff
@@ -64,13 +70,43 @@ export function getRegistry(): FileSystemRegistry {
...
-    proxyUrl: REGISTRY_PROXY,
-    logger: rootLogger.child({ module: 'infra-registry' }),
+/**
+ * Creates a new MergedRegistry using the provided URIs
+ * The intention of the MergedRegistry is to join the common data
+ * from a primary URI (such as a filesystem)
+ * and an override one (such as a github directory)
+ * @returns a new MergedRegistry
+ */
+export function getMergedRegistry(
```
nambrot, line 80: Could we just import the CLI function here and then call this with `primaryRegistryUri` from `DEFAULT_REGISTRY_URI`? Why introduce another ENV var and this logic?

ltyu: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5398

PR #5385, opened 2025-02-06
feat: enroll berachain

paulbalaji:
### Description

feat: enroll berachain

### Drive-by changes

- add renzo/luganodes validators to default unichain set
- add bera multisend address overrides in SDK
- support reading native token transfers in the GovernTransactionReader
- add deploy-only tx override for matchain
- fix default strategy resolution in `MultiChainResolver`
	- encountered issue when attempting `hyperlane status`

### DIFF
--- a/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
+++ b/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
@@ -134,10 +134,6 @@ export class MultiChainResolver implements ChainResolver {
       chains.push(argv.origin);
     }
 
-    if (argv.destination) {
-      chains.push(argv.destination);
-    }
-
     if (argv.chain) {
       chains.push(argv.chain);
     }
@@ -149,9 +145,13 @@ export class MultiChainResolver implements ChainResolver {
       return Array.from(new Set([...chains, ...additionalChains]));
     }
 
-    return chains.length > 0
-      ? chains
-      : Array.from(this.getEvmChains(multiProvider));
+    // If no destination is specified, return all EVM chains
+    if (!argv.destination) {
+      return Array.from(this.getEvmChains(multiProvider));
+    }
+
+    chains.push(argv.destination);
+    return chains;
   }
 
   private async getWarpRouteConfigChains(
--- a/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
+++ b/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
@@ -128,14 +128,14 @@ export class MultiChainResolver implements ChainResolver {
     argv: Record<string, any>,
   ): Promise<ChainName[]> {
     const { multiProvider } = argv.context;
-    const chains = [];
+    const chains = new Set<ChainName>();
 
     if (argv.origin) {
-      chains.push(argv.origin);
+      chains.add(argv.origin);
     }
 
     if (argv.chain) {
-      chains.push(argv.chain);
+      chains.add(argv.chain);
     }
 
     if (argv.chains) {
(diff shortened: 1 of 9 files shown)


typescript/sdk/src/utils/gnosisSafe.js
```diff
@@ -51,6 +51,11 @@ const chainOverrides = {
     multiSend: '0x0dFcccB95225ffB03c6FBB2559B530C2B7C8A912',
     multiSendCallOnly: '0xf220D3b4DFb23C4ade8C88E526C1353AbAcbC38F',
   },
+  // berachain
+  80094: {
+    multiSend: '0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761',
+    multiSendCallOnly: '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
```
daniel-savu, line 57: we use different multisend contracts for reads and writes?

typescript/infra/src/tx/govern-transaction-reader.ts
```diff
@@ -169,6 +174,23 @@ export class GovernTransactionReader {
...
+  private isNativeTokenTransfer(tx: AnnotatedEV5Transaction): boolean {
+    return !tx.data && !!tx.value && !!tx.to;
+  }
+
+  private async readNativeTokenTransfer(
+    chain: ChainName,
+    tx: AnnotatedEV5Transaction,
+  ): Promise<GovernTransaction> {
+    const { symbol } = await this.multiProvider.getNativeToken(chain);
+    const valueInEth = ethers.utils.formatEther(tx.value ?? BigNumber.from(0));
```
daniel-savu, line 186: what if the native token's not eth

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -215,6 +218,10 @@ export const icas: Partial<
   matchain: '0x66af72e46b3e8DFc19992A2A88C05d9EEFE01ffB',
   unitzero: '0x66af72e46b3e8DFc19992A2A88C05d9EEFE01ffB',
   trumpchain: '0x56895bFa7f7dFA5743b2A0994B5B0f88b88350F9',
+
+  // Q5, 2025 batch
+  // ----------------------------------------------------------
+  // berachain: '0x56895bFa7f7dFA5743b2A0994B5B0f88b88350F9',
```
daniel-savu, line 224: rm these commented out lines?
also you mean q5 2024 lol

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -134,26 +134,24 @@ export class MultiChainResolver implements ChainResolver {
...
-    if (argv.destination) {
-      chains.push(argv.destination);
-    }
-
     if (argv.chain) {
       chains.push(argv.chain);
     }
 
-    if (!argv.chains && chains.length === 0) {
+    if (argv.chains) {
```
daniel-savu, line 141: does this mean that if `destination` is defined we don't want to include it in here? also wondering what happens if someone passes `", "` as `chains` input and `additionalChains` ends up empty

daniel-savu (approved):

typescript/sdk/src/utils/gnosisSafe.js
paulbalaji: this is just taken from the safe-deployments repo

https://github.com/safe-global/safe-deployments/blob/9d89b7ae83b84e02209d95e436ea230f0b10b307/src/assets/v1.3.0/multi_send.json#L7
https://github.com/safe-global/safe-deployments/blob/9d89b7ae83b84e02209d95e436ea230f0b10b307/src/assets/v1.3.0/multi_send_call_only.json#L7

and cross-referenced with the deployed contracts on berachain

paulbalaji: multisend vs multisendcallonly is not read vs write, it's whether there are nested delegate calls or not

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
paulbalaji: yes - we're giving precedence to `origin` and `chain` and `chains` before finally looking at `destination`

PR #5391, opened 2025-02-07
fix(checker): Add MissingRouterViolation when config misses enrolled routers

Mo-Hussain:
### Description

- In response to an error thrown when there are actual enrolled routers that are not present in the config, this was encountered running the warp checker against a route not managed by AW. 

Here is the violation output
image

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/router/types.ts
+++ b/typescript/sdk/src/router/types.ts
@@ -48,16 +48,23 @@ export interface ClientViolation extends CheckerViolation {
 }
 
 export enum RouterViolationType {
-  EnrolledRouter = 'EnrolledRouter',
+  MisconfiguredEnrolledRouter = 'MisconfiguredEnrolledRouter',
+  MissingRouter = 'MissingRouter',
 }
 
 export interface RouterViolation extends CheckerViolation {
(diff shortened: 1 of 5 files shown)


Mo-Hussain:
Can readd the decent manage zeronetwork route to the check-warp-deploy cron job if we are happy with this change

typescript/sdk/src/router/types.ts
```diff
@@ -48,11 +48,12 @@ export interface ClientViolation extends CheckerViolation {
 }
 
 export enum RouterViolationType {
-  EnrolledRouter = 'EnrolledRouter',
```
paulbalaji, line 51: Does this need to be renamed? Could be smaller diff if we just add the new enum without changing this one

paulbalaji (commented):
1 comment otherwise lgtm

typescript/sdk/src/router/types.ts
Mo-Hussain: Renamed to make the intention clearer given the new violation type

paulbalaji (approved):

Issue #5320, opened 2025-01-29
CLI Warp Checker should reference deploy/warpcore config sanely


nambrot:
Current behavior:

- `warp check` reads deploy config from [DEFAULT_WARP_ROUTE_DEPLOYMENT_CONFIG_PATH](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/79f8197f379eed08e6c0312532a8e8f17a9cb84f/typescript/cli/src/commands/options.ts#L95)
- `warp check` can be configured to read warp deploy config with the `--config/-i` flag to a file path
- `warp check` reads warpcore config from specifying the `--symbol` flag and if a specific warpCore config couldn't be found, it will give an interactive wizard to pick it
- `warp check` can be configured to read warpcore config from the `--warp/-wc` flag to a file path

Expected behavior:

- `warp check` can be configured with a `--warpRouteId` flag to find both the warpdeploy config (now in the configured registry due to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5237) and the warpcoreconfig
- when not specified, an interactive wizard should give me the options of all warp route Ids in the specified warp registry. I should be able to use typeahead search to find the warp route that I want
- individual file paths to `--config/-i` for warp deploy config and `--warp/-wc` should still apply, however both have to be given together

nambrot:
@mshojaei-txfusion this should be assinged to you and is in review yeah?

PR #5395, opened 2025-02-07
feat: Add warp route ID lookup and validation to `warp check` command

ljankovic-txfusion:
### Description

This PR introduces `--warpRouteId` flag to `hl warp check` command that allows users to directly specify a warp route ID to check. When `--warpRouteId` is not specified, a user is prompted to search for a route ID from the registry.

1. Added a new option `warpRouteIdCommandOption` to specify warp routes by ID, making it easier to reference specific routes without needing full configuration files.

2. Modified the `check` command to accept either a warp route ID, symbol, or configuration file or , making it more flexible for different use cases, also to support running without any parameters, providing an interactive search interface that displays all available warp routes from the registry.

3. Removed the requirement for the `config` option to be mandatory in the `check` command, allowing users to use other identification methods.

4. Added `WarpRouteDeployConfigMailboxRequired` type import from SDK to ensure proper type safety when working with warp route configurations.

### Drive-by changes

1. Refactored the `runWarpRouteRead` function to separate concerns by:
   - Moving XERC20 limit checking to a standalone `logXerc20Limits` function
   - Adding input validation with clear error messages
   - Creating a helper function `deriveWarpRouteConfigs` to handle the common logic

6. Created a new function `getWarpRouteConfigsByCore` to specifically get warp route configurations from a core config, enabling registry-based lookup.

7. Added `validateEvmCompatibility` to centralize chain validation logic, improving code maintainability.

8. Removed `ChainSelectionMode.WARP_READ` and related resolution method from `MultiChainResolver`, simplifying the chain resolution process. [As we do not need signer for warp read anymore and as in new version of `runWarpRouteRead` we do not need context state to get warp core.

9. Removed the `warpCoreConfig` property from `CommandContext` interface, suggesting a move toward more explicit parameter passing rather than state in context.

10. Added utility function `getWarpConfigs` that provides a unified way to retrieve warp configurations from multiple sources (registry ID lookup, file paths, symbol-based lookup, or interactive search), eliminating redundant code and simplifying the configuration retrieval process.

11. Added utility function `filterWarpConfigsToMatchingChains` that ensures compatibility between configurations by comparing chain sets, highlighting mismatches with clear visual indicators, and filtering to only include chains that exist in both configs, preventing errors when configs have different chain coverage.

12. Modify `hyperlaneWarpCheckRaw` to support additional parameters: `warpCoreConfigPath` and `warpRouteId`.

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5320

Resolves ENG-1032

### Backward compatibility

Yes

### Testing

Updated cli e2e tests


typescript/cli/src/commands/options.ts
```diff
@@ -258,3 +258,9 @@ export const avsChainCommandOption: Options = {
   demandOption: true,
   choices: ['holesky', 'ethereum'],
 };
+
+export const warpRouteIdCommandOption: Options = {
+  type: 'string',
+  description: 'Warp route ID to check',
```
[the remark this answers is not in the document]
nambrot: So don't just say check

typescript/cli/src/read/warp.ts
```diff
@@ -32,25 +32,57 @@ export async function runWarpRouteRead({
...
-    // TODO: merge with XERC20TokenAdapter and WarpRouteReader
-    const xerc20Limits = await Promise.all(
-      warpCoreConfig.tokens
-        .filter(
-          (t) =>
-            t.standard === TokenStandard.EvmHypXERC20 ||
-            t.standard === TokenStandard.EvmHypXERC20Lockbox,
-        )
-        .map(async (t) => {
+  if (warpCoreConfig) {
```
nambrot, line 39: This is changed with the other PRs yeah?

mshojaei-txfusion: Yes it is, will refactor and simplify this file more there.

typescript/cli/src/utils/warp.ts
```diff
@@ -33,3 +43,78 @@ export async function getWarpCoreConfigOrExit({
 
   return warpCoreConfig;
 }
+
+/**
```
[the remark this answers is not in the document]
mshojaei-txfusion: created this issue for that function https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5747

typescript/cli/src/commands/options.ts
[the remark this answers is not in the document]
mshojaei-txfusion: resolved

typescript/cli/src/utils/warp.ts
```diff
@@ -33,3 +43,78 @@ export async function getWarpCoreConfigOrExit({
...
+  context,
+  warpRouteId,
+  config,
+  warp,
+  symbol,
+}: {
+  context: CommandContext;
+  warpRouteId?: string;
+  config?: string;
+  warp?: string;
```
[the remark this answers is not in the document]
mshojaei-txfusion: I did change name on functions, but not on commands as I was not sure about it.

typescript/cli/src/commands/warp.ts
```diff
@@ -358,35 +361,47 @@ export const check: CommandModuleWithContext<{
       demandOption: false,
     },
     config: inputFileCommandOption({
-      defaultPath: DEFAULT_WARP_ROUTE_DEPLOYMENT_CONFIG_PATH,
       description: 'The path to a warp route deployment configuration file',
+      demandOption: false,
+      alias: 'wd',
```
nambrot, line 366: Don't know that I love this alias, can it be `wdConfig`?

typescript/cli/src/commands/warp.ts
```diff
@@ -358,35 +361,47 @@ export const check: CommandModuleWithContext<{
...
+      warpCoreConfigPath: warp,
+    });
 
-    if (!warpCoreConfig) {
-      throw new Error('No warp core config found');
+    // First validate that warpCoreConfig chains match warpDeployConfig
+    const deployConfigChains = Object.keys(warpDeployConfig);
+    const coreConfigChains = warpCoreConfig.tokens.map((t) => t.chainName);
+
+    if (!setEquality(new Set(deployConfigChains), new Set(coreConfigChains))) {
```
nambrot, line 386: I assume this is to address the `Drive-by change` of https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5320. Imo this is different from what I wanted to specify though. I.e. we should still run through the check, its just that the check output doesn't mention the extra chains

typescript/cli/src/read/warp.ts
```diff
@@ -22,82 +24,83 @@ export async function runWarpRouteRead({
...
+    );
+    process.exit(1);
+  }
 
-  let addresses: ChainMap<string>;
-  if (symbol || warp) {
-    const warpCoreConfig =
-      context.warpCoreConfig ?? // this case is be handled by MultiChainHandler.forWarpCoreConfig() interceptor
-      (await getWarpCoreConfigOrExit({
+  const warpCoreConfig = hasWarpConfig
```
nambrot, line 44: Should this be imeplemented with similar logic to `getWarpConfigs`?

typescript/cli/src/utils/warp.ts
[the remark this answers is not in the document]
nambrot: So this is different from the `getWarpConfigs function now?

typescript/cli/src/read/warp.ts
```diff
@@ -107,14 +110,47 @@ export async function runWarpRouteRead({
...
-  const config = await promiseObjAll(
-    objMap(addresses, async (chain, address) =>
-      new EvmERC20WarpRouteReader(multiProvider, chain).deriveWarpRouteConfig(
-        address,
-      ),
-    ),
+/**
+ * Logs XERC20 token limits for the given warp core config
+ */
+export async function logXerc20Limits(
```
nambrot, line 118: Is this change from this PR or a different one?

typescript/cli/src/read/warp.ts
mshojaei-txfusion: what do you mean by similar logic? if you mean the way we can specify ID and can select between all routes, I would say better be another PR focus on `warp read` improvement after this PR merged.

typescript/cli/src/read/warp.ts
mshojaei-txfusion: Refactored and simplified as much as possible here on this PR.

typescript/cli/src/utils/warp.ts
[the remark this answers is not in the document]
mshojaei-txfusion: "we'll want to simplify this" did you mean `getWarpConfigs`?
I assumed you meant `getWarpCoreConfigOrExit ` as you said not your fault.

typescript/cli/src/commands/warp.ts
mshojaei-txfusion: As the other one was wc I made it wd, dose that make sense we select `wdConfig` that is longer than `config` itself?

typescript/cli/src/commands/warp.ts
mshojaei-txfusion: Got it now, good idea, changed it that way now

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -48,8 +47,6 @@ export class MultiChainResolver implements ChainResolver {
     switch (this.mode) {
       case ChainSelectionMode.WARP_CONFIG:
         return this.resolveWarpRouteConfigChains(argv);
-      case ChainSelectionMode.WARP_READ:
```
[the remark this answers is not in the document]
mshojaei-txfusion: Yes this is true, will mention it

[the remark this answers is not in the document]
mshojaei-txfusion: And this PR refactor kind of made it possible so I decided to do it here.

mshojaei-txfusion:
@antigremlin please let me know if any further clarification needed on drive-by

antigremlin:
> @antigremlin please let me know if any further clarification needed on drive-by

Oh wow, thanks for adding so many details. I wasn't expecting that much, TBH. Anyway, some items clearly describe the main change, and some are drive-by. I've moved a few into the top section, as an example. Please see if any others should be moved.

antigremlin (commented):
Looks good but I don't fully understand the code here.

@xeno097 could you have a look, please?

typescript/cli/src/commands/warp.ts
```diff
@@ -358,35 +365,37 @@ export const check: CommandModuleWithContext<{
...
+    // Get on-chain config
+    const onChainWarpConfig = await getWarpRouteConfigsByCore({
+      context,
+      warpCoreConfig,
+    });
 
     const expandedWarpDeployConfig = await expandWarpDeployConfig(
       context.multiProvider,
-      warpRouteConfig,
+      warpDeployConfig as WarpRouteDeployConfigMailboxRequired,
```
xeno097, line 398: Isn't this cast a bit risky here? What happens if any other function starts relying on the mailbox property, but the `warpDeployConfig` does not have it? Shouldn't we parse the configuration first so we can at least validate that the type matches? 


```typescript
    const deployConfigWithMailbox =
      WarpRouteDeployConfigMailboxRequiredSchema.parse(warpDeployConfig);
    const expandedWarpDeployConfig = await expandWarpDeployConfig(
      context.multiProvider,
      deployConfigWithMailbox,
      getRouterAddressesFromWarpCoreConfig(warpCoreConfig),
    );
```

typescript/cli/src/utils/warp.ts
```diff
@@ -33,3 +45,163 @@ export async function getWarpCoreConfigOrExit({
...
 }
+
+/**
+ * Gets both warp configs based on the provided inputs. Handles all cases:
+ * - warpRouteId: gets configs directly from registry
+ * - warpDeployConfigPath & warpCoreConfigPath: reads from files
+ * - symbol: prompts user to select from matching routes
+ * - no inputs: prompts user to search and select from all routes
+ */
+export async function getWarpConfigs({
```
xeno097, line 56: Is this going to replace the`getWarpCoreConfigOrExit` or are they going to be merged? because I see a lot of overlap between the 2

typescript/cli/src/commands/warp.ts
mshojaei-txfusion: Good catch, I taught this wont be a problem as `expandWarpDeployConfig` wont care about `mailbox`,
but yes you are right and it would effect final result type, and this made me think about further refactor and simplification here.
resolved!

typescript/cli/src/utils/warp.ts
mshojaei-txfusion: This is right I may use it there as well, but will do it on another PR, as we need some features on `thegetWarpCoreConfigOrExit ` and also `thegetWarpCoreConfigOrExit ` is used on different commands.

mshojaei-txfusion: related issues:
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5747
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5922

typescript/cli/src/utils/warp.ts
```diff
@@ -33,3 +50,148 @@ export async function getWarpCoreConfigOrExit({
...
+    const warpCoreConfig = readWarpCoreConfig(warpCoreConfigPath);
+    return { warpDeployConfig, warpCoreConfig };
+  }
+
+  let selectedId = warpRouteId;
+  if (!selectedId) {
+    const { ids: routeIds } = filterWarpRoutesIds(
+      (await context.registry.listRegistryContent()).deployments.warpRoutes,
+      symbol ? { symbol } : undefined,
+    );
```
xeno097, line 96: I think you can simplify your code by using the `getWarpRoutes` method and passing as filter the token symbol instead of having to do this process manually

```typescript
if (!selectedId) {
    const warpRoutes = await context.registry.getWarpRoutes({ symbol });
    const routeIds = Object.keys(warpRoutes);
    ...
```

mshojaei-txfusion: Good point, but this simplification would make it much slower as it would call `readConfigs` for each route and this would take time for the Github registry routes.

ltyu: agree that we should not hit the Registry too many times considering Github registry rate limits

typescript/cli/src/read/warp.ts
```diff
@@ -22,82 +24,83 @@ export async function runWarpRouteRead({
...
+  warpCoreConfig: WarpCoreConfig;
+}): Promise<ChainMap<HypTokenRouterConfig>> {
+  const addresses = Object.fromEntries(
+    warpCoreConfig.tokens.map((t) => [t.chainName, t.addressOrDenom!]),
+  );
+
+  return deriveWarpRouteConfigs(context, addresses, warpCoreConfig);
+}
+
+async function deriveWarpRouteConfigs(
```
antigremlin, line 74: A better name would be `deriveWarpDeployConfigs` because what we read from the `EvmERC20WarpRouteReader` is a deployment config.

In general, we should steer our naming towards a smaller set of terms, avoiding synonyms. My understanding is that we want to call the two types of warp route configs as "core" and "deploy" -- please correct me if I'm wrong here.

antigremlin (approved):

typescript/cli/src/read/warp.ts
```diff
@@ -22,82 +24,83 @@ export async function runWarpRouteRead({
...
-    addresses = {
-      [chain]: address,
-    };
-  } else {
-    logRed(`Please specify either a symbol, chain and address or warp file`);
-    process.exit(1);
+  return deriveWarpRouteConfigs(context, addresses, warpCoreConfig);
+}
+
+export async function getWarpRouteConfigsByCore({
```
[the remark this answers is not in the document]
mshojaei-txfusion: Agree config naming in CLI Are confusing, but `WarpRouteConfig` is different than `WarpDeployConfig` and I think this renaming maybe confusing again.

Reference:

image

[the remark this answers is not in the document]
mshojaei-txfusion: Yes, I agree, as it is right now is really unclear.
Maybe @ltyu  can help us to understand why jrossy did this that time.
so we can create an issue for it.

mshojaei-txfusion:
If all good we can merge this as other PRs depends on it.
@antigremlin @ltyu @xeno097

antigremlin (approved):

typescript/cli/src/utils/warp.ts
```diff
@@ -33,3 +50,147 @@ export async function getWarpCoreConfigOrExit({
...
+    if (matchingChains.size === 0) {
+      logRed('Error: No matching chains found between configs');
+      process.exit(1);
+    }
+
+    logRed(
+      `Continuing with check for matching chains: ${Array.from(
+        matchingChains,
+      ).join(', ')}\n`,
+    );
```
xeno097, line 175: Just a small nit on my side, but I would prefer the command to exit if there are mismatches between the input files, as only reading the matching chains might silently hide user errors that might end up in a wrong check result too

xeno097 (approved):

PR #5399, opened 2025-02-07
refactor: add shared refund logic to post dispatch hooks

yorhodes:
### Description

Authorized message ID hook quoted after sending unnecessarily. Introduces shared refund logic for this, protocol fee hook, and others in future.

### Drive-By Changes

Bumps forge-std libs and fixes false positive expect reverts.

### Backward compatibility

Yes

### Testing

Unit Tests


solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
```diff
@@ -88,9 +88,11 @@ abstract contract AbstractMessageIdAuthHook is
             "AbstractMessageIdAuthHook: msgValue must be less than 2 ** 255"
         );
 
-        _sendMessageId(metadata, message);
+        uint256 quote = _quoteDispatch(metadata, message);
```
nambrot, line 91: I forget our convention, but should it be `_quote`?

solidity/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@hyperlane-xyz/core",
   "description": "Core solidity contracts for Hyperlane",
-  "version": "5.11.1",
+  "version": "5.11.2",
   "dependencies": {
     "@arbitrum/nitro-contracts": "^1.2.1",
     "@eth-optimism/contracts": "^0.6.0",
-    "@hyperlane-xyz/utils": "8.5.0",
+    "@hyperlane-xyz/utils": "8.6.0",
```
nambrot, line 8: Why is this required?

yorhodes: its not, just merge fail

solidity/test/isms/MultisigIsm.t.sol
```diff
@@ -187,7 +187,8 @@ abstract contract AbstractMultisigIsmTest is Test {
         // changing single byte in metadata should fail signature verification
         uint256 index = uint256(seed) % metadata.length;
         metadata[index] = ~metadata[index];
-        assertFalse(ism.verify(metadata, message));
+        vm.expectRevert();
```
nambrot, line 190: Is this intended? I dont see why this is now expected to revert when returning false previously?

nambrot: Actually generally seems like an unrelated change?

nambrot (approved):

solidity/contracts/hooks/libs/AbstractMessageIdAuthHook.sol
```diff
@@ -90,17 +90,7 @@ abstract contract AbstractMessageIdAuthHook is
...
-        if (_overpayment > 0) {
-            address _refundAddress = metadata.refundAddress(
-                message.sender().bytes32ToAddress()
-            );
-            require(
-                _refundAddress != address(0),
-                "AbstractPostDispatchHook: no refund address"
-            );
-            payable(_refundAddress).sendValue(_overpayment);
-        }
```
ameten, line 103: I wonder if it is intentional to replace `_overpayment` with `address(this).balance`?

yorhodes: good question!
These hooks are not intended to custody funds but simply to forward fees of external bridges (in contrast to hooks like the IGP or protocol fee hook). There is no functionality to move value out of this contract and thus leaving any balance on it would be lost.

ameten: Ok. I see.

Our assumption here are as follows:
1. nobody can fund the smart contract since it does not have either `receive` or `fallback` function (true for contracts in our repo, although I suppose that `receive` and `fallback` functions can be added in inheriting contracts)
2. 1. means that before this function (or transaction with it) is being executed, `address(this).balance` is zero (assuming that smart contract was deployed with zero balance, did not receive funds from coinbase transactions or from other contracts during destruction)
3. `address(this).balance` is increased by `msg.value` before this function is executed (true for Ethereum).
4. some balance from `msg.value` was sent further in actual implementation of `_sendMessageId` and this balance is equal to `_quoteDispatch(metadata, message)`.

Under these assumptions `address(this).balance` is actually the same as `msg.value - _quoteDispatch(metadata, message)`.

If these assumptions are not true, we are okay to give back more (or less) than `msg.value - _quoteDispatch(metadata, message)` to the sender of the message.

Issue #5398, opened 2025-02-07
chore(warp speed): Import getRegistry from the CLI


ltyu:
## Problem

Could we just import the CLI function here and then call this with `primaryRegistryUri` from `DEFAULT_REGISTRY_URI`? Why introduce another ENV var and this logic?

_Originally posted by @nambrot in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5373#discussion_r1942053181_

PR #5400, opened 2025-02-07
feat: infra checker use registry

nambrot:
### Description

uses the CLI's getRegistry function to get an `IRegistry` from registry URLs

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5398


typescript/infra/config/warp.ts
```diff
@@ -1,3 +1,4 @@
+import { getRegistry } from '@hyperlane-xyz/cli';
```
paulbalaji, line 1: why is this command in the cli and not the sdk itself if we're reusing it across diff places

not a blocker for this PR but wondering for future

typescript/infra/package.json
paulbalaji, line 1: pull latest main, we just pushed a patch release so this should be 8.6.1 now

paulbalaji (commented):
lgtm once you pull latest main

ltyu (approved):

typescript/infra/config/warp.ts
nambrot: Good point yes. Should this even be in the SDK, but registry package, no?

ltyu: sdk shouldn't use FileSystem because the UI uses it.

PR #5401, opened 2025-02-07
feat: graceful and alertable relayer reorg detection

tkporter:
### Description

- Only panics in the validator if there is a reorg detected
- Mostly solves https://github.com/hyperlane-xyz/issues/issues/1394 (does not have super clear alert conditions for a reorg occurring, this should come later)
- Whereas before we would use `Result<Option<Vec<u8>>` as a return type for metadata building, where the Result indicates some error occurred (e.g. an RPC issue or something), Ok(Some(_)) meant that metadata building was successful, and Ok(None) meant that no error occurred but simply that fetching metadata wasn't successful -- we now have a new `Metadata` enum:
```
#[derive(Clone, Debug)]
pub enum Metadata {
    /// Able to fetch metadata
    Ok(Vec<u8>),
    /// Unable to fetch metadata, but no error occurred
    CouldNotFetch,
    /// While building metadata, encountered something that should
    /// prohibit all metadata for the message from being built.
    /// Provides the reason for the refusal.
    Refused(String),
}
```
- When we are building metadata and encounter a reorg flag in a validator that's used by the message, we handle this and return `Metadata::MetadataBuildingRefused`. The aggregation ISM will propagate this to make sure that we have end behavior of flatly refusing to build metadata for a message even if a nested metadata builder encounters this. If this happens, a new reprepare reason `MessageMetadataRefused` is introduced that's used.
- e2e test is added to confirm this behavior

### Testing

e2e, some unit

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
@@ -137,9 +137,7 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
             .iter()
             .find_map(|result| match result {
                 Ok(sub_module_and_meta) => match &sub_module_and_meta.metadata {
-                    Metadata::MetadataBuildingRefused(reason) => {
-                        Some(Metadata::MetadataBuildingRefused(reason.clone()))
-                    }
+                    Metadata::Refused(reason) => Some(Metadata::Refused(reason.clone())),
                     _ => None,
                 },
                 Err(_) => None,
--- a/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
@@ -9,7 +9,7 @@ use tracing::{info, instrument};
 
 use hyperlane_core::{HyperlaneMessage, InterchainSecurityModule, ModuleType, H256, U256};
 
-use super::{MessageMetadataBuilder, MetadataBuilder};
+use super::{MessageMetadataBuilder, Metadata, MetadataBuilder};
 
 /// Bytes used to store one member of the (start, end) range tuple
 /// Copied from `AggregationIsmMetadata.sol`
@@ -119,11 +119,7 @@ impl AggregationIsmMetadataBuilder {
 impl MetadataBuilder for AggregationIsmMetadataBuilder {
     #[instrument(err, skip(self, message), ret)]
     #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
-    async fn build(
-        &self,
-        ism_address: H256,
-        message: &HyperlaneMessage,
-    ) -> eyre::Result<Option<Vec<u8>>> {
+    async fn build(&self, ism_address: H256, message: &HyperlaneMessage) -> eyre::Result<Metadata> {
         const CTX: &str = "When fetching AggregationIsm metadata";
         let ism = self.build_aggregation_ism(ism_address).await.context(CTX)?;
         let (ism_addresses, threshold) = ism.modules_and_threshold(message).await.context(CTX)?;
@@ -136,6 +132,22 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
         )
         .await;
 
+        // If any inner ISMs are refusing to build metadata, we propagate just the first refusal.
+        if let Some(first_refusal) = sub_modules_and_metas
+            .iter()
+            .find_map(|result| match result {
+                Ok(sub_module_and_meta) => match &sub_module_and_meta.metadata {
+                    Metadata::MetadataBuildingRefused(reason) => {
+                        Some(Metadata::MetadataBuildingRefused(reason.clone()))
+                    }
+                    _ => None,
+                },
+                Err(_) => None,
+            })
+        {
+            return Ok(first_refusal);
+        }
+
         // Partitions things into
         // 1. ok_sub_modules: ISMs with metadata with valid metadata
         // 2. err_sub_modules: ISMs with invalid metadata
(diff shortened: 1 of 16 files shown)


rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -44,17 +47,27 @@ pub enum MetadataBuilderError {
     MaxDepthExceeded(u32),
 }
 
+#[derive(Clone, Debug)]
+pub enum Metadata {
+    /// Able to fetch metadata
+    Ok(Vec<u8>),
```
tkporter, line 53: open to `Some` or another name

rust/main/hyperlane-base/src/settings/checkpoint_syncer.rs
```diff
@@ -225,28 +233,12 @@ mod test {
...
-    ),
-}. Please resolve the reorg to continue."#
-            );
-        } else {
-            panic!(
-                "Caught panic has a different type than the expected one (`String`): {:?}",
-                startup_result
-            );
+        let result = checkpoint_syncer_conf.build_and_validate(None).await;
+        match result {
```
tkporter, line 237: not sure if there's a better way to test for equality tbh given I can't have `CheckpointSyncerBuildError` be partialeq with the eyre::Report, this feels ugly

rust/main/agents/relayer/src/msg/metadata/base.rs
ameten: `Success`

rust/main/utils/run-locally/src/main.rs
```diff
@@ -460,6 +462,21 @@ fn main() -> ExitCode {
         return report_test_result(test_passed);
     }
 
+    // Kill validator 1 and simulate a reorg, which we'll later use
+    // to ensure the relayer handles reorgs correctly.
+    stop_validator(&mut state, 1);
```
daniel-savu, line 467: why does the validator need to be stopped?

rust/main/utils/run-locally/src/main.rs
```diff
@@ -571,6 +616,25 @@ fn restart_relayer(config: &Config, state: &mut State, rocks_db_dir: &TempDir) {
     log!("Restarted relayer...");
 }
 
+fn relayer_reorg_handling_invariants_met() -> eyre::Result<bool> {
+    let lengths = fetch_metric(
```
daniel-savu, line 620: nit: rename from `lengths` to `refused_messages`?

daniel-savu (approved):

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -136,6 +132,20 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
         )
         .await;
 
+        // If any inner ISMs are refusing to build metadata, we propagate just the first refusal.
+        if let Some(first_refusal) = sub_modules_and_metas
```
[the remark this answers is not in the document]
tkporter: Yeah this was intentional, if one refuses I think we should fully refuse the whole thing. We can always change later but imo this is the most conservative and with little cost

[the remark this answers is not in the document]
daniel-savu: it's a good point

on the one hand, you can argue that stopping that the first refusal violates the spirit of the aggregation ISM, where as long as one of the ISMs does build in the 1/2 example, the message should be valid

on the other hand you have instances like the renzo ISM where it's an aggregation of multisigs where the validators are reputable operators, so it's unlikely they'd post a reorg flag by accident. we do risk false positives even in this case, as was the case with blockhunter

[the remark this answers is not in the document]
daniel-savu: will still merge as is but I opened https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5420 to track this

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -44,17 +47,28 @@ pub enum MetadataBuilderError {
...
+    /// Unable to fetch metadata, but no error occurred
+    CouldNotFetch,
```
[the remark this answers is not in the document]
daniel-savu: `CouldNotFetch` is used in other parts of the codebase so I think it's consistent. Example: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/fa15cd634be80e3986fd27cadee82b5c20f2f2f9/rust/main/hyperlane-core/src/traits/pending_operation.rs#L235

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -112,30 +109,41 @@ impl<T: MultisigIsmMetadataBuilder> MetadataBuilder for T {
...
-        let checkpoint_syncer = self
+        let checkpoint_syncer = match self
             .as_ref()
             .build_checkpoint_syncer(message, &validators, self.as_ref().app_context.clone())
             .await
-            .context(CTX)?;
+        {
+            Ok(syncer) => syncer,
+            Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event)) => {
+                return Ok(Metadata::Refused(format!(
```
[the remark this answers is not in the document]
daniel-savu: tracked in https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5420 but I think it's ok to merge as is for now

rust/main/utils/run-locally/src/main.rs
daniel-savu: ah I see, the reorg check is only done at startup time in the validator but it's ok to kill it to avoid future flakes. Will include this context in the comment

rust/main/agents/relayer/src/msg/metadata/base.rs
daniel-savu: went for `Found` since the existence of metadata doesn't guarantee it's successfull (e.g. non-malicious)

PR #5404, opened 2025-02-07
feat(cli): optionally check ism and hook config if provided in input config

xeno097:
### Description

This PR updates the `warp check` command to optionally check the ism and hook config if they are provided in the input file

### Drive-by changes

- NO

### Backward compatibility

- YES

### Testing

- E2E

### DIFF
--- a/typescript/cli/src/check/warp.ts
+++ b/typescript/cli/src/check/warp.ts
@@ -21,13 +21,13 @@ export async function runWarpRouteCheck({
   // Go through each chain and only add to the output the chains that have mismatches
   const [violations, isInvalid] = Object.keys(warpRouteConfig).reduce(
     (acc, chain) => {
-      const formattedOnchainConfig = formatCheckerInput({
+      formatCheckerInput({
         warpRouteConfig: warpRouteConfig[chain],
         onChainWarpConfig: onChainWarpConfig[chain],
       });
 
       const { mergedObject, isInvalid } = diffObjMerge(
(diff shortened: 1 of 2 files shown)


typescript/cli/src/check/warp.ts
```diff
@@ -38,3 +48,40 @@ export async function runWarpRouteCheck({
...
+  // If the hook config is not defined in the input file,
+  // we need to remove it from the onChainWarpConfig if it was derived
+  if (!warpRouteConfig.hook) {
+    onChainWarpConfig.hook = undefined;
+  }
+
+  // if the hook config is defined the input file, it means the user wants to check
+  // the hook config, so we need to add the default hook address to the onChainWarpConfig
+  // in case the default hook is currently used by the token.
+  if (warpRouteConfig.hook && !onChainWarpConfig.hook) {
```
nambrot, line 68: I'm not sure I understand this case

typescript/cli/src/check/warp.ts
```diff
@@ -38,3 +48,40 @@ export async function runWarpRouteCheck({
...
+    onChainWarpConfig.hook = constants.AddressZero;
+  }
+
+  // If the ism config is not defined in the input file,
+  // we need to remove it from the onChainWarpConfig if it was derived
```
nambrot, line 73: Why do we remove it? it the ISM is supposed to be the default, but the actualy ISM derived is something else, sholdn't we show that?

PR #5410, opened 2025-02-08
fix: don't derive testnet domains in IGP config derivation on mainnet

nambrot:
### Description

Right now, because IGP's domains are not an enumerable set, we have to query all known domains on it. That is becoming very expensive. This change at least avoids querying for testnet domains when looking at a mainnet IGP and vice versa.


paulbalaji (approved):
pls add changeset, then lgtm

PR #5412, opened 2025-02-09
feat: HyperlaneCCIPDeployer

paulbalaji:
### Description

- Introduce a `HyperlaneCCIPDeployer` that deploys the CCIPIsm/CCIPHook pairings for a set of chains.
	- Automatically recovers ISMs/Hooks for already deployed (origin->destination) pairs.
	- e.g. if we're on op,base,arb and decide to add blast, it'll only deploy the delta for adding blast. see example here: https://app.warp.dev/block/no2GBosdDs0IsZ58usAE8u
- Introduce CCIPContractCache for easy write/read of the relevant contracts.
- Update hook/ism deployers to read from the cached contracts instead of deploying their own.
	- Treating each (origin->destination) Hook/ISM pair as a singleton.
- Add a schema check to ensure a warp config with CCIP ism/hook as the corresponding hook/ism.

### Related issues

https://www.notion.so/hyperlanexyz/CCIP-Hook-ISM-Deployer-1946d35200d680cfab36e8bf58548651


typescript/infra/scripts/deploy.ts
```diff
@@ -241,6 +250,16 @@ async function main() {
...
+  } else if (module === Modules.CCIP) {
+    if (environment !== 'mainnet3') {
+      throw new Error('CCIP is only supported on mainnet3');
+    }
+    config = getCCIPDeployConfig(filteredTargetNetworks);
+    deployer = new HyperlaneCCIPDeployer(
+      multiProvider,
+      getEnvAddresses(environment),
+      contractVerifier,
+    );
```
paulbalaji, line 262: with the new design, the intention is that these hooks/isms are reused for each (origin,destination) pair. As we don't really want users to deploy these themselves (they'll just use the lookups), I kept the config generation + intended deployer usage in infra

typescript/sdk/src/utils/ccip.ts
```diff
@@ -33,3 +37,71 @@ export function getCCIPChainSelector(chainName: string): string | undefined {
...
+  }
+
+  writeBack(cachedAddresses: HyperlaneAddressesMap<any>): void {
+    for (const [origin, destinations] of Object.entries(this.cachedAddresses)) {
+      if (!cachedAddresses[origin]) {
+        cachedAddresses[origin] = {};
+      }
+      for (const [key, address] of Object.entries(destinations)) {
+        cachedAddresses[origin][key] = address;
+      }
```
paulbalaji, line 76: only want to write-back keys with the ccip hook/ism prefix

PR #5416, opened 2025-02-10
fix: try using a longer randomization period for retrying messages

daniel-savu:
### Description

We have so many messages with similar retry counts that now even by spreading them over 1h we still starve new messages. This PR increases the random period to 6h, hoping to more evenly distribute retries of old messages.

A side effect of this is that even newly indexed messages may take 6h to get delivered after a restart

Some data to back this, you can see that after prepare failures start spiking there's also a diff in the prep queue, indicating starvation
Screenshot_2025-02-10_at_12 03 53
Screenshot_2025-02-10_at_12 03 44
Screenshot_2025-02-10_at_12 03 18

### DIFF
--- a/rust/main/agents/relayer/src/msg/pending_message.rs
+++ b/rust/main/agents/relayer/src/msg/pending_message.rs
@@ -657,10 +657,10 @@ impl PendingMessage {
                 let hour: u64 = 60 * 60;
                 // To be extra safe, `max` to make sure it's at least 1 hour.
                 let target = hour.max((num_retries - 47) as u64 * hour);
-                // Schedule it at some random point in the next hour to
+                // Schedule it at some random point in the next 6 hours to
                 // avoid scheduling messages with the same # of retries
-                // at the exact same time.
-                target + (rand::random::<u64>() % hour)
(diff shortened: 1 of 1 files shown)


ameten (approved):

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -657,10 +657,10 @@ impl PendingMessage {
...
                 // To be extra safe, `max` to make sure it's at least 1 hour.
                 let target = hour.max((num_retries - 47) as u64 * hour);
-                // Schedule it at some random point in the next hour to
+                // Schedule it at some random point in the next 6 hours to
                 // avoid scheduling messages with the same # of retries
-                // at the exact same time.
-                target + (rand::random::<u64>() % hour)
+                // at the exact same time and starve new messages.
+                target + (rand::random::<u64>() % (6 * hour))
             }
```
paulbalaji, line 664: why not do the 6* at the `let hour:` and rename that to something like reschedule interval

kamiyaa (approved):

PR #5418, opened 2025-02-10
feat(sealevel): Support claim tx via SealevelIgpAdapter

Mo-Hussain:
### Description

- add `populateClaimTx` implementation to the `SealevelIgpAdapter`
- support priority fee config for sealevel chains
- This is a step towards supporting sealevel chains on the key funder

### Testing

Manual

### DIFF
--- a/typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts
+++ b/typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts
@@ -191,7 +191,7 @@ export class SealevelIgpAdapter extends SealevelIgpProgramAdapter {
   }
 
   /**
-   * Constructs a TransactionInstruction for the Claim instruction.
+   * Constructs a Transaction for .
    * @param {PublicKey} beneficiary - The IGP’s configured beneficiary.
    * @returns {Promise<TransactionInstruction>} The claim instruction.
    */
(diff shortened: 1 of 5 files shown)


typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts
```diff
@@ -158,6 +162,59 @@ export class SealevelIgpAdapter extends SealevelIgpProgramAdapter {
...
+    // check if the passed in beneficiary is same as stored beneficiary
+    if (!igpData.beneficiary_pub_key.equals(beneficiary)) {
+      throw new Error('Beneficiary account mismatch');
+    }
+
+    const keys = this.getClaimInstructionKeyList(beneficiary);
+
+    const data = Buffer.from([SealeveIgpInstruction.Claim]);
+
+    const claimIgpInstruction = new TransactionInstruction({
```
Mo-Hussain, line 210: Would you recommend we set a compute unit price using the median propriety fee paid to tx related to the igp account? @tkporter

tkporter: Yeah I think we should set a priority fee on Solana, but the other SVM chains no need

Figuring out priority fees is super tricky on SVM, the `getRecentPrioritizationFees` RPC is generally insufficient and the Helius API or quicknode API (https://www.quicknode.com/gas-tracker/solana) is better, but instead of doing that I'd recommend for now just setting a hardcoded prio fee for solanamainnet and accept that we may overpay. It should ideally be infrequent and we can always adjust later. I'd recommend like 200k or 500k microlamports?

paulbalaji (approved):

PR #5424, opened 2025-02-10
fix: HookModule update deploys due to unnormalized config

ltyu:
### Description
This PR fixes an issue with HookModule that causes HookModule to trigger a new deployment due to unnormalized config

### Backward compatibility
Yes

### Testing
Unit Tests


paulbalaji (approved):

nambrot:
Agree with @paulbalaji

PR #5429, opened 2025-02-11
feat: implement HyperlaneCCIPDeployer

paulbalaji:
### Description

feat: implement HyperlaneCCIPDeployer
- implement `HyperlaneCCIPDeployer` for deploying and initializing CCIP ISM/Hooks for pairs of supported CCIP chains.
- implement `CCIPContractCache` for wrapping writing/reading of contract addresses for given CCIP (origin,destination) pairs.
- set up config generation and deployment of the hook/ism pairs via infra.
- add `ZERO_ADDRESS_HEX_32` to constants.

### DIFF
--- a/typescript/sdk/src/ccip/HyperlaneCCIPDeployer.ts
+++ b/typescript/sdk/src/ccip/HyperlaneCCIPDeployer.ts
@@ -1,0 +1,246 @@
+import { CCIPHook__factory, CCIPIsm__factory } from '@hyperlane-xyz/core';
+import {
+  Address,
+  ZERO_ADDRESS_HEX_32,
+  addressToBytes32,
+  assert,
+  rootLogger,
+} from '@hyperlane-xyz/utils';
+
+import {
+  HyperlaneAddressesMap,
+  HyperlaneContracts,
+} from '../contracts/types.js';
+import { CoreAddresses } from '../core/contracts.js';
+import { HyperlaneDeployer } from '../deploy/HyperlaneDeployer.js';
+import { ContractVerifier } from '../deploy/verify/ContractVerifier.js';
+import { MultiProvider } from '../providers/MultiProvider.js';
+import { ChainMap, ChainName } from '../types.js';
+
+import {
+  CCIPContractCache,
+  getCCIPChainSelector,
+  getCCIPRouterAddress,
+} from './utils.js';
+
+export class HyperlaneCCIPDeployer extends HyperlaneDeployer<
+  Set<ChainName>,
+  {}
+> {
+  private ccipContractCache: CCIPContractCache = new CCIPContractCache();
+
+  constructor(
+    multiProvider: MultiProvider,
+    readonly core: ChainMap<Partial<CoreAddresses>>,
+    contractVerifier?: ContractVerifier,
+  ) {
+    super(
+      multiProvider,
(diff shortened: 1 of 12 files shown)


typescript/sdk/src/ccip/utils.ts
```diff
@@ -41,3 +46,63 @@ export function getCCIPRouterAddress(chainName: string): string | undefined {
...
+        cachedAddresses[origin][key] = address;
+      }
+    }
+  }
+
+  setHook(origin: ChainName, destination: ChainName, ccipHook: CCIPHook): void {
+    if (!this.cachedAddresses[origin]) {
+      this.cachedAddresses[origin] = {};
+    }
+    this.cachedAddresses[origin][`${CCIP_HOOK_KEY_PREFIX}-${destination}`] =
```
nambrot, line 85: Nit: Can we use `_` instead of `-` so that JSON keys are simpler?

typescript/sdk/src/ccip/HyperlaneCCIPDeployer.ts
```diff
@@ -0,0 +1,246 @@
...
+    //On the origin chain, deploy hooks for each destination chain in series
+    for (const destination of config) {
+      // Grab the ISM from the cache
+      const ccipIsmAddress = this.ccipContractCache.getIsm(origin, destination);
+      assert(
+        ccipIsmAddress,
+        `CCIP ISM not found for ${origin} -> ${destination}`,
+      );
+
+      await this.deployCCIPHook(origin, destination, ccipIsmAddress);
```
nambrot, line 73: Why not make this part of the promise above?, i.e. why split this up over multiple loops of destinations?

typescript/sdk/src/ccip/HyperlaneCCIPDeployer.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // Authorize hooks for each destination chain concurrently
+    await Promise.all(
+      Array.from(config).map(async (destination) => {
```
nambrot, line 78: Ditto

typescript/sdk/src/ccip/HyperlaneCCIPDeployer.ts
```diff
@@ -0,0 +1,246 @@
...
+  ): Promise<HyperlaneContracts<{}>> {
+    // Deploy ISMs from chain to eachdestination chain concurrently
+    await Promise.all(
+      Array.from(config).map(async (destination) => {
+        // Deploy CCIP ISM for this origin->destination pair
+        await this.deployCCIPIsm(origin, destination);
+      }),
+    );
+
+    //On the origin chain, deploy hooks for each destination chain in series
```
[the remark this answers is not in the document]
paulbalaji: I was worried about nonce contention. We deploy ISMs across distinct destination chains so it can be concurrent, however all the hooks will be deployed on the same origin chain, so I kept that serial. Which is why once we go back to authorizing the hooks on each destination ISM, we can do that concurrently again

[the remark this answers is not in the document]
nambrot: Great answer

nambrot (approved):

PR #5432, opened 2025-02-11
feat: enable usage of CCIP Hooks and ISMs in warp routes

paulbalaji:
### Description

feat: enable usage of CCIP Hooks and ISMs in warp routes
- use CCIPContractCache in ism/hook deployers
- pass CCIPContractCache through from registry down to deployers
- add example ccip warp route configs
- add zod validation for warp route configs to verify that CCIP hooks are paired with CCIP ISMs


typescript/cli/src/deploy/warp.ts
```diff
@@ -243,13 +244,16 @@ async function resolveWarpIsmAndHook(
 ): Promise<WarpRouteDeployConfig> {
   return promiseObjAll(
     objMap(warpConfig, async (chain, config) => {
-      const chainAddresses = await context.registry.getChainAddresses(chain);
+      const registryAddresses = await context.registry.getAddresses();
+      const ccipContractCache = new CCIPContractCache(registryAddresses);
```
nambrot, line 248: We did obviously discuss this, but still is tough to read. We'll have to come back and clean this up later

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -64,6 +65,7 @@ export class EvmIsmModule extends HyperlaneModule<
       IsmConfig,
       HyperlaneAddresses<ProxyFactoryFactories> & IsmModuleAddresses
     >,
+    ccipContractCache?: CCIPContractCache,
```
nambrot, line 68: Probably a premature optimization, but generally like capturing somewhat extra arguments like this one in an `options` parameter in case we accumulate more of these

typescript/sdk/src/token/types.ts
```diff
@@ -166,3 +215,108 @@ function isCollateralRebasePairedCorrectly(
...
+
+  for (const [chainName, config] of Object.entries(warpRouteDeployConfig)) {
+    traverseHookConfig(chainName, config.hook, ccipHookMap);
+    traverseIsmConfig(chainName, config.interchainSecurityModule, ccipIsmMap);
+  }
+
+  return { ccipHookMap, ccipIsmMap };
+}
+
+function traverseHookConfig(
```
nambrot, line 238: This is another maybe not practical comment, but this as far as I can create a generic version of this (always forget if they are called functors or monads) and then express the extraction (i.e. the population of existsCCIPHookMap) as a generic function. Until then I would probably name this function specific to cciphook extraction

typescript/sdk/src/token/types.ts
```diff
@@ -166,3 +215,108 @@ function isCollateralRebasePairedCorrectly(
   );
   return allOthersSynthetic;
 }
+
+type CCIPContractExistsMap = ChainMap<ChainMap<boolean>>;
```
nambrot, line 219: Might be useful to have a comment on the intended population of the nested map, iiuc, its origin domain => destination domain => boolean exists

typescript/sdk/src/token/types.ts
```diff
@@ -147,6 +151,51 @@ export const WarpRouteDeployConfigSchema = z
...
+  })
+  // Verify that CCIP hooks are paired with CCIP ISMs
+  .transform((warpRouteDeployConfig, ctx) => {
+    // Get all the CCIP hooks and ISMs by origin->destination chain
+    const { ccipHookMap, ccipIsmMap } = getCCIPConfigMaps(
+      warpRouteDeployConfig,
+    );
+
+    // Check that each CCIP hook has a corresponding ISM
+    for (const [originChain, hooks] of Object.entries(ccipHookMap)) {
```
nambrot, line 163: IIuc, isn't `hooks` a map of destination chains to booleans? If so, calling it `hooks` feels off. Maybe `claimedDestinationsInHooks` or something like that?

typescript/sdk/src/token/types.ts
```diff
@@ -147,6 +151,51 @@ export const WarpRouteDeployConfigSchema = z
...
+      if (!ccipIsmMap[originChain]) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `No CCIP ISM found in config for origin chain ${originChain}`,
+        });
+        return z.NEVER;
+      }
+
+      for (const [destinationChain] of Object.entries(hooks)) {
```
nambrot, line 172: If you are only going to take the keys, imo use `Object.keys`? However since you also are taking the value, why not `[destination, doesExist]`. But then again, when would the `destinationChain` be in here, but the boolean false? Wonder if it would make it simpler to have the inner `ChainMap` as an array or set?

typescript/sdk/src/ism/EvmIsmModule.ts
paulbalaji: i think this would apply generally for a bunch of the deployers, since we could make the cache/verifier both part of the "options" arg

made a ticket to track https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5443

typescript/sdk/src/token/types.ts
paulbalaji: changed it from ChainMap<ChainMap<boolean>> into a ChainMap<Set<ChainName>> as suggested, also fixed naming as requested

typescript/cli/examples/ccip/multi-chain-ccip-warp-route.yaml
```diff
@@ -0,0 +1,81 @@
...
+      polygon:
+        type: ccipIsm
+        originChain: polygon
+      base:
+        type: ccipIsm
+        originChain: base
+  hook:
+    type: aggregationHook
+    hooks:
+      - '0x3b6044acd6767f017e99318AA6Ef93b7B06A5a22'
```
yorhodes, line 18: whats this?

typescript/sdk/src/token/types.ts
```diff
@@ -166,3 +222,109 @@ function isCollateralRebasePairedCorrectly(
...
+    case IsmType.OP_STACK:
+    case IsmType.STORAGE_MERKLE_ROOT_MULTISIG:
+    case IsmType.STORAGE_MESSAGE_ID_MULTISIG:
+    case IsmType.TEST_ISM:
+    case IsmType.TRUSTED_RELAYER:
+    case IsmType.WEIGHTED_MERKLE_ROOT_MULTISIG:
+    case IsmType.WEIGHTED_MESSAGE_ID_MULTISIG:
+      break;
+  }
+}
```
yorhodes, line 330: I believe this could be replaced with `deepFind` and not account for every enum variant

typescript/sdk/src/token/types.ts
```diff
@@ -147,7 +151,59 @@ export const WarpRouteDeployConfigSchema = z
...
+        return Array.from(validDestinationChains).some((chain) => {
+          if (!ccipIsmMap[originChain].has(chain)) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              path: [chain, 'interchainSecurityModule', '...'],
+              message: `Required CCIP ISM not found in config for CCIP Hook with origin chain ${originChain} and destination chain ${chain}`,
+            });
+            return true;
+          }
+          return false;
```
yorhodes, line 181: a bit confused how this is different from the above case

typescript/cli/examples/ccip/multi-chain-ccip-warp-route.yaml
paulbalaji: ah i think this is a leftover from before the default hook config - will replace with that

typescript/sdk/src/token/types.ts
paulbalaji: i got caught in two minds with the loops - resolved

nambrot (approved):

Issue #4738, opened 2024-10-23
Sealevel RPC failure / success metrics and alerts


tkporter:
## Problem

The `hyperlane_request_count` metrics have been useful in observing RPC failures for EVM chains. We don't have this for any other VM.

## Solution

Add success / failure count metrics when interacting with Sealevel RPCs. The abstractions probably aren't as nice as they are on the EVM tooling to do this.

Base criteria:
- should isolate the method of the RPC being done
- should increment the status of success / failure, just like how the EVM metrics work
- should give an indication as to what RPC url is responsible (stripping secrets)

Nice to haves:
- the same metric as we use on EVM, this way all our alerting / queries work out of the box. I think this should be possible because SVM also uses JSON RPC and we can isolate the method
- I think we should prefer to increment the metric when an RPC is actually made, but in practice due to the abstractions I could see it being easier for us to imply based off what function we're calling on the client which method to increment

tkporter:
closing as done

PR #5434, opened 2025-02-11
feat: Sealevel RPC failure / success metrics and alerts

kamiyaa:
### Description

 - Add sealevel metrics

### Drive-by changes
 - add some builders to certain structs

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4738

### Backward compatibility
Yes


rust/main/hyperlane-metric/src/sealevel/prometheus_sender.rs
```diff
@@ -0,0 +1,160 @@
...
+            "status" => if res.is_ok() { "success" } else { "failure" }
+        };
+        if let Some(counter) = &self.metrics.request_count {
+            counter.with(&labels).inc()
+        }
+        if let Some(counter) = &self.metrics.request_duration_seconds {
+            counter
+                .with(&labels)
+                .inc_by((Instant::now() - start).as_secs_f64())
+        };
```
daniel-savu, line 79: this logic looks like the one for instatiating the ethers metrics provider, maybe it can be refactored

rust/main/hyperlane-metric/src/sealevel/prometheus_sender.rs
```diff
@@ -0,0 +1,160 @@
...
+                }
+                Err(err) => Err(RpcError::RpcRequestError(format!(
+                    "Failed to deserialize RPC error response: {} [{}]",
+                    serde_json::to_string(&json["error"]).unwrap(),
+                    err
+                ))
+                .into()),
+            };
+        }
+        return Ok(json["result"].take());
```
daniel-savu, line 158: this looks pretty complicated, do we have to do this parsing and retrying manually here? My expectation is that we can wrap a provider that does all this for us, and we just collect usage metrics, like we do for the ethers provider

kamiyaa: Yeah initially I was planning to use HttpSender from solana-client, but its private. So I ended up copying most of the code over. I'll add documentation on where the code comes from

rust/main/hyperlane-metric/src/sealevel/prometheus_sender.rs
kamiyaa: Moved into JsonMetrics method :+1:

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -35,27 +36,14 @@ lazy_static! {
...
-        // Set the `processed` commitment at rpc level
-        let rpc_client = Arc::new(SealevelRpcClient::new(conf.url.to_string()));
-        let native_token = conf.native_token.clone();
-
-        Self {
-            domain,
-            rpc_client,
-            native_token,
-        }
-    }
```
daniel-savu, line 57: my preference is to keep `conf: &ConnectionConf` as a constructor argument so `native_token` can be read from it, rather than have the token passed explicitly - which feels like it'd break some encapsulation

also since `SealevelRpcClient` is not generic we can also instantiate it inside the constructor rather than have it passed as an argument, to further encapsulate

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
```diff
@@ -572,12 +571,11 @@ pub struct SealevelMailboxIndexer {
...
     /// Create a new SealevelMailboxIndexer
     pub fn new(
-        conf: &ConnectionConf,
-        locator: ContractLocator,
+        mailbox: SealevelMailbox,
+        locator: &ContractLocator,
         advanced_log_meta: bool,
     ) -> ChainResult<Self> {
         let program_id = Pubkey::from(<[u8; 32]>::from(locator.address));
-        let mailbox = SealevelMailbox::new(conf, locator, None)?;
```
daniel-savu, line 579: so i better understand, why is it preferable to pass the `mailbox` as an arg rather than build it in here? it looks like it adds some boilerplate in the three sites calling this

rust/main/chains/hyperlane-sealevel/src/tx_submitter.rs
```diff
@@ -25,23 +26,16 @@ pub trait TransactionSubmitter: Send + Sync {
...
-    rpc_client: SealevelRpcClient,
-}
-
-impl RpcTransactionSubmitter {
-    pub fn new(url: String) -> Self {
-        Self {
-            rpc_client: SealevelRpcClient::new(url),
-        }
-    }
+    pub rpc_client: SealevelRpcClient,
```
daniel-savu, line 38: does this need to be public?

rust/main/chains/hyperlane-sealevel/src/validator_announce.rs
```diff
@@ -21,8 +21,7 @@ pub struct SealevelValidatorAnnounce {
 
 impl SealevelValidatorAnnounce {
     /// Create a new Sealevel ValidatorAnnounce
-    pub fn new(conf: &ConnectionConf, locator: ContractLocator) -> Self {
-        let provider = SealevelProvider::new(locator.domain.clone(), conf);
+    pub fn new(provider: SealevelProvider, locator: &ContractLocator) -> Self {
```
daniel-savu, line 24: in here and other places, I think it'd result in less boilerplate to build `SealevelProvider` inside this constructor but open to pushback

daniel-savu (commented):
looks great, one more thing that's not in this PR yet and would be good to have before merging is an e2e (sealevel) check that the new metrics are indeed there. Since the metrics are the same in both evm and sealevel it may actually be easy to add this check to both (to then have it easily added to cosmos)

rust/main/chains/hyperlane-sealevel/src/rpc/client.rs
```diff
@@ -45,19 +45,14 @@ pub struct SealevelTxCostEstimate {
...
+pub struct SealevelRpcClient(pub RpcClient);
 
 impl SealevelRpcClient {
     /// The max amount of compute units for a transaction.
     const MAX_COMPUTE_UNITS: u32 = 1_400_000;
 
-    pub fn new(rpc_endpoint: String) -> Self {
-        Self(RpcClient::new_with_commitment(
-            rpc_endpoint,
-            CommitmentConfig::processed(),
```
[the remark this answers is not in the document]
kamiyaa: Yeah, that works too :+1:

rust/main/chains/hyperlane-sealevel/src/provider.rs
kamiyaa: I can update it to keep `conf: &ConnectionConf`, but for `SealevelRpcClient` I'm not sure what you mean?
I pass it in this way so we can initialize `RpcClient` and `SealevelRpcClient` with metrics outside the constructor, so we don't need to pass metrics around everywhere.

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
kamiyaa: To me, it makes more sense to move more of the logic to a higher level, where we can do all the configurations, metrics, etc we need. Then injecting it into the primitives we use.

In the future, if we ever want to reduce the amount of providers/rpc_clients we instantiate, then it makes sense to move this logic out and pass providers/rpc_clients into these structs instead.

rust/main/chains/hyperlane-sealevel/src/tx_submitter.rs
kamiyaa: no :+1:

rust/main/chains/hyperlane-sealevel/src/validator_announce.rs
kamiyaa: If we instantiate `SealevelProvider` inside `SealevelValidatorAnnounce`'s constructor, then it kind of becomes a  weird situation where whenever `SealevelProvider`'s constructor changes (ie. we add a new feature like metrics, that requires passing in `PrometheusMetrics` and `PrometheusConfig`), we would need to change `SealevelValidatorAnnounce`'s constructor as well.
If we ever wanted to share a provider between `SealevelValidatorAnnounce` and `SealevelMailboxIndexer` or something, we won't be able to because they each do their own instantiation inside their constructor.
So I think moving the logic out would eliminate this weird relationship and not break things whenever we add a new parameter to `SealevelProvider`'s constructor.

Downside is we might have duplicate code and utility functions for instantiation.

Alternatively, I can also add both constructors, one that takes in provider and one that constructs the provider inside.

What do you think?

rust/main/chains/hyperlane-sealevel/src/mailbox.rs
daniel-savu: > more of the logic to a higher level, where we can do all the configurations, metrics, etc we need. Then injecting it into the primitives we use.

doesn't this mean that higher level code needs to be aware of assumptions in lower-level code, making them more coupled? 

on the inecting point, we can later add an alternative constructor `from_mailbox(...)` to use if we ever want to mock the mailbox in tests, but by default I think it's more ergonomic to not be aware of what `SealevelMailboxIndexer` needs under the hood

>  if we ever want to reduce the amount of providers/rpc_clients we instantiate, then it makes sense to move this logic out and pass providers/rpc_clients into these structs instead.

this can also be achieved by passing a reference to the provider in the constructor

rust/main/chains/hyperlane-sealevel/src/validator_announce.rs
daniel-savu: > If we ever wanted to share a provider between SealevelValidatorAnnounce and SealevelMailboxIndexer or something, we won't be able to because they each do their own instantiation inside their constructor.

makes sense, would it actually be possible to put the provider behind a reference like Arc, as you suggested? But if it's a pain we can revisit later

daniel-savu (approved):

rust/main/chains/hyperlane-sealevel/src/provider.rs
kamiyaa: We also need to pass in `Option<ChainInfo>`

PR #5444, opened 2025-02-12
refactor: warp read logic

ljankovic-txfusion:
### Description

Refactored `runWarpRouteRead` function to improve logic flow.  

Prerequisite to merging #5395 feat: Add warp route ID lookup and validation to warp check command

### Drive-by changes

1. Extract `addresses` from `warpCoreConfig`
2. Removed redundant variable assignment in return statement

### Related issues

None

### Backward compatibility

Yes

### Testing

cli e2e


typescript/cli/src/read/warp.ts
```diff
@@ -32,25 +32,57 @@ export async function runWarpRouteRead({
...
+  } else if (symbol || warp) {
+    warpCoreConfig = await getWarpCoreConfigOrExit({
+      context,
+      warp,
+      symbol,
+    });
+    addresses = Object.fromEntries(
+      warpCoreConfig.tokens.map((t) => [t.chainName, t.addressOrDenom!]),
+    );
+  } else if (chain && address) {
```
nambrot, line 52: I wonder if its better to have two different methods for where warpCoreConfig is available, and where not, that imo would make it easier to read

ljankovic-txfusion: Then we would have to repeat the rest of the function for validation EVM chains and printing XERC20 limits within those special function. We either provide `warpCoreConfig` or extract if from `symbol || warp`, and the only time we don't use it is if `chain && address` are provided

ljankovic-txfusion: Introduced `getAddressesFromWarpCoreConfig` and `validateEvmCompatibility` utils to further improve readability. Util for logging XERC20Limits is also a good contender for refactoring, but I am afraid the diff will be too large

nambrot: Could we not extract those into shared private functions that are referenced on both these methods? I just think there are too many considtionals for my taste. And I do think extracting xerc20 limit logging would also be a good contendor for extraction, this is a pretty small PR, so I wouldn't worry about the diff change here

ljankovic-txfusion: One of the challenges we had was either having do dig through code to find the implementation of a class or staring at the non wrapped implementation of some obscure code, thats why I think its not a bad idea to have more private/pure function that are named by the underlying logic.

I further reduced conditionals, extracted xerc20 limit logging, and reverted private funcs.

nambrot (approved):

PR #5446, opened 2025-02-12
feat: Add multicast tx submission for Ethereum fallback provider

ameten:
### Description

Add multicast tx submission for Ethereum fallback provider.

`EthereumFallbackProvider` will use different strategies to query chain RPCs:
1. If the method is `eth_sendRawTransaction`, it will make a parallel query to all RPC endpoints and return the first successful result. If no success, it will retry 4 times.
2. For other methods, the current functionality does not change: it will sequentially query RPC endpoints according to their priorities and adjust priorities depending on success/failure. It also retries 4 times.

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5452

### Backward compatibility

Yes

### Testing

Added unit tests


rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -92,35 +93,102 @@ where
...
+                let mut unordered = FuturesUnordered::new();
+                self.inner
+                    .providers
+                    .iter()
+                    .for_each(|p| unordered.push(Self::provider_request(p, method, &params)));
+
+                let mut errors = vec![];
+
+                loop {
+                    let Some(resp) = unordered.next().await else {
```
daniel-savu, line 114: nit: this can be a `while let Some(resp) = ...` instead of a loop

ameten: Fixed

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,216 @@
...
+    ];
+    let fallback_provider = fallback_provider_builder.add_providers(providers).build();
+    let ethereum_fallback_provider = EthereumFallbackProvider::new(fallback_provider);
+    ethereum_fallback_provider
+        .multicast::<_, u64>(BLOCK_NUMBER_RPC, ())
+        .await
+        .unwrap();
+    let provider_call_count: Vec<_> =
+        ProviderMock::get_call_counts(&ethereum_fallback_provider).await;
+    assert_eq!(provider_call_count, vec![1, 0, 0]);
```
tkporter, line 121: why would we expect the first provider to succeed and not the others?

confused for a couple reasons:
- wouldn't we expect all 3 to be queried concurrently? I guess only the first one is queried because with this mock setup we never end up with a future that isn't ready so it just immediately resolves to the first completed 0th provider future before trying any of the others?
- I would've thought the use of `FuturesUnordered` wouldn't guarantee that the first one always wins the race?

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,216 @@
...
+    ];
+    let fallback_provider = fallback_provider_builder.add_providers(providers).build();
+    let ethereum_fallback_provider = EthereumFallbackProvider::new(fallback_provider);
+    ethereum_fallback_provider
+        .multicast::<_, u64>(BLOCK_NUMBER_RPC, ())
+        .await
+        .unwrap();
+    let provider_call_count: Vec<_> =
+        ProviderMock::get_call_counts(&ethereum_fallback_provider).await;
+    assert_eq!(provider_call_count, vec![1, 1, 0]);
```
tkporter, line 140: ditto do we guarantee these first 2 if we are using `FuturesUnordered`?

feels like these test cases aren't very meaningfully different from the behavior we'd expect with a fallback, is there any other way to test that all providers are requested concurrently?

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,216 @@
...
+    ];
+    let fallback_provider = fallback_provider_builder
+        .add_providers(providers)
+        .with_max_block_time(Duration::from_secs(0))
+        .build();
+    let ethereum_fallback_provider = EthereumFallbackProvider::new(fallback_provider);
+    ethereum_fallback_provider.fallback_test_call().await;
+    let provider_call_count: Vec<_> =
+        ProviderMock::get_call_counts(&ethereum_fallback_provider).await;
+    assert_eq!(provider_call_count, vec![0, 0, 2]);
```
tkporter, line 212: I know this isn't new but why is this `vec![0, 0, 2]` instead of `vec![0, 1, 0]`?

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -90,8 +94,65 @@ where
...
+            if i > 0 {
+                // sleep starting from the second attempt
+                sleep(Duration::from_millis(100)).await;
+            }
+
+            // future which visits all providers as they fulfill their requests
+            let mut unordered = self.populate_unordered_future(method, &params);
+
+            while let Some(resp) = unordered.next().await {
+                match categorize_client_response(method, resp) {
```
tkporter, line 137: curious if you think we should do a `_span` somewhere in here like we do for fallback, or if we already have all we can possibly need for the logs

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
ameten: It is mostly the implementation detail of `FutureUnordered`.

I'll change the test case to simulate that we call all mock before anyone succeeds.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
ameten: Since the first provider gets deprioritised, it is moved from 0th to 2nd position, so, the last entry in `provider_call_count` corresponds to the first provider. 

The value `2` instead of `1` is explained by the implementation of `handle_stalled_provider` where we query block number to identify if the provider stalled. So, it just happed that the kind of query we implemented the mock providers with is the same as the kind of query we identify stalled providers.

ameten: I'll add a comment into the test case.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
ameten: I added a warn span in the case when we did not receive a successful result from all the providers. It means that we shall see the warn span at most 4 times, printed at every attempts of the outer `for` loop.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,297 @@
...
+    let fallback_provider_builder = FallbackProviderBuilder::default();
+    let providers = vec![
+        EthereumProviderMock::new(None, None),
+        EthereumProviderMock::new(None, Some(2)),
+        EthereumProviderMock::new(None, Some(3)),
+    ];
+    let fallback_provider = fallback_provider_builder.add_providers(providers).build();
+    let ethereum_fallback_provider = EthereumFallbackProvider::new(fallback_provider);
+    let provider_id = ethereum_fallback_provider
+        .multicast::<_, u64>(BLOCK_NUMBER_RPC, ())
```
daniel-savu, line 129: where we could still improve coverage is making sure `.request` only uses multicast for `METHOD_SEND_RAW_TRANSACTION`. IMO in each of these unit tests we could call `.request(METHOD_SEND_RAW_TRANSACTION)` instead of `.multicast` and return a similar value that's unique in each provider (like we're doing with the block number current, but it could be any arbitrary string really)

I like how the fallback provider unit tests use `fallback_test_call` to call a method that's realistically expected to be routed to the fallback logic (`BLOCK_NUMBER_RPC`)

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,297 @@
...
+        .await
+        .unwrap();
+    let provider_call_count: Vec<_> =
+        ProviderMock::get_call_counts(&ethereum_fallback_provider).await;
+    assert_eq!(provider_id, 2);
+    assert_eq!(provider_call_count, vec![1, 1, 1]);
+}
+
+#[tokio::test]
+async fn test_multicast_first_provider_succeeds_slow_third_succeeds_immediately() {
```
daniel-savu, line 199: yess this is the future of finance

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
```diff
@@ -0,0 +1,297 @@
...
+async fn test_multicast_none_provider_succeeds() {
+    let fallback_provider_builder = FallbackProviderBuilder::default();
+    let providers = vec![
+        EthereumProviderMock::new(None, None),
+        EthereumProviderMock::new(None, None),
+        EthereumProviderMock::new(None, None),
+    ];
+    let fallback_provider = fallback_provider_builder.add_providers(providers).build();
+    let ethereum_fallback_provider = EthereumFallbackProvider::new(fallback_provider);
+    let _ = ethereum_fallback_provider
```
daniel-savu, line 248: should we check that this is indeed an error?

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -90,8 +94,75 @@ where
...
     {
         use CategorizedResponse::*;
+
+        let params = serde_json::to_value(params).expect("valid");
+
+        // errors reported by providers
+        let mut errors = vec![];
+
+        // retry 4 times if all providers returned a retryable error
+        for i in 0..=3 {
```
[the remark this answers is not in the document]
ameten: The number of requests which are done by multicast and fallback is the same in the case of unsuccessful retry attempt: all providers have failed on such attempt, hence both multicast and fallback made the same number of requests. 
The only difference is one additional request per provider for multicast in the case when one of the providers returned success at the last retry attempt. It is the same difference between providers independently if we have retrying or not.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
ameten: Good point. I added such a check.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
ameten: Good point. I added `multicast_test_call` and now we cover the decision on multicast/fallback choice.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -1,21 +1,25 @@
...
 use ethers_prometheus::json_rpc_client::{JsonRpcBlockGetter, PrometheusJsonRpcClientConfigExt};
+use hyperlane_core::rpc_clients::{BlockNumberGetter, FallbackProvider};
 
 use crate::rpc_clients::{categorize_client_response, CategorizedResponse};
 
+const METHOD_SEND_RAW_TRANSACTION: &str = "eth_sendRawTransaction";
+
 /// Wrapper of `FallbackProvider` for use in `hyperlane-ethereum`
 #[derive(new)]
 pub struct EthereumFallbackProvider<C, B>(FallbackProvider<C, B>);
```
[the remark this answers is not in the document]
ameten: I added doc comment.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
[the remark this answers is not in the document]
daniel-savu: I see, you're right yep

daniel-savu (approved):
🙏

rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback/tests.rs
tkporter: makes sense! might be nice to have a quick comment for future readers but happy to merge without it

tkporter (approved):
🙏

PR #5455, opened 2025-02-12
feat(relayer): don't retry old messages

daniel-savu:
### Description

We've recently noticed the submitter attempts to process many old messages at the same time, because they have very similar `next_retry_attempt`s. These old messages are unlikely to be processable, and in fact starve new "healthy" messages, sometimes leading to lowered throughput.

There was an attempt to make this less of an issue in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5416, but we decided old messages are not worth retrying at all given we offer the ability to deliver messages on-demand in the hyperlane CLI.

Not retrying old messages means we don't even have to push them to the submitter when loaded from the db - we just skip them.

This PR:
- introduces a `HYP_MAXMESSAGERETRIES` env var which defaults to 66 if not set. The default of 66 is picked under the assumption that the first 48 retries take about 1 day ([formula](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4d6fd2f8d72dbc882572e73ba599b80870f2bf67/rust/main/agents/relayer/src/msg/pending_message.rs#L674)). The remaining 22 retries will take at least `21 * 22 / 2` hours (see the formula), which is about 11 days. In total that is almost 2 weeks, which @nambrot confirmed is how long we should retry for. If the default is incorrect we can set a higher env var to override. I also manually checked the queues for the arbitrum and base destinations, and there was no message with a retry count of 100 but most old ones had more than 70 attempts.
- skips messages read from the db whose retry count is >= MAX_MESSAGE_RETRIES
- for messages already loaded in the submitter, sets `Duration::from_secs(u32::MAX)` as the next retry attempt, which is far enough into the future as to not be attempted at all

Prepare queues are expected to get much smaller because of the skipping, which hopefully also lowers memory pressure.

### Drive-by changes

- Makes the db field in `MessageContext` a trait obj for easier future mocking

### Backward compatibility

Yes - if `HYP_MAXMESSAGERETRIES` is set to a very high value, it's as if this feature is disabled

### Testing

Unit testing of the message backoff logic and of inner logic in `maybe_from_persisted_retries`. The env var loading isn't tested but is very similar to how we load env vars [elsewhere](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4d6fd2f8d72dbc882572e73ba599b80870f2bf67/rust/main/chains/hyperlane-sealevel/src/rpc/client.rs#L412) - longer term it should be turned into a proper agent config (the other ones too).


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -562,15 +553,45 @@ impl PendingMessage {
                 0
             }
         };
+        // Skip this message if it has been retried too many times
+        if let Some(max_retries) = retries_before_skipping {
+            if num_retries >= max_retries {
+                return None;
+            }
+        }
```
daniel-savu, line 561: this is new logic, the rest is refactored out of `from_persisted_retries`

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -562,15 +553,45 @@ impl PendingMessage {
...
+        Some(num_retries)
+    }
 
-        let mut pm = Self::new(message, ctx, message_status, app_context);
-        if num_retries > 0 {
-            let next_attempt_after =
-                PendingMessage::calculate_msg_backoff(num_retries).map(|dur| Instant::now() + dur);
-            pm.num_retries = num_retries;
-            pm.next_attempt_after = next_attempt_after;
+    fn get_message_status(
```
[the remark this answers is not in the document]
kamiyaa: Nice!

kamiyaa (approved):

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -516,44 +517,37 @@ impl PendingOperation for PendingMessage {
...
+            pending_message.next_attempt_after = next_attempt_after;
+        }
+        Some(pending_message)
+    }
 
-        let num_retries = match ctx
-            .origin_db
+    fn maybe_get_num_retries(
+        origin_db: Arc<dyn HyperlaneDb>,
+        message: HyperlaneMessage,
```
ameten, line 547: This method does not need to take ownership of `message`. I wonder if passing as a reference will be sufficient.

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -562,15 +556,45 @@ impl PendingMessage {
...
 
-        let mut pm = Self::new(message, ctx, message_status, app_context);
-        if num_retries > 0 {
-            let next_attempt_after =
-                PendingMessage::calculate_msg_backoff(num_retries).map(|dur| Instant::now() + dur);
-            pm.num_retries = num_retries;
-            pm.next_attempt_after = next_attempt_after;
+    fn get_message_status(
+        origin_db: Arc<dyn HyperlaneDb>,
+        message: HyperlaneMessage,
```
ameten, line 570: This method does not need to take ownership of message. I wonder if passing as a reference will be sufficient.

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -516,44 +517,37 @@ impl PendingOperation for PendingMessage {
...
-        };
+        retries_before_skipping: Option<u32>,
+    ) -> Option<Self> {
+        let num_retries = Self::maybe_get_num_retries(
+            ctx.origin_db.clone(),
+            message.clone(),
+            retries_before_skipping,
+        )?;
+
+        let message_status = Self::get_message_status(ctx.origin_db.clone(), message.clone());
```
ameten, line 534: We can avoid the clone here if `get_message_status` does not take ownership of its `message` parameter.

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -711,3 +742,200 @@ impl MessageSubmissionMetrics {
...
+
+    #[test]
+    fn test_calculate_msg_backoff_does_not_overflow() {
+        use super::PendingMessage;
+        use std::time::Duration;
+        const SECS_PER_MINUTE: u64 = 60;
+        const MINS_PER_HOUR: u64 = 60;
+        const HOURS_PER_DAY: u64 = 24;
+        const DAYS_PER_WEEK: u64 = 7;
+        const SECS_PER_WEEK: u64 = SECS_PER_MINUTE * MINS_PER_HOUR * HOURS_PER_DAY * DAYS_PER_WEEK;
```
ameten, line 878:
```suggestion
        let duration = Duration::from_seconds(chrono::Duration::weeks(1).num_seconds())
```

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -711,3 +742,200 @@ impl MessageSubmissionMetrics {
...
 }
+
+#[cfg(test)]
+mod test {
+    use crate::msg::pending_message::DEFAULT_MAX_MESSAGE_RETRIES;
+    use hyperlane_base::db::*;
+    use hyperlane_core::*;
+    use std::{fmt::Debug, sync::Arc};
+
+    use super::PendingMessage;
```
ameten, line 753:
```suggestion
    use std::{fmt::Debug, sync::Arc};

    use hyperlane_base::db::*;
    use hyperlane_core::*;

    use crate::msg::pending_message::DEFAULT_MAX_MESSAGE_RETRIES;
    
    use super::PendingMessage;
```

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -711,3 +742,200 @@ impl MessageSubmissionMetrics {
...
+        let db = dummy_db_with_retries(mock_retries);
+        let num_retries = PendingMessage::maybe_get_num_retries(
+            Arc::new(db),
+            HyperlaneMessage::default(),
+            retries_before_skipping,
+        );
+
+        // retry count is the same, because `retries_before_skipping` is `None`
+        assert_eq!(num_retries, expected_retries);
+    }
```
ameten, line 910: nit: move this auxiliary methods to the end

ameten (approved):

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -516,44 +517,37 @@ impl PendingOperation for PendingMessage {
 
 impl PendingMessage {
     /// Constructor that tries reading the retry count from the HyperlaneDB in order to recompute the `next_attempt_after`.
+    /// If the message has been retried more than `max_retries`, it will return `None`.
```
tkporter, line 520: nit: max_retries seems stale

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -562,15 +556,45 @@ impl PendingMessage {
                 0
             }
         };
+        // Skip this message if it has been retried too many times
+        if let Some(max_retries) = retries_before_skipping {
```
tkporter, line 560: feels a little awkward for this function to be the place where the decision of whether to skip is made?

I would've thought that we'd have `get_num_retries` always return a u32 that is accurate (easier to reason about it then imo), and then in `maybe_from_persisted_retries` we make the judgement call whether to ignore the message based off that number. The benefit being that this retry skipping logic isn't so intertwined in multiple places

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -516,44 +517,37 @@ impl PendingOperation for PendingMessage {
...
-                    } else {
-                        Level::TRACE
-                    },
-                    "Message status not found in db"
-                );
-                PendingOperationStatus::FirstPrepareAttempt
-            }
-        };
+        retries_before_skipping: Option<u32>,
+    ) -> Option<Self> {
```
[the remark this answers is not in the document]
daniel-savu: we'd need to first set the MAX_RETRY var to a high number, load those messages in, and then we can retry them

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -651,6 +675,10 @@ impl PendingMessage {
     /// given the number of retries.
     /// `pub(crate)` for testing purposes
     pub(crate) fn calculate_msg_backoff(num_retries: u32) -> Option<Duration> {
+        let max_retries = std::env::var("MAX_MESSAGE_RETRIES")
```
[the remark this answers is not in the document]
daniel-savu: yeah i also think the config logic is too hard to work with, I was thinking we can add a ticket to sweep all the env var reads at once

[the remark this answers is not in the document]
daniel-savu: moved into relayer settings

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -562,15 +556,45 @@ impl PendingMessage {
                 0
             }
         };
+        // Skip this message if it has been retried too many times
+        if let Some(max_retries) = retries_before_skipping {
+            if num_retries >= max_retries {
+                return None;
```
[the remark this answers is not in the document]
daniel-savu: I'll add a log for in-memory messages but not for those skipped at startup since there'll be lots

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: great point, I refactored so `get_num_retries` returns the true value but we can test both `get_num_retries` and the decision making at the same time

rust/main/agents/relayer/src/settings/mod.rs
```diff
@@ -307,6 +310,12 @@ impl FromRawConf<RawRelayerSettings> for RelayerSettings {
             })
             .unwrap_or_default();
 
+        let max_message_retries = p
+            .chain(&mut err)
+            .get_opt_key("maxMessageRetries")
```
daniel-savu, line 315: this will have to be set as `HYP_MAX_MESSAGE_RETRIES`

tkporter: not super important but would be `HYP_MAXMESSAGERETRIES` right?

rust/main/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
tkporter: gotcha, happy to merge as is and just know that we have a path forward if this is required.

Thinking mostly of messages that involve non-EVM chains where we do need the relayer

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -65,6 +66,8 @@ pub struct PendingMessage {
     ctx: Arc<MessageContext>,
     status: PendingOperationStatus,
     app_context: Option<String>,
+    #[serde(skip_serializing)]
+    max_message_retries: u32,
```
tkporter, line 70: nit: def don't feel strongly but could maybe do `max_retries` as the message is redundant (& to more closely relate this to the `num_retries`

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -516,61 +519,90 @@ impl PendingOperation for PendingMessage {
...
+        max_message_retries: u32,
+    ) -> Option<u32> {
+        let num_retries = Self::get_num_retries(origin_db.clone(), message);
+        if Self::should_skip(num_retries, max_message_retries) {
+            return None;
+        }
+        Some(num_retries)
+    }
+
+    pub fn should_skip(retry_count: u32, retries_before_skipping: u32) -> bool {
```
tkporter, line 562: nit: any reason to have a different name for `retries_before_skipping` vs elsewhere? just imagine consistency within `PendingMessage` is easier to track

tkporter (approved):
✂️✉️

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: fixed

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: updated

PR #5474, opened 2025-02-14
fix: CCIP deploy

paulbalaji:
### Description

Fixes deploying CCIP hooks and isms by correctly importing the right CCIP router interface.

### Drive-by changes

simplify config generation in infra


typescript/infra/scripts/deploy.ts
```diff
@@ -254,7 +253,12 @@ async function main() {
     if (environment !== 'mainnet3') {
       throw new Error('CCIP is only supported on mainnet3');
     }
-    config = getCCIPDeployConfig(filteredTargetNetworks);
+    config = Object.fromEntries(
+      targetNetworks.map((origin) => [
```
nambrot, line 257: Was the use of `targetNetworks` instead of `filteredTargetNetworks` intentional?

typescript/sdk/src/ccip/utils.ts
```diff
@@ -121,7 +121,6 @@ export async function isSupportedCCIPLane({
 }): Promise<boolean> {
   const originRouter = getCCIPRouterAddress(origin);
   const destinationSelector = getCCIPChainSelector(destination);
-
```
nambrot, line 124: Highly recommend using `git add -p` to conciously check in the desired changes

paulbalaji: this was intended to remove extraneous whitespace in a small function, but can revert if we want to minimise diff

typescript/infra/scripts/deploy.ts
paulbalaji: good catch - doesn't make a big difference to what ends up getting deployed with the current inputs but yeah meant to use filtered - will fix + undo the whitespace change

nambrot (approved):

PR #5477, opened 2025-02-14
feat: deploy CDX token

tkporter:
### Description

- Deploys the CDX token


typescript/infra/config/environments/mainnet3/warp/warpIds.ts
```diff
@@ -70,4 +70,5 @@ export enum WarpRouteIds {
   SolanaSonicsvmUSDStar = 'USDStar/solanamainnet-sonicsvm',
   SolanaSonicsvmSSOL = 'sSOL/solanamainnet-sonicsvm',
   ArbitrumEthereumSolanaTreasureSMOL = 'SMOL/arbitrum-ethereum-solanamainnet-treasure',
+  BaseSolanaCDX = 'CDX/base-solanamainnet',
```
daniel-savu, line 73: are we not including this in check-warp-deploy? since we didn't add an entry in `warpConfigGetterMap`

daniel-savu (approved):

PR #5481, opened 2025-02-15
feat: Expand warp deploy config in warp check

nambrot:
### Description

When using `WarpDeployConfig` we often assume certain defaults, however when we compare against `warp read` config, it includes all the "expanded" config. This PR adds "config expansion", so that checking actually can be done. As part of that, it removes the practice of `warp read` after deploy in the warp check e2e tests so that the actual comparison against the intended deploy config can happen (instead of the inferred config from warp read).

This is related to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5358 and in fact lifts some of the convenience function into `configUtils.ts` fyi @mshojaei-txfusion 

### Drive-by changes

The warp-check e2e test setup changes as mentioned

### Backward compatibility

Yes

### Testing

Local e2e tests and manual

### DIFF
--- a/typescript/cli/src/check/warp.ts
+++ b/typescript/cli/src/check/warp.ts
@@ -6,6 +6,16 @@ import { ObjectDiff, diffObjMerge } from '@hyperlane-xyz/utils';
 import { log, logGreen } from '../logger.js';
 import { formatYamlViolationsOutput } from '../utils/output.js';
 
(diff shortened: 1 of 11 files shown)


nambrot:
Closing in favor of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5483 to keep changes more reviewable

PR #5478, opened 2025-02-14
feat: add collateral checks to warp route checker

yorhodes:
### Description

Implement checking proxy and ownership of warp route collateral contracts.

### Backward compatibility

Yes

### Testing

Unit tests

PR #5482, opened 2025-02-15
feat: Support xerc20 token config derivation in warp check

nambrot:
### Description

Allow `HypXERC20` and `HypXERC20Lockbox` to be derived in the warp reader, similar to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5478, just for CLI

### Backward compatibility

Yes

### Testing

Manual/Unit Tests

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
@@ -9,6 +9,7 @@ import {
   ERC4626Test__factory,
   Mailbox,
   Mailbox__factory,
+  XERC20Test__factory,
 } from '@hyperlane-xyz/core';
 import {
   HyperlaneContractsMap,
@@ -164,6 +165,58 @@ describe('ERC20WarpRouterReader', async () => {
       expect(derivedConfig.decimals).to.equal(TOKEN_DECIMALS);
     }
   });
+
+  it('should derive xerc20 config correctly', async () => {
+    // Create token
+    // @ts-ignore
+    const xerc20Token = await new XERC20Test__factory(signer).deploy(
+      TOKEN_NAME,
+      TOKEN_NAME,
+      TOKEN_SUPPLY,
+      TOKEN_DECIMALS,
+    );
+    // Create config
+    const config: WarpRouteDeployConfig = {
+      [chain]: {
+        type: TokenType.XERC20,
+        token: xerc20Token.address,
+        hook: await mailbox.defaultHook(),
+        interchainSecurityModule: await mailbox.defaultIsm(),
+        ...baseConfig,
+      },
+    };
+    // Deploy with config
+    const warpRoute = await deployer.deploy(config);
+
+    // Derive config and check if each value matches
+    const derivedConfig = await evmERC20WarpRouteReader.deriveWarpRouteConfig(
+      warpRoute[chain].xERC20.address,
+    );
+    for (const [key, value] of Object.entries(derivedConfig)) {
+      const deployedValue = (config[chain] as any)[key];
+      if (deployedValue && typeof value === 'string')
+        expect(deployedValue).to.equal(value);
+    }
+
+    // Check hook because they're potentially objects
+    expect(derivedConfig.hook).to.deep.equal(
+      await evmERC20WarpRouteReader.evmHookReader.deriveHookConfig(
+        config[chain].hook as string,
+      ),
+    );
+    // Check ism
+    expect(
+      (derivedConfig.interchainSecurityModule as DerivedIsmConfig).address,
+    ).to.be.equal(await mailbox.defaultIsm());
+
+    // Check if token values matches
+    if (derivedConfig.type === TokenType.XERC20) {
+      expect(derivedConfig.name).to.equal(TOKEN_NAME);
+      expect(derivedConfig.symbol).to.equal(TOKEN_NAME);
+      expect(derivedConfig.decimals).to.equal(TOKEN_DECIMALS);
+    }
+  });
+
   it('should derive synthetic rebase config correctly', async () => {
     // Create config
     const config: WarpRouteDeployConfig = {
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
@@ -1,7 +1,6 @@
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers.js';
 import { expect } from 'chai';
 import hre from 'hardhat';
-import { zeroAddress } from 'viem';
 
 import {
   ERC20Test,
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
@@ -169,7 +169,6 @@ describe('ERC20WarpRouterReader', async () => {
 
   it('should derive xerc20 config correctly', async () => {
     // Create token
-    // @ts-ignore
     const xerc20Token = await new XERC20Test__factory(signer).deploy(
       TOKEN_NAME,
       TOKEN_NAME,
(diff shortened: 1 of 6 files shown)


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -126,6 +128,27 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
         await warpRoute[method]();
+
+        // If its a collateral, it could be an xerc20 token
+        if (tokenType === TokenType.collateral) {
+          // Try if its HypXERC20Lockbox
+          const hypXERC20Lockbox = HypXERC20Lockbox__factory.connect(
+            warpRouteAddress,
+            this.provider,
+          );
+          try {
```
ltyu, line 139: Why can’t these be added to the above contractTypes, similar to what we do for collateralVaults?

nambrot: Ah yes, good point

nambrot: Unfortunately, doesn't work for the HypXERC20 one, we have to go into the token

typescript/sdk/src/token/types.ts
```diff
@@ -31,28 +31,44 @@ export const NativeTokenConfigSchema = TokenMetadataSchema.partial().extend({
...
+};
+
+export const CollateralTokenConfigSchema = TokenMetadataSchema.partial()
+  .extend({
     type: z.enum([
       TokenType.collateral,
       TokenType.collateralVault,
       TokenType.collateralVaultRebase,
-      TokenType.XERC20,
-      TokenType.XERC20Lockbox,
```
Mo-Hussain, line 41: Should we consider adding a new TokenTypes, say `XERC20VS` and `XERC20VSLockbox` given that the limits are implemented differently? This would mean we would have to check for this type in several places across the codebase.

We don't have the need to manage limits for a standard xERC20s now but it we did, we would have to revisit this setup. One option to manage this in the future would be to add `maxMintLimit` and `maxBurnLimit` optional fields to the config and use the presence of a particular limit related field the distinguish the two implementations

typescript/sdk/src/token/types.ts
```diff
@@ -31,28 +31,44 @@ export const NativeTokenConfigSchema = TokenMetadataSchema.partial().extend({
...
-      ),
-  },
-);
+  })
+  .extend(sharedCollateralTokenConfig);
+
 export type CollateralTokenConfig = z.infer<typeof CollateralTokenConfigSchema>;
 export const isCollateralTokenConfig = isCompliant(CollateralTokenConfigSchema);
 
+const xERC20LimitConfigSchema = z.object({
```
Mo-Hussain, line 57: Could it be cleaner to define the config like 

```
xERC20:
        limits:
               burnCap: 10000
               rateLimitPerSecond: 10
        owner: "..."
```

I realise that for the current setup, the xERC20 has the same owner as the Warp Route but it may be helpful to be able to set a differen owner in future

typescript/sdk/src/token/types.ts
nambrot: Yeah i think eventually it probably makes sense, but for now, I think just optional fields and attempting to read the values is good enough to me

typescript/sdk/src/token/types.ts
nambrot: Agreed pushed [76f8a5c](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5482/commits/76f8a5cb4ae5afb6abdf6fd185b79155a6990627)

Mo-Hussain: Could we set a different owner for the xERC20 so that we can manage limits without more nimbly??

nambrot: It's the reverse, the xerc20 limits are the most sensitive. In any case, we don't have a plan for this right now, if we do, we'll add it

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -185,40 +201,102 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+    ];
+    const xERC20 = new Contract(xERC20Address, rateLimitsABI, this.provider);
+    try {
+      return {
+        xERC20: {
+          limits: {
+            rateLimitPerSecond: (
+              await xERC20.rateLimitPerSecond(warpRouteAddress)
+            ).toString(),
+            bufferCap: (await xERC20.bufferCap(warpRouteAddress)).toString(),
```
yorhodes, line 222: string?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -185,40 +201,102 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+        xerc20Token = await hypXERC20Lockbox.xERC20();
+        token = xerc20Token;
+        lockbox = await hypXERC20Lockbox.lockbox();
+      } else {
+        const erc20 = HypERC20Collateral__factory.connect(
+          warpRouteAddress,
+          this.provider,
+        );
+        token = await erc20.wrappedToken();
+      }
```
yorhodes, line 269: just noting this is duped with the checker's `getCollateralToken`

typescript/sdk/src/token/types.ts
```diff
@@ -31,28 +31,53 @@ export const NativeTokenConfigSchema = TokenMetadataSchema.partial().extend({
...
 export const isNativeTokenConfig = isCompliant(NativeTokenConfigSchema);
 
-export const CollateralTokenConfigSchema = TokenMetadataSchema.partial().extend(
-  {
+const sharedCollateralTokenConfig = {
+  // For xerc20lockbox and collateral vault, we use the token as the lockbox/vault respectively
+  token: z
+    .string()
+    .describe('Existing token address to extend with Warp Route functionality'),
+};
```
yorhodes, line 39: should you have the types discriminated here?

typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
```diff
@@ -164,6 +166,109 @@ describe('ERC20WarpRouterReader', async () => {
       expect(derivedConfig.decimals).to.equal(TOKEN_DECIMALS);
     }
   });
+
+  it('should derive xerc20 config correctly', async () => {
+    // Create token
+    // @ts-ignore
```
yorhodes, line 172: why ts-ignore?

yorhodes (approved):

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
nambrot: Yeah so I originally had it as number as well, but string seemed "safer"

typescript/sdk/src/token/types.ts
nambrot: wdym?

typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
nambrot: I'll remove

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
nambrot: I think it is very similar, but slightly different, no? Here the input is the `warpRouteAddress`, while in the checker the input is the `token` address, right?

yorhodes: I guess, though you can get the token from `.wrappedToken`

typescript/sdk/src/token/types.ts
yorhodes: like this type is no different from the one below with the same `token` field

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
nambrot: Well in this case, we actaully want the xerc20 token when the type is XERC20Lockbox, not the "inner" token

PR #5483, opened 2025-02-16
feat: Expand warp deploy config in warp check

nambrot:
### Description

When using `WarpDeployConfig` we often assume certain defaults, however when we compare against `warp read` config, it includes all the "expanded" config. This PR adds "config expansion", so that checking actually can be done. As part of that, it removes the practice of `warp read` after deploy in the warp check e2e tests so that the actual comparison against the intended deploy config can happen (instead of the inferred config from warp read).

This is related to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5358 and in fact lifts some of the convenience function into `configUtils.ts` fyi @mshojaei-txfusion 

### Drive-by changes

The warp-check e2e test setup changes as mentioned

### Backward compatibility

Yes

### Testing

Local e2e tests and manual

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -1,7 +1,12 @@
 import { stringify as yamlStringify } from 'yaml';
 import { CommandModule } from 'yargs';
 
-import { ChainName, ChainSubmissionStrategySchema } from '@hyperlane-xyz/sdk';
+import {
+  ChainName,
+  ChainSubmissionStrategySchema,
+  expandWarpDeployConfig,
+  getRouterAddressesFromWarpCoreConfig,
+} from '@hyperlane-xyz/sdk';
 import { assert, objFilter } from '@hyperlane-xyz/utils';
 
 import { runWarpRouteCheck } from '../check/warp.js';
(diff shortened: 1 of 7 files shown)


typescript/cli/src/commands/warp.ts
```diff
@@ -367,9 +372,21 @@ export const check: CommandModuleWithContext<{
       symbol,
     });
 
+    const warpCoreConfig = context.warpCoreConfig;
+
+    if (!warpCoreConfig) {
+      throw new Error('No warp core config found');
+    }
```
yorhodes, line 379: seems like this should happen in the context function?

typescript/sdk/src/token/configUtils.ts
```diff
@@ -0,0 +1,95 @@
...
+  deployedRoutersAddresses: ChainMap<Address>,
+): Promise<WarpRouteDeployConfig> {
+  const derivedTokenMetadata = await HypERC20Deployer.deriveTokenMetadata(
+    multiProvider,
+    warpDeployConfig,
+  );
+  return promiseObjAll(
+    objMap(warpDeployConfig, async (chain, config) => {
+      const [remoteRouters, destinationGas] =
+        getDefaultRemoteRouterAndDestinationGasConfig(
```
yorhodes, line 79: I dont think this needs to be async

typescript/sdk/src/token/configUtils.ts
```diff
@@ -0,0 +1,95 @@
...
+          chain,
+          deployedRoutersAddresses,
+          warpDeployConfig,
+        );
+
+      return {
+        ...derivedTokenMetadata,
+        remoteRouters,
+        destinationGas,
+        proxyAdmin: { owner: config.owner },
```
yorhodes, line 90: this feels like a hack

nambrot: Can you share why?

typescript/cli/src/commands/warp.ts
nambrot: IIUC, the type is just not expressive enough?

yorhodes (approved):

PR #5484, opened 2025-02-16
feat: Support ISM/Hook checking when using defaults

nambrot:
### Description

This PR changes the canonical definition of using the default ism/hooks from `undefined` to `zeroAddress`. This ultimately makes cli checking work. Builds on top of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5483 and should be based against it once CI tests are passing

### Backward compatibility

Mostly yes (because I don't think this behavior has been leveraged)

### Testing

Unit tests/e2e tests/ manual tests

### DIFF
--- a/typescript/sdk/src/token/configUtils.ts
+++ b/typescript/sdk/src/token/configUtils.ts
@@ -1,3 +1,5 @@
+import { zeroAddress } from 'viem';
+
 import { Address, objMap } from '@hyperlane-xyz/utils';
 
 import { MultiProvider } from '../providers/MultiProvider.js';
@@ -79,6 +81,8 @@ export async function expandWarpDeployConfig(
       ...derivedTokenMetadata,
       remoteRouters,
       destinationGas,
+      hook: zeroAddress,
+      interchainSecurityModule: zeroAddress,
       proxyAdmin: { owner: config.owner },
 
       // User-specified config takes precedence
--- a/.changeset/lazy-otters-boil.md
+++ b/.changeset/lazy-otters-boil.md
@@ -1,0 +1,6 @@
+---
+'@hyperlane-xyz/cli': minor
+'@hyperlane-xyz/sdk': minor
+---
+
+Change semantics of ism/hook config from undefined to 0x0 for reading/checking purposes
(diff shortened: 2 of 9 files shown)


typescript/cli/src/check/warp.ts
```diff
@@ -6,6 +6,16 @@ import { ObjectDiff, diffObjMerge } from '@hyperlane-xyz/utils';
...
 
+const KEYS_TO_IGNORE = ['totalSupply'];
+
+function sanitizeConfig(obj: any): any {
+  // Remove keys from obj
+  const filteredObj = Object.fromEntries(
+    Object.entries(obj).filter(([key]) => !KEYS_TO_IGNORE.includes(key)),
+  );
+  return normalizeConfig(filteredObj);
+}
```
yorhodes, line 17: this seems a bit dangerous
for eg HypERC20 initial supply this is important

typescript/sdk/src/token/configUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+import { WarpRouteDeployConfig } from './types.js';
+
+/**
+ * Gets router address for a chain
+ */
+const getRemoteRouterAddress = (
+  deployedRoutersAddresses: ChainMap<Address>,
+  chain: string,
+): Address => deployedRoutersAddresses[chain];
+
```
yorhodes, line 21: should this really be a function?

typescript/sdk/src/token/configUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Gets gas configuration for a chain
+ */
+const getGasConfig = (
+  warpDeployConfig: WarpRouteDeployConfig,
+  chain: string,
+): string =>
+  warpDeployConfig[chain].gas?.toString() ||
+  gasOverhead(warpDeployConfig[chain].type).toString();
```
yorhodes, line 30: why to string?

typescript/sdk/src/token/configUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export function getDefaultRemoteRouterAndDestinationGasConfig(
```
yorhodes, line 32: this feels a bit unrelated to the scope of the PR description

yorhodes (commented):
I dont really understand why this change was made, seems breaking unnecessarily

yorhodes (approved):

PR #5485, opened 2025-02-16
feat: Support xERC20 limit checking

nambrot:
### Description

This PR supports checking the bufferCap and rateLimitPerSecond configuration of the warp route directly in the warp deploy config. 

### Backward compatibility

Yes

### Testing

Manually tested, since the velo xerc20 contracts are not yet referenced in the monorepo


nambrot:
Merged this into #5482

PR #5487, opened 2025-02-16
fix: overflow in new `next_attempt_after` calculation

daniel-savu:
### Description

Due to our lack of unit testing coverage, an overflow bug caused by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5455 wasn't caught. This bug happened on 7 chains on RC and took down their submitters.

I've noticed it now on RC due to the rising prep queues ([source](https://abacusworks.grafana.net/goto/kbQcXdcNR?orgId=1)), and also confirmed using these logs: https://cloudlogging.app.goo.gl/zJKqhTCTPAxETTg79. 

I'm not sure why but for crashed submitters, the prep queue keeps increasing although the only logic pushing to it lives in the submitter. Maybe's there's a future cancellation issue with `receive_task` and as a result it's still alive ([source](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ab2917a424534cb034400ad85836b9f833832aaf/rust/main/agents/relayer/src/msg/op_submitter.rs#L222)). [Logs](https://cloudlogging.app.goo.gl/qEjQq5AaFPWpvrv58) show that messages are successfully sent over [this channel](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ab2917a424534cb034400ad85836b9f833832aaf/rust/main/agents/relayer/src/msg/processor.rs#L306) to the submitter (to `base` in the logs linked), and this operation would fail if no receiving end existed.

This PR adds a unit test that would've caught the overflow, and lowers the max backoff period from `u32::MAX` to 10 weeks into the future.

### Drive-by changes

- Refactors the logic that calculates `next_attempt` after so it can be tested
- adds `chrono` as a relayer dependency to easily calcualte how many seconds are in 10 weeks

### Backward compatibility

Yes

### Testing

Adds a unit test that would have caught the overflow

### DIFF
--- a/rust/main/agents/relayer/src/msg/pending_message.rs
+++ b/rust/main/agents/relayer/src/msg/pending_message.rs
@@ -531,13 +531,16 @@ impl PendingMessage {
         let message_status = Self::get_message_status(ctx.origin_db.clone(), &message);
         let mut pending_message = Self::new(message, ctx, message_status, app_context, max_retries);
         if num_retries > 0 {
-            let next_attempt_after =
-                PendingMessage::calculate_msg_backoff(num_retries, max_retries, None)
-                    .map(|dur| Instant::now() + dur);
+            let next_attempt_after = Self::next_attempt_after(num_retries, max_retries);
             pending_message.num_retries = num_retries;
             pending_message.next_attempt_after = next_attempt_after;
         }
         Some(pending_message)
+    }
+
+    fn next_attempt_after(num_retries: u32, max_retries: u32) -> Option<Instant> {
+        PendingMessage::calculate_msg_backoff(num_retries, max_retries, None)
+            .map(|dur| Instant::now() + dur)
     }
 
     fn get_retries_or_skip(
@@ -716,7 +719,7 @@ impl PendingMessage {
                         "Message has been retried too many times, skipping",
                     );
                 }
-                u64::MAX
+                chrono::Duration::weeks(10).num_seconds() as u64
             }
         }))
     }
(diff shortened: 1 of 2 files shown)


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -716,7 +719,7 @@ impl PendingMessage {
                         "Message has been retried too many times, skipping",
                     );
                 }
-                u64::MAX
+                chrono::Duration::weeks(10).num_seconds() as u64
```
daniel-savu, line 722: this is the fix

tkporter, line 722: more for my learning, why chrono instead of std time?

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -888,22 +895,23 @@ mod test {
     fn test_calculate_msg_backoff_does_not_overflow() {
```
tkporter, line 895: why did this test not fail before? I don't understand why it was overflowing as well

tkporter (approved):

aroralanuk (approved):
lgtm

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: `std::time` doesn't have a `weeks` method (at least on the version we're on)

rust/main/agents/relayer/src/msg/pending_message.rs
daniel-savu: discussed irl, it was overflowing because the backoff period was added to `Instant::now()` inside `maybe_from_persisted_retries`. The unit test only checked that the call to `calculate_msg_backoff` didn't overflow, which I hadn't realised was used in an addition afterwards

Issue #5466, opened 2025-02-13
`tokio::spawn` should set a critical error metric on a panic


daniel-savu:
create wrapper logic for this

PR #5490, opened 2025-02-17
feat: tokio task metrics

daniel-savu:
### Description

We recently had a critical bug on RC where the submitter task crashed due to a panic (https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5487). We've had similar issues in the past, where one tokio task getting dropped wasn't detected by alerts and was only discovered later.

This PR integrates tokio-metrics with our Prometheus setup and exposes one of the metrics available: `dropped_tasks`, which tracked in prometheus as the cumulative count of all tasks ever dropped. In e2e this metric is zero after invariants pass - in prod I think we'll alert if this ever becomes nonzero.

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5466

### Drive-by changes

No

### Backward compatibility

Yes

### Testing

Checked in e2e that the new metric exists, otherwise relying on correctness from tokio-metrics

### DIFF
--- a/rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
+++ b/rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
@@ -9,9 +9,7 @@ use tracing::{info_span, instrument::Instrumented, Instrument};
 
 use super::CoreMetrics;
 
-/// Expected label names for the `gas_price` metric.
-/// Help string for the metric.
-pub const RUNTIME_DROPPED_TASKS_HELP: &str = "The number of tasks dropped";
+const RUNTIME_DROPPED_TASKS_HELP: &str = "The number of tasks dropped";
 
 /// Metrics for the runtime
 pub struct RuntimeMetrics {
--- a/rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
+++ b/rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
@@ -1,0 +1,66 @@
+use std::{fmt::Debug, time::Duration};
+
+use eyre::Result;
+use hyperlane_core::metrics::agent::METRICS_SCRAPE_INTERVAL;
+use prometheus::IntGauge;
+use tokio::{task::JoinHandle, time::MissedTickBehavior};
+use tokio_metrics::{TaskMetrics, TaskMonitor};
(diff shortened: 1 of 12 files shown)


rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
```diff
@@ -0,0 +1,64 @@
...
+        };
+        Ok(chain_metrics)
+    }
+
+    fn update(&mut self, metrics: TaskMetrics) {
+        self.dropped_tasks.add(metrics.dropped_count as i64);
+    }
+
+    /// Periodically updates the metrics
+    pub async fn start_updating_on_interval(mut self, period: Duration) {
```
daniel-savu, line 44: this is duplicated from `agent_metrics.rs` but couldn't find a nice refactoring with closures in 15 mins so gave up

rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
```diff
@@ -0,0 +1,64 @@
...
+            .with_label_values(&[]);
+        let chain_metrics = Self {
+            producer: task_monitor,
+            dropped_tasks,
+        };
+        Ok(chain_metrics)
+    }
+
+    fn update(&mut self, metrics: TaskMetrics) {
+        self.dropped_tasks.add(metrics.dropped_count as i64);
```
daniel-savu, line 40: this `.add`s rather than `.set`s to get the cumulative value over time

kamiyaa (approved):
Looks good :+1:

rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
ameten: I wonder why we choose `IntGauge` over `IntCounter`?

ameten (approved):

tkporter (approved):

rust/main/hyperlane-base/src/metrics/runtime_metrics.rs
daniel-savu: ah didn't know the gauge is signed while counter is unsigned, will change

PR #5491, opened 2025-02-17
refactor: move registry functions to @hyperlane-xyz/registry

mshojaei-txfusion:
### Description

This PR moves the registry-related functionality from the CLI package to the dedicated `@hyperlane-xyz/registry` package. The main changes include:

- Relocating `getRegistry()` function and its dependencies from CLI to registry package
- Updating imports across the codebase to use the new location
- Adding logger parameter to `getRegistry()` function for better logging control

### Drive-by changes

None

### Related issues

N/A

### Backward compatibility

Yes - This is a pure refactor with no functional changes

### Testing

Unit Tests - Existing tests were updated to use the new import paths

### DIFF
--- a/typescript/cli/src/index.ts
+++ b/typescript/cli/src/index.ts
@@ -1,1 +1,0 @@
-export { getRegistry } from './context/context.js';
--- a/.changeset/rude-wasps-do.md
+++ b/.changeset/rude-wasps-do.md
@@ -1,0 +1,6 @@
+---
+'@hyperlane-xyz/infra': minor
+'@hyperlane-xyz/cli': minor
+---
+
+Move getRegistry function from the CLI to `@hyperlane-xyz/registry` package.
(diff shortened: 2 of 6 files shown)


nambrot (approved):
Approving, but pending merge based on a registry release

ltyu (approved):

typescript/infra/package.json
```diff
@@ -15,7 +15,7 @@
     "@google-cloud/secret-manager": "^5.5.0",
     "@hyperlane-xyz/cli": "8.8.1",
```
ltyu, line 16: can prob remove this considering we no longer have anything to import :)

typescript/cli/src/index.ts
```diff
@@ -1 +0,0 @@
-export { getRegistry } from './context/context.js';
```
ltyu, line 1: can remove this file

mshojaei-txfusion: This is already a removed file!

typescript/infra/package.json
mshojaei-txfusion: Good catch, double checked, nowhere used, removed

PR #5492, opened 2025-02-17
fix: constrain max mint limit for super XERC20

yorhodes:
### Description

- Updates XERC20 collateral warp route adapter for [Super XERC20 variant](https://github.com/hyperlane-xyz/superchain-xerc20).
- Introduces a new token standard to represent this. 
- Adds logic for comparing to buffer cap limits. If the mint limit is decreasing, latency of message delivery can cause the transfer to appear mintable at dispatch time but not at process time. Thus we cap the mint limit at the buffer midpoint (where it will stop decreasing).

### Backward compatibility

Yes

### Testing

TODO

### DIFF
--- a/typescript/sdk/src/warp/WarpCore.ts
+++ b/typescript/sdk/src/warp/WarpCore.ts
@@ -493,9 +493,13 @@ export class WarpCore {
       destinationToken.standard === TokenStandard.EvmHypXERC20 ||
       destinationToken.standard === TokenStandard.EvmHypXERC20Lockbox
     ) {
-      destinationBalance = await (
-        adapter as IHypXERC20Adapter<unknown>
-      ).getMintLimit();
+      const adapter = destinationToken.getAdapter(
+        this.multiProvider,
+      ) as IHypXERC20Adapter<unknown>;
+      const mintLimit = await adapter.getMintLimit();
+      const mintMaxLimit = await adapter.getMintMaxLimit();
+      const max = mintMaxLimit / 2n;
+      destinationBalance = mintLimit < max ? mintLimit : max;
     } else {
       destinationBalance = await adapter.getBalance(
         destinationToken.addressOrDenom,
--- a/typescript/sdk/src/warp/WarpCore.ts
+++ b/typescript/sdk/src/warp/WarpCore.ts
@@ -491,15 +491,28 @@ export class WarpCore {
     const adapter = destinationToken.getAdapter(this.multiProvider);
     if (
       destinationToken.standard === TokenStandard.EvmHypXERC20 ||
-      destinationToken.standard === TokenStandard.EvmHypXERC20Lockbox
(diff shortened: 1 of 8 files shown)


typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -491,11 +491,28 @@ export class WarpCore {
...
+        this.multiProvider,
+      ) as IHypXERC20Adapter<unknown>;
+      destinationBalance = await adapter.getMintLimit();
+
+      if (
+        destinationToken.standard === TokenStandard.EvmHypVSXERC20 ||
+        destinationToken.standard === TokenStandard.EvmHypVSXERC20Lockbox
+      ) {
+        const bufferCap = await adapter.getMintMaxLimit();
+        const max = bufferCap / 2n;
```
Mo-Hussain, line 508: If the limit has not replenished to the midpoint and is above it i.e we have recently burned more than minted, we would be allowed to mint more than half the buffer cap at this point.

We should instead check if `destinationBalance` is greater than `getMintLimit()`

yorhodes: @nambrot wants to avoid this situation where the mint limit is decreasing and the latency of message delivery can cause the transfer to appear mintable at dispatch time but not at process time

yorhodes: otherwise we would not have made any changes to this adapter

Mo-Hussain: Makes sense

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -491,11 +491,28 @@ export class WarpCore {
...
-      destinationBalance = await (
-        adapter as IHypXERC20Adapter<unknown>
-      ).getMintLimit();
+      const adapter = destinationToken.getAdapter(
+        this.multiProvider,
+      ) as IHypXERC20Adapter<unknown>;
+      destinationBalance = await adapter.getMintLimit();
+
+      if (
+        destinationToken.standard === TokenStandard.EvmHypVSXERC20 ||
```
nambrot, line 504: How will this standard get set? Right now, it is not yeah? wouldn't we need a corresponding change in the warp core config artifact?

yorhodes: yes, the config artifact is only in the UI afaik?

yorhodes: this doesnt need to be urgently merged as @Xaroz already made the change on the UI
this just protects against onchain limit changes

yorhodes: I have asked @Xaroz to make this change and test with the UI.

Xaroz (approved):

PR #5495, opened 2025-02-17
feat: feb 17 chain deploy batch

paulbalaji:
### Description

feat: feb 17 chain deploy batch
- deploy to bouncebit, arcadia, ronin, sophon, story, subtensor

### Backward compatibility

yes

### Testing

manual
- known issue with sending from subtensor, will continue investigating but shouldn't block running our infra


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -67,12 +67,12 @@ export const hyperlaneContextAgentChainConfig: AgentChainConfig<
     appchain: true,
     arbitrum: true,
     arbitrumnova: true,
+    arcadia: true,
     artela: true,
     arthera: true,
     astar: true,
     astarzkevm: true,
     aurora: true,
-    flame: true,
```
paulbalaji, line 75: this moved to ensure the list is in alphabetical order

i think it's out of order because the name was originally `astaria` or `astariaflame`

.registryrc
```diff
@@ -1 +1 @@
-26e513e8fe567428cb58e7634e0ed8c659430707
+d230d7ec8a4bda00640d043202ea72aef6b77197
```
paulbalaji, line 1: update after merging https://github.com/hyperlane-xyz/hyperlane-registry/pull/596

Mo-Hussain (approved):

.registryrc
```suggestion
954b3b2982c7fe712b47bdc510b1414b3f27d2d7
```

tkporter (approved):

PR #5499, opened 2025-02-18
feat: Make number of worker threads in relayer equal to CPU requested

ameten:
### Description

Make number of worker threads in relayer equal to CPU requested.

We request 14 CPUs for Relayer pod. This change makes the number of threads in the tokio runtime equal to the number of CPUs we request for Relayer pod so that threads do not compete for CPU resources.

### Related issues

### Backward compatibility

Yes

### Testing

Will be tested in RC


rust/main/agents/relayer/src/main.rs
```diff
@@ -16,7 +16,9 @@ use relayer::Relayer;
 #[cfg(feature = "memory-profiling")]
 mod memory_profiler;
 
-#[tokio::main(flavor = "multi_thread", worker_threads = 20)]
+/// Number of worker threads matches the number of CPU requested for relayer
+/// https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4ddbd60e1f2e150e58dfeab72171c4fa88203db4/typescript/infra/config/environments/mainnet3/agent.ts#L595
+#[tokio::main(flavor = "multi_thread", worker_threads = 14)]
```
daniel-savu, line 21: I think we can just remove `worker_threads` as it will default to the number of cores for the architecture it's run on
https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#default
```suggestion
#[tokio::main(flavor = "multi_thread")]
```

ameten: Discussed in person.

Unfortunately, Relayer pod is not isolated in this manner. Currently, Relayer pod sees 32 CPUs while we limit to 14 CPUs. So, tokio will create 32 threads with the default setting.

from `lscpu`
```
CPU(s):                   32
  On-line CPU(s) list:    0-31
```

PR #5500, opened 2025-02-18
refactor: optional mailbox on warp config

mshojaei-txfusion:
### Description

This PR updates the warp route deployment configuration to handle mailbox addresses more consistently. Key changes:
- Makes mailbox optional in the WarpRouteDeployConfig schema
- Introduces a new WarpRouteDeployConfigMailboxRequired type for internal use
- Updates the config filling logic to handle mailbox addresses through the context
- Removes direct mailbox address prompting during config creation

### Drive-by changes

- Cleaned up some unused imports
- Improved type safety around mailbox requirements
- Updated tests to reflect new mailbox handling

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5258

### Backward compatibility

Yes - This change maintains backward compatibility while improving the internal handling of mailbox addresses.

### Testing

Unit Tests - Existing tests have been updated to accommodate the new mailbox handling approach


typescript/cli/src/config/warp.ts
```diff
@@ -96,15 +91,18 @@ async function fillDefaults(
...
     throw new Error(`No warp route deploy config found at ${filePath}`);
-  if (context) {
-    config = await fillDefaults(context, config as any);
-  }
-  return WarpRouteDeployConfigSchema.parse(config);
+
+  config = await fillDefaults(context, config as any);
+
+  //fillDefaults would have added a mailbox to the config if it was missing
+  return WarpRouteDeployConfigSchema.parse(
```
nambrot, line 103: Can we not use `WarpRouteDeployConfigMailboxRequired` to parse then to avoid the type cast?

mshojaei-txfusion: The thing for simplicity we did not define `WarpRouteDeployConfigMailboxRequired` schema and just defined the type.

```
export type WarpRouteDeployConfigMailboxRequired = z.infer<
  ReturnType<
    typeof WarpRouteDeployConfigSchema.and<typeof _RequiredMailboxSchema>
  >
>;
```

typescript/infra/config/warp.ts
```diff
@@ -125,6 +125,8 @@ async function getConfigFromMergedRegistry(
     true,
   ).getWarpDeployConfig(warpRouteId);
   assert(warpRoute, `Warp route Config not found for ${warpRouteId}`);
+  //TODO: fix this and manage mailbox
```
mshojaei-txfusion, line 128: @nambrot Can you please take a look at this?
should I handle mailbox on infra side or on registry itself?
So I need to do similar thing `fillDefaults` did on cli.

mshojaei-txfusion: As mailbox may not return when we call `getWarpDeployConfig` from registry

typescript/cli/src/config/warp.ts
ltyu: Can you use [omit](https://zod.dev/?id=pickomit) on the existing `WarpRouteDeployConfigSchema`?

something like:
```
WarpRouteDeployConfigSchema.omit('mailbox').parse(...)
```

ltyu: or maybe even [partial](https://zod.dev/?id=partial) to set mailbox to optional

typescript/infra/config/warp.ts
ltyu: I think the Registry should handle this, given that the mailbox address comes from the Registry

mshojaei-txfusion: Make sense, so we would return required mailbox deploy config type on registry.
And registry would get it from warp config in case provided, otherwise would get it from core contracts. 
can you both approve this?
@ltyu @nambrot

typescript/cli/src/config/warp.ts
mshojaei-txfusion: As we already discussed this was the issue with omit:
`Property 'omit' does not exist on type 'ZodIntersection...`

And for partial, we kind of already using something similar.

typescript/sdk/src/token/types.ts
```diff
@@ -150,8 +156,20 @@ export const WarpRouteDeployConfigSchema = z
   });
 export type WarpRouteDeployConfig = z.infer<typeof WarpRouteDeployConfigSchema>;
 
+const _RequiredMailboxSchema = z.record(
+  z.object({
+    mailbox: z.string(),
+  }),
+);
+
+export type WarpRouteDeployConfigMailboxRequired = z.infer<
```
ltyu, line 165: Prefer a schema here

typescript/cli/src/config/warp.ts
ltyu: isn't this not validating the mailbox because its no longer part of `WarpRouteDeployConfigSchema`?

typescript/sdk/src/token/types.ts
```diff
@@ -101,8 +101,14 @@ export const HypTokenRouterConfigSchema = HypTokenConfigSchema.and(
...
 
+const HypTokenRouterConfigMailboxOptionalSchema = HypTokenConfigSchema.and(
+  GasRouterConfigSchema.extend({
+    mailbox: z.string().optional(),
+  }),
+);
+
 export const WarpRouteDeployConfigSchema = z
-  .record(HypTokenRouterConfigSchema)
+  .record(HypTokenRouterConfigMailboxOptionalSchema)
```
ltyu, line 111: isn't the issue asking to remove the mailbox? this makes it optional. perhaps need clarification @nambrot

typescript/cli/src/config/warp.ts
ltyu: we also prefer schemas for runtime validation and then infer the types from them

typescript/infra/config/warp.ts
ltyu: on second thought, i am kind of leaning towards keeping that Registry function as is i.e. to fetch the non-derived. 

do the callers actually need the mailbox here?

typescript/sdk/src/token/types.ts
nambrot: Yeah I really mean optional, can be more precise going forward

typescript/cli/src/config/warp.ts
nambrot: Agree with @ltyu to have an explicit schema here, isn't the extra "cost" just 5 extra lines?

typescript/infra/config/warp.ts
nambrot: This is a great question, I'm inclined to change the return type here and then see where it breaks. Generally speaking would prefer to pass around the "simpler" version and then only "hydrate" the config when necessary

typescript/cli/src/config/warp.ts
mshojaei-txfusion: ok then make sense, done!

typescript/infra/config/warp.ts
mshojaei-txfusion: I believe this would be needed on deploy script of infra, so we really needed and should decide for the way we get it.

ltyu: yeah lets keep the Registry as is and then do some sort of filldefault() here

mshojaei-txfusion: got it, done!

nambrot (approved):
Would like explicit approval from @ltyu as well but lgtm

ltyu (approved):

paulbalaji (approved):

PR #5508, opened 2025-02-18
feat: CCIP and Open USDT warp route

nambrot:
### Description

For this push to support CCIP on the warp route, we had to manage a separate feature branch to ensure that `main` only has audited smart contract changes. Given that we have received Chainlight's audit, we can now merge these changes back into `main`

- [ ] Use merge instead of squash merging


.github/workflows/test.yml
```diff
@@ -3,7 +3,7 @@ name: test
 on:
   # Triggers the workflow on pushes to main branch, ignoring md files
   push:
-    branches: [main]
+    branches: [main, ccip-warp-route]
```
yorhodes, line 6: I guess revert this

yorhodes (approved):

PR #5518, opened 2025-02-19
feat: add SOL/hyperevm-solanamainnet

tkporter:
### Description

- Adds SOL/hyperevm-solanamainnet
- Tested in UI
- Includes hyperevm in the Solana IGP now, and did a drive-by quote update on Solana
- Includes https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5519


rust/sealevel/environments/mainnet3/gas-oracle-configs.json
paulbalaji, line 1: how can i validate that these changes are legit? is there a log somewhere of expected tx costs?

rust/sealevel/environments/mainnet3/chain-config.json
paulbalaji, line 1: why can we not reuse the mainnet_config json here?

paulbalaji (commented):
seems ok to me but couple qs

rust/sealevel/environments/mainnet3/gas-oracle-configs.json
tkporter: No logs 😅 but this is something that we'll do https://github.com/hyperlane-xyz/issues/issues/1422

PR #5519, opened 2025-02-19
feat: promote solanamainnet rc ISM to default ISM

tkporter:
### Description

- Promotes the updated `rc` ISM to be the default
- Will change the warp routes to use the default ISM that we currently still own and are using this instead of the default ISM later
- There is an accompanying multisig tx


tkporter:
Merging as the multisig tx was executed

PR #5520, opened 2025-02-19
feat: Support extra lockboxes in config

nambrot:
### Description

Expands the config to support the configuration of extra lockboxes


typescript/sdk/src/token/types.ts
```diff
@@ -61,7 +61,15 @@ export type XERC20LimitConfig = z.infer<typeof xERC20LimitConfigSchema>;
 const xERC20TokenMetadataSchema = z.object({
   xERC20: z
     .object({
-      limits: xERC20LimitConfigSchema,
+      warpRouteLimits: xERC20LimitConfigSchema,
```
paulbalaji, line 64: the rename from limits -> warpRouteLimits is breaking atm

PR #5528, opened 2025-02-20
feat(limits): Handle safe transaction scenarios

Mo-Hussain:
### Description
- Generally handle Safe or Signer transactions properly
- Throw when safe tx service is not available
- Specifically fallback to Signer tx when the expected owner is a safe in the config but the actual owner is the deployer (Signer), this special case occurs post deployment before ownership transfers are applied 

### Testing

Manual

### DIFF
--- a/typescript/infra/src/xerc20/utils.ts
+++ b/typescript/infra/src/xerc20/utils.ts
@@ -63,7 +63,6 @@ export async function addBridgeToChain({
     bridgeAddress,
     bufferCap,
     rateLimitPerSecond,
-    owner: expectedOwner,
     decimals,
   } = bridgeConfig;
 
@@ -126,7 +125,6 @@ export async function addBridgeToChain({
       await sendTransactions(
         envMultiProvider,
         chain,
-        expectedOwner,
         [tx],
         xERC20Address,
         bridgeAddress,
@@ -161,7 +159,6 @@ export async function updateChainLimits({
 }) {
   const {
     bridgeAddress,
-    owner,
     bufferCap,
     rateLimitPerSecond,
     decimals,
@@ -209,7 +206,6 @@ export async function updateChainLimits({
     await sendTransactions(
       envMultiProvider,
       chain,
-      owner,
       txsToSend,
       xERC20Address,
       bridgeAddress,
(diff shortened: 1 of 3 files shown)


typescript/infra/src/xerc20/utils.ts
```diff
@@ -279,24 +286,87 @@ async function prepareRateLimitTx(
...
+  let safeService;
+  try {
+    safeService = getSafeService(chain, multiProvider);
+  } catch (error) {
+    rootLogger.error(
+      chalk.red(
+        `[${chain}][${bridgeAddress}] Safe service not available, cannot send safe transactions, please add the safe service url to registry and try again.`,
+      ),
+    );
+    throw { chain, error };
```
Mo-Hussain, line 365: throw if safe tx service not set and avoid silently falling back to Signer txs @nambrot

typescript/infra/src/xerc20/utils.ts
```diff
@@ -376,35 +446,81 @@ function getTxCallData(transactions: PopulatedTransaction[]): CallData[] {
 async function sendTransactions(
   multiProvider: MultiProvider,
   chain: string,
-  owner: Address,
+  expectedOwner: Address,
```
nambrot, line 449: Is this owned?

Mo-Hussain: Not used, good spot

typescript/infra/src/xerc20/utils.ts
```diff
@@ -279,24 +286,87 @@ async function prepareRateLimitTx(
   );
 }
 
-async function checkSafeOwner(
+async function checkOwnerIsSafe(
   proposer: Address,
   chain: string,
   multiProvider: MultiProvider,
   safeAddress: Address,
```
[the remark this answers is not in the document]
Mo-Hussain: Renamed to owner as it is the actual owner

typescript/infra/src/xerc20/utils.ts
```diff
@@ -314,7 +380,7 @@ async function sendAsSafeMultiSend(
 
   try {
     const safeMultiSend = new SafeMultiSend(multiProvider, chain, safeAddress);
-
+    // TODO: SafeMultiSend.sendTransactions does not wait for the receipt
```
nambrot, line 383: What is this todo?

typescript/infra/src/xerc20/utils.ts
```diff
@@ -376,35 +442,80 @@ function getTxCallData(transactions: PopulatedTransaction[]): CallData[] {
...
     multiProvider,
-    owner,
+    actualOwner,
+    bridgeAddress,
   );
 
-  if (isSafeOwner) {
-    await sendAsSafeMultiSend(
+  if (isOwnerSafe) {
+    const isSafeProposer = await checkSafeProposer(
```
nambrot, line 472: Probably should keep for a followup PR, but this seems rather ineffient, as we will call `getSafe` twice

nambrot (approved):

typescript/infra/src/xerc20/utils.ts
Mo-Hussain: I want to look into how much of a lift it would be to wait for a receipt on safe propsal tx

PR #5529, opened 2025-02-20
feat: support ownership transfers in the SVM warp route tooling, finalize Sonic deploy

tkporter:
### **Note!**
This includes https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5575. I'd recommend just looking at the `rust/` changes when reviewing

- Supports changing the owner and upgrade authority (assumed to be the same address) in the SVM warp route tooling
- So when running e.g. the following, it'll also idempotently transfer ownership / set the program upgrade authority. It defaults to the payer as the owner.
```
cargo run -- -k ~/solana-devnet-deployer-keypair.json warp-route deploy --warp-route-name SOL-solanatestnet-sonicsvmtestnet --environment testnet4 --environments-dir ../environments --built-so-dir ../../target/deploy --token-config-file ../environments/testnet4/warp-routes/SOL-solanatestnet-sonicsvmtestnet/token-config.json  --chain-config-file ../environments/testnet4/chain-config.json --ata-payer-funding-amount 1000000
```
- I haven't gone back to update the configs for all the other deploys we did that had manual ownership transfer. Hoping to do something like this soon

### Drive-by changes

- none

### Related issues

- none

### Backward compatibility

- backward compatible

### Testing

- tried it out on a testnet warp deploy, worked well


rust/sealevel/client/src/router.rs
```diff
@@ -555,6 +566,82 @@ fn configure_owner(
...
+    }
+
+    // The program id must actually be a program
+    let programdata_address = if let Ok(UpgradeableLoaderState::Program {
+        programdata_address,
+    }) = program_account.state()
+    {
+        programdata_address
+    } else {
+        return;
```
aroralanuk, line 592: nit: can we have an UpgradeWarning enum we return with the specified reasons here, or is it a bit unnecessary?

aroralanuk, line 592: nit: can we have an UpgradeWarning enum we return with the specified reasons here, or is it a bit unnecessary?

rust/sealevel/client/src/router.rs
```diff
@@ -555,6 +566,82 @@ fn configure_owner(
...
+    ctx: &mut Context,
+    program_id: &Pubkey,
+    router_config: &RouterConfig,
+    chain_config: &ChainMetadata,
+) {
+    let client = chain_config.client();
+
+    let program_account = client.get_account(program_id).unwrap();
+    // If the program isn't upgradeable, exit
+    if program_account.owner == bpf_loader_upgradeable::id() {
```
aroralanuk, line 581: not too sure about solana specific stuff, but I would think equality would mean the program is upgradable

rust/sealevel/client/src/router.rs
```diff
@@ -555,6 +566,82 @@ fn configure_owner(
...
+    }) = program_account.state()
+    {
+        programdata_address
+    } else {
+        return;
+    };
+
+    let program_data_account = client.get_account(&programdata_address).unwrap();
+
+    // The program data account must actually a program data account
```
aroralanuk, line 597: ?

rust/sealevel/client/src/router.rs
```diff
@@ -555,6 +566,82 @@ fn configure_owner(
...
+                ),
+                format!(
+                    "Setting upgrade authority for chain: {} ({}) to {:?}",
+                    chain_config.name,
+                    chain_config.domain_id(),
+                    expected_upgrade_authority,
+                ),
+            )
+            .with_client(&client)
+            .send_with_payer();
```
aroralanuk, line 641: Is there a case where this fails silently and we don't end up upgrading? Maybe we can check if the upgrade authority is set correctly after this tx?

aroralanuk (approved):
looks idempotent to me

rust/sealevel/client/src/router.rs
```diff
@@ -540,6 +541,16 @@ fn configure_owner(
...
     if actual_owner != expected_owner {
+        if actual_owner != Some(ctx.payer_pubkey) {
+            println!(
+                "WARNING: Ownership transfer cannot be completed for chain: {} ({}) from {:?} to {:?}, the existing owner is not the payer account",
+                chain_config.name,
+                chain_config.domain_id(),
+                actual_owner,
+                expected_owner,
+            );
+            return;
```
ameten, line 552: If we exit `configure_owner` here, we don't need to run `configure_upgrade_authority`, correct? I wonder if we can  return some data here so that we can decide if we need to run `configure_upgrade_authority`.

rust/sealevel/client/src/router.rs
tkporter: Imo the quality bar for the SVM tooling is far below that 😅

rust/sealevel/client/src/router.rs
tkporter: this was a bug, great call

rust/sealevel/client/src/router.rs
tkporter: I'd expect things to fail if it was unsuccessful

PR #5530, opened 2025-02-20
fix: Specify chunk size for Lumia Prism chain

ameten:
### Description

Specify chunk size for Lumia Prism chain

### Backward compatibility

Yes

### Testing

Fix for failing RPC calls. Tested with direct change into Relayer RC.

### DIFF
--- a/rust/main/config/mainnet_config.json
+++ b/rust/main/config/mainnet_config.json
@@ -5353,7 +5353,8 @@
       ],
       "technicalStack": "polygoncdk",
       "index": {
-        "from": 1923136
+        "from": 1923136,
+        "chunk": 1000
       },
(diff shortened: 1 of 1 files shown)


rust/main/config/mainnet_config.json
```diff
@@ -6211,7 +6211,8 @@
       "timelockController": "0x0000000000000000000000000000000000000000",
       "validatorAnnounce": "0x803d2A87429E20d4B52266bF97Ca1c7e4f4F5Dfa",
       "index": {
-        "from": 3144538
+        "from": 3144538,
+        "chunk": 1000
```
daniel-savu, line 6215: wondering if we should update this for the `lumia` entry as well, since I see it's using the same RPC

ameten: Yes, we should

daniel-savu (approved):

tkporter (approved):

PR #5534, opened 2025-02-20
feat: update warp route monitor for extra lockboxes

xeno097:
### Description

Updates the infra monitor to monitor the limits and balances of extra lockboxes

### Backward compatibility

- YES

### Testing

- Manual

### DIFF
--- a/typescript/infra/config/registry.ts
+++ b/typescript/infra/config/registry.ts
@@ -13,6 +13,7 @@ import {
   ChainMetadata,
   ChainName,
   WarpCoreConfig,
+  WarpRouteDeployConfig,
   getDomainId as resolveDomainId,
   getReorgPeriod as resolveReorgPeriod,
 } from '@hyperlane-xyz/sdk';
@@ -99,6 +100,20 @@ export function getWarpCoreConfig(warpRouteId: string): WarpCoreConfig {
   if (!warpRouteConfig) {
     throw new Error(
       `Warp route config for ${warpRouteId} not found in registry`,
+    );
+  }
+  return warpRouteConfig;
+}
+
+export function getWarpDeployConfig(
+  warpRouteId: string,
+): WarpRouteDeployConfig {
+  const registry = getRegistry();
+  const warpRouteConfig = registry.getWarpDeployConfig(warpRouteId);
+
+  if (!warpRouteConfig) {
+    throw new Error(
+      `Warp route deploy config for ${warpRouteId} not found in registry`,
     );
   }
   return warpRouteConfig;
--- a/typescript/infra/scripts/warp-routes/monitor/metrics.ts
+++ b/typescript/infra/scripts/warp-routes/monitor/metrics.ts
@@ -61,12 +61,13 @@ export function updateTokenBalanceMetrics(
   token: Token,
   balanceInfo: WarpRouteBalance,
   collateralTokenSymbol: string,
+  extraLockbox?: Address,
 ) {
   const metrics: WarpRouteMetrics = {
     chain_name: token.chainName,
     token_address: token.collateralAddressOrDenom || token.addressOrDenom,
     token_name: token.name,
-    wallet_address: token.addressOrDenom,
+    wallet_address: extraLockbox ?? token.addressOrDenom,
     token_standard: token.standard,
     warp_route_id: createWarpRouteConfigId(
       collateralTokenSymbol,
--- a/typescript/infra/scripts/warp-routes/monitor/metrics.ts
+++ b/typescript/infra/scripts/warp-routes/monitor/metrics.ts
@@ -119,7 +119,7 @@ export function updateNativeWalletBalanceMetrics(balance: NativeWalletBalance) {
 export function updateXERC20LimitsMetrics(
   token: Token,
   limits: XERC20Limit,
-  bridgeAddress?: Address,
+  bridgeAddress: Address,
 ) {
   for (const [limitType, limit] of Object.entries(limits)) {
     xERC20LimitsGauge
--- a/typescript/infra/scripts/warp-routes/monitor/metrics.ts
+++ b/typescript/infra/scripts/warp-routes/monitor/metrics.ts
@@ -2,6 +2,7 @@ import { Gauge, Registry } from 'prom-client';
 
 import { createWarpRouteConfigId } from '@hyperlane-xyz/registry';
 import { ChainName, Token, TokenStandard, WarpCore } from '@hyperlane-xyz/sdk';
+import { Address } from '@hyperlane-xyz/utils';
 
 import { getWalletBalanceGauge } from '../../../src/utils/metrics.js';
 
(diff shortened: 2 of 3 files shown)


typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -143,9 +164,63 @@ async function updateTokenMetrics(
     promises.push(
       tryFn(async () => {
         const limits = await getXERC20Limits(warpCore, token);
-        updateXERC20LimitsMetrics(token, limits);
+
+        const tokenAddress =
```
nambrot, line 168: We really do have this logic everywhere @yorhodes

nambrot (approved):
Personally lgtm, but i"m not as familiar with the code, so probably get @Mo-Hussain to also look at this

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -143,9 +164,63 @@ async function updateTokenMetrics(
...
+        tryFn(async () => {
+          const balance = await getExtraLockboxBalance(
+            token,
+            warpCore.multiProvider,
+            tokenPriceGetter,
+            lockbox.lockbox,
+          );
+
+          if (balance) {
+            updateTokenBalanceMetrics(
```
Mo-Hussain, line 213: Having some doubts now as to whether capturing this balance as a `hyperlane_warp_route_token_balance` is the right approach.

The intention is to make sure this collateral is accounted for in our warp diff alerts but technically this collateral is not in a token router.

We will have to change the implementation of `updateTokenBalanceMetrics` or probably define a new function as the currently we will set the `token_standard` label to synthetic when it should be collateral

Curious if you have thought or can foresee issues with approach

Mo-Hussain: @tkporter

Mo-Hussain: Looking at the labels for `hyperlane_warp_route_token_balance`

```
interface WarpRouteMetrics {
  chain_name: ChainName;
  token_address: string;
  token_name: string;
  wallet_address: string;
  token_standard: TokenStandard;
  warp_route_id: string;
  related_chain_names: string;
}
```

given that the collateral in this lockbox will a different token

token_address, token_name will currently be set incorrectly currently . 

token_standard is a tricky one as this extraLockbox contract is not a Token Router. We have `EvmHypXERC20Lockbox` which we could use but this is a fudge

Mo-Hussain: Looking at the easy and warp routes dashboards and the `All Warp Routes: Collateral and Synthetic amounts don't match` alert, we rely on the token_standard and warp_route_id to collect the balances to compare collateral vs synthetic diffs.

I think `EvmHypXERC20Lockbox` will be ok for now, we could also define a new standard eventually that better represents this contract

typescript/infra/scripts/warp-routes/monitor/metrics.ts
```diff
@@ -52,20 +53,21 @@ const xERC20LimitsGauge = new Gauge({
   name: 'hyperlane_xerc20_limits',
   help: 'Current minting and burning limits of xERC20 tokens',
   registers: [metricsRegister],
-  labelNames: ['chain_name', 'limit_type', 'token_name'],
+  labelNames: ['chain_name', 'limit_type', 'token_name', 'bridge_address'],
```
Mo-Hussain, line 56: Adding the xerc20 token address will be helpful also

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
tkporter: Mo and I chatted and I agree with token_address and token_name being incorrect and we should probably fix. Iiuc Mo will add this and make a view call to get the token name

nambrot: Agree that `EvmHypXERC20Lockbox` is fine

typescript/infra/scripts/warp-routes/monitor/metrics.ts
Mo-Hussain: We actually dont have a label with warp id but even if we did we would have multiple sets of limits on the same chain for the same warp id. bridge address is the only differentiator

typescript/infra/config/registry.ts
```diff
@@ -104,6 +105,20 @@ export function getWarpCoreConfig(warpRouteId: string): WarpCoreConfig {
...
 }
 
+export function getWarpDeployConfig(
+  warpRouteId: string,
+): WarpRouteDeployConfig {
+  const registry = getRegistry();
+  const warpRouteConfig = registry.getWarpDeployConfig(warpRouteId);
+
+  if (!warpRouteConfig) {
+    throw new Error(
```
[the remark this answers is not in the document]
xeno097: Ok, you're right! I'll change that

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -143,9 +164,63 @@ async function updateTokenMetrics(
...
+          token.collateralAddressOrDenom
+            ? token.collateralAddressOrDenom
+            : token.addressOrDenom;
+
+        updateXERC20LimitsMetrics(token, limits, tokenAddress);
       }, 'Getting xERC20 limits'),
     );
+
+    // Check if the current xERC20 has also extra lockboxes
+    warpDeployConfig[token.chainName].type === TokenType.XERC20 ||
```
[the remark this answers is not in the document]
xeno097: Noup, removing it

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
```diff
@@ -261,10 +336,89 @@ async function getXERC20Limit(
...
-    burn: formatBigInt(burnCurrent),
-    burnMax: formatBigInt(burnMax),
+    mint: formatBigInt(token, mintCurrent),
+    mintMax: formatBigInt(token, mintMax),
+    burn: formatBigInt(token, burnCurrent),
+    burnMax: formatBigInt(token, burnMax),
+  };
+}
+
+const managedLockBoxMinimalABI = [
```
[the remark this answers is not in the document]
xeno097: updated it didn't notice that we had a `IXERC20Lockbox` interface so i replaced this using the interface

typescript/infra/config/registry.ts
```diff
@@ -104,6 +105,20 @@ export function getWarpCoreConfig(warpRouteId: string): WarpCoreConfig {
   return warpRouteConfig;
 }
 
+export function getWarpDeployConfig(
```
tkporter, line 108: seems like we can remove this now

tkporter (approved):

PR #5536, opened 2025-02-21
feat: xerc20 safe tx submission improvements

paulbalaji:
### Description

feat: xerc20 safe tx submission improvements
- use getSafeAndService infra util where possible instead of SDK directly (to have retry logic)
- refactor sendTransaction logic to better leverage MultiSend
- add --chains filter

image

### Testing

manual

### DIFF
--- a/typescript/infra/scripts/xerc20/xerc20vs-add-bridge.ts
+++ b/typescript/infra/scripts/xerc20/xerc20vs-add-bridge.ts
@@ -10,8 +10,8 @@ import {
 import {
   addBridgeToChain,
   deriveBridgesConfig,
+  getAndValidateBridgesToUpdate,
   getWarpConfigsAndArtifacts,
-  validateConfigChains,
 } from '../../src/xerc20/utils.js';
 import {
   getArgs,
@@ -42,11 +42,11 @@ async function main() {
 
   // if chains are provided, validate that they are in the warp config
   // throw an error if they are not
-  validateConfigChains(chains, bridgesConfig);
+  const bridgesToUpdate = getAndValidateBridgesToUpdate(chains, bridgesConfig);
 
   const erroredChains: string[] = [];
 
-  for (const [_, bridgeConfig] of Object.entries(bridgesConfig)) {
+  for (const bridgeConfig of bridgesToUpdate) {
     try {
       await addBridgeToChain({
         chain: bridgeConfig.chain,
--- a/typescript/infra/scripts/xerc20/xerc20vs-add-bridge.ts
+++ b/typescript/infra/scripts/xerc20/xerc20vs-add-bridge.ts
@@ -11,6 +11,7 @@ import {
   addBridgeToChain,
   deriveBridgesConfig,
   getWarpConfigsAndArtifacts,
+  validateConfigChains,
 } from '../../src/xerc20/utils.js';
 import {
   getArgs,
@@ -39,19 +40,9 @@ async function main() {
     envMultiProvider,
   );
 
-  const configChains = Object.keys(bridgesConfig);
-  if (chains) {
-    const missingChains = chains.filter(
-      (chain) => !configChains.includes(chain),
-    );
-    if (missingChains.length > 0) {
-      throw new Error(
-        `The following chains are not in the provided warp config: ${missingChains.join(
-          ', ',
-        )}`,
-      );
-    }
-  }
+  // if chains are provided, validate that they are in the warp config
+  // throw an error if they are not
+  validateConfigChains(chains, bridgesConfig);
 
   const erroredChains: string[] = [];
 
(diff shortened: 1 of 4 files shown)


typescript/infra/scripts/xerc20/xerc20vs-add-bridge.ts
```diff
@@ -38,6 +39,20 @@ async function main() {
     envMultiProvider,
   );
 
+  const configChains = Object.keys(bridgesConfig);
+  if (chains) {
+    const missingChains = chains.filter(
```
Mo-Hussain, line 44: nit: can we rename this to some thing like `nonConfigChains`

typescript/infra/scripts/xerc20/xerc20vs-set-limits.ts
```diff
@@ -37,6 +38,20 @@ async function main() {
     envMultiProvider,
   );
 
+  const configChains = Object.keys(bridgesConfig);
```
Mo-Hussain, line 41: Could make this a helper

typescript/infra/src/xerc20/utils.ts
```diff
@@ -283,17 +286,13 @@ async function prepareRateLimitTx(
...
   multiProvider: MultiProvider,
   owner: Address,
   bridgeAddress: Address,
 ): Promise<boolean> {
-  // check if safe service is available
-  await getSafeTxService(chain, multiProvider, bridgeAddress);
-
   try {
-    await getSafe(chain, multiProvider, owner);
+    await getSafeAndService(chain, multiProvider, owner);
```
Mo-Hussain, line 295: The problem with this approach is that we wont know if the `gnosisSafeTransactionServiceUrl` is not set right? We will throw here and assume that the owner is not a safe. This caught me out initially. 

That's why I defined `getSafeTxService` 

On a separate note, looking at the definition of `getSafeAndService`

```
export async function getSafeAndService(
  chain: ChainNameOrId,
  multiProvider: MultiProvider,
  safeAddress: Address,
) {
  const safeSdk: Safe.default = await retryAsync(
    () => getSafe(chain, multiProvider, safeAddress),
    5,
    1000,
  );
  const safeService: SafeApiKit.default = getSafeService(chain, multiProvider);
  return { safeSdk, safeService };
}
```

I think we should call
```
const safeService: SafeApiKit.default = getSafeService(chain, multiProvider);
```
first, that way we through if the safe tx url is not set we throw before trying to get the safe which will always fail

typescript/infra/scripts/xerc20/xerc20vs-set-limits.ts
Hugoo: 👍  yes there is duplicated code with `xerc20vs-add-bridge.ts`

typescript/infra/src/xerc20/utils.ts
```diff
@@ -482,39 +438,44 @@ async function sendTransactions(
           `[${chain}][${bridgeAddress}] Signer ${signerAddress} is not a proposer on Safe (${actualOwner}), cannot submit safe transaction. Exiting...`,
         ),
       );
-      throw new Error('Signer is not a safe proposer');
```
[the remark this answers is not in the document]
Mo-Hussain: Hmm I guess there may be another EOA that we own that can propose on this safe, something that is not the deployer

typescript/infra/scripts/xerc20/xerc20vs-set-limits.ts
paulbalaji: i assumed we were gonna merge the two scripts together at some point so didn't do a helper

typescript/infra/src/xerc20/utils.ts
paulbalaji: that's fine with me - i just don't want to be in a place where we're calling the sdk fns directly without the retry logic

typescript/infra/src/xerc20/utils.ts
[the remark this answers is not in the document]
paulbalaji: imo if we end up in a place where we have to do manual submission, then something has gone wrong in the process. therefore i'd usually prefer to always log regardless of the failure mode, and then the onus is on the submitter to ensure they have correctly identified and submitted to the right place/with the right proposer

typescript/infra/src/xerc20/utils.ts
paulbalaji: nah switching the order should be enough, because we throw from the SDK if the tx service URL isn't set 
image

the real problem is swallowing the error without logging

so this below should be enough to surface the error imo 
image

Mo-Hussain: From that message you cannot distinguish whether there is no `gnosisSafeTransactionServiceUrl` set or that the address is not a safe, it ambiguous right?

Mo-Hussain: Seen your change nm

Mo-Hussain (approved):

typescript/infra/src/xerc20/utils.ts
paulbalaji: ah yeah should've updated my comment - this is what the output looks like 
image

PR #5538, opened 2025-02-21
feat: remove bridge if bufferCap is set to 0

nambrot:
### Description

Removes a bridge when the buffer cap is set to 0

### DIFF
--- a/typescript/infra/src/xerc20/utils.ts
+++ b/typescript/infra/src/xerc20/utils.ts
@@ -189,21 +189,24 @@ export async function updateChainLimits({
       xERC20Adapter,
       bridgeAddress,
     );
-    if (!dryRun) {
-      await sendTransactions(
-        envMultiProvider,
-        chain,
-        [removeBridgeTx],
-        xERC20Address,
-        bridgeAddress,
-      );
-    } else {
+
+    if (dryRun) {
       rootLogger.info(
         chalk.gray(
           `[${chain}][${bridgeAddress}] Dry run, no transactions sent, exiting...`,
         ),
       );
+      return;
     }
+
+    await sendTransactions(
+      envMultiProvider,
+      chain,
+      [removeBridgeTx],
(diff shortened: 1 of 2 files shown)


typescript/infra/src/xerc20/utils.ts
```diff
@@ -174,6 +174,39 @@ export async function updateChainLimits({
...
+        bridgeAddress,
+      );
+    } else {
+      rootLogger.info(
+        chalk.gray(
+          `[${chain}][${bridgeAddress}] Dry run, no transactions sent, exiting...`,
+        ),
+      );
+    }
+    return;
```
Hugoo, line 207: i suggest to use early return and avoid use negative logic (it is just easier to read)


```suggestion

    if (dryRun) {
      rootLogger.info(
        chalk.gray(
          `[${chain}][${bridgeAddress}] Dry run, no transactions sent, exiting...`,
        ),
      );
      return;
    }

    await sendTransactions(
        envMultiProvider,
        chain,
        [removeBridgeTx],
        xERC20Address,
        bridgeAddress,
      );
```

something in that direction

typescript/infra/src/xerc20/utils.ts
```diff
@@ -174,6 +174,39 @@ export async function updateChainLimits({
     bufferCap: currentBufferCap,
   } = await xERC20Adapter.getRateLimits(bridgeAddress);
 
+  // if buffer cap is 0, remove the bridge
+  if (bufferCap === 0) {
```
Hugoo, line 178: maybe early return here too?

```
if (bufferCap !== 0) {
  return;
}
```

nambrot:
@Mo-Hussain yes we have to remove the bridge because the contract won't let us set the limits to 0, we tried that and 1 wei

Mo-Hussain (approved):

paulbalaji (approved):

PR #5539, opened 2025-02-21
feat: check the right proxy admin for xerc20 token

nambrot:
### Description

Right now we are checking ownables in the infra checker for the collateral token (and its proxy admin). However, we want to be testing the xerc20 owners.

### DIFF
--- a/typescript/sdk/src/token/checker.ts
+++ b/typescript/sdk/src/token/checker.ts
@@ -54,8 +54,6 @@ export class HypERC20Checker extends ProxiedRouterChecker<
 
       // XERC20s are Ownable
       const expectedConfig = this.configMap[chain];
-      const tokenType = expectedConfig.type;
-
       if (expectedConfig.type === TokenType.XERC20Lockbox) {
         const lockbox = IXERC20Lockbox__factory.connect(
           expectedConfig.token,
@@ -71,7 +69,7 @@ export class HypERC20Checker extends ProxiedRouterChecker<
         );
       }
 
-      if (tokenType === TokenType.XERC20) {
+      if (expectedConfig.type === TokenType.XERC20) {
         contracts['collateralToken'] = Ownable__factory.connect(
           collateralToken.address,
           provider,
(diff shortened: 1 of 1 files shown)


Mo-Hussain (approved):

Mo-Hussain:
@yorhodes - nam fixed the issue we discussed yday

typescript/sdk/src/token/checker.ts
```diff
@@ -48,16 +48,30 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
       // XERC20s are Ownable
-      const tokenType = this.configMap[chain].type;
-      if (
-        tokenType === TokenType.XERC20Lockbox ||
-        tokenType === TokenType.XERC20
-      ) {
+      const expectedConfig = this.configMap[chain];
+      const tokenType = expectedConfig.type;
+
+      if (expectedConfig.type === TokenType.XERC20Lockbox) {
```
Hugoo, line 59:
```suggestion
      if (tokenType === TokenType.XERC20Lockbox) {
```

typescript/sdk/src/token/checker.ts
```diff
@@ -48,16 +48,30 @@ export class HypERC20Checker extends ProxiedRouterChecker<
...
+          await lockbox.callStatic['XERC20()'](),
+          provider,
+        );
+        contracts['collateralToken'] = Ownable__factory.connect(
+          collateralToken.address,
+          provider,
+        );
+      }
+
+      if (tokenType === TokenType.XERC20) {
```
Hugoo, line 74: maybe using a switch statement over 2 if would be better

typescript/sdk/src/token/checker.ts
nambrot: Agreed, changing it to the former

typescript/sdk/src/token/checker.ts
nambrot: Keeping it like this for now, this probably needs to be refactored anyways

Issue #5452, opened 2025-02-12
Multicast sending Ethereum transactions


ameten:
* any RPCs we make other than sending a tx, we still have normal fallback provider
* any time eth_sendRawTransaction is sent, multicast it and require at least one success

PR #5540, opened 2025-02-21
fix: Make only one retrying attempt if we get non-retryable error

ameten:
### Description

Make only one retrying attempt if we get non-retryable error.

At the moment, we are doing 12 RPC calls even if all of them returned a non-retryable error in multicast approach of EthereumFallbackProvider. While we still would like to wait for all providers to return their result when we do multicast, we would like to avoid retrying since it produces a lot of requests with the same failure result. It leads to elevated RPC failure metrics and higher number of alerts.

### Related issues

- Contributes into https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5452

### Backward compatibility

Yes

### Testing

TBD

### DIFF
--- a/rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
+++ b/rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
@@ -132,6 +132,9 @@ where
         // errors reported by providers
         let mut errors = vec![];
 
+        // non-retryable error from previous attempt
+        let mut non_retriable = None;
+
         // retry 4 times if all providers returned a retryable error
         for i in 0..=3 {
             if i > 0 {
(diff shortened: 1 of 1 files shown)


rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -159,6 +166,12 @@ where
 
                 return Ok(value);
             }
+
+            // if we are here, it means that all providers failed
+            // if one of the errors was non-retryable, we stop doing retrying attempts
+            if let Some(e) = non_retriable {
+                return Err(e.into());
```
daniel-savu, line 173: should we still collect all the errors (not just the first one) and return `Err(FallbackError::AllProvidersFailed(errors).into())` here too?

so we could gather the non_retriable errors into a separate array, and if that array is nonempty at this point, we `break`

Then at the end of the funciton, instead of returning `Err(FallbackError::AllProvidersFailed(errors).into())` we return something like
```rust
let all_errors = retryable_errors.extend(nonretryable_errors);
Err(FallbackError::AllProvidersFailed(all_errors).into())
```

daniel-savu (approved):
sweet

PR #5541, opened 2025-02-21
fix: optimize hook module read in warp apply process by using provide…

mshojaei-txfusion:
[temp: Just to have a stacked PR]

### DIFF
--- a/typescript/sdk/src/hook/EvmHookModule.ts
+++ b/typescript/sdk/src/hook/EvmHookModule.ts
@@ -129,6 +129,7 @@ export class EvmHookModule extends HyperlaneModule<
     this.txOverrides = multiProvider.getTransactionOverrides(this.chain);
   }
 
+  //optimize to not fetch the config from the chain in case actual config is provided
   public async read(): Promise<HookConfig> {
     return typeof this.args.config === 'string'
       ? this.args.addresses.deployedHook


typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -132,7 +129,7 @@ export class EvmHookModule extends HyperlaneModule<
   public async read(): Promise<HookConfig> {
     return typeof this.args.config === 'string'
       ? this.args.addresses.deployedHook
-      : this.reader.deriveHookConfig(this.args.addresses.deployedHook);
+      : { ...this.args.config };
```
ltyu, line 132: we generally still want the Modules to derive onchain values. however, i do see your intention. 

the challenge here is that we need some way to force `update()` to derive, but have another way to detect if `this.args.config` is an already derived config. 

One way we can do this is by checking if the field `address` exists. This should mean that it's a [DerivedHookConfig](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/719d022ec18c5d310e50fa0a49a31788bb6b1820/typescript/sdk/src/hook/EvmHookReader.ts#L50)

ltyu (approved):
good progress

typescript/sdk/src/hook/EvmHookModule.ts
mshojaei-txfusion: I see, can you take another look and confirm I understood correctly?

.changeset/eighty-planets-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Optimize HookModule.read() by removing redundant hook config derivation
```
ltyu, line 5: needs update?

ltyu (commented):
lgtm

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -130,9 +131,17 @@ export class EvmHookModule extends HyperlaneModule<
   }
 
   public async read(): Promise<HookConfig> {
-    return typeof this.args.config === 'string'
-      ? this.args.addresses.deployedHook
-      : this.reader.deriveHookConfig(this.args.addresses.deployedHook);
+    const { config, addresses } = this.args;
```
nambrot, line 134: I know you didn't create this, but can you add some comments to both the class, and the update and read function so that consumers know how to think about the semantics and API? It's not super obvious to me why you are able to just skip the hook derivation. IIUC, its that under only some circumstances we pass the actual config in, and the module can just assume that as the result of the read? If that is the case, imo the api has to be made pretty clear, when I call `.read`, i would expect that the module actually reads. Probably worth adding tests here if they dont exist cc @ltyu

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -147,13 +156,13 @@ export class EvmHookModule extends HyperlaneModule<
...
     }
 
-    // Update the config
-    this.args.config = targetConfig;
-
     // We need to normalize the current and target configs to compare.
     const normalizedCurrentConfig = normalizeConfig(await this.read());
     const normalizedTargetConfig = normalizeConfig(targetConfig);
 
+    // Update the config
```
nambrot, line 163: Why does moving this here make any difference? I feel like if we have logic that depends on the module state, then we should make that a lot more obvious

.changeset/eighty-planets-tan.md
mshojaei-txfusion: updated

typescript/sdk/src/hook/EvmHookModule.ts
mshojaei-txfusion: As now we use `this.args.config` on read as well, we can't change it before read.
What do you exactly mean by "we should make that a lot more obvious"?

typescript/sdk/src/hook/EvmHookModule.ts
mshojaei-txfusion: @ltyu can you please first clarify this and then I write the comments?

ltyu: The reason read() has this branching logic is because the Module was designed to allow
1. initialization through the constructor, which is used when we have an existing deployed Hook config.
2. initialization through `create()`, in this case, it is used to create a new Hook.

In both cases, we want to allow the caller to `read()`. The reason I suggested using the `address` is due to the first point. If a user initializes a config through the constructor, we can assume they have an existing deployed config i.e., a DerivedHookConfig. For the latter point, `deployedHook` address will be populated so we will derive from it. 

alternatively, we can also populate this.args.config in the [create()]((https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/cc3af7d0a7599d6b84a579624b51f77c95d9e253/typescript/sdk/src/hook/EvmHookModule.ts#L264) ) with a DerivedConfig to simplify `read()` 

I'm not sure what the `if (typeof config === 'string')` is used for. perhaps @paulbalaji can clarify

typescript/sdk/src/hook/EvmHookModule.ts
nambrot: We should have clear semantics as to when instance variables (like `this.args.config` are populated and readable). In this case, it sounds like this.args.config is expected to be not usable until ` read` happens, and that is not obvious at all

typescript/sdk/src/hook/EvmHookModule.ts
nambrot: Filing https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5610 as somewhat related

PR #5551, opened 2025-02-24
feat(cli/sdk): extra lockbox deriving and checking

xeno097:
### Description

This PR introduces the necessary changes to allow the CLI to read and check extra lockboxes linked to an xERC20 token

### Drive-by changes

- Replaces logic to get the deployment tx for a contract in the verify utils with the newly implemented `tryGetContractDeploymentTransaction `
- Adds the `XERC20VSTest` in the solidity contracts
- Adds testing utils for deploying the new `XERC20VSTest` and lockboxes contracts
- Changes to testing registry block explorer metadata from `etherscan` to `other` to force the CLI to use the RPC for fetching the associated extra lockboxes.
- Fixes a bug in warp check tests caused by not resetting the warp config to the initial value after modifying it in the ism and hook tests 

### Related issues

-

### Backward compatibility

- YES

### Testing

- Manual
- e2e


typescript/sdk/src/token/xerc20.ts
```diff
@@ -0,0 +1,197 @@
...
+    const contractDeploymentTx = await getContractDeploymentTransaction(
+      { apiUrl, apiKey: apiKey },
+      { contractAddress: xERC20Address },
+    );
+
+    const deploymentTransactionReceipt = await provider.getTransactionReceipt(
+      contractDeploymentTx.txHash,
+    );
+
+    startBlock = deploymentTransactionReceipt.blockNumber;
```
nambrot, line 81: When using etherscan, imo we don't need to add this logic, since we already filter on the event topic no? Simplifies the logic

typescript/sdk/src/token/types.ts
```diff
@@ -58,22 +58,25 @@ const xERC20LimitConfigSchema = z.object({
 });
 export type XERC20LimitConfig = z.infer<typeof xERC20LimitConfigSchema>;
 
+const xERC20ExtraLockboxLimitConfigsSchema = z.object({
```
nambrot, line 61: It just occurs to me that there is really no reason why we should call this extra lockboxes, from the perspective of the xerc20 token, they are just other bridges, wdyt of us renaming this?

typescript/sdk/src/token/xerc20.ts
xeno097: Unfortunately, it seems that we can't remove it because depending on the explorer API, it will start complaining that the block range params are required
image

typescript/sdk/src/token/xerc20.ts
```diff
@@ -0,0 +1,221 @@
...
+    (log) =>
+      ({
+        address: log.address,
+        data: log.data,
+        blockNumber: BigInt(log.blockNumber),
+        transactionHash: log.transactionHash,
+        logIndex: Number(log.logIndex),
+        transactionIndex: Number(log.transactionIndex),
+        topics: log.topics,
+      } as Log),
```
yorhodes, line 107: why is this cast necessary?

typescript/sdk/src/token/xerc20.ts
```diff
@@ -0,0 +1,221 @@
...
+    logs,
+  });
+
+  const dedupedBridges = parsedLogs.reduce((acc, log) => {
+    if (log.blockNumber > (acc[log.args.bridge]?.blockNumber ?? 0n)) {
+      acc[log.args.bridge] = log;
+    }
+
+    return acc;
+  }, {} as Record<string, ConfigurationChangedLog>);
```
yorhodes, line 186: I find this block quite hard to read

yorhodes (approved):
lgtm
can you open an updated config PR on the registry too?

typescript/sdk/src/token/xerc20.ts
```diff
@@ -0,0 +1,221 @@
...
+          `Contract at address ${log.args.bridge} on chain ${chain} is not a XERC20Lockbox contract.`,
+        );
+        return undefined;
+      }
+    });
+
+  const lockboxes = await Promise.all(lockboxPromises);
+  return lockboxes
+    .filter((log) => log !== undefined)
+    .map((log) => log as ConfigurationChangedLog)
```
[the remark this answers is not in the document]
xeno097: Unfortunately the type cast is necessary otherwise ts will complain when building the project because it seems that it doesn't understand that there are no undefined values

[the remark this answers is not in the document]
yorhodes: in future you can make the return type of the filter a type predicate
https://www.geeksforgeeks.org/what-is-type-predicates-in-typescript/

typescript/sdk/src/token/xerc20.ts
```diff
@@ -149,6 +149,7 @@ async function getConfigurationChangedLogsFromRpc({
   const currentBlockNumber = await provider.getBlockNumber();
   return provider.getLogs({
     address: xERC20Address,
+    fromBlock: 0,
```
nambrot, line 152: I'm actually not sure you can do this, because some or maybe even most rpc providers don't let you query for more than 2000 blocks

xeno097: You are right, but filtering by topic without setting the block boundaries should be enough in this case, I'll test with some public rpc to see if it works well

xeno097: update: seems to work fine just by setting the upper block number and filtering by topic

PR #5554, opened 2025-02-24
fix: Prevent try on some transaction methods and errors

### LINKED ISSUE #3119: SmartProvider shouldn't retry send_transaction calls

ltyu:
### Description
This PR stops retries in the SmartProvider by parsing the transaction method and error response. It prevents them when:
1. There are multiple providers
2. Within the `retryAsync`

### Related issues

- Fixes #3119

### DIFF
--- a/typescript/sdk/src/providers/SmartProvider/SmartProvider.foundry-test.ts
+++ b/typescript/sdk/src/providers/SmartProvider/SmartProvider.foundry-test.ts
@@ -2,6 +2,7 @@ import { expect } from 'chai';
 import { errors as EthersError, Wallet, constants } from 'ethers';
 
 import { ERC20__factory } from '@hyperlane-xyz/core';
+import { randomInt } from '@hyperlane-xyz/utils';
 
 import { randomAddress } from '../../test/testUtils.js';
 
@@ -17,6 +18,7 @@ const NETWORK = 31337;
 const URL = 'http://127.0.0.1:8545';
 
 describe('SmartProvider', async () => {
+  const maxRetries = randomInt(50, 0);
   let signer: Wallet;
   let smartProvider: HyperlaneSmartProvider;
   let contractAddress: string;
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -55,6 +55,36 @@ const DEFAULT_STAGGER_DELAY_MS = 1000; // 1 seconds
...
+  Known = 'known',
+  Nonce = 'nonce',
+  Underpriced = 'underpriced',
+  InsufficientFunds = 'insufficient funds',
+  InsufficientBalance = 'insufficient balance',
+  Revert = 'revert',
+}
+export function hasNonRetryableError(params: { method: string; error: Error }) {
+  const { method, error } = params;
+  const TX_METHODS = ['sendRawTransaction', 'sendTransaction'];
```
tkporter, line 72: I would've thought these would need to include the `eth_` prefix, e.g. `eth_sendRawTransaction`, `eth_call`, etc

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -310,10 +341,21 @@ export class HyperlaneSmartProvider
...
           providerResultErrors.push(result.error);
-          pIndex += 1;
+
+          const nonRetryable = hasNonRetryableError({
+            method,
+            error: result.error as Error,
+          });
+
+          pIndex += nonRetryable
+            ? providers.length // Setting this to providers.length will force exit the loop and not retry with the next provider
```
tkporter, line 353: feels a little bit like a hack, wonder if there's a nicer way to do it?

worried that it's so dependent on the `if (pIndex < providers.length) {` and if that were to change, it's not obvious to change it here too

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -310,10 +341,21 @@ export class HyperlaneSmartProvider
...
+            : 1;
+
+          this.logger.debug(
+            isLastProvider || nonRetryable ? '' : 'Triggering next provider.',
```
tkporter, line 357: Should we log something about it being the last time it's attempted and why? feels useful to know if it's non retryable

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
ltyu: good point. I've changed this to instead throw

PR #5558, opened 2025-02-25
fix: More exhaustive search for a relevant transaction in Solana block for Scraper

ameten:
### Description

Currently, Scraper cannot make progress (either forward or backward) on Solana chain if it encounters a transaction which contains two separate entries of Mailbox program acting on two different Hyperlane messages. If Scraper encounters such a transaction on the forward indexing, it stalls and new dispatched messages are not being indexed anymore. If Scraper encounters such a transaction on the backward indexing, Scraper will not index historical dispatched messages.

The root cause of the issue is that Scraper encounters the first Mailbox program entry in the transaction, but since this entry does not act an a given Hyperlane message, the whole transaction is considered to be irrelevant to the given Hyperlane message.

This change fixes the issue: Scraper will continue iterating over the list of instructions (programs) included into a transaction until it exhaust all of them, so that it does not miss the entry of Mailbox program which acts on the Hyperlane message in question.

### Backward compatibility

Yes

### Testing

Unit test

### DIFF
--- a/rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
+++ b/rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
@@ -56,22 +56,31 @@ pub fn test_search_dispatched_message_versioned_transaction() {
 pub fn test_search_dispatched_message_relevant_instruction_out_of_two_in_single_transaction() {
     // given
     let mailbox_program_id = decode_pubkey("E588QtVUvresuXq2KoNEwAmoifCzYGpRBdHByN9KQMbi").unwrap();
-    let dispatched_message_pda_account =
+    let dispatched_message_pda_account_one =
         decode_pubkey("HkS7U5adrqR4PZfn6DUEtHzwHxF8hZzmrEkJE8UuqFmz").unwrap();
+    let dispatched_message_pda_account_two =
+        decode_pubkey("9qg84RiHnQmi8Qk7ZYWRs7VTtkRB42sB4UFdhLNZ15qD").unwrap();
     let transactions = transactions(&read_json(
         "dispatch_message_two_instructions_in_one_txn.json",
     ));
 
     // when
-    let transaction_hashes = search_transactions(
-        transactions,
+    let transaction_hashes_one = search_transactions(
+        transactions.clone(),
         &mailbox_program_id,
-        &dispatched_message_pda_account,
+        &dispatched_message_pda_account_one,
+        is_message_dispatch_instruction,
+    );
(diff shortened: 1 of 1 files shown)


rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
```diff
@@ -52,6 +52,28 @@ pub fn test_search_dispatched_message_versioned_transaction() {
...
+    // when
+    let transaction_hashes = search_transactions(
+        transactions,
+        &mailbox_program_id,
+        &dispatched_message_pda_account,
+        is_message_dispatch_instruction,
+    );
+
+    // then
+    assert!(!transaction_hashes.is_empty());
```
daniel-savu, line 74: should we assert that there are exactly two matching? otherwise maybe we're returning the wrong number of txs

daniel-savu (approved):

rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
ameten: Fixed

rust/main/chains/hyperlane-sealevel/src/log_meta_composer/tests.rs
```diff
@@ -49,7 +49,29 @@ pub fn test_search_dispatched_message_versioned_transaction() {
...
+    // when
+    let transaction_hashes = search_transactions(
+        transactions,
+        &mailbox_program_id,
+        &dispatched_message_pda_account,
+        is_message_dispatch_instruction,
+    );
+
+    // then
+    assert_eq!(transaction_hashes.len(), 1);
```
daniel-savu, line 74: hm, maybe i'm misunderstanding here - is this always expected to be 1 because we specify the `dispatched_message_pda_account`? and the purpose of this unit test is to make sure we identify the second message? in that case I assume we want to make sure we find a tx hash for both pdas in the tx json

ameten: There is a single transaction in the file used in this unit test. So, the possible outcomes are either 0 or 1 in terms of length of the vector we get back from the method.

ameten: I see what you mean. The search should work for both messages. Let me check what should be done there.

ameten: Added

PR #5560, opened 2025-02-25
feat: chain metadata availability

Xaroz:
### Description

- Add `availability` field to `ChainMetadata`. It includes the `status` and a `reasons` array when status is `disabled`
- Add test for the new field

### Drive-by changes

No

### Backward compatibility

Yes
### Testing

Unit tests


typescript/sdk/src/metadata/chainMetadataTypes.ts
```diff
@@ -34,6 +34,19 @@ export enum ChainTechnicalStack {
...
+  Disabled = 'disabled',
+}
+
+export enum ChainDisabledReason {
+  BadRpc = 'badrpc',
+  Deprecated = 'deprecated',
+  Private = 'private',
+  Unavailable = 'unavailable',
+  Other = 'other',
+}
```
xeno097, line 48: I'd add a small doc comment on top of each variant to explain in which scenarios we might want to use them, at least for the non obvious one like `unavailable` and `private`

Xaroz: nice suggestion!

xeno097 (approved):

PR #5567, opened 2025-02-26
feat: add ism limit

kamiyaa:
### Description

Add an total ISM limit. This works in addition to the ISM max depth.
If either is reached, we won't build the metadata and drop the message.

- combined `Metadata` with `MetadataBuildError`
  - This fixes an issue where the incorrect error is returned when using AggregationIsm + max ism count error occurs
    - instead of returning MaxIsmCount error, it returns CouldNotFetch
  - This also simplifies error handling and makes it easier for us to respond to errors (see [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/jeff/ism-limit-2/rust/main/agents/relayer/src/msg/pending_message.rs#L307))

```rust
// old
pub enum MetadataBuildError {
    UnsupportedModuleType(ModuleType),
    MaxDepthExceeded(u32),
}

pub enum Metadata {
    Found(Vec<u8>),
    CouldNotFetch,
    Refused(String),
}

// new
#[derive(Clone, Debug, PartialEq, thiserror::Error)]
pub enum MetadataBuildError {
    #[error("Some external error causing the build to fail")]
    FailedToBuild,
    /// While building metadata, encountered something that should
    /// prohibit all metadata for the message from being built.
    /// Provides the reason for the refusal.
    #[error("Refused")]
    Refused(String),
    /// Unable to fetch metadata, but no error occurred
    #[error("Could not fetch metadata")]
    CouldNotFetch,
    #[error("Unknown or invalid module type ({0})")]
    UnsupportedModuleType(ModuleType),
    #[error("Exceeded max depth when building metadata ({0})")]
    MaxIsmDepthExceeded(u32),
    #[error("Exceeded max count when building metadata ({0})")]
    MaxIsmCountReached(u32),
    #[error("Aggregation threshold not met ({0})")]
    AggregationThresholdNotMet(u32),
}

#[derive(Clone, Debug, new)]
pub struct Metadata(Vec<u8>);
```

 - moved previous recursive logic into function: `build_message_metadata`
   - I wanted to move more of the logic out of the builders and just have functions, but it was harder than I thought. So I just have the recursive function building the Builders.

 - created a new trait `BaseMetadataBuilderTrait` for `BaseMetadataBuilder`
   - This way we can create a Mock version for testing
 - Add `MockBaseMetadataBuilder`, `MockInterchainSecurityModule`, `MockRoutingIsm`
   - essentially just a queue of responses we can populate to get the response we want, so we can test logic easier
   - Not sure if mockall has an easy way to do this, but I didn't really look into it

### Drive-by changes

 - moved some code into separate files. I find it easier to read multiple files than reading a single long file :eyes:

### Related issues

https://github.com/hyperlane-xyz/issues/issues/1424

### Backward compatibility

 - the new error handling might cause different behaviour for errors
   - In the case of UnsupportedModuleType, MaxIsmDepth, MaxIsmCount, we want to reprepare instead of just dropping these messages

### Testing
Unit tests
 - depth already reached
 - ism count already reached
 - max depth exceeded
 - max ism count exceeded
 - max ism count exceeded with aggregation ism


rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -2,61 +2,50 @@
...
+    /// Provides the reason for the refusal.
+    #[error("Refused")]
+    Refused(String),
+    /// Unable to fetch metadata, but no error occurred
+    #[error("Could not fetch metadata")]
+    CouldNotFetch,
     #[error("Unknown or invalid module type ({0})")]
     UnsupportedModuleType(ModuleType),
     #[error("Exceeded max depth when building metadata ({0})")]
     MaxDepthExceeded(u32),
```
ameten, line 30: Since we call new error `MaxIsmCountReached` (with `Ism` infix), I would add infix `Ism` into `MaxDepthExceeded` error as well.

rust/main/agents/relayer/src/msg/metadata/metadata_builder.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub max_ism_count: u32,
+}
+
+#[async_trait::async_trait]
+pub trait MetadataBuilder: Send + Sync {
+    /// Given a message, build it's ISM metadata
+    async fn build(
+        &self,
+        message: &HyperlaneMessage,
+        mut params: MessageMetadataBuildParams,
```
ameten, line 39: I wonder if we really need `mut` here.

```suggestion
       params: MessageMetadataBuildParams,
```

rust/main/agents/relayer/src/msg/metadata/base.rs
kamiyaa: Ah good point!

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -94,11 +95,16 @@ pub trait MultisigIsmMetadataBuilder: AsRef<MessageMetadataBuilder> + Send + Syn
 
 #[async_trait]
 impl<T: MultisigIsmMetadataBuilder> MetadataBuilder for T {
-    async fn build(&self, ism_address: H256, message: &HyperlaneMessage) -> Result<Metadata> {
+    async fn build(
+        &self,
+        message: &HyperlaneMessage,
+        params: MessageMetadataBuildParams,
+    ) -> eyre::Result<Metadata> {
```
ameten, line 102: Unnecessary prefix `eyre::`

```suggestion
    ) -> Result<Metadata> {
```

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -286,13 +292,32 @@ impl PendingOperation for PendingMessage {
...
+                    }
+                    // These errors cannot be recovered from, so we drop them
+                    MetadataBuildError::UnsupportedModuleType(reason) => {
+                        warn!(?reason, "Unsupported module type");
+                        return PendingOperationResult::Drop;
+                    }
+                    MetadataBuildError::MaxDepthExceeded(depth) => {
+                        warn!(depth, "Max ISM depth reached");
+                        return PendingOperationResult::Drop;
+                    }
```
ameten, line 315: Previously, we called `on_reprepare` to deal with these two errors, but now we drop the operation straight away. Is this new behaviour expected/intentional?

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+/// Builds metadata for a message.
+pub async fn build_message_metadata(
+    message_builder: &MessageMetadataBuilder,
+    message: &HyperlaneMessage,
+    params: MessageMetadataBuildParams,
+) -> eyre::Result<IsmWithMetadataAndType> {
+    let res = build_message_metadata_recursive(message_builder.clone(), message, params).await?;
+    Ok(res)
```
ameten, line 82: I wonder if this will work here:

```suggestion
    build_message_metadata_recursive(message_builder.clone(), message, params).await
```

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+async fn build_message_metadata_recursive(
```
ameten, line 85: I maybe missing something, but I cannot find recursion in this function. I wonder what I am missing?

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+#[derive(Debug, Clone)]
+pub struct MessageMetadataBuilder {
+    pub base: Arc<Box<dyn BaseMetadataBuilderTrait>>,
+    pub app_context: Option<String>,
+    /// current ISM depth.
+    /// ISMs can be structured recursively. We keep track of the depth
+    /// of the recursion to avoid infinite loops.
+    pub depth: u32,
+    /// current ISM count
+    pub ism_count: Arc<Mutex<u32>>,
```
ameten, line 35: We update `depth` and `ism_count` in the same place. I wonder why we need `Arc<Mutex<u32>>` for `ism_count` and we don't need it for `depth`?

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -271,8 +275,10 @@ impl PendingOperation for PendingMessage {
             }
         };
 
+        let metadata_build_params =
+            MessageMetadataBuildParams::new(ism_address, ISM_MAX_DEPTH, ISM_MAX_COUNT);
```
ameten, line 279: It looks like this is the only place where we construct params. In other places, we copy the options. I wonder why we cannot pass `ISM_MAX_DEPTH` and `ISM_MAX_COUNT` directly into the place we use them instead of passing through all the hierarchy of builders?

rust/main/agents/relayer/src/msg/pending_message.rs
kamiyaa: Yeah thats a good point. What do you think @daniel-savu @tkporter ?
Intuitively, it made sense to me to drop these if they are unprocessable with the ISM limits we have.

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
kamiyaa: Ah good point. I can remove this function.
The recursion logic is kind of confusing, but basically:
```
1. MessageMetadataBuilder tries to build the metadata
2. Another *Builder struct wraps around MessageMetadataBuilder
3. *Builder updates some build options, ie ism address
4. Calls MessageMetadataBuilder.build() with the new parameters
5. repeat to step 1
```
I wanted to move the logic out of these Builder wrappers but for now its too much code changes :sweat_smile:

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
kamiyaa: For AggregationISM, we spawn tokio threads to run the ISM metadata building in parallel.
These threads don't really need to share the depth because the depths are independent. But they need to share the count because one ISM might take up a lot of ISMs.

rust/main/agents/relayer/src/msg/pending_message.rs
kamiyaa: I think this is a bit more future proof. Where maybe in the future, depth and count is chain-specific.

daniel-savu: I think it's safe to assume they'll always be global, since the point is to prevent DoS-ing the relayer which could be attempted on any chain / ISM

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,420 @@
...
+            .build(message, params)
+            .await
+            .context("When building metadata")?,
+        ModuleType::CcipRead => CcipReadIsmMetadataBuilder::new(message_builder)
+            .build(message, params)
+            .await
+            .context("When building metadata")?,
+        ModuleType::Null => Metadata::Found(vec![]),
+        _ => return Err(MetadataBuildError::UnsupportedModuleType(module_type).into()),
+    };
```
ameten, line 143: It looks like the original implementation manages to extract method `build` call so that we don't need to repeat it every time. I wonder what may have prevented us to do the same here?

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+                metadata: Metadata::Failed(MetadataBuildError::MaxIsmCountReached(
+                    options.max_ism_count,
+                )),
+                module_type,
+            });
+        }
+
+        // update depth and ism count
+        message_builder.depth = message_builder.depth.saturating_add(1);
+        *ism_count = ism_count.saturating_add(1);
```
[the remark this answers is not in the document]
kamiyaa: Yeah, that's why I had to add the `Arc<Mutex<>>` for the ism_count

rust/main/agents/relayer/src/msg/pending_message.rs
tkporter: imo we should not drop these and just reprepare, a recipient can always change the ISM configuration and we should allow for that to happen

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
```diff
@@ -0,0 +1,147 @@
...
+    accumulator::merkle::Proof, AggregationIsm, CcipReadIsm, Checkpoint, HyperlaneDomain,
+    HyperlaneMessage, InterchainSecurityModule, MultisigIsm, RoutingIsm, H256,
+};
+
+use crate::msg::metadata::{BaseMetadataBuilderTrait, IsmAwareAppContextClassifier};
+
+type ResponseList<T> = Arc<Mutex<VecDeque<T>>>;
+
+#[derive(Debug, Default)]
+pub struct MockBaseMetadataBuilderResponses {
```
kamiyaa, line 17: Kind of verbose mocking, but I think its relatively straightforward to understand and use.
Not sure if mockall has something similar.

rust/main/agents/relayer/src/msg/metadata/ccip_read.rs
```diff
@@ -5,7 +5,7 @@ use derive_more::Deref;
 use derive_new::new;
 use ethers::{abi::AbiDecode, core::utils::hex::decode as hex_decode};
 use eyre::Context;
-use hyperlane_core::{utils::bytes_to_hex, HyperlaneMessage, RawHyperlaneMessage};
+use hyperlane_core::{utils::bytes_to_hex, HyperlaneMessage, RawHyperlaneMessage, H256};
 use hyperlane_ethereum::OffchainLookup;
```
ameten, line 9: minor: let's move these imports just after the external imports with a new line separation.

rust/main/agents/relayer/src/msg/metadata/null_metadata.rs
```diff
@@ -0,0 +1,22 @@
+use super::{Metadata, MetadataBuilder};
+use async_trait::async_trait;
+use derive_new::new;
+use tracing::instrument;
+
+use hyperlane_core::{HyperlaneMessage, H256};
```
ameten, line 6:
```suggestion
use async_trait::async_trait;
use derive_new::new;
use tracing::instrument;

use hyperlane_core::{HyperlaneMessage, H256};

use super::{Metadata, MetadataBuilder};
```

rust/main/agents/relayer/src/msg/metadata/routing.rs
```diff
@@ -2,12 +2,10 @@ use async_trait::async_trait;
...
 use derive_new::new;
 use eyre::Context;
-use hyperlane_core::HyperlaneMessage;
+use hyperlane_core::{HyperlaneMessage, H256};
 use tracing::instrument;
 
-use super::{
-    metadata_builder::MessageMetadataBuildParams, MessageMetadataBuilder, Metadata, MetadataBuilder,
-};
+use super::{MessageMetadataBuilder, Metadata, MetadataBuilder};
```
ameten, line 8:
```suggestion
use hyperlane_core::{HyperlaneMessage, H256};

use super::{MessageMetadataBuilder, Metadata, MetadataBuilder};
```

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
```diff
@@ -145,3 +145,63 @@ impl BaseMetadataBuilderTrait for MockBaseMetadataBuilder {
...
+#[cfg(test)]
+mod tests {
+    use hyperlane_core::ModuleType;
+
+    use crate::test_utils::mock_ism::MockInterchainSecurityModule;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_mock_001() {
```
ameten, line 158: Could you please give more descriptive name to the method?

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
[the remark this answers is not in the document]
ameten: Discussed in person. All good.

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
ameten: Discussed in person. We'll use Box as in original implementation.

ameten (approved):

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+
+    #[tokio::test]
+    async fn max_ism_count_reached() {
+        let base_builder = build_mock_base_builder();
+
+        for _ in 0..11 {
+            insert_mock_routing_ism(&base_builder);
+        }
```
daniel-savu, line 334: I think using the routing ISM here can make it less easy to understand because depth will always be equal to the ism count. An aggregation ISM that points to a bunch of other ISMs would probably be more common and easier to understand. In the unit test I'd add a check that depth is 1 and count is the number of sub-isms + 1, so the reader has an easier time figuring out what the test does

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,430 @@
...
+                panic!("Metadata found when it should have failed");
+            }
+            Metadata::Failed(err) => {
+                assert_eq!(err, MetadataBuildError::MaxIsmCountReached(ISM_MAX_COUNT));
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn max_depth_reached() {
```
daniel-savu, line 298: i'd add a check for both the max_depth and max_count in all unit tests so the reader can more easily understand the ISM layout. Or maybe the depth and count can be passed as parameters to the initialization function

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,430 @@
...
+        // push aggregation ISMs
+        //
+        // R = Routing (9)
+        // Agg = Aggregation
+        //
+        //           Agg
+        //     /      |      \
+        //    R       R       R
+        //    |       |       |
+        //    R       R       R
```
daniel-savu, line 373: for instance in this case I'd want us to check that the tracked height never exceeds 2 but that we do hit a depth of 2 three times

daniel-savu (changes requested):
metadata building is pretty complex, nice work! Left some comments

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,430 @@
...
+    multisig::{MerkleRootMultisigMetadataBuilder, MessageIdMultisigMetadataBuilder},
+    null_metadata::NullMetadataBuilder,
+    routing::RoutingIsmMetadataBuilder,
+    Metadata, MetadataBuilder,
+};
+
+/// Builds metadata for a message.
+#[derive(Debug, Clone)]
+pub struct MessageMetadataBuilder {
+    pub base: Arc<Box<dyn BaseMetadataBuilderTrait>>,
```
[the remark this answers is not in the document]
kamiyaa: Ah, I didn't realized we didn't need a box here. Thanks!

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -40,6 +42,8 @@ pub const CONFIRM_DELAY: Duration = if cfg!(any(test, feature = "test-utils")) {
 };
 
 pub const RETRIEVED_MESSAGE_LOG: &str = "Message status retrieved from db";
+pub const ISM_MAX_DEPTH: u32 = 13;
+pub const ISM_MAX_COUNT: u32 = 1000;
```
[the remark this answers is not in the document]
kamiyaa: ah okay :+1:

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -2,61 +2,56 @@
...
+}
+
+impl Metadata {
+    pub fn ok(&self) -> bool {
+        match self {
+            Self::Found(_) => true,
+            Self::Failed(_) => false,
+        }
+    }
+}
```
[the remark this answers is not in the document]
kamiyaa: Yeah, I was thinking of doing that originally too. Will look into it

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -121,27 +123,24 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
-                    _ => None,
-                }
-            })
-        }) {
-            return Ok(first_refusal);
+        for sub_module in sub_modules_and_metas.iter().flatten() {
+            if !sub_module.metadata.ok() {
+                return Ok(sub_module.metadata.clone());
+            }
         }
```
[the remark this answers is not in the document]
kamiyaa: ah good point

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -206,225 +191,277 @@ mod test {
...
-                panic!("Metadata found when it should have failed");
-            }
-            Metadata::Failed(err) => {
-                assert_eq!(err, MetadataBuildError::MaxIsmCountReached(ISM_MAX_COUNT));
-            }
-        }
+            .err()
+            .expect("Metadata found when it should have failed");
+        assert_eq!(err, MetadataBuildError::AggregationThresholdNotMet(2));
+        assert!(*(params.ism_count.lock().await) <= 4);
```
daniel-savu, line 439: if you add an error log where the depth is checked, you can the make sure the depth error was hit by adding a check like this below (the fn is auto-included if you decorate with `#[tracing_test::traced_test]`)
```rust
        logs_contain("depth limit reached (5)");
```

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -206,225 +191,277 @@ mod test {
...
-            Metadata::Found(_) => {
-                panic!("Metadata found when it should have failed");
-            }
-            Metadata::Failed(err) => {
-                assert_eq!(err, MetadataBuildError::MaxIsmCountReached(ISM_MAX_COUNT));
-            }
-        }
+            .err()
+            .expect("Metadata found when it should have failed");
+        assert_eq!(err, MetadataBuildError::AggregationThresholdNotMet(2));
```
daniel-savu, line 464: same here, would be good to add a `logs_include` check

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
```diff
@@ -168,13 +186,10 @@ mod tests {
...
             base_builder
                 .responses
-                .build_ism
-                .lock()
-                .unwrap()
-                .push_back(Ok(Box::new(mock_ism)));
+                .push_build_ism_response(H256::zero(), Ok(Box::new(mock_ism)));
         }
         {
-            let mock_ism = MockInterchainSecurityModule::default();
```
daniel-savu: are the new and old versions not equivalent?

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
```diff
@@ -198,7 +210,7 @@ mod tests {
         assert_eq!(module_type, ModuleType::Routing);
 
         let ism = base_builder
-            .build_ism(H256::zero())
+            .build_ism(H256::from_low_u64_be(10))
```
daniel-savu, line 213: wondering why the change

rust/main/agents/relayer/src/msg/metadata/ccip_read.rs
```diff
@@ -27,26 +32,40 @@ pub struct CcipReadIsmMetadataBuilder {
...
+        _params: MessageMetadataBuildParams,
+    ) -> Result<Metadata, MetadataBuildError> {
         const CTX: &str = "When fetching CcipRead metadata";
-        let ism = self.build_ccip_read_ism(ism_address).await.context(CTX)?;
+        let ism = self
+            .base_builder()
+            .build_ccip_read_ism(ism_address)
+            .await
+            .context(CTX)
+            .map_err(|_| MetadataBuildError::FailedToBuild)?;
```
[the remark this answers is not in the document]
kamiyaa: makes sense, I'll add it :+1:

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -0,0 +1,467 @@
...
+            .err()
+            .expect("Metadata found when it should have failed");
+        assert_eq!(err, MetadataBuildError::MaxIsmCountReached(0));
+        assert_eq!(*(params.ism_count.lock().await), 0);
+    }
+
+    #[tokio::test]
+    async fn max_depth_reached() {
+        let base_builder = build_mock_base_builder();
+        insert_ism_test_data_001(&base_builder);
```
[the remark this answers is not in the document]
kamiyaa: was just thinking in case we have more test data. But I will remove it :+1:

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
kamiyaa: So originally I had these mock structs just have a basic mock queue response setup.
FIFO in terms of the mocks we feed it.
But if we want to test multiple aggregation ISM, they all run in parallel, so its not possible to guarantee the responses they get.

So my solution is to have a hashmap of responses instead, `Hashmap<H256, Ism>`.
This way I can structure the AggregationISM in a way that yields reproducible results in the tests.

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
kamiyaa: ah cool, I didn't know tracing test provided these :100:

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -134,13 +133,11 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
             .base_builder()
             .build_aggregation_ism(ism_address)
             .await
-            .context(CTX)
-            .map_err(|_| MetadataBuildError::FailedToBuild)?;
+            .map_err(|_| MetadataBuildError::FailedToBuild(CTX.into()))?;
```
daniel-savu, line 136: what I had in mind is something like this, so we can see the actual error messages. Everywhere else as well
```suggestion
            .map_err(|err| MetadataBuildError::FailedToBuild(err.to_string()))?;
```

rust/main/agents/relayer/src/msg/metadata/ccip_read.rs
```diff
@@ -56,13 +54,18 @@ impl MetadataBuilder for CcipReadIsmMetadataBuilder {
                 return Err(MetadataBuildError::CouldNotFetch);
             }
             Err(raw_error) => {
-                let matching_regex = Regex::new(r"0x[[:xdigit:]]+")
-                    .map_err(|_| MetadataBuildError::FailedToBuild)?;
+                let matching_regex = Regex::new(r"0x[[:xdigit:]]+").map_err(|_| {
+                    MetadataBuildError::FailedToBuild("Invalid matching regex".into())
+                })?;
```
daniel-savu, line 59: here and elsewhere as well
```suggestion
                let matching_regex = Regex::new(r"0x[[:xdigit:]]+")
                    .map_err(|err| MetadataBuildError::FailedToBuild(err.to_string()))?;
```

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -96,17 +96,22 @@ pub async fn build_message_metadata(
...
+        .map_err(|_| MetadataBuildError::FailedToBuild("When fetching module type".into()))?;
 
     // check if max depth is reached
     if params.ism_depth >= message_builder.max_ism_depth {
+        tracing::error!(
+            ism_depth = message_builder.max_ism_depth,
+            ism_address = ?ism_address,
+            message_id = ?message.id(),
+            "Max ISM depth reached ({})",
+            message_builder.max_ism_depth,
```
daniel-savu, line 113: ah my bad for misleading here, ideally we shouldn't include variables in the log message for easier filtering in GCP. So the message should be `"Max ISM depth reached"` and maybe we could still read the `ism_depth` property in the unit test with a `tracing` util. If not, it's ok to just check for the log message without the depth

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
```diff
@@ -116,6 +121,13 @@ pub async fn build_message_metadata(
         // check if max ism count is reached
         let mut ism_count = params.ism_count.lock().await;
         if *ism_count >= message_builder.max_ism_count {
+            tracing::error!(
+                ism_count = message_builder.max_ism_count,
+                ism_address = ?ism_address,
+                message_id = ?message.id(),
+                "Max ISM count reached ({})",
```
daniel-savu, line 128: same here

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
```diff
@@ -21,6 +21,13 @@ pub struct MockBaseMetadataBuilderResponses {
     pub get_proof: ResponseList<eyre::Result<Proof>>,
     pub highest_known_leaf_index: ResponseList<Option<u32>>,
     pub get_merkle_leaf_id_by_message_id: ResponseList<eyre::Result<Option<u32>>>,
+    /// build_ism uses a hashmap of VecDeque responses instead.
+    /// This is because AggregationISMs run in parallel, so having just
+    /// a single VecDeque shared between the different threads yields unpredictable
+    /// results. One thread might run for a long time and starve the other threads
+    /// of responses.
+    /// In order to fix this in tests, the mock ISMs are built with specific addresses
+    /// in place. And gets responses from this based on the address.
```
daniel-savu, line 30: nice that's useful to know. Wondering when you say "uses a hashmap of VecDeque responses instead", instead of what it is you mean? Assuming a concurrent hashmap instead of an arcmutex. But since the mutex is still locked across the entire hashmap and the tests run fine I think it's not starvation that's the issue? More likely tests overwriting each other's state - we should update the comment to reflect this if that's the case

rust/main/agents/relayer/src/msg/metadata/aggregation.rs
kamiyaa: ah I see. Let me change it xd

rust/main/agents/relayer/src/msg/metadata/message_builder.rs
kamiyaa: The logs look like
```
2025-03-18T13:37:48.329343Z ERROR depth_already_reached: relayer::msg::metadata::message_builder: Max ISM depth reached ism_depth=0 ism_address=0x0000000000000000000000000000000000000000000000000000000000000000 message_id=0xf87c5995df379ae771337e27f463102a4c2ca5fd0218d63ead2fa2ef1ffc111b
2025-03-18T13:37:48.329388Z ERROR ism_count_already_reached: relayer::msg::metadata::message_builder: Max ISM count reached ism_count=0 ism_address=0x0000000000000000000000000000000000000000000000000000000000000000 message_id=0xf87c5995df379ae771337e27f463102a4c2ca5fd0218d63ead2fa2ef1ffc111b
```
and `logs_contain()` only takes in a string to match.

But I ordered the `ism_count` and `ism_depth` to be the first variable.
So we are able to test for strings like `Max ISM count reached ism_count=0` :ok_hand: 

Let me remove the variable in the log message :+1:

rust/main/agents/relayer/src/test_utils/mock_base_builder.rs
kamiyaa: ah, what I meant by instead is because all the other responses still just uses an `Arc<Mutex<VecDeque>>`.
If we need more complicated behaviour in the future, we can always come back and change them to a hashmap or some other data structure that suits our needs.

daniel-savu: I see, can you update the comment to clarify pls. This point also still stands I believe, so should the starvation point should be replaced:
> But since the mutex is still locked across the entire hashmap and the tests run fine I think it's not starvation that's the issue? More likely tests overwriting each other's state (vecdeque) - we should update the comment to reflect this if that's the case

daniel-savu (approved):
nice work, one outstanding comment and we're good to merge 🚢

Issue #5533, opened 2025-02-20
Warp read/check commands should not require signers


nambrot:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/8fc28501ff0c2f0351b17707e2252102ed9d7c35/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts#L142

![Image](https://github.com/user-attachments/assets/565c9bdf-bbbf-46b2-b75b-ea3d1749ff98)

PR #5574, opened 2025-02-26
fix: warp read check signer

mshojaei-txfusion:
### Description

This PR makes `warp:read` and `warp:check`/`warp:verify` operations independent of signer requirements. The changes include:

1. Removing temporary command handling for warp read/check/verify commands that previously required signers unnecessarily
2. Modifying the `EvmERC20WarpRouteReader` to use a zero address instead of requiring a signer address for read operations
3. Removing `WARP_READ` and `WARP_CHECK` from the command types that require specific chain resolvers

These changes improve the user experience by making read-only operations truly read-only, without requiring private keys or signers.

### Drive-by changes

None

### Related issues

resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5533

### Backward compatibility

Yes - These changes are backward compatible. They only remove unnecessary signer requirements for read operations and don't change any core functionality.

### Testing

Manual testing of warp:read and warp:check/verify commands without signer configuration


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -137,11 +137,15 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
-        value: BigNumber.from(0),
-      });
+      await this.multiProvider.estimateGas(
+        this.chain,
+        {
+          to: warpRouteAddress,
+          value: BigNumber.from(0),
+        },
+        // Use zero address as the sender as signer is not provided for read commands
+        constants.AddressZero,
```
nambrot, line 147: Do we even have to provide an address at all? Have we tested that this works? Just wondering if some rpcs might actually spcial case the 0x0 address since they know it definitely cannot be the from addrss ever, so maybe we need the 0x1 address?

nambrot (approved):

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
mshojaei-txfusion: Read works without even specifying from (tested with specifying `from: undefined`) but the way functions implemented we need to pass it, and I am not really sure would this estimate gas work for all casing without from.

So prefer for now use `0x1` as a safe option. (based on what you said)

nambrot:
Looks like some tests are failing, but feel free to merge once passing

mshojaei-txfusion:
> Looks like some tests are failing, but feel free to merge once passing

Just was missing some new tests from main, updated and passed!

PR #5570, opened 2025-02-26
feat: add MINT/mint-solanamainnet

xeno097:
### Description

This PR adds the config getter for the new mint <-> solana route

related registry PR:
https://github.com/hyperlane-xyz/hyperlane-registry/pull/626

### Drive-by changes

- Updates the igp configuration on solana

### Backward compatibility

- YES

### Testing

- Manual
- UI

PR #5580, opened 2025-02-27
chore: fix broken sealevel e2e build

daniel-savu:
### Description

We have at least 2 PRs whose CIs are failing due to the sealevel e2e: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5570 and https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5517

TIL that `cargo install`ing from a repo ignores the lockfile. This means that any patch bumps will be automatically applied. This wouldn't be a problem if crates upheld the promise that patches don't break the MSRV, but they do :) and this breaks our sealevel e2e because it is on `1.76.0` (a rust version that doesn't conform to the MSRV after applying the patch updates)

So the fix is to pass a `--locked` flag to `cargo install` to force it to use the lockfile. ~~I've also added the `--force` flag to invalidate any cache so we make sure the lockfile versions are always used.~~

### Testing

E2E should pass now

### DIFF
--- a/rust/sealevel/client/src/warp_route.rs
+++ b/rust/sealevel/client/src/warp_route.rs
@@ -747,6 +747,8 @@ pub fn install_spl_token_cli() {
             "dan/create-token-for-mint",
             "--rev",
             "ae4c8ac46",
+            "--locked",
+            "--force",
         ])
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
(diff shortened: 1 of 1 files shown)


ameten (approved):

daniel-savu:
> Do you know if `--force` will slow down CI?

it might but what it also does is install yanked crate versions and apparently our lockfile points to some, so I removed it and CI passed

PR #5581, opened 2025-02-27
feat(widgets): allow chains to be disabled in ChainSearchMenu

Xaroz:
### Description

Updates `<ChainSearchMenu />` to be able to disabled chains when the status of the chain is `disabled` 

- Add `showDisabledChains` boolean to props. When false it will make the chain items unable to be selected. Defaults to `true`
- Add disabled to the chain metadata because its required by `SearchMenu` component
- Now disabled items will be shown at the bottom
- New storybook for this use case

### Backward compatibility

Yes
### Testing

Manual and visual testing with storybook

### DIFF
--- a/typescript/widgets/src/chains/ChainSearchMenu.tsx
+++ b/typescript/widgets/src/chains/ChainSearchMenu.tsx
@@ -65,6 +65,10 @@ export interface ChainSearchMenuProps {
   showAddChainButton?: boolean;
   // Field by which data will be sorted by default
   defaultSortField?: DefaultSortField;
+  /**
+   * Include disabled chains in the shown list. Defaults to `true`
+   */
+  showDisabledChains?: boolean;
 }
 
 export function ChainSearchMenu({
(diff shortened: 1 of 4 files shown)


typescript/widgets/src/chains/ChainSearchMenu.tsx
```diff
@@ -64,6 +65,10 @@ export interface ChainSearchMenuProps {
   showAddChainButton?: boolean;
   // Field by which data will be sorted by default
   defaultSortField?: DefaultSortField;
+  /**
+   * Include disabled chains in the shown list. Defaults to `true`
+   */
+  showDisabledChains?: boolean;
```
cmcewen, line 71: suggest `shouldDisableChains`

typescript/widgets/src/chains/ChainSearchMenu.tsx
```diff
@@ -257,6 +276,14 @@ function chainSearch({
       })
       // Sort options
       .sort((c1, c2) => {
+        if (showDisabledChains === false) {
```
cmcewen, line 279: with name change, switch to true

cmcewen (approved):

PR #5588, opened 2025-02-28
feat: support for intents

fernandomg:
### Description

This PR is the result of multiple iterations towards a non-invasive inclusion of intents into the SDK.


typescript/sdk/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@arbitrum/sdk": "^4.0.0",
     "@aws-sdk/client-s3": "^3.577.0",
+    "@bootnodedev/intents-framework-core": "^0.1.0",
```
fernandomg, line 8: we need this to interact with the Hyperlane7683 contracts

typescript/sdk/src/token/Token.ts
```diff
@@ -175,6 +196,7 @@ export class Token implements IToken {
   getHypAdapter(
     multiProvider: MultiProtocolProvider<{ mailbox?: Address }>,
     destination?: ChainName,
+    fillDeadline: number = Math.floor(Date.now() / 1000) + 60 * 60 * 24,
```
fernandomg, line 199: `fillDeadline` is an intent-specific value that can only be accessed if it's passed via parameters at runtime.

It has a default value to 1 day (in seconds) if it's not defined in the request.

typescript/sdk/src/token/TokenStandard.ts
```diff
@@ -60,6 +62,8 @@ export const TOKEN_STANDARD_TO_PROTOCOL: Record<TokenStandard, ProtocolType> = {
   EvmHypSyntheticRebase: ProtocolType.Ethereum,
   EvmHypXERC20: ProtocolType.Ethereum,
   EvmHypXERC20Lockbox: ProtocolType.Ethereum,
+  EvmIntent: ProtocolType.Ethereum,
+  EvmIntentNative: ProtocolType.Ethereum,
```
fernandomg, line 66: `EmvIntent` and `EvmIntentNative` are the standars that will identify the intents

typescript/sdk/src/token/deploy.hardhat-test.ts
```diff
@@ -67,7 +67,7 @@ describe('TokenDeployer', async () => {
     await deployer.deploy(config);
   });
 
-  for (const type of [TokenType.collateral, TokenType.synthetic]) {
+  for (const type of [TokenType.collateral, TokenType.synthetic] as const) {
```
fernandomg, line 70: these `as const` were a contribution to avoid TS complaining about types.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -99,10 +99,31 @@ export class WarpCore {
...
+              }
+              if (t.standard === TokenStandard.EvmIntentNative) {
+                return true;
+              }
+            }
+            return false;
+          } else {
+            return t.addressOrDenom === addressOrDenom;
+          }
+        });
```
fernandomg, line 126: properly finding the connections among intent-based tokens

typescript/sdk/src/token/Token.ts
```diff
@@ -292,6 +314,44 @@ export class Token implements IToken {
       const connection = this.getConnectionForChain(destination);
       assert(connection, `No connection found for chain ${destination}`);
       return this.getIbcAdapter(multiProvider, connection);
+    } else if (standard === TokenStandard.EvmIntent) {
+      assert(
+        this.collateralAddressOrDenom,
+        'collateralAddressOrDenom required for EvmIntent tokens',
+      );
```
fernandomg, line 321: `addressOrDenom` hosts the address for the Hyperlane7683 instance (router), and `collateralAddressOrDenom` will be the ERC20 token address. That's why it's required when dealing with intents.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -102,10 +102,31 @@ export class WarpCore {
...
```
Xaroz, line 129: reduce conditions nesting by doing something like:

`if (t.standard !== TokenStandard.EvmIntent && t.standard !== TokenStandard.EvmIntentNative) return t.address === addressOrDenom`

and so on.

fernandomg: fixed, check [`bb77b36` (#5588)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5588/commits/bb77b36364ba84f9a2a4c7e46abf47b95cb8d218)

I went a step further and did some refactor to prevent creating the variable `token1` when it wasn't necessary, and changed the `token2` name to make it more clear. But if it was done to improve readability, let me know and I'll revert those changes.

PR #5592, opened 2025-03-03
feat: Add names to tokio tasks in Relayer

ameten:
### Description

Add names to tokio tasks in Relayer so that each task is recognisable in tokio-console

### Drive-by changes

- Minor method rename `run_merkle_tree_hook_syncs` to `run_merkle_tree_hook_sync`.

### Related issues

### Backward compatibility

Yes

### Testing

Locally

### DIFF
--- a/rust/main/agents/relayer/src/merkle_tree/processor.rs
+++ b/rust/main/agents/relayer/src/merkle_tree/processor.rs
@@ -42,6 +42,12 @@ impl Debug for MerkleTreeProcessor {
 
 #[async_trait]
 impl ProcessorExt for MerkleTreeProcessor {
+    fn name(&self) -> String {
+        let mut name = "processor::merkle_tree::".to_string();
+        name.push_str(self.domain().name());
+        name
+    }
+
     /// The domain this processor is getting merkle tree hook insertions from.
     fn domain(&self) -> &HyperlaneDomain {
         self.db.domain()
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -324,13 +324,15 @@ impl BaseAgent for Relayer {
 
         let task_monitor = tokio_metrics::TaskMonitor::new();
         if let Some(tokio_console_server) = self.tokio_console_server.take() {
-            let console_server =
-                tokio::spawn(TaskMonitor::instrument(&task_monitor.clone(), async move {
+            let console_server = tokio::task::Builder::new()
+                .name("tokio_console_server")
+                .spawn(TaskMonitor::instrument(&task_monitor.clone(), async move {
                     info!("Starting tokio console server");
                     if let Err(e) = tokio_console_server.serve().await {
                         error!(error=?e, "Tokio console server failed to start");
                     }
-                }));
+                }))
+                .expect("spawning tokio task from Builder is infalliable");
             tasks.push(console_server.instrument(info_span!("Tokio console server")));
         }
         let sender = BroadcastSender::new(ENDPOINT_MESSAGES_QUEUE_SIZE);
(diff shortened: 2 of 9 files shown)


rust/main/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -42,6 +42,12 @@ impl Debug for MerkleTreeProcessor {
 
 #[async_trait]
 impl ProcessorExt for MerkleTreeProcessor {
+    fn name(&self) -> String {
+        let mut name = "processor::merkle_tree::".to_string();
```
tkporter, line 46: any reason to prefer this over a one-liner `format!`? here and elsewhere

rust/main/hyperlane-base/src/server/base_server.rs
```diff
@@ -41,13 +41,16 @@ impl Server {
...
+        tokio::task::Builder::new()
+            .name("agent::server")
+            .spawn(async move {
+                let addr = SocketAddr::from(([0, 0, 0, 0], port));
+                axum::Server::bind(&addr)
+                    .serve(app.into_make_service())
+                    .await
+                    .expect("Failed to start server");
+            })
+            .expect("spawning tokio task from Builder is infalliable")
```
tkporter, line 53: nit, also in a couple other places
```suggestion
            .expect("spawning tokio task from Builder is infallible")
```

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -626,16 +647,21 @@ impl Relayer {
...
+            .spawn(TaskMonitor::instrument(&task_monitor, async move {
+                // Propagate task panics
+                serial_submitter.spawn().await.unwrap_or_else(|err| {
+                    panic!(
+                        "destination submitter panicked for destination {}: {:?}",
+                        destination, err
+                    )
+                });
+            }))
+            .expect("spawning tokio task from Builder is infalliable")
```
tkporter, line 663: why does it return a Result then? I assume there's some condition in which an error could occur, but it's just unlikely?

tkporter (approved):

rust/main/agents/relayer/src/merkle_tree/processor.rs
kamiyaa: should this be `merkle_tree::processor`? Because I see its under `agents/relayer/src/merkle_tree/processor`

ameten: > any reason to prefer this over a one-liner format!? here and elsewhere

Good point! No actual reason apart from not thinking about `format!`.

ameten: > should this be `merkle_tree::processor`? Because I see its under `agents/relayer/src/merkle_tree/processor`

I've reverted it since `MerkleTreeProcessor` is a processor, so, processor is a more general term for this object.

rust/main/hyperlane-base/src/server/base_server.rs
ameten: Fixed

rust/main/agents/relayer/src/relayer.rs
ameten: It is a good question on the API design, but the `spawn` method just cannot fail ever in `tokio::task::Builder`:

```
    pub fn spawn<Fut>(self, future: Fut) -> io::Result<JoinHandle<Fut::Output>>
    where
        Fut: Future + Send + 'static,
        Fut::Output: Send + 'static,
    {
        let fut_size = mem::size_of::<Fut>();
        Ok(if fut_size > BOX_FUTURE_THRESHOLD {
            super::spawn::spawn_inner(Box::pin(future), SpawnMeta::new(self.name, fut_size))
        } else {
            super::spawn::spawn_inner(future, SpawnMeta::new(self.name, fut_size))
        })
    }
```

PR #5604, opened 2025-03-05
fix(cli): checker remoteRouters check fix

xeno097:
### Description

This fixes the `warp check` command because it did not correctly check the `remoteRouters` object when included in the config. The issue was caused by the `normalizeConfig` function, which removes occurrences of the `address` field from an object. This converted the remoteRouter config to an empty object, causing the check to succeed when it shouldn't have.

image

A new `HypTokenRouterConfigToCheck` type has been implemented, which renames the field to `routerAddress` for checking purposes

image

### Drive-by changes

- None

### Backward compatibility

- YES

### Testing

- Manual
- E2E

### DIFF
--- a/typescript/cli/src/check/warp.ts
+++ b/typescript/cli/src/check/warp.ts
@@ -1,10 +1,55 @@
 import { stringify as yamlStringify } from 'yaml';
 
-import { WarpRouteDeployConfig, normalizeConfig } from '@hyperlane-xyz/sdk';
-import { ObjectDiff, diffObjMerge } from '@hyperlane-xyz/utils';
+import {
+  HypTokenRouterConfig,
+  WarpRouteDeployConfig,
+  normalizeConfig,
+} from '@hyperlane-xyz/sdk';
+import {
+  Address,
+  ChainId,
+  ObjectDiff,
+  diffObjMerge,
+} from '@hyperlane-xyz/utils';
 
 import { log, logGreen } from '../logger.js';
 import { formatYamlViolationsOutput } from '../utils/output.js';
+
+type HypTokenRouterConfigToCheck = Omit<
+  HypTokenRouterConfig,
+  'remoteRouters'
+> & {
+  remoteRouters: Record<ChainId, { routerAddress: Address }>;
+};
+
+// Changes address fields occurrences that should be checked in the config to have a
+// different name as the normalizeConfig function removes all the address
+// fields from a given object
+function formatTokenConfigToCheck(
+  config: HypTokenRouterConfig,
+): HypTokenRouterConfigToCheck {
+  const formattedConfig: any = {};
+  for (const [key, value] of Object.entries(config)) {
+    if ((key as keyof HypTokenRouterConfig) !== 'remoteRouters') {
+      formattedConfig[key as keyof HypTokenRouterConfig] = value;
+    }
+  }
+
+  if (config.remoteRouters) {
+    formattedConfig.remoteRouters = Object.entries(config.remoteRouters).reduce(
+      (acc, [chain, config]) => {
+        acc[chain] = {
+          routerAddress: config.address,
+        };
+
+        return acc;
+      },
+      {} as HypTokenRouterConfigToCheck['remoteRouters'],
+    );
+  }
+
+  return formattedConfig;
+}
 
 export async function runWarpRouteCheck({
   warpRouteConfig,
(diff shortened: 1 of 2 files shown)


typescript/cli/src/check/warp.ts
```diff
@@ -1,19 +1,66 @@
...
 
 import { log, logGreen } from '../logger.js';
 import { formatYamlViolationsOutput } from '../utils/output.js';
 
+const formatter: FormatObjectFormatter = (
+  obj: any,
+  propPath: ReadonlyArray<string>,
+) => {
+  // Needed to check if we are currently inside the remoteRouters object
+  const maybeRemoteRoutersKey = propPath[propPath.length - 3];
```
nambrot, line 23: Don't we need to bound checking? I.e. what if the path is only 2 elements, then it will be -1, thus it will check for the last element?

typescript/utils/src/objects.ts
```diff
@@ -337,3 +337,82 @@ export function mustGet<T>(obj: Record<string, T>, key: string): T {
...
+        .map((obj) =>
+          internalFormatObj(obj, formatter, [...propPath], maxDepth),
+        )
+        .filter((obj) => obj.shouldInclude)
+        .map((obj) => obj.formattedValue),
+      shouldInclude: true,
+    };
+  } else if (isObject(obj)) {
+    const newObj = Object.entries(obj).reduce((formattedObj, [key, value]) => {
+      const { formattedValue, shouldInclude } = internalFormatObj(
```
nambrot, line 400: Could we simplify this? I.e. if the value is undefined, just omit the key?

typescript/cli/src/check/warp.ts
xeno097: If an element does not exist at a given index, we will get undefined as value, it will behave like this even with negative indexes

typescript/sdk/src/token/configUtils.ts
```diff
@@ -90,3 +96,51 @@ export async function expandWarpDeployConfig(
...
+};
+
+type HypTokenRouterConfigKey = keyof HypTokenRouterConfig;
+const KEYS_TO_IGNORE: HypTokenRouterConfigKey[] = ['totalSupply'];
+
+function sanitizeConfig(obj: HypTokenRouterConfig): any {
+  // Remove keys from obj
+  return Object.fromEntries(
+    Object.entries(obj).filter(
+      ([key]) => !KEYS_TO_IGNORE.includes(key as HypTokenRouterConfigKey),
```
nambrot, line 131: Should we not move this into `transformWarpDeployConfigToCheck`?

typescript/utils/src/objects.ts
```diff
@@ -337,3 +337,67 @@ export function mustGet<T>(obj: Record<string, T>, key: string): T {
...
+export type TransformObjectTransformer = (
+  obj: any,
+  propPath: ReadonlyArray<string>,
+) => any;
+
+/**
+ * Recursively applies `formatter` to the provided object
+ *
+ * @param obj
+ * @param transformer a function that takes an object and formats it, if the `shouldInclude` property in the return value is set to false, the current value won't be included in the final result
```
nambrot, line 350: This comment is no longer valid yes?

typescript/utils/src/objects.ts
```diff
@@ -337,3 +337,67 @@ export function mustGet<T>(obj: Record<string, T>, key: string): T {
...
+  if (propPath.length > maxDepth) {
+    throw new Error(`transformObj went too deep. Max depth is ${maxDepth}`);
+  }
+
+  if (Array.isArray(obj)) {
+    return obj.map((obj) =>
+      internalTransformObj(obj, transformer, [...propPath], maxDepth),
+    );
+  } else if (isObject(obj)) {
+    const newObj = Object.entries(obj).reduce(
```
nambrot, line 378: This is a nit, but IMO what reads better is to map over the entries with `internalTransformObj` and then filter out where the values are empty, before caling `fromEntries`. Wdyt?

typescript/utils/src/objects.ts
```diff
@@ -337,3 +337,67 @@ export function mustGet<T>(obj: Record<string, T>, key: string): T {
...
+        if (
+          transformedValueValue !== undefined &&
+          transformedValueValue !== null
+        ) {
+          transformedObj[key] = transformedValueValue;
+        }
+
+        return transformedObj;
+      },
+      {} as any,
```
nambrot, line 396: Why is this cast necessary?

nambrot (approved):

typescript/utils/src/objects.ts
xeno097: agree! making the change now

typescript/utils/src/objects.ts
xeno097: It was needed because of this

image

typescript/sdk/src/index.ts
```diff
@@ -13,10 +13,10 @@ export {
 export { S3Config, S3Receipt, S3Wrapper } from './aws/s3.js';
 export { S3Validator } from './aws/validator.js';
 export {
-  TOKEN_EXCHANGE_RATE_DECIMALS_ETHEREUM,
-  TOKEN_EXCHANGE_RATE_SCALE_ETHEREUM,
   getProtocolExchangeRateDecimals,
   getProtocolExchangeRateScale,
+  TOKEN_EXCHANGE_RATE_DECIMALS_ETHEREUM,
```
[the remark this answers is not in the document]
xeno097: It is. I'm going to revert it to how it was, it seems that the file was formatted on save changing the order of the exports

PR #5607, opened 2025-03-05
feat: add provider instance metrics

kamiyaa:
### Description

Add provider instance metrics.
 - two int counters, one for creation, one for drop
 - This will allow us to better track how many providers we're constantly using.
 - Which will allow us to figure out if theres memory optimizations we can make by consolidating provider instances

```
hyperlane_provider_create_count
hyperlane_provider_drop_count
```

### Drive-by changes

 - updated cosmos e2e test to be more chain specific with metrics query
 - move ethereum termination_invariants into `ethereum` module
   - leave chain agnostic termination invariant functions intact

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5672

### Backward compatibility

Yes

### Testing

 - add e2e tests to test metrics exists


rust/main/hyperlane-base/src/settings/chains.rs
```diff
@@ -978,6 +985,11 @@ fn build_sealevel_rpc_client(
     let middleware_metrics = chain_conf.metrics_conf();
     let rpc_client_url = connection_conf.url.clone();
     let client_metrics = metrics.client_metrics();
+
+    // increment provider metric count
+    let chain_name = PrometheusConfig::chain_name(&middleware_metrics.chain);
```
kamiyaa, line 990: Not sure if we should increment this when we create an RPC Client or when we create an actual **Provider** struct. I think its usually 1:1 anyways though.

rust/main/hyperlane-base/src/settings/chains.rs
```diff
@@ -978,6 +985,11 @@ fn build_sealevel_rpc_client(
     let middleware_metrics = chain_conf.metrics_conf();
     let rpc_client_url = connection_conf.url.clone();
     let client_metrics = metrics.client_metrics();
+
```
kamiyaa, line 988: Sealevel metric increment here

rust/main/hyperlane-base/src/settings/chains.rs
```diff
@@ -1036,6 +1053,11 @@ fn build_cosmos_wasm_provider(
 ) -> ChainResult<CosmosWasmRpcProvider> {
     let middleware_metrics = chain_conf.metrics_conf();
     let client_metrics = metrics.client_metrics();
+
+    // increment provider metric count
```
kamiyaa, line 1057: cosmos wasm provider metric increment

rust/main/hyperlane-base/src/settings/chains.rs
```diff
@@ -960,10 +960,17 @@ impl ChainConf {
             signer = self.ethereum_signer().await?;
         }
         let metrics_conf = self.metrics_conf();
-        let rpc_metrics = Some(metrics.client_metrics());
+        let client_metrics = metrics.client_metrics();
+
+        // increment provider metric countclient_metrics
```
kamiyaa, line 965: ethereum provider metric increment here

rust/main/hyperlane-base/src/settings/chains.rs
daniel-savu: it'd be much cleaner to increment in the provider constructor. We want new code that creates providers to automatically increment this metric, rather than needing it done manually

daniel-savu (commented):
Some refactoring may be needed to avoid duplication. For instance, I noticed that `SealevelRpcClient` can be instantiated in two ways: 1) via the `new` method, and 2) via the `from_rpc_client` method. 1) should call into 2) so that 2) can increment the metric.

Wherever `Clone` is derived, it should be manually implemented to increment the metric.

Also I'm not sure whether dropped providers will end up decrementing the metric - we should manually implement drop as well on the provider structs

rust/main/hyperlane-base/src/settings/chains.rs
kamiyaa: yeah, that makes sense

rust/main/chains/hyperlane-cosmos/src/providers/rpc/client.rs
```diff
@@ -132,6 +136,27 @@ impl CosmosRpcClient {
...
+    fn drop(&mut self) {
+        // decrement provider metric count
+        let chain_name = PrometheusConfig::chain_name(&self.metrics_config.chain);
+        self.metrics.decrement_provider_instance(chain_name);
+    }
+}
+
+impl Clone for CosmosRpcClient {
+    fn clone(&self) -> Self {
+        // decrement provider metric count
```
ameten, line 149:
```suggestion
        // increment provider metric count
```

ameten (approved):

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
```diff
@@ -283,6 +283,8 @@ pub fn provider_metrics_invariant_met(
     .iter()
     .sum::<u32>();
 
+    log!("Provider instance count: {}", provider_count);
+
     if provider_count == 0 {
```
daniel-savu, line 288: I think there'll be at least `expected_messages` providers since we create ephemeral ones to prep each metadata, so let's use that

rust/main/chains/hyperlane-cosmos/src/providers/rpc/client.rs
```diff
@@ -132,6 +136,27 @@ impl CosmosRpcClient {
...
+impl Clone for CosmosRpcClient {
+    fn clone(&self) -> Self {
+        // increment provider metric count
+        let chain_name = PrometheusConfig::chain_name(&self.metrics_config.chain);
+        self.metrics.increment_provider_instance(chain_name);
+        Self {
+            client: self.client.clone(),
+            metrics: self.metrics.clone(),
+            metrics_config: self.metrics_config.clone(),
+        }
```
daniel-savu, line 156: nit: rename the current `CosmosRpcClient::new` to `CosmosRpcClient::from_url` and create a new method called `new` that only increments the metric and calls self like is done here, to be able to just call `CosmosRpcClient::new` from here as well

daniel-savu (approved):
some nits and then we're good to merge

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
kamiyaa: Yeah, but we made it a guage, so it will always be in flux. Once the provider gets dropped, it will decrease. I'm not sure we can make any guarantees that it will be some number

daniel-savu: hm good point. In that case we can have two gauges instead, one tracking how many providers were ever created and one for how many were ever dropped. Then we'd plot the difference while still being able to check for correctness in e2e. Wdyt?

kamiyaa: oh good idea :+1:

rust/main/chains/hyperlane-cosmos/src/providers/rpc/client.rs
```diff
@@ -132,6 +145,27 @@ impl CosmosRpcClient {
...
```
daniel-savu, line 165: now that we have the `new` method, this should be
```suggestion
        Self::new(
            client: self.client.clone(),
            metrics: self.metrics.clone(),
            metrics_config: self.metrics_config.clone(),
        )
```

rust/main/utils/run-locally/src/sealevel/termination_invariants.rs
```diff
@@ -67,6 +67,7 @@ pub fn termination_invariants_met(
         RELAYER_METRICS_PORT,
         total_messages_expected,
         &hashmap! {"chain" => "sealeveltest1", "connection" => "rpc", "status" => "success"},
+        &hashmap! {},
```
daniel-savu, line 70: how come this is empty? IIUC we _are_ counting SVM provider instances, assuming an empty hashmap means we're not actually reading the provider count metric

kamiyaa: For some reason, `sealeveltest1` was only instantiating a small amount of providers. that was less than the expected message count.

daniel-savu: ah makes sense, we need to use the sealevel message count instead

PR #5612, opened 2025-03-06
fix: Move access to database into a blocking task

ameten:
### Description

Move access to database into a blocking task

The access to rocksdb is synchronous and we discovered that it take about 1 millisecond to make one retrieval of an insertion into a merkle tree from the database when we build internal merkle tree in relayer from scratch. As the result, the worker thread which executes the task, is blocked by access to the database and cannot be used to execute other scheduled tasks.

This change moves the access to the database into a separate thread in the blocking pool. It frees up the worker thread to execute other tasks.

### Related issues

### Backward compatibility

Yes

### Testing

Manual run of relayer locally

### DIFF
--- a/rust/main/agents/relayer/src/merkle_tree/processor.rs
+++ b/rust/main/agents/relayer/src/merkle_tree/processor.rs
@@ -21,6 +21,8 @@ use crate::processor::ProcessorExt;
 
 use super::builder::MerkleTreeBuilder;
 
+const PREFIX: &str = "processor::merkle_tree";
+
 /// Finds unprocessed merkle tree insertions and adds them to the prover sync
 #[derive(new)]
 pub struct MerkleTreeProcessor {
@@ -44,7 +46,7 @@ impl Debug for MerkleTreeProcessor {
 #[async_trait]
 impl ProcessorExt for MerkleTreeProcessor {
     fn name(&self) -> String {
-        format!("processor::merkle_tree::{}", self.domain().name())
+        format!("{}::{}", PREFIX, self.domain().name())
     }
 
     /// The domain this processor is getting merkle tree hook insertions from.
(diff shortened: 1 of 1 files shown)


rust/main/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -81,34 +83,57 @@ impl ProcessorExt for MerkleTreeProcessor {
...
-                .merkle_tree_retrieve_insertion_total_elapsed_micros
-                .inc_by(begin.elapsed().as_micros() as u64);
-            self.metrics.merkle_tree_retrieve_insertions_count.inc();
+    async fn next_unprocessed_leaf_task(&mut self) -> Result<Option<MerkleTreeInsertion>> {
+        let index = self.leaf_index;
+        let db = self.db.clone();
+        let metrics = self.metrics.clone();
+        let name = format!("{}::retrieval::{}::{}", PREFIX, self.domain(), index);
+        let insertion = tokio::task::Builder::new()
+            .name(&name)
```
daniel-savu, line 92: nice

rust/main/agents/relayer/src/merkle_tree/processor.rs
```diff
@@ -81,34 +83,57 @@ impl ProcessorExt for MerkleTreeProcessor {
...
+            Self::update_metrics(metrics, &insertion, &begin);
             Some(insertion)
         } else {
-            trace!(leaf_index=?self.leaf_index, "No merkle tree insertion found in DB for leaf index, waiting for it to be indexed");
+            trace!(
+                ?leaf_index,
+                "No merkle tree insertion found in DB for leaf index, waiting for it to be indexed"
+            );
             None
         };
```
daniel-savu, line 114: what's the upside of having these in `spawn_blocking`? Would've expected to only have the db read in there

daniel-savu (approved):

rust/main/agents/relayer/src/merkle_tree/processor.rs
ameten: Good catch. No actual upside. Let me refactor it.

ameten: Done

PR #5613, opened 2025-03-06
ci: Pack and Install CLI in ci

### LINKED ISSUE #5616: Build and Install the CLI in CI to test successfull builds

xeno097:
### Description

This PR adds a CI step where the CLI is locally packaged and installed

### Drive-by changes

- Update e2e tests to use locally built cli instead of relying on `yarn run`

### Issue

Fixes #5616 

### Backward compatibility

- Yes

### Testing

- Manual


.github/workflows/test.yml
paulbalaji, line 1: is there a yarn-based equivalent for this?

paulbalaji: also, i think we should move all the `yarn workspace` style cli invocations to this way in the cli e2e tests

xeno097: `yarn pack` would be the equivalent

xeno097: > also, I think we should move all the yarn workspace style cli invocations to this way in the cli e2e tests

Yes, it would be better to test using the packaged version, I'm going to update them

typescript/cli/src/tests/commands/helpers.ts
yorhodes, line 1: instead of duplicating all the invocations can we just change the command prefix?

xeno097: doesn't work, unfortunately, I tried to change the prefix programmatically, but the resulting command is not formatted correctly, resulting in the command not being correct:

image

yorhodes: you are changing the prefix with the log level var so thats a little odd to me
the args are also programmatically set in the string

xeno097: Fixed it! found out how to make it work well

.github/workflows/test.yml
```diff
@@ -119,9 +119,29 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: install-hyperlane-cli
+        id: install-hyperlane-cli
+        uses: ./.github/actions/install-cli
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: Test run the CLI
+        run: hyperlane --version
```
paulbalaji, line 138: i don't think test is required now that cli-e2e uses the packed/installed cli

.github/workflows/test.yml
```diff
@@ -158,13 +178,20 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
       - name: Checkout registry
         uses: ./.github/actions/checkout-registry
 
       - name: CLI e2e tests (${{ matrix.test }})
         run: yarn --cwd typescript/cli test:e2e
         env:
           CLI_E2E_TEST: ${{ matrix.test }}
+          IS_CI: true
```
paulbalaji, line 194: instead of using `IS_CI`, just use `CI` - as github actions should automatically set this to true in CI https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

.github/workflows/test.yml
xeno097: Wouldn't it be better to leave it for clarity so that instead of having to go to the CI logs, we can just know why tests are failing by knowing that the single cli install action failed?

paulbalaji: in that case, can we make cli-e2e not depend on this test? otherwise that's an extra 3m added to each CI run for no good reason

paulbalaji: so cli-e2e-matrix and cli-install-test can both run in parallel and overall ci time is unchanged

xeno097: sounds good!

paulbalaji (dismissed):

typescript/cli/package.json
```diff
@@ -45,13 +45,16 @@
...
-    "typescript": "5.3.3"
+    "typescript": "5.3.3",
+    "webpack": "^5.98.0",
+    "webpack-cli": "^6.0.1"
   },
   "scripts": {
     "hyperlane": "node ./dist/cli.js",
-    "build": "yarn version:update && tsc",
+    "build": "yarn version:update && tsc && yarn bundle",
+    "bundle": "rm -rf ./bundle && webpack --config webpack.config.mjs",
```
paulbalaji, line 55: do we need to bundle on every build? can we just call `yarn bundle` as part of the `yarn pack` CI step?

typescript/cli/package.json
```diff
@@ -73,7 +76,7 @@
     "node": ">=16"
   },
   "bin": {
-    "hyperlane": "./dist/cli.js"
+    "hyperlane": "./cli-bundle/cli.js"
   },
```
paulbalaji, line 80: why this change?

typescript/cli/package.json
xeno097: yes, to update the bundled build so that when we install the cli we always have the last version

paulbalaji:
lgtm but would like @cmcewen and @yorhodes to review the webpack changes

paulbalaji:
```
npm error code 127
npm error git dep preparation failed
npm error command /usr/local/bin/node /usr/local/lib/node_modules/npm/bin/npm-cli.js install --force --cache=/home/runner/.npm --prefer-offline=false --prefer-online=false --offline=false --no-progress --no-save --no-audit --include=dev --include=peer --include=optional --no-package-lock-only --no-dry-run
npm error npm warn using --force Recommended protections disabled.
npm error npm error code 127
npm error npm error path /home/runner/.npm/_cacache/tmp/git-cloneMG8r0k
npm error npm error command failed
npm error npm error command sh -c husky
npm error npm error sh: 1: husky: not found
```
https://github.com/hyperlane-xyz/hyperlane-monorepo/actions/runs/14082321382/job/39437992157?pr=5613#step:3:636

typescript/cli/package.json
```diff
@@ -49,9 +50,10 @@
...
   "scripts": {
     "hyperlane": "node ./dist/cli.js",
-    "build": "yarn version:update && tsc",
+    "build": "yarn version:update && tsc && yarn bundle",
+    "bundle": "rm -rf ./cli-bundle && ncc build ./dist/cli.js -o cli-bundle && node ./scripts/ncc.post-bundle.mjs",
```
antigremlin, line 54: Why don't we invoke `bundle` from `prepack`? (that would require removing `yarn bundle` from `build` of course)

antigremlin (commented):
The `ncc` part looks great.

typescript/cli/scripts/ncc.post-bundle.mjs
```diff
@@ -0,0 +1,42 @@
...
+const shebang = '#! /usr/bin/env node';
+const dirnameDef = `import { fileURLToPath } from 'url';
+import path from 'path';
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);`;
+
+async function prepend() {
+  try {
+    const content = await readFile(outputFile, 'utf8');
+    if (!content.startsWith(shebang) && content.includes(dirnameDef)) {
```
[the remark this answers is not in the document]
xeno097: When working on this, I tried to run the CLI without the `__dirname` definition and the CLI was crashing because it seems that there is something that requires it. Adding it fixed the issue

.github/workflows/test.yml
```diff
@@ -153,6 +169,12 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
 
+      - name: install-hyperlane-cli
+        id: install-hyperlane-cli
+        uses: ./.github/actions/install-cli
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
```
yorhodes, line 177: is this used?

typescript/cli/package.json
yorhodes: agree that we dont need to bundle in build

typescript/cli/package.json
```diff
@@ -2,78 +2,75 @@
...
-    "version:update": "echo \"export const VERSION = '$npm_package_version';\" > src/version.ts"
+    "version:update": "echo \"export const VERSION = '$npm_package_version';\" > src/version.ts",
+    "prepack": "yarn bundle"
   },
   "files": [
-    "./dist",
-    "./examples",
-    "package.json"
+    "./cli-bundle",
+    "./examples"
```
yorhodes, line 64: should we not still include the pjson?

typescript/cli/scripts/ncc.post-bundle.mjs
yorhodes, line 1: im surprised this isnt handled by ncc

.github/actions/install-cli/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      id: pack-cli
+      run: |
+        cd typescript/cli
+        yarn pack
+
+    - name: Install the packaged CLI
+      shell: bash
+      run: |
+        cd typescript/cli
+        npm i -g package.tgz
```
yorhodes, line 28: as discussed you will add the matrix for os and nodejs version in a followup

yorhodes (approved):

typescript/cli/package.json
xeno097: Even if not included in the `files` array explicitly, it will be when publishing 
ref https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files:~:text=regardless%20of%20settings%3A-,package.json,-README

typescript/cli/scripts/ncc.post-bundle.mjs
xeno097: Yes, but at least it is something simple, and the CLI works fine without any major changes

.github/workflows/test.yml
xeno097: yep, we install the cli and then when we run the e2e tests we use the bundled cli instead of relying on the monorepo code

typescript/cli/package.json
yorhodes: got it, so this diff is an explicit removal but implicit inclusion? seems better to be explicit

xeno097: makes sense,  i'll add it back!

PR #5614, opened 2025-03-06
feat: Add chains to relayer balance overrides

Mo-Hussain:
### Description

- Add arthera, sophon and tangle to the overrides, we have to keep the balances temporarily low as we cannot acquire gas tokens for these chains at the the moment
- All three chains have had very small to no burn over the last month. 
- These overrides set the desired balance to a level higher than the already existing balance

### Testing

Manual


typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
```diff
@@ -8,7 +8,7 @@
   "arbitrumnova": 0.05,
   "arcadia": 0.00952,
   "artela": 2220,
-  "arthera": 1010,
```
tkporter, line 11: I assume this was originally so high just because of the min $ usd thing?

tkporter (approved):

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalances.json
Mo-Hussain: Yes

PR #5621, opened 2025-03-06
feat: check-deploy with --registry

### LINKED ISSUE #5240: infra's check-deploy can be run with a command line arg for registry commit
This will allow verifies to not have to check out the registry commit locally themselves.

ltyu:
### Description
This PR 
- allows check-deploy to run with a `--registry` instead of having to check out the Registry branch. This is similar to how CLI's parameter.
- Update getConfigFromMergedRegistry() to fetch from the array in `--registry`
- Adds new functions getWarpCoreConfigFromMergedRegistry() and getWarpAddressesFromMergedRegistry() to fetch from the array in `--registry`
- requires `--warpRouteId` to be filled in and no longer depend on the warpRouteIds mapping

### Related issues
- Fixes #5240 

### Backward compatibility
Yes - We now fetch from the FileSystem and Github Registry instead of just FileSystem

### Testing
Manual


typescript/infra/scripts/check/check-utils.ts
```diff
@@ -63,7 +67,9 @@ export function getCheckWarpDeployArgs() {
 }
 
 export function getCheckDeployArgs() {
-  return withWarpRouteId(withModule(getCheckBaseArgs()));
+  return withGithubCommit(
+    withWarpRouteIdNoChoices(withModule(getCheckBaseArgs())),
+  );
 }
```
paulbalaji, line 73: this will break a lot of existing flows if we're now demanding that a github commit has to be required with every check-deploy call - imo it should still be optional

ltyu: it is optional though

typescript/infra/config/registry.ts
```diff
@@ -7,7 +7,10 @@ import {
   PartialRegistry,
   warpConfigToWarpAddresses,
 } from '@hyperlane-xyz/registry';
-import { FileSystemRegistry } from '@hyperlane-xyz/registry/fs';
+import {
+  FileSystemRegistry,
+  getRegistry as getMergedRegistry,
```
nambrot, line 12: Why call this merged registry? Just call this registry, to the caller, it shouldn't matter if its a mergedRegistry no?

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -63,7 +67,9 @@ export function getCheckWarpDeployArgs() {
...
 
 export function getCheckDeployArgs() {
-  return withWarpRouteId(withModule(getCheckBaseArgs()));
+  return withregistryUris(
```
nambrot, line 70: Nit: 

```suggestion
  return withRegistryUris(
```

typescript/infra/scripts/github-utils.ts
```diff
@@ -0,0 +1,8 @@
+import { Argv } from 'yargs';
+
+export function withregistryUris<T>(args: Argv<T>) {
```
nambrot, line 3: Nit:

```suggestion
export function withRegistryUris<T>(args: Argv<T>) {
```

nambrot (approved):
Would like @paulbalaji 's review on this too

typescript/infra/scripts/check/check-utils.ts
paulbalaji: the change from withWarpRouteId -> withWarpRouteIdNoChoices has the implication that `warpRouteIds` are now demanded, as per this line:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/faabe65cf579101d6d68a5badd6ba049848768cf/typescript/infra/scripts/agent-utils.ts#L213

which makes sense in check-warp-deploy but not in check-deploy

typescript/infra/config/registry.ts
ltyu: there's an existing getRegistry

typescript/infra/scripts/check/check-utils.ts
ltyu: Gotcha. removed!

typescript/infra/config/registry.ts
```diff
@@ -104,6 +108,36 @@ export function getWarpCoreConfig(warpRouteId: string): WarpCoreConfig {
   return warpRouteConfig;
 }
 
+export async function getWarpCoreConfigFromGithubFileSystem(
```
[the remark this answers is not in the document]
ltyu: updated

typescript/infra/scripts/agent-utils.ts
```diff
@@ -208,6 +208,10 @@ export function withWarpRouteId<T>(args: Argv<T>) {
     .choices('warpRouteId', Object.values(WarpRouteIds));
 }
 
+export function withWarpRouteIdNoChoices<T>(args: Argv<T>) {
+  return args.describe('warpRouteId', 'warp route id').string('warpRouteId');
+}
```
paulbalaji, line 213: i wonder if this is the chance to rename this to `withWarpRouteId` and rename the existing `withWarpRouteId` -> `withKnownWarpRouteId`

and then in your follow up PR the `withWarpRouteIdRequired` makes more sense

ltyu: Sure, i wanted to reduce the diffs but can do this as well

paulbalaji (approved):

PR #5624, opened 2025-03-06
feat: Expose Additional Validator Lifecycle Metrics in Prometheus

keefertaylor:
### Description

This change adds three additional prometheus metrics that are useful for understanding the health and states of a validator:

- `announced`: Whether the validator has been announced, or is waiting for announcement in order to sign and submit checkpoints
- `backfill_complete`: Whether the validator has successfully migrated through the entire checkpoint backlog. Useful to know that the nodes you are talking to have adequate historical state.
- `reached_initial_consistency`: Whether the validator has hit the chain head at least once. Useful to know that the validator is in sync (vs just looking at `contract_sync_height`, where a validator could be syncing forward but syncing slower than the tip of the chain. 

These metrics are a simple 0/1 flag that flip to 1 when the state has been achieved.

### Testing

Tessellated has these flags deployed in our infrastructure for the last ~3 weeks, and use them in our monitoring to diagnose validator issues


keefertaylor:
Addressing comment from the [initial pr](https://github.com/Tessellated-io/hyperlane-monorepo/pull/3#discussion_r1981793387): 

> I suppose that validator will set the metric once when it (re)-started. 

Yes that's correct. It gets set pretty quickly at startup. On our side, our monitoring generally looks for bad values for some period of time (~5min) before escalating to on-call for exactly this reason. It's pretty common across software for some values to be out of sync on a restart, ex. Nodes often reported `synced=false` for a moment after they restart, even if they are in sync. 

It's also worth noting that if the config for a validator could cause it to be reannounced across restarts (if the service configuration changed). In that case, monitoring to know that something is wrong is helpful. 

> I wonder if we really need it as a field. We could have just a method which gets the metric from core_metrics and sets it.

I'm not a native rust developer and as familiar with the hyperlane codebase so I wasn't sure what the convention was. Would you like me to make a change for it?

rust/main/agents/validator/src/validator.rs
```diff
@@ -119,6 +120,10 @@ impl BaseAgent for Validator {
             )
             .await?;
 
+        let announced = metrics
+            .announced()
+            .with_label_values(&[&settings.origin_chain.name().to_string()]);
```
ameten, line 125: Since we set `announced` metrics once, this code can run when validator is announced, namely line `self.announced.set(1);` will include the code above, but probably, better just to extract getting the metric and setting it into a separate method.

keefertaylor: I've combined the two lines into a single helper inside of core metrics. I *think* think is what you're asking for, but if I misunderstood please correct me.

ameten (approved):

PR #5628, opened 2025-03-07
feat(balances): Update write alert scripts with helpul prompts

Mo-Hussain:
### Description

- Updates the write alert scripts with helpful prompts and general QoL improvements
- Checks if the proposed change introduces a regression i.e remove a chain from the existing alert, logs an error and aborts
- Read the actual alert threshold via grafana api, compares with proposed thresholds, prompts the user to confirm that they want to introduce these changes
image
- Check if proposed update will lead to alert firing and prompts the user to confirm with these changes 
image

### Testing

Manual

### DIFF
--- a/typescript/infra/scripts/funding/write-alert.ts
+++ b/typescript/infra/scripts/funding/write-alert.ts
@@ -258,19 +258,19 @@ function generateDiffTable(
 }
 
 function handleMissingChainErrors(missingChainErrors: RegressionError[]) {
-  if (missingChainErrors.length > 0) {
-    for (const error of missingChainErrors) {
-      rootLogger.error(
-        `Missing thresholds for chains: ${error.missingChains.join(', ')} for ${
-          error.alertType
-        } config`,
-      );
-    }
+  if (missingChainErrors.length === 0) return;
+
+  for (const error of missingChainErrors) {
     rootLogger.error(
-      `Aborting updating alerts due to missing thresholds in config`,
+      `Missing thresholds for chains: ${error.missingChains.join(', ')} for ${
+        error.alertType
+      } config`,
     );
-    process.exit(1);
   }
+  rootLogger.error(
+    `Aborting updating alerts due to missing thresholds in config`,
+  );
+  process.exit(1);
 }
 
 async function confirmFiringAlerts(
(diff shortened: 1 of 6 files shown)


typescript/infra/src/funding/alerts.ts
```diff
@@ -34,7 +34,7 @@ function getBalanceAlertRegex(walletName: WalletName): RegExp {
     case WalletName.KeyFunder:
       return /wallet_name="key-funder", chain="([^"]+)"[^-]+ - ([0-9.]+)/g;
     case WalletName.Relayer:
-      return /wallet_name="relayer", chain="([^"]+)"[^-]+ - ([0-9.]+)/g;
+      return /wallet_name="relayer", chain="([^"]+)", hyperlane_context="hyperlane"[^-]+ - ([0-9.]+)/g;
```
Mo-Hussain, line 37: We were mixing up threshold reads for arbitrum and neutron where we have thresholds for the neutron context. As we are deprecating this context soon, we will just read hyperlane context values

typescript/infra/scripts/funding/write-alert.ts
```diff
@@ -9,66 +12,297 @@ import {
...
+  // fetch alerting chains
+  const alertingChains = await fetchFiringThresholdAlert(query);
 
-    // only change the query
-    await updateGrafanaAlert(
-      alertConfigMapping[alert].grafanaAlertId,
-      alertRule.rawData,
-      query,
-      saToken,
+  if (alertingChains.length > 0) {
```
paulbalaji, line 285: nit: invert then early out, ditto in handleMissingChainErrors

typescript/infra/scripts/funding/write-alert.ts
```diff
@@ -9,66 +12,297 @@ import {
...
+      })),
     );
 
-    rootLogger.info(`Updated ${alert} alert`);
+    const confirmed = await confirm({
+      message: `Do you want to proceed with updating the alert thresholds for ${alert}?`,
+    });
+    if (!confirmed) {
+      rootLogger.info(
+        `Exiting without updating any alerts, this is to avoid thresholds from being out of sync`,
```
paulbalaji, line 302: "this is to avoid thresholds from being out of sync"

wdym by this if we're not updating the alert

typescript/infra/src/config/funding/grafanaAlerts.ts
```diff
@@ -57,6 +58,10 @@ export const alertConfigMapping: Record<AlertType, AlertConfig> = {
       header: LOW_URGENCY_KEY_FUNDER_HEADER,
       footer: LOW_URGENCY_KEY_FUNDER_FOOTER,
     },
+    writePriority:
```
paulbalaji, line 61: why is the multiplier the priority?

typescript/infra/scripts/funding/write-alert.ts
Mo-Hussain: I could make this clearer but the intention is the exit the whole script without updating any alerts, this is to avoid selectively updating some alerts that may lead to a high urgency threshold being higher than a low urgency threshold

typescript/infra/src/config/funding/grafanaAlerts.ts
Mo-Hussain: Because we want to write the alerts with the highest thresholds first, given that we will be increasing thresholds in the main, this will reduce the risk of threshold being out of sync e.g.

1. LowUrgency - write succeeds
2. LowUrgencyEng - write fails  (due to API error for example)
3. HighUrgencyEng - write skipped (If this was not skipped, it could have increased HighUrgencyEng thresholds above LowUrgencyEng) 

I'll add a comment to explain

typescript/infra/scripts/funding/write-alert.ts
Mo-Hussain: Can you expand?

paulbalaji (approved):

PR #5629, opened 2025-03-07
feat(balances): Use desiredRelayerBalances config in key funder

Mo-Hussain:
### Description

- use the `desiredRelayerBalances.json` config when running key-funder
- change the threshold for which we fund the relayer from 50% of the desired balance to 60%.

### Drive-by changes
- remove abstract for desired relayer overrides

### Testing

Manual - Tested running key-funder locally with new desired relayer balances


typescript/infra/scripts/funding/fund-keys-from-deployer.ts
```diff
@@ -81,8 +81,8 @@ const walletBalanceGauge = getWalletBalanceGauge(
   Object.keys(constMetricLabels),
 );
 
-// Min delta is 50% of the desired balance
-const MIN_DELTA_NUMERATOR = ethers.BigNumber.from(5);
+// Min delta is 60% of the desired balance
```
Mo-Hussain, line 84: Context for this change https://www.notion.so/Balances-Alerts-Revamp-17c6d35200d680648350c247f292f6b0?d=17d6d35200d680cbbb43001c73385f1e&pvs=4#17c6d35200d6801dbd2cf508583b5b01

typescript/infra/config/environments/mainnet3/funding.ts
```diff
@@ -25,143 +34,7 @@ export const keyFunderConfig: KeyFunderConfig<
...
-    // ignore non-evm chains
-    injective: '0',
-    neutron: '0',
-    osmosis: '0',
-    solanamainnet: '0',
-    eclipsemainnet: '0',
-    soon: '0',
-    sonicsvm: '0',
-  },
+  desiredBalancePerChain: desiredRelayerBalancePerChain,
```
paulbalaji, line 37: is the typing still sound here? e.g. if you remove a chain from the desired relayer balances does it correctly inform you the type is wrong

paulbalaji (approved):

PR #5631, opened 2025-03-07
feat: rate limit message for MINT_LIMITED Token standards

Xaroz:
### Description

Add rate limit message to `validateDestinationCollateral` depending if the token standard is is part of `MINT_LIMITED_STANDARDS`, I decided to put this here instead of modifying `isDestinationCollateralSufficient` because these were used on other places (like the [UI](https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/blob/fa79e3b4e57460e6995f01eda8b0a74f1e6acd28/src/features/transfer/useTokenTransfer.ts#L121))

### Drive-by changes

Clean up condition check to use less code

### Backward compatibility

Yes
### Testing

UI
Screenshot 2025-03-07 at 12 48 50 PM

### DIFF
--- a/typescript/sdk/src/warp/WarpCore.ts
+++ b/typescript/sdk/src/warp/WarpCore.ts
@@ -26,6 +26,7 @@ import {
   TOKEN_COLLATERALIZED_STANDARDS,
   TOKEN_STANDARD_TO_PROVIDER_TYPE,
   TokenStandard,
+  VSXERC20_STANDARDS,
 } from '../token/TokenStandard.js';
 import { EVM_TRANSFER_REMOTE_GAS_ESTIMATE } from '../token/adapters/EvmTokenAdapter.js';
 import { IHypXERC20Adapter } from '../token/adapters/ITokenAdapter.js';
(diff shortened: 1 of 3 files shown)


typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -790,7 +786,19 @@ export class WarpCore {
...
     });
-    if (!valid) return { amount: 'Insufficient collateral on destination' };
+
+    if (!valid) {
+      const destinationName = this.multiProvider.getChainName(destination);
+      const destinationToken =
+        originTokenAmount.token.getConnectionForChain(destinationName)?.token;
+      if (
+        destinationToken &&
+        VSXERC20_STANDARDS.includes(destinationToken.standard)
```
yorhodes, line 796: I think this should maybe apply to all `MINT_LIMITED_STANDARDS`

yorhodes (approved):

typescript/sdk/src/warp/WarpCore.ts
Xaroz: done

PR #5640, opened 2025-03-10
feat: deploy to infinityvm, plume

paulbalaji:
### Description

feat: deploy to infinityvm, plume

### Drive-by changes

Estimate gas + add buffer on mailbox initialization, setting ISMs, setting IGP configs, setting routing hooks.

### Related issues

https://github.com/hyperlane-xyz/hyperlane-registry/pull/659

### Testing

manual


typescript/infra/config/environments/mainnet3/igp.ts
```diff
@@ -36,6 +36,17 @@ export function getOverheadWithOverrides(local: ChainName, remote: ChainName) {
   return overhead;
 }
 
+function getOracleConfigWithOverrides(chain: ChainName) {
+  const oracleConfig = storageGasOracleConfig[chain];
+  if (chain === 'infinityvm') {
+    // For InfinityVM, override all remote chain gas configs to use 0 gas
+    for (const remoteConfig of Object.values(oracleConfig)) {
+      remoteConfig.gasPrice = '0';
```
tkporter, line 44: does this mean that we quote nothing when sending to infinityvm right now?

fwiw we just care about Solana <> InfinityVM being free but all the rest would want it to not be free

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
```diff
@@ -392,9 +393,15 @@ export class HyperlaneHookDeployer extends HyperlaneDeployer<
...
         'Setting routing hooks',
       );
+      const estimatedGas = await routingHook.estimateGas.setHooks(
+        routingConfigs,
+      );
       return this.multiProvider.handleTx(
         chain,
-        routingHook.setHooks(routingConfigs, overrides),
+        routingHook.setHooks(routingConfigs, {
+          gasLimit: addBufferToGasLimit(estimatedGas),
```
tkporter, line 402: is all this stuff bc of infinity's gasless setup?

paulbalaji: not explicitly because of infinity, I've noticed it on other chains as well but never looked into the root cause til now

typescript/infra/config/environments/mainnet3/igp.ts
paulbalaji: so e.g. eth->infinity would still quote a fee, but infinity->eth would not

Mo-Hussain (approved):

PR #5645, opened 2025-03-11
feat: add OpStack L2 to L1 ITokenBridge

gitmp01:
### Description

 1. Adds `ValueTransferBridgeNative` router abstracting the transfer of ETH from L2 to L1
 2. Adds `OPValueTransferBridgeNative` offering an implementation for 1. on Optimism
 3. Adds hooks and required ISM for 2.
 4. Adds an `OPStackService` to the CCIP-read server handling the CCIP-read requests


solidity/contracts/ValueTransferBridge.sol
```diff
@@ -0,0 +1,194 @@
...
+            unchecked {
+                ++i;
+            }
+        }
+    }
+
+    function transferRemote(
+        uint32 destinationDomain,
+        bytes32 recipient,
+        address[] memory tokens,
```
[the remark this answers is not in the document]
gitmp01: Apologies, I totally forgot about that, changing

solidity/contracts/ValueTransferBridge.sol
```diff
@@ -0,0 +1,194 @@
...
+        Quotes memory quotes = quoteTransferRemote(
+            destinationDomain,
+            recipient,
+            tokens,
+            amounts
+        );
+
+        bytes memory data = abi.encode(tokens);
+
+        uint256 proveFees = mailbox.quoteDispatch(
```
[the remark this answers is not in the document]
gitmp01: I see, but how do we expect folks would be able to set one way or the other (secure message and/or value transfer)? Through the `metadata` parameter of the `postDispatch` method or through the contract storage (i.e. if the isms and the ERC20 are set) ?

solidity/contracts/ValueTransferBridge.sol
```diff
@@ -0,0 +1,194 @@
...
+            data
+        );
+
+        uint256 finalizeFees = mailbox.quoteDispatch(
+            destinationDomain,
+            finalizeWithdrawalIsm,
+            data
+        );
+
+        transferId = keccak256(
```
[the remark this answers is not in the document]
gitmp01: Yeah I'd definitely need to revise this, good point

solidity/contracts/ValueTransferBridge.sol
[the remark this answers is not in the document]
gitmp01: Another option would be to make the `ValueTransferBridge` implementing the `IPostDispatchHook` interface and pass it as a `childHook` to the OpL2ToL1Hook contract. It's cleaner actually, but I'll wait for a confirmation before proceeding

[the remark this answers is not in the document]
gitmp01: There would be the simpler option to make the `ValueTransferBridge` implement the `IPostDispatchHook` interface and provide it as an argument for the `OPL2ToL1Hook` (as a childHook), in this way the value transfer logic gets integrated straight away. 

What do you think?

solidity/contracts/ValueTransferBridge.sol
```diff
@@ -0,0 +1,194 @@
...
+import {OPL2ToL1Hook} from "./hooks/OPL2ToL1Hook.sol";
+
+struct Quotes {
+    address[] tokens;
+    uint256[] amounts;
+    uint256 dispatchingFees;
+}
+
+// TODO: add IValueTransferBridge
+contract ValueTransferBridge is OPL2ToL1Hook {
```
nambrot, line 18: So @yorhodes made a good point that it would be nice if we can generally fulfill some more of the warp route APIs (maybe even inherit from `HypNative`) here for the following reasons:

- Have a destination "router" contract (as the recipient of a Hyperlane message) that allows to do the transfer in reverse (don't have to implement the reverse). Since you need some kind of message recipient for both the prove/finzalizewithdrawal ISM anyways, it seems to make sense to make the latter the "canonical" one. (for the proveWithdrawal one, it probably makes sense to just have a no-op recipient?)
- Have the message (probably the one using finalizeWithdrawal) that gets send to a destination "router" contract confirm to the warp route message format (https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b14f997810ebd7dbdff2ac6622a149ae77010ae3/solidity/contracts/token/libs/TokenMessage.sol#L4)  so that offchain tooling like the explorer can show the information (like on https://explorer.hyperlane.xyz/message/0x0a75864aa47a2b037df9437c137a2740e8ea0e79256f92ee5d291a22322114b0). This would also allow us to track completion of a transfer like all other warp routes (by tracking this message)
- Minor, but supporting the `balanceOf` function like https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/88970a78c1ec356d60c0aa640f4be0a82abca619/solidity/contracts/token/HypNative.sol#L79 would be nice

solidity/contracts/ValueTransferBridge.sol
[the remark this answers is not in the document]
nambrot: I think as per https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5645#discussion_r1992558842, I think it would make the most sense to me to have the messageId of the message using the finalizeWithdrawalIsm to be the "transferID"

solidity/contracts/ValueTransferBridge.sol
gitmp01: This is my feedback, we can maybe continue on a call:

- Having a _"destination router"_ and "reverse" capabilities means the message recipient would implement the `HypNative` interface as well? Does the recipient become effectively a `ValueTransferBridge`?

- As per our previous discussions, I thought the prove/finalize ISM had to be the message recipient for the hyperlane msgs due to the CCIP-read requirement. I can just change the recipient of the finalize dispatch towards the ValueTransferBridge on L1 and let it specify the CCIPReadFinalizeIsm for the verification part. The recipient for the prove withdrawal message would be kept as is (the CCIPReadProveWithdrawalIsm)
- Ack on other bullets

From my understanding and also to have a recap of the requirements, we would need to:

1. Create an abstract contract `AbstractValueTransferBridge` derived from `HypNative` which abstraction should be compatible with any rollup bridges (OPStack, Orbit, etc). This would be implemented also by the recipient of the transfer. The hyperlane message sent to the recipient should be compliant with the `TokenMessage` format. The configured hook then will be responsible of dispatching hyperlane messages to the destination recipient accordingly with the bridge in use (i.e. prove/finalize for OP, `Outbox.executeTransaction` on Orbit)
2. Create an `OPValueTransferBridge` which provides an implementation of `AbstractValueTransferBridge` using the OP rollup bridge (this will be used for the Renzo api extension)
5. Change the `OPL2ToL1Hook` in order to derive from `OPValueTransferBridge` and provide an implementation for the `transferRemote` by using OP bridge for transferring value to L1. The `OPL2ToL1Hook`'s childHook will then be used for dispatching the prove/finalize messages as defined in 1.
6. Create an `IPostDispatchHook` (possible name `OPL2toL1ValueHook`?) which dispatches the messages required to finalize the withdrawal on L1. For instance when using Orbit there will be an `ArbL2ToL1ValueHook`. (Note: this will be the hook configured on `OPL2ToL1Hook` and `OPValueTransferBridge` )

Does this match what you had in mind? Sorry for the long comment, but it could be a valuable note for me to make amendments on a possible call.

yorhodes (commented):
1 variant that inherits/extends from `HypNative` (and calls `bridgeEthTo`)
1 variant that inherits/extends from `HypERC20Collateral` (and calls `bridgeERC20To`)

yorhodes (commented):
the (first) message we use to initiate the withdrawal will be sent by an `AbstractAuthorizedMessageId` hook extension and on the destination will initiate the withdrawal on the `ProveIsm`
the (second) message that the warp route sends will forward value over the OP bridge and on the destination will finalize the withdrawal on the `FinalizeIsm` -- benefit of this is the warp message is a progress indicator of the withdrawal and the value will be delivered to the warp route recipient atomically with this message delivery

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
```diff
@@ -0,0 +1,59 @@
...
+        uint32 _destinationDomain,
+        bytes32 _ism
+    ) AbstractMessageIdAuthHook(_mailbox, _destinationDomain, _ism) {}
+
+    /// @inheritdoc AbstractPostDispatchHook
+    function _quoteDispatch(
+        bytes calldata metadata,
+        bytes calldata message
+    ) internal view override returns (uint256) {
+        return mailbox.quoteDispatch(destinationDomain, ism, message, metadata);
```
nambrot, line 37: IIUC, this is for the quote in the sendMessageId dispatch yeah? If so, imo the metadata should include the gasLimit that the `OPL2ToL1ProveWithdrawalIsm` needs for the proving of the withdrawal, no?

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
```diff
@@ -0,0 +1,59 @@
...
+// ============ Internal Imports ============
+import {TokenMessage} from "../token/libs/TokenMessage.sol";
+import {StandardHookMetadata} from "./libs/StandardHookMetadata.sol";
+import {AbstractPostDispatchHook} from "./libs/AbstractPostDispatchHook.sol";
+import {AbstractMessageIdAuthHook} from "./libs/AbstractMessageIdAuthHook.sol";
+import {InterchainGasPaymaster} from "../hooks/igp/InterchainGasPaymaster.sol";
+
+/**
+ * @title OPL2ToL1CcipReadHook
+ * @notice Inform an OPL2ToL1ProveWithdrawalIsm that a withdrawal has been executed
```
nambrot, line 13: Initiated, yeah?

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
```diff
@@ -0,0 +1,59 @@
...
+        bytes calldata metadata,
+        bytes calldata message
+    ) internal view override returns (uint256) {
+        return mailbox.quoteDispatch(destinationDomain, ism, message, metadata);
+    }
+
+    /// @inheritdoc AbstractMessageIdAuthHook
+    function _sendMessageId(
+        bytes calldata metadata,
+        bytes calldata message
```
nambrot, line 43: It might be useful to name or at least comment the usages of metadata/message here because IIUC, this is the "parent" full Hyperlane message, that we send as the body of the "prove withdrawal" message yeah? Not sure if you talked with @yorhodes about this, but I wonder if its better if we deliniate the message via the body vs. the recipient?

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,136 @@
...
+    // TODO: remove
+    function setUrls(string[] memory _urls) external {
+        require(_urls.length > 0, "URLs array is empty");
+        urls = _urls;
+    }
+
+    function getOffchainVerifyInfo(
+        bytes calldata _message
+    ) external view override {
+        bytes memory ccipReadCallData = _areWeMessageRecipient(_message)
```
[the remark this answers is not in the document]
gitmp01: Of course, I can check that offchain :+1:

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
gitmp01: Yes, that's it. I did it but it looked like it didn't change anything. For instance if I override the StandardHookMetadata gas limit into the `sendMessageId` dispatch, I still see just one single payment to the IGP, is that correct? Shouldn't be there two payment events?

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
gitmp01: You mean wrapping it through `TokenMessage` ? If so, I got it should be done at a higher level (when sending the first message)

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,154 @@
...
+        bytes32,
+        bytes calldata _message
+    ) internal virtual override {
+        bytes32 recipient = _message.recipient();
+        uint256 amount = _message.amount();
+
+        Address.sendValue(payable(recipient.bytes32ToAddress()), amount);
+
+        emit ReceivedTransferRemote(_origin, recipient, amount);
+    }
```
yorhodes, line 153: this is identical to `HypNative.handle` behavior, why override?

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,154 @@
...
+ * @author Substance Labs
+ * @dev Derives from the Hyperlane native token router, but supports
+ * transfer of ERC20 token value
+ */
+abstract contract ValueTransferBridgeNative is HypNative {
+    using TypeCasts for bytes32;
+    using TypeCasts for address;
+    using TokenMessage for bytes;
+
+    uint32 constant HOOK_METADATA_GAS_LIMIT = 700_000;
```
yorhodes, line 23: this is quite high, how did you benchmark this?

yorhodes (commented):
just looked at the contracts so far
could you add comments explaining that prove and finalize must be by invoked by the same caller which is why we have both implemented in the same ISM?

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,136 @@
...
+        return true;
+    }
+
+    /**
+     * @dev no-op handle
+     */
+    function handle(
+        uint32 _origin,
+        bytes32 _sender,
+        bytes calldata _message
```
yorhodes, line 86: this is `body` just to be clear

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
```diff
@@ -0,0 +1,59 @@
...
+    ) internal override {
+        uint256 relayFees = mailbox.quoteDispatch(
+            destinationDomain,
+            ism,
+            message,
+            metadata
+        );
```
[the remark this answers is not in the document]
yorhodes: there is a `_refund` helper for this as well

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
yorhodes: I actually think its preferable for the hook to be agnostic to the message body

nambrot: Yeah the ISM as the recipient is totally fine, just saying that the differentiation (i.e. `_areWeMessageRecipient`) seems better suited via something in the message body of the extra message for proving. Think of this like a application-specific leading type byte that some folks are doing.

The `message` (i.e. message body) passed in the "extra" message for proving, is right now the underlying full hyperlane message (i.e. the messageBody + the recipient/destination domainParameters). I agree that the hook should be agnostic to the message body, but not sure if it should use that as the body for its (i.e. extra) message itself. You could imagine that it should simply by something like `"proving message" + message.id"` as the body.

solidity/contracts/token/ValueTransferBridgeNative.sol
gitmp01: Yeah that's a value for testing, the real one should be around 450k gas

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,136 @@
...
+import {AbstractCcipReadIsm} from "../ccip-read/AbstractCcipReadIsm.sol";
+import {ICcipReadIsm} from "../../interfaces/isms/ICcipReadIsm.sol";
+import {IMessageRecipient} from "../../interfaces/IMessageRecipient.sol";
+import {IOptimismPortal} from "../../interfaces/optimism/IOptimismPortal.sol";
+import {IInterchainSecurityModule, ISpecifiesInterchainSecurityModule} from "../../interfaces/IInterchainSecurityModule.sol";
+
+contract OPL2ToL1CcipReadIsm is
+    AbstractCcipReadIsm,
+    IMessageRecipient,
+    ISpecifiesInterchainSecurityModule
```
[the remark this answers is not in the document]
yorhodes: not that important, we just have client side tooling for deploying/configuring `MailboxClient`

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,154 @@
...
+    function transferRemote(
+        uint32 _destination,
+        bytes32 _recipient,
+        uint256 _amountOrId
+    ) external payable override returns (bytes32) {
+        bytes32 _router = _mustHaveRemoteRouter(_destination);
+        uint256 _value = msg.value - _amountOrId;
+        bytes memory _extraData = bytes("");
+
+        _l2BridgeTransferRemote(_destination, _router, _amountOrId, _extraData);
```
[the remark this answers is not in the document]
gitmp01: I'm afraid we wouldn't be able to get the correct value to pass to `mailbox.dispatch()` after the amount has been sent over the rollup bridge. I can move everything to the internal `_transferRemote` instead

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,154 @@
...
+        bytes memory _extraData = bytes("");
+
+        fees =
+            _Router_quoteDispatch(
+                _destination,
+                _tokenMessage,
+                _getHookMetadata(),
+                address(hook)
+            ) +
+            _l2BridgeQuoteTransferRemote(
```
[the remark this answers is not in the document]
gitmp01: See my previous comment

solidity/contracts/token/ValueTransferBridgeNative.sol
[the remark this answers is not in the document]
gitmp01: I managed to move the logic in `_transferFromSender`. Had to remove `IValueTransferBridge` interface though, since the derived contract required overriding `transferRemote` and at the same time I couldn't left it unimplemented because the same function is defined in `HypNative` (base contract)
(cc @nambrot)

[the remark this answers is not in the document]
gitmp01: Another consequence of this change is this
```diff
--- a/solidity/contracts/token/HypNative.sol
+++ b/solidity/contracts/token/HypNative.sol
@@ -85,7 +85,7 @@ contract HypNative is TokenRouter {
      */
     function _transferFromSender(
         uint256
-    ) internal pure override returns (bytes memory) {
+    ) internal virtual override returns (bytes memory) {
         return bytes(""); // no metadata
     }
```

Which had to be `virtual` in order to be implemented in the derived class and `pure` had been removed because we have an internal `_getL2Extradata()` function which can be implemented to attach extra bytes to the `_l2TransferRemote()`. I thought that devs may want to label the withdrawal with custom data, let me know if it makes sense.

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
gitmp01: Given the different nature of the second message (`TokenMessage`) we can't differentiate through a simple flagging byte (or boolean) in the message body without complicating things. Furthermore I can see `TokenMessage` has the metadata field, but `Message` doesn't. I'm not saying it's impossible, just saying that we would introduce unneccessary/ugly logic to achieve something that may not be that useful in the end given the inconsistency above. Just wanted to make you aware of that, let me know what you think

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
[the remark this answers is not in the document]
gitmp01: Something I didn't foresee is that as a result of this change we would have a bigger calldata on the finalization call, so more expensive. Shall I restore the check on chain? Thinking about it, this is just a function that does not consume anything, so we shouldn't focus on optimizations here

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
nambrot: Yeah I think that is a fair callout. I could imagine you could discriminate based on  something like message length (given that you don't really need to convey any info), but its debatable whether that's cleaner as well

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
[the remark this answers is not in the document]
nambrot: sgtm

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,131 @@
...
+        l1Domain = _l1Domain;
+        l2Bridge = _l2Bridge;
+        _transferOwnership(_msgSender()); // TODO: remove
+    }
+
+    function setL2Bridge(address _l2Bridge) external onlyOwner {
+        l2Bridge = _l2Bridge;
+    }
+
+    function quoteTransferRemote(
```
nambrot, line 50: This method should return `Quote[]` so that consumers of the `IValueTransferBridge` contracts can support bridges where the quote is in the token being transferred (OpStack bridge admittedly is not one of them)

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    function transferRemote(
+        uint32 /* _destination */,
+        bytes32 /* _recipient */,
+        uint256 /* _amount */,
+        bytes calldata /* _hookMetadata */,
+        address /* _hook */
+    ) external payable override returns (bytes32) {
+        revert NotImplemented();
```
nambrot, line 77: This is an interesting question, we originally offered this to allow callers to override the hook to pay lets say another relayer's IGP. In this case, it sounds like it might actually work, but not sure if that generalizes well to subclasses of `ValueTransferBridgeNative`.

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * @dev see MailboxClient's initializer for other configurables
+     */
+    constructor(address _l2Bridge, address _mailbox) HypNative(_mailbox) {
+        l2Bridge = _l2Bridge;
+        _transferOwnership(_msgSender()); // TODO: remove
+    }
+
+    function setL2Bridge(address _l2Bridge) external onlyOwner {
```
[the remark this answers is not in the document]
nambrot: I did actually miss it, but in this context I meant that you could imagine if you implement let's say Across or Stargate in this interface, you would still want to use this as a parent (because of its name), but then not sure if they would actually use this classes abstractions and variables like `l2Bridge`.

It seems like its either we should rename this to be ValueTransferBridgeRollupNative` (what a mouthful), or make this class less L2 specific?

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,131 @@
...
+ */
+abstract contract ValueTransferBridgeNative is HypNative {
+    using TypeCasts for bytes32;
+    using TypeCasts for address;
+    using TokenMessage for bytes;
+
+    uint32 constant HOOK_METADATA_GAS_LIMIT = 450_000;
+
+    // L2 bridge used to initiate the withdrawal
+    address public l2Bridge; // TODO: immutable
```
nambrot, line 27: Is there actually a reason to keep this variable in this class? It's not actually used here, right? So subclasses can just define this themselves? IIUC, the primary value is to provide the external function implementations and allow children to only implement the bridge specific quoting and funds pulling?

solidity/contracts/token/ValueTransferBridgeNative.sol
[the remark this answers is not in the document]
nambrot: So iiuc, the tricky thing here is that we are trying to change the warp route api (today: quoteGasPayment + transferRemote) to the new `IValueTransferBridge` API (quoteTransferRemote + transferRemote). Ideally this integration supports both.

Subclassing HypNative accomplishes the former, the latter requires implementing the external `quoteTransferRemote` function. However, I don't quite follow you had to remove the `IValueTransferBridge` interface? In this case, couldn't the existing `HypNative` implementation fulfill that interface?

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,131 @@
...
+ * @author Substance Labs
+ * @dev Derives from the Hyperlane native token router, but supports
+ * transfer of ERC20 token value
+ */
+abstract contract ValueTransferBridgeNative is HypNative {
+    using TypeCasts for bytes32;
+    using TypeCasts for address;
+    using TokenMessage for bytes;
+
+    uint32 constant HOOK_METADATA_GAS_LIMIT = 450_000;
```
nambrot, line 24: Should this be in this class?

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
```diff
@@ -0,0 +1,79 @@
...
+import {AbstractPostDispatchHook} from "./libs/AbstractPostDispatchHook.sol";
+import {AbstractMessageIdAuthHook} from "./libs/AbstractMessageIdAuthHook.sol";
+import {InterchainGasPaymaster} from "../hooks/igp/InterchainGasPaymaster.sol";
+
+/**
+ * @title OPL2ToL1CcipReadHook
+ * @notice Inform an OPL2ToL1ProveWithdrawalIsm that a withdrawal has been initiated on L2
+ * @dev We expect a single CCIP-read ISM executing portal.proveWithdrawal() and portal.finalizeWithdrawal() on L1 after 7 days. This is due to the fact that OP Stack expect the proof submitter and the finalizer to be the same caller.
+ */
+contract OPL2ToL1CcipReadHook is AbstractPostDispatchHook {
```
nambrot, line 19: So I haven't been able to think too much about it yet, but have you given it some thought how to share/merge logic with the existing OpStackL2toL1 hook and isms which are used for securing a message (while these ones are primarily concerned with moving value), curious on your assessment. I know @yorhodes would like to ideally see it unified, but admittedly we haven't really given much guidance there

solidity/contracts/token/ValueTransferBridgeNative.sol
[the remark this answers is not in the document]
gitmp01: I see your points, make sense, I'll let the subclasses define their bridge implementation then :+1:

solidity/contracts/token/ValueTransferBridgeNative.sol
gitmp01: This is a sort of leftover because I was previously ovverriding `transferRemote(uint32,address,uint256)`, while now I'm overriding `_transferFromSender` instead, so I think it's safe to remove this

solidity/contracts/token/ValueTransferBridgeNative.sol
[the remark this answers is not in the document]
gitmp01: that's definitely possible yeah, I thought that you may not want to touch that contract, but since it's just an interface should make no harm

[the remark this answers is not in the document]
nambrot: declare what as an abstract? `IValueTransferBridge` is just an interface no?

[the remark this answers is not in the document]
nambrot: declare what as an abstract? `IValueTransferBridge` is just an interface no?

solidity/contracts/hooks/OPL2ToL1CcipReadHook.sol
gitmp01: > What do you think?
I think it depends on the use-case, the former fits better when the devs want to secure messages depending on some runtime conditions/logic. While the latter it would be a better fit when you always want the messages delivered securely, regardless

> Did you come up with that yourself or chatted with @yorhodes about?
That's my personal brainstorming, happy to hear @yorhodes opinion too

solidity/contracts/token/ValueTransferBridgeNative.sol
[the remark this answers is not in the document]
gitmp01: In order to sort this out, I've removed `ValueTransferBridgeNative` because it useless and I've just let OPValueTransferBridge implement the `IValueTransferBridge` interfaces as it should be

solidity/contracts/interfaces/IValueTransferBridge.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity >=0.8.0;
+
+struct Quotes {
```
larryob, line 4: Nit: This should be `Quote`.

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,206 @@
...
+            (IOptimismPortal.WithdrawalTransaction)
+        );
+
+        bytes32 withdrawalHash = OPL2ToL1Withdrawal.hashWithdrawal(_tx);
+
+        if (opPortal.finalizedWithdrawals(withdrawalHash)) {
+            return;
+        }
+
+        opPortal.finalizeWithdrawalTransaction(_tx);
```
larryob, line 187: Maybe slightly clearer would be
```solidity
if (!opPortal.finalizedWithdrawals(withdrawalHash)) {
    opPortal.finalizeWithdrawalTransaction(_tx);
 }
       
```

solidity/contracts/token/ValueTransferBridgeNative.sol
```diff
@@ -0,0 +1,40 @@
...
+        uint32 _destination,
+        bytes32 _recipient,
+        uint256 _amount
+    )
+        external
+        payable
+        override(HypNative, IValueTransferBridge)
+        returns (bytes32)
+    {
+        require(msg.value >= _amount, "Native: amount exceeds msg.value");
```
larryob, line 24: Maybe add an internal function to HypNative that does this check and calls `TokenRouter._transferRemote`, then use that here?

larryob (commented):
Looks good, there's a file missing though. 

Can you update the E2E test after addressing comments?

solidity/contracts/token/ValueTransferBridgeNative.sol
gitmp01: Is this due to the similar body of `HypNative.transferRemote()`?
Because that doesn't pass the overridable `_getHookMetadata()` among the arguments, so the functions body (the one in `HypNative` and the one in `ValueTransferBridgeNative`) look the same but they are different, just bringing this up that maybe has gone unnoticed.

Also note that `_getHookMetadata()` is an internal function added to `ValueTransferBridgeNative`, shall I move that into `HypNative` when making the change?

gitmp01 (commented):
Sorted few comments, I've got questions on a couple left though

solidity/script/ValueTransferBridge.s.sol
```diff
@@ -0,0 +1,497 @@
...
+import {Script, console} from "forge-std/Script.sol";
+import {TypeCasts} from "../contracts/libs/TypeCasts.sol";
+import {Message} from "../contracts/libs/Message.sol";
+import {TokenMessage} from "../contracts/token/libs/TokenMessage.sol";
+import {OPValueTransferBridgeNative} from "../contracts/token/extensions/OPValueTransferBridgeNative.sol";
+import {OPL2ToL1CcipReadIsm} from "../contracts/isms/hook/OPL2ToL1CcipReadIsm.sol";
+import {OPL2ToL1CcipReadHook} from "../contracts/hooks/OPL2ToL1CcipReadHook.sol";
+import {IOptimismPortal} from "../contracts/interfaces/optimism/IOptimismPortal.sol";
+import {Quotes} from "../contracts/interfaces/IValueTransferBridge.sol";
+import {SecureMerkleTrie} from "./SecureMerkleTrie.sol";
```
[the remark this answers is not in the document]
gitmp01: I've removed it and other few utility functions that they were added just for debugging purposes

solidity/contracts/interfaces/optimism/IOptimismPortal.sol
```diff
@@ -20,9 +20,39 @@ interface IOptimismPortal {
         bytes data;
     }
 
+    /// @notice Struct representing the elements that are hashed together to generate an output root
```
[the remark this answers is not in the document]
larryob: Yeah, you can go ahead and remove those comments. Replace them with a link to the repo.

solidity/contracts/token/ValueTransferBridgeNative.sol
larryob: Yes, the overloading makes things a bit unclear. I'm suggesting that you add an internal function to `HypNative` that calculates the hook payment, and then reuse that function in `ValueTransferBridgeNative`.

E.g. for `HypNative`
```solidity
  function transferRemote(
        uint32 _destination,
        bytes32 _recipient,
        uint256 _amount,
        bytes calldata _hookMetadata,
        address _hook) external payable virtual override returns (bytes32 messageId) {

      return _transferRemoteWithHook(
                _destination,
                _recipient,
                _amount,
                _hookMetadata,
                _hook);
    }

    function _transferRemoteWithHook(
        uint32 _destination,
        bytes32 _recipient,
        uint256 _amount,
        bytes calldata _hookMetadata,
        address _hook
    ) internal virtual override returns (bytes32 messageId) {
        require(msg.value >= _amount, "Native: amount exceeds msg.value");
        uint256 _hookPayment = msg.value - _amount;
        return TokenRouter._transferRemote(_destination, _recipient, _amount, _hookPayment, _hookMetadata, _hook);
    }
```

And then for `ValueTransferBridgeNative`
```solidity
  function transferRemote(
        uint32 _destination,
        bytes32 _recipient,
        uint256 _amount
    )
        external
        payable
        override(HypNative, IValueTransferBridge)
        returns (bytes32)
    {
        HypNative._transferRemoteWithHook( 
                _destination,
                _recipient,
                _amount,
                _getHookMetadata(),
                address(hook)
        );
    }

```

larryob (approved):
Looks good, thanks!

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,181 @@
...
+    IMessageRecipient,
+    ISpecifiesInterchainSecurityModule
+{
+    using Message for bytes;
+    using TypeCasts for address;
+
+    // CCIP-read gateways URLs
+    string[] public urls;
+    // mailbox on L1
+    IMailbox mailbox;
```
yorhodes, line 32: this should be `immutable` to avoid incurring `SLOAD` costs
there is also no mutatot fn included

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,181 @@
...
+        } else {
+            _finalizeWithdrawal(_metadata, _message);
+        }
+
+        return true;
+    }
+
+    /**
+     * @dev no-op handle
+     */
```
yorhodes, line 97: emitting an event and receiving value isnt really "no-op"

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,181 @@
...
+        override
+        returns (IInterchainSecurityModule)
+    {
+        return IInterchainSecurityModule(address(this));
+    }
+
+    function _areWeMessageRecipient(
+        bytes calldata _message
+    ) internal view returns (bool) {
+        return _message.recipient() == address(this).addressToBytes32();
```
yorhodes, line 118: I think its slightly cheaper to use `recipientAddress()` on the message

solidity/contracts/token/HypNative.sol
```diff
@@ -89,14 +89,34 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
...
+    function _transferRemoteWithHook(
+        uint32 _destination,
+        bytes32 _recipient,
+        uint256 _amount,
+        bytes memory _hookMetadata,
+        address _hook
+    ) internal virtual returns (bytes32 messageId) {
+        require(msg.value >= _amount, "Native: amount exceeds msg.value");
+        uint256 _hookPayment = msg.value - _amount;
+        return
```
yorhodes, line 101: do we really need to create a third code location with this check?

solidity/contracts/token/extensions/OPTokenBridgeNative.sol
```diff
@@ -0,0 +1,93 @@
...
+            )
+        );
+    }
+
+    function _getHookMetadata() internal view override returns (bytes memory) {
+        return
+            StandardHookMetadata.overrideGasLimit(
+                FINALIZE_WITHDRAWAL_GAS_LIMIT
+            );
+    }
```
yorhodes, line 67: `GasRouter` is already inherited and therefore should be used for this

solidity/contracts/libs/OPL2ToL1Withdrawal.sol
```diff
@@ -0,0 +1,81 @@
...
+    function getWithdrawalMetadata(
+        address payable _l2Bridge,
+        address _l2MessagePasser,
+        uint32 _l1MinGasLimit,
+        address _remoteRouter,
+        uint256 _amountOrId,
+        bytes memory _extraData
+    ) internal view returns (bytes memory metadata) {
+        IStandardBridge l2Bridge = IStandardBridge(_l2Bridge);
+        ICrossDomainMessenger messenger = l2Bridge.MESSENGER();
```
yorhodes, line 45: im surprised to see a view call in this message encoding library

solidity/contracts/libs/OPL2ToL1Withdrawal.sol
```diff
@@ -0,0 +1,81 @@
...
+                IOptimismPortal.WithdrawalTransaction({
+                    nonce: messagePasserNonce,
+                    sender: address(messenger),
+                    target: messenger.OTHER_MESSENGER(),
+                    value: _amountOrId,
+                    gasLimit: messenger.baseGas(message, _l1MinGasLimit),
+                    data: data
+                })
+            )
+        );
```
yorhodes, line 79: I am quite confused why this needs to be committed to in the message? isnt all of this accessible offchain for the relayer

solidity/contracts/token/HypNative.sol
yorhodes: would prefer we remove this

solidity/contracts/token/TokenBridgeNative.sol
yorhodes, line 1: I think this abstraction should be removed

solidity/contracts/token/extensions/OPTokenBridgeNative.sol
yorhodes, line 1: this is also L2ToL1 and should be named as such

solidity/contracts/token/extensions/OPTokenBridgeNative.sol
```diff
@@ -0,0 +1,93 @@
...
+        // IMPORTANT: this must be placed before the l2Bridge.bridgeETHto()
+        // call in order to work (nonce will change during the withdrawal)
+        metadata = OPL2ToL1Withdrawal.getWithdrawalMetadata(
+            payable(l2Bridge),
+            address(OP_MESSAGE_PASSER),
+            OP_MIN_GAS_LIMIT_ON_L1,
+            remoteRouter,
+            _amountOrId,
+            extraData
+        );
```
yorhodes, line 85: can you explain why this needs to be passed in the message?

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /// @dev We check the withdrawal hash here in order to prevent someone
+    /// DDoS-ing the transfer execution by providing a message relative to a legit withdrawal X
+    /// and a valid proof relative to withdrawal Y
```
yorhodes, line 123: I do not understand this comment at all

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        bytes32 untrustedWithdrawalHash = OPL2ToL1Withdrawal.hashWithdrawal(
+            _tx
+        );
+        if (withdrawalHash != untrustedWithdrawalHash) {
+            revert InvalidWithdrawalHash(
+                untrustedWithdrawalHash,
+                withdrawalHash
+            );
+        }
```
yorhodes, line 153: I dont think the withdrawal hash is trusted because the `proveWithdrawalTransaction` would revert if the `_tx` was invalid?

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
yorhodes: can you elaborate?

solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
```diff
@@ -0,0 +1,178 @@
...
+    /// @dev we handle the access to the provenWithdrawal mapping
+    /// for OptimismPortal version 1 and version 2
+    function _isWithdrawalProvenAlready(
+        bytes32 _withdrawalHash
+    ) internal view returns (bool) {
+        try opPortal.provenWithdrawals(_withdrawalHash) returns (
+            IOptimismPortal.ProvenWithdrawal memory provenWithdrawal
+        ) {
+            return provenWithdrawal.timestamp > 0;
+        } catch {
```
yorhodes, line 155: i'm not a fan of this try catch for handling multiple versions
can we have an immutable version in the constructor we switch on here?
usually I would suggest having two different implementations but don't want to delay any further

yorhodes (approved):

yorhodes (approved):
lgtm

PR #5647, opened 2025-03-11
feat: add Turtle

christopherbrumm:
### Description

This PR adds the Turtle config getter and the warp ID.

### Backward compatibility

Yes

### Testing

Manual


typescript/infra/config/environments/mainnet3/warp/configGetters/getEthereumLineaTurtle.ts
```diff
@@ -0,0 +1,39 @@
...
+  RouterConfigWithoutOwner,
+  tokens,
+} from '../../../../../src/config/warp.js';
+
+const turtleOwners = {
+  ethereum: '0x2e0355922EF3a5b77d29287C808aEafB4e7f25B2',
+  linea: '0x2e0355922EF3a5b77d29287C808aEafB4e7f25B2',
+};
+
+const ISM_CONFIG = ethers.constants.AddressZero; // Default ISM
```
ltyu, line 15: I know that we have this as an example, but we no longer need to set this as address 0. In other words, this can be undefined

christopherbrumm:
Duplicate of #5649.

PR #5650, opened 2025-03-11
feat(sdk): add minimal support for `starknet` protocolType

aroralanuk:
### Description

- https://github.com/hyperlane-xyz/hyperlane-registry/pull/619 is failing because of starknet type not present in the monorepo so I moved the minimal changes from https://github.com/txfusion/hyperlane-monorepo/tree/feat/starknet-send-relayer here 

### Drive-by changes

none

### Related issues

none

### Backward compatibility

yes

### Testing

none


paulbalaji:
i think is only partially complete based on yarn build failing

typescript/utils/src/types.ts
```diff
@@ -5,6 +5,7 @@ export enum ProtocolType {
   Ethereum = 'ethereum',
   Sealevel = 'sealevel',
   Cosmos = 'cosmos',
+  Starknet = 'starknet',
```
paulbalaji, line 8: imo it'd be good if you could also add a starknet explorer family type since it looks like most starknet-chains share a similar explorer type https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2cdeff43df7aba45d014dfe8989b1ae1a92f7ecd/typescript/sdk/src/metadata/chainMetadataTypes.ts#L21

paulbalaji (commented):
pls add an SDK changeset for adding the new protocol type + Widgets changeset for integrating starknetkit

paulbalaji:
would also appreciate a review from @Xaroz on the widgets changes

typescript/widgets/tsconfig.json
```diff
@@ -7,6 +7,8 @@
     "noImplicitAny": false,
     "rootDir": "./src",
     "outDir": "./dist",
+    "moduleResolution": "bundler",
+    "module": "ES2015"
```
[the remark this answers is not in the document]
aroralanuk: This is from txfusion's branch; otherwise, imports from starknetkit weren't working. Not too familiar with bundlers so will defer to @ljankovic-txfusion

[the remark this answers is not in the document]
ljankovic-txfusion: tried without 'bundler' but imports don't work without it [Source](https://www.starknetkit.com/docs/latest/starknetkit-with-starknet-react/basics#:~:text=module%20to%20use-,Bundler,-and%20ES2015%20respectively)

typescript/sdk/src/token/Token.test.ts
```diff
@@ -200,6 +200,11 @@ const STANDARD_TO_TOKEN: Record<TokenStandard, TokenArgs | null> = {
     name: 'TIA.n',
   },
   [TokenStandard.CwHypSynthetic]: null,
+
+  //TODO: check this and manage it.
+  [TokenStandard.StarknetHypCollateral]: null,
```
tkporter, line 205: does this need updating? or no

typescript/sdk/src/token/nativeTokenMetadata.ts
```diff
@@ -22,4 +22,10 @@ export const PROTOCOL_TO_DEFAULT_NATIVE_TOKEN: Record<
     name: 'Atom',
     symbol: 'ATOM',
   },
+  [ProtocolType.Starknet]: {
+    decimals: 18,
+    denom: '0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7',
```
tkporter, line 27: is this tied to a specific network? and is it required?

typescript/widgets/src/walletIntegrations/MultiProtocolWalletModal.tsx
```diff
@@ -57,6 +57,15 @@ export function MultiProtocolWalletModal({
             Cosmos
```
tkporter, line 57: nit: can you change `an Cosmos` above to `a Cosmos` as a driveby

typescript/sdk/src/token/nativeTokenMetadata.ts
ljankovic-txfusion: Its pretty much universal on starknet chains for ETH token, in addition in chain registry denom address can be provided which has priority over this address - so this is a fallback basically

typescript/sdk/src/token/Token.test.ts
ljankovic-txfusion: Curious about how how these devnets are run in tests? For starknet we use local dockerized node, should that be added to ci/cd?

Xaroz (changes requested):
Some minor comments but overall the widgets hooks seem to be pretty in line with what Rossy created, would suggest testing everything in the UI to make sure a wallet can be connected and a transaction can be signed

typescript/sdk/src/token/Token.test.ts
aroralanuk: not right now @tkporter but good point, we should fix this with the sdk changes

paulbalaji: i think there's a larger story here around e2e with different VMs e.g. using `zksync-cli` for zksync e2e too

Xaroz (approved):

typescript/sdk/src/token/Token.test.ts
aroralanuk: made into a separate issue: https://github.com/hyperlane-xyz/issues/issues/1443

PR #5653, opened 2025-03-11
chore: separate mint limit check and add test

Xaroz:
### Description

- Separate mint limit checking from `isDestinationCollateralSufficient` into its own function: `validateDestinationRateLimit`
- Add new test for XERC20 tokens for checking `isDestinationCollateralSufficient` and `validateTransfer` 
- Add lockbox token standard to `TOKEN_COLLATERALIZED_STANDARDS` 
- Separated `MINT_LIMITED_STANDARDS` from `XERC20_STANDARDS` tokens
- Remove early return in `isDestinationCollateralSufficient` for `MINT_LIMITED_STANDARDS` tokens

### Backward compatibility

Yes
### Testing

Unit tests

### DIFF
--- a/typescript/sdk/src/warp/WarpCore.ts
+++ b/typescript/sdk/src/warp/WarpCore.ts
@@ -22,6 +22,7 @@ import { Token } from '../token/Token.js';
 import { TokenAmount } from '../token/TokenAmount.js';
 import { parseTokenConnectionId } from '../token/TokenConnection.js';
 import {
+  MINT_LIMITED_STANDARDS,
   TOKEN_COLLATERALIZED_STANDARDS,
   TOKEN_STANDARD_TO_PROVIDER_TYPE,
   TokenStandard,
@@ -771,22 +772,6 @@ export class WarpCore {
     return null;
   }
 
-  protected getMintTokenStandardAdapter(
-    token: IToken,
-  ): IHypXERC20Adapter<unknown> | null {
-    switch (token.standard) {
-      case TokenStandard.EvmHypXERC20:
-      case TokenStandard.EvmHypXERC20Lockbox:
-      case TokenStandard.EvmHypVSXERC20:
-      case TokenStandard.EvmHypVSXERC20Lockbox:
-        return token.getAdapter(
-          this.multiProvider,
-        ) as IHypXERC20Adapter<unknown>;
-      default:
-        return null;
-    }
-  }
-
   /**
    * Ensure the sender has sufficient balances for minting
    */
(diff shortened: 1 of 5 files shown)


typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -787,18 +770,64 @@ export class WarpCore {
...
+        `${destinationToken.symbol} does not have rate limit constraint, skipping`,
+      );
+      return null;
+    }
 
+    let destinationBalance: bigint;
+    const adapter = destinationToken.getAdapter(
+      this.multiProvider,
+    ) as IHypXERC20Adapter<unknown>;
+    destinationBalance = await adapter.getMintLimit();
```
nambrot, line 802: I don't think `destinationBalance`, what about `destinationMintLimit`?

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -787,18 +770,64 @@ export class WarpCore {
...
+      this.logger.debug(
+        `${destinationToken.symbol} does not have rate limit constraint, skipping`,
+      );
+      return null;
+    }
 
+    let destinationBalance: bigint;
+    const adapter = destinationToken.getAdapter(
+      this.multiProvider,
+    ) as IHypXERC20Adapter<unknown>;
```
nambrot, line 801: Is this cast necessary? Aren't there are mint standards that have a different type and thus interface?

Xaroz: afaik yes, seems that getAdapter only returns an adapter of type `ITokenAdapter`, looks like something we could improve on

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -787,18 +770,64 @@ export class WarpCore {
...
+      }
+    }
+
+    const destinationMintLimitInOriginDecimals = convertDecimalsToIntegerString(
+      destinationToken.decimals,
+      originToken.decimals,
+      destinationMintLimit.toString(),
+    );
+
+    const isSufficient = BigInt(destinationMintLimitInOriginDecimals) >= amount;
```
yorhodes, line 824: kinda weird that we convert to string and then big int

yorhodes (approved):

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -486,33 +486,10 @@ export class WarpCore {
...
-          );
-          destinationBalance = max;
-        }
-      }
-    } else {
-      destinationBalance = await adapter.getBalance(
-        destinationToken.addressOrDenom,
-      );
-    }
+    const destinationBalance: bigint = await adapter.getBalance(
```
nambrot, line 490: What if the target is a mint limited token standard, but not a token collateralized one? Then this logic doesn't apply no?

typescript/sdk/src/warp/WarpCore.ts
nambrot: But if the standard is a mint limited standard, but not xerc20, then this will fail no?

Xaroz: The mint limited token standard uses the XERC20 Token standard adapters
[here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/7d9613262239535853b6e37afc780f464fc1d36e/typescript/sdk/src/token/Token.ts#L226)

typescript/sdk/src/warp/WarpCore.ts
Xaroz: Don't understand this, the mint limited token standard implements from the base TokenAdapter so this should still work no?

nambrot (approved):

PR #5654, opened 2025-03-12
feat: cosmos module sdk support

troykessler:
### Description

This PR adds a new package `@hyperlane-xyz/cosmos-sdk` which is a standalone TypeScript SDK client for the Hyperlane Cosmos Module SDK. Furthermore, this new package is used in the main SDK to support the new Protocol Type of Cosmos Module. The main implementation can be found in the `CosmosModuleTokenAdapter.ts` Finally, the new protocol type has been added to the widgets.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

This is a WIP PR, tests will be added


paulbalaji:
This PR is really big. Could you split this up further please? Suggested:
- N PRs for the cosmos-sdk package itself (I'm unsure how much you can split this up but any split here is helpful)
- 1 PR for adding the new cosmos module support to the SDK/Widgets

paulbalaji:
> Yes, happy to split up the addition of the cosmos-sdk package itself and the rest because this PR is indeed very huge, but I don't think it makes sense to split it up further since the cosmos-sdk is tightly coupled and for the cosmos module support I had to implement the widgets already or else the build would fail there with the new registered Protocol Type. Let me know if this is fine for you or if I should try to further split it up.

yup that's probably fine then!

PR #5657, opened 2025-03-12
fix: Scraper Sealevel to support complex transactions in recipient calculation

ameten:
### Description

Currently, Scraper has very rudimentary logic to calculate a recipient of a transaction, namely, the recipient is identified as any non-native program which was invoked by the transaction. Unfortunately, when Mailbox is invoked as part of another program, transaction contains more than one non-native programs and Mailbox is featured as inner instruction, not as instruction on the top level of the transaction. It makes Scraper to report error of too many non-native transactions or incorrect recipient.

This change will allow Scraper to search for top level program which invokes Mailbox or IGP. Usually, this top level program is a Warp Route, which we prefer to have as recipient.

If Scraper cannot identify Mailbox or IGP, the first non-native program will be reported as recipient.

### Drive-by changes

- Calculation of recipient is moved to separate module.

### Related issues

- Contributes into https://github.com/hyperlane-xyz/issues/issues/1432

### Backward compatibility

Yes

### Testing

Unit tests


rust/main/chains/hyperlane-sealevel/src/provider/recipient/tests.rs
```diff
@@ -0,0 +1,39 @@
...
+    ));
+    let mailbox_address = decode_h256("EitxJuv2iBjsg2d7jVy2LDC1e2zBrx4GB5Y9h2Ko3A9Y").unwrap();
+    let provider = RecipientProvider::new(mailbox_address);
+
+    // when
+    let recipient = provider.recipient(&H512::zero(), &transaction);
+
+    // then
+    assert!(recipient.is_ok());
+    assert_eq!(recipient.unwrap(), mailbox_address);
```
tkporter, line 27: I was actually expecting `FJu4E1BDYKVg7aTWdwATZRUvytJZ8ZZ2gQuvPfMWAz9y` to be the "recipient", as this is the warp route program ID. I wouldn't say it's a giant deal but this is the behavior had before (if there weren't other non-native programs invoked) instead of considering the mailbox the recipient. The heuristic could be "search `transaction.message.instructions` for any elements that have the Mailbox in the `accounts`".

But then again, I don't think this is that critical as this address isn't used for anything too important. So if you think it's better to have us just always know that the tx recipient in the scraper db isn't super accurate on SVM, I'm open to it

rust/main/chains/hyperlane-sealevel/src/provider/recipient.rs
```diff
@@ -0,0 +1,86 @@
...
+            })
+            .map(|ii| match ii {
+                UiParsedInstruction::Parsed(iii) => &iii.program_id,
+                UiParsedInstruction::PartiallyDecoded(iii) => &iii.program_id,
+            })
+            .filter(|program_id| !NATIVE_PROGRAMS.contains(*program_id))
+            .filter(|program_id| self.program_id == **program_id)
+            .collect::<Vec<&String>>();
+
+        if programs.len() > 1 {
```
tkporter, line 69: wonder if we actually care about this case. Feels like it's sufficient to just use the first? e.g. if the Mailbox is called twice, it doesn't change the intended returned recipient

rust/main/chains/hyperlane-sealevel/src/interchain_gas.rs
```diff
@@ -41,7 +41,12 @@ impl SealevelInterchainGasPaymaster {
         conf: &ConnectionConf,
         igp_account_locator: &ContractLocator<'_>,
     ) -> ChainResult<Self> {
-        let provider = SealevelProvider::new(rpc_client, igp_account_locator.domain.clone(), conf);
+        let provider = SealevelProvider::new(
+            rpc_client,
+            igp_account_locator.domain.clone(),
+            igp_account_locator.address,
```
tkporter, line 47: I guess another observation is the IGP and Message indexing will race one another when it comes to the tx's recipient. Iiuc each task calls this https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/agents/scraper/src/db/txn.rs#L72 somewhere along the way, but it doesn't update if the tx hash already exists. So txs that include an IGP and Mailbox call (most) will be in the db with either

rust/main/chains/hyperlane-sealevel/src/provider.rs
```diff
@@ -92,41 +85,6 @@ impl SealevelProvider {
...
-                }
-            })
-            .map(|ii| match ii {
-                UiParsedInstruction::Parsed(iii) => &iii.program_id,
-                UiParsedInstruction::PartiallyDecoded(iii) => &iii.program_id,
-            })
-            .filter(|program_id| !NATIVE_PROGRAMS.contains(*program_id))
-            .collect::<Vec<&String>>();
-
-        if programs.len() > 1 {
```
tkporter, line 115: tbf I could also see it being valid for us to just have the heuristic be "pick whatever the first non native program is, regardless of whether it's the instruction responsible for the message or IGP payment"

tkporter (commented):
sorry I know the requirements around this weren't set at all, just questioning whether we want the IGP or Mailbox address to be the tx recipient. I guess there's no very satisfactory answer

rust/main/chains/hyperlane-sealevel/src/provider/recipient/tests.rs
ameten: Good point. Fixed

rust/main/chains/hyperlane-sealevel/src/provider/recipient.rs
ameten: Removed

tkporter (approved):
🤝

rust/main/chains/hyperlane-sealevel/src/provider/recipient.rs
```diff
@@ -0,0 +1,70 @@
...
+            })
+            .filter_map(|ii| match ii {
+                UiParsedInstruction::Parsed(iii) => {
+                    println!("parsed: {:?}", iii);
+                    None
+                } // only native programs are fully parsed
+                UiParsedInstruction::PartiallyDecoded(iii) => {
+                    println!("decoded: {:?}", iii);
+                    Some(iii)
+                }
```
ameten, line 47: These `println`s will be removed.

tkporter (approved):

Issue #5661, opened 2025-03-13
Separate metrics between GRPC and RPC


kamiyaa:
So that RPC and GRPC metrics are separated even if their host urls are the same

PR #5662, opened 2025-03-13
feat: add grpc/rpc separation in rpc metrics

kamiyaa:
### Description

 - add a `connection` label to client metrics
   - so we can filter based on `rpc/grpc` connection types

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5661

### Backward compatibility

Yes

### Testing

Updated E2E to filter by connection

### DIFF
--- a/rust/main/utils/run-locally/src/cosmos/termination_invariants.rs
+++ b/rust/main/utils/run-locally/src/cosmos/termination_invariants.rs
@@ -119,6 +119,14 @@ pub fn termination_invariants_met(
         return Ok(false);
     }
 
+    if !provider_metrics_invariant_met(
+        &relayer_metrics_port.to_string(),
+        messages_expected,
+        &hashmap! {"connection" => "grpc", "status" => "success"},
+    )? {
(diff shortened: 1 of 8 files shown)


ameten (approved):

rust/main/utils/run-locally/src/cosmos/termination_invariants.rs
```diff
@@ -114,7 +114,7 @@ pub fn termination_invariants_met(
     if !provider_metrics_invariant_met(
         &relayer_metrics_port.to_string(),
         messages_expected,
-        &hashmap! {"status" => "success"},
+        &hashmap! {"connection" => "rpc", "status" => "success"},
```
daniel-savu, line 117: we should add a check for grpc in the invariants as well

kamiyaa: Good call 👍

daniel-savu (approved):

PR #5664, opened 2025-03-13
fix: Make scraper to start on dedicated node

ameten:
### Description

Make scraper to start on dedicated node. It helps to avoid timeouts on Eclipse RPC with archival node of Luganodes

### Related issues

- Fixes https://github.com/hyperlane-xyz/issues/issues/1431

### Backward compatibility

Yes

### Testing

In testnet context

### DIFF
--- a/rust/main/helm/hyperlane-agent/templates/scraper-statefulset.yaml
+++ b/rust/main/helm/hyperlane-agent/templates/scraper-statefulset.yaml
@@ -71,8 +71,9 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


daniel-savu (approved):
as far as i understand, setting a taint and toleration prevents other pods from being scheduled on the same node as the scraper but does nothing to guarantee the same dedicated node is always used?

ameten:
> as far as i understand, setting a taint and toleration prevents other pods from being scheduled on the same node as the scraper but does nothing to guarantee the same dedicated node is always used?

I configured `scraper-pool` to have a single node with proper taint configured, so, Scraper will start on the same node every time.

PR #5671, opened 2025-03-14
feat: enroll infinityvm, plume

paulbalaji:
### Description

- add validators for infinityvm, plume
- update fuse validator for fusemainnet

### Drive-by changes

- igp updates
- estimate+add buffer when deploying new ICAs


.changeset/khaki-bottles-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Update default ISMs with new validators for infinityvm, plume, fuse.
```
paulbalaji, line 5:
```suggestion
Update default ISMs with new validators for infinityvm, plume, fuse. Add gas buffer when deploying Interchain Accounts.
```

.changeset/khaki-bottles-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Update default ISMs with new validators for infinityvm, plume, fuse. Add gas buffer when deploying Interchain Accounts. Add gas buffer when transferring ownership of contracts in HyperlaneDeployer.
```
paulbalaji, line 5:
```suggestion
Update default ISMs with new validators for infinityvm, plume, fuse. Add gas buffer when deploying Interchain Accounts. Add gas buffer when transferring ownership of contracts in HyperlaneDeployer. Migrate safe signing from signTransactionHash -> signTypedData.
```

typescript/infra/scripts/safes/migrate-to-typed-signatures.ts
```diff
@@ -0,0 +1,180 @@
+import { confirm } from '@inquirer/prompts';
```
Mo-Hussain, line 1: What is the context for this script?

typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -778,16 +778,19 @@ export abstract class HyperlaneDeployer<
...
+            owner,
+          );
           return this.multiProvider.handleTx(
             chain,
-            ownable.transferOwnership(
-              owner,
-              this.multiProvider.getTransactionOverrides(chain),
-            ),
+            ownable.transferOwnership(owner, {
+              gasLimit: addBufferToGasLimit(estimatedGas),
```
Mo-Hussain, line 791: nit: would it make sense to flip the order incase there is a gasLimit in the override? Or would we want to use the override instead?

Mo-Hussain (approved):
Approved with a question

typescript/infra/scripts/safes/migrate-to-typed-signatures.ts
paulbalaji: safe imposed new restrictions on their service and no longer supported txs created with `eth_sign`, instead they required `eth_signTypedData`. Rather than delete all txs and run the check-deploy again it was deemed easier to migrate the txs by reading the existing proposals and the re-proposing with `eth_signTypedData`

typescript/sdk/src/deploy/HyperlaneDeployer.ts
paulbalaji: I also wondered this, but it looks like in most places we want to use the override instead when provided. Also there are some chains like rootstock where you kind of have to manually use the 6.7M gas limit to have more reliable txs

paulbalaji:
> Approved with a question

thanks, answered your questions

PR #5680, opened 2025-03-17
feat(monitor): support per chain value at risk monitoring

Mo-Hussain:
### Description

- Add a new time series `hyperlane_warp_route_value_at_risk` which captures value at risk (VaR) per chain + warp_route_id combination. The goal is to be able to sum all of these metrics by chain to calculate a total VaR per chain.
- This data will feed this dashboard: https://abacusworks.grafana.net/d/ae1u4ofhvo4jkc/total-usd-value-at-risk?orgId=1&from=now-5m&to=now&timezone=browser&var-chain_name=$__all
- Notice how it currently has a row per chain, even when there is not data for this chain, it makes the dashboard difficult to use/not useful.
- This change will enable us to setup the dashboard so that all the available `chain_name` values on `hyperlane_warp_route_value_at_risk` are used as the chain variable on the dashboard, this will mean we will only see chains where there is data available. I will also allow us to setup alerts that check if a the # of validators for chains default ISM is higher than configured threshold if the total VaR is higher than another threshold i.e. # of validators should be higher than 4 if VaR is higher than $10 mil. I imagine we could create a separate rotation for Product for these alerts
- Current VaR promQL query

```
sum by (chain_name) (
    hyperlane_warp_route_collateral_value{related_chain_names=~".*${chain_name}.*", environment="mainnet3"} 
    or 
    hyperlane_warp_route_collateral_value{chain_name="${chain_name}"}
)
```

Proposed VaR promQL query

```
sum by (chain_name) (
    hyperlane_warp_route_value_at_risk{chain_name="${chain_name}", environment="mainnet3"} 
)
```

I did experiment with different ways to derive the chain name variables for the dashboard to only use chain names that have data without success before proposing this change. I'm also confident that we would not able to setup alert for # of validators without this change

### Drive-by changes

- Consolidate some interfaces and types to maximise reuse

### Backward compatibility

Introducing a new time series so yes

### Testing

Locally running the monitor


typescript/infra/scripts/warp-routes/monitor/metrics.ts
```diff
@@ -87,11 +120,8 @@ export function updateTokenBalanceMetrics(
       collateralTokenSymbol,
       warpCore.getTokenChains(),
     ),
-    related_chain_names: warpCore
-      .getTokenChains()
-      .filter((chainName) => chainName !== token.chainName)
-      .sort()
-      .join(','),
+    // TODO: consider deprecating this label given that we have the value at risk metric
```
Mo-Hussain, line 123: I haven't done this as I believe will be similar to a breaking change, where old metrics would be viewed as a a different time series. Not sure if there is a path to avoid this if we remove this label, also I think there is no big harm in leaving this in

paulbalaji (approved):

PR #5683, opened 2025-03-18
feat(cli): authenticated GitHub registry

mshojaei-txfusion:
### Description

Adds support for GitHub authentication token in the CLI to access private registry repositories. Implements:
- New `--authToken` CLI option
- Environment variable support via `GH_AUTH_TOKEN`
- Type definitions and context updates for auth token handling

### Related issues

resolves https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5501

### Backward compatibility

Yes - Optional feature that maintains existing functionality


.changeset/calm-hounds-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
```
paulbalaji, line 2: i'd say this is a minor than a patch, since it's not really a "fix"
```suggestion
'@hyperlane-xyz/cli': minor
```

paulbalaji (approved):

typescript/cli/src/context/context.ts
```diff
@@ -101,6 +102,8 @@ export async function getContext({
   skipConfirmation,
   disableProxy = false,
   strategyPath,
+  // @ts-ignore - To be used when getRegistry function is updated to include authToken parameter
```
paulbalaji, line 105: is there a corresponding PR for this? I wouldn't be opposed to merging this registry change first, releasing, and then updating this PR so we can do it all together

mshojaei-txfusion: yep will send it here, if you are fine with release
I would update this one after that release then

paulbalaji: yup happy to!

mshojaei-txfusion: thanks please check it here:
https://github.com/hyperlane-xyz/hyperlane-registry/pull/684

paulbalaji: registry `11.0.0` has been released https://www.npmjs.com/package/@hyperlane-xyz/registry/v/11.0.0

mshojaei-txfusion: updated, pushed with latest one

paulbalaji (approved):

mkykadir (approved):

antigremlin (approved):

PR #5686, opened 2025-03-18
chore: remove total supply from token metadata

### LINKED ISSUE #5687: remove totalSupply from token metadata

yorhodes:
### Description

Remove `totalSupply` from `TokenMetadataSchema`. The field is not needed in most cases and could cause misconfiguration issues, on the synthetic token contract `totalSupply` was used, which mints tokens to the message sender.

Instead we now apply a `initialSupply` directly to the synthetic token without the need to use `totalSupply`

### Drive-by changes

- Upgrade registry version
- Fix and remove tests that would depend on `totalSupply`

### Related issues

fixes #5687 
### Backward compatibility

Yes
### Testing

Unit tests

### DIFF
--- a/typescript/cli/package.json
+++ b/typescript/cli/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@aws-sdk/client-kms": "^3.577.0",
     "@aws-sdk/client-s3": "^3.577.0",
-    "@hyperlane-xyz/registry": "10.1.0",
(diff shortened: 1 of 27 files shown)


yorhodes (commented):
unfortunately I cant approve a PR I opened but thanks again for picking up @Xaroz

ltyu (approved):

PR #5706, opened 2025-03-19
feat: created cosmos sdk for the cosmos hyperlane module

troykessler:
### Description

This PR adds a TypeScript SDK for the Hyperlane Cosmos Module. It extends the base SDK from CosmJs and adds all the Hyperlane specific queries and transactions to it so it can be used in the `typescript/sdk` package to support the Hyperlane Cosmos Module there.

The next steps include using this new SDK as the CosmJs Provider in the Hyperlane SDK and adding the CosmosHypCollateral and CosmosHypSynthetic token adapter.

### How to review

This PR is huge, but only because there is lots of autogenerated code which can be skipped during the review process. A good overview of the cosmos-sdk package can be found in the README under `typescript/cosmos-sdk/README.md`. In summary, only the code under `src/client/*` needs to be reviewed which account for ~1500 lines of code, the other ~23,000 lines of code originating from the autogenerated `src/types` directory can be ignored for the review.

### Backward compatibility

Yes

### Testing

None - did manual e2e testing


typescript/cosmos-sdk/src/client/hyperlane/core/aminomessages.ts
```diff
@@ -0,0 +1,109 @@
+import { AminoMsg } from '@cosmjs/amino';
+
+import {
+  MsgCreateMailbox,
+  MsgProcessMessage,
+  MsgSetMailbox,
+} from '../../../types/hyperlane/core/v1/tx';
```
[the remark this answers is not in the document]
troykessler: Linting does pass according to the github actions and the `yarn lint` command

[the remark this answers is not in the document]
troykessler: Just saw that every other package in the monorepo is of type "module" (defined in the package.json). This is something I can change if required, then all import paths of local files will have the `.js` extension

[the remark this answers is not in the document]
ljankovic-txfusion: Yeah, thats a good idea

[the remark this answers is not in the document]
troykessler: just pushed the changes

typescript/cosmos-sdk/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "repository": "https://github.com/hyperlane-xyz/hyperlane-monorepo",
+  "keywords": [
+    "hyperlane",
+    "cosmos",
+    "blockchain",
+    "sdk"
+  ],
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "rimraf dist && tsc",
```
xeno097, line 23: Would it be possible to add the types building step to the `build` script and add the types to the `.gitignore`? Considering that the types folder contains autogenerated code I think it would be better to just build it when running `yarn build` instead of committing it with the non-generated files, this would also help in reducing the pr size

typescript/cosmos-sdk/src/client/hyperlane/interchain_security/aminomessages.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import {
+  MsgAnnounceValidator,
+  MsgCreateMerkleRootMultisigIsm,
+  MsgCreateMessageIdMultisigIsm,
+  MsgCreateNoopIsm,
+} from '../../../types/hyperlane/core/interchain_security/v1/tx.js';
+
+export interface AminoMsgAnnounceValidator extends AminoMsg {
+  readonly type: 'hyperlane/v1/MsgAnnounceValidator';
```
xeno097, line 11: Can the values of the `type` field for each one of the types in this file be part of an enum? I see that the same string is used in 2 places, so one unfamiliar with the code might change it in one place but not in the other. Example:

```typescript
enum MsgType {
 AnnounceValidator = "hyperlane/v1/MsgAnnounceValidator",
 MsgCreateMerkleRootMultisigIsm = "hyperlane/v1/MsgCreateMerkleRootMultisigIsm",
 ...
}

I'd this to be done also for the other hardcoded strings that are used in more than once

File: typescript/cosmos-sdk/src/client/registry/index.ts [hunk 4]
```diff
@@ -0,0 +1,60 @@
...
+  MsgRemoteTransfer,
+  MsgSetToken,
+  MsgUnrollRemoteRouter,
+} from '../../types/hyperlane/warp/v1/tx.js';
+
+export const REGISTRY: Record<string, any> = {
+  // Core transactions
+  '/hyperlane.core.v1.MsgCreateMailbox': MsgCreateMailbox,
+  '/hyperlane.core.v1.MsgSetMailbox': MsgSetMailbox,
+  '/hyperlane.core.v1.MsgProcessMessage': MsgProcessMessage,
```
xeno097, line 34: Same enum comment as in the other file

typescript/cosmos-sdk/package.json
troykessler: It would be possible, however the types are currently generated with docker and this is for a good reason since the user needs to have the proto buf cli installed for the proto generation (where the output can differ from version and architecture). So either docker must be active during the build step when building the monorepo or the user has to install the correct buf cli when we want to generate the types before the build step.

xeno097: We can add docker to the required tools for working with the monorepo and update the README so that people setting it up know that they need to have it active, wdyt?

paulbalaji:
copying my comment from slack:

how about if we split up so that it's a stack of PRs:
- new package boiler plate: package jsons, eslint rules, very basic index.ts file, very basic test. just to ensure that everything is hooked up correctly in ci and with the monorepo linter/prettier etc
- ci/local build changes to generate the types at build time instead of checking them in.
- src/client changes

another q i had is why the type generation has to live in this repo in the first place, could you not set it up to release from the cosmos module repo and then have some sort of cosmos-sdk-types package as just another dependency for the cosmos-sdk package? keeps the code cleaner in the monorepo + also more directly ties the types to the repo it is being generated from

troykessler:
> copying my comment from slack:
> 
> how about if we split up so that it's a stack of PRs:
> 
> * new package boiler plate: package jsons, eslint rules, very basic index.ts file, very basic test. just to ensure that everything is hooked up correctly in ci and with the monorepo linter/prettier etc
> * ci/local build changes to generate the types at build time instead of checking them in.
> * src/client changes
> 
> another q i had is why the type generation has to live in this repo in the first place, could you not set it up to release from the cosmos module repo and then have some sort of cosmos-sdk-types package as just another dependency for the cosmos-sdk package? keeps the code cleaner in the monorepo + also more directly ties the types to the repo it is being generated from

So regarding the discussion about moving the types generation into the ci/local build stage I agree with Lazar here: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5706#discussion_r2010464840

I also totally get that we could split up the types into a separate cosmos-sdk-types package, this is something we did at KYVE with our sdk, and would be no problem for me to do.

Regarding moving this cosmos-sdk out of the monorepo is actually totally fine for me. I'm happy to create the cosmos-sdk repo in our [bcp-innovations](https://github.com/bcp-innovations) organisation where the actual hyperlane cosmos module is located too (https://github.com/bcp-innovations/hyperlane-cosmos). Then we can publish the cosmos-sdk npm package under `@bcp-innovations/cosmos-sdk` and import it in the monorepo to implement the adapters. Let me know what you prefer

paulbalaji:
> So regarding the discussion about moving the types generation into the ci/local build stage I agree with Lazar here: [#5706 (comment)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5706#discussion_r2010464840)
> I also totally get that we could split up the types into a separate cosmos-sdk-types package, this is something we did at KYVE with our sdk, and would be no problem for me to do.

imo it'd be great to split up the types into it's own thing, that way how we generate the types becomes a separate concern that we can change over time if need be. even just to start idm going with the approach of just committing the generated types directly into the types package and working from there
 
> Regarding moving this cosmos-sdk out of the monorepo is actually totally fine for me. I'm happy to create the cosmos-sdk repo in our [bcp-innovations](https://github.com/bcp-innovations) organisation where the actual hyperlane cosmos module is located too ([bcp-innovations/hyperlane-cosmos](https://github.com/bcp-innovations/hyperlane-cosmos)). Then we can publish the cosmos-sdk npm package under `@bcp-innovations/cosmos-sdk` and import it in the monorepo to implement the adapters. Let me know what you prefer

ah no, i meant we can keep the cosmos-sdk package itself here no problem but the types package might make more sense to live in hyperlane-cosmos as that's where it's generated from

troykessler:
> imo it'd be great to split up the types into it's own thing, that way how we generate the types becomes a separate concern that we can change over time if need be. even just to start idm going with the approach of just committing the generated types directly into the types package and working from there

I would definitely prefer going forward with this approach. I would then create a new package called `cosmos-types` or `cosmos-sdk-types` (whatever you prefer) which I then import in `cosmos-sdk`. I would then create a first PR adding the types package and then creating a second PR adding the sdk package

paulbalaji:
> I would definitely prefer going forward with this approach. I would then create a new package called `cosmos-types` or `cosmos-sdk-types` (whatever you prefer) which I then import in `cosmos-sdk`. I would then create a first PR adding the types package and then creating a second PR adding the sdk package

fantastic, sounds like a plan!

troykessler:
This PR will be split up by first implementing a separate types package in the first PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5776 and using this in the sdk as a dependency which gets added in a second PR

PR #5709, opened 2025-03-19
feat(submitter): scaffolding

daniel-savu:
### Description

Creates scaffolding for the new `submitter` module, which will include the PayloadDispatcher and ChainTxAdapter

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5698

### Backward compatibility

Yes

### Testing

It passes compilation


rust/main/Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
   "ethers-prometheus",
   "hyperlane-base",
   "hyperlane-core",
-  "hyperlane-test",
+  "hyperlane-test", "submitter",
```
ameten, line 16:
```suggestion
  "hyperlane-test", 
  "submitter",
```

daniel-savu: I think it got auto-populated or formatted like this, let me see

ameten (approved):

PR #5710, opened 2025-03-20
feat(submitter): `AdaptsChain` trait definition and dependency structs

daniel-savu:
### Description

Defines the `Transaction` struct, `Payload` struct, and `AdaptsChain` trait which unblocks work on the chain adapters.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5697

### Backward compatibility

Yes

### Testing

The code compiles


rust/main/submitter/src/chain_tx_adapter/adapter.rs
```diff
@@ -0,0 +1,39 @@
+// TODO: re-enable clippy warnings
+#![allow(dead_code)]
+
+use eyre::Result;
+use hyperlane_core::U256;
+use uuid::Uuid;
```
ameten, line 6:
```suggestion
use eyre::Result;
use uuid::Uuid;

use hyperlane_core::U256;
```

rust/main/submitter/src/chain_tx_adapter/adapter.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// uses BatchManager, returns any reverted Payload IDs sent in a Transaction. Called in the Finality Stage (PayloadDispatcher)
+    async fn reverted_payloads(&self, tx: &Transaction) -> Result<Vec<Uuid>>;
+
+    // methods below are excluded from the MVP
+
+    /// Checks if a nonce gap exists in transactions submitted by the PayloadDispatcher via this adapter
+    async fn nonce_gap_exists(&self) -> bool;
+
+    /// Replaces calldata in this tx with a transfer-to-self, to use its payload(s) for filling a nonce gap
+    async fn replace_tx(&self, tx: &Transaction) -> Result<()>;
```
ameten, line 38: I wonder if it makes sense to give these method a default implementation?

rust/main/submitter/src/chain_tx_adapter/payload.rs
ameten, line 1: I wonder if definition of payload should be in `payload_dispatcher` module or in an independent module which is used by both `chain_tx_adapter` and `payload_dispatcher`?

rust/main/submitter/src/chain_tx_adapter/adapter.rs
daniel-savu: will provide a default that's `todo()` since we don't want them used yet

rust/main/submitter/src/chain_tx_adapter/payload.rs
daniel-savu: since `payload_dispatcher` will depend on `chain_tx_adapter`, I think declaring Payload and Transaction there would create a circular dependency. Two types seems too small to create a new crate IMO

rust/main/submitter/src/chain_tx_adapter/transaction.rs
```diff
@@ -0,0 +1,33 @@
+// TODO: re-enable clippy warnings
+#![allow(dead_code)]
+
+use hyperlane_core::H256;
+use uuid::Uuid;
+
+type PayloadId = Uuid;
+type SignerAddress = H256;
```
ameten, line 8: I wonder if we can share this aliases with `payload_dispatcher` as well. `PayloadId` alias will be useful there.

And I wonder if we need a similar alias for identifier of a transaction?

rust/main/submitter/src/chain_tx_adapter/transaction.rs
```diff
@@ -0,0 +1,33 @@
...
+
+/// Full details about a transaction
+pub struct Transaction {
+    uuid: Uuid,                         // unique tx identifier. Used as primary key in the db.
```
ameten, line 12: Rename field from `uuid` to `id` to match `id` field of `PayloadDetails` struct.

```suggestion
    id: Uuid,                         // unique tx identifier. Used as primary key in the db.
```

rust/main/submitter/src/chain_tx_adapter/transaction.rs
daniel-savu: hm, sure I'll rename both `id` fields and export

rust/main/submitter/src/chain_tx_adapter/transaction.rs
```diff
@@ -0,0 +1,33 @@
...
+    tx_hash: Option<H256>, // tx identifier obtained by hashing its contents. This may change when gas price is escalated
```
ameten, line 13: nit: rename field `tx_hash` into `hash` since it is inside of `Transaction` struct.

```suggestion
    hash: Option<H256>, // tx identifier obtained by hashing its contents. This may change when gas price is escalated
```

And I wonder if we want to introduce an alias for `H256` when it is used for a hash (as contrast against `SignerAddress`).

daniel-savu: will rename the field, but I think H256 stands for "hash" so I think it's descriptive enough

ameten (approved):

rust/main/submitter/src/chain_tx_adapter/transaction.rs
ameten: I don't mind using `H256`. Last time we needed to migrate from `H256` to `H512`, we changed it in many places.

By the way, we are using `H512` in `HyperlaneProvider` as transaction hash type.

daniel-savu: happy to use an alias everywhere but it feels like scope creep for this particular PR

rust/main/submitter/src/chain_tx_adapter/payload.rs
ameten: I suppose we don't need a new crate, but we can have a separate module?

I don't mind to keep it as is and see how it evolves.

rust/main/submitter/src/chain_tx_adapter/transaction.rs
daniel-savu: great point on H512, will use that

Issue #4585, opened 2024-09-27
Set `from` address when performing `eth_estimateGas` calls


paulbalaji:
## Problem

**Is your feature request related to a problem? Please describe.**

When performing an `eth_estimateGas`, some chains estimate as if the call is made from the zero address, if no `from` address is specified. On newer chains, there may not be any funds on the zero address which means that when trying to perform message deliveries we think that there are not enough funds to deliver a message, despite the relayer itself having more than enough.

**Additional context**

Example failure:
```
[...]

PR #5718, opened 2025-03-21
fix: Zksync gas estimation fix for aggregation ISMs

tkporter:
### Description

- Essentially https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5220, but only setting a random from address for zkSync chains to be conservative
- Some context here https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4585 as to why we don't want to do this for non-zksync chains 
- Opening a new PR just because I can't push to https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5220, but originally authored by Caldera


rust/main/chains/hyperlane-ethereum/src/ism/interchain_security_module.rs
```diff
@@ -94,6 +95,14 @@ where
     }
 }
 
+// The address 0x69BE704F62F7CbC1a30E35E0153D89e2b0A6Aa55 as a byte array.
+// This address was randomly generated in order to estimate gas better than
```
daniel-savu, line 99: > to estimate gas better

is it better because it doesn't mislead the reader that any pre-existing balance / usage is required on the address?

rust/main/chains/hyperlane-ethereum/src/ism/interchain_security_module.rs
```diff
@@ -116,10 +125,17 @@ where
         message: &HyperlaneMessage,
         metadata: &[u8],
     ) -> ChainResult<Option<U256>> {
-        let tx = self.contract.verify(
+        let mut tx = self.contract.verify(
             metadata.to_owned().into(),
             RawHyperlaneMessage::from(message).to_vec().into(),
         );
+        if self.domain.domain_technical_stack() == HyperlaneDomainTechnicalStack::ZkSync {
```
daniel-savu, line 132: maybe worth mentioning in a comment that since we have duplicate logic for estimating the gas limit, this special casing is also duplicated here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/cd1d0334411a40950b59efee8b42e341a219cad4/rust/main/chains/hyperlane-ethereum/src/tx.rs#L220

daniel-savu (approved):

rust/main/chains/hyperlane-ethereum/src/ism/interchain_security_module.rs
tkporter: I believe it's because of some compression gains if an address is low entropy, I'll add it to the comment

rust/main/chains/hyperlane-ethereum/src/ism/interchain_security_module.rs
tkporter: Ah good call, added

Issue #5711, opened 2025-03-20
Flaky Sealevel E2E: Invalid instruction in sealevel E2E tests


kamiyaa:
Originally, this bug was only seen in Github CI, but I ran into this bug running it locally as well.

 - We originally thought it had something to do with our Github CI caching setup.
   - This may no longer be the case because I was able to produce it locally.
 - It could be because my `/tmp` was filled up with lots of previous sealevel e2e runs

### Logs
```rust
<hyperlane-sealevel-client> accounts=[
<hyperlane-sealevel-client>     AccountMeta {
<hyperlane-sealevel-client>         pubkey: 11111111111111111111111111111111,
<hyperlane-sealevel-client>         is_signer: false,
<hyperlane-sealevel-client>         is_writable: false,
<hyperlane-sealevel-client>     },
<hyperlane-sealevel-client>     AccountMeta {
<hyperlane-sealevel-client>         pubkey: noopb9bkMVfRPU8AsbpTUg8AQkHtKwMYZiFUjNRtMmV,
<hyperlane-sealevel-client>         is_signer: false,
<hyperlane-sealevel-client>         is_writable: false,
<hyperlane-sealevel-client>     },
<hyperlane-sealevel-client>     AccountMeta {
<hyperlane-sealevel-client>         pubkey: 3hr9XFusXXL1kQYesLSfr1mzByJchCJPo5ADyZyZkPWr,
<hyperlane-sealevel-client>         is_signer: false,
<hyperlane-sealevel-client>         is_writable: false,
<hyperlane-sealevel-client>     },
<hyperlane-sealevel-client>     AccountMeta {
<hyperlane-sealevel-client>         pubkey: 692KZJaoe2KRcD6uhCQDLLXnLNA5ZLnfvdqjE4aX9iu1,
<hyperlane-sealevel-client>         is_signer: false,
<hyperlane-sealevel-client>         is_writable: false,
<hyperlane-sealevel-client>     },
<hyperlane-sealevel-client>     AccountMeta {
<hyperlane-sealevel-client>         pubkey: 4N9SspjvX6aWRJJsQ7jjS6QKGoXJbLryHDun2d4ajHc6,
<hyperlane-sealevel-client>         is_signer: false,
<hyperlane-sealevel-client>         is_writable: true,
<hyperlane-sealevel-client>     },
<hyperlane-sealevel-client>     AccountMeta {
[...]

PR #5713, opened 2025-03-20
feat(submitter): Payload and Transaction stores

daniel-savu:
### Description

Creates rocksdb stores for payloads and transactions. This required implementing encoding/decoding, for which I just used json for simplicity.

### Drive-by changes

Creates standalone modules for Payload and Transaction as @ameten suggested

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5696

### Backward compatibility

Yes

### Testing

The code builds

PR #5729, opened 2025-03-23
feat(submitter): enforce min coverage %

daniel-savu:
### Description

- Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5713
- Uses `cargo-llvm-cov` to measure test coverage of the `submitter` crate and enforces a minimum (currently configured to 1%)
- The action is added under the `test-rs` job 

### Drive-by changes

- adds a sleep in the sealevel e2e after the scraper is spun up. The assumption is that the scraper uses up a bunch of hardware resources at first which may affect the solana-test-validator and cause flakes sending warp transfers. The sleep should make the transfers less likely to fail. Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5711

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5728

### Backward compatibility

Yes - only the `submitter` crate is measured

### Testing

Yes - the CI action will fail unless there is enough coverage


.github/workflows/rust.yml
```diff
@@ -27,8 +27,60 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: full
+  MIN_SUBMITTERCOVERAGE_PERCENTAGE: 1
 
 jobs:
+  submitter-coverage:
+    runs-on: buildjet-8vcpu-ubuntu-2204
```
paulbalaji, line 34: does this need 8vcpu or can we go lower? available images: https://buildjet.com/for-github-actions/docs/runners/hardware#amd

note that adding one more job that requires a buildjet agent means we need to boost our CI capacity ASAP (or we accept slightly longer waits for PRs)

.github/workflows/rust.yml
```diff
@@ -109,3 +161,5 @@ jobs:
       - name: Check WASM for hyperlane-core
         run: cargo check --release -p hyperlane-core --features=strum,test-utils --target wasm32-unknown-unknown
         working-directory: ./rust/main
+        
+
```
paulbalaji, line 165: fix file ending

PR #5733, opened 2025-03-24
fix: ensure non-EVM testnets are enrolled on EVM chains

paulbalaji:
### Description

fix: testnet4 core config generation
- add ethereumChainOwners to testnet4 owners.ts
- renamed ownerConfig -> owner to reduce diff with mainnet3 core.ts
- deliberately didn't unify the mainnet3/testnet4 config generation scripts for now

this fixes the issue where non-EVM testnets were not being enrolled on the EVM chains

### Testing

manual


tkporter (approved):
Any idea if there's an easy unit test we can add to make sure we're for sure covered?

PR #5740, opened 2025-03-24
feat: fix url parsing when there is no protocol provided

kamiyaa:
### Description

When the rpc/grpc url provided does not have a protocol attached (ie. `https://`, `ftp://`, etc), it will show up as unknown in our metrics.
This adds a fix for when `Url` lib is not able to parse it properly because of missing protocol, we manually parse the url for our metrics.

 - I made it so, default ports also show up in the metrics
   - ie. `https://rpc.example.com/1235/1243243` -> `rpc.example.com:443`
   - because we strip the `https://` in our metrics, so its hard to tell which port its actually using
   - not sure how useful this is

### Backward compatibility

yes

### Testing

Added unit test to make sure different url formats work with metrics


rust/main/hyperlane-metric/src/utils.rs
```diff
@@ -5,13 +5,18 @@ use url::Url;
...
             write!(&mut s, ":{port}").unwrap();
         }
         Some(s)
     } else {
-        None
+        let url_str = url.to_string();
+        match url_str.split_once('/') {
+            Some((left, _)) => Some(left.into()),
+            None => Some(url_str),
+        }
```
daniel-savu, line 20: This assumes the only reason parsing could fail is a missing [scheme](https://docs.rs/url/latest/url/struct.Url.html#method.scheme). Wdyt of appending a dummy scheme if one is missing and still parsing using `host_str`, since that's more battle tested.
```suggestion
// `url` has to be `mut`
if url.scheme().is_empty() {
            let with_dummy_scheme = format!("http://{}", input);
            url = Url::parse(&with_scheme)?;
}

    if let Some(host) = url.host_str() {
        s.push_str(host);
        if let Some(port) = url.port_or_known_default() {
            write!(&mut s, ":{port}").unwrap();
        }
        Some(s)
    } else {
        None


```

kamiyaa: yeah, that works too, let me update it

daniel-savu (approved):

PR #5742, opened 2025-03-24
feat: add forma as a scraper-only chain

paulbalaji:
### Description

feat: add forma as a scraper-only chain

- add `scraperOnlyChains` to BaseScraperConfig
- use `scraperOnlyChains` when constructing the agent config
- easy configuration of other scraper-only chains in the future

> note: scraper-only chains are those who are not on default ISMs but we still want to scraper and show in explorer. Notably PI chains such as forma.

### Drive-by changes

- add checks to ensure all required `HyperlaneDeploymentArtifacts` fields are in the recovered addresses

### Related issues

https://linear.app/hyperlane-xyz/issue/BACK-106

### Testing

running rn: https://explorer.hyperlane.xyz/?origin=forma


typescript/infra/scripts/agent-utils.ts
```diff
@@ -617,15 +617,28 @@ function getInfraLandfillPath(environment: DeployEnvironment, module: Modules) {
   return path.join(getModuleDirectory(environment, module), 'addresses.json');
 }
 
-export function getAddresses(environment: DeployEnvironment, module: Modules) {
+export function getAddresses(
+  environment: DeployEnvironment,
+  module: Modules,
+  chains: ChainName[] = [],
```
tkporter, line 623: nit: may be more natural to have this be `ChainName[] | undefined` instead of treating `[]` as a no filter

typescript/infra/src/config/agent/scraper.ts
```diff
@@ -9,8 +10,7 @@ import { HelmStatefulSetValues } from '../infrastructure.js';
 import { AgentConfigHelper, RootAgentConfig } from './agent.js';
 
 export interface BaseScraperConfig {
-  // no configs at this time
-  __placeholder?: undefined;
+  scraperOnlyChains?: ChainMap<boolean>;
```
tkporter, line 13: why ChainMap? do we ever honor the boolean being true or false?

typescript/infra/src/config/agent/scraper.ts
```diff
@@ -20,6 +20,27 @@ export interface HelmScraperValues extends HelmStatefulSetValues {
...
+ */
+export function getCombinedChainsToScrape(
+  contextChainNames: string[],
+  scraperOnlyChains: ChainMap<boolean> = {},
+): string[] {
+  const chainsToScrape = new Set(contextChainNames);
+
+  // Add all scraper-only chains so we don't need to rebuild images to enable them
+  for (const chain of Object.keys(scraperOnlyChains)) {
+    chainsToScrape.add(chain);
```
tkporter, line 38: would've thought we'd honor the bool here of `scraperOnlyChains`

tkporter (approved):

typescript/infra/src/config/agent/scraper.ts
paulbalaji: resolved in 132fa02a5c0545f1020358c69da32691ce96c5d3

PR #5743, opened 2025-03-24
feat(ci): migrate to depot

paulbalaji:
### Description

feat(ci): migrate to depot
- move rust+monorepo docker builds to depot
- move big test runners from `buildjet-8vcpu-ubuntu-2204` to same-sized `depot-ubuntu-24.04-8`

### Testing

ci

### DIFF
--- a/.github/workflows/rust-docker.yml
+++ b/.github/workflows/rust-docker.yml
@@ -27,7 +27,10 @@ jobs:
         run: echo "defined=true" >> $GITHUB_OUTPUT
 
   build-and-push-to-gcr:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
 
     # uses check-env to determine if secrets.GCLOUD_SERVICE_KEY is defined
     needs: [check-env]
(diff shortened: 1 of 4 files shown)


daniel-savu (approved):
🤍

.github/workflows/rust.yml
```diff
@@ -82,28 +75,25 @@ jobs:
...
           sudo rm -rf /opt/ghc
           sudo rm -rf "/usr/local/share/boost"
           sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+      - name: Install mold linker
+        uses: rui314/setup-mold@v1
+        with:
+          mold-version: 2.0.0
+          make-default: true
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.8
```
daniel-savu, line 96: I think we used to use sccache prior to the new caching setup and it was pretty slow, concerned about increased misses. I assume we're switching to this because depot doesn't support `Swatinem/rust-cache@v2`

paulbalaji: if you look at the workflow summary this action provides the breakdown of hits/misses

paulbalaji: also moving to sccache gives us wayyyy more cache to play in GBs on depot

.github/workflows/rust-docker.yml
```diff
@@ -27,7 +27,10 @@ jobs:
         run: echo "defined=true" >> $GITHUB_OUTPUT
 
   build-and-push-to-gcr:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: ubuntu-latest
```
daniel-savu, line 30: why use the default runner instead of depot here? assuming to parallelize further?

daniel-savu (approved):

.github/workflows/rust-docker.yml
paulbalaji: previously this task would do the build on the agent itself, now we're just triggering the build on depot's own docker build runners which are separate to the depot runners we have in e2e/rust.yml

paulbalaji: see: https://depot.dev/docs/container-builds/overview

Issue #2806, opened 2023-10-13
Migrate fast token router unit tests to forge"


yorhodes:

abhishek818:
@avious00 @yorhodes Is there a bounty on this.. Can i take this up?

Hugoo:
@yorhodes could you add more context please? I see this issue is a bit old so maybe it is not as relevant anymore.

I believe the contract that needs to be tested is:

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/token/libs/FastTokenRouter.sol

But I can't find any "unit test" that need to be migrated.

1. Are there tests to migrate - if so - can you provide a link?
2. If not, can you confirm tests need to be added for this contract?
3. If none of these options - please advise.

Thanks

yorhodes:
Hey @Hugoo , thanks for reaching out
There were javascript unit tests for this a long time ago but I think we will remove from the repo for now as it is poorly maintained and not integrated to any product surface area

PR #5752, opened 2025-03-25
fix: remove `FastTokenRouter`

yorhodes:
### Description

`FastTokenRouter` is untested and not consistent with the rest of our product surface area. For example, it requires new relayer/solver behavior which is not implemented anywhere. 

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/2806

### Backward compatibility

No

### Testing

NA


nambrot (approved):

ltyu (changes requested):
still a few places where the type ref exists (e.g. `cli`)

ltyu (approved):

PR #5760, opened 2025-03-26
feat: mar26 deploy batch

paulbalaji:
### Description

feat: mar26 deploy batch
- coti, deepbrainchain, nibiru, opbnb, reactive.

### Drive-by changes

- igp updates
- daily burn updates with `--skipReview`

### Related issues

https://github.com/hyperlane-xyz/hyperlane-registry/pull/703
https://linear.app/hyperlane-xyz/issue/BACK-115/march-26th-chain-deployments

### Testing

manual
- coti → deepbrainchain ✅ (by 8x overhead to dbc)
    - self-relay ✅
- deepbrainchain → nibiru ✅
- nibiru → opbnb ✅ (by adding negligible buffer to IGP quote)
    - self-relay ✅
- opbnb → reactive ✅
- reactive → coti ✅


rust/main/config/mainnet_config.json
```diff
@@ -2368,6 +2368,7 @@
       "gasCurrencyCoinGeckoId": "ethereum",
       "gnosisSafeTransactionServiceUrl": "https://safe-transaction-zkevm.safe.global/",
       "index": {
+        "chunk": 999,
```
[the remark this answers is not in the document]
paulbalaji: spoke about it with trevor last night but i noticed in the scraper logs we were getting warnings along the lines of "block range too large, max size 1024" for a few chains - this is one such chain where i observed the log and reducing the index chunk size so it's within an accepted range for the chain

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
```diff
@@ -1,8 +1,12 @@
 {
```
Mo-Hussain, line 1: Do we need to add all these new chains here? Are we temporarily maintaining a lower than expected desired relayer balances

rust/main/config/mainnet_config.json
```diff
@@ -3873,6 +3874,7 @@
       "testRecipient": "0xbB22547D1dc681fe925f568f637Ff67aC06c20fc",
       "validatorAnnounce": "0xb4fc9B5fD57499Ef6FfF3995728a55F7A618ef86",
       "index": {
+        "chunk": 30,
```
tkporter, line 3877: this is so small 🥹 does it need to be this small?

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
tkporter, line 1: there's a process for removing these right? are any of the existing ones eligible for getting removed?

typescript/infra/config/environments/mainnet3/igp.ts
```diff
@@ -24,6 +24,10 @@ const tokenPrices: ChainMap<string> = rawTokenPrices;
 
 export function getOverheadWithOverrides(local: ChainName, remote: ChainName) {
   let overhead = getOverhead(local, remote, ethereumChainNames);
+  if (remote === 'deepbrainchain') {
```
tkporter, line 27: can you add a comment as to why and possibly link to the slack convo

Mo-Hussain, line 27: nit: a comment to explain why we need to do this

typescript/infra/scripts/print-gas-prices.ts
```diff
@@ -40,14 +40,27 @@ async function main() {
...
+          return [
+            chain,
+            await getGasPrice(
+              mpp,
+              chain,
+              gasPrices[chain as keyof typeof gasPrices],
+            ),
+          ];
+        } catch (error) {
+          console.error(`Error getting gas price for ${chain}:`, error);
```
tkporter, line 54: curious why motivated this? do we need it?

tkporter (approved):

rust/main/config/mainnet_config.json
paulbalaji: unfortunately so :(
```
(code: -32000, message: requested too many blocks from 39272316 to 39272422, maximum is set to 30, data: None)
```

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
paulbalaji: I've only added the ones here where we've been funded by the teams and the calculated amounts were blowing past what we have, `opbnb` is not included because we can self-fund there

paulbalaji: I'm not totally sure what the process is exactly, and I would've thought it would be ad-hoc inbetween deploys and not as part of a new deploy itself

typescript/infra/scripts/print-gas-prices.ts
paulbalaji: yes because as we add more chains we may not have 100% reliability of the getGasPrice function due to the variability of RPCs

therefore by try-catching we log the failures but still log the updated gasprices for chains that didnt fail

paulbalaji: notably encountered issues with infintyvm, tangle, glue recently

typescript/infra/config/environments/mainnet3/balances/desiredRelayerBalanceOverrides.json
paulbalaji: ticked to do this afterwards https://linear.app/hyperlane-xyz/issue/BACK-122/review-desired-relayed-balance-overrides-and-remove-any-where-possible

rust/main/config/mainnet_config.json
paulbalaji: for reference we set it to 50 for hyperevm https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/7dbf7e4fa7aae231815efe03ea49470b7c805a49/rust/main/config/mainnet_config.json#L8594

PR #5763, opened 2025-03-26
chore(ci): exit e2e earlier if we detect no rust changes

paulbalaji:
### Description

chore(ci): exit e2e earlier if we detect no rust changes


kamiyaa (approved):

paulbalaji:
skips correctly
| CosmWasm | EVM | Sealevel |
|----------|-----|-----------|
| image | image | image |

Issue #5746, opened 2025-03-24
Implement a dummy Building Stage, Inclusion Stage and Finality Stage which only have a (Build) queue  / (Inclusion, Finality) channel


daniel-savu:
Main piece of work is defining the Building Stage queue

daniel-savu:
Completed by other PRs

PR #5767, opened 2025-03-26
feat(submitter): Building Queue and Building Stage skeleton

daniel-savu:
### Description

Started off looking to implement the Building Queue and Inclusion / Finality channels, with only dummy implementations for the actual stages, to be able test the Queue and channels.

But after implementing the Building Queue and some of its logic, the PR got quite big due to the size of tests. I'll stop here to keep it a digestable size.

Design doc references:
- [Building Queue](https://www.notion.so/hyperlanexyz/Relayer-Submitter-Redesign-17b6d35200d68020b516c6c9bc455213?pvs=4#1a56d35200d680f68e86e9bafd1bc5ef). The full MVP scope is implemented, since reordering via an API endpoint is excluded from the initial scope.
- [Building Stage](https://www.notion.so/hyperlanexyz/Relayer-Submitter-Redesign-17b6d35200d68020b516c6c9bc455213?pvs=4#1a56d35200d680f08da4cd70dbf3f965). 3 of the 5 bullet points are already implemented: 1) being event-driven by the queue; 2) building txs from payloads via the adapter; 3) sending the tx to the next stage via a channel

### Drive-by changes

- `PayloadDispatcherState` now stores two db trait objects: the PayloadDb and the TransactionDb. They are clones of the same `Arc` though so it's lightweight
- the `build_transactions` method of `AdaptsChains` returns a `Result` now
- some test utils from `entrypoint.rs` are moved to `test_utils.rs`, to reuse them in the tests of other modules

### Related issues

Partly fixes two issues: 
    - https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5746 (the Building Queue)
    - https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5766 (part of the Building Stage logic)

### Backward compatibility

Yes - until we release anything this will always be true

### Testing

Unit tests


rust/main/submitter/src/payload/types.rs
```diff
@@ -78,4 +78,8 @@ impl FullPayload {
     pub fn set_status(&mut self, status: PayloadStatus) {
         self.status = status;
     }
+
+    pub fn details(&self) -> PayloadDetails {
+        self.details.clone()
+    }
```
ameten, line 84: I wonder if we want user of the method to decide if they want to clone the instance?

rust/main/submitter/src/payload_dispatcher/building_stage.rs
```diff
@@ -0,0 +1,134 @@
...
+    inclusion_stage_sender: mpsc::Sender<Transaction>,
+    state: PayloadDispatcherState,
+}
+
+impl BuildingStage {
+    pub async fn run(&self) -> Result<()> {
+        loop {
+            let payload = self.queue.lock().await.pop_front();
+            if let Some(payload) = payload {
+                let txs = self.state.adapter.build_transactions(vec![payload]).await?;
```
ameten, line 30: I wonder if we are going to pop several payloads at once in the future so that we can benefit from batching?

rust/main/submitter/src/payload_dispatcher/building_stage.rs
```diff
@@ -0,0 +1,134 @@
...
+            building_stage,
+            test_utils::tests::{dummy_tx, tmp_dbs, MockAdapter},
+        },
+        transaction::Transaction,
+    };
+
+    use super::{BuildingStage, BuildingStageQueue};
+
+    async fn run_building_stage_once(
+        building_stage: &super::BuildingStage,
```
ameten, line 60:
```suggestion
        building_stage: &BuildingStage,
```

ameten (approved):

rust/main/submitter/src/payload/types.rs
daniel-savu: good idea

PR #5769, opened 2025-03-27
refactor: Optimize TokenType derivation with caching and parallelization

ljankovic-txfusion:
### Description

This PR optimizes the `deriveTokenType` method in the `EvmERC20WarpRouteReader` class by implementing two key improvements:
1. Static Token Type Cache: Added a memory cache for derived `TokenType` to avoid redundant contract calls for previously queried warp route addresses
2. Parallel Contract Method Execution: Replaced `for loop` contract calls with `Promise.allSettled` for parallelization

This significantly improves performance for token type derivation is important for `warp` commands:

| Command                    | Before                                         | After                                        |
|---------------------------------------|-----------------------------------------------|----------------------------------------------|
| Warp Read  ETH/alfajores-sepolia      | 6:54 (m:ss)                          | 1:11 (m:ss)                         |
| Warp Deploy - enrollRemoteRoutes ETH/sepolia-starknetsepolia (native>synthetic) | 28:00 (m:ss) (derive token type called twice)  | 2:40 (m:ss) (derive token type called once due to caching) |

### Drive-by changes

1. Refactored code structure with cleaner organization
2. Improved code readability by removing nested try/catch blocks

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual with performance timing measurements using `console.time`

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -49,6 +49,8 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
   evmHookReader: EvmHookReader;
   evmIsmReader: EvmIsmReader;
 
+  private static tokenTypeCache: Map<Address, TokenType> = new Map();
+
   constructor(
     protected readonly multiProvider: MultiProvider,
     protected readonly chain: ChainNameOrId,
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -49,6 +49,8 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
   evmHookReader: EvmHookReader;
   evmIsmReader: EvmIsmReader;
 
+  private static tokenTypeCache: Map<Address, TokenType> = new Map();
```
ljankovic-txfusion, line 52: cache just for `TokenType` since that seems to be the biggest bottleneck in the whole class

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -94,77 +96,104 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
-      [TokenType.syntheticRebase]: {
-        factory: HypERC4626__factory,
-        method: 'collateralDomain',
-      },
-      [TokenType.synthetic]: {
-        factory: HypERC20__factory,
-        method: 'decimals',
-      },
-    };
+    const cached = EvmERC20WarpRouteReader.tokenTypeCache.get(warpRouteAddress);
```
mshojaei-txfusion, line 99: Do we need to make this dependent to `chain` as well? just in case same address on different chains
[or as in that case it would be instance with different constructor args it would be handled automatically?]

ljankovic-txfusion: rn cache is static and it's shared across all instances of the class, regardless of the constructor parameters (chain param). If the same contract address exists on different chains, they could be different token types in theory, but I dont think that adding more code to cover that edge case is worth it TBH

nambrot: Unfortunately this is actually rather common since we don't use counterfactual addresses, so we should definitely not assume that contract addresses with different contracts can be the same across chains

paulbalaji: would prefer to use the chain name or a `domainid` rather than chainids

ljankovic-txfusion: ...actually `EvmERC20WarpRouteReader` class takes `ChainNameOrId` for param, I'd figure it would be more straight forward to use `chainId` instead of `domainId`, otherwise to add intermediary step to fetch `domainId` either in constructor or method like this:
```  
async deriveTokenType(warpRouteAddress: Address): Promise<TokenType> {
    const domainId = this.multiProvider.getDomainId(this.chain);
    const cacheKey = `${domainId}-${warpRouteAddress}`;
    const cached = EvmERC20WarpRouteReader.tokenTypeCache.get(cacheKey);
    if (cached) return cached;```

Comment (paulbalaji) [reply typescript/sdk/src/token/EvmERC20WarpRouteReader.ts:99]: ah then we can just use `this.chain` directly as it should be the name or domainId already iirc, we didn't rename it when doing [https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798) to avoid creating a big diff

Comment (ljankovic-txfusion) [reply typescript/sdk/src/token/EvmERC20WarpRouteReader.ts]: updated

**Review by paulbalaji — APPROVED** [review]

**Review by ltyu — APPROVED** [review]

**Comment by paulbalaji** [comment]
yarn-test failing?

File: typescript/sdk/src/token/EvmERC20WarpRouteReader.ts [hunk 3]
```diff
@@ -179,103 +181,111 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+                wrappedToken,
+                this.provider,
+              );
+              await xerc20['mintingCurrentLimitOf(address)'](warpRouteAddress);
+              return TokenType.XERC20;
+            } catch {
+              return TokenType.collateral;
+            }
+          }),
+        contracts[TokenType.collateralFiat].callStatic
```
ljankovic-txfusion, line 250: @paulbalaji for some reason test is failing for this case

PR #5776, opened 2025-03-27
feat: added cosmos-types package

troykessler:
### Description
This PR adds a TypeScript types and encoding/decoding package for the Hyperlane Cosmos Module. It autogenerates these types and converters from the proto files from the original [Cosmos SDK Hyperlane Module](https://github.com/bcp-innovations/hyperlane-cosmos) implementation. This PR is the result of the discussion from the previous PR https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5706 implementing the cosmos-sdk. It was decided to move the types in this cosmos-types package to make it cleaner.

The next steps include creating an SDK which uses those types and converters to support queries and transaction submission. Furthermore, using this new SDK as the CosmJs Provider in the Hyperlane SDK and adding the CosmosHypCollateral and CosmosHypSynthetic token adapter will be the final step in order to achieve the first milestone of supporting Cosmos in the Nexus UI.

### How to review
This PR is huge, but only because there is lots of autogenerated code which can be skipped during the review process. The only code which was written manually can be found under the `proto` directory and in the `src/index.ts` file.

### Backward compatibility
Yes

### Testing
None - did manual e2e testing


typescript/cosmos-types/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+After this command has finished the newly generated types can be found under `src/types`.
+
+### Building the project
+
+You can build the project with `yarn build`, the build output can be found under `dist`.
+
+## License
+
+Apache 2.0
```
paulbalaji, line 37: can omit this as we have the license at the repo root https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/LICENSE.md

typescript/cosmos-types/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "types"
+  ],
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "rimraf dist && tsc",
+    "format": "prettier --write .",
+    "lint": "eslint -c ./eslint.config.mjs",
+    "prettier": "prettier --write ./src",
+    "clean": "rm -rf ./dist ./cache",
+    "test": "exit 0"
```
paulbalaji, line 29: is there anything we can test? maybe add a test to check that if we regenerate types in CI that we don't get a diff?

typescript/cosmos-types/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@eslint/js": "^9.15.0",
+    "@typescript-eslint/eslint-plugin": "^8.1.6",
+    "@typescript-eslint/parser": "^8.1.6",
+    "eslint": "^9.15.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-import-resolver-typescript": "^3.6.3",
+    "eslint-plugin-import": "^2.31.0",
+    "globals": "^15.14.0",
+    "prettier": "^2.8.8",
+    "rimraf": "^6.0.1",
```
paulbalaji, line 41: what's this for?

typescript/cosmos-types/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "ts-node": "^10.8.0",
+    "tslib": "^2.8.1",
+    "typescript": "5.3.3",
+    "typescript-eslint": "^8.23.0"
```
paulbalaji, line 45: do we need this if we have the other two?

typescript/cosmos-types/proto/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM bufbuild/buf:latest as BUILDER
```
paulbalaji, line 1: probably want to pin to a specific version

typescript/cosmos-types/proto/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    fi
+
+# Create user if it doesn't exist
+RUN if ! getent passwd $USER_ID; then \
+    adduser -D -u $USER_ID -G $(getent group $USER_GID | cut -d: -f1) myuser; \
+    fi
+
+WORKDIR /app
+
+RUN apk add --no-cache git openssh-client
```
paulbalaji, line 24: swap so we can do the RUN commands together for better layer caching
```suggestion
RUN apk add --no-cache git openssh-client

WORKDIR /app
```

typescript/cosmos-types/proto/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+COPY buf.gen.yaml ./
+
+# Create a mount point for the types directory
+VOLUME /app/types
```
paulbalaji, line 29: do we not need to actually set the user to the one we setup earlier in the file?
```suggestion
VOLUME /app/types

USER $USER_ID:$USER_GID
```

typescript/cosmos-types/proto/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Override default entrypoint
+ENTRYPOINT []
+
+# Use shell form of CMD
+CMD ["/bin/sh", "-c", "git clone https://github.com/bcp-innovations/hyperlane-cosmos.git && \
+    cp -rf hyperlane-cosmos/proto/* ./ && \
+    rm -rf hyperlane-cosmos && \
+    buf generate && \
+    rm -rf ./types/amino*"]
```
paulbalaji, line 39: -eo pipefail & --depth 1
```suggestion
CMD ["/bin/sh", "-c", "\
    set -eo pipefail && \
    git clone --depth 1 https://github.com/bcp-innovations/hyperlane-cosmos.git && \
    cp -rf hyperlane-cosmos/proto/* ./ && \
    rm -rf hyperlane-cosmos && \
    buf generate && \
    rm -rf ./types/amino*"]

```

typescript/cosmos-types/src/index.ts
paulbalaji, line 1: any chance we can specify the specific things we want to export? better for tree shaking

typescript/cosmos-types/package.json
paulbalaji: even if we don't test anything, would be good to add an echo "no tests in cosmos-sdk-types" or something to log that we do try this

troykessler: unfortunately there is nothing to test since all is autogenerated and just exported, will echo `no tests in cosmos-types`

typescript/cosmos-types/package.json
troykessler: removed unused dependencies

typescript/cosmos-types/package.json
troykessler: unused, removed dependency

typescript/cosmos-types/proto/Dockerfile
troykessler: Adding this fails with `fatal: could not create work tree dir 'hyperlane-cosmos': Permission denied`. Leaving users as is

typescript/cosmos-types/src/index.ts
troykessler: I think we have to export everything since almost everything is used in our sdk anyway and at the end of the day the user of this package decides what he wants to use, so because of that I would export everything

paulbalaji:
need to run `yarn` and push the updated lockfile because of the package.json change

paulbalaji (commented):
overall lgtm

typescript/cosmos-types/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "keywords": [
+    "hyperlane",
+    "cosmos",
+    "blockchain",
+    "sdk",
+    "types"
+  ],
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "rm -rf ./dist && tsc",
```
ltyu, line 24: can we run the prettier after this? when i run `build`, it gives me a bunch of diffs that resolve after using `yarn prettier --fix`

troykessler: what are those diffs? I don't get any after running `yarn build` inside `typescript/cosmos-types`

ltyu (approved):

typescript/cosmos-types/package.json
troykessler: running prettier after `build` now

PR #5779, opened 2025-03-28
fix: token adapter initialization logic in the sdk to derive spl token data correctly

xeno097:
### Description

Fixes a bug in the`SealevelHypCollateralAdapter` initialization logic where the `isSpl2022` variable was set to false by default, preventing correct data derivation when interacting with SPL_2022 tokens on chain (balance equal to 0 for a given address or account not existing error) 

### Backward compatibility

- NO

### Testing

- Manual testing with local UI using the patched package


typescript/sdk/src/token/Token.ts
```diff
@@ -172,10 +176,10 @@ export class Token implements IToken {
    * @throws If multiProvider does not contain this token's chain.
    * @throws If token is an NFT (TODO NFT Adapter support)
    */
-  getHypAdapter(
+  async getHypAdapter(
     multiProvider: MultiProtocolProvider<{ mailbox?: Address }>,
     destination?: ChainName,
-  ): IHypTokenAdapter<unknown> {
+  ): Promise<IHypTokenAdapter<unknown>> {
```
tkporter, line 182: this feels like a pretty major change so would like for someone closer than me to the SDK/CLI/registry to have an opinion here. E.g. I imagine this will require lots of changes on the warp UI, some of which may not be easily possible if this function is called by non-async functions

Other ideas that come to mind:
1. Leave the constructor as non-async, and have a state variable internal to the SealevelHypCollateralAdapter that is initially uninitialized. Then we make whatever functions we have that require the knowledge of the SPL token program ID to be async, and it has logic to have a one-time initialization of the SPL token type or to use the already initialized value
2. Add the SPL token type to the config schema, so we don't need to interact with the chain at all

typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts
yorhodes, line 1: I really dont like these casts :(

typescript/sdk/src/token/Token.ts
yorhodes: I think the `WarpCore` api mostly encapsulates this for other consumers like the warp UI but curious to get @Xaroz 's thoughts

yorhodes: agree though that there are other less breaking approaches

typescript/sdk/src/token/Token.ts
```diff
@@ -245,6 +249,29 @@ export class Token implements IToken {
...
+        new PublicKey(collateralAddressOrDenom),
+      );
+
+      assert(
+        accountInfo,
+        'collateralAddressOrDenom must be an existing account for Sealevel hyp collateral tokens',
+      );
+
+      assert(
+        accountInfo.owner.equals(TOKEN_2022_PROGRAM_ID) ||
```
yorhodes, line 270: how is this different from line 283?

typescript/sdk/src/token/Token.ts
yorhodes: it sounds like @xeno097 tested this with a local UI build

yorhodes: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/adapters/ITokenAdapter.ts#L32-L57
the adapter API is already async everywhere so deriving internally seems like the move

xeno097: Oky I'm going to change the approach to derive internally and retest with a local ui

yorhodes (approved):
lgtm pending UI tests

xeno097:
> lgtm pending UI tests

@yorhodes Retested on the ui with the latest changes and it worked fine, merging

http://localhost:3000/?origin=solanamainnet&destination=soon&token=9TxKLb6RBo4hEfFNN6up6obvebnEUcm2gkBXrzQFq8s8
with the current SDK version the balance would have been reported as 0

PR #5781, opened 2025-03-28
feat: cache metadata for retries

kamiyaa:
### Description

 - store metadata in `PendingMessage`
 - check if metadata already exists. if it does, don't try to recreate it again

cached metadata is considered stale if:
 - more than 1 day old
 - fails gas estimation
 - exceeds max gas
 - fails on-chain

### Related issues

 - fixes https://linear.app/hyperlane-xyz/issue/BACK-91/cache-metadata-for-messages-that-had-a-successful-simulation-but-are

### Backward compatibility

yes

### Testing

added e2e check to check cache was reused


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -343,6 +288,7 @@ impl PendingOperation for PendingMessage {
                     .clarify_reason(ReprepareReason::ErrorEstimatingGas)
                     .await
                     .unwrap_or(ReprepareReason::ErrorEstimatingGas);
+                self.clear_metadata();
```
tkporter, line 291: iiuc this means that if we use a cached metadata that we should no longer be using and we now fail when estimating gas, we will only actually rebuild metadata in the next retry?

Because time between retries can get pretty long, I think I'd rather have us not need to wait for a second retry just to build the correct metadata, and for us to have the logic of "if we are using cached metadata, try estimating costs, and if that fails then just rebuild metadata and try estimating the costs again" explicitly in here

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -250,81 +253,23 @@ impl PendingOperation for PendingMessage {
...
-            Ok(metadata) => {
-                let metadata_bytes = metadata.to_vec();
-                self.metadata = Some(metadata_bytes.clone());
-                metadata_bytes
+        let metadata_bytes = match self.metadata.as_ref() {
+            // if metadata is already built and is not too stale
+            Some((metadata, built_time))
+                if Instant::now().duration_since(*built_time).as_secs() < STALE_METADATA_SECS =>
+            {
+                tracing::debug!("Reusing cached metadata");
```
tkporter, line 261: would suggest either having this as a const or adding a comment that e2e relies on this

rust/main/agents/relayer/src/msg/pending_message.rs
kamiyaa: Ah that makes sense, will update it

rust/main/agents/relayer/src/msg/pending_message.rs
kamiyaa: Good point

daniel-savu (approved):

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -45,6 +45,9 @@ pub const RETRIEVED_MESSAGE_LOG: &str = "Message status retrieved from db";
 pub const ISM_MAX_DEPTH: u32 = 13;
 pub const ISM_MAX_COUNT: u32 = 100;
 
+/// 1 day
+pub const STALE_METADATA_SECS: u64 = 60 * 60 * 24;
```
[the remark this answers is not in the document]
kamiyaa: removing

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -250,81 +253,23 @@ impl PendingOperation for PendingMessage {
...
-                    }
-                    MetadataBuildError::AggregationThresholdNotMet(threshold) => {
-                        warn!(threshold, "Aggregation threshold not met");
-                        return self
-                            .on_reprepare(Some(err), ReprepareReason::CouldNotFetchMetadata);
-                    }
+            _ => match self.build_metadata().await {
+                Ok(metadata) => {
+                    self.metadata = Some((metadata.to_vec(), Instant::now()));
+                    metadata.to_vec()
```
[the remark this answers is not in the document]
kamiyaa: This doesn't work because `prepare()` returns `PendingOperationResult`, but `build_metadata()` is `Result<Metadata, PendingOperationResult>`

rust/main/utils/run-locally/src/main.rs
```diff
@@ -572,6 +576,33 @@ fn relayer_restart_invariants_met() -> eyre::Result<bool> {
     Ok(true)
 }
 
+/// Check relayer reused already built metadata
+fn relayer_cached_metadata_invariant_met() -> eyre::Result<bool> {
```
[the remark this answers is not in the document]
kamiyaa: no, good point :+1:

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -250,99 +252,64 @@ impl PendingOperation for PendingMessage {
...
-                    MetadataBuildError::AggregationThresholdNotMet(threshold) => {
-                        warn!(threshold, "Aggregation threshold not met");
-                        return self
-                            .on_reprepare(Some(err), ReprepareReason::CouldNotFetchMetadata);
-                    }
+        // if gas estimate fails, we want to retry building metadata,
+        // and try gas estimation again
+        if tx_cost_estimate.is_err() {
+            self.clear_metadata();
+            match self.build_metadata().await {
```
tkporter, line 283: can we make it so we only build metadata twice if the metadata wasn't built this time? i.e. only if it was previously cached

tkporter (approved):
tyvm

rust/main/agents/relayer/src/msg/pending_message.rs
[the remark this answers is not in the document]
tkporter: assuming this is good to resolve

PR #5785, opened 2025-03-28
feat(submitter): inclusion stage implementation

daniel-savu:
### Description

- Finishes the Building Stage and Inclusion Stage implementations and unit tests them to high coverage
- Adds unlimited retries to all fallible network calls. The reasoning is that if the network is problematic, it doesn't make sense to move on to a different payload / tx since those will have the same issue
- Changes return type of `AdaptsChain::build_transactions` to `TxBuildingResult`, to know which payloads to drop if building a tx for that was unsuccessful
- Refactors testing utils to reuse across all stages

### Drive-by changes

- Adds enum error types across the `submitter` crate. The point is to eventually use these to know exactly when it's worth escalating the price or drop the tx
- Changes the sealevel `get_estimated_costs_for_instruction` to return an `Option`, where `None` represents failed simulation and `Some` represents successful simulation. This helps inform whether to drop payloads / txs
    - Similarly changes the `estimate` and `simulate` methods on `AdaptsChain`

### Related issues
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5766
- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5789

### Backward compatibility

Yes

### Testing

Unit tests. See the cargo-llvm-cov report below (the columns are `Function Coverage | Line Coverage | Region Coverage`)

Screenshot 2025-03-31 at 02 49 48


rust/main/submitter/src/chain_tx_adapter/error.rs
```diff
@@ -0,0 +1,29 @@
+#![allow(dead_code)]
+
+use hyperlane_base::db::DbError;
+
+use crate::transaction::Transaction;
+
+#[derive(Debug, thiserror::Error)]
+pub enum DispatcherError {
```
ameten, line 8: I would call it `AdapterError` since they will be reported from adapter, as far as I understand it.

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
```diff
@@ -1 +1,406 @@
...
+            );
+        }
+    }
+
+    async fn receive_txs(
+        mut building_stage_receiver: mpsc::Receiver<Transaction>,
+        pool: InclusionStagePool,
+    ) -> Result<()> {
+        loop {
+            let tx = building_stage_receiver.recv().await.unwrap();
```
ameten, line 68: If `recv` returns `None`, the channel is closed. I suppose we can raise a critical error if it happens so that we don't crash the whole relayer.

ameten (approved):

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
```diff
@@ -1 +1,406 @@
...
+                    let mut tx_clone_inner = tx_clone.clone();
+                    state.adapter.submit(&mut tx_clone_inner).await
+                }
+            },
+            "Submitting transaction",
+        )
+        .await;
+
+        // update tx submission attempts
+        tx.submission_attempts += 1;
```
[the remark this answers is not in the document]
daniel-savu: agreed IRL that `PayloadDispatcher` will update `submission_attempts` since this field is vm-agnostic

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
```diff
@@ -1 +1,406 @@
...
+        pool: &InclusionStagePool,
+    ) -> Result<()> {
+        info!(?tx, "Dropping tx");
+        let new_tx_status = TransactionStatus::Dropped(reason);
+        Self::update_tx_status(state, tx, new_tx_status.clone()).await?;
+        // drop the payloads as well
+        state
+            .update_status_for_payloads(
+                &tx.payload_details,
+                PayloadStatus::InTransaction(new_tx_status),
```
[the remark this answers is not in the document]
daniel-savu: discussed IRL that we can only reach `PayloadStatus::Dropped(DropReason)` if simulation fails in the Building Stage. Will have another look at statuses to see if I can simplify

rust/main/submitter/src/chain_tx_adapter/error.rs
daniel-savu: agreed that of the two options below to first do 1), and aim to switch to 2) later
1) have a single error type across the entire Submitter. We rename the current enum to SubmitterError and move it to submitter/src to reduce confusion. The drawback here is that the errors for various components are mixed into one type
2) have component specific error types (AdapterError, DispatcherError), but also a submitter-wide SubmitterError which is an enum that combines AdapterError, DispatcherError. The drawback here is that some errors might need two conversion steps. For instance, ChainCommunicationError -> AdapterError(ChainCommunicationError) -> SubmitterError(AdapterError(ChainCommunicationError)). This would require us to use .map_err in many places, which is verbose and I would like to avoid

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
```diff
@@ -1 +1,406 @@
...
+                pool.lock().await.remove(&tx_id);
+                return Ok(());
+            }
+            TransactionStatus::Finalized | TransactionStatus::Dropped(_) => {
+                error!(
+                    ?tx,
+                    ?tx_status,
+                    "Transaction has invalid status for inclusion stage"
+                );
+            }
```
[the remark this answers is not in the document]
daniel-savu: changed to handle the `Finalized` status the same as `Included` (the tx is sent to the finality stage to reuse its logic, even if this will cause one more call to fetch the status). I think it's highly unlikely that `Dropped` will be reported in the inclusion step, because `Dropped` assume the tx was first `Included`. @ameten also pointed out that identifying dropped txs requires the adapter to be stateful (read the tx db) in order to know which txs used to be included

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
daniel-savu: ah my bad, this `unwrap` wasn't supposed to make its way here

PR #5787, opened 2025-03-28
feat: short circuit when preparing a message that can certainly not meet the gas payment enforcement policy

tkporter:
### Description

- Prior to building metadata or simulating the message, first perform a check if we would pass the gas payment enforcement policy even if there was zero cost for the transaction
- This means that:
  -  if the relayer is configured to strictly require at least some kind of payment, we avoid spending resources preparing (building metadata, simulating, etc) on messages that have not paid our IGP
  - we continue to fully prepare messages if they meet a policy that does not require a payment

Example in action:
When we had large Ethereum prep queues to chug through due to a bunch of messages without IGP payments, we were severely bottlenecked on the prep queue, only popping every ~2-3 mins:
Screen Shot 2025-03-28 at 9 28 21 PM

After the change, now popping much more frequently:
Screen Shot 2025-03-28 at 9 44 39 PM

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/BACK-78/dont-build-metadata-for-a-message-that-certainly-has-not-met-the-gas


rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter/tests.rs
```diff
@@ -129,7 +129,7 @@ async fn test_estimate_gas_limit() {
     let result = adapter.estimate_gas_limit(&payload).await;
 
     // then
-    assert!(matches!(result, Ok(_)));
+    assert!(result.is_ok());
```
tkporter, line 132: just clippy's doing

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -649,6 +643,71 @@ impl PendingMessage {
...
+        // which may violate the gas payment enforcement policies depending on
+        // the configuration, but also allows us to be tolerant of the configuration
+        // allowing no payment at all.
+        let zero_cost = TxCostEstimate {
+            gas_limit: U256::zero(),
+            gas_price: FixedPointNumber::zero(),
+            l2_gas_limit: None,
+        };
+
+        self.meets_gas_payment_requirement(&zero_cost)
```
daniel-savu, line 669: so rather than using `GasPaymentPolicyMinimum`, this reuses the configured IGP policy

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -649,6 +643,71 @@ impl PendingMessage {
...
+            .map(|_| ())
+    }
+
+    /// Returns the gas limit if the message meets the gas payment requirement,
+    /// otherwise returns an Err(PendingOperationResult), with the PendingOperationResult intended
+    /// to be propagated up by the prepare fn.
+    async fn meets_gas_payment_requirement(
+        &mut self,
+        tx_cost_estimate: &TxCostEstimate,
+    ) -> Result<U256, PendingOperationResult> {
```
daniel-savu, line 681: IMO this overloads the meaning of `Result`, since the `Err` variant isn't really an error, which can be confusing. Would be cleaner to define an enum
```rust
enum IgpPolicyOutcome {
    MeetsPolicy(U256),
    PolicyNotMet(PendingOperationResult)
}
```
```suggestion
    ) -> IgpPolicyOutcome {
```

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -250,6 +250,14 @@ impl PendingOperation for PendingMessage {
             return PendingOperationResult::Drop;
         }
 
+        // Perform a preflight check to see if we can short circuit the gas
+        // payment requirement check early without performing expensive
+        // operations like metadata building or gas estimation.
+        if let Err(op_result) = self.meets_gas_payment_requirement_preflight_check().await {
```
daniel-savu, line 256: noting this will slightly increase the duration of time spent in the `prepare` step

daniel-savu (approved):

rust/main/agents/relayer/src/msg/pending_message.rs
tkporter: exactly, to still allow the relayer to relay messages without requiring any payments if it's configured to do so

rust/main/agents/relayer/src/msg/pending_message.rs
tkporter: yeah you're def right, will do that

rust/main/agents/relayer/src/msg/pending_message.rs
tkporter: yeah I think this is worth it - I'll actually add some screenshots to the description to give some more background on the impact of this

PR #5788, opened 2025-03-29
feat: add metadata build metrics

kamiyaa:
### Description

adds metrics for metadata building

```rust
hashmap! {
    "app_context" => ...
    "origin" => ...
    "destination" => ...
    "status" => ...
};
```

### Drive-by
 - add `rust/sealevel/environments/local-e2e/igp/` to gitignore

### Related issues

 - fixes https://linear.app/hyperlane-xyz/issue/BACK-120/metrics-for-metadata-build-times

### Backward compatibility

yes

### Testing

e2e test to make sure metrics are available


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -293,10 +294,29 @@ impl PendingOperation for PendingMessage {
 
         let params = MessageMetadataBuildParams::default();
 
-        let metadata_bytes = match message_metadata_builder
+        let build_metadata_start = Instant::now();
```
tkporter, line 297: nit: might be nice to put this metric stuff in helper function or something, just so that the `prepare` logic can stay lean. Either a function in PendingMessage or maybe even internal to MessageMetadataBuilder, happy either way

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -293,10 +294,29 @@ impl PendingOperation for PendingMessage {
...
+        let metadata_res = message_metadata_builder
             .build(ism_address, &self.message, params)
-            .await
-        {
+            .await;
+        let build_metadata_end = Instant::now();
+
+        let origin_chain = KnownHyperlaneDomain::try_from(self.message.origin).ok();
```
tkporter, line 303: we should avoid using `KnownHyperlaneDomain::try_from` - the reason is that KnownHyperlaneDomain is only aware of a small subset of chains that we've explicitly hardcoded into the enum https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/hyperlane-core/src/chain.rs#L130. So if we run the relayer with a chain that is not hardcoded into the binary (which we actually do today all the time) then this will give an Err and the ok() will give us None. Feel free to also document this aversion to `KnownHyperlaneDomain` unless dealing with special casing specific chains, as this is understandably confusing

I actually think the better move would be to leverage the existing [`MessageSubmissionMetrics`](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f3618fe26b93a41865d5d92ecdd9e71c4b9b63e8/rust/main/agents/relayer/src/msg/pending_message.rs#L863). With these we already have the origin and destination names in the constructor, and the tracking of the metadata building is conceptually a message submission metric more than it is a client metric imo. If you move the gauge/counter into there I think everything will fit more nicely and we won't have this case of having origin or dest chains be None

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -293,10 +294,29 @@ impl PendingOperation for PendingMessage {
...
+        let dest_chain = KnownHyperlaneDomain::try_from(self.message.destination).ok();
+
+        let metrics_params = MetadataBuildMetric {
+            app_context: self.app_context.clone(),
+            origin: origin_chain,
+            destination: dest_chain,
+            success: metadata_res.is_ok(),
+            duration: build_metadata_end - build_metadata_start,
```
tkporter, line 311: should we use `.saturating_duration_since` instead?

rust/main/agents/relayer/src/msg/pending_message.rs
kamiyaa: Ah nice, I'll move it to MessageSubmissionMetrics

tkporter (approved):
🫡

PR #5792, opened 2025-03-31
feat(validator): public rpc flag & rpc info in `metadata_latest.json`

### LINKED ISSUE #5590: Validators should only use the default public RPCs if they explicitly opt into using it
## Problem

- We generally want to discourage validators from not using custom RPC URLs. By default, if no custom urls are provided, the default ones in the agent config jsons are used.
- The aim is to have:
  - Validators use their own private URLs
  - Validators not all use the same URL, so if that one URL goes down they all go down

## Solution

- Introduce some config like "allow-unsafe-default-rpcs" or "allow-unsafe-public-rpcs" where only if it's explicitly specified does the validator use default RPCs
- Imo for the relayer / scraper we don't care as much about safety so we wouldn't need or want this config required for those
- Would want to document this
- We should also consider that some operators prefer to not use the customRpcUrls path, and instead just use their own agent config JSONs. These modified JSONs may include private RPC URLs, and we shouldn't necessarily require them to have this config "allow-unsafe-default-rpcs" imo. Maybe we should change things such that the default RPCs we ship with the JSONs are: `{ "http": "...", public: true }` and that's how we determine if a URL is by default an unsafe / public one?

## Nice to Have

### LINKED ISSUE #5591: Using a validator's metadata_latest.json, it should be possible to know if they are just using default public URLs
## Problem

- Want to know if multiple validators are using the same RPC URLs, or if they are using the default public ones
- We don't want the RPCs themselves to be publicly exposed

## Solution

- Add to the `metadata_latest.json` some info about the RPC URLs that a validator is using
- Leading idea to add a hash / commitment of the RPC URLs used
  - This has the slight downside of us now needing some way to understand if this commitment matches the default/public URLs we ship with the image, which themselves may change over time (e.g. if the registry URLs change)
  - I wonder if there's a way to just have a bool of "uses public default URLs" or something, but this is also sorta hard because sometimes validator operators choose not to use the customRpcUrls config and instead use their own YAMLs. So it's not enough to just look at the presence of the `customRpcUrls` to decide if someone is using public default URLs or not

## Nice to Have

yjamin:
### Description

This PR introduces a new flag `--allowPublicRpcs` that has to be explicitly set to `true` if one wants to run a validator on public RPCs.

The Validator detects whether or not a RPC is public based on an extra property per RPC Url. If the property is not set, the RPC is assumed to be private. This way backwards compatibility is ensured and we can incrementally rollout this feature by explicitly setting `public: true` for each RPC in the `hyperlane-registry`

Also includes: #5793 

### Related issues

Closes #5590
Closes #5591

### Backward compatibility
Full, until the Registry is updated to account for the new `public` attribute

### Testing

Added unit tests to check for correct config parsing.


rust/main/agents/validator/src/settings.rs
```diff
@@ -230,3 +309,72 @@ fn parse_checkpoint_syncer(syncer: ValueParser) -> ConfigResult<CheckpointSyncer
...
+            })
+            .collect::<Vec<_>>();
+        let rpcs = serde_json::json!({
+            "rpcurls": rpcs
+        });
+
+        let mut err = ConfigParsingError::default();
+        let value_parser = ValueParser::new(ConfigPath::default(), &rpcs);
+        let parsed = get_rpc_urls(&value_parser, "rpcUrls", &mut err); // why does it convert to lowercase?
+
```
yjamin, line 346: Why does the `get_opt_key` method convert the key into lowercase?

rust/main/agents/validator/src/settings.rs
```diff
@@ -145,10 +174,60 @@ impl FromRawConf<RawValidatorSettings> for ValidatorSettings {
...
+                    Some(url) => Some(RpcConfig {
+                        url: url.to_owned(),
+                        public: false,
+                    }),
+                    _ => None,
+                }
+            })
+            .collect_vec()
+        })
+        .unwrap_or_default()
```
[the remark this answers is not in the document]
yjamin: Sadly there is no `Default` for `ValueParse` so we can't do that. However I could remove that match statement with a map like this: 

```rust
    chain
        .chain(err)
        .get_opt_key(rpc_key)
        .into_array_iter()
        .map(|urls| {
            urls.filter_map(|v| {
                let url = v.chain(err).get_key("http").parse_string().end();
                url.map(|url| RpcConfig {
                    url: url.to_owned(),
                    public: v
                        .chain(err)
                        .get_opt_key("public")
                        .parse_bool()
                        .unwrap_or(false),
                })
            })
            .collect()
        })
        .unwrap_or_default()
 ```
 
 What do you think?

File: typescript/sdk/src/metadata/chainMetadataTypes.ts [hunk 3]
```diff
@@ -95,6 +95,10 @@ export const RpcUrlSchema = z.object({
     .describe(
       'Default retry settings to be used by a provider such as MultiProvider.',
     ),
+  public: z
+    .boolean()
+    .optional()
+    .describe('Flag if the RPC is publicly available.'),
 });
```
yjamin, line 102: Would be great to get a confirmation that this is not breaking anything @paulbalaji

paulbalaji: as of right now it won't break anything - but it might later down the line when we start populating the registry. imo as long as we leave sufficient time between the SDK release containing this schema change and the Registry release containing the use of this updated schema, we're fine

rust/main/agents/validator/src/settings.rs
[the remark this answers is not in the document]
ameten: Does method `.into_array_iter()` returns `Option<Iter>` and default will be an empty iterator?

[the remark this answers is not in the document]
yjamin: It does but `the trait `Default` is not implemented for `impl Iterator<Item = ValueParser<'_>>``

rust/main/agents/validator/src/settings.rs
ameten: It was made well before my time in the org. @daniel-savu, do you know why we have it?

daniel-savu: part of the reasoning was to make it really easy to run your own agents and configure however's easier for you, but retrospectively it was over-engineered

some docs on this in case they're helpful: https://docs.hyperlane.xyz/docs/operate/agent-config#config-layers

rust/main/agents/validator/src/settings.rs
```diff
@@ -145,10 +174,60 @@ impl FromRawConf<RawValidatorSettings> for ValidatorSettings {
...
+                    Some(url) if public => Some(RpcConfig {
+                        url: url.to_owned(),
+                        public: true,
+                    }),
+                    Some(url) => Some(RpcConfig {
+                        url: url.to_owned(),
+                        public: false,
+                    }),
+                    _ => None,
+                }
```
ameten, line 224:
```suggestion
                match url {
                    Some(url) => Some(RpcConfig {
                        url: url.to_owned(),
                        public,
                    }),
                    _ => None,
                }
```

paulbalaji (commented):
you need to add a `minor` SDK changeset as well to mark the change to `chainMetadataTypes`

ameten (approved):

ameten (approved):

PR #5793, opened 2025-03-31
feat(validator): public rpc info in `metadata_latest.json`

### LINKED ISSUE #5591: Using a validator's metadata_latest.json, it should be possible to know if they are just using default public URLs

yjamin:
### Description

This PR announces a list of RPCs that the validator uses in `metadata_latest.json`. It also shows, whether or not the validator opt into using public RPCs. RPCs are kept private by only publishing the Keccak Hash of the URL.

Metadata structure: 
```rust
pub struct ValidatorMetadata {
    git_sha: String,
    rpcs: Vec<H256>,  // list of SHA3 to all RPCs, this includes gRPC & RPC
    allows_public_rpcs: bool,
}
```

### Drive-by changes

The `BaseAgent` trait now includes a type for the metadata that each agent has. This type has to implement `MetadataFromSettings`, which builds the agents metadata based on the agents settings.
```rust
pub trait BaseAgent {
    ...
    /// The agents metadata type
    type Metadata: MetadataFromSettings<Self::Settings>;
    ....
}
```

### Related issues

Closes #5591

> Note: Builds on top of #5792.
### Backward compatibility
Full backward compatibility.

### Testing
Manual testing of the `metadata_latest.json`


rust/main/agents/validator/src/validator.rs
```diff
@@ -52,17 +56,43 @@ pub struct Validator {
...
+}
+
+impl MetadataFromSettings<ValidatorSettings> for ValidatorMetadata {
+    /// Create a new instance of the agent metadata from the settings
+    fn build_metadata(settings: &ValidatorSettings) -> ValidatorMetadata {
+        // Hash all the RPCs for the metadata
+        let rpcs = settings
+            .rpcs
+            .iter()
+            .map(|rpc| H256::from_slice(&keccak256(&rpc.url)))
```
yjamin, line 77: Maybe we want to use a different hashing function?

ameten (approved):

Issue #4750, opened 2024-10-24
Generic non-native token warp route fees


nambrot:
Related to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3060, which is about charging the "mandatory" IGP fee from a users warp token balance (vs. as the local gas token). This issue is somewhat related, in that there are use cases where a non-native fees are charged independent of the IGP fee itself. For example the owner may want to charge a fee for all bridge transfers. Additionally, #3060 is limited in its quoting transparency, i.e. the `amount` parameter of `transferRemote` is not what is actually getting transferred and a UI like the warp template UI would be unable to display a quote to the user.

This issue tries to lay out a potential solution to this by adding a new additional quoting behavior for warp routes.

`struct Amount { address token; uint256 amount; }`
`function quoteTransferRemote(uint32 destination, bytes32 recipient, uint256 amountOrId) returns Amount[]`

`quoteTransferRemote` returns the amount of token(s) that this warp route will pull from the `amount` that is passed to `transferRemote` before embedding the amount in the `TokenMessage`. This allows these warp routes to be "backwards compatible" with warp route tooling like previous warp template UIs that don't know about this new quoting (and it will just not display the full quote).

`quoteTransferRemote` can also be extended with CCIP-read like behavior to make the client fetch a quote from an off-chain source. This would work well for intent protocols where the fee is very dynamic and depends on off-chain quotes from solvers. The downside of this option would be that the contract wouldn't be able to revert if not enough was passed.

You can also imagine this as kind of an experiment for alternative quoting mechanisms for the mailbox. I.e. a new kind of IGP can be used that emits the `GasPayment` event inside `transferRemote` instead of `postDispatch`.

open question:
- what other context should be passed into the quote?

PR #5794, opened 2025-03-31
feat: Add FeeTokenRouter

larryob:
We want to be able to charge fees in ERC20 tokens, as well in ETH/msg.value.

### Description

A new Router that can charge fees in ERC20 tokens.

### Drive-by changes
N/A

### Related issues

#4750 

### Backward compatibility

Yes

### Testing

Unit


solidity/contracts/token/libs/FeeTokenRouter.sol
```diff
@@ -0,0 +1,60 @@
...
+            amountOut
+        );
+        return quotes;
+    }
+
+    function quoteExternalFees(
+        uint32 destination,
+        bytes32 recipient,
+        uint256 amountOut
+    ) public view virtual returns (Quotes[] memory);
```
nambrot, line 59: Should this provide a default implementation of no quotes?

solidity/contracts/token/libs/FeeTokenRouter.sol
```diff
@@ -0,0 +1,60 @@
...
+
+        return _transferRemote(destination, recipient, amountIn, msg.value);
+    }
+
+    function quoteTransferRemote(
+        uint32 destination,
+        bytes32 recipient,
+        uint256 amountOut
+    ) public view virtual returns (Quotes[] memory) {
+        Quotes[] memory quotes = quoteExternalFees(
```
nambrot, line 47: Are IGP fees expected to be charged inside `quoteExternalFees`?

solidity/contracts/token/libs/FeeTokenRouter.sol
```diff
@@ -0,0 +1,60 @@
...
+struct Quotes {
+    address token;
+    uint256 amount;
+}
+
+/**
+ * @title Warp Route that charges non-native (ERC20) fees in addition to the native IGP fee.
+ * @author Abacus Works
+ */
+abstract contract FeeTokenRouter is TokenRouter {
```
nambrot, line 16: This should probably implement the `IValueTransferBridge` interface?

solidity/contracts/token/libs/FeeTokenRouter.sol
```diff
@@ -0,0 +1,60 @@
...
+    constructor(address _mailbox) TokenRouter(_mailbox) {}
+
+    function transferRemote(
```
nambrot, line 19: I feel like we still don't have a great story for the various overrides that exist for transferRemote, i.e. for overriding the hooks and hook metadata? @yorhodes how do you think we should handle it? This is also related to a feature for warp route owners to be able to charge fees for using the warp route, but today with the approach of using `PostDispatchHooks` users can technically just override that hook

yorhodes: hmm I mean I guess there are warp routes that would want to charge eg 5bps on the transfer amount but still allow overriding the hook for self relay

yorhodes: although that could all be implemented in the `_transferFromSender` hook

solidity/contracts/token/libs/FeeTokenRouter.sol
larryob: Done

solidity/contracts/token/libs/FeeTokenRouter.sol
larryob: Made the IGP fees more explicit

solidity/contracts/token/libs/FeeTokenRouter.sol
larryob: Done

larryob:
Closing because of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6562

Issue #5700, opened 2025-03-19
Build SVM ChainTxAdapter


daniel-savu:
Design doc for reference: https://www.notion.so/hyperlanexyz/Relayer-Submitter-Redesign-17b6d35200d68020b516c6c9bc455213?pvs=4

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5699

- [ ] Defines an `AdaptsSvm` trait which provides default implementations for subcomponents
- [ ] The subcomponents should consolidate existing logic for SVM chains (implemented [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/rust/main/chains/hyperlane-sealevel/src/trait_builder.rs#L1), [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/rust/main/chains/hyperlane-sealevel/src/tx_submitter.rs#L1), [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/rust/main/chains/hyperlane-sealevel/src/provider.rs#L1), [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/rust/main/chains/hyperlane-sealevel/src/rpc/client.rs#L1), and configured [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/typescript/infra/config/environments/mainnet3/agent.ts#L477))
- [ ] Up to the implementer of this task, but I assume the main adapters will be `GasPriceEstimator`, `GasLimitEstimator` and `TxSubmitter`. Since reading the blockhash is so simple I assume it can be done in the `TxSubmitter`
- [ ] It looks like the TxSubmitter for solana uses a different RPC, so the implementation should fit this well: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2d5db4ba25cf3258c387eaed7c7b3cdf0f3a295/typescript/infra/config/environments/mainnet3/agent.ts#L510

PR #5795, opened 2025-03-31
feat(submitter): Add max batch size and estimated block time implementation to Sealevel chain adapter

ameten:
### Description

Add max batch size and estimated block time implementation to Sealevel chain adapter

### Drive-by changes

- parse estimated block time from json
- place unit tests into separate modules

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5700

### Backward compatibility

Yes

### Testing

Unit tests


rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter/tests.rs
```diff
@@ -1,311 +1,7 @@
-use std::sync::Arc;
```
daniel-savu, line 1: i personally didn't mind having all the tests here but up to you :P

daniel-savu (approved):

PR #5801, opened 2025-03-31
feat(submitter): dispatcher finality stage 

daniel-savu:
### Description

Fully implements the Dispatcher's Finality Stage and unit tests it.

Design doc section for reference: https://www.notion.so/hyperlanexyz/Relayer-Submitter-Redesign-17b6d35200d68020b516c6c9bc455213?pvs=4#1a56d35200d6805d80b6c03598559c20

### Drive-by changes

- `AdaptsChain::reverted_payloads` now returns `Vec<PayloadDetails>` instead of `Vec<Uuid>`, for consistency with other methods and to avoid DB reads

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5800

### Backward compatibility

Yes

### Testing

Unit tests. See the cargo-llvm-cov report below (the columns are Function Coverage | Line Coverage | Region Coverage)
Screenshot 2025-04-01 at 00 01 13


rust/main/submitter/src/payload_dispatcher/stages/finality_stage.rs
```diff
@@ -1 +1,533 @@
...
+use super::{
+    building_stage::BuildingStageQueue, utils::retry_until_success, PayloadDispatcherState,
+};
+
+pub type FinalityStagePool = Arc<Mutex<HashMap<TransactionId, Transaction>>>;
+
+#[derive(new)]
+struct FinalityStage {
+    pool: FinalityStagePool,
+    inclusion_stage_receiver: mpsc::Receiver<Transaction>,
```
ameten, line 36: I was a little confused by the name `inclusion_stage_receiver`. I understand that it is a receiver which expects transactions from inclusion stage. I suppose finality stage can receive transactions either from inclusion stage or database. 

I wonder if it makes sense to call it just a `receiver` as a field while it will be `finality_stage_receive` when struct `FinalityStage` is created?

ameten (approved):

rust/main/submitter/src/payload_dispatcher/stages/finality_stage.rs
```diff
@@ -1 +1,533 @@
...
+    async fn receive_txs(
+        mut inclusion_stage_receiver: mpsc::Receiver<Transaction>,
+        pool: FinalityStagePool,
+    ) -> Result<(), DispatcherError> {
+        loop {
+            if let Some(tx) = inclusion_stage_receiver.recv().await {
+                pool.lock().await.insert(tx.id.clone(), tx.clone());
+                info!(?tx, "Received transaction");
+            } else {
+                sleep(Duration::from_millis(500)).await;
```
ameten, line 76: I think that `None` return value means that the channel is closed and no more messages will be delivered. If it happens, I suppose we report a critical error.

daniel-savu: since we don't have the metrics set up for the MVP, I'll just log an error

PR #5803, opened 2025-03-31
feat(submitter): dispatcher db loader

daniel-savu:
### Description

Core logic is duplicated from the `ForwardBackwardIterator` used in [the relayer](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ac008ba4155465b5500a791b2d4f345ed6cc8c29/rust/main/agents/relayer/src/msg/processor.rs#L54) to load messages from the DB, with some additions:
- made it generic in terms of the type being loaded
- made the forward cursor optional, to cater for loading `Transaction`s (since new txs will be received over the stage channels and there is no need to load them from the DB)
- added unit tests to make sure the loader works correctly

Based on the coverage report, it seems like the last bit of Dispatcher that definitely needs testing coverage is the `LoadableFromDb` for the Payload and Transaction.

### Drive-by changes

- added unit tests for the PayloadDb and TransactionDb trait implementations, luckily did catch some bugs where the wrong prefixes / order of values were used
- `test_utils.rs` had an unnecessary `mod test` inside the `test_utils` module that contained all the logic. I removed the `mod test` since it was just being wrapped by `test_utils`

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5802

### Backward compatibility

Yes

### Testing

Unit testing. See the cargo-llvm-cov report below (the columns are Function Coverage | Line Coverage | Region Coverage)
Screenshot 2025-04-03 at 01 08 26


rust/main/submitter/src/payload/db.rs
```diff
@@ -23,20 +26,48 @@ pub trait PayloadDb: Send + Sync {
     /// Store a payload by its unique ID
     async fn store_payload_by_id(&self, payload: &FullPayload) -> DbResult<()>;
 
+    /// Retrieve a payload by its unique ID
```
ameten, line 29:
```suggestion
    /// Retrieve a payload index by its unique ID
```

rust/main/submitter/src/payload/db.rs
```diff
@@ -23,20 +26,48 @@ pub trait PayloadDb: Send + Sync {
...
+    async fn retrieve_payload_index_by_id(&self, payload_id: &PayloadId) -> DbResult<Option<u32>>;
+
+    async fn store_payload_index_by_id(&self, index: u32, payload_id: &PayloadId) -> DbResult<()>;
```
ameten, line 32:
```suggestion
    /// Store a payload index by payload unique ID
    async fn store_payload_index_by_id(&self, index: u32, payload_id: &PayloadId) -> DbResult<()>;
```

rust/main/submitter/src/payload/db.rs
```diff
@@ -23,20 +26,48 @@ pub trait PayloadDb: Send + Sync {
     /// Store a payload by its unique ID
     async fn store_payload_by_id(&self, payload: &FullPayload) -> DbResult<()>;
 
+    /// Retrieve a payload by its unique ID
+    async fn retrieve_payload_index_by_id(&self, payload_id: &PayloadId) -> DbResult<Option<u32>>;
```
ameten, line 30:
```suggestion
    /// Retrieve a payload index by payload unique ID
    async fn retrieve_payload_index_by_id(&self, payload_id: &PayloadId) -> DbResult<Option<u32>>;
```

rust/main/submitter/src/payload/db.rs
```diff
@@ -23,20 +26,48 @@ pub trait PayloadDb: Send + Sync {
...
+
+    async fn store_payload_index_by_id(&self, index: u32, payload_id: &PayloadId) -> DbResult<()>;
+
+    /// Retrieve a payload by its unique ID
+    async fn retrieve_payload_id_by_index(&self, index: u32) -> DbResult<Option<PayloadId>>;
+
+    /// Store a payload ID by its index
+    async fn store_payload_id_by_index(&self, index: u32, payload_id: &PayloadId) -> DbResult<()>;
```
ameten, line 38:
```suggestion
    /// Store a payload unique ID by payload index
    async fn store_payload_id_by_index(&self, index: u32, payload_id: &PayloadId) -> DbResult<()>;
```

rust/main/submitter/src/payload/db.rs
```diff
@@ -23,20 +26,48 @@ pub trait PayloadDb: Send + Sync {
...
+
+    /// Retrieve a payload by its unique ID
+    async fn retrieve_payload_by_index(&self, index: u32) -> DbResult<Option<FullPayload>> {
```
ameten, line 41:
```suggestion
    /// Retrieve a payload by its index
    async fn retrieve_payload_by_index(&self, index: u32) -> DbResult<Option<FullPayload>> {
```

rust/main/submitter/src/transaction/db.rs
ameten, line 1: Same feedback as for `rust/main/submitter/src/payload/db.rs` regarding comments.

ameten:
I wonder if it would make sense to move everything related to database under `payload_dispatcher`. The modules `payload` and `transaction` shared with other parts of the code which don't need to know about database owned by PayloadDispatcher.

ameten (approved):

PR #5811, opened 2025-04-01
feat: stub out support for ProtocolType.CosmosModule

paulbalaji:
### Description

feat: stub out support for ProtocolType.CosmosNative
- added in agents https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/c214952e7da0dfce3ef307920762c9af8d80656c/rust/main/hyperlane-core/src/chain.rs#L282-L283
- just adding the new enum, and then updating anything that references it
- reuse existing cosmos logic for now until @troykessler's SDK update PRs flesh out the implementation

This is so we can have agree on the ProtocolType name and stub out functionality so that we can then update the registry schema and start getting real cosmosmodule chain metadata in there.


typescript/utils/src/types.ts
```diff
@@ -5,6 +5,7 @@ export enum ProtocolType {
   Ethereum = 'ethereum',
   Sealevel = 'sealevel',
   Cosmos = 'cosmos',
+  CosmosModule = 'cosmosnative',
```
paulbalaji, line 8: this one line of addition necessitated all the other additions in this PR

chose `CosmosModule` to avoid confusion/clash with the existing `CosmosNative` TokenStandard

xeno097: I'd have no problems with having the enum variant being called `CosmosNative` but happy to keep it as it is too, because in any case we would need to use the value as `ProtocolType. CosmosNative` \ `ProtocolType. CosmosModule`

paulbalaji: ok cool lets keep it CosmosNative then to make it easy for ctrl+f

xeno097 (approved):

PR #5817, opened 2025-04-02
fix(test): e2e increate timeout

yjamin:
### Description
Increase chain spin up time for cosmos native E2E tests


paulbalaji:
think you accidentally have commits from a different PR here

paulbalaji (approved):

PR #5820, opened 2025-04-02
refactor: add DerivedTokenRouterConfig

### LINKED ISSUE #4932: chore: add DerivedTokenRouterConfig to EvmERC20WarpModule
## Problem
Currently, we use HypTokenRouterConfig within the EvmERC20WarpModule functions. However, when we derive a config using the other modules, namely Hook and ISM, `address` is added to the respective object/config. 

This causes some casting logic and is unclear to devs that derived configs contain an `address`. eg. https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/sdk/src/token/EvmERC20WarpModule.ts#L300-L314

## Solution
Create and replace the HypTokenRouterConfig with DerivedTokenRouterConfig for derived configs (actualConfig).

antigremlin:
### Description

This PR introduces a DerivedTokenRouterConfig type that extends hooks and ISMs with the address field. It pulls the derived hook and ISM config types into the corresponding types.js files, and adds functions to safely extract the addresses from both (as it's not a trivial expression even with better static types).

### Related issues

Resolves #4932 

### Testing

Unit tests and hardhat tests pass locally.


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -68,23 +68,25 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
     const destinationGas = await this.fetchDestinationGas(warpRouteAddress);
 
     return {
-      ...baseMetadata,
+      ...mailboxClientConfig,
       ...tokenConfig,
       remoteRouters,
       proxyAdmin,
       destinationGas,
       type,
```
antigremlin, line 88: This might be a bug. We already calculate a type for this TokenRouterConfig in `fetchTokenConfig()` and that returns a correct type. If you remove this line, the cast on the next line can also be removed.

ltyu (approved):

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: can prob remove `type` because it's being returned in `tokenConfig`

PR #5821, opened 2025-04-02
feat: Use salts in ICA derivation

### LINKED ISSUE #5353: ICA callers should be able to pass a salt
Use cases:

- A middleware contract wants to use ICAs and expose behavior to its callers. However, it needs differential permissioning on the destination chain for the different callers.
- Somewhat similar to https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5352, a caller just wants to be able to segment state and have control over which state a call should apply to.

Proposed solution:
- allow an optional salt to be passed by the ICARouter caller that is part of the ICA derivation.

larryob:
### Description

You can now pass a salt to `callRemoteWithOverrides`. See #5353 for context.

### Drive-by changes

Some code in `InterchainAccountMessage` was moved around.

### Related issues

Fixes #5353 

### Backward compatibility

No, see the changes to `_getSalt`.

### Testing

Unit


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -650,15 +813,20 @@ contract InterchainAccountRouter is Router {
...
         bytes32 _owner,
         bytes32 _router,
-        bytes32 _ism
+        bytes32 _ism,
+        bytes32 _userSalt
     ) private pure returns (bytes32) {
-        return keccak256(abi.encodePacked(_origin, _owner, _router, _ism));
+        return
+            keccak256(
+                abi.encodePacked(_origin, _owner, _router, _ism, _userSalt)
```
yorhodes, line 828: I think this is the right tradeoff but just noting this means already deployed proxies are incompatible with this derivation and we should not roll this out as an ICA upgrade

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -32,7 +32,29 @@ library InterchainAccountMessage {
     ) internal pure returns (bytes memory) {
         CallLib.Call[] memory _calls = new CallLib.Call[](1);
         _calls[0] = CallLib.build(_to, _value, _data);
-        return abi.encode(TypeCasts.addressToBytes32(_owner), _ism, _calls);
+        return
+            abi.encode(
+                TypeCasts.addressToBytes32(_owner),
+                _ism,
+                _calls,
+                bytes32(0) // Salts are expected when decoding.
```
yorhodes, line 40: should we pull this out into a shared constant?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -802,9 +802,31 @@ contract InterchainAccountRouterTest is InterchainAccountRouterTestBase {
...
+        OwnableMulticall ica2 = destinationIcaRouter
+            .getDeployedInterchainAccount(
+                origin,
+                owner.addressToBytes32(),
+                address(originIcaRouter).addressToBytes32(),
+                address(environment.isms(destination)),
+                keccak256("i am a different salt")
+            );
+        assertNotEq(address(ica), address(ica2));
+    }
```
yorhodes, line 831: should we also test the same with the same salt?

yorhodes (approved):
nice work!

yorhodes (changes requested):
for changes I mentioned in slack thread

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Done

solidity/test/InterchainAccountRouter.t.sol
larryob: Done

yorhodes (approved):

PR #5825, opened 2025-04-03
feat: added cosmos-sdk

troykessler:
### Description

This PR adds a TypeScript SDK for the Hyperlane Cosmos Module. It extends the base SDK from CosmJs and adds all the Hyperlane specific queries and transactions to it so it can be used in the typescript/sdk package to support the Hyperlane Cosmos Module there.

This package still lacks tests, I would like to discuss whether we should implement e2e tests where a local Cosmos chain is running and transactions from this sdk are submitted against it or if unit tests which assert the transaction output before they are broadcasted are enough.

The next steps include using this new SDK as the CosmJs Provider in the Hyperlane SDK and adding the CosmosHypCollateral and CosmosHypSynthetic token adapter.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Testing is planned, still needs to be discussed if e2e or unit tests should be implemented


antigremlin:
I suggest we rename this PR and the corresponding commit to `feat: add cosmos-sdk`. Grammatically, it should use "imperative mood", add vs. added.

antigremlin:
As this is adding a new module, do we need to crate a changeset as per the changeset-bot?

typescript/cosmos-sdk/src/index.ts
```diff
@@ -0,0 +1,603 @@
...
+  WarpExtension &
+  CoreExtension &
+  InterchainSecurityExtension &
+  PostDispatchExtension;
+
+export interface TxResponse<R> extends DeliverTxResponse {
+  response: R;
+}
+
+export class HyperlaneModuleClient extends StargateClient {
```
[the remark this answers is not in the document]
xeno097: ++

typescript/cosmos-sdk/src/registry/index.ts
```diff
@@ -0,0 +1,416 @@
+import * as types from '@hyperlane-xyz/cosmos-types';
+
+// amino converters can be null by default since the
+// converters for proto can be taken. In rare cases
+// they can differ, in that instance the amino
+// converters can be overridden here by adding the
+// methods "toJSON" and "fromJSON".
+export const REGISTRY = {
```
xeno097, line 8: Considering that this constant is exported from the cosmos SDK package, can we rename it to something like `COSMOS_MODULE_MESSAGE_REGISTRY` (or better name). Otherwise, it might be confusing with the actual hyperlane registry

typescript/cosmos-sdk/src/registry/index.ts
```diff
@@ -0,0 +1,416 @@
...
+  },
+  MsgCreateMailboxResponse: {
+    proto: {
+      type: '/hyperlane.core.v1.MsgCreateMailboxResponse' as const,
+      converter: types.coreTx.MsgCreateMailboxResponse,
+    },
+    amino: {
+      type: '',
+      converter: null,
+    },
```
xeno097, line 28: Does anythig break if we don't define the `amino` property at all when bot fields are empty/no defined

typescript/cosmos-sdk/src/index.ts
```diff
@@ -0,0 +1,603 @@
...
+      options,
+    );
+  }
+
+  private async submitTx<T>(
+    msg: EncodeObject,
+    options?: {
+      fee?: StdFee | 'auto' | number;
+      memo?: string;
+    },
```
xeno097, line 228: nit: would it be possible to refactor this type  into a named type so we can avoid "duplicating" it because I see the same type being used in multiple methods in this class. Could be something like this:

```typescript
type TxOptions = {
  fee?: StdFee | 'auto' | number;
  memo?: string;
}
```

typescript/cosmos-sdk/src/registry/index.ts
troykessler: no, just wanted to already define it since in the future it can be the case that those converters need to be defined. Let me know if we rather should remove them now

troykessler: So when removing the amino part the inferred type breaks, making it quite difficult to work with the registry object when removin only the unused amino parts

typescript/cosmos-sdk/src/registry/index.ts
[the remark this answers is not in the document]
antigremlin: I mean a `registry.js` file side-by-side with `index.js`, not nested in a subdirectory like now - `src/registry/registry.js` - 😄 But that's OK.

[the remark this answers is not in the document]
troykessler: Ah, that of course makes more sense, move to `src/registry.ts` now

typescript/cosmos-sdk/src/registry/index.ts
xeno097: What if we do something like this?

```typescript
type RawCosmosModuleMessageRegistry<T> = {
  [Key in keyof Omit<
    T,
    'protobufPackage' | 'MsgServiceName' | 'MsgClientImpl'
  >]: {
    proto: {
      type: string;
      converter: T[Key];
    };
    amino?: {
      type: string;
      converter?: T[Key];
    };
  };
};

type CoreCosmosModuleMesageRegistry = RawCosmosModuleMessageRegistry<
  typeof types.coreTx
>;
type IsmCosmosModuleMessafeRegistry = RawCosmosModuleMessageRegistry<
  typeof types.isTx
>;
type PostDispatchCosmosModuleMessageRegistry = RawCosmosModuleMessageRegistry<
  typeof types.pdTx
>;
type WarpTransactionCosmosModuleMessageRegistry =
  RawCosmosModuleMessageRegistry<typeof types.warpTx>;

type CosmosModuleMessageRegistry = CoreCosmosModuleMesageRegistry &
  IsmCosmosModuleMessafeRegistry &
  PostDispatchCosmosModuleMessageRegistry &
  WarpTransactionCosmosModuleMessageRegistry;

export const COSMOS_MODULE_MESSAGE_REGISTRY: CosmosModuleMessageRegistry = {
...
}
```

troykessler: works perfectly, awesome!

xeno097 (approved):

paulbalaji:
> As this is adding a new module, do we need to crate a changeset as per the changeset-bot?

Nope, changesets will automatically pickup the new package at release time

PR #5830, opened 2025-04-03
feat: Add Starknet contract ABI fetching and contract artifact generation [STARKNET-01]

ljankovic-txfusion:
### Description

1st Starknet PR

This PR adds the `@hyperlane-xyz/starknet-core` package for fetching Starknet contract ABI and generating contract artifacts. The package provides TypeScript tooling for interacting with Cairo contracts ABI on Starknet, including auto-generated compiled contract ABIs from Astraly Labs Hyperlane repo. 

### Drive-by changes

None

### Related issues

Resolves ENG-1270

### Backward compatibility

This is a new package that doesn't modify existing functionality

### Testing

Manual testing of contract fetching and artifact generation


starknet/scripts/generate-artifacts.ts
```diff
@@ -0,0 +1,314 @@
...
+      ? ContractClass.CASM
+      : ContractClass.SIERRA;
+  }
+
+  /**
+   * @notice Retrieves paths of all relevant artifact files
+   */
+  async getArtifactPaths() {
+    const sierraPattern = `${RELEASE_DIR}/**/*${CONFIG.CONTRACT_FILE_SUFFIXES.SIERRA_JSON}`;
+    const [sierraFiles] = await Promise.all([globby(sierraPattern)]);
```
ljankovic-txfusion, line 114: We only glob `SIERRA` files which contain `abi`, and skip `CASM` files with deployable bytecode

starknet/scripts/fetch-contracts-if-missing.ts
paulbalaji, line 1: this could just be part of the fetch-contracts-release.sh script

perhaps a "dev" or "release" arg to switch between the operation modes

package.json
```diff
@@ -39,7 +39,8 @@
   },
   "workspaces": [
     "solidity",
-    "typescript/*"
+    "typescript/*",
+    "starknet"
```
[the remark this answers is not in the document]
ljankovic-txfusion: Semantically speaking, there are no cairo contracts there. But if we are going for that approach then I could rename package to `@hyperlane-xyz/cairo`

starknet/scripts/fetch-contracts-if-missing.ts
ljankovic-txfusion: what would be the difference between "dev" or "release"? contract release version?

paulbalaji: dev --> does what fetch-contracts-if-missing.ts does
release -> existing behaviour of fetch-contracts-release.sh

ljankovic-txfusion: What is the benefit? fetch-contracts-if-missing just check if the contract artifacts exist locally in releases folder, so we don't double fetch, something like `cache` for contract artifacts that already exists in solidity folder.

package.json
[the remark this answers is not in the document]
paulbalaji: ah true - starknet probably fine to keep

starknet/scripts/fetch-contracts-if-missing.ts
ljankovic-txfusion: Initially @mshojaei-txfusion worked on it not to bloat `.sh` too much, which can be difficult to understand on its own, but since check for releases is universal and it does not add too much code I refactored to a single bash script, and removed fetch-contracts-if-missing.ts entirely

antigremlin (commented):
Please see my comments

starknet/scripts/generate-artifacts.ts
ljankovic-txfusion: I am trying to avoid doing double work, this PR is indeed incomplete feature wise, and there is a branch for generating CASM objects working. So after the contract audit is complete, we would only need to make couple changes to this file. If you insist, I can remove logic related to casm?

starknet/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "types": "./dist/index.d.ts",
+  "files": [
+    "/dist"
```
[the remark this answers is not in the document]
ljankovic-txfusion: In all other packages dist is prefixed by slash such as [solidity/package.json](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/package.json)

starknet/scripts/generate-artifacts.ts
```diff
@@ -0,0 +1,314 @@
...
+enum ContractType {
+  CONTRACT = 'contracts_',
+  TOKEN = 'token_',
+  MOCK = 'mock_',
+}
+
+/**
+ * @notice Templates for TypeScript artifact generation
+ */
+const TEMPLATES = {
```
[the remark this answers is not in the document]
ljankovic-txfusion: This looks really good, was wondering if there was a neater solution

starknet/scripts/generate-artifacts.ts
antigremlin: I don't insist, of course 😄 - I was just trying to understand why we're writing more code than using.

starknet/package.json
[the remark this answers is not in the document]
antigremlin: Even if that works fine, we could start using the relative paths as they are expected by the tools. I understand that the confusion comes from the NPM's spec not saying the path should be relative explicitly. But if you search GitHub with `path:package.json "files"` you can see it's relative in virtually all cases.

starknet/src/index.ts
```diff
@@ -0,0 +1,63 @@
+import { CairoAssembly, CompiledContract } from 'starknet';
+
+import { starknetContracts } from './artifacts/index.js';
```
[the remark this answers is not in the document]
antigremlin: I'm sorry, I wasn't very clear. Using dynamic imports would require more work. So I intended that comment as a question (have we considered this?) rather than a suggestion to implement (please do it). If there's enough value in dynamic imports, we should create a separate task.

starknet/scripts/generate-artifacts.ts
ljankovic-txfusion: No, really, if you don't like this I will remove it no problem 😄

starknet/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  try {
+    const group = getContractGroup(contractType);
+    const contract = group[name];
+
+    if (!contract?.contract_class) {
+      throw new Error('Contract not found or missing Sierra class');
+    }
+
+    return contract.contract_class;
+  } catch (_error) {
```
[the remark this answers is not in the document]
ljankovic-txfusion: Updated

starknet/src/contract-retriever.ts
```diff
@@ -0,0 +1,66 @@
+import { CompiledContract } from 'starknet';
+
+import { starknetContracts } from './artifacts/index.js';
+import { CONSTANTS } from './const.js';
+import { ContractError } from './errors.js';
+import { ContractType, StarknetContractGroup } from './types.js';
+
+/**
+ * @notice Retrieves a compiled contract
+ * @param name The name of the contract to retrieve
```
antigremlin, line 10: Should we document the 2nd param?

starknet/src/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class ContractError extends Error {
```
antigremlin, line 1: nit: It would be easier to read if the known error codes were declared in the same file. Also, we can give the object a shorter name, like ERR_CODES, and that'll make the `throw new ContractError(code...` lines format better:

```typescript
throw new ContractError(ERR_CODES.INVALID_CONTRACT_GROUP,
```

starknet/src/contract-retriever.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * @notice Helper function to get the correct contract group
+ */
+function getContractGroup(type: ContractType): StarknetContractGroup {
+  const group = starknetContracts[type];
+  if (!group) {
+    throw new ContractError(
+      CONSTANTS.CONTRACT_ERROR_CODES.INVALID_CONTRACT_GROUP,
```
antigremlin, line 59: Naming inconsistency: we pass a ContractType but the error code is INVALID_CONTRACT_GROUP. Should it be INVALID_CONTRACT_TYPE?

antigremlin (approved):

starknet/src/contract-retriever.ts
ljankovic-txfusion: Resolved

starknet/src/contract-retriever.ts
ljankovic-txfusion, line 1: @antigremlin Do you feel like I should include some tests for these contract retrieval functions in this PR?

antigremlin: `contract-retriever.ts` seems trivial to test. It would be more useful to `generate-artifacts.ts` as that code is complex and is more likely to have bugs.

ljankovic-txfusion: Actually I'll make a separate PR just for tests, had to install more packages etc.. & changes in `generate-artifacts.ts` for it to be compatible with tests so I don't want to bloat this PR. Test with current generate-artifacts.ts are working

starknet/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@hyperlane-xyz/starknet-core",
```
paulbalaji, line 2: you'll need to ensure the Dockerfile knows about this new package too

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0cd15ae26e9348d7c13132daf2bb6c7606509082/Dockerfile#L24

ie just add after the solidity line
```Dockerfile
 COPY starknet/package.json ./starknet/ 
```

paulbalaji (commented):
lgtm once Dockerfile is updated

PR #5831, opened 2025-04-03
chore: move rust caching to depot

paulbalaji:
### Description

depot claim:
> we transparently redirect anything bound for GitHub's Actions cache to our own for extra speed

this PR moves the swatinem backend to github, to test whether this redirects to depot on the depot runner

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,15 +69,6 @@ jobs:
           # check out full history
           fetch-depth: 0
 
-      - name: yarn-cache
-        uses: buildjet/cache@v4
-        with:
-          path: |
-            **/node_modules
-            .yarn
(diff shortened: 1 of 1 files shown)


daniel-savu (approved):

xeno097 (approved):

.github/workflows/test.yml
```diff
@@ -274,7 +260,7 @@ jobs:
         with:
           prefix-key: 'v3-rust-e2e'
           shared-key: ${{ matrix.e2e-type }}
-          cache-provider: 'buildjet'
+          cache-provider: 'github'
```
daniel-savu, line 263: do we not need to specify depot anywhere in here?

paulbalaji: nope! depot does a bunch of magic such that their cache/image builds etc work nicely with their runners

image

PR #5834, opened 2025-04-03
feat: easy RPC configuration for agents for all chains

mkykadir:
### Description

This PR introduces a new common configuration field `rpcUrls` which is a string argument expecting list of comma separated RPC URLs. Regardless of chains, any chain RPC URL can be provided in list and while configuration is loaded each RPC will be queried for its chain id so its utilized within related chain configuration.

This makes configuration very easy, especially for automated deployment environments.

### Drive-by changes

Configuration loading is now async

### Related issues

NA

### Backward compatibility

Changes are backward compatible. Common `rpcUrls` field is optional, when provided URL preferred over chain specific fields so that its easy to configure RPC URLs by default. If common (new) field is not provided RPC URL configuration works as usual.

### Testing

Manual, `rpcUrls` provided and tested


mkykadir:
Docs PR https://github.com/hyperlane-xyz/v3-docs/pull/377

mkykadir:
updated with latest main

PR #5838, opened 2025-04-04
feat: Starknet SDK logic integration [STARKNET-03]

ljankovic-txfusion:
### Description

3nd Starknet PR

[Diff with previous PR - feat: Starknet address utils #5851](https://github.com/txfusion/hyperlane-monorepo/pull/28/files)

This PR introduces Starknet logic to enable token warping on the UI. 

### Drive-by changes

- Utilities for working with Hyperlane messaging on Starknet
- Starknet testChain obj

### Related issues

None

### Backward compatibility

Yes

### Testing

Manual testing of token warping using warp ui template


typescript/sdk/src/providers/transactionFeeEstimators.ts
```diff
@@ -280,9 +282,28 @@ export function estimateTransactionFee({
...
+  transaction: _transaction,
+  provider: _provider,
+  sender: _sender,
+}: {
+  transaction: StarknetJsTransaction;
+  provider: StarknetJsProvider;
+  sender: Address;
+}): Promise<TransactionFeeEstimate> {
+  return { gasUnits: 0, gasPrice: 0, fee: 0 };
+}
```
xeno097, line 309: Can't we use the trick used for Zksync chains where we use a known address? We might use one of the priv key + address combos from starknet devnet to create a fake account instance

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -186,12 +186,14 @@ export class WarpCore {
...
     sender,
+    recipient,
     senderPubKey,
     interchainFee,
   }: {
     originToken: IToken;
     destination: ChainNameOrId;
     sender: Address;
+    recipient?: Address;
     senderPubKey?: HexString;
```
xeno097, line 197: Why do we need these changes if they aren't doing anything new at the moment?

typescript/sdk/src/providers/transactionFeeEstimators.ts
ljankovic-txfusion: 1. In starknet all addresses are contracts that need to be deployed. You can generate account address but it wont pass getting nonce for estimation.
2. I took a look at starknet golang rpc node, and they are doing `simulate` tx to estimate gas for `Invoke` transactions such as this one, so If we input random address+priv key for tx that we know wont work, simulation will fail. 
3. IIUC for gas estimation rpc_method, there are `from_address` field and a `signature` field generated using strk signer, which need to match which leads to bullet point 2.

typescript/sdk/src/utils/starknet.ts
```diff
@@ -0,0 +1,129 @@
...
+  MERKLE_TREE_HOOK = 'merkle_tree_hook',
+  NOOP_ISM = 'noop_ism',
+  HOOK = 'hook',
+  PROTOCOL_FEE = 'protocol_fee',
+  VALIDATOR_ANNOUNCE = 'validator_announce',
+  MESSAGE_RECIPIENT = 'message_recipient',
+  DOMAIN_ROUTING_HOOK = 'domain_routing_hook',
+  FALLBACK_DOMAIN_ROUTING_HOOK = 'fallback_domain_routing_hook',
+  STATIC_AGGREGATION_HOOK = 'static_aggregation_hook',
+}
```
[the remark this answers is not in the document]
ljankovic-txfusion: `StarknetContractName` is enum dedicated only to contracts used typescript/SDK, `starknet-core` has many more contracts

typescript/sdk/src/utils/starknet.ts
```diff
@@ -0,0 +1,129 @@
...
+export function getStarknetMailboxContract(
+  address: string,
+  providerOrAccount?: ProviderInterface | AccountInterface,
+): Contract {
+  return getStarknetContract(
+    StarknetContractName.MAILBOX,
+    address,
+    providerOrAccount,
+  );
+}
```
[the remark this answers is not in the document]
ljankovic-txfusion: We don't know Mailbox contract `ABI` until we call `getStarknetContract` bc abi is extracted like :

```
const {abi} = getStarknetContract
```

But its actually a good idea to to see if I can get typings for these contracts somehow

typescript/sdk/src/warp/WarpCore.ts
ljankovic-txfusion: removed unused recipient

typescript/utils/src/addresses.ts
```diff
@@ -24,10 +31,12 @@ const COSMOS_FACTORY_TOKEN_REGEX = new RegExp(
 const EVM_TX_HASH_REGEX = /^0x([A-Fa-f0-9]{64})$/;
 const SEALEVEL_TX_HASH_REGEX = /^[a-zA-Z1-9]{88}$/;
 const COSMOS_TX_HASH_REGEX = /^(0x)?[A-Fa-f0-9]{64}$/;
+const STARKNET_TX_HASH_REGEX = /^(0x)?[0-9a-fA-F]{64}$/;
```
xeno097, line 34: Wouldn't this fail if the address omits leading 0s but is still a valid starknet address? i have seen the stark token address being referenced both as `0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7` and `0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7`

typescript/sdk/src/utils/starknet.ts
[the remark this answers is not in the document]
xeno097: Gucci, I assume you are going to look at this in another PR, right?

typescript/utils/src/addresses.ts
ljankovic-txfusion: Took it from starknetjs itself https://github.com/starknet-io/starknet.js/blob/e369fe291f18a0eaf7d498afc7aa7b7c4abdab58/src/utils/address.ts#L55

typescript/sdk/src/utils/starknet.ts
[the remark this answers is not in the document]
ljankovic-txfusion: Yeah, need to figure out how to get typings in a way not to introduce technical debt

xeno097 (approved):
I'll approve but i'd like someone else to also have a look too @paulbalaji @ltyu

antigremlin (commented):
This is beyond my understanding so far. I'll do my best going through the SDK changes at least, but please don't wait for my comments.

paulbalaji (approved):

PR #5840, opened 2025-04-04
chore: remove mailboxes from warp configs

### LINKED ISSUE #5374: Remove mailbox addresses from warp deploy configs in the registry

antigremlin:
### Description

This change removes the mailbox addresses from all warp routes that are written into the registry. Instead of extensive changes to the code generating the warp configs, we simply remove the mailbox addresses from the configs before adding them to the registry.

### Related issues

Resolves #5374 

### Backward compatibility

From https://github.com/hyperlane-xyz/hyperlane-registry/pull/679:

> Yes. This change is backward compatible as it only removes redundant configuration fields that are not used in the deployment process. The removal of mailbox configurations does not affect the functionality of the warp routes or their deployment process.

### Testing

Tested by exporting warp configs and comparing with the ones in the registry branch for https://github.com/hyperlane-xyz/hyperlane-registry/pull/679

### DIFF
--- a/typescript/infra/scripts/warp-routes/export-warp-configs.ts
+++ b/typescript/infra/scripts/warp-routes/export-warp-configs.ts
@@ -1,3 +1,5 @@
+import { WarpRouteDeployConfig } from '@hyperlane-xyz/sdk';
+
 import { getRegistry } from '../../config/registry.js';
 import { getWarpConfig, warpConfigGetterMap } from '../../config/warp.js';
 import { getArgs, withOutputFile } from '../agent-utils.js';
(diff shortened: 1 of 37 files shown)


typescript/infra/scripts/warp-routes/export-warp-configs.ts
```diff
@@ -14,14 +17,22 @@ async function main() {
...
+    const warpConfigs = await getWarpConfig(
       multiProvider,
       envConfig,
       warpRouteId,
     );
 
+    const registryConfig: WarpRouteDeployConfig = objMap(
+      warpConfigs,
+      (chain, config) => {
+        const { mailbox: _mailbox, ...rest } = config;
```
paulbalaji, line 29: `chain` isn't used anywhere either, could make that `_`

typescript/infra/scripts/warp-routes/export-warp-configs.ts
```diff
@@ -14,14 +17,22 @@ async function main() {
   for (const warpRouteId of warpIdsToCheck) {
     console.log(`Generating Warp config for ${warpRouteId}`);
 
-    const warpConfig = await getWarpConfig(
+    const warpConfigs = await getWarpConfig(
```
paulbalaji, line 20: it's technically all still part of one `warpConfig` - but with subconfigs per-chain

mshojaei-txfusion (approved):

PR #5841, opened 2025-04-04
chore(ci): tidy up e2e rust-changes check

paulbalaji:
### Description

chore(ci): tidy up e2e rust-changes check
- for PRs: use the gh cli instead for better diff detection
- for main: use git diff just on the latest commit to main

basically i noticed on andrey's ts-only pr https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5840 it was still doing the full e2e, because his branch was not up to date with main which did have new rust changes - so doing a naive diff will give you way more rust-change hits than you really need. but using the gh cli you get a more accurate reflection of files changed in a PR

### Related issues

something i would like to do in a future PR is to extract this check out into its own action so we only do some of the other rust jobs only when we detect changes


.github/workflows/test.yml
```diff
@@ -221,15 +221,35 @@ jobs:
       # - on the main branch
       # - PR with changes to ./rust
       # - PR with changes to this workflow
-      - name: Check for Rust file changes
+      - name: Check for Rust file changes using GitHub CLI
```
paulbalaji, line 224:
```suggestion
      - name: Check for Rust file changes
```

.github/workflows/test.yml
```diff
@@ -221,15 +221,35 @@ jobs:
...
+              echo "rust_changes=true" >> $GITHUB_OUTPUT
+              echo "Changes detected in Rust files or workflow on main branch:"
+              echo "$CHANGES"
+            else
+              echo "rust_changes=false" >> $GITHUB_OUTPUT
+            fi
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            # For PRs, use GitHub CLI to check changes
+            PR_NUMBER=${{ github.event.pull_request.number }}
+            CHANGES=$(gh pr view $PR_NUMBER --json files -q '.files[].path | select(startswith("rust/") or . == ".github/workflows/test.yml")')
```
paulbalaji, line 244: nice
image

ameten (approved):

PR #5843, opened 2025-04-04
fix: add hardcoded fail message to TestSendReceiver

kamiyaa:
### Description

 - add `--body` parameter to `kathy` to allow for custom message body content
   - defaults to `0x1234` like before
 - hardcoded `handle()` function fail on `0xfa17ed` message body
   - this is so we can deterministically create a message that succeeds on `dispatch()` but fails on `handle()`
   - so in e2e we can catch these and invalidate cache

### Drive-by changes
 - remove unnecessary log messages

### Backward compatibility
yes

### Testing
 - update foundry tests to test new message body fails for `handle()` as expected
 - e2e tests work now


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -272,7 +272,6 @@ impl PendingOperation for PendingMessage {
         // If gas estimation fails, invalidate cache and rebuild it again.
         let tx_cost_estimate = match self.metadata.as_ref() {
             Some(metadata) => {
-                tracing::debug!(USE_CACHE_METADATA_LOG);
```
tkporter, line 275: did you still want this log in the `Ok(s) => Some(s),` ?

solidity/contracts/test/TestSendReceiver.sol
```diff
@@ -47,10 +49,15 @@ contract TestSendReceiver is IMessageRecipient {
         );
     }
 
-    function handle(uint32, bytes32, bytes calldata) external payable override {
+    function handle(
```
tkporter, line 52: can we actually keep the probabilistic reverting as well, and have this just be a special case?

the idea is that we want e2e coverage in the event that a message isn't immediately deliverable but is later

solidity/contracts/test/TestSendReceiver.sol
```diff
@@ -15,6 +15,8 @@ contract TestSendReceiver is IMessageRecipient {
     using TypeCasts for address;
 
     uint256 public constant HANDLE_GAS_AMOUNT = 50_000;
+    // taken from Message.sol
+    uint256 public constant BODY_OFFSET = 77;
```
tkporter, line 19: instead of this, you can import `../libs/Message.sol`, add `using Message for bytes;`, and then do `message.body()`. e.g. https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/test/TestMessage.sol#L24

rust/main/utils/run-locally/src/main.rs
```diff
@@ -418,7 +418,9 @@ fn main() -> ExitCode {
         loop_start,
         || {
             Ok(
-                relayer_restart_invariants_met()? && relayer_reorg_handling_invariants_met()?,
+                relayer_restart_invariants_met()?
+                    && relayer_reorg_handling_invariants_met()?
+                    && relayer_cached_metadata_invariant_met()?,
                 // TODO: fix and uncomment
                 // && relayer_cached_metadata_invariant_met()?
```
paulbalaji, line 425: remove old comments if fixed

solidity/contracts/test/TestSendReceiver.sol
```diff
@@ -15,6 +16,7 @@ contract TestSendReceiver is IMessageRecipient {
     using TypeCasts for address;
 
     uint256 public constant HANDLE_GAS_AMOUNT = 50_000;
+    uint256 public constant BODY_OFFSET = 77;
```
tkporter, line 19: you can rm this now

solidity/contracts/test/TestSendReceiver.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity >=0.8.0;
 
 import {TypeCasts} from "../libs/TypeCasts.sol";
+import {Message} from "../libs/Message.sol";
```
tkporter, line 5: and you can rm now too

solidity/test/test/TestSendReceiver.t.sol
```diff
@@ -111,4 +111,23 @@ contract TestSendReceiverTest is Test {
...
+        );
+    }
+
+    function testHandle_withHardcodedBodyHex() public {
+        vm.expectRevert("failMessageBody");
+        testSendReceiver.handle(
+            0,
+            address(testSendReceiver).addressToBytes32(),
+            // failMessageBody but hex value
+            hex"6661696c4d657373616765426f6479"
```
tkporter, line 130: could you do
```
bytes memory hardcodedFail = "failMessageBody";
```
and use `hardcodedFail` as the param instead of the hex? just trying to avoid magic numbers

tkporter (approved):

solidity/test/test/TestSendReceiver.t.sol
kamiyaa: sure

PR #5845, opened 2025-04-04
feat: add support for Khalani solvers

nicosampler:
### Description
This PR is to add support for the Khalani network. 

### Drive-by changes

This PR adds a few changes:
SDK:
- A new TokenStandard was created for Khalani.
- WarpCore: When the token is Khalani, addressOrDenom and collateralAddressOrDenom are treated specially.
- New Adapters: EvmKhalaniIntentTokenAdapter and EvmKhalaniHypAdapter.

Widgets:
- Users are required to sign an EIP712 message. We have added a hook helper for that.

### Backward compatibility
yes

### Testing
Manual.

Nexus URL integrating this version of the SDK: TBD. 
Loom video: TBD.

### DIFF
--- a/typescript/sdk/package.json
+++ b/typescript/sdk/package.json
@@ -18,7 +18,7 @@
     "@safe-global/safe-deployments": "1.37.32",
     "@solana/spl-token": "^0.4.9",
     "@solana/web3.js": "^1.95.4",
-    "@tvl-labs/arcadia-sdk": "^0.0.3",
+    "@tvl-labs/arcadia-sdk": "^0.0.6",
     "bignumber.js": "^9.1.1",
     "compare-versions": "^6.1.1",
     "cosmjs-types": "^0.9.0",
--- a/typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
+++ b/typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
@@ -58,7 +58,7 @@ import {
   TransferParams,
   TransferRemoteParams,
 } from './ITokenAdapter.js';
-import { NetworkType } from '@tvl-labs/arcadia-sdk/config.js';
+import { NetworkType } from '@tvl-labs/arcadia-sdk/config';
 
 // An estimate of the gas amount for a typical EVM token router transferRemote transaction
 // Computed by estimating on a few different chains, taking the max, and then adding ~50% padding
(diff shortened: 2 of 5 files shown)


typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
```diff
@@ -762,3 +766,268 @@ export class EvmXERC20VSAdapter
     });
   }
 }
+
+export class EvmKhalaniIntentTokenAdapter
```
nicosampler, line 770: All the communication between HL and Khalani happens in this adapter.

Besides the `approve` and `deposit` TXs, Khalani creates a `refine` from the user inputs and returns an object with the bridge stats. This object is then signed by the user and used to create an intent, which is forwarded to the network of solvers.

typescript/sdk/src/token/adapters/ITokenAdapter.ts
```diff
@@ -53,6 +55,7 @@ export interface IHypTokenAdapter<Tx> extends ITokenAdapter<Tx> {
   quoteTransferRemoteGas(
     destination: Domain,
     sender?: Address,
+    amount?: string,
```
nicosampler, line 58: `quoteTransferRemoteGas` returns the amount the user will receive on the destination chain. In order to provide this information, Khalani needs to know the amount that will be sent.

typescript/widgets/src/walletIntegrations/ethereum.ts
```diff
@@ -159,6 +160,38 @@ export function useEthereumTransactionFns(
   return { sendTransaction: onSendTx, switchNetwork: onSwitchNetwork };
 }
 
+export function useEthereumSignTypedData(multiProvider: MultiProtocolProvider) {
```
nicosampler, line 163: This hook was added to allow users to sign an `EIP-712` message. It’s data-agnostic, so it can be reused if needed.

typescript/sdk/package.json
```diff
@@ -16,6 +16,7 @@
     "@safe-global/safe-deployments": "1.37.23",
     "@solana/spl-token": "^0.4.9",
     "@solana/web3.js": "^1.95.4",
+    "arcadia-sdk-wip": "^1.2.4",
```
nicosampler, line 19: To be replaced once the productive version is ready.

PR #5849, opened 2025-04-07
feat(agents): cosmosnative routing ism

yjamin:
### Description

Added new routing ism to the agents.

### Drive-by changes
Decrease provider count for `cosmosnative` by implementing the Drop trait.

### Related issues

[Linear Issue](https://linear.app/hyperlane-xyz/issue/BACK-127/cosmosnative-relayer-routing-ism-support)

### Backward compatibility
Full backwards compatibility.

### Testing
Added new `RoutingIsm` to the E2E `cosmosnative` tests.

### DIFF
--- a/rust/main/utils/run-locally/src/cosmosnative/cli.rs
+++ b/rust/main/utils/run-locally/src/cosmosnative/cli.rs
@@ -178,6 +178,22 @@ impl SimApp {
             "1",
         ]);
 
+        // create routing ism and configure it to use the merkle tree ism just created
+        // cmd is following: create-routing
+        // expected ism address: 0x726f757465725f69736d00000000000000000000000000010000000000000001
+        self.tx(vec!["hyperlane", "ism", "create-routing"]);
+
+        // configure the routing ism to use the merkle tree ism
+        // cmd is following: set-routing-ism-domain [routing-ism-id] [domain] [ism-id]
+        self.tx(vec![
+            "hyperlane",
+            "ism",
+            "set-routing-ism-domain",
+            "0x726f757465725f69736d00000000000000000000000000010000000000000001",
+            destination_domain,
+            "0x726f757465725f69736d00000000000000000000000000040000000000000000",
+        ]);
+
         // create mailbox
         // cmd is following: default-ism local-domain
         // expected mailbox address: 0x68797065726c616e650000000000000000000000000000000000000000000000
@@ -185,7 +201,7 @@ impl SimApp {
             "hyperlane",
             "mailbox",
             "create",
-            "0x726f757465725f69736d00000000000000000000000000040000000000000000",
+            "0x726f757465725f69736d00000000000000000000000000010000000000000001",
             local_domain,
         ]);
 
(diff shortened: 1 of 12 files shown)


rust/main/utils/run-locally/src/cosmosnative/cli.rs
```diff
@@ -178,14 +178,30 @@ impl SimApp {
...
+        // cmd is following: create-routing
+        // expected ism address: 0x726f757465725f69736d00000000000000000000000000010000000000000001
+        self.tx(vec!["hyperlane", "ism", "create-routing"]);
+
+        // configure the routing ism to use the merkle tree ism
+        // cmd is following: set-routing-ism-domain [routing-ism-id] [domain] [ism-id]
+        self.tx(vec![
+            "hyperlane",
+            "ism",
+            "set-routing-ism-domain",
```
yjamin, line 191: > [!WARNING]  
> The E2E Tests still point to a release where the `RoutingIsm` is not implemented. Will work, once we have a release with the `RoutingIsm` included.

rust/main/utils/run-locally/src/cosmosnative/cli.rs
```diff
@@ -178,14 +178,30 @@ impl SimApp {
             "1",
         ]);
 
+        // create routing ism and configure it to use the merkle tree ism just created
+        // cmd is following: create-routing
+        // expected ism address: 0x726f757465725f69736d00000000000000000000000000010000000000000001
+        self.tx(vec!["hyperlane", "ism", "create-routing"]);
```
daniel-savu, line 184: wondering why the routing ISM address is so similar to that of the merkle tree multisig ISM. Would've expected the hash used to generate the address to have more randomness? Or is it a common thing with cosmos sdk

daniel-savu: also would be nice to define some consts with these addresses, since they're hard to tell apart (for instance when setting the default ISM on the mailbox) and it'll only get harder as we add more "contracts"

daniel-savu (approved):

rust/main/utils/run-locally/src/cosmosnative/cli.rs
yjamin: They look so similar, because that Address is not a hash. It's a fixed format that we use for generation:
```
{prefixBytes}{moduleId}{sequence}
```

daniel-savu: until we get to that it's cleaner to hardcode them as consts rather than inline strings IMO

rust/main/utils/run-locally/src/cosmosnative/cli.rs
```diff
@@ -151,6 +151,9 @@ impl SimApp {
         // TODO: test against the tx result to see if everything was created correctly
         self.tx(vec!["hyperlane", "hooks", "igp", "create", DENOM]);
 
+        const IGP_ADDRESS: &str =
```
daniel-savu, line 154: nit: consts are usually defined at the top of the file (like `GENESIS_FUND`), so they're easy to export elsewhere if needed

daniel-savu (approved):

PR #5856, opened 2025-04-08
feat(ci): update dockerfile for depot caching improvements

paulbalaji:
### Description

chore(ci): docker improvements
- multiplatform monorepo/agent builds (no longer need to specify `--platform linux/amd64` when pulling an image)
- Dockerfile caching improvements: agent build with warm cache can now go as low as 2-3 mins (previous low was 5-6)
	- leverage https://docs.docker.com/build/building/multi-stage/#use-multi-stage-builds

Couldn't get cargo-chef to play nice but this still is an improvement i suppose. Shoutout to depot for the helpful docs: https://depot.dev/docs/container-builds/how-to-guides/optimal-dockerfiles/rust-dockerfile


rust/Dockerfile
```diff
@@ -1,65 +1,64 @@
...
 FROM ubuntu:22.04
-RUN apt-get update && \
-    apt-get install -y \
-        openssl \
-        ca-certificates \
-        tini && \
-    rm -rf /var/lib/apt/lists/*
-
 WORKDIR /app
-RUN mkdir -p /app/config
```
daniel-savu, line 54: why this this dir no longer being created?

rust/Dockerfile
```diff
@@ -1,65 +1,64 @@
...
+    rustup target add x86_64-unknown-linux-musl && \
+    cargo install --locked sccache
+
+ENV RUSTC_WRAPPER=sccache
+ENV SCCACHE_DIR=/sccache
 
-RUN mkdir -p rust/main
-RUN mkdir -p rust/sealevel
+# -------- Builder Stage --------
+FROM base AS builder
```
daniel-savu, line 14: what's this?

paulbalaji: previously we just had the build step and then the runtime image that gets created. now we're separating into a `base` image, and then later a `builder` image which builds from this base image. This is so that we have better separation/caching of the parts that don't change frequently (toolchain & setup) vs things that are more likely to change (the actual rust compilation)

daniel-savu: oh i see it here: `FROM rust:1.80.1 AS base`

that's sick, seeing here: https://docs.docker.com/build/building/multi-stage/#use-multi-stage-builds

but yeah as Mo mentioned, would be good to add some docs in the PR descriptions since none of us are docker experts so it's hard to tell what's going on

paulbalaji: added comments throughout the file to explain what it's doing better - also documented existing steps for future readers

daniel-savu (approved):

PR #5859, opened 2025-04-08
fix: skip balance alert test if grafana api key cannot be read from secrets

Mo-Hussain:
### Description
- Skip balance alert test if grafana api key cannot be read from secrets
- This will allow CI to pass for external PRs

### Testing

Manual

### DIFF
--- a/typescript/infra/test/balance-alerts.test.ts
+++ b/typescript/infra/test/balance-alerts.test.ts
@@ -19,8 +19,17 @@ const DEFAULT_TIMEOUT = 30_000;
 describe('Balance Alert Thresholds', async function () {
   this.timeout(DEFAULT_TIMEOUT);
 
-  it('should have matching thresholds between Grafana alerts and threshold config files', async () => {
-    const saToken = await fetchGrafanaServiceAccountToken();
+  it('should have matching thresholds between Grafana alerts and threshold config files', async function () {
+    let saToken: string;
+    try {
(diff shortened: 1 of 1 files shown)


typescript/infra/test/balance-alerts.test.ts
```diff
@@ -19,8 +19,17 @@ const DEFAULT_TIMEOUT = 30_000;
...
+    let saToken: string;
+    try {
+      saToken = await fetchGrafanaServiceAccountToken();
+    } catch (error) {
+      console.log(
+        'Error fetching grafana service account token, skipping test',
+        error,
+      );
+      this.skip();
+    }
```
paulbalaji, line 32: try `safelyAccessEnvVar` from utils

Mo-Hussain: I do think it's cleaner if https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/c18510746e5342cefb8bdac67b208c0a79897b1d/typescript/infra/src/infrastructure/monitoring/grafana.ts#L65 returns a `string` instead of `string | undefined` so unless you have very strong opinions on this, I'll keep it as is

paulbalaji: ah ye i forgot that had a ci if case

paulbalaji (approved):

PR #5862, opened 2025-04-08
test: add cosmos-sdk e2e tests

troykessler:
### Description

This PR adds e2e tests for the cosmos-sdk. It spawns a local cosmos chain inside a docker container with just the Hyperlane Module installed while transactions are submitted with the cosmos-sdk. It is asserted if the transactions succeed and if there are no errors.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

e2e testing

### DIFF
--- a/typescript/cosmos-sdk/scripts/run-e2e-test.sh
+++ b/typescript/cosmos-sdk/scripts/run-e2e-test.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+function cleanup() {
+  docker compose down
+}
+
+cleanup
+
+echo "Preparing E2E tests"
+docker compose up --detach --wait
+
+if [[ $? -ne 0 ]]; then
+  echo "Failure starting local cosmos chain"
+  exit 1
+fi
+
+echo "Running E2E tests"
+yarn mocha --config .mocharc-e2e.json
+
+cleanup
+
+echo "Completed E2E tests"
--- a/typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
+++ b/typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
@@ -1,0 +1,166 @@
+import { expect } from 'chai';
+import { step } from 'mocha-steps';
+
+import {
+  MerkleRootMultisigISM,
+  MessageIdMultisigISM,
+} from '../../../cosmos-types/dist/types/hyperlane/core/interchain_security/v1/types.js';
+import {
+  bytes32ToAddress,
+  isValidAddressEvm,
+} from '../../../utils/dist/addresses.js';
+import { SigningHyperlaneModuleClient } from '../index.js';
+
+import { createSigner } from './utils.js';
+
+describe('1. cosmos sdk interchain security e2e tests', async function () {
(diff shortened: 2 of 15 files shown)


typescript/cosmos-sdk/scripts/run-e2e-test.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+function cleanup() {
+  docker compose down
+}
```
xeno097, line 5: This is not something that you need to do in this PR, but it might be worth exploring later. This allows to control docker containers from nodejs https://node.testcontainers.org/

typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
```diff
@@ -0,0 +1,166 @@
...
+    expect(isms.isms).to.have.lengthOf(1);
+
+    const threshold = 2;
+    const validators = [
+      '0x3C24F29fa75869A1C9D19d9d6589Aae0B5227c3c',
+      '0xf719b4CC64d0E3a380e52c2720Abab13835F6d9c',
+      '0x98A56EdE1d6Dd386216DA8217D9ac1d2EE7c27c7',
+    ];
+
+    // note that the validators need to be sorted alphabetically
```
xeno097, line 70: What happens if they are not sorted?

xeno097 (commented):
Are there any not so happy paths that we can test? like incorrect arguments passed to the contracts or expected errors when miscondifguring something?

typescript/cosmos-sdk/scripts/run-e2e-test.sh
troykessler: nice, will try implement it

typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
troykessler: this is a requirement from the chain module side, if we would submit them not sorted the tx would fail. Testing such a thing would be more on the chain side instead of the TS sdk side

troykessler:
> Are there any not so happy paths that we can test? like incorrect arguments passed to the contracts or expected errors when miscondifguring something?

So my idea for the cosmos-sdk e2e tests were that I expect that the cosmos chain module works and I only test if the TypeScript SDK is able to submit the txs without issues to the chain. Can still add a few misconfigurations and see if they are handled correctly, but would not go any further since any logic testing would be expected on the chain side in my opinion

xeno097:
> > Are there any not so happy paths that we can test? like incorrect arguments passed to the contracts or expected errors when miscondifguring something?
> 
> So my idea for the cosmos-sdk e2e tests were that I expect that the cosmos chain module works and I only test if the TypeScript SDK is able to submit the txs without issues to the chain. Can still add a few misconfigurations and see if they are handled correctly, but would not go any further since any logic testing would be expected on the chain side in my opinion

fair enough

xeno097 (approved):
I'll approve but would love some one else to also have a look

paulbalaji (approved):

PR #5865, opened 2025-04-08
fix: corect synthetic token types

antigremlin:
### Description

This PR fixes a couple mistakes in token config type definitions. Some of the synthetic token types didn't match to config schemas correctly.

### Backward compatibility

Incompatibilities should be caught by static typing but we can't be sure it doesn't cause breakage in any dependent code.

### Testing

The existing test suite wasn't extended to expose the bugs.


yorhodes (approved):
nice catch

PR #5868, opened 2025-04-08
feat(submitter): subtask spawner + sealevel e2e

daniel-savu:
### Description

- In the SVM adapter, if `submit` is called on a tx that already has a `hash`, the submission is skipped. This is because in solana we shouldn't resubmit unless the hash was dropped, and that would be detected when status is queried within the inclusion stage, causing the tx to be dropped, payloads sent back to the building queue, and a new tx built. This can be optimized but there was no obvious simple way to fix for now. If we do submit again, a new hash would be generated, creating a race condition between the old tx and the new tx. In e2e I noticed that the old tx always lands and the new tx reverts, though the submitter is unaware of it because its `hash` field gets set to the new one, so it keeps the tx stuck in the inclusion stage instead of promoting it to the finality stage.
- Adds PayloadDispatcher `spawn` method, which initializes and connects all dispatcher components
- Enables the Lander in sealevel e2e for both chains
- Fixes several bugs in the SVM adapter that were causing e2e to fail
- Adds unit testing for the PayloadDispatcherEntrypoint, which found a bug in the DbLoader
- Adds an integration test that runs the whole Dispatcher (from entrypoint to finalization), with a mocked adapter
- Adds logs to all stages of the dispatcher, to easily track payloads and transactions
- Allows configuring `PayloadDispatcherSettings` with a DB instance rather than a path. This was to fix a bug where the same path cannot be used by two separate DB instances (the MessageProcessor one and the Dispatcher one), as an OS lock is acquired on that path by the first DB

### Drive-by changes

- comments out the `gas_expenditure_log_count` e2e invariant because it is specific to the old submitter
- adds more logs to the e2e invariant checker, to have a better view of where things fail

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5824

### Backward compatibility

Yes, but changes to the old sealevel submitter won't be covered in e2e

### Testing

Unit testing and e2e.

Current unit testing coverage report (the columns are Function Coverage | Line Coverage | Region Coverage)
Screenshot 2025-04-09 at 17 15 37

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -34,7 +34,7 @@ use hyperlane_core::{
     QueueOperation, SubmitterType, ValidatorAnnounce, H512, U256,
 };
 use hyperlane_operation_verifier::ApplicationOperationVerifier;
-use submitter::{PayloadDispatcherEntrypoint, PayloadDispatcherSettings};
+use submitter::{DatabaseOrPath, PayloadDispatcherEntrypoint, PayloadDispatcherSettings};
 
 use crate::{
     merkle_tree::builder::MerkleTreeBuilder,
@@ -166,6 +166,7 @@ impl BaseAgent for Relayer {
             &settings,
             core_metrics.clone(),
             &chain_metrics,
+            db.clone(),
         )
         .await;
 
@@ -774,6 +775,7 @@ impl Relayer {
         settings: &RelayerSettings,
         core_metrics: Arc<CoreMetrics>,
         chain_metrics: &ChainMetrics,
+        db: DB,
     ) -> HashMap<HyperlaneDomain, PayloadDispatcherEntrypoint> {
         settings.destination_chains.iter()
             .filter(|chain| SubmitterType::Lander == settings.chains[&chain.to_string()].submitter)
@@ -781,7 +783,7 @@ impl Relayer {
                 chain_conf: settings.chains[&chain.to_string()].clone(),
                 raw_chain_conf: Default::default(),
                 domain: chain.clone(),
-                db_path: settings.db.clone(),
+                db: DatabaseOrPath::Database(db.clone()),
                 metrics: core_metrics.clone(),
             }))
             .map(|(chain, s)| (chain, PayloadDispatcherEntrypoint::try_from_settings(s)))
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -165,6 +165,7 @@ impl BaseAgent for Relayer {
             Self::build_validator_announces(&settings, &core_metrics, &chain_metrics).await;
         debug!(elapsed = ?start_entity_init.elapsed(), event = "initialized validator announces", "Relayer startup duration measurement");
 
+        start_entity_init = Instant::now();
         let dispatcher_entrypoints = Self::build_payload_dispatcher_entrypoints(
             &settings,
             core_metrics.clone(),
@@ -172,7 +173,9 @@ impl BaseAgent for Relayer {
             db.clone(),
         )
         .await;
+        debug!(elapsed = ?start_entity_init.elapsed(), event = "initialized dispatcher entrypoints", "Relayer startup duration measurement");
 
+        start_entity_init = Instant::now();
         let dispatchers = Self::build_payload_dispatchers(
             &settings,
             core_metrics.clone(),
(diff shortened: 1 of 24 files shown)


rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
```diff
@@ -267,6 +270,9 @@ impl AdaptsChain for SealevelTxAdapter {
     }
 
     async fn submit(&self, tx: &mut Transaction) -> Result<(), SubmitterError> {
+        if tx.hash.is_some() {
+            return Ok(());
+        }
```
daniel-savu, line 275: this is the most consequential change, see the PR description as well:
> In the SVM adapter, if submit is called on a tx that already has a hash, the submission is skipped. This is because in solana we shouldn't resubmit unless the hash was dropped, and that would be detected when status is queried within the inclusion stage, causing the tx to be dropped, payloads sent back to the building queue, and a new tx built. This can be optimized but there was no obvious simple way to fix for now

rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
```diff
@@ -292,43 +298,26 @@ impl AdaptsChain for SealevelTxAdapter {
...
-                    err
-                )
-            })
-            .unwrap_or(false);
-
-        info!(?tx, "confirmed transaction with commitment level processed");
-
-        if !executed {
-            return Err(SubmitterError::TxSubmissionError(
-                "Process transaction is not confirmed with commitment level processed".to_string(),
```
daniel-savu, line 311: this step would always return an error in e2e, causing txs to be dropped. It's better to rely on querying the status via the txhash imo

rust/main/utils/run-locally/src/sealevel/mod.rs
```diff
@@ -98,6 +98,8 @@ fn run_locally() {
         .hyp_env("DB", relayer_db.to_str().unwrap())
         .hyp_env("CHAINS_SEALEVELTEST1_SIGNER_KEY", RELAYER_KEYS[0])
         .hyp_env("CHAINS_SEALEVELTEST2_SIGNER_KEY", RELAYER_KEYS[1])
+        .hyp_env("CHAINS_SEALEVELTEST1_SUBMITTER", "Lander")
+        .hyp_env("CHAINS_SEALEVELTEST2_SUBMITTER", "Lander")
```
ameten, line 102: I wonder if it makes sense to test situation when one chain uses `Lander` submitter while the other one `Classic`. Probably, not that easy to test since we are sending messages in one direction in E2E tests. And it does not add a lot of value since `Lander` and `Classic` submitter are independent. Only value is that we keep making sure that we did not break `Classic` submitter by implementing `Lander` submitter.

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -166,6 +169,15 @@ impl BaseAgent for Relayer {
...
+        )
+        .await;
+
+        let dispatchers = Self::build_payload_dispatchers(
+            &settings,
+            core_metrics.clone(),
+            &chain_metrics,
+            db.clone(),
         )
         .await;
```
ameten, line 182: Let's have a debug log message about how much time we spend initialising `dispatchers` as we do for other entities (mailboxes, validator announces, etc)

rust/main/submitter/src/payload_dispatcher/dispatcher.rs
```diff
@@ -42,12 +57,89 @@ impl PayloadDispatcher {
...
+            tokio::sync::mpsc::channel::<Transaction>(SUBMITTER_CHANNEL_SIZE);
+        let (finality_stage_sender, finality_stage_receiver) =
+            tokio::sync::mpsc::channel::<Transaction>(SUBMITTER_CHANNEL_SIZE);
+
+        let building_stage = BuildingStage::new(
+            building_stage_queue.clone(),
+            inclusion_stage_sender.clone(),
+            self.inner.clone(),
+        );
+        let building_task = tokio::spawn(
```
ameten, line 73: It would be better to use `tokio::task::Builder::new().name("<name>")` so that each task has a name and we can identify it in Tokio Console.

ameten (approved):

rust/main/utils/run-locally/src/sealevel/mod.rs
daniel-savu: non-svm e2e still uses classic, so that goes to show that the `Lander` integration didn't break anything. For SVM specifically, we'll need to remember to update e2e to use `Classic` if any changes to that submitter are made

PR #5877, opened 2025-04-09
feat: higher s3 DoS resilience

yjamin:
### Description

Contains following changes:
- New `MetadataBuildError` when the MultiSig contains more than `50` validators
- Fetch metadata in S3 buckets and check for size constraints first (50KiB)
- Stricter S3 Request Timeout when fetching objects (30s -> 10s)
- Parallelized checkpoint fetching logic:
  - Use `1-10` workers to fetch signatures for one checkpoint index

### Drive-by changes
None

### Related issues
[Github](https://github.com/hyperlane-xyz/issues/issues/401) - [Linear](https://linear.app/hyperlane-xyz/issue/BACK-130/pre-tge-fixes-for-known-dos-vectors-are-documented-known)
Also includes https://linear.app/hyperlane-xyz/issue/BACK-146/make-multisigcheckpointsyncerget-validator-latest-checkpoints-and

### Testing
Wrote a unit test for ensuring the logic stays the same. Can be executed with
```
cargo test --package hyperlane-base --lib -- types::multisig::test::test_s3_checkpoint_syncer --exact --show-output
```


rust/main/hyperlane-base/src/types/multisig.rs
```diff
@@ -129,11 +139,27 @@ impl MultisigCheckpointSyncer {
...
-                if let Ok(Some(checkpoint)) =
-                    self.fetch_checkpoint(validators, threshold, index).await
-                {
-                    return Ok(Some(checkpoint));
+
+            // Batch the fetching of a specific index
+            // This enables us to fetch a quorum faster
+            let batch_size = ((start_index - minimum_index) / 4) as usize;
+            let batch_size = batch_size.clamp(1, 3);
+
```
yjamin, line 147: What is the avg distance between `minimum_index` & `start_index`? I'd expect that to always be small, except when we catch up to live height.

rust/main/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -21,7 +24,9 @@ use crate::{
 /// The timeout for S3 requests. Rusoto doesn't offer timeout configuration
 /// out of the box, so S3 requests must be wrapped with a timeout.
 /// See https://github.com/rusoto/rusoto/issues/1795.
-const S3_REQUEST_TIMEOUT_SECONDS: u64 = 30;
+const S3_REQUEST_TIMEOUT_SECONDS: u64 = 10;
+const S3_METADATA_TIMEOUT_SECONDS: u64 = 3;
+const S3_MAX_OBJECT_SIZE: i64 = 1024 * 1024; // 1MiB
```
yjamin, line 30: Valid Checkpoints are quite limit in the size they can be - should we make this more strict? I've seen many around the 1KiB size so with 1MiB we have >1000x threshold

rust/main/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -69,8 +74,40 @@ impl S3Storage {
...
+            self.anonymous_client().head_object(metadata_req),
+        )
+        .await?;
+
+        match get_object_result {
+            Ok(value) => match value.content_length {
+                Some(length) if length >= S3_MAX_OBJECT_SIZE => {
+                    bail!("Object size for key {key} is too big: {}KiB", length / 1024);
+                }
+                _ => Ok(true),
```
yjamin, line 97: Should we assume that valid objects always have `content_length` set? Right now we assume that object is valid if it has no `Content-Length` even though it might be greater than our size restriction.

rust/main/hyperlane-base/src/types/multisig.rs
```diff
@@ -216,12 +268,131 @@ impl MultisigCheckpointSyncer {
...
+    use hyperlane_core::KnownHyperlaneDomain;
+    use rusoto_core::Region;
+
+    use crate::S3Storage;
+
+    use super::*;
+
+    #[tokio::test]
+    #[tracing_test::traced_test]
+    async fn test_s3_checkpoint_syncer() {
```
yjamin, line 292: These validators are taken from Arbitrum Mainnet. I'm unsure if we want to have those storage locations hard-coded in tests, maybe we remove it again?

tkporter (commented):
With the new S3 client coming up, wondering if it makes sense to block this on the new client? Sorry to create more work for you there

rust/main/hyperlane-base/src/types/multisig.rs
yjamin: Yes, I'm good with that. I agree, we would make a lot of concurrent S3 requests when this number is high and I don't expect this to be the bottleneck in normal operations. Will remove the batching for this method 👍

rust/main/hyperlane-base/src/types/multisig.rs
yjamin: Yes I agree! Also this test serves more as a performance tests anyways.

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -33,6 +33,8 @@ pub enum MetadataBuildError {
     MaxIsmDepthExceeded(u32),
     #[error("Exceeded max count when building metadata ({0})")]
     MaxIsmCountReached(u32),
+    #[error("Exceeded max validator count when building metadata ({0})")]
+    MaxValidatorCountReached(u32),
```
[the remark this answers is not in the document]
tkporter: imo it's ok for now

rust/main/hyperlane-base/src/types/multisig.rs
```diff
@@ -31,42 +30,54 @@ impl MultisigCheckpointSyncer {
...
+                if let Some(checkpoint_syncer) = self.checkpoint_syncers.get(&v) {
+                    Some((v, checkpoint_syncer))
+                } else {
+                    warn!(validator=%v, "Checkpoint syncer is not provided for validator");
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let futures = syncer
```
tkporter, line 53: I don't feel strongly but just noting that this doesn't limit the concurrency, wondering if this was intentional. To be clear I'm happy as it's written, just curious

tkporter (approved):
🫡

rust/main/hyperlane-base/src/types/multisig.rs
yjamin: At the time of writing this code I wasn't aware of a nice way to limit the number of concurrency. I didn't want to use the same chunking logic, because we'd need to await all futures anyway and can't short circuit - which was the main benefit of that approach. 

Since then I learned about `futures::stream` and its [buffere_unordered](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered) function, which fits perfectly. I'll use that instead and limit the concurrency to 10.

PR #5878, opened 2025-04-09
feat: Add MovableCollateralRouter

larryob:
### Description
Add abstract contract that allows a router's collateral to be moved.

### Drive-by changes

N/A

### Related issues

N/A

### Backward compatibility

Yes
### Testing

Unit

### DIFF
--- a/solidity/contracts/token/libs/MovableCollateralRouter.sol
+++ b/solidity/contracts/token/libs/MovableCollateralRouter.sol
@@ -36,12 +36,12 @@ abstract contract MovableCollateralRouter is AccessControl {
 
     function rebalance(
         uint32 domain,
-        bytes32 recipient,
         uint256 amount,
         ValueTransferBridge bridge
     ) external payable onlyRole(REBALANCER_ROLE) {
         address rebalancer = _msgSender();
-        if (allowedDestinations[domain] != recipient) {
+        bytes32 recipient = allowedDestinations[domain];
+        if (recipient == bytes32(0)) {
             revert BadDestination({rebalancer: rebalancer, domain: domain});
         }
 
--- a/solidity/contracts/token/libs/MovableCollateralRouter.sol
+++ b/solidity/contracts/token/libs/MovableCollateralRouter.sol
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.8.0;
+
+import {FungibleTokenRouter} from "./FungibleTokenRouter.sol";
+import {ValueTransferBridge} from "./ValueTransferBridge.sol";
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+import {Context} from "@openzeppelin/contracts/utils/Context.sol";
+import {ContextUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
+
+abstract contract MovableCollateralRouter is AccessControl {
+    constructor() {
+        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+    }
+
+    bytes32 public constant REBALANCER_ROLE = keccak256("REBALANCER_ROLE");
+
+    mapping(uint32 destinationDomain => bytes32 recipient)
+        public allowedDestinations;
+    mapping(uint32 destinationDomain => mapping(ValueTransferBridge bridge => bool isValidBridge))
+        public allowedBridges;
+
+    event CollateralMoved(
+        uint32 indexed domain,
+        bytes32 recipient,
+        uint256 amount,
+        address indexed rebalancer
+    );
+
+    error BadDestination(address rebalancer, uint32 domain);
+    error BadBridge(address rebalancer, ValueTransferBridge bridge);
+
+    function moveCollateral(
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount,
+        ValueTransferBridge bridge
+    ) external onlyRole(REBALANCER_ROLE) {
+        address rebalancer = _msgSender();
+        if (allowedDestinations[domain] != recipient) {
+            revert BadDestination({rebalancer: rebalancer, domain: domain});
+        }
+
+        if (!(allowedBridges[domain][bridge])) {
+            revert BadBridge({rebalancer: rebalancer, bridge: bridge});
+        }
+
+        _moveCollateral(domain, recipient, amount, bridge);
+        emit CollateralMoved({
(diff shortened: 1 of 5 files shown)


solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -0,0 +1,129 @@
...
+    function moveCollateral(
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount
+    ) external payable onlyRole(REBALANCER_ROLE) {
+        _moveCollateral(
+            domain,
+            recipient,
+            amount,
+            ValueTransferBridge(address(this))
```
yorhodes, line 48: interesting

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -0,0 +1,129 @@
...
+        uint256 _scale,
+        address _mailbox
+    ) FungibleTokenRouter(_scale, _mailbox) {
+        _grantRole(DEFAULT_ADMIN_ROLE, owner());
+    }
+
+    bytes32 public constant REBALANCER_ROLE = keccak256("REBALANCER_ROLE");
+
+    mapping(uint32 destinationDomain => bytes32 recipient)
+        public allowedDestinations;
```
yorhodes, line 25: should we allow multiple destinations if we allow multiple bridges?
I am still partial to reusing the router table for this to narrow the rebalancing interface

yorhodes (commented):
are you working on tests still?

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: Resolving since we discussed this offline.

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: Yeah, but I'll remove it since we could always just call `addBridge` on the router and pass in its own address.

larryob:
> are you working on tests still?

Yep, following up with tests. I just wanted to get a little demo version out first and get some eyes on it.

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -0,0 +1,83 @@
...
+        uint32 indexed domain,
+        bytes32 recipient,
+        uint256 amount,
+        address indexed rebalancer
+    );
+
+    error BadDestination(address rebalancer, uint32 domain);
+    error BadBridge(address rebalancer, ValueTransferBridge bridge);
+
+    function moveCollateral(
```
yorhodes, line 33: move collateral sounds a little scary
should we rename to rebalance?

yorhodes: still hesitant to allow arbitrary recipients here, feels like a footgun

solidity/contracts/token/libs/ValueTransferBridge.sol
yorhodes, line 1: prefer interface if the implementations are empty?

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -0,0 +1,83 @@
...
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount,
+        ValueTransferBridge bridge
+    ) internal virtual {
+        bridge.transferRemote({
+            destinationDomain: domain,
+            recipient: recipient,
+            amountOut: amount
+        });
```
yorhodes, line 67: how does the rebalancer get the quote for the bridge?

yorhodes: I expect we will want to pass `msg.value`

solidity/contracts/token/libs/ValueTransferBridge.sol
```diff
@@ -0,0 +1,21 @@
...
+    address token;
+    uint256 amount;
+}
+
+abstract contract ValueTransferBridge {
+    function quoteTransferRemote(
+        uint32 destinationDomain,
+        bytes32 recipient,
+        uint amountOut
+    ) public view virtual returns (Quote[] memory) {}
```
yorhodes, line 14: still think memory arrays are pretty annoying to work with and gas inefficient
what bridge implementations have multiple fees in different denominations? I'm personally not aware of one

yorhodes (commented):
please add a changeset

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: Done!

solidity/contracts/token/libs/ValueTransferBridge.sol
larryob: I think interfaces are more for external integrators. We can just auto-generate interfaces for integrators and use abstract contracts for base classes.

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: Fixed. I assume the rebalancer would call `quoteTransferRemote` on the bridge.

solidity/contracts/token/libs/ValueTransferBridge.sol
larryob: Yeah, probably not too useful at the moment. I think we just want to give bridges the flexibility of charging multiple tokens (e.g. charge fees in the token being bridged).

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    error BadDestination(address rebalancer, uint32 domain);
+    error BadBridge(address rebalancer, ValueTransferBridge bridge);
+
+    function rebalance(
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount,
+        ValueTransferBridge bridge
+    ) external payable onlyRole(REBALANCER_ROLE) {
```
yorhodes, line 42: should we/could we add a version that doesnt take a recipient? as I expect most initial usage patterns to do

solidity/contracts/token/libs/ValueTransferBridge.sol
yorhodes: what if someone forgets to implement an override? we presumably dont want these to be empty

yorhodes (approved):
approving but you still need to add a changeset

PR #5881, opened 2025-04-09
feat(core): use safeApprove in HypXERC20Lockbox.approveLockbox()

paulbalaji:
### Description

Use SafeERC20.safeApprove instead of IERC20.approve to approve expenditure of the wrapped token in HypXERC20Lockbox.approveLockbox(). Useful for cases such as USDT on ethereum whose `approve()` method is non-standard to regular ERC20.

### Testing

ousdt staging route

### DIFF
--- a/.changeset/early-cooks-wink.md
+++ b/.changeset/early-cooks-wink.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/core': patch
+---
+
+Use SafeERC20.safeApprove instead of IERC20.approve in HypXERC20Lockbox.approveLockbox()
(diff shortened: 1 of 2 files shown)


solidity/contracts/token/extensions/HypXERC20Lockbox.sol
```diff
@@ -33,10 +36,7 @@ contract HypXERC20Lockbox is HypERC20Collateral {
...
      */
     function approveLockbox() public {
-        require(
-            IERC20(wrappedToken).approve(address(lockbox), MAX_INT),
-            "erc20 lockbox approve failed"
-        );
+        IERC20(wrappedToken).safeApprove(address(lockbox), MAX_INT);
         require(
             xERC20.approve(address(lockbox), MAX_INT),
             "xerc20 lockbox approve failed"
```
yorhodes, line 42: I think we should migrate both of these

.changeset/early-cooks-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/core': minor
+---
```
yorhodes, line 3: I might argue this is patch

yorhodes (approved):

PR #5883, opened 2025-04-09
feat [DO NOT MERGE YET]: warp fork command

xeno097:
### Description

PR opened for tracking purposes at the moment

### DIFF
--- a/typescript/cli/src/fork/fork.ts
+++ b/typescript/cli/src/fork/fork.ts
@@ -156,8 +156,8 @@ async function forkChain(
       await killAnvilProcess(false);
     }
   } catch (error) {
-    // Kill any anvil process as if not, the process will keep running
-    // in the background
+    // Kill any running anvil process otherwise the process will keep running
+    // in the background.
     if (killAnvilProcess) {
       await killAnvilProcess(true);
     }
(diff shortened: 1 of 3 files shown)


typescript/cli/src/fork/fork.ts
```diff
@@ -0,0 +1,302 @@
...
+): Promise<void> {
+  if (accountsToImpersonate.length === 0) {
+    return;
+  }
+
+  logGray(
+    `Impersonating accounts ${accountsToImpersonate} on chain ${chainName}`,
+  );
+  await Promise.all(
+    accountsToImpersonate.map((address) =>
```
ltyu, line 182: as discussed, would be an improvement to simplify the config by removing this. we would just preprocess the `transactions.from` into this array if batch impersonations are absolutely necessary.

ltyu:
pretty good progress. some thoughts for improvement:

- I was hoping I can just give it my submitter artifacts (gnosisSafeTxBuilder) as is.
- I found it challenging to configure the `fork-config` mostly because I don't understand the fields entirely. Hope we can reduce this config, or ideally not have it. (e.g. derive the owner (used to populate `transactions.from`), etc.)

ltyu (approved):

PR #5884, opened 2025-04-09
feat: alert if merkle root from validators differ from relayer internal root

kamiyaa:
### Description

 - add metrics to indicate merkle root mismatch

### Related issues

https://linear.app/hyperlane-xyz/issue/BACK-72/pre-tge-relayer-should-high-urgency-alert-if-a-known-app-context-has-a

### Backward compatibility

Yes/No

### Testing
None so far

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
@@ -153,7 +153,17 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
         // If any inner ISMs are refusing to build metadata, we propagate just the first refusal.
         for sub_module_res in sub_modules_and_metas.iter() {
             if let Err(MetadataBuildError::Refused(s)) = sub_module_res {
-                return Err(MetadataBuildError::Refused(s.to_string()));
+                return Err(MetadataBuildError::Refused(s.clone()));
+            }
+            if let Err(MetadataBuildError::MerkleRootMismatch {
+                root,
+                canonical_root,
+            }) = sub_module_res
+            {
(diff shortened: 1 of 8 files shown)


rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -199,10 +199,24 @@ impl MetadataBuilder for AggregationIsmMetadataBuilder {
...
+        // we propagate just the first refusal.
         for sub_module_res in sub_modules_and_metas.iter() {
+            // This occurs when there is a reorg
             if let Err(MetadataBuildError::Refused(s)) = sub_module_res {
-                return Err(MetadataBuildError::Refused(s.to_string()));
+                return Err(MetadataBuildError::Refused(s.clone()));
+            }
+            // This likely occurs when there is a reorg as well,
+            // but it could also be because validators are colluding
+            if let Err(MetadataBuildError::MerkleRootMismatch {
```
kamiyaa, line 211: I'm assuming we want to stop building the metadata if there are any indication of a merkle root mismatch.

rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -1012,6 +1012,20 @@ impl PendingMessage {
...
+                MetadataBuildError::MerkleRootMismatch {
+                    root,
+                    canonical_root,
+                } => {
+                    warn!(
+                        checkpoint_root=?root,
+                        ?canonical_root,
+                        "Checkpoint root does not match canonical root from merkle proof"
+                    );
+                    self.ctx
```
kamiyaa, line 1024: top-level logic to `set_merkle_root_mismatch` metrics

PR #5885, opened 2025-04-09
feat: configurable default ISM caching policy

tkporter:
### Description

- Allows you to configure a different caching policy when a default ISM is used by a recipient. Config includes:
  - The caching policy, `MessageSpecific` or `IsmSpecific`
  - Module types to cache in this way
  - Chains to apply this to
- The initial configuration is to just `IsmSpecific` cache aggregation and multisig ISMs on EVM chains. SVM is explicitly not included because multisig ISMs include routing functionality internal to it. Other VMs are less of a concern atm so to be conservative they've been omitted. The practical impact of the IsmSpecific caching policy is to allow for cache hits for certain ISM configurations regardless of the message that's passed into the ISM config getter
- I definitely believe there's a more expressive config that we'll want to use eventually, but with the primary motivations laid out in https://linear.app/hyperlane-xyz/issue/BACK-82/caching-ism-configurations-in-the-relayer I think this is satisfactory to get us in a good spot for the upcoming traffic
- Tested on testnet4 rc - see hits here for the `module_and_threshold` and `validators_and_threshold`!
Screen Shot 2025-04-10 at 12 32 15 PM

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/BACK-144/pre-tge-more-generously-cache-certain-isms-based-off-configuration


typescript/infra/config/environments/testnet4/agent.ts
```diff
@@ -225,6 +228,20 @@ const scraperResources = {
...
+  moduleTypes: [
+    ModuleType.AGGREGATION,
+    ModuleType.MERKLE_ROOT_MULTISIG,
+    ModuleType.MESSAGE_ID_MULTISIG,
+  ],
+  // SVM is explicitly not cached as the default ISM is a multisig ISM
+  // that routes internally.
+  chains: ethereumChainNames,
+  cachePolicy: IsmCachePolicy.IsmSpecific,
+};
```
paulbalaji, line 243: do we anticipate splitting the strategies for mainnet3/testnet4? if not could we move this config somewhere more central

typescript/sdk/src/metadata/agentConfig.ts
```diff
@@ -398,6 +420,12 @@ export const RelayerAgentConfigSchema = AgentConfigSchema.extend({
     .describe(
       'A list of app contexts and their matching lists to use for metrics. A message will be classified as the first matching app context.',
     ),
+  defaultIsmCacheConfig: z
+    .union([IsmCacheConfigSchema, z.string().min(1)])
```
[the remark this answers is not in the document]
tkporter: I'm following the pattern for other places we've taken in JSON config - the idea is you can either pass in a stringified JSON (e.g. via an env var) or the JSON structure itself (e.g. via a config file)

typescript/infra/config/environments/testnet4/agent.ts
tkporter: I think for now the flexibility will be nice just in case there are issues in a pinch around the chains in here

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -187,3 +192,189 @@ impl AppContextClassifier {
...
 }
+
+/// An ISM caching policy.
+#[derive(Copy, Clone, Debug, PartialEq, Default, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum IsmCachePolicy {
+    /// Default cache policy, includes the message in the cache key
+    /// when querying config that may be message-specific.
+    #[default]
+    MessageSpecific,
```
daniel-savu, line 203: worth clarifying in a comment that `MessageSpecific` is the default because it makes the fewest assumptions about the mutability of the ISM result

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -187,3 +192,189 @@ impl AppContextClassifier {
...
+}
+
+/// Configuration for ISM caching behavior.
+/// Fields are renamed to be all lowercase / without underscores to match
+/// the format expected by the settings parsing.
+#[derive(Debug, Clone, Default, Deserialize)]
+pub struct IsmCacheConfig {
+    #[serde(deserialize_with = "deserialize_module_types", rename = "moduletypes")]
+    module_types: HashSet<ModuleType>,
+    chains: Option<HashSet<String>>,
```
daniel-savu, line 216: is there a noticeable reduction in RPC calls on chain where this was enabled and had a good hit rate?

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -187,3 +192,189 @@ impl AppContextClassifier {
...
+}
+
+impl IsmCacheConfig {
+    fn matches_chain(&self, domain_name: &str) -> bool {
+        if let Some(chains) = &self.chains {
+            chains.contains(domain_name)
+        } else {
+            // If no domains are specified, match all domains
+            true
+        }
```
daniel-savu, line 245: nit: I thought that when `self.chains` was `None` we wouldn't cache anything. Seems like we always cache, and `self.chains` is used for deciding which chains to apply `cache_policy` to. To wasn't obvious to me but now I see the
```
        if root_ism == default_ism
            && self.default_ism_cache_config.matches_chain(domain.name())
            && self
                .default_ism_cache_config
                .matches_module_type(ism_module_type)
```
In that same if statement we could add a chain-specific flag to enable/disable caching, and have it off by default?

ameten (approved):

typescript/infra/config/environments/testnet4/agent.ts
```diff
@@ -225,6 +228,20 @@ const scraperResources = {
   },
 };
 
+const defaultIsmCacheConfig: IsmCacheConfig = {
+  // Default ISM Routing ISMs change configs based off message content,
+  // so they are not specified here.
+  moduleTypes: [
+    ModuleType.AGGREGATION,
+    ModuleType.MERKLE_ROOT_MULTISIG,
+    ModuleType.MESSAGE_ID_MULTISIG,
```
paulbalaji, line 237: do we care about the weighted versions of these for now or nah?

rust/main/agents/relayer/src/settings/mod.rs
```diff
@@ -378,6 +388,43 @@ fn parse_matching_list(p: ValueParser) -> ConfigResult<MatchingList> {
     err.into_result(ml)
 }
 
+fn parse_json_object(p: ValueParser) -> Option<(ConfigPath, Value)> {
```
daniel-savu, line 391: we could eventually switch to this for parsing matching lists as well and remove the current complicated logic

typescript/infra/config/environments/testnet4/agent.ts
tkporter: Good q, atm we don't have the weighted functionality in the relayer yet so we don't care

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
```diff
@@ -105,14 +107,29 @@ pub trait MultisigIsmMetadataBuilder: AsRef<MessageMetadataBuilder> + Send + Syn
...
+                self.as_ref().root_ism,
+                multisig_ism.domain(),
+                self.module_type(),
+            )
+            .await
+        {
+            // To have the cache key be more succinct, we use the message id
+            IsmCachePolicy::MessageSpecific => (multisig_ism.address(), message.id()),
+            IsmCachePolicy::IsmSpecific => (multisig_ism.address(), H256::zero()),
+        };
```
daniel-savu, line 126: using either the (ism_address, message_id) or (ism_address, zero_id) as caching keys seems ISM agnostic and could be refactored out

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -187,3 +192,189 @@ impl AppContextClassifier {
...
+            .get_cache_policy(H256::repeat_byte(0xfe), &domain, ModuleType::Routing)
+            .await;
+        assert_eq!(cache_policy, IsmCachePolicy::MessageSpecific);
+
+        // Different domain, should not match
+        let domain = HyperlaneDomain::new_test_domain("barchain");
+        let cache_policy = classifier
+            .get_cache_policy(default_ism, &domain, ModuleType::Routing)
+            .await;
+        assert_eq!(cache_policy, IsmCachePolicy::MessageSpecific);
```
daniel-savu, line 378: wrote this before editing one of my comments above:

so once this PR is merged we will always be using cache? My initial understanding was that only `chains` in the `IsmCacheConfig` get to use cache, but seeing the result for `barchain` here shows otherwise. Would be nice to turn off caching if needed

daniel-savu (approved):

kamiyaa (approved):

rust/main/agents/relayer/src/msg/metadata/base.rs
tkporter: yeah good point, I'll add a config to globally allow caching

tkporter: added `allow_contract_call_caching` to make caching in general optional

rust/main/agents/relayer/src/msg/metadata/multisig/base.rs
tkporter: definitely agree the whole caching setup is clunky and could use a refactor - I want us to revisit this later but due to the urgency I'm gonna go ahead with merging

rust/main/agents/relayer/src/msg/metadata/base.rs
tkporter: added a config to just enable / disable globally

PR #5900, opened 2025-04-10
fix(sealevel): return an error regardless of how simulation fails

daniel-savu:
### Description

this change had been made as part of the new submitter but introduced an alerting regression in the old submission logic


tkporter (approved):
I don't have any context on the adapter stuff so I wouldn't give any weight to my review there

PR #5905, opened 2025-04-11
feat: support registry override for check warp deploy job

Mo-Hussain:


typescript/infra/helm/check-warp-deploy/values.yaml
```diff
@@ -4,6 +4,7 @@ image:
 hyperlane:
   runEnv: mainnet3
   chains: []
+  registryCommit: '' # Registry commit to use
```
ltyu, line 7: if you use `main` here, would it just default to it?

ltyu (approved):
looks good. i believe i need to merge this in order for deploy-check-warp-deploy.ts to pick up the image

Issue #5547, opened 2025-02-21
Ensure consistent sorting of WarpCore config


ltyu:
## Problem
Currently, when we write the WarpCoreConfig to the Registry after `warp deploy` or `warp apply`, we are using `sortMapEntries: true`, which is called via `writeDeploymentArtifacts()`. 

This flag will "sort by comparing key values using the native less-than < operator." However, this flag is not predictable and creates inconsistent sorting, which creates large git diffs when we update the WarpCoreConfig (and other configs).

For example, consider this [PR that adds Uni and Bera chain to PZETH](https://github.com/hyperlane-xyz/hyperlane-registry/pull/594/files). In this case, `berachain-ethereum-swell-unichain-zircuit-config.yaml` is a completely new file instead of a renamed file with additions.

## Note
It is understood that github shows diffs in a heuristic and unpredictable manner, especially when there is a rename + large diff. The point of this PR is not to spend too much figuring out how Github show diffs, but to solve the sorting problem such that if we did not rename the file, we'd see the correct diffs.

## Solution
- Configure `toYamlString()` in the Registry to sort correctly.

## Nice to Have
- Consider all the places where we use `sortMapEntries: true` and use the same configuration

nambrot:
I would also say that we should have prettier/CI checks in the registry that ensures the sorting

mshojaei-txfusion:
I think prettier is not well configured for registry, so there might be huge changes if we apply formatting to all files.
[not sure why it is not configured as pre commit hook like mono-repo]

mshojaei-txfusion:
@ltyu I wrote a eslint plugin for this case, please take a look at PR and tell me what do you think.
As the issue was just with sorting different types of arrays in yaml, we need to specify how we want to order them so I made it dynamic we can customize it.

mshojaei-txfusion:
Added this for registry formatting as well:
https://github.com/hyperlane-xyz/hyperlane-registry/pull/635

PR #5909, opened 2025-04-11
feat: yaml sorting eslint plugin

mshojaei-txfusion:
### Description

This PR adds a new ESLint rule `sort-yaml-arrays` to the utils package, which allows for sorting arrays in YAML files based on specified keys. This helps maintain consistent and predictable ordering in YAML configuration files.

Key features:
- New utility function for array sorting by key
- YAML transformation with complete comment preservation 
- Support for sorting nested arrays at specific paths
- Comprehensive test coverage for all sorting scenarios and comment handling

### Drive-by changes

- Enhanced the comment preservation logic in YAML transformation
- Improved unit test coverage for YAML transformation utilities

### Related issues
Required to resolve
https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5547

### Backward compatibility

Yes - This change is fully backward compatible. It adds a new ESLint rule that can be optionally enabled.

### Testing

Unit Tests - Comprehensive test suite covers:
- Basic array sorting functionality
- Nested array sorting
- Comment preservation
- Various edge cases


typescript/utils/src/arrays.ts
```diff
@@ -20,3 +22,15 @@ export function exclude<T>(item: T, list: T[]) {
 export function randomElement<T>(list: T[]) {
   return list[randomInt(list.length)];
 }
+
+export function sortArrayByKey<T extends Record<string, any>>(
+  array: T[],
+  sortKey: string,
```
xeno097, line 28: Can you update this to be `keyof T` instead of a plain string? so we can have type safety there too

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +22,406 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+    });
+  });
+
+  return matchingConfig?.sortKey || null;
+}
+
+/**
+ * Sorts arrays nested within objects according to configuration
+ */
+export function sortNestedArrays<T = any>(
```
xeno097, line 397: Can we use [this](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/utils/src/objects.ts#L355) in here for the recursive logic?

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +22,406 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+  data: T,
+  config: ArraySortConfig,
+  path: string[] = [],
+): T {
+  // Handle primitive values
+  if (!isObject(data) && !isArray(data)) {
```
antigremlin, line 403: Why not reverse the conditions?

```typescript
if (Array.isArray(data)) {}
if (typeof data === 'object' && data !== null) {}
return data;
```

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +22,406 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+  if (isArray(data)) {
+    const sortKey = findSortKeyForPath(path, config);
+
+    // Process each array item recursively
+    const processedArray = data.map((item, idx) =>
+      sortNestedArrays(item, config, [...path, idx.toString()]),
+    );
+
+    return (
+      sortKey ? sortArrayByKey(processedArray, sortKey) : processedArray
```
antigremlin, line 417: Do we really need a dedicated function?

```typescript
!sortKey
  ? processedArray
  : processedArray.sort((a, b) => 
      a[sortKey] < b[sortKey] ? -1 : a[sortKey] > b[sortKey] ? 1 : 0
```

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +22,406 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+
+    if (path.length !== patternParts.length) {
+      return false;
+    }
+
+    return path.every((part, idx) => {
+      const patternPart = patternParts[idx];
+      if (patternPart === '*') return true;
+      if (patternPart.endsWith('[]')) return patternPart.slice(0, -2) === part;
+      return patternPart === part;
```
antigremlin, line 387: What is the purpose of a pattern like `foo[].bar`? Or does it work as intended
Will this handle a path like `foo.4.bar`?

It seems we may need tests for this algorithm.

typescript/utils/src/eslint-rules/sort-yaml-arrays.ts
```diff
@@ -0,0 +1,87 @@
...
+          if (!yamlData) return;
+
+          const sortedData = sortNestedArrays(yamlData, sortConfig);
+
+          const sortedYaml = YAML.stringify(sortedData);
+          const originalYaml = YAML.stringify(yamlData);
+
+          if (sortedYaml !== originalYaml) {
+            context.report({
+              node,
```
antigremlin, line 65: Does it report errors on a particular line of the YAML file, or always on the first line?

typescript/utils/src/yaml.test.ts
```diff
@@ -31,3 +37,366 @@ describe('tryParseJsonOrYaml', () => {
...
+  it('should sort a simple array by key', () => {
+    const data = {
+      items: [
+        { name: 'c', value: 3 },
+        { name: 'a', value: 1 },
+        { name: 'b', value: 2 },
+      ],
+    };
+
+    const config: ArraySortConfig = {
```
antigremlin, line 51: How about some objects in the array not having the expected key?

antigremlin (commented):
This is a great effort, considering all the complexity of preserving comments. It may need more work though, considering the existing test cases.

Have we considered extending the rules in `eslint-plugin-yml`? It takes a parser-driven approach.

mshojaei-txfusion:
> This is a great effort, considering all the complexity of preserving comments. It may need more work though, considering the existing test cases.
> 
> Have we considered extending the rules in `eslint-plugin-yml`? It takes a parser-driven approach.

First tried with `yaml-eslint-parser` but found it so complex and this way was more straight forward and easier to test.

typescript/utils/src/arrays.ts
mshojaei-txfusion: Done!

typescript/utils/src/yaml.ts
mshojaei-txfusion: For this we need to apply this change there, that I think would break other functionalities

xeno097 (approved):
approved on my side but wait for @antigremlin to check the pending comments

typescript/utils/src/eslint-rules/sort-yaml-arrays.ts
mshojaei-txfusion: No it wont mention line, do you think we need to add it?

typescript/utils/src/yaml.ts
antigremlin: It's OK, let it stay as it is. Just for clarity, here's why I raised this:

- I agree, that having a dedicated utility function can make the code more readable for someone familiar with the codebase, especially when this function is used often.
- However, such functions often are in a different file (or even a different package). This introduces dependencies, and increases cognitive load when reading/writing code. It goes like this:

> What does it do? I don't exactly remember... Let me navigate to source and read it. Oh, it's the same as that lodash func, right...

- Making these generic utilities type-safe requires extra effort
- You wrote tests for this, even though the underlying `sortBy` is likely well-tested
- This code depends on `lodash`

I'm taking this a bit too far, maybe. But you see my point 😄

typescript/utils/src/eslint-rules/sort-yaml-arrays.ts
antigremlin: No, with the current approach it would be quite difficult. I wanted to be sure I understood this.

typescript/utils/src/yaml.test.ts
```diff
@@ -31,3 +37,366 @@ describe('tryParseJsonOrYaml', () => {
...
+          { code: 'cat-a', items: [] },
+        ],
+      },
+    };
+
+    const config: ArraySortConfig = {
+      arrays: [
+        { path: 'level1.items', sortKey: 'id' },
+        { path: 'level1.categories', sortKey: 'code' },
+        { path: 'level1.categories.*.items', sortKey: 'priority' },
```
antigremlin, line 90: It seems it works the same as `level1.categories[].items`. But we're not testing the array syntax. Do we really need both `categories[]` and `categories.*`?

typescript/utils/src/yaml.test.ts
```diff
@@ -31,3 +37,366 @@ describe('tryParseJsonOrYaml', () => {
...
+            { id: 'y', completed: false },
+          ],
+        },
+      ],
+    };
+
+    const config: ArraySortConfig = {
+      arrays: [
+        { path: 'users', sortKey: 'name' },
+        { path: 'users.*.tasks', sortKey: 'id' },
```
antigremlin, line 140: Same comment about `users[]` vs. `users.*`

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +21,436 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+    pathIndex++;
+  }
+
+  return pathIndex === path.length;
+}
+
+/**
+ * Sorts arrays nested within objects according to configuration
+ */
+export function sortNestedArrays<T = any>(
```
antigremlin, line 427: It seems the type param doesn't need the `= any` part.

antigremlin (approved):
Nice!

typescript/utils/src/arrays.ts
```diff
@@ -20,3 +22,15 @@ export function exclude<T>(item: T, list: T[]) {
...
+
+export function sortArrayByKey<T extends Record<keyof T, any>>(
+  array: T[],
+  sortKey: keyof T,
+): T[] {
+  return sortBy(array, [
+    (item) => {
+      const value = item[sortKey];
+      return value;
+    },
```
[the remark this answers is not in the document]
antigremlin: 👎 Incorrect. `_.sortBy()` appears to put undefined and null values at the end of the sort.

typescript/utils/src/arrays.ts
```diff
@@ -20,3 +22,15 @@ export function exclude<T>(item: T, list: T[]) {
 export function randomElement<T>(list: T[]) {
   return list[randomInt(list.length)];
 }
+
+export function sortArrayByKey<T extends Record<keyof T, any>>(
+  array: T[],
+  sortKey: keyof T,
+): T[] {
+  return sortBy(array, [
```
antigremlin, line 30: Why not `return sortBy(array, [(item) => item[sortKey]);` ?

typescript/utils/src/yaml.test.ts
mshojaei-txfusion: categories[] is for arrays, while categories.* allows any object or array under a category, but the next level matters. So, `*` works for both, but `[]` covers cases where we might not intend to target objects.

mshojaei-txfusion: Anyway will improve tests to cover this edge case

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +21,436 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+      pathIndex++;
+      continue;
+    }
+
+    if (pattern !== pathSegment) return false;
+    pathIndex++;
+  }
+
+  return pathIndex === path.length;
+}
```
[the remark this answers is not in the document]
antigremlin: @mshojaei-txfusion WDYT?

antigremlin (commented):
Please have a look at the remaining comments from the Graphite bot.

ltyu (approved):

typescript/utils/src/yaml.ts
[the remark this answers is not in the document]
mortezashojaei: Anyway added this test case to cover this scenario: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5909/commits/1b49b9486af1bc3245d5e379a1954f1f4f5a0b54

typescript/utils/src/yaml.ts
```diff
@@ -15,3 +21,436 @@ export function tryParseJsonOrYaml<T = any>(input: string): Result<T> {
...
+      const prefix = pattern.slice(0, -2);
+      if (!prefix || pathSegment !== prefix) return false;
+
+      pathIndex++;
+      if (pathIndex >= path.length) return false;
+
+      if (isNaN(Number(path[pathIndex]))) return false;
+      pathIndex++;
+      continue;
+    }
```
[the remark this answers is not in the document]
mortezashojaei: We dont want to match in case `anything.[].anything` as we already have * for this case!

typescript/utils/src/eslint-rules/sort-yaml-arrays.ts
```diff
@@ -0,0 +1,87 @@
...
+            });
+          }
+        } catch (error: unknown) {
+          context.report({
+            node,
+            message: `Error processing YAML: ${
+              error instanceof Error ? error.message : String(error)
+            }`,
+          });
+        }
```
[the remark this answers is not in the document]
mortezashojaei: No need as this is already part of a return:

```
    return {
      Program(node: Rule.Node) {
```

typescript/utils/src/eslint-rules/sort-yaml-arrays.test.ts
```diff
@@ -0,0 +1,647 @@
...
+      expected: `features:
+  - name: search
+    enabled: false
+  - name: auth
+    enabled: true
+  - name: notifications
+    enabled: true`,
+      options: {
+        arrays: [{ path: 'features', sortKey: 'enabled' }],
+      },
```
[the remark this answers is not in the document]
mortezashojaei: Good catch, fixed

antigremlin (approved):

Issue #5911, opened 2025-04-11
Pre-TGE: Cache validator announcements


ameten:
* With https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4860 merged, we have a foundation for using a cache for certain RPCs

* We should cache the lookup into the ValidatorAnnounce smart contract that the relayer does when building metadata here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/agents/relayer/src/msg/metadata/base_builder.rs#L164

* Note that the call we make is get_announced_storage_locations which takes in multiple validators, i.e. the validators in the set in question. A single RPC is done under the hood. So we really have 2 options for caching:

* Cache the entire call, where the cache key includes the Vec<H256> validators. This means that we only get a cache hit if we do a later lookup for this exact same validator set

* Or we could try to split this up to cache each validator's storage location separately, i.e. having the cache key just include the H256 of a validator, and we'd then have some more complicated logic to figure out cache hits/misses (because we'd still want to only have a single RPC get_announced_storage_locationswith the validators that are missed in the cache)

* Given our primary motivation is just feeling good for the default ISMs used by the HYPER / stHYPER, I think the first approach is totally acceptable for now. The complexity for the second approach feels annoying

* Example of how the cache is used here: https://github.com/Mantas-M/hyperlane-monorepo/blob/ba85e5a444f2aeb4a9ed15f24ef8a7d29d38d7f0/rust/main/agents/relayer/src/msg/metadata/multisig/base.rs#L101

PR #5912, opened 2025-04-11
feat: Add caching to retrieving announced storage locations of validators

ameten:
### Description

Add caching to retrieving announced storage locations of validators

- We cache storage locations for each validator separately.
- We retrieve storage locations for validators which we don't have storage locations for.

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5911

### Backward compatibility

Yes

### Testing

Unit tests


rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    if missing_validators.is_empty() {
+        // Cache contains storage locations for all validators
+        return Ok(storage_locations);
+    }
+
+    // Fetch from validator_announce for missing validators
+    let fetched_locations = validator_announce
```
paulbalaji, line 41: what happens for validators that have multiple announced locations e.g. started with one location but then migrated their infra and reannounced elsewhere

rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let mut storage_locations = Vec::new();
+    let mut missing_validators = Vec::new();
+
+    for (index, validator) in validators.iter().enumerate() {
+        let key = generate_cache_key(validator);
+
+        // Attempt to retrieve from cache
+        if let Some(cached) = cache
+            .get_cached_call_result::<Vec<String>>(DOMAIN_NAME, METHOD_NAME, &key)
```
tkporter, line 25: the validator announce is specific to the origin domain (i.e. the domain that the `&dyn ValidatorAnnounce` relates to). I think as written with the `DOMAIN_NAME` being `""`, this would mean that if a validator address is used for multiple different chains (which is often the case, many operators use the same validator key for validating different chains) then we'd get wrong data in our cache hit

I think you can use `validator_announce.domain().name()` to get the domain name

rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages.rs
tkporter: I think it'll essentially be the same behavior as before, but with some period where we have a stale value. The default expiry is 2 mins which I think is acceptable

rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages/tests.rs
```diff
@@ -0,0 +1,147 @@
...
+        fn address(&self) -> H256;
+    }
+
+    impl HyperlaneChain for ValidatorAnnounceMock {
+        fn domain(&self) -> &HyperlaneDomain;
+        fn provider(&self) -> Box<dyn HyperlaneProvider>;
+    }
+}
+
+#[tokio::test]
```
tkporter, line 39: would probably be good to test that different domains don't overlap in the cache

tkporter (commented):
lgtm, only thing is the domain stuff

daniel-savu (approved):
lgtm, agree with the comment on using a domain to avoid clashes

rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages.rs
ameten: Added

rust/main/agents/relayer/src/msg/metadata/base_builder/validator_announced_storages/tests.rs
ameten: Added

PR #5928, opened 2025-04-14
feat: use `aws-sdk-s3` for S3 operations

tkporter:
### Description

- Moving to `aws-sdk-s3` instead of rusoto for all S3 operations. We still require rusoto for KMS
- Moved to a static set of S3 clients that are used globally for all anonymous S3 operations. Clients are region specific so there are some ugly types. I preferred the static set to minimize how invasive this change is
- Some context here https://hyperlaneworkspace.slack.com/archives/C08GR6PBPGT/p1744561237142619?thread_ts=1744392583.189179&cid=C08GR6PBPGT
- Web Identity support is still there despite it being deleted, see https://docs.aws.amazon.com/sdk-for-rust/latest/dg/credproviders.html and https://docs.rs/aws-config/latest/aws_config/web_identity_token/index.html#environment-variable-configuration

### Drive-by changes

- sets `resolver = "2"`. This resulted in some slightly different compile warnings / errors due to various features now being activated or not activated

### Testing

- I tested locally with my local prepares but tbh there isn't a nice way of testing functionality in e2e or unit tests with a lot of this. So I'll roll out to RC cautiously

### DIFF
--- a/rust/main/agents/validator/src/settings.rs
+++ b/rust/main/agents/validator/src/settings.rs
@@ -260,12 +260,12 @@ fn parse_checkpoint_syncer(syncer: ValueParser) -> ConfigResult<CheckpointSyncer
                 .parse_string()
                 .end()
                 .map(str::to_owned);
-            let region = syncer
+            // Using rusoto_core::Region just to get some input validation
+            let region: Option<rusoto_core::Region> = syncer
                 .chain(&mut err)
                 .get_key("region")
-                .parse_string()
-                .end()
-                .map(str::to_owned);
+                .parse_from_str("Expected aws region")
+                .end();
             let folder = syncer
                 .chain(&mut err)
                 .get_opt_key("folder")
@@ -276,7 +276,7 @@ fn parse_checkpoint_syncer(syncer: ValueParser) -> ConfigResult<CheckpointSyncer
             cfg_unwrap_all!(&syncer.cwp, err: [bucket, region]);
             err.into_result(CheckpointSyncerConf::S3 {
                 bucket,
-                region: Region::new(region),
+                region: Region::new(region.name().to_owned()),
                 folder,
             })
         }
--- a/rust/main/agents/validator/src/settings.rs
+++ b/rust/main/agents/validator/src/settings.rs
@@ -6,6 +6,7 @@
 
 use std::{collections::HashSet, path::PathBuf, time::Duration};
 
+use aws_config::Region;
 use derive_more::{AsMut, AsRef, Deref, DerefMut};
 use eyre::{eyre, Context};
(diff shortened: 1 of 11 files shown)


rust/main/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -156,7 +156,6 @@ pub trait PendingOperation: Send + Sync + Debug + TryBatchAs<HyperlaneMessage> {
     fn reset_attempts(&mut self);
 
     /// Set the number of times this operation has been retried.
-    #[cfg(any(test, feature = "test-utils"))]
```
tkporter, line 159: this was sorta needed bc of the new resolver, otherwise some things wouldn't compile. Didnt want to deal with the headache of it so figured we can just add it

rust/main/hyperlane-base/src/types/s3_storage.rs
```diff
@@ -54,73 +57,87 @@ impl fmt::Debug for S3Storage {
...
             Err(e) => bail!(e),
         }
     }
 
-    /// Gets an authenticated S3Client, creating it if it doesn't already exist.
-    fn authenticated_client(&self) -> &S3Client {
-        self.authenticated_client.get_or_init(|| {
-            S3Client::new_with(
-                utils::http_client_with_timeout().unwrap(),
-                AwsChainCredentialsProvider::new(),
```
tkporter, line 100: this is a regression actually i'll add

daniel-savu (commented):
worth pushing the prepare step unit test and `#[ignore]` it if you still have it around. Maybe we can use it to break the op_submitter up into testable components

rust/main/agents/validator/src/settings.rs
```diff
@@ -274,7 +276,7 @@ fn parse_checkpoint_syncer(syncer: ValueParser) -> ConfigResult<CheckpointSyncer
             cfg_unwrap_all!(&syncer.cwp, err: [bucket, region]);
             err.into_result(CheckpointSyncerConf::S3 {
                 bucket,
-                region,
+                region: Region::new(region),
```
daniel-savu, line 279: any reason not to parse `Region` with `rusoto` in here as well?

daniel-savu (approved):
I take it you tested the anonymous client in your unit test, so we should run a validator with this change too to make sure the authenticated client works too?

tkporter:
correct on the authenticated client note

rust/main/agents/validator/src/settings.rs
tkporter: I didn't want to add it as a dep to the validator but might as well, I'll add

PR #5932, opened 2025-04-14
feat: Routing ISM caching

kamiyaa:
### Description

adds caching via `message.origin` for routing ISM

### Related issues

 - fixes https://linear.app/hyperlane-xyz/issue/ENG-1283/routing-ism-caching-for-the-default-ism-hyper-warp-route

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/routing.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/routing.rs
@@ -4,7 +4,7 @@ use derive_new::new;
 use hyperlane_base::cache::FunctionCallCache;
 use tracing::instrument;
 
-use hyperlane_core::{HyperlaneMessage, KnownHyperlaneDomain, ModuleType, H256};
+use hyperlane_core::{HyperlaneMessage, ModuleType, H256};
 
 use super::{
     base::MessageMetadataBuildParams, IsmCachePolicy, MessageMetadataBuilder, Metadata,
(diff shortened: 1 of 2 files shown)


rust/main/agents/relayer/src/msg/metadata/routing.rs
```diff
@@ -30,10 +31,57 @@ impl MetadataBuilder for RoutingIsmMetadataBuilder {
             .build_routing_ism(ism_address)
             .await
             .map_err(|err| MetadataBuildError::FailedToBuild(err.to_string()))?;
-        let module = ism
-            .route(message)
+
+        let message_domain = KnownHyperlaneDomain::try_from(message.origin)
```
tkporter, line 35: imo we should avoid `KnownHyperlaneDomain` here, let's just use `self.base_builder().origin_domain()` (or `message.origin` if that's not satisfactory for some reason)

https://hyperlaneworkspace.slack.com/archives/C08GR6PBPGT/p1744653023063519?thread_ts=1744652424.257009&cid=C08GR6PBPGT

rust/main/agents/relayer/src/msg/metadata/routing.rs
```diff
@@ -30,10 +31,57 @@ impl MetadataBuilder for RoutingIsmMetadataBuilder {
...
+        // Depending on the cache policy, make use of the message ID
+        let params_cache_key = match self
+            .base_builder()
+            .ism_cache_policy_classifier()
+            .get_cache_policy(self.root_ism, ism.domain(), ModuleType::Routing)
             .await
-            .map_err(|err| MetadataBuildError::FailedToBuild(err.to_string()))?;
+        {
+            // To have the cache key be more succinct, we use the message id
+            IsmCachePolicy::IsmSpecific => (ism.address(), H256::zero()),
```
tkporter, line 49: Instead of `H256::zero()` we should have this be the message.origin. The idea is that we want to take advantage of us knowing that it's a DomainRoutingIsm used, which means that we can only cache the result depending on the origin domain of a message. As written this would result in a cache hit for a message from chain A and chain B when we know that we should be getting different results bc they are from a different domain

Might be able to massage the type of `params_cache_key` to tolerate this as a different type without converting to H256 by matching the type expected by `get_cached_call_result`

rust/main/agents/relayer/src/msg/metadata/routing.rs
```diff
@@ -30,10 +31,57 @@ impl MetadataBuilder for RoutingIsmMetadataBuilder {
...
+            IsmCachePolicy::MessageSpecific => (ism.address(), message.id()),
+        };
+
+        let cache_result: Option<H256> = self
+            .base_builder()
+            .cache()
+            .get_cached_call_result(message_domain.as_str(), fn_key, &params_cache_key)
```
tkporter, line 56: the first param here should be the `ism.domain().name()`. this is meant to be the domain of the ISM that we are caching

rust/main/agents/relayer/src/msg/metadata/routing.rs
```diff
@@ -30,10 +31,57 @@ impl MetadataBuilder for RoutingIsmMetadataBuilder {
...
+            None => {
+                let module = ism
+                    .route(message)
+                    .await
+                    .map_err(|err| MetadataBuildError::FailedToBuild(err.to_string()))?;
+
+                // store result in cache
+                self.base_builder()
+                    .cache()
+                    .cache_call_result(message_domain.as_str(), fn_key, &params_cache_key, &module)
```
tkporter, line 75: ditto here for the first param

tkporter (approved):
🫡

PR #5933, opened 2025-04-14
feat: make all CLI deps as dev deps to make bundle install self sufficient

xeno097:
### Description

This PR implements the proposed solution to avoid installing NPM dependencies when installing the CLI by setting al the dependencies as dev dependencies so that they will be ignored when installing the CLI with `npm install -g` and avoid third party deps breaking the install process

### Drive-by changes

- Updates the workspace commands to use `--topological-dev` flag instead of the `--topological` to avoid the cli being built when depencies in the workspace haven't been built yet

### Backward compatibility

- YES

### Testing

- Manual
- E2E


xeno097:
> Looks good, assuming this is part of the bigger PR.

yup I'm going to merge this in the parent PR

paulbalaji (approved):

PR #5937, opened 2025-04-15
feat: Add mixing of message processing depending on configuration

ameten:
### Description

Currently we sort new messages in prepare queue depending on nonce and their ids. It gives us the same order of messages in every instance of Relayer we are running. It means that all instances of relayer will attempt to deliver the same message and one of them will succeed while others will fail. It is quite wasteful on resources.

We would like to make the order of new messages in prepare queue different for each instance of relayer so that each instance mostly working on a distinct message while merely checking if a message was delivered if it was processed by another instance of relayer. We are fine with the order being stochastic, i.e. the order of messages will be different for each instance of relayer but it will be deterministic for each instance of relayer.

We introduce a salt which will be configured with distinct value for each relayer and we will hash message id with the salt producing the stochastic ordering of messages.

This behaviour will be configured with environment variables: `HYPERLANE_RELAYER_MIXING_ENABLED` to enable mixing of messages in each relayer instance and `HYPERLANE_RELAYER_MIXING_SALT` to specify the salt.

We add a server endpoint so that environment variables can be set on the fly in Relayer. This endpoint will be exposed if environment variable `HYPERLANE_RELAYER_ENVIRONMENT_VARIABLE_ENDPOINT_ENABLED` is set to `true` when relayer starts.

We don't cache the values of the environment variables so that the behaviour of relayer can be changes without restarting it.

### Backward compatibility

Yes -  when environment variable `HYPERLANE_RELAYER_MIXING_ENABLED` is not specified, relayer behaves as usual.

### Testing

Manual


daniel-savu (approved):

rust/main/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -363,18 +365,34 @@ impl Eq for QueueOperation {}
...
             // No time means it should come before
             (None, Some(_)) => Less,
             (Some(_), None) => Greater,
             (None, None) => {
-                if self.origin_domain_id() == other.origin_domain_id() {
-                    // Should execute in order of nonce for the same origin
-                    self.priority().cmp(&other.priority())
+                let mixing =
+                    env::var("HYPERLANE_RELAYER_MIXING_ENABLED").map_or(false, |v| v == "true");
+                if !mixing {
```
daniel-savu, line 381: worth adding unit tests with different values for these two environment vars

rust/main/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -363,18 +365,34 @@ impl Eq for QueueOperation {}
...
+                    }
                 } else {
-                    // There is no priority between these messages, so arbitrarily use the id
-                    self.id().cmp(&other.id())
+                    let salt = env::var("HYPERLANE_RELAYER_MIXING_SALT")
+                        .map_or(0, |v| v.parse::<u32>().unwrap_or(0))
+                        .to_vec();
+                    let self_hash = salted_hash(&self.id(), &salt);
+                    let other_hash = salted_hash(&other.id(), &salt);
+                    self_hash.cmp(&other_hash)
```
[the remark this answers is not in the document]
ameten: It makes sense. Since we calculate message id on-the-fly as well, implementation with additional hashing makes it  2x more expensive.

rust/main/hyperlane-core/src/traits/pending_operation.rs
```diff
@@ -363,18 +365,34 @@ impl Eq for QueueOperation {}
...
+                        // Should execute in order of nonce for the same origin
+                        self.priority().cmp(&other.priority())
+                    } else {
+                        // There is no priority between these messages, so arbitrarily use the id
+                        self.id().cmp(&other.id())
+                    }
                 } else {
-                    // There is no priority between these messages, so arbitrarily use the id
-                    self.id().cmp(&other.id())
+                    let salt = env::var("HYPERLANE_RELAYER_MIXING_SALT")
```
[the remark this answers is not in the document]
ameten: At the start, we can set it in StatefulSet. We'll use API only if we need to change it mid-fligth.

rust/main/hyperlane-core/src/traits/pending_operation.rs
ameten: Added unit tests

rust/main/agents/relayer/src/server/environment_variable.rs
```diff
@@ -0,0 +1,188 @@
...
+        message: message.to_string(),
+    };
+
+    Ok(Json(response))
+}
+
+impl EnvironmentVariableApi {
+    pub fn router(&self) -> Router {
+        Router::new()
+            .route("/", routing::get(get_environment_variable))
```
tkporter, line 67: Definitely get what you mean about security concerns 😅

I'd be nervous that someone would publicly expose their 9090 port thinking that it's just metrics, and then they end up exposing some secrets. I'm glad though that this isn't for the validator and relayer only - the worst thing here I guess is just halting liveness or being able to fetch the relayer private key

I do wonder if we should only enable this API if explicitly configured to?

tkporter (approved):
logic all lgtm, curious what you think of conditionally exposing this new API

rust/main/agents/relayer/src/server/environment_variable.rs
ameten: I've added a environment variable (sic!) which enables/disables environment variable endpoint.

PR #5938, opened 2025-04-15
feat(infra): support multiple governance types

paulbalaji:
### Description

feat(infra): support multiple governance types
- add separate owners file for matrix of owner x governance types
	- [safe,ica] owner x [regular,irregular,exceptional,AW]
- add reusable utils/types for reasoning with multiple governance types/owners
- update any script dealing with safes/icas to correctly operate on the corresponding set for a given governance type
- streamline owners.ts
- fix warp config generation by manually configuring legacy-style owner/ownerOverrides

### Drive-by changes

- switch logging from `console` to `rootLogger` in more places
- refactor signers configuration
- refactor old create-safes script

### Related issues

https://linear.app/hyperlane-xyz/issue/AW-75

### Testing

manual check-deploy


typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -28,247 +31,27 @@ export function localAccountRouters(): ChainMap<Address> {
...
+      awIcas[local] ??
+      awSafes[local] ??
+      DEPLOYER;
 
     return [
       local,
       {
         owner,
         ownerOverrides: {
-          proxyAdmin: upgradeTimelocks[local] ?? owner,
```
Mo-Hussain, line 262: Did we get clarity on whether the arbitrum proxyAdmin would be a timelock moving forward?

typescript/infra/config/environments/mainnet3/owners.ts
```diff
@@ -28,247 +31,27 @@ export function localAccountRouters(): ChainMap<Address> {
...
-  reactive: '0x9312B04076efA12D69b95bcE7F4F0EA847073E6a',
-} as const;
-
 export const DEPLOYER = '0xa7ECcdb9Be08178f896c26b7BbD8C3D4E844d9Ba';
 
 export const ethereumChainOwners: ChainMap<OwnableConfig> = Object.fromEntries(
   ethereumChainNames.map((local) => {
-    const owner = icas[local] ?? safes[local] ?? DEPLOYER;
+    const owner =
+      regularIcas[local] ??
```
Mo-Hussain, line 40: Is this why the maps are commented out for now? After ownership transfer will be remove `awIcas[local]` and `awSafes[local]`

typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -160,25 +169,10 @@ export abstract class HyperlaneAppGovernor<
     const sendCallsForType = async (
       submissionType: SubmissionType,
       multiSend: MultiSend,
+      governanceType?: GovernanceType,
     ) => {
       const callsForSubmissionType = [];
-      const filteredCalls = filterCalls(submissionType);
-
-      // If calls are being submitted via a safe, we need to check for any safe owner changes first
```
Mo-Hussain, line 167: What's the reason behind removing this?

paulbalaji: i think i did it before the signers config refactor, but can add it back if you'd like

typescript/infra/src/config/environment.ts
```diff
@@ -89,7 +95,28 @@ export async function getRouterConfigsForAllVms(
     envConfig.environment,
     multiProvider,
   );
-  const evmRouterConfig = core.getRouterConfig(envConfig.owners);
+
```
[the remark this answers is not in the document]
paulbalaji: atm it piggybacks off of the core ownership config
```
const evmRouterConfig = core.getRouterConfig(envConfig.owners);
```
but we are changing the core ownership here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2c038739cccd3f0251ac8e6e139891b01e22fa84/typescript/infra/config/environments/mainnet3/owners.ts#L39-L44

so if unchanged, then it'll expect the warp route ownership to transfer - keeping this unchanged for now to detangle warp route changes from core changes

once a decision is made on the warp route situation we can change

typescript/infra/config/environments/mainnet3/owners.ts
paulbalaji: that is the intention yup

typescript/infra/config/environments/mainnet3/owners.ts
paulbalaji: not yet - but i would suspect it won't be initially? unsure why arb is timelocked in the first place

typescript/infra/src/govern/HyperlaneAppGovernor.ts
Mo-Hussain: I'd want to understand if we'd lose anything if we remove. We ould be checking if the safe signers are different before doing a governance action?

paulbalaji: for context this check was added initially as a way to easily update the signers of an abacus works safe within the same multisend tx as the ism updates (to save on #txs a signer would need to sign)

imo since we're still checking the safes with a separate script, we're not really losing anything by removing it from here

typescript/infra/src/config/environment.ts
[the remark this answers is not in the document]
Mo-Hussain: @nambrot confirmed that warp routes can still be owned by the AW multisig

Mo-Hussain (approved):

PR #5941, opened 2025-04-15
feat: async checkpoint syncer builder

yjamin:
### Description
Build the `checkpoint_syncer` in parallel for each validator in the `BaseMetadataBuilder`
- limits the number of concurrent requests to 10

### Related issues
[Linear](https://linear.app/hyperlane-xyz/issue/BACK-148/make-basemetadatabuilderbuild-checkpoint-syncer-more-concurrent)

### Testing
Ran E2E tests

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/base_builder.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/base_builder.rs
@@ -226,6 +226,7 @@ impl BuildsBaseMetadata for BaseMetadataBuilder {
                             Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event)) => {
                                 // If a reorg event has been posted to a checkpoint syncer,
                                 // we refuse to build
+                                // This will result in a short circuit and return an error for the entire build process of all syncers 
                                 return Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event));
                             }
                             Err(err) => {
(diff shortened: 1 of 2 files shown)


rust/main/agents/relayer/src/msg/metadata/base_builder.rs
```diff
@@ -180,63 +181,88 @@ impl BuildsBaseMetadata for BaseMetadataBuilder {
...
+                Some(future)
+            })
+            .collect::<Vec<_>>();
+
+        let checkpoint_syncers_results = stream::iter(result)
+            .buffer_unordered(10) // Limit the number of concurrent tasks
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, _>>()? // Collect results into a single vector
```
tkporter, line 257: This means that if one of the futures gives an Err, that we short circuit / error out of this function right?

This seems like the intended behavior with the whole `Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event));` logic where we want to be conservative and not build a `MultisigCheckpointSyncer` if one of the validators says it's experienced a reorg - but the reason I call it out is that I think it's pretty subtle. Could you maybe add a comment or something about it? Either here or next to the `return Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event));`. I just wanna avoid us introducing more Err opportunities in the future that may not have that intended behavior

tkporter (approved):
🙏

PR #5945, opened 2025-04-15
feat: app-context specific ISM cache configs, reuse Ethereum HTTP reqwest clients

tkporter:
### Description

- Moves away from the default ISM cache config to a vec of ISM cache configs with a selector that can match either default ISM use or a specific app context
- Motivating use case here is the testnet4 load testing doesn't actually use the default ISM. This will also allow us to cache other routes like ezETH etc that don't use the default ISM directly
- Reuses Ethereum HTTP reqwest clients instead of creating a bunch of new ones

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/base.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/base.rs
@@ -208,11 +208,20 @@ pub enum IsmCachePolicy {
     IsmSpecific,
 }
 
+#[derive(Debug, Clone, Default, Deserialize, PartialEq)]
+#[serde(tag = "type", content = "data", rename_all = "camelCase")]
+pub enum IsmCacheSelector {
+    #[default]
+    DefaultIsm,
+    AppContext(String),
+}
+
 /// Configuration for ISM caching behavior.
 /// Fields are renamed to be all lowercase / without underscores to match
 /// the format expected by the settings parsing.
 #[derive(Debug, Clone, Default, Deserialize)]
 pub struct IsmCacheConfig {
+    selector: IsmCacheSelector,
     #[serde(deserialize_with = "deserialize_module_types", rename = "moduletypes")]
     module_types: HashSet<ModuleType>,
     chains: Option<HashSet<String>>,
@@ -257,7 +266,7 @@ impl IsmCacheConfig {
 #[derive(Debug, new)]
 pub struct IsmCachePolicyClassifier {
     default_ism_getter: DefaultIsmCache,
-    default_ism_cache_config: IsmCacheConfig,
+    ism_cache_configs: Vec<IsmCacheConfig>,
 }
 
 impl IsmCachePolicyClassifier {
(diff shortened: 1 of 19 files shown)


ameten (approved):

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -208,11 +208,20 @@ pub enum IsmCachePolicy {
     IsmSpecific,
 }
 
+#[derive(Debug, Clone, Default, Deserialize, PartialEq)]
+#[serde(tag = "type", content = "data", rename_all = "camelCase")]
```
daniel-savu, line 214: I found the `data` naming of the content a bit confusing, how about calling it `context`

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -267,28 +276,40 @@ impl IsmCachePolicyClassifier {
...
-                tracing::warn!(?err, "Error fetching default ISM for ISM cache policy, falling back to default cache policy");
-                return IsmCachePolicy::default();
+        for config in &self.ism_cache_configs {
+            let matches_module = match &config.selector {
+                IsmCacheSelector::DefaultIsm => {
+                    let default_ism = match self.default_ism_getter.get().await {
+                        Ok(default_ism) => default_ism,
+                        Err(err) => {
+                            tracing::warn!(?err, "Error fetching default ISM for ISM cache policy, falling back to default cache policy");
+                            return IsmCachePolicy::default();
```
daniel-savu, line 290: what if fetching the default ISM fails here, but another cache config selector would've matched? Does this stop searching for a match too soon? Since we return `IsmCachePolicy::default()` at the end of this function, maybe we could `continue` on error

rust/main/agents/relayer/src/msg/metadata/base.rs
```diff
@@ -269,28 +278,40 @@ impl IsmCachePolicyClassifier {
...
+                        }
+                    };
+                    root_ism == default_ism
+                }
+                IsmCacheSelector::AppContext(selector_app_context) => {
+                    if let Some(app_context) = app_context {
+                        app_context == selector_app_context
+                    } else {
```
daniel-savu, line 298: nit: replace with `app_context.map_or(false, |ctx| ctx == selector_app_context)`

daniel-savu (approved):

typescript/infra/config/environments/testnet4/agent.ts
```diff
@@ -228,19 +230,60 @@ const scraperResources = {
...
-    ModuleType.MESSAGE_ID_MULTISIG,
-  ],
-  // SVM is explicitly not cached as the default ISM is a multisig ISM
-  // that routes internally.
-  chains: ethereumChainNames,
-  cachePolicy: IsmCachePolicy.IsmSpecific,
-};
+const kesselMatchingList = [
+  {
+    recipientAddress: '0x492b3653A38e229482Bab2f7De4A094B18017246',
```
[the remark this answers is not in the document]
tkporter: yeah I'll clarify

rust/main/agents/relayer/src/msg/metadata/base.rs
tkporter: agreed, good call

PR #5946, opened 2025-04-16
feat: Add rebalancing to HypNative

larryob:
### Description
Allow collateral in HypNative to be rebalanced.

### Drive-by changes

N/A

### Related issues

N/A

### Backward compatibility

Yes
### Testing

Unit.

### Other Info

Builds on #5925


solidity/contracts/token/HypNativeMovable.sol
```diff
@@ -0,0 +1,63 @@
...
+contract HypNativeMovable is HypNative, MovableCollateralRouter {
+    constructor(uint256 _scale, address _mailbox) HypNative(_scale, _mailbox) {}
+
+    function _rebalance(
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount,
+        ValueTransferBridge bridge
+    ) internal override {
+        bridge.transferRemote{value: amount}({
```
yorhodes, line 37: the value required by the call may be greater than amount, how should we handle that?

solidity/contracts/token/HypNativeMovable.sol
yorhodes, line 1: I think @nambrot prefers us to just adopt this in `HypNative` and `HypERC20Collateral` directly

larryob: Done

solidity/contracts/token/HypNativeMovable.sol
larryob: The same issue exists with the ERC20 collateral contracts. If we want to send 10 usdc and the bridge charges a 10% fee, should we call `usdc.approve(bridge, 11)` or `usdc.approve(bridge, 10)` (bridge keeps 1 usdc, and we receive 9 usdc on the remote chain)? 

Seems simpler to just send the bridge an amount (whether the bridge takes it from us with `transferFrom` or we send it with msg.value), and let it take a fee from that amount.

larryob: Updated this function to pass in `msg.value` as a fee in addition to the amount we wish to move. If the value required by the call is greater than `amount`, it will have to be covered with `msg.value`.

solidity/contracts/token/HypNative.sol
```diff
@@ -33,8 +39,9 @@ contract HypNative is FungibleTokenRouter {
         address _hook,
         address _interchainSecurityModule,
         address _owner
-    ) public initializer {
+    ) public virtual reinitializer(2) {
         _MailboxClient_initialize(_hook, _interchainSecurityModule, _owner);
+        _MovableCollateralRouter_initialize(_owner);
```
yorhodes, line 44: can we add a comment about the initialized version?

yorhodes (approved):
lgtm
is the 66% patch coverage accurate?

larryob:
> lgtm is the 66% patch coverage accurate?

Yeah, the two lines where I used the `INSUFFICIENT_NATIVE_AMOUNT` constant aren't covered by tests. Btw, are there any existing tests for HypNative?

solidity/contracts/token/HypNative.sol
larryob: Fixed

solidity/contracts/token/HypNative.sol
```diff
@@ -107,4 +114,18 @@ contract HypNative is FungibleTokenRouter {
...
     }
+
+    function _rebalance(
+        uint32 domain,
+        bytes32 recipient,
+        uint256 amount,
+        ValueTransferBridge bridge
+    ) internal override {
+        uint fee = msg.value;
+        bridge.transferRemote{value: fee + amount}({
```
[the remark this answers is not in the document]
larryob: Fixed

PR #5956, opened 2025-04-17
feat: apply prepare queue backpressure

tkporter:
### Description

- Applies backpressure to the prepare queue based off the value of the env var `HYP_MAXSUBMITQUEUELENGTH`
- Drive-by to lower the bsctestnet gas price to 1 gwei instead of 8, which is far higher and seemingly a relic of past times
- make IGP indexing optional
- make tx id indexing optional


rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -339,6 +352,20 @@ async fn prepare_task(
     // Prepare at most `max_batch_size` ops at a time to avoid getting rate-limited
     let ops_to_prepare = max_batch_size as usize;
     loop {
+        // Apply backpressure to the prepare queue if the submit queue is too long.
+        if let Some(max_len) = max_submit_queue_len() {
```
daniel-savu, line 356: should this default to 2-3x the max batch size?

daniel-savu (approved):

rust/main/agents/relayer/src/msg/op_submitter.rs
tkporter: As idk if we'll use this going forward I think just will keep it manual for now, we can add later if it's useful though

daniel-savu (approved):
the indexing changes make sense too

PR #5958, opened 2025-04-17
chore: log eip1559 price estimation error

daniel-savu:


tkporter (approved):
would we maybe get an error from the lower lvl stuff already? like the fallback provider etc

ameten (approved):

kamiyaa (approved):

PR #5963, opened 2025-04-17
feat: add debug message for unknown domains for chain metrics

kamiyaa:
### Description
Add debug message so we can better debug missing block height metrics for certains chains:
https://linear.app/hyperlane-xyz/issue/BACK-94/hyperlane-block-height-metrics-are-missing-for-some-chains

### Related issues

- related: https://linear.app/hyperlane-xyz/issue/BACK-94/hyperlane-block-height-metrics-are-missing-for-some-chains

### Backward compatibility
Yes


ameten (approved):

daniel-savu (changes requested):
let's keep this unmerged until tomorrow eve to minimize new code for a while

daniel-savu (approved):

rust/main/hyperlane-base/src/metrics/agent_metrics.rs
```diff
@@ -206,7 +206,16 @@ impl ChainSpecificMetricsUpdater {
...
+            ..
+        } = &self.conf.domain
+        {
+            debug!(
+                domain_id,
+                domain_name, "Unknown domain, skipping chain metrics"
+            );
             return;
         };
         let chain = self.conf.domain.name();
```
daniel-savu, line 221: could also make the `No chain metrics available` log a `debug`, since we wouldn't see the `trace`

PR #5964, opened 2025-04-17
fix: zeronetwork igp check

daniel-savu:


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -580,6 +580,8 @@ const gasPaymentEnforcement: GasPaymentEnforcement[] = [
       // Infinity VM is gasless, so enforcing min 1 wei here ensures outbound txs
       // outside of Solana are ignored.
       { originDomain: getDomainId('infinityvm') },
+      // Temporary workaround due to funky Zeronetwork gas amounts.
+      { destinationDomain: getDomainId('zeronetwork') },
```
paulbalaji, line 584: do we need to change any of the other zksync chains?

paulbalaji (approved):

typescript/infra/config/environments/mainnet3/agent.ts
daniel-savu: hopefully not, we'll see if @ameten gets any alerts there

PR #5972, opened 2025-04-17
fix: request latest block and fee history in parallel and filter out failed calls and estimate only successful calls

ameten:
### Description

Request latest block and fee history in parallel so that the latency to update cache with block number and fees is reduced. 

We filter out failed calls from the batch so that the batch is smaller and we don't waist resources.

### Backward compatibility

Yes

### Testing

Unit test with real RPC calls

### DIFF
--- a/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
+++ b/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
@@ -611,32 +611,16 @@ where
                 )
             })
             .collect::<ChainResult<Vec<_>>>()?;
+        // these two we do in parallel
+        let res = self
+            .simulate_batch(&mut multicall, contract_calls.clone())
+            .await?;
         if let Some(contract_call) = contract_calls.first() {
             self.refresh_block_and_fee_cache(&contract_call.tx).await?;
         }
-        let filled_tx_params_futures = contract_calls.iter().map(|tx| {
-            fill_tx_gas_params(
-                tx.clone(),
-                self.provider.clone(),
-                &self.conn.transaction_overrides,
-                &self.domain,
-                true,
-                self.cache.clone(),
-            )
-        });
-        let contract_calls = join_all(filled_tx_params_futures)
-            .await
-            .into_iter()
-            .collect::<ChainResult<Vec<_>>>()?;
 
-        if self.conn.operation_batch.bypass_batch_simulation {
-            // submit the tx without checking if subcalls would revert
-            self.submit_multicall(&mut multicall, contract_calls, self.cache.clone())
-                .await
-        } else {
-            self.simulate_and_submit_batch(&mut multicall, contract_calls, self.cache.clone())
-                .await
-        }
+        // and then submit
+        res.try_submit(self.cache.clone()).await
     }
 
     #[instrument(skip(self), fields(msg=%message, metadata=%bytes_to_hex(metadata)))]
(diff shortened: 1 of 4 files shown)


rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -438,19 +440,13 @@ where
...
-            .await
-            .map_err(ChainCommunicationError::from_other)?
-            .ok_or_else(|| ProviderError::CustomError("Latest block not found".into()))?;
-        let eip1559_fee =
-            estimate_eip1559_fees(self.provider.clone(), None, &latest_block, &self.domain, tx)
+        let (eip1559_fee, latest_block) =
+            estimate_eip1559_fees(self.provider.clone(), None, &self.domain, tx)
                 .await
-                .ok();
+                .map_err(ChainCommunicationError::from_other)?;
```
ameten, line 446: Change in behaviour: before we did not throw an error when no fee were found, we returned None for fee. Now we throw an error and exist.

rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -304,32 +321,38 @@ struct ZksyncEstimateFeeResponse {
...
+        EIP1559_FEE_ESTIMATION_PAST_BLOCKS,
+        BlockNumber::Latest,
+        &[EIP1559_FEE_ESTIMATION_REWARD_PERCENTILE],
+    );
+
+    let (latest_block, fee_history) = join!(latest_block, fee_history);
+    let (latest_block, fee_history) = (
+        latest_block?,
+        fee_history.map_err(ChainCommunicationError::from_other)?,
+    );
```
ameten, line 339: This parallel execution of these two remote calls is the reason for this change

rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -163,10 +165,10 @@ where
     }
 
     let eip1559_fee_result = match cached_eip1559_fee {
-        Some(fee) => Ok(fee),
-        None => estimate_eip1559_fees(provider.clone(), None, &latest_block, domain, &tx.tx).await,
+        Some(fee) => Ok((fee, latest_block)),
+        None => estimate_eip1559_fees(provider.clone(), None, domain, &tx.tx).await,
```
ameten, line 169: We request latest block two times in `fill_tx_gas_params`, but since the request of the latest block inside `estimate_eip1559_fees` happens in parallel to requesting fee history, the overall latentcy is not impacted for regular chains. It is impacted for zksync since we don't request fee history there.

ameten: It is ok. Discussed in real life.

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
ameten: We will not fail batch in this case. Implemented in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5974

rust/main/chains/hyperlane-ethereum/src/tx.rs
```diff
@@ -223,32 +225,44 @@ pub type Eip1559Fee = (
...
 async fn estimate_eip1559_fees_zksync<M>(
     provider: Arc<M>,
-    latest_block: &Block<TxHash>,
     tx: &TypedTransaction,
-) -> ChainResult<Eip1559Fee>
+) -> ChainResult<(Eip1559Fee, Block<TxHash>)>
 where
     M: Middleware + 'static,
 {
+    let latest_block = latest_block(provider.clone()).await?;
```
daniel-savu, line 264: this could be awaited in parallel with `zksync_estimate_fee` but nbd

daniel-savu (approved):

PR #5973, opened 2025-04-17
feat: revoke approvals in warp core

xeno097:
### Description

This PR modifies the `ITokenAdapter` interface to add a new `isRevokeApprovalRequired` function used in WarpCore to check if allowance should be reset before setting a new allowance amount. This is useful for tokens like USDT, where they require that the current allowance is 0 when setting a new amount. 
ref https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7#code#L205

### Backward compatibility

- Yes

### Testing

- Manual testing locally with the UI


typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
```diff
@@ -154,6 +161,13 @@ export class CwTokenAdapter
     return false;
   }
 
+  async isRevokeApprovalRequired(
```
ltyu, line 164: should this be in `BaseAppAdapter`?

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -332,16 +333,42 @@ export class WarpCore {
...
     const hypAdapter = token.getHypAdapter(this.multiProvider, destinationName);
 
-    if (await this.isApproveRequired({ originTokenAmount, owner: sender })) {
-      this.logger.info(`Approval required for transfer of ${token.symbol}`);
+    const [isApproveRequired, isRevokeApprovalRequired] = await Promise.all([
+      this.isApproveRequired({
+        originTokenAmount,
+        owner: sender,
+      }),
+      hypAdapter.isRevokeApprovalRequired(
```
ltyu, line 341: why does isApproveRequired use getAdapter and isRevokeApprovalRequired use getHypAdapter?

whats the diff here?

typescript/sdk/src/token/adapters/CosmWasmTokenAdapter.ts
xeno097: I saw that the base adapter was something more generic, to build general adapters for interacting with a specific protocol, that is why I didn't add it there

typescript/sdk/src/warp/WarpCore.ts
xeno097: Because in transfer remote we are interacting with a HypToken, so I want to be sure that the adapter we are using is for only `HyptToken`s while the `getAdapter` might return an adapter that is not for a HypToken, which would be wrong in this context. getHypAdapter would fail if let's say that the provided token is a simple native or ERC20 one

ltyu (approved):

PR #5968, opened 2025-04-17
fix: Add simulation back in parallel with refreshing cache

ameten:
### Description

Add simulation back in parallel with refreshing cache so that we can see which calls to Mailbox.process failed in the batch.

### Backward compatibility

Yes

### Testing

Unit test with real RPC calls

PR #5974, opened 2025-04-18
fix: filter out failed calls and estimate only successful calls

ameten:
### Description

We filter out failed calls from the batch so that the batch is smaller and we don't waist resources.

### Backward compatibility

Yes

### Testing

Unit tests

### DIFF
--- a/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
+++ b/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
@@ -691,7 +691,7 @@ mod test {
         providers::{MockProvider, Provider},
         types::{Block, Transaction, U256 as EthersU256},
     };
-
+    use ethers_core::types::FeeHistory;
     use hyperlane_core::{
         ContractLocator, HyperlaneDomain, HyperlaneMessage, KnownHyperlaneDomain, Mailbox,
         TxCostEstimate, H160, H256, U256,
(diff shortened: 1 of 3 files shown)


rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -363,34 +362,27 @@ where
...
         // only send a batch if there are at least two successful calls
-        let call_count = contract_calls.len();
-        let successful_calls = call_count - failed_calls.len();
-        if successful_calls >= 2 {
+        if successful_calls_len >= 2 {
             Ok(BatchSimulation::new(
-                Some(self.submittable_batch(batch)),
-                failed_calls,
+                Some(self.submittable_batch(estimated)),
+                failed,
```
ameten, line 382: Since we don't do serial submission anymore, we may want to allow batches to have a single transaction. It will make building a new batch more reliable and we will avoid re-building the same batch as on a previous iterations which will produce a single successful call again.

rust/main/chains/hyperlane-ethereum/src/contracts/multicall.rs
```diff
@@ -85,15 +119,13 @@ where
                 None
             }
         };
-        multicall.add_call(call, ALLOW_BATCH_FAILURES);
     });
```
ameten, line 122: This aggregation is not relevant anymore since we removed population of individual transaction parameters in this PR - https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5968/files#diff-5f49b3c719e488330745b8f07595760a4c875fc78e39621fdefb97bc0c8e6b63L617-L630

daniel-savu (approved):

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -465,7 +464,9 @@ impl<M: Middleware + 'static> BatchSimulation<M> {
         self,
         cache: Arc<Mutex<EthereumMailboxCache>>,
     ) -> ChainResult<BatchResult> {
-        if let Some(submittable_batch) = self.call {
+        if let Some(mut submittable_batch) = self.call {
+            let estimated = multicall::estimate(submittable_batch.call, self.successful).await?;
```
daniel-savu, line 468: could this be run in parallel to the `fill_tx_gas_params` in `submit`?

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
ameten: We keep it as is.

rust/main/chains/hyperlane-ethereum/src/contracts/multicall.rs
ameten: We brought the aggregation back in this PR.

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -607,14 +608,28 @@ where
...
+        let (simulate_result, _) = join!(simulate_future, refresh_cache_future);
+        let mut simulation = simulate_result?;
+
+        let filled_tx_params_futures = simulation.successful.iter().map(|tx| {
+            fill_tx_gas_params(
+                tx.clone(),
+                self.provider.clone(),
+                &self.conn.transaction_overrides,
+                &self.domain,
+                true,
```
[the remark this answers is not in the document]
ameten: Done

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
ameten: Function `fill_tx_gas_params` uses gas limit estimate which we got from `multicall::estimate`. So, we cannot.

daniel-savu (approved):

PR #5981, opened 2025-04-19
fix: Add script to run relayer locally

ameten:
### Description

Add script to run relayer locally

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/rust/docker-compose.yaml
+++ b/rust/docker-compose.yaml
@@ -1,0 +1,20 @@
+services:
+  relayer-service:
+    image: gcr.io/abacus-labs-dev/hyperlane-agent:da3978b-20250414-155929
+    command: >
+      /bin/sh -c "./relayer
+      --db /relayer/database
+      --relayChains arbitrum,base,bsc,ethereum,optimism
+      --defaultSigner.key $DEFAULT_SIGNER_KEY
+      > /relayer/logs/relayer-$(date +%Y-%m-%d-%H-%M-%S).log 2>&1 && tail -f /dev/null"
+    ports:
+      - "9090:9090"
+    volumes:
+      - /tmp/relayer:/relayer
+    environment:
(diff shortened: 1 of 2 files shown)


rust/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/zsh
+
+source ~/.zshrc
+
+export HYP_CHAINS_ARBITRUM_CUSTOMRPCURLS=$(rpcs mainnet3 arbitrum)
+export HYP_CHAINS_BASE_CUSTOMRPCURLS=$(rpcs mainnet3 base)
+export HYP_CHAINS_BSC_CUSTOMRPCURLS=$(rpcs mainnet3 bsc)
+export HYP_CHAINS_ETHEREUM_CUSTOMRPCURLS=$(rpcs mainnet3 ethereum)
+export HYP_CHAINS_OPTIMISM_CUSTOMRPCURLS=$(rpcs mainnet3 optimism)
```
paulbalaji, line 9: link to the `rpcs` rune in notion as well - since it's new and some people may confuse it with the other `rpc` rune

rust/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/zsh
+
```
paulbalaji, line 2: would appreciate some quick comments on expected usage as well - i see `DEFAULT_SIGNER_KEY` in the compose but can't see where we're configuring it

paulbalaji: ah just saw your slack message - ignore this

rust/docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
+services:
+  relayer-service:
+    image: gcr.io/abacus-labs-dev/hyperlane-agent:8117739-20250420-201559
+    command: >
+      /bin/sh -c "./relayer
+      --db /relayer/database
+      --relayChains arbitrum,base,bsc,ethereum,optimism
```
daniel-savu, line 7: would be useful to make the `relayChains` and gcr image tag configurable via env vars

daniel-savu (approved):

PR #5996, opened 2025-04-23
fix: Report all operations as failed when we are failing a batch

ameten:
### Description

When there is a single successful operation in a batch and decide to fail the batch as a whole, currently, we will reports only failed operations. We should include the successful operation into the failed list.

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
+++ b/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
@@ -366,6 +366,7 @@ where
         let batch = multicall::batch::<_, ()>(multicall, contract_calls.clone());
         let call_results = batch.call().await?;
 
+        let call_count = contract_calls.len();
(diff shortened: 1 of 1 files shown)


rust/main/agents/relayer/src/msg/op_batch.rs
```diff
@@ -51,7 +51,7 @@ impl OperationBatch {
 
             if mailbox.supports_batching() {
                 warn!(excluded_ops=?excluded_ops, "Either operations reverted in the batch or the txid wasn't included. Sending them back to prepare queue.");
-                let reason = ReprepareReason::ErrorEstimatingGas;
```
daniel-savu, line 54: this should stay as is, otherwise we get high urgency errors. `ErrorSubmitting` should be set only when simulation was successful but submission still failed

daniel-savu (approved):

PR #6000, opened 2025-04-23
fix: Submit failed operations from a batch in a serial manner

ameten:
### Description

* Submit failed operations from a batch in a serial manner
* Allow batches to contains a single successful operation

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/rust/main/agents/relayer/src/msg/op_batch.rs
+++ b/rust/main/agents/relayer/src/msg/op_batch.rs
@@ -4,11 +4,11 @@ use derive_new::new;
 use hyperlane_core::{
     rpc_clients::DEFAULT_MAX_RPC_RETRIES, total_estimated_cost, BatchResult,
     ChainCommunicationError, ChainResult, ConfirmReason, HyperlaneDomain, Mailbox,
-    PendingOperation, PendingOperationStatus, QueueOperation, ReprepareReason, TxOutcome,
+    PendingOperation, PendingOperationStatus, QueueOperation, TxOutcome,
 };
 use itertools::{Either, Itertools};
 use tokio::time::sleep;
-use tracing::{error, info, instrument, warn};
+use tracing::{info, instrument, warn};
 
 use super::{
     op_queue::OpQueue,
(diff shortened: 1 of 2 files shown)


tkporter (commented):
lgtm but deferring to Dan

rust/main/agents/relayer/src/msg/op_batch.rs
```diff
@@ -43,26 +43,11 @@ impl OperationBatch {
...
-                    prepare_queue.push(op, status.clone()).await;
-                }
-            } else {
-                info!(excluded_ops=?excluded_ops, "Chain does not support batching. Submitting serially.");
-                OperationBatch::new(excluded_ops, self.domain)
-                    .submit_serially(prepare_queue, confirm_queue, metrics)
-                    .await;
-            }
+        if !excluded_ops.is_empty() {
+            warn!(excluded_ops=?excluded_ops, "Either operations reverted in the batch or the txid wasn't included. Submitting them serially.");
```
daniel-savu, line 47: let's reuse the original warning message for consistent log filtering:
```suggestion
            warn!(excluded_ops=?excluded_ops, "Either operations reverted in the batch or the txid wasn't included. Falling back to serial submission.");
```

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -366,26 +366,19 @@ where
...
 
         if successful.is_empty() {
             return Ok(BatchSimulation::failed(failed.len()));
         }
 
-        let successful_calls_len = successful.len();
         let successful_batch = multicall::batch::<_, ()>(multicall, successful.clone());
 
-        // only send a batch if there are at least two successful calls
-        if successful_calls_len >= 2 {
```
daniel-savu, line 380: we should probably still return the error if `successful_calls_len` is zero, to avoid submitting an empty batch onchain

daniel-savu: discussed that there's an `is_empty` check on a line above already

ameten: Discussed in real life. We don't submit empty batches.

daniel-savu (approved):

PR #6002, opened 2025-04-23
feat: test cli cross platform compatibility in ci

xeno097:
### Description

Updates the `test-cli-install` action to install the cli on macos and windows too and with different node.js version

### Drive-by changes

- None

### Backward compatibility

- YES

### Testing

- Manual


yorhodes (approved):
love how easy github makes this
we might consider moving this to the release workflow as it might be a bit more expensive and doesnt need to be run on every PR imo

.github/workflows/release.yml
```diff
@@ -10,6 +10,30 @@ on:
...
+      - name: install-hyperlane-cli
+        id: install-hyperlane-cli
+        uses: ./.github/actions/install-cli
+        with:
+          ref: ${{ github.sha }}
+
+      - name: Test run the CLI
+        run: hyperlane --version
+
   release:
```
paulbalaji, line 37: you'll want this job to require `cli-install-cross-platform-release-test` first

.github/workflows/release.yml
```diff
@@ -10,6 +10,30 @@ on:
...
+  cli-install-cross-platform-release-test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [18, 20, 21, 22, 23]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
```
paulbalaji, line 22: can remove these lines as the action defaults to this anyway

.github/workflows/release.yml
```diff
@@ -10,6 +10,30 @@ on:
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version}}
```
paulbalaji, line 26:
```suggestion
          node-version: ${{ matrix.node-version }}
```

.github/workflows/release.yml
```diff
@@ -10,6 +10,30 @@ on:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
+  cli-install-cross-platform-release-test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
paulbalaji, line 16: i just prefer using depot for anything beyond `ubuntu-latest` https://depot.dev/docs/github-actions/runner-types
```suggestion
        os: [depot-ubuntu-latest, depot-macos-latest, depot-windows-2022]
```

.github/workflows/release.yml
```diff
@@ -10,6 +10,30 @@ on:
...
+        node-version: [18, 20, 21, 22, 23]
```
paulbalaji, line 17: why no 19 - also is 24 out yet i can't recall

xeno097: I'll be honest as i never used v19 i completely forgot about it 🤣

xeno097: btw no v24 yet https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch

paulbalaji (approved):

PR #6004, opened 2025-04-23
chore: upgrade prettier

antigremlin:
### Description

This PR upgrades Prettier and the `@trivago/prettier-plugin-sort-imports` plugin to support the now-standard JSON import syntax. This upgrade is required for another PR that transitions to the new import syntax.

The CI job that runs Prettier has been updated to check the formatting of only the files modified by the current PR. When this job runs on pushes to the `main` branch, it verifies all files without raising errors. The formatting check results will still appear as warnings in the annotations section of the job run page.

### Related Issues

Resolves ENG-1335.

### Backward Compatibility

This change is not entirely backward compatible, as the latest Prettier version introduces different formatting in some cases. The updated formatter will only apply to modified files. Remaining formatting inconsistencies will be addressed in a future PR.


.github/workflows/test.yml
```diff
@@ -77,14 +79,30 @@ jobs:
...
-          fi
+          gh pr view ${{ github.event.pull_request.number }} --json files -q '.files[] | select (.status != "removed") | .path' > changed_files.txt
+          echo "Changed files: $(cat changed_files.txt)"
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Run prettier (changed files)
+        if: github.event_name == 'pull_request'
+        run: >-
+          cat changed_files.txt |
```
antigremlin, line 93: Only verify formatting of files changed in this PR

.github/workflows/test.yml
```diff
@@ -77,14 +79,30 @@ jobs:
...
+          xargs -r yarn exec prettier --check --ignore-unknown
+
+      - name: Run prettier (all files)
+        if: github.event_name != 'pull_request'
+        run: >-
+          yarn exec prettier --check --ignore-unknown .
```
antigremlin, line 99: Verify all files but don't flag as a failure. A root-level recursive prettier invocation works much faster.

.github/workflows/test.yml
```diff
@@ -65,8 +65,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          # check out full history
-          fetch-depth: 0
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
paulbalaji, line 71: this says 20 but we use 18 in the yarn-install step - should we consolidate? also in the yarn-install we setup before checking out, i don't mind which order we do it but would prefer consistency

solidity/package.json
```diff
@@ -90,7 +90,7 @@
     "slither": "slither .",
     "storage": "./storage.sh",
     "version:update": "sh ./bytecodeversion.sh",
-    "version:changed": "yarn version:update && git diff --exit-code",
+    "version:changed": "yarn version:update && git diff --exit-code contracts/PackageVersioned.sol",
```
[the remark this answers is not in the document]
paulbalaji: what's the implication of this, given the PackageVersioned.sol is only updated by the release PRs e.g. https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6018

paulbalaji (commented):
mostly lgtm

.github/workflows/test.yml
```diff
@@ -235,7 +252,7 @@ jobs:
         run: |
           cd typescript/infra
           yarn tsx ./scripts/agents/update-agent-config.ts -e ${{ matrix.environment }}
-          CHANGES=$(git status -s)
+          CHANGES=$(git status -s .)
```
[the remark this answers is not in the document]
paulbalaji: also the actual action we're checking here is the `update-agent-config` script causes any diff in the rust/main/config dir https://github.com/hyperlane-xyz/hyperlane-monorepo/tree/main/rust/main/config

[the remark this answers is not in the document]
paulbalaji: also the actual action we're checking here is the `update-agent-config` script causes any diff in the rust/main/config dir https://github.com/hyperlane-xyz/hyperlane-monorepo/tree/main/rust/main/config

.github/workflows/test.yml
antigremlin: I used 20 because that's what is set in the `.nvmrc` file. Maybe we could start sourcing the version from that file as it directs our dev machine workflow. I'll create a separate PR for this.

solidity/package.json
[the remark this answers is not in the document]
antigremlin: I don't fully understand why we're doing this here, but this `version:changed` script regenerates `PackageVersioned.sol` and checks if it's the same as committed to git. My change is to `git diff` only that file. It was necessary because the changes to prettier and a weirdness in the cosmos-types build process (which I don't want to change in this PR) lead to unrelated changes showing up in `git diff`.

.github/workflows/test.yml
paulbalaji: can you also update the Dockerfile as a drive-by with that PR please 🙏 https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6101959f7799faa1e92056fb4afb39e3edf8d867/Dockerfile#L1

antigremlin: There's a built-in way of reading the version from `.nvmrc` -- I'm testing if it works.

paulbalaji (approved):

Issue #3850, opened 2024-05-27
Validator doesn't announce without funds on the validator key (but the signer)


nambrot:
## Problem

**Is your feature request related to a problem? Please describe.**
Validators can specify a separate chain signer key which submits the tx on-chain, so that the validator key never needs to have funds on-chain. However, that functionality is broken right now, as the validator requires funds on the validator key itself, even if there is enough on the chain signer. 

**Additional context**
Looks like the issue is here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f2cb6ebc1b75612cb22b5135dd0ad1f133b3a197/rust/chains/hyperlane-ethereum/src/contracts/validator_announce.rs#L152

## Solution

**Describe the solution you'd like**
Validators should be able to announce with just funds on the chain signer

hydrogenbond007:
Hey @nambrot  didnt fully understand but do you want the announcement to happen from validator signer account? I think then we could just make new txn for that 


```
 let tx = contract_call
        .from(eth_h160)
        .send()
        .await?;

    let receipt = tx.await?;
    Ok(receipt.into())
```

avious00:
hey @daniel-savu or @tkporter could you help clarify for a new community contributor?

nambrot:
The announcement should happen from the signer indeed if specified, its just that we look up the balance of the validator key

avious00:
assigned to @hydrogenbond007 btw

avious00:
hey @hydrogenbond007 any update here

avious00:
hey @hydrogenbond007 will need to ship a fix for this soon, can you wrap up your WIP?

hydrogenbond007:
will be pushing new change in a day or two, sorry been afk because of travel

PR #6007, opened 2025-04-23
feat: validator announce uses chain signer instead of validator key

kamiyaa:
### Description

Use the chain's signer instead of validator key when checking balance.
- updated `announce_tokens_needed` to take a `chain_signer: H256,` parameter, to be used for checking balance instead of validator key
- add `address_h256()` to `ChainSigner` trait

### Related issues

 - fixes: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3850
 - fixes: https://linear.app/hyperlane-xyz/issue/BACK-178/validator-doesnt-announce-without-funds-on-the-validator-key-but-the

### Backward compatibility

Yes

### Testing

Add tests


rust/main/chains/hyperlane-ethereum/src/contracts/validator_announce.rs
```diff
@@ -159,7 +163,11 @@ where
             return None;
         };
 
-        let Ok(balance) = self.provider.get_balance(eth_h160, None).await else {
+        let chain_signer = chain_signer
+            .and_then(|s| ethers::types::H160::from_str(&s).ok())
```
kamiyaa, line 167: Not sure if we should fallback to validator key or force to always use chain signer by making chain_signer not an `Option`

rust/main/chains/hyperlane-ethereum/src/contracts/validator_announce.rs
```diff
@@ -150,7 +150,11 @@ where
     }
 
     #[instrument(ret, skip(self))]
-    async fn announce_tokens_needed(&self, announcement: SignedType<Announcement>) -> Option<U256> {
+    async fn announce_tokens_needed(
+        &self,
+        announcement: SignedType<Announcement>,
+        chain_signer: Option<String>,
```
tkporter, line 156: my other suggestion would've been to have this be an Option<H256> and add a new fn to `ChainSigner` to get the `address_h256` to avoid needing to go into String and back, but I think this is probably fine

rust/main/chains/hyperlane-ethereum/src/contracts/validator_announce.rs
tkporter: I guess might as well make it not optional?

tkporter (approved):

PR #6009, opened 2025-04-24
fix: broadcast channel send only unique tx

### LINKED ISSUE #5618: ContractSync's broadcast channel should only send unique tx ids for the current batch of logs
Recall we have a mechanism sometimes called "hook indexing" where if we index a dispatched message, we send the responsible tx id via a channel to the other indexing tasks.

An interesting edge case: if a single transaction sends many messages, we end up sending that transaction id multiple times via the broadcast channel. So we end up doing much more work than we need to

Example log: https://cloudlogging.app.goo.gl/t1tdmfPVQJz4bjRQ8 and how we then try to process the tree insertion multiple times: https://cloudlogging.app.goo.gl/LvuGFk7r9fzDJuF57

Suggested fix: before sending txs here, https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/d6ddf5b9ed165cab113099dca078c915df374e75/rust/main/hyperlane-base/src/contract_sync/mod.rs#L219 dedupe the transaction IDs using a set

yjamin:
### Description
Only send unique TX Ids via the channel when hook indexing. 

### Drive-by changes
None

### Related issues
[Linear](https://linear.app/hyperlane-xyz/issue/BACK-71/contractsyncs-broadcast-channel-should-only-send-unique-tx-ids-for-the)
Closes #5618

### Backward compatibility
Yes

### Testing
Local E2E testing


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -222,8 +222,15 @@ where
                 );
 
                 if let Some(tx) = self.broadcast_sender.as_ref() {
+                    // Deduplicate txids in logs to only send unique txids
```
daniel-savu, line 225: nit - would slightly reword, since it may be hard to understand what's happening without more context: "If multiple logs occur in the same transaction they'll have the same `transaction_id`. Deduplicate their txids to avoid doing wasteful queries in txid indexer"

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -222,8 +222,15 @@ where
...
+                    let mut unique_txids = HashSet::new();
                     for (_, meta) in &logs {
-                        if let Err(err) = tx.send(meta.transaction_id).await {
+                        unique_txids.insert(meta.transaction_id);
+                    }
```
daniel-savu, line 229: could do
```suggestion
                    let unique_txids: HashSet<_> =
                        logs.iter().map(|(_, meta)| meta.transaction_id).collect();
```

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -222,8 +222,15 @@ where
...
+
+                    // Send unique txids to the receiver
```
daniel-savu, line 231: nit: I think the variable naming (`unique_txids`) is suggestive enough to not warrant this comment

daniel-savu (approved):
👌

ameten (approved):

yjamin:
There are some issues with the Sealevel E2E tests, will look into this

ameten (approved):

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
```diff
@@ -127,17 +129,24 @@ pub fn relayer_termination_invariants_met(
...
         .expect("Failed to get tx id indexing log count");
+
+    // there are 3 txid-indexed events:
+    // - relayer: merkle insertion and gas payment
+    // - scraper: gas payment
+    // some logs are emitted for multiple events, so requiring there to be at least
+    // `config.kathy_messages` logs is a reasonable approximation, since all three of these events
+    // are expected to be logged for each message.
+    // Note: sealevel relayer does not require tx id indexing, so we don't check for it
+    let expected_td_id_logs = if solana_tx_id_indexing {
```
daniel-savu, line 140: can you pls add a comment about why the expected tx_id logs for solana is zero

rust/main/utils/run-locally/src/invariants/termination_invariants.rs
```diff
@@ -127,17 +129,24 @@ pub fn relayer_termination_invariants_met(
     let total_tx_id_log_count = *log_counts
         .get(&tx_id_indexing_line_filter)
         .expect("Failed to get tx id indexing log count");
+
+    // there are 3 txid-indexed events:
```
daniel-savu, line 133: this comment block should probably go above the assert with `total_tx_id_log_count as u64 >= expected_td_id_logs,`

PR #6012, opened 2025-04-24
refactor(agents): CosmWasm <> CosmosNative deduplication

yjamin:
### Description
Merged CosmWasm & CosmosNative implementations. CosmWasm and CosmosNative now share a lot of chain logic when it comes to:
- chain communication
- Event Fetching / Parsing
- Connection Config
- Account Structs & Signer

Structure overview:
CosmWasm and CosmosNative now share one crate which implements both Protocol Types. However there is still a logical distinction necessary between `HyperlaneDomainProtocol::CosmosNative` & `HyperlaneDomainProtocol::Cosmos`.

Provider logic is now shared between implementations and split into two categories: 
- **RPC:** that handles general chain communication like Tx submission, estimation, Block & Event Fetching etc.
- **GRPC:** implementation specific query client that either queries contract states for CW or module state for CosmosNative.

```
hyperlane-cosmos/                      
└── src/                               
    ├── application/                   # Untouched - Core application logic
    ├── cw/                            # Previous CW Contracts, mostly the same except query client
    │   └── cw_query_client.rs         # Query client for CosmWasm contracts
    ├── indexer.rs                     # Event Indexer helper for both CW & CosmosNative
    ├── libs/                          # libs shared between both implementations
    ├── native/                        # CosmosNative Contracts, mostly the same except query client
    │   └── module_query_client.rs     # Client for querying native modules
    ├── providers/                     # Generic providers that can be used for both CW & CosmosNative
    │   ├── cosmos.rs                  # Cosmos chain provider
    │   ├── grpc.rs                    # the query for each implementation client builds on top of this
    │   └── rpc.rs                     # Same for CW & CosmosNative. Handles raw chain communication like: TX Submission etc.
    └── utils.rs                       
```

### Related issues
[Linear](https://linear.app/hyperlane-xyz/issue/BACK-125/remove-duplication-in-cw-and-cosmos-native)

### Backward compatibility
Full

### Testing
Local E2E Tests


rust/main/chains/hyperlane-cosmos/src/cw/cw_query_client.rs
```diff
@@ -0,0 +1,321 @@
...
+            .filter(|a| a.type_url == "/cosmwasm.wasm.v1.MsgExecuteContract")
+            .cloned()
+            .collect::<Vec<Any>>();
+
+        let contract_execution_messages_len = contract_execution_messages.len();
+        if contract_execution_messages_len > 1 {
+            let msg = "transaction contains multiple contract execution messages";
+            Err(HyperlaneCosmosError::ParsingFailed(msg.to_owned()))?
+        }
+
```
yjamin, line 110: In cosmos you can have multiple messages in a single transactions. Meaning there could be more than one `RemoteTransfer` in one transaction and it would still be valid.  We currently don't account for that and throw an error, which I think is wrong.

Will this be fixed with the submitter redesign / how do we deal with that on the ethereum side with a multicall?

Btw this applies to both `cosmosnative` and `cosmwasm`

rust/main/chains/hyperlane-cosmos/src/cw/validator_announce.rs
```diff
@@ -89,18 +87,19 @@ impl ValidatorAnnounce for CosmosValidatorAnnounce {
                 signature: hex::encode(announcement.signature.to_vec()),
             },
         };
+        let payload = self.provider.query().wasm_encode_msg(announce_request)?;
 
-        let response: TxResponse = self
-            .provider
-            .grpc()
-            // TODO: consider transaction overrides for Cosmos.
-            .wasm_send(announce_request, None)
```
yjamin, line 97: Is this still a TODO?

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -1,5 +1,452 @@
...
+    ) -> ChainResult<SignDoc> {
+        // As this function is only used for estimating gas or sending transactions,
+        // we can reasonably expect to have a signer.
+        let signer = self.get_signer()?;
+        let account_info = self.get_account(signer.address.clone()).await?;
+
+        // timeout height of zero means that we do not have a timeout height TODO: double check
+        let tx_body = tx::Body::new(msgs, String::default(), 0u32);
+        let signer_info = SignerInfo::single_direct(Some(signer.public_key), account_info.sequence);
+
```
yjamin, line 343: Did we ever run into TX Timeout issues? Might still be wise to set a responsible timeout even if thats 1000 blocks in the future.

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -1,5 +1,452 @@
...
+        let signed_tx = signed_tx.to_bytes()?;
+
+        // broadcast tx commit blocks until the tx is included in a block
+        self.provider
+            .call(|client| {
+                let signed_tx = signed_tx.clone(); // TODO: this is very memory inefficient, we should actually be able to just reference the slice
+                let future = async move {
+                    Self::track_metric_call(&client, "send", || {
+                        client.client.broadcast_tx_commit(signed_tx.clone())
+                    })
```
yjamin, line 433: `broadcast_tx_commit` should actually wait for that TX to be mined, and fix nonce issues. However we still run into some nonce issues in the E2E tests, implying it does not. We should probably add extra logic that fetches the TX state to and waits for confirmation. 

I'm unsure if this will be required for the Submitter anyways, meaning it wont make sense to implement right now.

rust/main/chains/hyperlane-cosmos/src/utils.rs
```diff
@@ -1,122 +1,103 @@
...
+    Ok(public_key_and_account_address_type)
+}
+
+pub(crate) fn tx_response_to_outcome(response: Response, gas_price: FixedPointNumber) -> TxOutcome {
+    // TODO: check if gas price is the literal price per gas unit or how much we paid in tokens for gas
+    // rn we assume that the underlying cosmos chain does not have gas refunds
+    // in that case the gas paid will always be:
+    // gas_wanted * gas_price
+    let gas_price = FixedPointNumber::from(response.tx_result.gas_wanted) * gas_price;
+
```
yjamin, line 96: Would be nice to get clarification about the question above.

rust/main/hyperlane-base/src/settings/parser/connection_parser.rs
```diff
@@ -186,11 +187,18 @@ pub fn build_cosmos_connection_conf(
 
     let native_token = parse_native_token(chain, err, 18);
 
+    let gas_multiplier = chain
+        .chain(err)
+        .get_opt_key("gasMultiplier")
+        .parse_f64()
+        .end()
+        .unwrap_or(1.35);
+
```
yjamin, line 196: This was a constant before, I think we should start to include this in the config as can heavily change per chain.

rust/main/chains/hyperlane-cosmos/Cargo.toml
```diff
@@ -48,6 +48,8 @@ tower.workspace = true
 tracing = { workspace = true }
 tracing-futures = { workspace = true }
 url = { workspace = true }
+hyperlane-cosmos-rs = "0.1.4"
```
ameten, line 51: Probably, it should be moved to workspace `Cargo.toml`.

rust/main/chains/hyperlane-cosmos/src/cw/cw_query_client.rs
ameten: I think we use this logic only when we need to identify a transaction by its hash and it happens only in Scraper when we need to store a transaction for Hyperlane message and their relationship.

Since this logic is not used to submit transaction on chain, there is no impact on the submitter redesign.

Having said that, our indexing logic does require improvement so that we can support AltVMs better.

rust/main/chains/hyperlane-cosmos/src/cw/interchain_gas.rs
```diff
@@ -176,59 +176,69 @@ impl CosmosInterchainGasPaymasterIndexer {
                     }
                 }
 
-                EventAttribute::V034(a) => {
+                EventAttribute::V034(_a) => {
                     unimplemented!();
                 }
```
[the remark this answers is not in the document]
ameten: Our implementation of CosmWasm relied on obsolete version of `cosmrs` where there were no support for v0.34, but an issue with obsolete TLS certificate made up to upgrade to a newer version of `cosmrs` which introduced additional version of `EventAttribute`. Since the goal was just to resolve the issue with TLS certificate, this branch stayed unimplemented.

rust/main/chains/hyperlane-cosmos/src/cw/validator_announce.rs
ameten: We are using transaction overrides to customise gas price and gas limit on Ethereum with the goal to make transaction submission more reliable. We may consider similar approach for Cosmos chains as well. Submitter implementation for Cosmos will inform this decision.

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
ameten: I cannot recall a case when we ran into tx timeout issue, but it is a good consideration for transaction submission logic. We'll keep it in mind for submitter redesign implementation.

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
ameten: Agreed. It does not make sense to deal with it at the moment. Submitter will have its own way to deal with nonce issues.

rust/main/chains/hyperlane-cosmos/src/utils.rs
ameten: It looks like this code only exists in Cosmos Native implementation. I am not sure that `TxOutcome.gas_price` can be calculated in this way. Presumably, this calculation gives the total cost of a transaction, not a price for a unit of gas.

And I wonder what question requires clarification?

rust/main/hyperlane-base/src/settings/parser/connection_parser.rs
ameten: It makes sense.

rust/main/chains/hyperlane-cosmos/src/providers/cosmos.rs
```diff
@@ -1,3 +1,328 @@
...
+
+    /// Whether or not the given address is a contract
+    async fn is_contract(&self, address: &H256) -> ChainResult<bool>;
+
+    /// Extract the message recipient contract address from the tx
+    /// this is implementation specific
+    fn parse_tx_message_recipient(&self, tx: &Tx, hash: &H512) -> ChainResult<H256>;
+}
+
+/// Cosmos Native Provider
```
ameten, line 42: I suppose `CosmosProvider<QueryClient>` is a provider which is used in both CosmWasm and Cosmos Native.

rust/main/chains/hyperlane-cosmos/src/cw/mailbox/delivery_indexer.rs
```diff
@@ -9,41 +10,40 @@ use tendermint::abci::EventAttribute;
...
-    CONTRACT_ADDRESS_ATTRIBUTE_KEY, CONTRACT_ADDRESS_ATTRIBUTE_KEY_BASE64,
-};
-use crate::{ConnectionConf, HyperlaneCosmosError, Signer};
+use crate::cw::CwQueryClient;
+use crate::indexer::{CosmosEventIndexer, ParsedEvent};
+use crate::utils::{CONTRACT_ADDRESS_ATTRIBUTE_KEY, CONTRACT_ADDRESS_ATTRIBUTE_KEY_BASE64};
+use crate::{CosmosAddress, CosmosProvider, HyperlaneCosmosError, RpcProvider};
 
 /// The message process event type from the CW contract.
-pub const MESSAGE_DELIVERY_EVENT_TYPE: &str = "mailbox_process_id";
```
ameten, line 24: I wonder if this change means that we should adopt and redeploy CosmWasm smart contract?

rust/main/chains/hyperlane-cosmos/src/cw/mailbox/dispatch_indexer.rs
```diff
@@ -9,42 +10,47 @@ use tracing::instrument;
...
-use crate::{ConnectionConf, CosmosMailbox, HyperlaneCosmosError, Signer};
+use crate::indexer::{CosmosEventIndexer, ParsedEvent};
+use crate::utils::{CONTRACT_ADDRESS_ATTRIBUTE_KEY, CONTRACT_ADDRESS_ATTRIBUTE_KEY_BASE64};
+use crate::{cw::CwQueryClient, CosmosAddress, CosmosProvider, HyperlaneCosmosError, RpcProvider};
+
+use super::CwMailbox;
 
 /// The message dispatch event type from the CW contract.
-pub const MESSAGE_DISPATCH_EVENT_TYPE: &str = "mailbox_dispatch";
+pub const MESSAGE_DISPATCH_EVENT_TYPE: &str = "wasm-mailbox_dispatch";
```
ameten, line 23: Same question is [here](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6012/files#r2059999080)

ameten (approved):

rust/main/chains/hyperlane-cosmos/src/cw/mailbox/delivery_indexer.rs
yjamin: Logic didn't change, meaning we don't have to deploy anything different it is still the same. CosmWasm event types are prefixed with `wasm-`, which was previously deeply hidden/encoded in the parsing logic - which i think we should not do and instead prioritize readability. 

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6101959f7799faa1e92056fb4afb39e3edf8d867/rust/main/chains/hyperlane-cosmos/src/providers/rpc/provider.rs#L123

rust/main/chains/hyperlane-cosmos/src/providers/cosmos.rs
yjamin: Correct! The generic part will take core of the implementation specific queries - this way they can share a large amount of logic

rust/main/chains/hyperlane-cosmos/Cargo.toml
yjamin: Not sure why we would need these types anywhere else in the codebase, but happy to move it there ofc.

rust/main/chains/hyperlane-cosmos/src/utils.rs
yjamin: This is used to give a TxCostEstimate and I guess my question was whether or not `gas_price` in this struct is the price per gas unit or the amount of tokens paid for the tx.

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/6101959f7799faa1e92056fb4afb39e3edf8d867/rust/main/hyperlane-core/src/types/mod.rs#L260-L272

rust/main/chains/hyperlane-cosmos/src/cw/cw_query_client.rs
yjamin: I take that as we should leave it as it is right now and need to some work on the scraper side instead.

rust/main/chains/hyperlane-cosmos/src/cw/interchain_gas.rs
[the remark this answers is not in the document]
yjamin: Okay, makes sense to me 👍

rust/main/chains/hyperlane-cosmos/src/cw/validator_announce.rs
yjamin: Interesting, this sounds like the`gasMultiplier` somewhat solves this issue?

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
yjamin: True, will add a reasonable timeout height here for now.

rust/main/chains/hyperlane-cosmos/src/cw/cw_query_client.rs
ameten: Agreed

rust/main/chains/hyperlane-cosmos/src/cw/validator_announce.rs
ameten: Yep

rust/main/chains/hyperlane-cosmos/src/providers/cosmos.rs
ameten: This is just a note on the comment.

```suggestion
/// Cosmos Provider
```

rust/main/chains/hyperlane-cosmos/src/utils.rs
ameten: Judging how the field is populated for Ethereum, it is the price per gas unit: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/ae9966b86536f841a509a8d379c9a5f963fcc17f/rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs#L679.

What would make you to assume that it can be the amount of tokens paid for the transaction?

rust/main/chains/hyperlane-cosmos/src/cw/mailbox/delivery_indexer.rs
ameten: It makes sense to me. Good point.

rust/main/chains/hyperlane-cosmos/src/cw/mailbox/dispatch_indexer.rs
ameten: Resolving with this - https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6012#discussion_r2060018613

rust/main/chains/hyperlane-cosmos/src/utils.rs
yjamin: Cool, I guess I was just confused.

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
daniel-savu: good point - the edge case you pointed out does happen in prod as well unfortunately @yjamin. As @ameten said, the new submitter will allow us to send txs reliably without needing to wait for tx inclusion / finality

rust/main/chains/hyperlane-cosmos/src/providers/rpc.rs
```diff
@@ -1,5 +1,455 @@
...
+    rpc_clients::{BlockNumberGetter, FallbackProvider},
+    ChainCommunicationError, ChainResult, FixedPointNumber, H256, H512, U256,
+};
+use hyperlane_metric::prometheus_metric::{
+    ClientConnectionType, PrometheusClientMetrics, PrometheusConfig,
+};
+
+use crate::{ConnectionConf, CosmosAmount, HyperlaneCosmosError, Signer};
+
+const TX_TIEMOUT_BLOCKS: u32 = 100;
```
daniel-savu, line 40: nit: typo

rust/main/chains/hyperlane-cosmos/src/native/ism.rs
```diff
@@ -26,13 +28,16 @@ pub struct CosmosNativeIsm {
     /// The address of the ISM contract.
     address: H256,
     /// The provider for the ISM contract.
-    provider: Box<CosmosNativeProvider>,
+    provider: Box<CosmosProvider<ModuleQueryClient>>,
```
daniel-savu, line 31: nit: the `Box` isn't needed here other than for not using up stack space, wondering why you went for this here but not in e.g. `CosmosNativeMerkleTreeHook`

rust/main/chains/hyperlane-cosmos/src/native/indexers/merkle_tree_hook.rs
```diff
@@ -19,20 +19,26 @@ use hyperlane_core::{
...
-use crate::{CosmosNativeProvider, HyperlaneCosmosError, RpcProvider};
+use crate::{
+    native::module_query_client::ModuleQueryClient, CosmosProvider, HyperlaneCosmosError,
+    RpcProvider,
+};
 
-use super::{CosmosEventIndexer, ParsedEvent};
+use crate::indexer::{CosmosEventIndexer, ParsedEvent};
 
 /// delivery indexer to check if a message was delivered
```
daniel-savu, line 29: the comment is outdated

rust/main/chains/hyperlane-cosmos/src/native/indexers/interchain_gas.rs
```diff
@@ -12,17 +12,18 @@ use hyperlane_core::{
...
 use crate::{
-    ConnectionConf, CosmosEventIndexer, CosmosNativeProvider, HyperlaneCosmosError, RpcProvider,
+    native::module_query_client::ModuleQueryClient, ConnectionConf, CosmosProvider,
+    HyperlaneCosmosError, RpcProvider,
 };
 
-use super::ParsedEvent;
+use crate::indexer::{CosmosEventIndexer, ParsedEvent};
 
 /// delivery indexer to check if a message was delivered
```
daniel-savu, line 21: ditto

daniel-savu (approved):
very nice refactoring work 👏 there's a ton of code changes, which make this PR very hard to review. Nothing stood out to me, but I'd like a gradual release of this since it could be breaking: first to one RC chain, then all RC cosmos chains, and only then to the hyperlane context

rust/main/chains/hyperlane-cosmos/src/native/validator_announce.rs
```diff
@@ -91,18 +95,10 @@ impl ValidatorAnnounce for CosmosNativeValidatorAnnounce {
...
-        let gas_price =
-            FixedPointNumber::from(response.tx_result.gas_wanted) * self.provider.rpc().gas_price();
-
-        Ok(TxOutcome {
-            transaction_id: H256::from_slice(response.hash.as_bytes()).into(),
-            executed: response.check_tx.code.is_ok() && response.tx_result.code.is_ok(),
-            gas_used: response.tx_result.gas_used.into(),
-            gas_price,
-        })
+        Ok(utils::tx_response_to_outcome(
```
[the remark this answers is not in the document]
yjamin: Nope, that was a wrong calculation.

rust/main/chains/hyperlane-cosmos/src/trait_builder.rs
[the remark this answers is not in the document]
yjamin: It was named the same for cosmos & other implementations like fuel.

rust/main/chains/hyperlane-cosmos/src/native/mailbox.rs
```diff
@@ -84,8 +86,8 @@ impl Mailbox for CosmosNativeMailbox {
         let height = self.provider.reorg_to_height(reorg_period).await?;
```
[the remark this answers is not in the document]
yjamin: Yeah it it. Funny enough that comment is actually from the trait itself. See here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2faf54edd961f7c738699333f3aefbf765cfb6ef/rust/main/hyperlane-core/src/traits/mailbox.rs#L20-L24

rust/main/chains/hyperlane-cosmos/src/native/ism.rs
yjamin: Good point, that was just an artifact from the refactoring.

PR #6016, opened 2025-04-24
feat: Store all payload ids for message id

ameten:
### Description

* Store all payload ids for message id
* Store as a vector
* Extend the vector with new payload ids

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5892

### Backward compatibility

Yes

### Testing

Unit test for encoding/decoding

### DIFF
--- a/rust/main/agents/relayer/src/msg/op_submitter.rs
+++ b/rust/main/agents/relayer/src/msg/op_submitter.rs
@@ -516,9 +516,9 @@ async fn submit_via_lander(
         return;
     }
 
-    if let Err(e) = db.store_payload_id_by_message_id(&message_id, &payload.details.id) {
-        let reason = ReprepareReason::ErrorStoringPayloadIdByMessageId;
-        let msg = "Error storing mapping from message id to payload id";
+    if let Err(e) = db.store_payload_ids_by_message_id(&message_id, &vec![payload.details.id]) {
+        let reason = ReprepareReason::ErrorStoringPayloadIdsByMessageId;
+        let msg = "Error storing mapping from message id to payload ids";
         prepare_op(op, prepare_queue, e, msg, reason).await;
         return;
     }
@@ -668,7 +668,7 @@ async fn confirm_lander_task(
             .into_iter()
             .map(|op| {
                 let message_id = op.id();
-                (op, db.retrieve_payload_id_by_message_id(&message_id))
+                (op, db.retrieve_payload_ids_by_message_id(&message_id))
             })
             .collect::<Vec<_>>();
 
--- a/rust/main/agents/relayer/src/msg/op_submitter.rs
+++ b/rust/main/agents/relayer/src/msg/op_submitter.rs
@@ -677,14 +677,19 @@ async fn confirm_lander_task(
             .map(|(op, result)| async {
                 let message_id = op.id();
                 match result {
-                    Ok(Some(payload_id)) => Some((op, entrypoint.payload_status(payload_id).await)),
-                    Ok(None) | Err(_) => {
+                    Ok(Some(ids)) if !ids.is_empty() => Some((
+                        op,
+                        entrypoint
+                            .payload_status(ids.into_iter().last().unwrap())
+                            .await,
+                    )),
+                    Ok(Some(_)) | Ok(None) | Err(_) => {
                         error!(
                             ?op,
                             %message_id,
                             "Error retrieving payload id by message id",
                         );
-                        send_back_on_failed_submisison(
+                        send_back_on_failed_submission(
                             op,
                             prepare_queue.clone(),
                             &metrics,
@@ -709,7 +714,7 @@ async fn confirm_lander_task(
             .map(|(op, status_result)| async {
                 let Ok(payload_status) = status_result else {
                     warn!(?op, "Error retrieving payload status",);
-                    send_back_on_failed_submisison(
+                    send_back_on_failed_submission(
                         op,
                         prepare_queue.clone(),
                         &metrics,
@@ -790,7 +795,7 @@ async fn process_confirm_result(
                 .await;
         }
         PendingOperationResult::Reprepare(reason) => {
-            send_back_on_failed_submisison(op, prepare_queue.clone(), &metrics, Some(reason)).await;
+            send_back_on_failed_submission(op, prepare_queue.clone(), &metrics, Some(reason)).await;
         }
         PendingOperationResult::Drop => {
             metrics.ops_dropped.inc();
@@ -800,7 +805,7 @@ async fn process_confirm_result(
     operation_result
 }
 
-async fn send_back_on_failed_submisison(
+async fn send_back_on_failed_submission(
     op: QueueOperation,
     prepare_queue: OpQueue,
     metrics: &SerialSubmitterMetrics,
--- a/rust/main/agents/relayer/src/msg/op_submitter.rs
+++ b/rust/main/agents/relayer/src/msg/op_submitter.rs
@@ -7,7 +7,6 @@ use std::time::Duration;
 
 use futures::future::join_all;
 use futures_util::future::try_join_all;
-use itertools::{Either, Itertools};
 use num_traits::Zero;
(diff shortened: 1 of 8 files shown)


rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -677,14 +677,19 @@ async fn confirm_lander_task(
...
                 let message_id = op.id();
                 match result {
-                    Ok(Some(payload_id)) => Some((op, entrypoint.payload_status(payload_id).await)),
-                    Ok(None) | Err(_) => {
+                    Ok(Some(ids)) if !ids.is_empty() => Some((
+                        op,
+                        entrypoint
+                            .payload_status(ids.into_iter().last().unwrap())
+                            .await,
+                    )),
```
daniel-savu, line 685: IMO all payloads linked to a message must be confirmed, as a requirement for the message to also count as confirmed. So here we should check the status for all payloads rather than just the last one

rust/main/agents/validator/src/submit.rs
```diff
@@ -588,8 +588,8 @@ mod test {
             ) -> DbResult<Option<u64>>;
             fn store_highest_seen_message_nonce_number(&self, nonce: &u32) -> DbResult<()>;
             fn retrieve_highest_seen_message_nonce_number(&self) -> DbResult<Option<u32>>;
-            fn store_payload_id_by_message_id(&self, message_id: &H256, payload_id: &UniqueIdentifier) -> DbResult<()>;
-            fn retrieve_payload_id_by_message_id(&self, message_id: &H256) -> DbResult<Option<UniqueIdentifier>>;
+            fn store_payload_id_by_message_id(&self, message_id: &H256, payload_id: &UniqueIdentifier) -> DbResult<Vec<UniqueIdentifier>>;
```
daniel-savu, line 591: wondering why this would be helpful, the pattern for the `store` methods so far has been to return `()`

rust/main/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -663,14 +663,19 @@ impl HyperlaneDb for HyperlaneRocksDB {
         &self,
         message_id: &H256,
         payload_id: &UniqueIdentifier,
-    ) -> DbResult<()> {
-        self.store_value_by_key(PAYLOAD_ID_BY_MESSAGE_ID, message_id, payload_id)
+    ) -> DbResult<Vec<UniqueIdentifier>> {
+        let mut ids = self
+            .retrieve_payload_id_by_message_id(message_id)?
+            .unwrap_or_default();
+        ids.push(payload_id.clone());
```
daniel-savu, line 670: I would instead rename this function to `store_payload_ids_by_message_id`, take in a vector of `UniqueIdentifier`, and write to the DB without reading first. All payloads associated with a message would be known from when the message is first processed, so there'd be no need to add extra payloads for a message later

rust/main/hyperlane-core/src/traits/encode.rs
```diff
@@ -303,8 +303,45 @@ impl<T: Decode> Decode for Indexed<T> {
...
+        W: std::io::Write,
+    {
+        let mut written = 0;
+        // Write the length of the vector as a u32
+        written += (self.len() as u64).write_to(writer)?;
+
+        // Write each `T` in the vector using its `Encode` implementation
+        written += self.iter().try_fold(0, |acc, item| {
+            item.write_to(writer).map(|bytes| acc + bytes)
+        })?;
```
daniel-savu, line 318: nice, I would've been ok with just encoding as json as well to avoid spending time on the implementation

rust/main/agents/validator/src/submit.rs
ameten: User of all other `store_` methods has everything available to them at the time they invoke the method. It is not true `store_payload_id_by_message_id` method: user has only `message_id` and last `payment_id`. So, it can be useful to return the whole list of payment ids. And it does not cost much since no copy is happening there.

rust/main/hyperlane-core/src/traits/encode.rs
ameten: ChatGPT helps a lot with such implementations. :)

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: It is part of the next task: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5894

rust/main/hyperlane-base/src/db/rocks/hyperlane_db.rs
ameten: Let me see how [this task](https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5894) should be implemented and understand if we need to change it.

ameten: I see what you mean. We shall create several payloads for the same message and we don't need them anymore if message is returned back to prepare queue.

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: Fixed

rust/main/agents/validator/src/submit.rs
ameten: Since payload ids will be known at the time when we store them, different return time is not needed. Fixed.

rust/main/hyperlane-base/src/db/rocks/hyperlane_db.rs
ameten: Fixed

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -711,8 +715,16 @@ async fn confirm_lander_task(
...
+            .map(|(op, status_results)| async {
+                let (successes, failures): (Vec<_>, Vec<()>) =
+                    status_results.into_iter().partition_map(|status| {
+                        if let Ok(payload_status) = status {
+                            return Either::Left(payload_status);
+                        }
+                        Either::Right(())
+                    });
+
+                if !failures.is_empty() {
```
daniel-savu, line 727: - since we don't actually use `failures` for anything other than their count, we could take the initial length of `status_results` and use `filter` for `Ok` responses instead of `partition_map`. The failures would be `status_results_length - filtered_results_length` and that might make the code here less verbose
- but then maybe we want to log the payload IDs that are still pending, in the `Operation not finalized yet` log, wdyt? If so, then I'd just move this `payload_status_results` partitioning logic into its own function to lower verbosity here

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -677,14 +679,21 @@ async fn confirm_lander_task(
...
                 let message_id = op.id();
                 match result {
-                    Ok(Some(payload_id)) => Some((op, entrypoint.payload_status(payload_id).await)),
-                    Ok(None) | Err(_) => {
+                    Ok(Some(ids)) if !ids.is_empty() => {
+                        let op_futures = ids
+                            .into_iter()
+                            .map(|id| entrypoint.payload_status(id))
+                            .collect::<Vec<_>>();
+                        let op_results = join_all(op_futures).await;
```
daniel-savu, line 687: noting that this join_all will block synchronously on rocksdb under the hood, for each payload - but nothing we can improve about this

rust/main/hyperlane-core/src/traits/encode.rs
daniel-savu: haha i see, as long as you're confident it's ok 👍 it does lgtm

daniel-savu (approved):

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: I adopted the proposal.

PR #6022, opened 2025-04-25
chore: cosmos addresses

troykessler:
### Description

This PR improves the way cosmos addresses are displayed. In cosmos there is like in Ethereum a private public key pair, but for every cosmos chain there is a different address prefix. Example Osmosis `osmo1sduu...`, Cosmos Hub `cosmos1j48f33..`, KYVE `kyve1jhf738...` and so on. Right now always the address of the first chain gets displayed which often has nothing to do with the chain the user has funds on and wants to swap from/to. This has caused confusion in the past.

This PR addresses this issue by showing the Cosmos Hub address by default if the user is connected with a cosmos wallet since Cosmos Hub is still the standard in the cosmos ecosystem. Once the user selects a cosmos origin chain that address is displayed instead.

This PR can be tested by checking out the Warp UI PR https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/536

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Testing with Warp UI


typescript/widgets/src/walletIntegrations/ConnectWalletButton.tsx
```diff
@@ -34,6 +40,11 @@ export function ConnectWalletButton({
 
   const numReady = readyAccounts.length;
   const firstAccount = readyAccounts[0];
+
+  const firstAddress = shortenAddress(
+    getAddressFromAccountAndChain(firstAccount, chainName),
+  );
```
Xaroz, line 46: I think here it would be more accurate to use something like
`shownAddress` or just `address`

Xaroz: also seems like before shortenAddress was called by a boolean, is that not necessary anymore?

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
```diff
@@ -127,14 +128,55 @@ export function getAccountAddressForChain(
...
+    if (cosmosAddress) {
+      // set cosmos address if one has been found for the chain name
+      address = cosmosAddress;
+    } else {
+      // search for the cosmos hub address if no address has been found
+      // for the chain name
+      address =
+        account?.addresses?.find((a) => a.chainName === cosmoshub.name)
+          ?.address ?? 'Unknown';
+    }
```
Xaroz, line 174: nit: is it possible to rely on the `multiProvider` instead of the registry for `cosmohub`? Just to keep things consistent

Xaroz (commented):
Mostly looks good

typescript/widgets/src/walletIntegrations/ConnectWalletButton.tsx
troykessler: forgot the boolean, good catch!

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
troykessler: what exactly do you want to do with the `multiProvider` here?

Xaroz: I guess from my understanding `multiProvider` would only include a `ChainMap` with existing chains, if we did `cosmohub.name` directly from the registry we wouldn't know if it actually exists unless we did something like `multiProvider.tryGetChainName()`

But you can just ignore this, seems unnecessary

Xaroz (approved):

paulbalaji (approved):

paulbalaji:
`yarn test` will be fixed when pulling main after either one of the following PRs gets merged:
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6036
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6039

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
```diff
@@ -127,14 +128,55 @@ export function getAccountAddressForChain(
...
+  ) {
     return account?.addresses.find((a) => a.chainName === chainName)?.address;
   } else {
     // Use first because only cosmos has the notion of per-chain addresses
     return account?.addresses[0]?.address;
   }
 }
 
+export function getAddressFromAccountAndChain(
+  account?: AccountInfo,
```
nambrot, line 143: When would these values be undefined?

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
```diff
@@ -127,14 +128,55 @@ export function getAccountAddressForChain(
...
+  }
+
+  // by default display the first address of the account
+  let address = account.addresses[0]?.address ?? 'Unknown';
+
+  // only in cosmos there are multiple addresses per account, in this
+  // case we display the cosmos hub address by default. If the user
+  // selects a cosmos based origin chain in the swap form that cosmos
+  // address is displayed instead
+  if (account?.protocol === ProtocolType.Cosmos) {
```
nambrot, line 157: Shouldn't tyepscript infer that account is defined now?

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
```diff
@@ -127,14 +128,55 @@ export function getAccountAddressForChain(
...
+export function getAddressFromAccountAndChain(
+  account?: AccountInfo,
+  chainName?: ChainName,
+) {
+  if (!account || !chainName) {
+    return 'Unknown';
+  }
+
+  // by default display the first address of the account
+  let address = account.addresses[0]?.address ?? 'Unknown';
```
nambrot, line 151: Instead of doing `let` here, I'd rather do the if clause for cosmos first, and directly return in there, and then otherwise just return this, mutable state i always find much harder to reason about

paulbalaji:
created https://linear.app/hyperlane-xyz/issue/ENG-1360/cosmos-pr-follow-ups to resolve comments afterwards

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
troykessler: During initialization of the page when the account is not "ready" yet it can be undefined

typescript/widgets/src/walletIntegrations/multiProtocol.tsx
troykessler: Yes it should, will remove the "?" in a follow up PR

Issue #4260, opened 2024-08-06
Validator checkpoint_latest_index.json is not set if the checkpoint latest index is 0


tkporter:
## Problem

The logic here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/3d4e6436d526b94a76afd613f70199dc2f0579e8/rust/hyperlane-base/src/types/gcs_storage.rs#L146 results in the checkpoint_latest_index.json only being updated if `self.latest_index()` is > 0. It's possible for `self.latest_index()` to be 0 if only a single message has been inserted into the merkle tree

## Solution

Some better condition

PR #6024, opened 2025-04-25
feat: don't default to 0 index when comparing checkpoint indexes

kamiyaa:
### Description

- instead of defaulting to 0 when comparing checkpoint indexes, we check if its None.
- if it's None, then we write checkpoint, if not, then we do checkpoint index comparison

### Drive-by changes

 - remove GcsStorageClient override of `update_latest_index()`

### Related issues

 - fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/4260

### Backward compatibility

Yes/No

### Testing

None

### DIFF
--- a/rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
+++ b/rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
@@ -14,9 +14,15 @@ pub trait CheckpointSyncer: Debug + Send + Sync {
     async fn write_latest_index(&self, index: u32) -> Result<()>;
     /// Update the latest index of this syncer if necessary
     async fn update_latest_index(&self, index: u32) -> Result<()> {
-        let curr = self.latest_index().await?.unwrap_or(0);
-        if index > curr {
-            self.write_latest_index(index).await?;
+        match self.latest_index().await? {
+            None => {
+                self.write_latest_index(index).await?;
+            }
+            Some(curr) => {
+                if index > curr {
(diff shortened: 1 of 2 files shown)


rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
```diff
@@ -14,9 +14,15 @@ pub trait CheckpointSyncer: Debug + Send + Sync {
     async fn write_latest_index(&self, index: u32) -> Result<()>;
     /// Update the latest index of this syncer if necessary
     async fn update_latest_index(&self, index: u32) -> Result<()> {
-        let curr = self.latest_index().await?.unwrap_or(0);
-        if index > curr {
-            self.write_latest_index(index).await?;
```
ameten, line 19: I wonder if we use `-1` as default, this code will work as needed?

daniel-savu: it would! Basically the current solution uses `None` as `-1`, which I think is more idiomatic to rust

daniel-savu (approved):

rust/main/hyperlane-base/src/traits/checkpoint_syncer.rs
kamiyaa: yeah agree with @daniel-savu . if we use -1, we will have to do a bunch of casting to `i64` and back

PR #6030, opened 2025-04-25
feat: configure HYPER routes

yorhodes:
### Description

Configures the HYPER and stHYPER warp routes (plus staging variants) for `mainnet3`.

- Adds `HyperToken` Solidity contract (HypERC20 + AccessControl + ERC20Burnable + ERC20Permit) for the Ethereum collateral leg of HYPER, with `MINTER_ROLE`/`BURNER_ROLE` for controlled supply management.
- Adds infra config getter `getHyperWarpConfig` with `STAGING` and `PRODUCTION` owner sets and initial supply (mainnet3 + testnet4).
- Wires four warp route IDs (`HYPER`, `stHYPER`, `HYPERSTAGE`, `stHYPERSTAGE`) into the warp config getter map, matching the registry path conventions.
- SDK plumbing for `TokenType.hyperToken` (treated as `EvmHypSynthetic` for WarpCore, distinct deploy type via `HyperToken__factory`).

### Drive-by changes

- `HypERC20.initialize()` / `decimals()` marked `virtual` to allow `HyperToken` to override; no behavior change to existing HypERC20 deployments.
- `EvmWarpRouteReader.deriveTokenType` selector probe extended to detect HyperToken via `MINTER_ROLE` selector; previously misclassified as plain synthetic.
- \`initialSupply\` added to \`FIELDS_TO_IGNORE\` in warp check — it's a write-only constructor arg, not recoverable from chain and not a runtime invariant (drifts with mint/burn/bridging).
- HYPER family route getter pins \`contractVersion: '7.0.0'\` to match deployed bytecode.

### Verification

Ran \`hyperlane warp check\` against all 4 HYPER routes via the infra HTTP registry. After the SDK fixes:
- Owner addresses match exactly on every leg (production + staging).
- \`type\` correctly detected as \`hyperToken\` on Ethereum HYPER/HYPERSTAGE.
- Only remaining diff is \`contractVerificationStatus: error\` (etherscan API quota in local env) and \`ownerStatus: inactive\` for counterfactual ICAs not yet deployed (the owner address is still correct — it's a CREATE2 ICA waiting to be initialized).

### Backward compatibility

Yes. The HypERC20 \`virtual\` modifiers are additive; no storage layout or bytecode behavior change for existing synthetic deployments.

### Testing

Manual: \`hyperlane warp check\` against live HYPER/stHYPER/HYPERSTAGE/stHYPERSTAGE produces clean output after this branch.


larryob (approved):
LGTM

paulbalaji (commented):
Left one deploy-flow issue inline. Not requesting changes.

solidity/contracts/token/extensions/HyperToken.sol
```diff
@@ -0,0 +1,104 @@
...
+        string memory _name,
+        string memory _symbol,
+        address _hook,
+        address _interchainSecurityModule,
+        address _owner
+    ) public override initializer {
+        __ERC20_init(_name, _symbol);
+        __ERC20Burnable_init();
+        __ERC20Permit_init(_name);
+        _mint(_owner, _initialSupply);
```
paulbalaji, line 61: This uses the initializer `_owner` as both the initial supply recipient and the `DEFAULT_ADMIN_ROLE` holder, but `HypERC20Deployer.initializeArgs()` currently passes the deployer signer here and relies on the later `transferOwnership()` step to move only `Ownable` ownership to the configured route owner. That leaves the configured HYPER owner without AccessControl admin and mints the initial HYPER supply to the deployer signer. I think the deploy flow needs to either pass/grant the final owner for `hyperToken` specifically, or transfer the AccessControl admin/initial supply intentionally before ownership handoff.

paulbalaji (approved):
Approving with the context that HYPER is already live and this PR is wiring existing route/config support.

PR #6035, opened 2025-04-25
feat: xERC20 extra lockboxes derivation updated logic

xeno097:
### Description

Updates the extra lockboxes logic to only rely on block explorers apis and skip derviation otherwise

### Drive-by changes

- Removes rpc based logic to service lockboxes
- Removes e2e tests for reading on chain lockboxes 

### Testing

- Manual

### DIFF
--- a/.changeset/lovely-melons-ring.md
+++ b/.changeset/lovely-melons-ring.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Remove fallback logic to derive extra lockboxes from rpc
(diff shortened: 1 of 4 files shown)


paulbalaji:
What's the context for this? Doesn't this break for any chain that doesn't have a working explorer?

nambrot (approved):

nambrot:
@paulbalaji context is here https://hyperlaneworkspace.slack.com/archives/C08GR6KJ067/p1745545470540629

Basically, deriving without the block explorer isn't really practical so we should just skip doing so

PR #6039, opened 2025-04-26
feat: cosmosnative fixes

paulbalaji:
### Description

feat: cosmosnative fixes
- correctly skip non-EVM chains when doing a warp apply
- adjust the `ZHash` zod regex to support cosmos addresses more explicitly
	- required because setting valid cosmos addresses such as `milk1326ley07fm6rpeqgxmxevnqevrsjfew2akzupg` or `neutron1fqf5mprg3f5hytvzp3t7spmsum6rjrw80mq8zgkc0h6rxga0dtzqws3uu7` as an owner in the warp deploy config would break the schema parsing

### Drive-by changes

- bring `CosmosNative` parity in more conditions where `Cosmos` protocol type exists
- don't fail balance thresholds test if an alert threshold is configured in grafana but undefined locally
	- this is common when doing new chain deploys, and we should not be breaking CI for everyone else just because 1 PR adds a new chain

### Testing

manual cli warp apply of the MILK/bsc-milkyway route

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -652,6 +652,11 @@ async function updateExistingWarpRoute(
 
   await promiseObjAll(
     objMap(expandedWarpDeployConfig, async (chain, config) => {
+      if (multiProvider.getProtocol(chain) !== ProtocolType.Ethereum) {
+        logBlue(`Skipping non-EVM chain ${chain}`);
+        return;
+      }
+
       await retryAsync(async () => {
         const deployedTokenRoute = deployedRoutersAddresses[chain];
         assert(deployedTokenRoute, `Missing artifacts for ${chain}.`);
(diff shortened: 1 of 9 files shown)


typescript/cli/src/deploy/warp.ts
```diff
@@ -652,6 +652,11 @@ async function updateExistingWarpRoute(
 
   await promiseObjAll(
     objMap(expandedWarpDeployConfig, async (chain, config) => {
+      if (multiProvider.getProtocol(chain) !== ProtocolType.Ethereum) {
```
nambrot, line 655: Optional Nit: Probably personal preference, but I'd filter them out from `expandedWarpDeployConfig`

nambrot (approved):

typescript/cli/src/deploy/warp.ts
paulbalaji: thought i'd keep the condition in the loop so we can always adjust if we support more protocol types here

PR #6050, opened 2025-04-28
feat: cosmos core cli

troykessler:
### Description

This PR is still in progress. Once finished the cli commands for core are also compatible with the new protocol type cosmos native. The following commands can then be used for cosmos native:

- core read
- core deploy
- core check
- core apply
- core init

Since this PR is already big and the e2e tests will require more changes a follow up PR will implement those

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual testing with local cosmos chain. e2e tests will follow in follow up PR


troykessler:
This PR will be split up to make it easier to review. It will be split up in 3 steps:
1. 1 pr per sdk module change
2. Pr to change the signers init strategy in the cli
3. 1 PR per command in the cli to add cosmos support

PR #6051, opened 2025-04-28
feat: Support commitments in ICA messages

larryob:
### Description

You can now make shielded calls using ICAs. We use a [commit/reveal](https://en.wikipedia.org/wiki/Commitment_scheme) scheme where the first message is a commitment, and the second message is a reveal. The commitment message looks like 

```solidity
struct {
  address owner;
  address ism;
  bytes32 salt;
  bytes32 commitment;
}
```

The commitment is `hash(bytes32 salt, Call[] calls)`. 

The reveal message looks like 
```solidity
struct {
  address offChainLookupIsm;
  bytes32 commitment;
}
```

The offChainLookupIsm will contain the urls that a relayer must query to retrieve the calls/salt given the commitment. The relayer will pass the retrieved calls/salt to this ism as **metadata**, and the ism will verify if the salt/calls match the commitment. For more, see #6069 and the [CCIP spec](https://eips.ethereum.org/EIPS/eip-3668#specification).
### Related issues
https://linear.app/hyperlane-xyz/issue/ENG-1425/ica-call-commitments-smart-contract-changes

### Backward compatibility

No.

### Testing
Unit.

### DIFF
--- a/solidity/contracts/middleware/InterchainAccountRouter.sol
+++ b/solidity/contracts/middleware/InterchainAccountRouter.sol
@@ -321,7 +321,7 @@ contract InterchainAccountRouter is Router {
             CallLib.Call[] memory calls = _message.calls();
             ica.multicall{value: msg.value}(calls);
         } else {
-            // This is definitely of a message of type COMMITMENT
+            // This is definitely a message of type COMMITMENT
             verifiedCommitments[_commitment] = ica;
         }
     }
--- a/solidity/contracts/middleware/InterchainAccountRouter.sol
+++ b/solidity/contracts/middleware/InterchainAccountRouter.sol
@@ -754,6 +754,47 @@ contract InterchainAccountRouter is Router {
             );
     }
 
+    function sendCommitment(
+        uint32 _destination,
+        bytes32 _router,
+        bytes32 _ism,
+        bytes memory _hookMetadata,
+        IPostDispatchHook _hook,
+        bytes32 _salt,
+        bytes32 _commitment
+    ) public payable returns (bytes32) {
+        bytes memory _body = InterchainAccountMessage.encodeCommitment({
+            _owner: msg.sender.addressToBytes32(),
+            _ism: _ism,
+            _commitment: _commitment,
+            _userSalt: _salt
+        });
+        return
+            _dispatchMessageWithHook(
+                _destination,
+                _router,
+                _ism,
+                _body,
+                _hookMetadata,
+                _hook
+            );
+    }
+
+    error InvalidCommitment(OwnableMulticall ICA, bytes32 badCommitment);
+
+    /// @dev The calls represented by the commitment can only be executed once.
+    function executeWithCommitment(
+        OwnableMulticall _ICA,
+        CallLib.Call[] calldata _calls
+    ) public payable {
+        bytes32 _commitment = keccak256(abi.encode(_calls));
+        if (verifiedCommitments[_ICA] != _commitment) {
+            revert InvalidCommitment(_ICA, _commitment);
+        }
+        _ICA.multicall{value: msg.value}(_calls);
+        delete verifiedCommitments[_ICA];
+    }
+
     // ============ Internal Functions ============
 
     /**
--- a/solidity/contracts/middleware/InterchainAccountRouter.sol
+++ b/solidity/contracts/middleware/InterchainAccountRouter.sol
@@ -15,7 +15,7 @@ pragma solidity ^0.8.13;
 
 // ============ Internal Imports ============
 import {OwnableMulticall} from "./libs/OwnableMulticall.sol";
-import {InterchainAccountMessage} from "./libs/InterchainAccountMessage.sol";
+import {InterchainAccountMessage, InterchainAccountMessageReveal} from "./libs/InterchainAccountMessage.sol";
 import {CallLib} from "./libs/Call.sol";
 import {MinimalProxy} from "../libs/MinimalProxy.sol";
 import {TypeCasts} from "../libs/TypeCasts.sol";
@@ -286,21 +286,27 @@ contract InterchainAccountRouter is Router {
         InterchainAccountMessage.MessageType _messageType = _message
             .messageType();
 
-        // If the message is a reveal, we don't do any derivation of the ica address
-        // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
-        // so here we just check that commitment -> ICA association has been cleared
+        bytes32 _commitment;
+        bytes32 _ism;
         if (_messageType == InterchainAccountMessage.MessageType.REVEAL) {
+            _commitment = InterchainAccountMessageReveal.commitment(_message);
+            _ism = InterchainAccountMessageReveal.ism(_message);
+
+            // If the message is a reveal, we don't do any derivation of the ica address
             // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
+            // so here we just check that commitment -> ICA association has been cleared
             require(
-                verifiedCommitments[_message.commitment(_messageType)] ==
+                verifiedCommitments[_commitment] ==
                     OwnableMulticall(payable(address(0))),
                 "Commitment was not executed"
             );
             return;
+        } else {
+            _commitment = _message.commitment();
+            _ism = _message.ism();
         }
 
         bytes32 _owner = _message.owner();
-        bytes32 _ism = _message.ism(_messageType);
         bytes32 _salt = _message.salt();
 
         OwnableMulticall ica = getDeployedInterchainAccount(
@@ -315,8 +321,8 @@ contract InterchainAccountRouter is Router {
             CallLib.Call[] memory calls = _message.calls();
             ica.multicall{value: msg.value}(calls);
         } else {
-            bytes32 commitment = _message.commitment(_messageType);
-            verifiedCommitments[commitment] = ica;
+            // This is definitely of a message of type COMMITMENT
+            verifiedCommitments[_commitment] = ica;
         }
     }
 
--- a/solidity/contracts/middleware/InterchainAccountRouter.sol
+++ b/solidity/contracts/middleware/InterchainAccountRouter.sol
@@ -36,6 +36,7 @@ contract InterchainAccountRouter is Router {
 
     using TypeCasts for address;
     using TypeCasts for bytes32;
+    using InterchainAccountMessage for bytes;
 
     // ============ Constants ============
 
@@ -44,7 +45,7 @@ contract InterchainAccountRouter is Router {
 
     // ============ Public Storage ============
     mapping(uint32 => bytes32) public isms;
-    mapping(OwnableMulticall ICA => bytes32 commitment)
+    mapping(bytes32 commitment => OwnableMulticall ICA)
         public verifiedCommitments;
 
     // ============ Upgrade Gap ============
@@ -280,13 +281,27 @@ contract InterchainAccountRouter is Router {
         bytes32 _sender,
         bytes calldata _message
     ) external payable override onlyMailbox {
-        InterchainAccountMessage.MessageType _messageType = InterchainAccountMessage
-                .messageType(_message);
-        bytes32 _owner = InterchainAccountMessage.owner(_message);
-        bytes32 _ism = InterchainAccountMessage.ism(_message);
-        bytes32 _salt = InterchainAccountMessage.salt(_message);
+        InterchainAccountMessage.MessageType _messageType = _message
+            .messageType();
 
-        OwnableMulticall _interchainAccount = getDeployedInterchainAccount(
+        // If the message is a reveal, we don't do any derivation of the ica address
+        // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
+        // so here we just check that commitment -> ICA association has been cleared
+        if (_messageType == InterchainAccountMessage.MessageType.REVEAL) {
+            // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
+            require(
+                verifiedCommitments[_message.commitment()] !=
+                    OwnableMulticall(payable(address(0))),
+                "Commitment ewas not executed"
+            );
+            return;
+        }
+
+        bytes32 _owner = _message.owner();
+        bytes32 _ism = _message.ism();
+        bytes32 _salt = _message.salt();
+
+        OwnableMulticall ica = getDeployedInterchainAccount(
             _origin,
             _owner,
             _sender,
@@ -295,13 +310,11 @@ contract InterchainAccountRouter is Router {
         );
 
         if (_messageType == InterchainAccountMessage.MessageType.CALLS) {
-            CallLib.Call[] memory _calls = InterchainAccountMessage.calls(
-                _message
-            );
-            _interchainAccount.multicall{value: msg.value}(_calls);
+            CallLib.Call[] memory calls = _message.calls();
+            ica.multicall{value: msg.value}(calls);
         } else {
-            bytes32 commitment = InterchainAccountMessage.commitment(_message);
-            verifiedCommitments[_interchainAccount] = commitment;
+            bytes32 commitment = _message.commitment();
+            verifiedCommitments[commitment] = ica;
         }
     }
 
@@ -730,7 +743,7 @@ contract InterchainAccountRouter is Router {
             );
     }
 
-    function sendCommitment(
+    function callRemoteCommitReveal(
         uint32 _destination,
         bytes32 _router,
         bytes32 _ism,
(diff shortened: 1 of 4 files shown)


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -300,12 +301,11 @@ contract InterchainAccountRouter is Router {
...
-            bytes32 _owner,
-            bytes32 _ism,
-            CallLib.Call[] memory _calls,
-            bytes32 _salt
-        ) = InterchainAccountMessage.decode(_message);
+        InterchainAccountMessage.MessageType _messageType = InterchainAccountMessage
+                .messageType(_message);
+        bytes32 _owner = InterchainAccountMessage.owner(_message);
+        bytes32 _ism = InterchainAccountMessage.ism(_message);
+        bytes32 _salt = InterchainAccountMessage.salt(_message);
```
yorhodes, line 308: should we adopt the `using InterchainAccountMessage for bytes` syntax?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -314,7 +314,16 @@ contract InterchainAccountRouter is Router {
...
-        _interchainAccount.multicall{value: msg.value}(_calls);
+
+        if (_messageType == InterchainAccountMessage.MessageType.CALLS) {
+            CallLib.Call[] memory _calls = InterchainAccountMessage.calls(
+                _message
+            );
+            _interchainAccount.multicall{value: msg.value}(_calls);
+        } else {
+            bytes32 commitment = InterchainAccountMessage.commitment(_message);
+            verifiedCommitments[_interchainAccount] = commitment;
```
yorhodes, line 325: does this mean an account cannot send multiple concurrent commitments?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -745,6 +754,47 @@ contract InterchainAccountRouter is Router {
...
 
+    function sendCommitment(
+        uint32 _destination,
+        bytes32 _router,
+        bytes32 _ism,
+        bytes memory _hookMetadata,
+        IPostDispatchHook _hook,
+        bytes32 _salt,
+        bytes32 _commitment
+    ) public payable returns (bytes32) {
```
yorhodes, line 765: wdyt of `callRemoteShielded` or something?

yorhodes: I also think we should have a version here which uses the default router, ism, hook, and hookMetadata

yorhodes: like the simplest usage should be
```solidity
callRemoteShielded(uint32 _destination, bytes32 _commitment)
```

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -745,6 +754,47 @@ contract InterchainAccountRouter is Router {
...
+                _destination,
+                _router,
+                _ism,
+                _body,
+                _hookMetadata,
+                _hook
+            );
+    }
+
+    error InvalidCommitment(OwnableMulticall ICA, bytes32 badCommitment);
```
yorhodes, line 783: prefer revert strings to custom errors
we can consider doing this migration wholesale in future but have found that custom error decoding is not universally supported across client side tooling

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -745,6 +754,47 @@ contract InterchainAccountRouter is Router {
...
+
+    /// @dev The calls represented by the commitment can only be executed once.
+    function executeWithCommitment(
+        OwnableMulticall _ICA,
+        CallLib.Call[] calldata _calls
+    ) public payable {
+        bytes32 _commitment = keccak256(abi.encode(_calls));
+        if (verifiedCommitments[_ICA] != _commitment) {
```
yorhodes, line 791: alternatively could we just check if `verifiedCommitMents[_ICA][_commitment]` is set? to allow for multiple calls

solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes: should we revert if its already populated?

yorhodes: lets add a test with multiple sequential invocations

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -34,13 +41,15 @@ library InterchainAccountMessage {
...
-            );
+
+        bytes memory prefix = abi.encodePacked(
+            MessageType.CALLS,
+            _owner.addressToBytes32(),
+            _ism,
+            EMPTY_SALT
+        );
+        bytes memory suffix = abi.encode(_calls);
+        return bytes.concat(prefix, suffix);
```
yorhodes, line 52: why use `bytes.concat` instead of `abi.encodePacked`? im actually not sure what the difference is

yorhodes, line 52: idk if the compiler would already optimize this away but I doubt it
```suggestion
        return abi.encodePacked(
            MessageType.CALLS,
            _owner.addressToBytes32(),
            _ism,
            EMPTY_SALT,
            abi.encode(_calls)
        );
```

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -107,27 +116,81 @@ library InterchainAccountMessage {
...
     ) internal pure returns (bytes memory) {
-        return abi.encode(_owner, _ism, _calls, _userSalt);
+        bytes memory prefix = abi.encodePacked(
+            MessageType.CALLS,
+            _owner,
+            _ism,
+            _userSalt
+        );
+        bytes memory suffix = abi.encode(_calls);
+        return bytes.concat(prefix, suffix);
```
yorhodes, line 126: ditto

yorhodes (commented):
looks like a good start
we will need to figure out sending the second message for the reveal step and how to quote fees for this new endpoint

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -745,6 +754,47 @@ contract InterchainAccountRouter is Router {
...
+            revert InvalidCommitment(_ICA, _commitment);
+        }
+        _ICA.multicall{value: msg.value}(_calls);
+        delete verifiedCommitments[_ICA];
```
trocher, line 795:
```suggestion
        delete verifiedCommitments[_ICA];
        _ICA.multicall{value: msg.value}(_calls);
```
Check-Effect-Interaction pattern would be good here!

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -107,27 +116,81 @@ library InterchainAccountMessage {
...
 
     /**
      * @notice Parses and returns the ISM address from the provided message
      * @param _message The interchain account message
      * @return The ISM encoded in the message
      */
-    function ism(bytes calldata _message) internal pure returns (address) {
-        return address(bytes20(_message[44:64]));
+    function ism(bytes calldata _message) internal pure returns (bytes32) {
+        return bytes32(_message[33:65]);
```
yorhodes, line 178: hmm why change this to bytes32?

solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes: discussed offline and we will invert this mapping to be `[commitment] => (ICA, status)`
lets make sure it only takes a single slot per value

solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes: thanks @trocher !

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -12,12 +12,19 @@ import {TypeCasts} from "../../libs/TypeCasts.sol";
  */
 library InterchainAccountMessage {
     using TypeCasts for bytes32;
+    using TypeCasts for address;
+
+    enum MessageType {
+        CALLS,
+        COMMITMENT
+    }
```
yorhodes, line 20: discussed offline
we will have 3 message types
calls, commit, and reveal

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -278,21 +281,41 @@ contract InterchainAccountRouter is Router {
...
             _salt
         );
-        _interchainAccount.multicall{value: msg.value}(_calls);
+
+        if (_messageType == InterchainAccountMessage.MessageType.CALLS) {
+            CallLib.Call[] memory calls = _message.calls();
+            ica.multicall{value: msg.value}(calls);
+        } else {
+            bytes32 commitment = _message.commitment();
+            verifiedCommitments[commitment] = ica;
```
yorhodes, line 317: what should we do if this mapping is already populated?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -720,6 +743,59 @@ contract InterchainAccountRouter is Router {
...
+    ) public payable returns (bytes32 _commitmentMsgId, bytes32 _revealMsgId) {
+        bytes memory _commitmentMsg = InterchainAccountMessage
+            .encodeCommitment({
+                _owner: msg.sender.addressToBytes32(),
+                _ism: _ism,
+                _commitment: _commitment,
+                _userSalt: _salt
+            });
+        bytes memory _revealMsg = InterchainAccountMessage.encodeReveal({
+            _ism: _ism,
```
yorhodes, line 763: is this the CCIP Read ISM? as written you force the user to use the same for security and liveness

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -278,21 +281,41 @@ contract InterchainAccountRouter is Router {
...
+            .messageType();
+
+        // If the message is a reveal, we don't do any derivation of the ica address
+        // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
+        // so here we just check that commitment -> ICA association has been cleared
+        if (_messageType == InterchainAccountMessage.MessageType.REVEAL) {
+            // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
+            require(
+                verifiedCommitments[_message.commitment()] !=
+                    OwnableMulticall(payable(address(0))),
```
yorhodes, line 294: don't you want to assert that it is equal to 0 (after the `delete` in `revealAndExecute`?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -44,6 +45,8 @@ contract InterchainAccountRouter is Router {
 
     // ============ Public Storage ============
     mapping(uint32 => bytes32) public isms;
+    mapping(bytes32 commitment => OwnableMulticall ICA)
```
yorhodes, line 48: add a comment for how this commitment is defined?

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -34,13 +42,15 @@ library InterchainAccountMessage {
...
-                _ism,
-                _calls,
-                EMPTY_SALT // Salts are expected when decoding.
-            );
+
+        bytes memory prefix = abi.encodePacked(
+            MessageType.CALLS,
+            _owner.addressToBytes32(),
+            _ism,
+            EMPTY_SALT
```
yorhodes, line 50: why is there a salt on the calls message type?

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -107,27 +117,84 @@ library InterchainAccountMessage {
...
      * @param _message The interchain account message
      * @return The ISM encoded in the message
      */
-    function ism(bytes calldata _message) internal pure returns (address) {
-        return address(bytes20(_message[44:64]));
+    function ism(bytes calldata _message) internal pure returns (bytes32) {
+        if (messageType(_message) == MessageType.REVEAL) {
+            return bytes32(_message[1:33]);
+        }
+        return bytes32(_message[33:65]);
```
yorhodes, line 179: I dont love branching in here
could we have separate codecs for each message type?

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -107,27 +117,84 @@ library InterchainAccountMessage {
...
+        return abi.decode(_message[97:], (CallLib.Call[]));
+    }
+
+    function commitment(
+        bytes calldata _message
+    ) internal pure returns (bytes32) {
+        if (messageType(_message) == MessageType.REVEAL) {
+            return bytes32(_message[33:65]);
+        }
+        return bytes32(_message[97:]);
```
yorhodes, line 198: when type == `CALLS`, this is invalid yes?

solidity/test/InterchainAccountRouter.t.sol
```diff
@@ -867,4 +867,65 @@ contract InterchainAccountRouterTest is InterchainAccountRouterTestBase {
...
+
+        // Destination ICA router should have the commitment after commit message
+        assertEq(
+            address(destinationIcaRouter.verifiedCommitments(commitment)),
+            address(ica)
+        );
+
+        // Process reveal message
+        environment.processNextPendingMessage();
+        destinationIcaRouter.revealAndExecute(calls, salt);
```
yorhodes, line 925: the `processNextPendingMessage` should do the execution and then the `revealAndExecute` should revert

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Done.

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: See discussion here: https://github.com/ethereum/solidity/issues/10903

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Added

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Because `callRemoteCommitReveal` takes in a `bytes32` ism, and so `encodeReveal` does as well.

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: We're using the salt to derive the ica, so we still need to include it, even it's zero.

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Yeah, it would just be the first 32 bytes of the encoded `Call[] calls`.

solidity/test/InterchainAccountRouter.t.sol
larryob: Updated

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Updated a couple functions to take message type as a param. This doesn't fix branching, but I think using a parameter is cleaner than using a different function signature for every different kind of message. If we have more message types in the future, using different functions could become unwieldly.

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
yorhodes: I would maybe assert against that then
what do you think of having separate libraries for each message type as opposed to branching within each function?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -278,21 +283,41 @@ contract InterchainAccountRouter is Router {
...
             _ism.bytes32ToAddress(),
             _salt
         );
-        _interchainAccount.multicall{value: msg.value}(_calls);
+
+        if (_messageType == InterchainAccountMessage.MessageType.CALLS) {
+            CallLib.Call[] memory calls = _message.calls();
+            ica.multicall{value: msg.value}(calls);
+        } else {
+            bytes32 commitment = _message.commitment(_messageType);
```
yorhodes, line 318: messageType is guaranteed to be `COMMIT` at this point so I would be explicit

solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes: what if they send the same commitment N times?
if I was a user of this API, I would expect the calls to be executed N times but as written they may only be executed once

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -107,27 +117,88 @@ library InterchainAccountMessage {
...
      * @return The ISM encoded in the message
      */
-    function ism(bytes calldata _message) internal pure returns (address) {
-        return address(bytes20(_message[44:64]));
+    function ism(
+        bytes calldata _message,
+        MessageType _type
+    ) internal pure returns (bytes32) {
+        if (_type == MessageType.REVEAL) {
+            return bytes32(_message[1:33]);
```
yorhodes, line 180: if we moved the ISM to the first field, could we remove this branching?

yorhodes (commented):
almost there

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Looks like we'll just end up moving to structs (https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6166). I've already added a library for reveal messages, but I'll skip adding another lib for commitment messages.

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Done

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Branching removed

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -193,12 +187,24 @@ library InterchainAccountMessage {
...
         return bytes32(_message[97:]);
     }
 }
+
+library InterchainAccountMessageReveal {
+    enum MessageType {
+        CALLS,
+        COMMITMENT,
+        REVEAL
+    }
```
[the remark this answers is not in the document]
larryob: Fixed

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Fixing in another PR, as discussed offline

PR #6054, opened 2025-04-28
feat: Integrate movable collateral contracts into warp init

fzavalia:


fzavalia:
Closing because it is not needed

PR #6058, opened 2025-04-28
chore: reformat cosmos-types code

antigremlin:
### Description

This PR formats the protobuf-generated code in `cosmos-types` to comply with the updated Prettier rules. We also remove the formatter from the `yarn build` script because it doesn't have to run every time.

### Related issues

Fixes ENG-1369.

### Backward compatibility

Yes

### Testing

Not needed for a formatting change.


paulbalaji:
i think yarn-test will pass again after pulling latest main

paulbalaji (approved):

PR #6065, opened 2025-04-28
chore: update formatting after Prettier upgrade

antigremlin:
### Description

This PR updates source code formatting after the recent Prettier upgrade. We also add more root-level Prettier ignores as we currently run it recursively from root. 

### Related issues

Fixes ENG-1388.

### Backward compatibility

Yes

### Testing

CI should catch any accidental breakage.


.github/actions/install-cli/action.yml
```diff
@@ -1,17 +1,17 @@
-name: "Install Hyperlane CLI"
-description: "Install the Hyperlane CLI by packaging from a local build"
+name: 'Install Hyperlane CLI'
```
paulbalaji, line 1: is it the prettier default to be ' instead of "" now?

antigremlin: Good question. The default is to use double quotes, but our configuration sets it to single quotes (which, incidentally, matches my personal preference).

I think the reason it’s changing in this PR is that we never applied formatting to some configuration files, including this directory.

yorhodes (approved):
can we merge this asap? I am getting prettier errors on my branches now

paulbalaji (approved):
lgtm once conflict with main is resolved

PR #6067, opened 2025-04-28
ci: only run infra tests when infra changes

yorhodes:


typescript/infra/package.json
```diff
@@ -77,7 +77,8 @@
     "kathy": "yarn tsx ./scripts/send-test-messages.ts",
     "prettier": "prettier --write ./src ./config ./scripts ./test",
     "test": "yarn test:unit && yarn test:hardhat",
-    "test:unit": "mocha --config ../sdk/.mocharc.json test/**/*.test.ts",
+    "test:unit": "mocha --config ../sdk/.mocharc.json \"test/**/*.test.ts\" --ignore \"test/{warp-configs,balance-alerts}.test.ts\"",
+    "test:infra": "mocha --config ../sdk/.mocharc.json \"test/{warp-configs,balance-alerts}.test.ts\"",
```
paulbalaji, line 81: this is nice

.github/workflows/test.yml
```diff
@@ -131,6 +131,30 @@ jobs:
...
+        uses: ./.github/actions/yarn-build-with-cache
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: Checkout registry
+        uses: ./.github/actions/checkout-registry
+
+      - name: Infra Tests
+        run: |
+          if git diff --name-only | grep -qE '^typescript/infra/|^.registryrc$'; then
```
paulbalaji, line 154: `git diff` doesn't work as nice as one hopes in PRs, but you can instead use the `gh` cli directly for better PR diffing like we do here
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/40603302dbe5b402c988e58121291dd078ddb276/.github/workflows/test.yml#L304-L339

yorhodes: ofc
chatgpt is dumb

Mo-Hussain: As we are doing this, we should only run the `balance-alerts` if we make changes to files in this directory `infra/config/environments/mainnet3/balances`

.github/workflows/test.yml
```diff
@@ -149,10 +150,16 @@ jobs:
...
+            yarn --cwd typescript/infra test:balance
+          fi
+
+      - name: Warp Config Tests
         run: |
-          if git diff --name-only | grep -qE '^typescript/infra/|^.registryrc$'; then
-            yarn --cwd typescript/infra test:infra
+          if git diff $PR_BASE HEAD --name-only | grep -qE '^typescript/infra/|^.registryrc$'; then
+            yarn --cwd typescript/infra test:balance
           fi
```
paulbalaji, line 163: if the PR branch has no changes to these files but you're out of date with HEAD which does have the changes, i think this still triggers a diff

paulbalaji (approved):

PR #6069, opened 2025-04-29
feat: Add CCIP Read ISM for Commitments

larryob:
### Description

Add CCIP Read ISM for commitments.

### Drive-by changes
Updated MockMailbox such that you can now pass in metadata that will be used during message verification. See `addInboundMetadata`.

### Related issues

Fixes https://linear.app/hyperlane-xyz/issue/PROT-23/add-ccip-read-ism

### Backward compatibility

Yes.

### Testing

Unit.

### Misc
Builds on #6051


solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -318,6 +323,37 @@ contract InterchainAccountRouter is Router {
         }
     }
 
+    function moduleType() external pure returns (uint8) {
+        return uint8(IInterchainSecurityModule.Types.ROUTING);
+    }
+
+    CommitmentReadIsm public immutable CCIP_READ_ISM;
```
yorhodes, line 330: declare this next to the other constants

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -318,6 +323,37 @@ contract InterchainAccountRouter is Router {
         }
     }
 
+    function moduleType() external pure returns (uint8) {
+        return uint8(IInterchainSecurityModule.Types.ROUTING);
+    }
```
yorhodes, line 328: inherit `AbstractRoutingIsm` and implement `route` which the relayer introspects

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -290,7 +295,7 @@ contract InterchainAccountRouter is Router {
         if (_messageType == InterchainAccountMessage.MessageType.REVEAL) {
             // The commitment should be executed in the `verify` method of the CCIP read ISM that verified this message
             require(
-                verifiedCommitments[_message.commitment()] !=
+                verifiedCommitments[_message.commitment()] ==
```
yorhodes, line 298: would have expected you to fix this in the other PR

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -159,21 +159,18 @@ library InterchainAccountMessage {
 
     function messageType(
         bytes calldata _message
-    ) internal pure returns (MessageType) {
+    ) internal view returns (MessageType) {
+        console.log("messageType");
+        console.logBytes(_message);
+        console.log(uint(uint8(_message[0])));
```
yorhodes, line 165: assume you will remove these

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
```diff
@@ -192,7 +189,7 @@ library InterchainAccountMessage {
 
     function commitment(
         bytes calldata _message
-    ) internal pure returns (bytes32) {
+    ) internal view returns (bytes32) {
```
yorhodes, line 192: why is this view now?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -318,6 +323,37 @@ contract InterchainAccountRouter is Router {
...
+            if (
+                _body.messageType() ==
+                InterchainAccountMessage.MessageType.REVEAL
+            ) {
+                _ism = CCIP_READ_ISM;
+            } else {
+                _ism = mailbox.defaultIsm();
+            }
+        }
+        return _ism.verify(_metadata, _message);
```
yorhodes, line 354: this logic replaces `InterchainAccountISM` yes?
we should remove that contract

yorhodes (commented):
looks good
I think we should have a simpler overload signature
```solidity
callRemoteCommitReveal(
  uint32 destination,
  bytes32 commitment
)
```

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: I can do this, but I actually would like to move to having variables defined next to the functions that use them. We can discuss more offline.

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Done

solidity/contracts/middleware/libs/InterchainAccountMessage.sol
larryob: Done

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Done

solidity/contracts/middleware/InterchainAccountRouter.sol
yorhodes: since these can only be set in the constructor I would expect them to be declared before the constructor

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -807,14 +856,34 @@ contract InterchainAccountRouter is Router {
...
+        bytes32 _router = routers(_destination);
+        bytes32 _ism = isms[_destination];
+
+        return
+            callRemoteCommitReveal(
+                _destination,
+                _router,
+                _ism,
+                bytes32(0),
+                bytes(""),
```
yorhodes, line 872: is empty hook metadata really what we want?

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -1045,4 +1114,7 @@ contract InterchainAccountRouter is Router {
                 address(hook)
             );
     }
+
+    /// @dev This allows us to receive gas refunds from the IGP
+    receive() external payable {}
```
yorhodes, line 1119: I guess without this, refunds would revert which means that overpayment would revert
that seems preferable

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -967,7 +1036,7 @@ contract InterchainAccountRouter is Router {
         );
         emit RemoteCallDispatched(_destination, msg.sender, _router, _ism);
         return
-            mailbox.dispatch{value: msg.value}(
+            mailbox.dispatch{value: address(this).balance}(
```
yorhodes, line 1039: upon further consideration I think we should keep `msg.value` and omit the `receive`

yorhodes: or alternatively specify the refund address as the `msg.sender` in hook metadata

yorhodes (approved):

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Done

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Updated to pass in a gas limit in the hook metadata

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: We now split the msg.value between the two dispatches.

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,78 @@
...
+    using Message for bytes;
+
+    string[] public urls;
+    Mailbox public immutable mailbox;
+
+    constructor(Mailbox _mailbox) {
+        mailbox = _mailbox;
+    }
+
+    function setUrls(string[] memory _urls) external onlyOwner {
```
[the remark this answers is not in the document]
larryob: @nambrot Is there an issue with the owner setting the urls?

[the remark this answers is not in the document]
nambrot: No i think that is ok

solidity/contracts/hooks/libs/StandardHookMetadata.sol
```diff
@@ -78,6 +79,15 @@ library StandardHookMetadata {
...
     }
 
+    function gasLimit(
+        bytes memory _metadata
+    ) internal pure returns (uint256 _gasLimit) {
+        if (_metadata.length < GAS_LIMIT_OFFSET + 32) return 50_000;
+        assembly {
+            _gasLimit := mload(add(_metadata, add(0x20, GAS_LIMIT_OFFSET)))
+        }
+    }
```
yorhodes, line 89: I dont like the implicit default behavior, and generally we avoid assembly for readability

yorhodes (changes requested):

solidity/contracts/hooks/libs/StandardHookMetadata.sol
larryob: 50_000 is taken from `DEFAULT_GAS_USAGE` in the IGP, which is being used as a default gas limit already: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/46eeda6c3a7110520f4b671a16ed775922d49983/solidity/contracts/hooks/igp/InterchainGasPaymaster.sol#L53

I used assembly because slicing isn't available for memory arrays. We're using `abi.encodePacked` to create the metadata so I can't `abi.decode` to extract the gas limit. I could update `formatMetadata` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/46eeda6c3a7110520f4b671a16ed775922d49983/solidity/contracts/hooks/libs/StandardHookMetadata.sol#L120 to use a standard ABI encoding, but would need to rework the rest of the functions in that lib to support that.

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,79 @@
...
+        bytes calldata body = _message.body();
+
+        if (actualHash != body.commitment()) {
+            return false;
+        }
+
+        InterchainAccountRouter(payable(msg.sender)).revealAndExecute(
+            calls,
+            salt
+        );
```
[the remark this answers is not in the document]
larryob: Fixed

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -31,17 +37,21 @@ import {Address} from "@openzeppelin/contracts/utils/Address.sol";
...
     using InterchainAccountMessage for bytes;
+    using Message for bytes;
+    using StandardHookMetadata for bytes;
 
     // ============ Constants ============
 
     address public immutable implementation;
     bytes32 public immutable bytecodeHash;
+    CommitmentReadIsm public immutable CCIP_READ_ISM;
+    uint public constant COMMIT_TX_GAS_USAGE = 10_000;
```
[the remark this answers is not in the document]
larryob: This was a guesstimate. I can update this number by running a simulation.

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -326,6 +337,37 @@ contract InterchainAccountRouter is Router {
...
+        );
+
+        // If the ISM is not set, we need to check if the message is a reveal
+        // If it is, we need to set the ISM to the CCIP read ISM
+        // Otherwise, we need to set the ISM to the default ISM
+        if (address(_ism) == address(0)) {
+            if (
+                _body.messageType() ==
+                InterchainAccountMessage.MessageType.REVEAL
+            ) {
```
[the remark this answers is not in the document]
larryob: Updated

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -1051,4 +1165,42 @@ contract InterchainAccountRouter is Router {
...
+        return
+            _Router_quoteDispatch(
+                _destination,
+                _messageBody,
+                StandardHookMetadata.overrideGasLimit(
+                    gasLimit + COMMIT_TX_GAS_USAGE
+                ),
+                address(hook)
+            );
+    }
```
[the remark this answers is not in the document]
larryob: Removed

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,79 @@
...
+     */
+    function verify(
+        bytes calldata _metadata,
+        bytes calldata _message
+    ) external returns (bool) {
+        (bytes32 salt, CallLib.Call[] memory calls) = abi.decode(
+            _metadata,
+            (bytes32, CallLib.Call[])
+        );
+        bytes32 actualHash = keccak256(_metadata);
```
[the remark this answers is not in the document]
larryob: We're already hashing the encoded metadata. The decoding is for the `revealAndExecute` call later on. I think keccak only works on memory  (see https://www.evm.codes/?fork=cancun#20) so `mstore` is unavoidable.

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,88 @@
...
+        (bytes32 salt, CallLib.Call[] memory calls) = abi.decode(
+            _metadata,
+            (bytes32, CallLib.Call[])
+        );
+
+        bytes32 revealedHash = keccak256(_metadata);
+        bytes32 msgCommitment = _message.body().commitment();
+        if (revealedHash != msgCommitment) {
+            return false;
+        }
```
yorhodes, line 72: should we revert with a more descriptive error here?

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,88 @@
...
+
+        // If the commitment has been executed, don't call revealAndExecute, just let caller know the message is legit
+        if (
+            address(icaRouter.verifiedCommitments(msgCommitment)) == address(0)
+        ) {
+            return true;
+        }
+
+        icaRouter.revealAndExecute(calls, salt);
+        return true;
```
yorhodes, line 86: nit: small preference for a single return at the end with the revealAndExecute inside the inverse condition

yorhodes (approved):

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
larryob: Done

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
larryob: Done

solidity/contracts/hooks/libs/StandardHookMetadata.sol
yorhodes: got it
I just dont think its obvious that a `hookMetadata.gasLimit()` defaults to this in the library

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        // Revert if this is not the ica's currently active commitment
+        require(
+            ica.commitment() == msgCommitment,
+            "Commitment ISM: Invalid Commitment"
+        );
+        InterchainAccountRouter icaRouter = InterchainAccountRouter(
+            _message.recipient().bytes32ToAddress()
+        );
+        icaRouter.revealAndExecute(calls, salt, ica);
```
yorhodes, line 83: why do you go through the ICA router instead of `ica.reveal()`?

solidity/contracts/middleware/InterchainAccountRouter.sol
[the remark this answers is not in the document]
yorhodes: as mentioned, this should be an immutable so we dont have to change the source code if we need to tweak this

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -813,18 +856,48 @@ contract InterchainAccountRouter is Router {
...
+    ) external payable returns (bytes32 executedCommitment) {
+        // If there is no active commitment, do nothing.
+        bytes32 icaCommitment = ica.commitment();
+        if (icaCommitment == bytes32(0)) return bytes32(0);
+
+        bytes32 revealedHash = keccak256(abi.encode(calls, salt, ica));
+        require(icaCommitment == revealedHash, "Invalid Reveal");
+
+        ica.deleteCommitment();
+        ica.multicall{value: msg.value}(calls);
```
yorhodes, line 899: I think it would be far more efficient if this all happened inside of a single call to the ICA

solidity/contracts/middleware/InterchainAccountRouter.sol
```diff
@@ -813,18 +856,48 @@ contract InterchainAccountRouter is Router {
...
+        return icaCommitment;
```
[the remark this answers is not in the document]
yorhodes: this seems like a valid comment

solidity/contracts/middleware/libs/OwnableMulticall.sol
```diff
@@ -26,6 +27,20 @@ contract OwnableMulticall {
...
     }
 
+    /// @notice The next commitment to execute.
+    bytes32 public commitment;
+
+    /// @notice Sets the commitment value that will be executed next
+    /// @param _commitment The new commitment value to be set
+    function setCommitment(bytes32 _commitment) external onlyOwner {
+        commitment = _commitment;
+    }
```
yorhodes, line 37: imo we should put the `require(commitment == bytes32(0))` in here

solidity/test/isms/CommitmentReadIsm.t.sol
```diff
@@ -0,0 +1,59 @@
...
+
+        // Setting urls doesn't work if you aren't the owner
+        vm.expectRevert("Ownable: caller is not the owner");
+        ism.setUrls(newUrls);
+    }
+
+    /**
+     * We don't need to test `verify` since it's well tested in InterchainAccountRouter.t.sol
+     * when we test the message processing flow for commit/reveal messages.
+     */
```
yorhodes, line 58: I think it would be clearer if the tests were in here but fine I guess

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
larryob: Left over from old iterations, moved into ica.

solidity/contracts/middleware/InterchainAccountRouter.sol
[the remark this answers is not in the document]
larryob: We're already following check-effects-interactions, and the right commitment would always be returned. Updated for clarity though.

solidity/contracts/middleware/InterchainAccountRouter.sol
larryob: Updated

solidity/test/isms/CommitmentReadIsm.t.sol
larryob: Will likely move some tests later, the code has been in flux so it wasn't clear which tests should go where.

solidity/contracts/middleware/libs/OwnableMulticall.sol
```diff
@@ -36,9 +36,25 @@ contract OwnableMulticall {
...
-    function deleteCommitment() external onlyOwner {
-        delete commitment;
+    /// @dev The calls represented by the commitment can only be executed once per commitment,
+    /// though you can submit the same commitment again after the calls have been executed.
+    function revealAndExecute(
+        CallLib.Call[] calldata calls,
+        bytes32 salt
+    ) external payable returns (bytes32 executedCommitment) {
+        // If there is no active commitment, do nothing.
+        if (commitment == bytes32(0)) return bytes32(0);
```
yorhodes, line 46: I expected you to move the
```
require(
            ica.commitment() == msgCommitment,
            "Commitment ISM: Invalid Commitment"
        );
```
block here
we shouldnt need to call into this contract multiple times from the ISM

solidity/contracts/middleware/libs/OwnableMulticall.sol
```diff
@@ -36,9 +36,25 @@ contract OwnableMulticall {
...
+
+        bytes32 revealedHash = keccak256(abi.encode(calls, salt, this));
+        require(commitment == revealedHash, "Invalid Reveal");
```
yorhodes, line 49: why have the "do nothing" block if we also have this revert?

larryob: See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6211

PR #6072, opened 2025-04-29
feat: Send operations to confirm queue if they have non-dropped payloads

ameten:
### Description

Send operations to confirm queue if they have non-dropped payloads

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5894

### Drive-by changes

- Fix of reprepare reason for database operations with payload

### Backward compatibility

Yes

### Testing

None


rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -342,86 +371,247 @@ async fn prepare_task(
...
-                    %submit_queue_len,
-                    max_submit_queue_len=%max_len,
-                    "Submit queue is too long, waiting to prepare more ops"
-                );
-                // The submit queue is too long, so give some time before checking again
-                sleep(Duration::from_millis(150)).await;
-                continue;
-            }
+        if apply_backpressure(&submit_queue, &max_submit_queue_len).await {
+            continue;
```
daniel-savu, line 376: the name `apply_backpressure` indicates sleeping isn't done inside the function. Maybe move out the sleep from there in here, before `continue`ing?

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -342,86 +371,247 @@ async fn prepare_task(
...
-        if batch.is_empty() {
-            // queue is empty so give some time before checking again to prevent burning CPU
-            sleep(Duration::from_millis(100)).await;
+
+        let batch = if let Some(batch) = get_batch_or_wait(&mut prepare_queue, max_batch_size).await
+        {
+            batch
+        } else {
+            continue;
+        };
```
daniel-savu, line 384: would this work?
```suggestion
        let Some(batch) = get_batch_or_wait(&mut prepare_queue, max_batch_size).await else {
            continue;
        };
```

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -342,86 +371,247 @@ async fn prepare_task(
...
+            &metrics,
+        )
+        .await;
+    }
+}
+
+/// This function checks the status of the payloads associated with the operations in the batch.
+/// If the payload is dropped or does not exist, the operation will go through prepare logic.
+/// If the payload is not dropped, the operation is pushed to the confirmation queue.
+/// If there are an error on retrieving payload ids or statuses for an operation, the operation
```
daniel-savu, line 447: In very simple terms, this function should decide if some messages should be sent straight to the confirm queue rather than filter. Given that, I would actually prepare the operation if retrieving the payload status fails, to keep things simple

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -342,86 +371,247 @@ async fn prepare_task(
...
+/// will be returned to prepare queue.
+async fn filter_ops_to_prepare(
```
daniel-savu, line 449: to me, the naming `filter` suggests purely applying a read-only predicate to see what should be ignored. In this case though, `filter_single_op_to_prepare` actually writes to the confirm queue / prep queues. 

There are really two outcomes:
1) we submit the operation to the confirm step
2) we forward the operation to the prepare logic

Based on that, I'd rename this function to `confirm_already_submitted_operations` and return the Vec of ops to prepare. For simplicity, I don't think it should be up to this function to push to the prepare queue, so would remove `prepare_queue` from the params

rust/main/agents/relayer/src/msg/op_submitter.rs
```diff
@@ -342,86 +371,247 @@ async fn prepare_task(
...
+        .await;
+
+        if let Some(op_to_prepare) = maybe_op_to_prepare {
+            ops_to_prepare.push(op_to_prepare);
+        }
+    }
+    ops_to_prepare
+}
+
+async fn filter_single_op_to_prepare(
```
daniel-savu, line 474: this function has quite a lot of cases that we can simplify IMO. Since there are only two outcomes for an operation - 1) either submit to the confirm queue or, 2) keep preparing this operation - in this function I would only categorize what needs to be done by returning a bool, representing whether to confirm this operation. Then I would do the "sending to the confirm step / adding to the Vec of ops to prepare" in the`confirm_already_submitted_operations` function above. This function should probably be named `has_operation_been_confirmed` given that

daniel-savu (commented):
i think we can get away with simplifying the flow a bit, left some suggestions

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: Done

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: Done

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: Done

rust/main/agents/relayer/src/msg/op_submitter.rs
ameten: Done

daniel-savu (approved):

PR #6078, opened 2025-04-29
fix: tracing unit test

kamiyaa:
### Description

 - remove trace subscriber and use #[traced_test] decorator

if you want to see logs, use `-- --nocapture`

### Related issues

- https://linear.app/hyperlane-xyz/issue/BACK-179/cargo-test-fails-unless-run-on-a-single-thread

### Backward compatibility

Yes


daniel-savu (approved):
great find!

ameten (approved):

PR #6095, opened 2025-04-30
Remove outputting isNft in warp init config

nambrot:
### Description

Remove outputting isNft in warp init config (since we don't support NFTs practically anyways)

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/typescript/cli/src/config/warp.ts
+++ b/typescript/cli/src/config/warp.ts
@@ -166,10 +166,6 @@ export async function createWarpRouteDeployConfig({
       choices: typeChoices,
     });
 
-    // TODO: restore NFT prompting
-    const isNft =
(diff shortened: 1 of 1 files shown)


typescript/cli/src/config/warp.ts
```diff
@@ -180,7 +176,6 @@ export async function createWarpRouteDeployConfig({
           type,
           owner,
           proxyAdmin,
-          isNft,
```
ltyu, line 183: to ensure we not break anything, can you just set `isNft: true` for collateralUri and syntheticUri?

nambrot: Done

ltyu (approved):

xeno097 (approved):

PR #6096, opened 2025-04-30
feat: optimize `InterchainAccountRouter.handle` proxy deployment

yorhodes:
### Description

- Remove account owners reverse mapping
- Update` InterchainAccountCreated` event for reverse mapping purpose
- Use CREATE2 for constructor implementation deploy
- Do not proxy ICA routers

### Backward compatibility

No

### Testing

Unit Tests

### DIFF
--- a/typescript/infra/src/govern/HyperlaneAppGovernor.ts
+++ b/typescript/infra/src/govern/HyperlaneAppGovernor.ts
@@ -366,7 +366,16 @@ export abstract class HyperlaneAppGovernor<
       };
     }
 
-    if (account.address !== icas[chain as keyof typeof icas]) {
+    let accountConfig = this.interchainAccount.knownAccounts[account.address];
+    if (account.address === icas[chain as keyof typeof icas]) {
+      const origin = 'ethereum';
+      accountConfig = {
+        origin,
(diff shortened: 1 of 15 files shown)


typescript/infra/src/govern/HyperlaneAppGovernor.ts
```diff
@@ -366,14 +366,25 @@ export abstract class HyperlaneAppGovernor<
...
+        chain,
+        call,
+      };
+    }
+
+    const origin = 'ethereum';
+    const accountConfig = {
+      origin,
+      owner: safes[origin],
+    };
```
paulbalaji, line 386: i suppose this is fine for now since we're not governing from this branch, but we'll want to use the `getGovernanceSafes` nomenclature when it's in main
```
const safeOwner = getGovernanceSafes(governanceType)[chain];
```

yorhodes: really config shouldnt be hardcoded in here but I will lob that to you 😅

yorhodes:
> LGTM. Maybe add a test for the constructor?

the `testFuzz_constructor(address _localOwner)` covers these changes

PR #6097, opened 2025-05-01
fix: Add initializer for MovableCollateralRouter

larryob:
### Description
Since HypERC20Collateral is an upgradeable contract, we can't use constructors to set storage. Here we add an initializer function to `MovableCollateralRouter` and call it in the initializer of `HypERC20Collateral`. 

We've also changed the initializer function on `HypERC20Collateral` to have a modifier of `reinitializer(2)` meaning that it can be called two times. This allows us to call `initialize` again on all existing instances of `HypERC20Collateral`.

### Backward compatibility

Yes

### Testing

Unit

### DIFF
--- a/solidity/contracts/token/HypERC20Collateral.sol
+++ b/solidity/contracts/token/HypERC20Collateral.sol
@@ -100,24 +100,4 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
             bridge: bridge
         });
     }
-
-    function _msgData()
-        internal
-        view
-        virtual
-        override(Context, ContextUpgradeable)
-        returns (bytes calldata)
-    {
-        return ContextUpgradeable._msgData();
-    }
-
-    function _msgSender()
-        internal
-        view
-        virtual
-        override(Context, ContextUpgradeable)
-        returns (address)
-    {
-        return ContextUpgradeable._msgSender();
-    }
 }
--- a/solidity/contracts/token/HypERC20Collateral.sol
+++ b/solidity/contracts/token/HypERC20Collateral.sol
@@ -52,8 +52,9 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
         address _hook,
         address _interchainSecurityModule,
         address _owner
-    ) public virtual initializer {
+    ) public virtual reinitializer(2) {
         _MailboxClient_initialize(_hook, _interchainSecurityModule, _owner);
+        _MovableCollateralRouter_initialize(_owner);
     }
 
     function balanceOf(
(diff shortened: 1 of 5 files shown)


yorhodes (approved):
ive never used the reinitializer
I guess we just need to call `upgradeAndCall` with initialize for existing instances? what happens with net new instances?

solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -52,8 +52,9 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
         address _hook,
         address _interchainSecurityModule,
         address _owner
-    ) public virtual initializer {
+    ) public virtual reinitializer(2) {
```
fzavalia, line 55: This will allow upgrading previous versions of the contract + deploy new versions of the contract because new versions will be deployed as init version 2 directly. However, upgraded contracts will call the _MailboxClient_initialize which I don't think is the expected behavior

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -13,8 +13,8 @@ import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol
 abstract contract MovableCollateralRouter is AccessControl {
```
fzavalia, line 13: This could use the AccessControlUpgradeable OZ contract as well, which uses ContextUpgradeable, which is used by the OwnableUpgradeable which is also in the chain of inheritance and could remove:

```solidity
function _msgData()
    internal
    view
    virtual
    override(Context, ContextUpgradeable)
    returns (bytes calldata)
{
    return ContextUpgradeable._msgData();
}

function _msgSender()
    internal
    view
    virtual
    override(Context, ContextUpgradeable)
    returns (address)
{
    return ContextUpgradeable._msgSender();
}
```

which was needed because of using both Context and ContextUpgradeable

fzavalia (commented):
MovableCollateralRouter has a bunch of storage variables that might break the layout in case of an upgrade. Haven't seen all mutables from FungibleTokenRouter, but at least `wrappedToken` will be affected in case of an upgrade I believe.

I believe the simplest approach to keep upgradeabily and the contract name would be to rename HypERC20Collateral to HypERC20CollateralBase and add all rebalance logic to a new HypERC20Collateral which inherits from both HypERC20CollateralBase and AccessControlUpgradeable

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: Fixed!

solidity/contracts/token/HypERC20Collateral.sol
larryob: I like this approach for being more explicit, but I think having multiple initialization functions sets us up for off-chain issues (e.g. forgetting to call `initialize` and only calling `initializeV2` on new routers).

PR #6106, opened 2025-05-01
fix(submitter): Cache estimate and use it only on the first submission attempt

ameten:
### Description

Simulation was removed to make Lander as quick and Classical Submitter. We re-introduce the simulation before submission in the form of estimation. We shall use the result of the estimation on the first attempt of submission so that this does not add latency.

### Drive-by changes

Removed the increment of the counter of submission attempts from within submission logic since currently the value which is produced is not correct:
- When we retry the transaction submission until success or non-retryable error, we don't retain the updates from unsuccessful submissions since we pass a cloned instance of transaction.
- When the submission is successful, we increment the counter in `process_pending_tx` method.

It means that in all cases the overall result is that the counter is updated twice.

### Backward compatibility

Yes

### Testing

None


rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
```diff
@@ -316,12 +323,39 @@ impl AdaptsChain for SealevelTxAdapter {
         Ok(success)
     }
 
+    async fn estimate_tx(&self, tx: &mut Transaction) -> Result<(), SubmitterError> {
+        info!(?tx, "estimating transaction");
+        let not_estimated = tx.precursor();
+        let estimated = self.estimate(not_estimated).await?;
```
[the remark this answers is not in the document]
ameten: Not relevant and just incorrect.

rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
```diff
@@ -316,12 +323,39 @@ impl AdaptsChain for SealevelTxAdapter {
...
-        let not_estimated = tx.precursor();
-        // TODO: the `estimate` call shouldn't happen here - the `Transaction` argument should already contain the precursor,
-        // set in the `build_transactions` method
-        let estimated = self.estimate(not_estimated.clone()).await?;
+
+        let previous = tx.precursor();
+        let estimated = {
+            let mut guard = self.estimate_type_cache.lock().await;
+            match guard.get(&tx.id) {
+                Some(EstimateType::Estimation) | None => self.estimate(previous).await?,
```
daniel-savu, line 351: iiuc, `EstimateType` represents whether cache should be refreshed or not. If so, I'd be inclined to call it slightly differently:
```rust
#[derive(Default)]
enum EstimationCacheFreshness {
    #[default]
    Stale,
    Fresh
}
```

with the default declared, we can then slightly simplify the `match` by doing `match guard.get(&tx.id).unwrap_or_default()`, since we're really matching over two cases (stale or fresh), but having a third case confused me (Estimation, Simulation, None)

rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
daniel-savu, line 329: i'm assuming `self.estimate()` also tell us if simulation failed, so we should be able to remove `simulate_tx` from `AdaptsChains`?

daniel-savu (approved):
looks good, left some suggestions to make it easier to understand

there are two tests still ignored due to previously removing the simulation, which we should fix in this PR as well:
- test_entrypoint_send_is_dropped_by_dispatcher
- test_entrypoint_payload_fails_simulation

rust/main/submitter/src/payload_dispatcher/stages/inclusion_stage.rs
```diff
@@ -213,6 +213,23 @@ impl InclusionStage {
         // }
         // info!(?tx, "Transaction simulation succeeded");
 
+        // Estimating transaction just before we submit it
```
daniel-savu, line 216: the commented out simulation code above can probably be removed now

rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
ameten: It makes sense. I'll adopt it.

daniel-savu (approved):

PR #6108, opened 2025-05-01
fix(submitter): svm adapter commitment levels

daniel-savu:
### Description

- All chain reads for confirming the status of a tx happened at the `finalized` [commitment](https://docs.anza.xyz/consensus/commitments) level, which explains why the Finality Pool is always empty in prod and why it takes 10-30s to get a tx included in Sealevel E2E
- This PR reads the tx hash at the `processed` status - if that fails, we return an error like before. Then, if the tx's slot can be read at `finalized` status, the tx is final. If the slot can be read at `confirmed` status, the tx is included. Otherwise the tx is pending inclusion, because the hash does exist.
- Removes the `reorgPeriod` from the SVM adapter, since I noticed we actually [set that to zero](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/f692b35be7bf9d7083877bfe259e18e8bf124ab6/rust/main/config/mainnet_config.json#L2870). We rely on using the right commitment levels in code (e.g. `finalized`) as opposed to setting an arbitrary number of blocks

### Related issues

Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/6103

### Backward compatibility

Yes

### Testing

sealevel E2E

### DIFF
--- a/rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
+++ b/rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
@@ -59,7 +59,6 @@ const TX_RESUBMISSION_MIN_DELAY_SECS: u64 = 15;
 pub struct SealevelTxAdapter {
     estimated_block_time: Duration,
     max_batch_size: u32,
-    reorg_period: ReorgPeriod,
     keypair: SealevelKeypair,
     client: Box<dyn SubmitSealevelRpc>,
     provider: Box<dyn SealevelProviderForSubmitter>,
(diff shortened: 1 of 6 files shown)


ameten (approved):

daniel-savu:
feedback from danil: check whether we can query the commitment level of a txhash. We can also simplify by just querying the tx hash at different commit levels

rust/main/submitter/src/chain_tx_adapter/chains/sealevel/adapter.rs
```diff
@@ -215,48 +215,45 @@ impl SealevelTxAdapter {
...
-                Ok(TransactionStatus::Included)
+                info!("transaction pending inclusion");
+                return Ok(TransactionStatus::PendingInclusion);
             }
-            Err(_) => {
-                info!("pending transaction");
-                Ok(TransactionStatus::PendingInclusion)
+            Err(err) => {
+                warn!(?err, "Failed to get transaction status by hash");
+                return Err(SubmitterError::TxSubmissionError(
```
daniel-savu, line 254: the semantics slightly change by returning an error here, but my assumption is that it's less likely to hit this case given we query at `processed` level. However, if the status queries for all historic hashes return errors, the tx is still assumed to be pending: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e77652913da075f111aa5420e6236dbd18ab48bb/rust/main/submitter/src/transaction/types.rs#L92

PR #3159, opened 2024-01-17
feat:remove create-key circular dependency

aroralanuk:
### Description

Diagnosis: circular dependency created by igp requiring multisig size and threshold for building coreConfig which is needed while creating new keys. This arises in the scenario where you add the chain to chainMetadata and not to defaultMultisigConfigs (say newChainConfigs)

- separated agents and helloworld configs from core and segregated scripts for `core-utils.ts` and `agent-utils.ts` for avoiding loading of envConfig.coreConfig
- add argv commands to prompt for new chains or updated chains and validator/threshold combo:
`yarn ts-node scripts/create-keys -e testnet4 --context hyperlane --new-chains basesepolia=2/3` (here 2 is desired threshold and 3 is desired multisig set size) 
  - if you provide a newChainConfigs chain and don't supply an argv, the create-keys script will revert
- persist addresses in gcp (already present) and under sdk (this PR) with types preserved

### Drive-by changes

- imports
- might break other scripts like delete-keys, etc in infra (not tested)

### Related issues

- fixes https://github.com/hyperlane-xyz/issues/issues/887

### Backward compatibility

Yes

### Testing

Manual

PR #6111, opened 2025-05-02
fix: persist key addresses in GCP

tkporter:
### Description

- Starts persisting key addresses in GCP again, fixing key funder infra
- Regression introduced in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3159/files#diff-604c29d425429647a37bc7b46d72cfddb8f5ba0c40d81609234c36bb2a203b35L372 ages ago, just never realized because we didn't add new keys


paulbalaji:
probs wanna pull cherry-pick `7cd88e3d41059a83c45416a2d5d1cf908f22e9a3` as well to fix infra-test
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6089/commits/7cd88e3d41059a83c45416a2d5d1cf908f22e9a3

PR #6113, opened 2025-05-02
chore: implemented cosmos native ism module

troykessler:
### Description

This PR implements the ISM module for Cosmos Native. Following PRs will use this to support the goal of having full cosmos native support in the hyperlane CLI.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual testing with local cosmos chain

### DIFF
--- a/typescript/cosmos-sdk/src/hyperlane/interchain_security/query.ts
+++ b/typescript/cosmos-sdk/src/hyperlane/interchain_security/query.ts
@@ -16,13 +16,19 @@ type QueryDecodedIsmsResponse = {
   pagination: pagination.PageResponse | undefined;
 };
 
+export const NoopIsmTypeUrl = '/hyperlane.core.interchain_security.v1.NoopISM';
+export const MerkleRootMultisigIsmTypeUrl =
+  '/hyperlane.core.interchain_security.v1.NoopISM';
+export const MessageIdMultisigTypeUrl =
+  '/hyperlane.core.interchain_security.v1.NoopISM';
+
 export const decodeIsm = (ism: any.Any | undefined): ISM => {
   switch (ism?.type_url) {
-    case '/hyperlane.core.interchain_security.v1.NoopISM':
+    case NoopIsmTypeUrl:
       return isTypes.NoopISM.decode(ism.value);
-    case '/hyperlane.core.interchain_security.v1.MerkleRootMultisigISM':
+    case MerkleRootMultisigIsmTypeUrl:
       return isTypes.MerkleRootMultisigISM.decode(ism.value);
-    case '/hyperlane.core.interchain_security.v1.MessageIdMultisigISM':
+    case MessageIdMultisigTypeUrl:
       return isTypes.MessageIdMultisigISM.decode(ism.value);
     default:
       throw new Error(`can not decode ISM with type url ${ism?.type_url}`);
--- a/typescript/sdk/src/ism/CosmosNativeIsmModule.ts
+++ b/typescript/sdk/src/ism/CosmosNativeIsmModule.ts
@@ -1,0 +1,163 @@
+import { SigningHyperlaneModuleClient } from '@hyperlane-xyz/cosmos-sdk';
+import {
+  Address,
+  ChainId,
+  Domain,
+  ProtocolType,
+  assert,
+  deepEquals,
+  rootLogger,
+} from '@hyperlane-xyz/utils';
+
+import {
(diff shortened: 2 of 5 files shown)


typescript/cosmos-sdk/src/hyperlane/interchain_security/query.ts
```diff
@@ -16,13 +16,19 @@ type QueryDecodedIsmsResponse = {
   pagination: pagination.PageResponse | undefined;
 };
 
+export const NoopIsmTypeUrl = '/hyperlane.core.interchain_security.v1.NoopISM';
+export const MerkleRootMultisigIsmTypeUrl =
+  '/hyperlane.core.interchain_security.v1.MerkleRootMultisigISM';
+export const MessageIdMultisigTypeUrl =
+  '/hyperlane.core.interchain_security.v1.MessageIdMultisigISM';
```
xeno097, line 23: This is just a personal preference, but I'd like this to be an enum instead of separate constants

typescript/sdk/src/ism/CosmosNativeIsmReader.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // perform type narrowing
+    if ('validators' in ism && 'threshold' in ism) {
+      return {
+        type: IsmType.MESSAGE_ID_MULTISIG,
+        address,
+        validators: ism.validators,
+        threshold: ism.threshold,
+      };
+    }
```
xeno097, line 83: is there a way to narrow the type without having to use the `in` operator? I say this because in the rare case where we might rename those fields we would need to remember to update the assertion here otherwise it will fail even if the underlying ISM is of the correct type

typescript/sdk/src/ism/CosmosNativeIsmReader.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // perform type narrowing
+    if ('validators' in ism && 'threshold' in ism) {
+      return {
+        type: IsmType.MERKLE_ROOT_MULTISIG,
+        address,
+        validators: ism.validators,
+        threshold: ism.threshold,
+      };
+    }
```
xeno097, line 60: Same here as https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6113/files#r2071738655

typescript/sdk/src/ism/CosmosNativeIsmReader.ts
troykessler: will do it with generics so I have the correct ism type directly

xeno097 (approved):

typescript/sdk/src/ism/CosmosNativeIsmModule.ts
```diff
@@ -0,0 +1,163 @@
...
+      );
+    }
+
+    // save current config for comparison
+    // normalize the config to ensure it's in a consistent format for comparison
+    const actualConfig = normalizeConfig(await this.read());
+    expectedConfig = normalizeConfig(expectedConfig);
+
+    assert(
+      typeof expectedConfig === 'object',
```
ltyu, line 84: is this needed? `if (typeof expectedConfig === 'string') { throw ... }` should ensure that it's not a string

typescript/cosmos-sdk/src/hyperlane/interchain_security/query.ts
```diff
@@ -7,22 +7,28 @@ type ISM =
...
 
-type QueryDecodedIsmsResponse = {
-  isms: ISM[];
+type QueryDecodedIsmsResponse<T> = {
+  isms: T[];
   pagination: pagination.PageResponse | undefined;
 };
 
+export enum IsmTypes {
+  NoopISM = '/hyperlane.core.interchain_security.v1.NoopISM',
```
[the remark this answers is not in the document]
troykessler: The `v1` in the type url path will most likely never change, I would leave it as is

typescript/sdk/src/ism/CosmosNativeIsmModule.ts
troykessler: so it is also in `EvmIsmModule`: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/troy/cosmos-ism-module/typescript/sdk/src/ism/EvmIsmModule.ts#L102, so I would also include it here

ltyu (approved):

PR #6114, opened 2025-05-02
feat: chunk message db insertions

kamiyaa:
### Description

Instead of inserting all the messages we receive in a single db transaction, we chunk it into smaller sizes.
This prevents us from running into `Too many arguments` errors from Postgres for exceeding their `u16::MAX` parameter count.

### Related issues
- fixes https://linear.app/hyperlane-xyz/issue/ENG-1390/scraper-cannot-insert-too-many-records

### Backward compatibility

Yes

### Testing

Manual. From Jeff:

Tried it locally with a postgres instance with 100k messages and it failed without chunking 
```
running 1 test
RES: Err(Query Error: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)

Caused by:
   0: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)
   1: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)

Location:
    agents/scraper/src/db/message.rs:320:9)
test db::message::tests::test_store_dispatched_messages_real_postgres ... ok
```

Tried it again with chunking + transaction and it didn't have any errors :+1:
So we can use the transaction if we want the action to be atomic, but the important part is the chunking.

### DIFF
--- a/rust/main/agents/scraper/src/db/message.rs
+++ b/rust/main/agents/scraper/src/db/message.rs
@@ -284,9 +284,11 @@ impl ScraperDb {
         messages: impl Iterator<Item = StorableMessage<'_>>,
     ) -> Result<u64> {
         let origin_mailbox = address_to_bytes(origin_mailbox);
+
         let latest_id_before = self
             .latest_dispatched_id(domain, origin_mailbox.clone())
             .await?;
+
         // we have a race condition where a message may not have been scraped yet even
         let models = messages
             .map(|storable| message::ActiveModel {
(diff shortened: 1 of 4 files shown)


rust/main/agents/scraper/src/db/message.rs
```diff
@@ -32,6 +32,14 @@ pub struct StorableMessage<'a> {
 }
 
 impl ScraperDb {
+    /// Used for store_dispatched_messages()
+    /// message::ActiveModel has 11 fields, on conflict has 3,
+    /// update has 6. So there should be about a maximum of 20 sql
+    /// parameters per ActiveModel
+    /// u16::MAX (65_535u16) is the maximum amount of parameters we can
+    /// have for Postgres. So 65000 / 20 = 3250
+    const STORE_MESSAGE_CHUNK_SIZE: usize = 3250;
```
kamiyaa, line 41: kind of a magic number I came up with, so open to changing this.

rust/main/agents/scraper/src/db/message.rs
```diff
@@ -305,25 +313,27 @@ impl ScraperDb {
...
-                    message::Column::Recipient,
-                    message::Column::MsgBody,
-                    message::Column::OriginTxId,
-                ])
-                .to_owned(),
-            )
-            .exec(&self.0)
-            .await?;
+        for chunk in models.chunks(Self::STORE_MESSAGE_CHUNK_SIZE) {
+            Insert::many(chunk.to_vec())
```
kamiyaa, line 317: Not sure if theres a way to prevent cloning the chunk here.

ameten: Ok

ameten:
I wonder how we can test that the query is executed inside a transaction?

kamiyaa:
> I wonder how we can test that the query is executed inside a transaction?

let me see if I can write a unittest for it :+1:

ameten (approved):

rust/main/agents/scraper/src/db/message.rs
```diff
@@ -336,3 +360,136 @@ impl ScraperDb {
...
+            .map(|v| v.to_vec())
+            .collect();
+
+        let mock_result: BTreeMap<&str, _> = [
+            ("num_items", Into::<Value>::into(10000i64)),
+            ("last_insert_id", Into::<Value>::into(10000i64)),
+        ]
+        .into_iter()
+        .collect();
+        let mock_db = MockDatabase::new(DatabaseBackend::Postgres)
```
daniel-savu, line 404: it seems like this doesn't actually create a postgres instance. Were you able to reproduce the `too many arguments` error with this setup?

daniel-savu: asking because i'm not convinced that creating a `transaction` of insertions will fix this. Would like to see a doc / reference that confirms that a "transaction" of smaller insertion won't run into the same error. Since atomicity for these insertions is not a requirement (because we don't expect write race conditions on these items), it may be safer to just not use a "transaction"

kamiyaa: Tried it locally with a postgres instance with 100k messages and it failed without chunking 
```
running 1 test
RES: Err(Query Error: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)

Caused by:
   0: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)
   1: encountered unexpected or invalid data: PgConnection::run(): too many arguments for query: 1000000 (sqlx_postgres::connection::executor:215)

Location:
    agents/scraper/src/db/message.rs:320:9)
test db::message::tests::test_store_dispatched_messages_real_postgres ... ok
```

Tried it again with chunking + transaction and it didn't have any errors :+1:
So we can use the transaction if we want the action to be atomic, but the important part is the chunking.

daniel-savu (approved):

rust/main/agents/scraper/src/db/message.rs
daniel-savu: nice! I updated the testing section of the PR description with this

kamiyaa: Thanks!

PR #6125, opened 2025-05-05
feat: Redesign key-funder to support multiple protocols  + Add Sealevel support

Mo-Hussain:


typescript/infra/scripts/funding/multi-protocol-key-funder.ts
```diff
@@ -0,0 +1,472 @@
...
+      chainsToFund.includes(chain),
+    ),
+  );
+
+  // Add ethereum metadata if we have L2 chains that need it for bridging
+  const needsL1 = chainsToFund.some((chain) => L2_CHAINS.includes(chain));
+
+  if (needsL1) {
+    filteredChainMetadata.ethereum = chainMetadata.ethereum;
+  }
```
paulbalaji, line 156: i suppose this is a design decision of whether we still want to do the L1->L2 bridging for some of these l2s? imo though it's cool we don't use it nearly as much as we should and it only adds complexity to the funder

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
```diff
@@ -0,0 +1,472 @@
...
+async function buildMultiProtocolProviderOptions(
+  environment: DeployEnvironment,
+  chainsToFund: ChainName[],
+): Promise<MultiProtocolProviderOptions> {
+  const options: MultiProtocolProviderOptions = {
+    signers: {},
+  };
+
+  // Set up signers for each chain
+  for (const chain of chainsToFund) {
```
paulbalaji, line 381: the `getKeyForRole` could be refactored out of the if statements and the only thing that changes between protocol types is how you set the signer from the key

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
```diff
@@ -0,0 +1,472 @@
...
+    for (const role of roles) {
+      let balance: string | undefined;
+
+      if (role === Role.Relayer) {
+        balance = desiredBalancePerChain[chain];
+      } else if (role === Role.Kathy) {
+        balance = desiredKathyBalancePerChain?.[chain];
+      }
+
+      if (!balance) continue;
```
paulbalaji, line 306: maybe worth a debug log

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
```diff
@@ -0,0 +1,472 @@
...
+      const fundingConfig: FundingConfig = {
+        skipIgpClaim: argv.skipIgpClaim,
+        fundingThresholdFactor: DEFAULT_FUNDING_THRESHOLD_FACTOR,
+      };
+
+      // add funding discount factor for RC context
+      let fundingDiscountFactor: number | undefined;
+      if (ctx === Contexts.ReleaseCandidate) {
+        fundingDiscountFactor = RC_FUNDING_DISCOUNT_FACTOR;
+      }
```
paulbalaji, line 210: maybe flip it so hyperlane is always normal and !hyperlane context has a discount factor

in case we decide to add neutron it means we don't have to add discount factors for every new context we add

typescript/infra/src/funding/MultiProtocolContextFunder.ts
```diff
@@ -0,0 +1,78 @@
...
+        } finally {
+          cleanup();
+        }
+      }),
+    );
+
+    if (results.some((result) => result.status === 'rejected')) {
+      this.logger.error('One or more chains failed to fund');
+      throw new Error('One or more chains failed to fund');
+    }
```
paulbalaji, line 79: log the failed chains just so it's also all in one line at the end?

typescript/infra/src/funding/adapters/EVMFundingAdapter.ts
```diff
@@ -0,0 +1,415 @@
...
+    };
+    this.L2Chains = L2_CHAINS;
+    this.L2ToL1 = {
+      ...L2_CHAINS.reduce(
+        (acc, chain) => {
+          acc[chain] = L1_CHAIN;
+          return acc;
+        },
+        {} as Record<ChainName, ChainName>,
+      ),
```
paulbalaji, line 78: see other comment about whether l1->l2 bridging is even a desirable thing in the keyfunder

typescript/infra/src/funding/adapters/EVMFundingAdapter.ts
```diff
@@ -0,0 +1,415 @@
...
+    );
+    return delta.gt(minDelta) ? delta.toString() : '0';
+  }
+
+  async fundKey(
+    key: BaseAgentKey,
+    desiredBalance: number,
+    fundingThresholdFactor: number,
+  ): Promise<void> {
+    if (await this.shouldBridgeToL2()) {
```
paulbalaji, line 122: said already but just marking the l1->l2 bridging design decision again next to some more of the code we could consider removing

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
```diff
@@ -0,0 +1,303 @@
...
+  ) {
+    super(chainName, multiProvider, fundingAddresses, logger);
+    this.logger = logger;
+    // TODO: this should be a config that is passed in
+    this.igpProgramIds = {
+      solanamainnet: 'BhNcatUDC2D5JTyeaqrdSukiVFsEHK7e3hVmKMztwefv',
+      sonicsvm: '3kd3K1yymsfYoqzNnktLwpKXKeB9Y2PRf9rfb31k4o13',
+      soon: '5AAMBemjUrAprKJzi22Si9pyYhK8vTr7Jpti7LEXo6d8',
+      eclipsemainnet: 'Hs7KVBU67nBnWhDPZkEFwWqrFMUfJbmY2DQ4gmCZfaZp',
+    };
```
paulbalaji, line 57: would like the TODO to be done before merging

svmbnb is missing from the list

also the addresses should all be available in the chain metadata that the multiprovider has right?

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
```diff
@@ -0,0 +1,303 @@
...
+    this.igpAdapter = new SealevelIgpAdapter(
+      this.chainName,
+      this.multiProvider,
+      {
+        igp: this.fundingAddresses.interchainGasPaymaster,
```
paulbalaji, line 62: asked earlier but if igp is the only thing in fundingAddresses - is it even needed? I believe mpp/mp extend the chain metadata manager or something so should have all the addresses+metadata we need 

ie getChainMetadata(chain).interchainGasPaymaster or something like that, but could be wrong

paulbalaji: ditto in the other adapter

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
paulbalaji, line 1: seems fine to me but maybe one for @xeno097 to check the changes as well since i'm not too familiar with sealevel practices

typescript/infra/src/funding/MultiProtocolFundingApp.ts
```diff
@@ -0,0 +1,103 @@
...
+  ) {
+    super(multiProtocolProvider, addresses, logger);
+    this.logger = logger;
+  }
+
+  override protocolToAdapter(
+    protocol: ProtocolType,
+  ): AdapterClassType<IFundingAdapter> {
+    if (protocol === ProtocolType.Ethereum) return EVMFundingAdapter;
+    if (protocol === ProtocolType.Sealevel) return SealevelFundingAdapter;
```
paulbalaji, line 54: qq: does this mean when this was run and accidentally funded the vanguard contexts, we've funded the vanguards on solana too? 

if so, then we have a bunch of sol sitting on the vanguard keys that haven't been claimed back yet?

paulbalaji (commented):
lot of changes here that would've been nice to see a design doc for before hand. overall idea seems fine, as mentioned offline i'd like a doc on the rollout plan to discuss in the next backend weekly

Asked in the comments but also reposting up top here: do we need the L1->L2 bridging if it's only used for like 3 chains and adds complexity to the code?

Another question overall though is if this is the right approach for a keyfunder redesign? yes it's a multi-protocol'd version of our existing keyfunder, but was that design the correct one to begin with? what if we had an even more generic keyfunder which had no concept of contexts or roles, but just a set of chains and balances to keep up?

Mo-Hussain:
> lot of changes here that would've been nice to see a design doc for before hand. overall idea seems fine, as mentioned offline i'd like a doc on the rollout plan to discuss in the next backend weekly
> 
> Asked in the comments but also reposting up top here: do we need the L1->L2 bridging if it's only used for like 3 chains and adds complexity to the code?
> 
> Another question overall though is if this is the right approach for a keyfunder redesign? yes it's a multi-protocol'd version of our existing keyfunder, but was that design the correct one to begin with? what if we had an even more generic keyfunder which had no concept of contexts or roles, but just a set of chains and balances to keep up?

Happy to do a doc on the role out plan. For context, Trevor and I were thinking to base a more general redesign of the key-funder on the abstractions developed during the rebalancer project. See here https://www.notion.so/hyperlanexyz/Multi-collateral-warp-route-rebalancer-agent-requirements-1c86d35200d68090bcbdedb29ea101c9?pvs=4#1c96d35200d6808aa986d0ed2206df77

The changes in this PR get us to sealevel funding parity as an incremental improvement. That is why I focused on making minimal changes. Having actually made the changes, I can appreciate the PR is larger than expected but it was honestly hard to shoehorn sealevel funding into the current implementation and reason about whether it was working as expected.

As for L1->L2 bridging, again I left it in to make a minimal changes to functionality but would be open to removing it in a more general redesign.

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
Mo-Hussain: Happy to remove it at this stage if you feel strongly

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
Mo-Hussain: The reason why we cant do that at the moment is that the evm keys do not have 'ethereum', the protocol, in the name of the secret i.e hyperlane-mainnet3-key-deployer

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
Mo-Hussain: The igp addresses are in the metadata, the program ids are not currently

typescript/infra/src/funding/MultiProtocolFundingApp.ts
Mo-Hussain: I didn't include sealevel chains in the call and the address for vanguards were not defined for sealevel so it wouldn't have done the funding

typescript/infra/src/funding/helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+  let cleanup: () => void;
+  const promise = new Promise<void>((_, reject) => {
+    const timeout = setTimeout(
+      () => reject(new Error(errorMessage)),
+      timeoutMs,
+    );
+    cleanup = () => clearTimeout(timeout);
+  });
+  return { promise, cleanup: cleanup! };
+}
```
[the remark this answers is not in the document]
Mo-Hussain: Replaced usage of this in the new implementation

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
Mo-Hussain: There is not a particularly clean way on handling this right now, one option is to pass this as an optional field on `FunderAddresses` like igpProgramId that we only set for sealevel chains. We would then need to maintain this mapping elsewhere, likely the entrypoint script 

As for having them live in registry, IIUC we don't have protocol specific contract addresses in there at the moment so this would be a broader change

xeno097: also considering that the program id is required for svm chains maybe maybe we can add a method/update the registry so that the user can list all the chains of a specified protocol type so we can check in the constructor if we have a value for all the svm chains. Probably it can be done with `getChains` and filtering by protocol type rn

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
```diff
@@ -0,0 +1,304 @@
...
+    const response = await this.getProvider().getBalanceAndContext(
+      new PublicKey(address),
+    );
+    // Get the raw value from the RPC response before it's converted to number
+    return BigInt(response.value);
+  }
+
+  async getFundingAmount(
+    address: string,
+    desiredBalance: number,
```
xeno097, line 78: Maybe the desired balances by chain can be provided as a config object in the constructor? wdyt?

typescript/infra/src/funding/adapters/SealevelFundingAdapter.ts
```diff
@@ -0,0 +1,304 @@
...
+          }),
+        );
+      }
+
+      // Send and confirm transaction
+      const signature = await sendAndConfirmTransaction(
+        this.getProvider(),
+        transaction,
+        [this.getSigner()],
+      );
```
xeno097, line 193: Can we have some retry logic here, solana is quite messy when trying to land txs, like if the error is `transaction blockchash expired` we retry otherwise we just throw

typescript/sdk/src/providers/MultiProtocolProvider.ts
```diff
@@ -60,7 +64,7 @@ export class MultiProtocolProvider<
   // Chain name -> provider type -> provider
   protected readonly providers: ChainMap<ProviderMap<TypedProvider>>;
   // Chain name -> provider type -> signer
-  protected signers: ChainMap<ProviderMap<never>> = {}; // TODO signer support
+  protected signers: ChainMap<ProviderMap<TypedSigner>>;
```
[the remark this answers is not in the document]
xeno097: lgtm too

typescript/infra/scripts/funding/multi-protocol-key-funder.ts
Mo-Hussain: Curious if you had thoughts @tkporter

PR #6135, opened 2025-05-05
feat(submitter): evm-specific fields in Payload and Transaction

daniel-savu:
### Description

The `VmSpecificTxData` enum is changed for the `Evm` to now wrap a `TypedTransaction`. No changes to `Payload` are required to support EVM chains, since @ameten suggested serializing everything as a vec of bytes in the `data` field.

- One concern I have is that the serialization of `VmSpecificTxData` may change as a result of this. Memory is allocated for Enums as: "The size will be the size of an integer specifying which of the two variants it is, plus the size of the largest variant, plus (sometimes) padding between the integer and the payload." ([source](https://www.reddit.com/r/rust/comments/t6fzcv/comment/hzaxlrx/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button), from one of the tokio maintainers). More details here too: https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#layout-of-a-data-carrying-enums-without-a-repr-annotation. 
- The changes may cause the `VmSpecificTxData` stored in rocksdb for sealevel to be interpreted differently now - but hopefully that's not a big concern since it's just for Transactions on RC.

### Drive-by changes

Adds `ethers` as a dependency on the submitter, to gain access to `TypedTransaction`

### Related issues

Fixes: https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/6127

### Backward compatibility

This change may be breaking to the `VmSpecificTxData` format stored in the RC db

### Testing

Covered by unit tests and E2E


ameten:
https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6135#issue-3039643711

We are serialising transaction into json before converting into a byte vector. I wonder if these concerns are relevant in this case?

ameten (approved):

daniel-savu:
> [#6135 (comment)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6135#issue-3039643711)
> 
> We are serialising transaction into json before converting into a byte vector. I wonder if these concerns are relevant in this case?

@ameten the byte vector format is compatible, but parsing that into json may fail due to the new enum format. Still don't think it's a big issue though

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/payload.rs
```diff
@@ -0,0 +1,15 @@
+use ethers::abi::Function;
+use ethers::types::transaction::eip2718::TypedTransaction;
+
+use crate::payload::FullPayload;
+
+pub(crate) trait Precursor {
```
daniel-savu, line 6: this trait is a duplicate of the one in the sealevel adapter, we should move it out of the vm-specific adapters and reuse it

PR #6137, opened 2025-05-05
fix(sdk, cli): fix cli sdk trying to derive configs from 0 address [eng-1551]

xeno097:
### Description

This PR fixes a bug in the warp reader that assumed that all warp routes are deployed as proxies when it doesn't seem to be the case, which resulted in the on-chain config derivation failing because the `.owner` field was called on the 0 address when deriving the proxy admin config 

### Drive-by changes

- None

### Related issues

- [ENG-1551](https://linear.app/hyperlane-xyz/issue/ENG-1551/fix-warp-checker-reader-trying-to-derive-info-from-the-0-address)

### Backward compatibility

- YES

### Testing

- Manual
- unit
- e2e

### DIFF
--- a/typescript/sdk/src/token/configUtils.ts
+++ b/typescript/sdk/src/token/configUtils.ts
@@ -2,12 +2,15 @@ import { zeroAddress } from 'viem';
 
 import {
   Address,
+  ProtocolType,
   TransformObjectTransformer,
   addressToBytes32,
   objMap,
+  promiseObjAll,
   transformObj,
 } from '@hyperlane-xyz/utils';
 
+import { isProxy } from '../deploy/proxy.js';
 import { MultiProvider } from '../providers/MultiProvider.js';
 import { DestinationGas, RemoteRouters } from '../router/types.js';
 import { ChainMap } from '../types.js';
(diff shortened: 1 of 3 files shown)


ltyu (approved):

typescript/sdk/src/token/configUtils.ts
```diff
@@ -109,7 +125,9 @@ export async function expandWarpDeployConfig(
       destinationGas,
       hook: zeroAddress,
       interchainSecurityModule: zeroAddress,
-      proxyAdmin: { owner: config.owner },
+      proxyAdmin: isDeployedAsProxyByChain[chain]
+        ? { owner: config.owner }
```
ltyu, line 129: why is this defaulted to the top level owner and not read from the proxyAdmin.owner?

xeno097: Spreading the `config` below will override it if it was defined in the config, but otherwise, the default behaviour is that both the token and the proxy admin have the same owner

ltyu: ah missed that

PR #6140, opened 2025-05-05
Implement `InterchainAccount` router program outbound from SVM

yorhodes:

### DIFF
--- a/rust/sealevel/programs/interchain-accounts/src/accounts.rs
+++ b/rust/sealevel/programs/interchain-accounts/src/accounts.rs
@@ -1,0 +1,86 @@
+//! InterchainAccount accounts.
+use std::collections::HashMap;
+
+use access_control::AccessControl;
+use account_utils::{AccountData, SizedData};
+use borsh::{BorshDeserialize, BorshSerialize};
+use hyperlane_core::H256;
+use hyperlane_sealevel_connection_client::{
+    router::{HyperlaneRouter, RemoteRouterConfig},
+    HyperlaneConnectionClient,
+};
+use hyperlane_sealevel_igp::accounts::InterchainGasPaymasterType;
+
+use solana_program::{program_error::ProgramError, pubkey::Pubkey};
+
+/// The storage account.
+pub type InterchainAccountStorageAccount = AccountData<InterchainAccountStorage>;
+
+/// The storage account's data.
+#[derive(BorshSerialize, BorshDeserialize, Debug, Default)]
+pub struct InterchainAccountStorage {
+    /// The local domain.
+    pub local_domain: u32,
+    /// The mailbox.
+    pub mailbox: Pubkey,
+    /// The ISM.
+    pub ism: Option<Pubkey>,
+    /// The IGP.
+    pub igp: Option<(Pubkey, InterchainGasPaymasterType)>,
+    /// The owner.
+    pub owner: Option<Pubkey>,
+    /// Keyed by domain, the router for the remote domain.
+    pub routers: HashMap<u32, H256>,
+}
+
+impl SizedData for InterchainAccountStorage {
+    fn size(&self) -> usize {
+        // local domain
+        std::mem::size_of::<u32>() +
+        // mailbox
+        32 +
+        // ism
+        1 + 32 +
+        // igp
+        1 + 32 + 1 + 32 +
+        // owner
+        1 + 32 +
+        // routers
+        (self.routers.len() * (std::mem::size_of::<u32>() + 32))
+    }
+}
+
+impl AccessControl for InterchainAccountStorage {
+    fn owner(&self) -> Option<&Pubkey> {
+        self.owner.as_ref()
+    }
+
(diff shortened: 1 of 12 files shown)


rust/sealevel/programs/interchain-accounts/src/processor.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    // Account 1: Payer.
+    let payer_info = next_account_info(accounts_iter)?;
+    if !payer_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
+
+    // Account 2: Storage PDA.
+    let storage_info = next_account_info(accounts_iter)?;
```
tkporter, line 143: believe we'll need to make sure that the program hasn't been initialized before (fwiw I wouldn't be surprised if helloworld doesn't have the correct safety checks, it was meant more for testing than anything)

Take a look at `verify_account_uninitialized` which we can use on the storage_info

rust/sealevel/programs/interchain-accounts/src/processor.rs
```diff
@@ -0,0 +1,507 @@
...
+        &storage
+            .ism
+            .try_to_vec()
+            .map_err(|err| ProgramError::BorshIoError(err.to_string()))?[..],
+    );
+
+    Ok(())
+}
+
+fn set_account_meta_return_data(program_id: &Pubkey) -> ProgramResult {
```
tkporter, line 444: nit: maybe better to name this something specific to the ISM account metas, as I imagine if/when this has a handle fn we'll have the account metas for that be very different

rust/sealevel/programs/interchain-accounts/src/processor.rs
```diff
@@ -0,0 +1,507 @@
...
+        Pubkey::find_program_address(program_storage_pda_seeds!(), program_id);
+    if storage_info.key != &expected_storage_pda_key {
+        return Err(ProgramError::InvalidArgument);
+    }
+    let mut storage =
+        InterchainAccountStorageAccount::fetch(&mut &storage_info.data.borrow()[..])?.into_inner();
+
+    // Account 2: Owner.
+    let owner_info = next_account_info(accounts_iter)?;
+    storage.ensure_owner_signer(owner_info)?;
```
tkporter, line 494: nit: I believe `enroll_remote_routers_only_owner` does this `ensure_owner_signer` as part of its logic, so no need to have this twice

rust/sealevel/programs/interchain-accounts/src/instruction.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub igp: Option<(Pubkey, InterchainGasPaymasterType)>,
+    /// The owner.
+    pub owner: Option<Pubkey>,
+}
+
+/// CallRemote instruction data.
+#[derive(BorshSerialize, BorshDeserialize, Debug)]
+pub struct CallRemoteMessage {
+    /// The address of the owner on the origin domain.
+    pub sender: Pubkey,
```
tkporter, line 34: Any reason to not have this be implied by the accounts passed in? this is the typical pattern on Solana and would save us 32 bytes in the tx as well

rust/sealevel/programs/interchain-accounts/src/processor.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    let instruction = InterchainAccountInstruction::try_from_slice(instruction_data)
+        .map_err(|_| ProgramError::InvalidInstructionData)?;
+    match instruction {
+        InterchainAccountInstruction::Init(mailbox) => init(program_id, accounts, mailbox),
+        InterchainAccountInstruction::CallRemote(remote_call) => {
+            send_call_remote(program_id, accounts, remote_call)
+        }
+        InterchainAccountInstruction::SetInterchainSecurityModule(ism) => {
```
tkporter, line 112: should also have a TransferOwnership instruction

tkporter (commented):
Great start, didn't do the deepest pass but nothing super blatant stands out atm

rust/sealevel/programs/interchain-accounts/src/accounts.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub local_domain: u32,
+    /// The mailbox.
+    pub mailbox: Pubkey,
+    /// The ISM.
+    pub ism: Option<Pubkey>,
+    /// The IGP.
+    pub igp: Option<(Pubkey, InterchainGasPaymasterType)>,
+    /// The owner.
+    pub owner: Option<Pubkey>,
+    /// Keyed by domain, the router for the remote domain.
```
tkporter, line 32: fwiw us keeping routers in the hashmap here means that we have some limits on how large the routers can get. I think it's something like 50 where we max out. Might be worth us considering a more scalable approach in this new program, or we can always make the changes later. The scalable approach would be to store each router in its own account

yorhodes: we actually can allow the caller to specify their own destination rotuer

PR #6141, opened 2025-05-05
feat: Add typescript relayer support for CCIP-read ISM

nambrot:
### Description

Adds CCIP-read ISM support for the typescript relayer (will rename in a follow-up PR). Adds sdk test for it.

### Backward compatibility

Yes

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/isms/ccip-read/TestCccipReadIsm.sol
+++ b/solidity/contracts/isms/ccip-read/TestCccipReadIsm.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity >=0.8.0;
+
+error OffchainLookup(
+    address sender,
+    string[] urls,
+    bytes callData,
+    bytes4 callbackFunction,
+    bytes extraData
+);
+
+import "../../interfaces/isms/ICcipReadIsm.sol";
+import "../../interfaces/IInterchainSecurityModule.sol";
+import "../../interfaces/IMailbox.sol";
+import "../../libs/Message.sol";
+import "./AbstractCcipReadIsm.sol";
+
+/**
+ * @title TestCcipReadIsm
+ * @notice A test CCIP-Read ISM that simply checks the passed metadata as a boolean.
+ */
+contract TestCcipReadIsm is AbstractCcipReadIsm {
+    string[] public urls;
+
+    constructor(string[] memory _urls) {
+        urls = _urls;
+    }
+
+    function getOffchainVerifyInfo(
+        bytes calldata _message
+    ) external view override {
+        // Revert with OffchainLookup to instruct off-chain resolution
+        revert OffchainLookup(address(this), urls, _message, bytes4(0), "");
+    }
+
+    function verify(
+        bytes calldata metadata,
+        bytes calldata
+    ) external view override returns (bool) {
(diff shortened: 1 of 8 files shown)


typescript/sdk/package.json
```diff
@@ -16,6 +16,7 @@
     "@safe-global/safe-deployments": "1.37.23",
     "@solana/spl-token": "^0.4.9",
     "@solana/web3.js": "^1.95.4",
+    "axios": "^1.9.0",
```
yorhodes, line 19: channeling my inner @jmrossy 
native fetch is sufficient in 2025

typescript/sdk/src/ism/types.ts
```diff
@@ -262,6 +268,10 @@ export const ArbL2ToL1IsmConfigSchema = z.object({
   bridge: z.string(),
 });
 
+export const CCIPReadIsmConfigSchema = z.object({
+  type: z.literal(IsmType.CCIP_READ),
+});
```
yorhodes, line 273: I would expect the URLs to be included here

typescript/sdk/src/ism/metadata/ccipread.ts
```diff
@@ -0,0 +1,69 @@
...
+  readonly type = IsmType.CCIP;
+  private core: HyperlaneCore;
+  private iface: utils.Interface;
+
+  constructor(core: HyperlaneCore) {
+    this.core = core;
+    this.iface = new utils.Interface([
+      'function getOffchainVerifyInfo(bytes) view',
+      'error OffchainLookup(address sender, string[] urls, bytes callData, bytes4 callbackFunction, bytes extraData)',
+    ]);
```
yorhodes, line 21: would prefer if this was derived from the source code and strongly typed

typescript/sdk/src/ism/metadata/ccipread.ts
```diff
@@ -0,0 +1,69 @@
...
+        .replace('{data}', callDataHex);
+      try {
+        const response = urlTemplate.includes('{data}')
+          ? await axios.get(url)
+          : await axios.post(url, { sender, data: callDataHex });
+        const rawHex = response.data.data as string;
+        return rawHex.startsWith('0x') ? rawHex : `0x${rawHex}`;
+      } catch {
+        continue;
+      }
```
yorhodes, line 64: this continue is not necessary as there is nothing else in the for loop body
also should we really fail silently here?

yorhodes (approved):

typescript/sdk/src/ism/types.ts
nambrot: I also thought so, but it turns out we don't require a view call from `ICcipReadIsm`, so it could technically be dynamic

nambrot: We discussed offline that we can derive this from the message, but will add this later

solidity/contracts/isms/ccip-read/TestCccipReadIsm.sol
paulbalaji, line 1: file name has an extra C

also a question on naming - should it stick with `*CcipRead*` or `*CCIPRead*`

nambrot: fukkkk

PR #6142, opened 2025-05-05
fix: Add E2E test with mocked bridge

fzavalia:
Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/44

Includes https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6087

Adds an e2e test that listens to the event emitted by the mocked value transfer bridge and performs an actual transfer to achieve balance


fernandomg (approved):
👨‍🍳

PR #6143, opened 2025-05-05
feat(sdk):  fiat collateral token derivation support in sdk [eng-1552]

xeno097:
### Description

This PR updates the `EvmERC20WarpRouteReader` to correctly derive the type of a fiat collateral warp route deployment that was incorrectly inferred as collateral.

### Drive-by changes

- Refactors the `fetchTokenConfig` to delegate token config derivation to a separate function instead of having the whole derivation logic in one place with nested if statements

### Related issues

- Fixes [ENG-1552](https://linear.app/hyperlane-xyz/issue/ENG-1552/add-derivation-support-for-fiat-collateral-in-the-sdk)

### Backward compatibility

- YES

### Testing

- Manual
- e2e
- unit


.changeset/wise-hairs-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+Adds fiatCollateral token on chain config derivation logic as it was correctly inferred as collateral
```
antigremlin, line 5:
```suggestion
Adds fiatCollateral token on chain config derivation logic as it was incorrectly inferred as collateral
```

typescript/sdk/src/token/EvmERC20WarpRouteReader.hardhat-test.ts
```diff
@@ -412,6 +423,33 @@ describe('ERC20WarpRouterReader', async () => {
     expect(derivedConfig.token).to.equal(token.address);
   });
 
+  // FiatTokenTest
+  it('should derive fiat collateral config correctly', async () => {
```
antigremlin, line 427:
```suggestion
  it('should derive collateral fiat token type correctly', async () => {
```

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -141,12 +145,30 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+            );
+
+            // Simulate minting tokens from the warp route contract
+            await fiatToken.callStatic.mint(NON_ZERO_SENDER_ADDRESS, 1, {
+              from: warpRouteAddress,
+            });
+
+            return TokenType.collateralFiat;
+            // eslint-disable-next-line no-empty
+          } catch {}
```
antigremlin, line 171: Do we always ignore exceptions in calls to contracts? Or should we at least check the exception type to come from `ethers` and re-throw unexpected exceptions (I'm not sure if this is possible)?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -261,87 +283,193 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
-      } else {
-        const erc20 = HypERC20Collateral__factory.connect(
-          warpRouteAddress,
-          this.provider,
-        );
-        token = await erc20.wrappedToken();
-      }
+    // Using null instead of undefined to force
+    // a compile error when adding a new token type
+    const mapConfig: Record<
```
antigremlin, line 288: This map is the same every time. Why not make it an instance variable, and initialize in the constructor?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -261,87 +283,193 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+      TokenType,
+      ((address: Address) => Promise<HypTokenConfig>) | null
```
antigremlin, line 290: Would it be cleaner if the type is `(arg => result) | undefined` and we skip the entries initialized to null?

antigremlin (commented):
Nice refactoring!

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
xeno097: it would have the same result as having the `| null` to avoid initialization entries that currently have null values I can wrap the record into a `Partial` and that would not require all keys of the map to be specified but that would go against my goal of having the map in the first place because i'd like users that add a new TokenType to be "notified" by the compiler that they need to implement (or not) a new token config derivation function instead of simply adding the new type and forgetting this part because there are no compile time errors only to discover that the new type is not derivable when using the cli and having an error or an incorrect data read like it was before. Wdyt? @antigremlin

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -141,12 +172,38 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+          }
+
+          try {
+            const fiatToken = IFiatToken__factory.connect(
+              wrappedToken,
+              this.provider,
+            );
+
+            // Simulate minting tokens from the warp route contract
+            await fiatToken.callStatic.mint(NON_ZERO_SENDER_ADDRESS, 1, {
```
ltyu, line 196: isn't mint() on every erc20 token?

should we use fiatToken specifc fns like minterAllowance()

https://github.com/circlefin/stablecoin-evm/blob/master/contracts/v1/FiatTokenV1.sol

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -141,12 +172,38 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
-            // eslint-disable-next-line no-empty
-          } catch {}
+          } catch (error) {
+            this.logger.debug(
+              `Warp route token at address "${warpRouteAddress}" on chain "${this.chain}" is not a ${TokenType.XERC20}`,
+              error,
+            );
+          }
+
+          try {
```
ltyu, line 189: why do we need separate try-catch? aren't these 2 mutually exclusive? in this case, we are inferring to the reader that the warp route can be xerc20 and fiat token.

perhaps we should make use of `continue`? or perhaps another way?

xeno097: It can't becuase if we find out that the token is an XERC20 then we just return inside the try catch, same for the fiat collateral :
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2c7ba45fa7f45db2e00baddf999ff5f9a544b70/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts#L175-L186
- https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e2c7ba45fa7f45db2e00baddf999ff5f9a544b70/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts#L189-L200

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: it seems circle specific
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/token/interfaces/IFiatToken.sol#L4 

perhaps the author only included this terse interface so we don't need to bring in FiatToken package completely

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: ah yes my mistake. misread this

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: discussed offline, FiatToken may be a misnomer and should probably be MintableToken

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -261,87 +318,168 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
-        throw new Error(
-          `Warp route config specifies native token but chain metadata for ${this.chain} does not provide native token details`,
-        );
-      }
-    } else {
+    return {
+      ...erc20TokenMetadata,
+      type: TokenType.XERC20,
+      token: collateralTokenAddress,
+      xERC20: xERC20Metadata.xERC20,
```
ltyu, line 350: in the above Promise.all, if this.fetchXERC20Config() throws then it would return an empty object. that means the `xERC20` would be undefined. is this okay, or should fetchXERC20Config _only_ throw?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -261,87 +318,168 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+      this.fetchERC20Metadata(xerc20Token),
+      this.fetchXERC20Config(xerc20Token, hypTokenAddress),
+      hypTokenInstance.lockbox(),
+    ]);
+
+    return {
+      ...erc20TokenMetadata,
+      type: TokenType.XERC20Lockbox,
+      token: lockbox,
+      xERC20: xERC20Metadata.xERC20,
```
ltyu, line 373: ditto

ltyu (approved):
thanks for the refactor

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
xeno097: only `fetchXERC20Config` should throw at the moment, we actually need to change something in there because we assume that all the XERC20 tokens comply with the same interface for the limits, which is not true, for example renzo's XERC20's don't have the same interface as velodrome's contracts

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
xeno097: se above 👆

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
xeno097: discussed in dms too

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
antigremlin: Good thinking! If the point of having `| null` is to get a warning from the compiler when we add a new element to `TokenType` but forget to add a derivation function to this map, then this typing is very useful. 🚀

antigremlin (approved):

PR #6145, opened 2025-05-06
feat: Manually reduce desired balance and alert threshold for chains

Mo-Hussain:
### Description
- adds a script `print-dollar-value-key-funder-threshold.ts`, that checks whether the proposed  deployer target balance values would trigger and low urgency key funder alert. (we cont have to merge this script)
- adds `target-dollar-balances.json`, this is the deployer target balance values in dollars.

image

- initial run of the scrip suggested that alerts would trigger for a set of chains (base, bsc, bitlayer, solana, linea)

- I checked without we could safely reduce the balance thresholds for these chains but running `calculate-relayer-daily-burn.ts` in review mode. For the following chains, we were maintaining a desired relayer balance that was higher than the recommended

image
 
 - I reduced the desired balance thresholds for these chains, not using the recommendations but values that are higher to be conservative.

image

- post these reductions, we were still going to trigger alerts for some chains so I increased the targets for some of these chains. I increased targets for Linea and Sonium to include a funding buffer 
- See the commits to view the changes in the different stages

### Drive-by changes

- pass remote chain for logging `Too much underflow when downscaling exchange rate` error, makes it easier to identify the culprit chain

### Testing

Manual


typescript/infra/scripts/funding/print-dollar-value-key-funder-threshold.ts
```diff
@@ -0,0 +1,109 @@
...
+      output += `\n✅ ${result.chain.padEnd(15)} (${result.symbol})`;
+      output += `\n   Alerting balance: ${result.balance.toFixed(6)} ${result.symbol} ($${result.dollarValue.toFixed(2)})`;
+      output += `\n   Target: $${result.targetDollarValue.toFixed(2)}`;
+      output += `\n   Current: ${result.percentageOfTarget.toFixed(1)}% of target`;
+      output += '\n';
+    }
+  }
+
+  // Single console.log for all output
+  console.log(output);
```
paulbalaji, line 103: why a single console.log at the end instead of logging the outputs as you go?

typescript/infra/scripts/funding/print-dollar-value-key-funder-threshold.ts
```diff
@@ -0,0 +1,109 @@
...
+function processChains(
+  chains: readonly ChainName[],
+  balances: Record<ChainName, number>,
+): ChainBalanceInfo[] {
+  const results: ChainBalanceInfo[] = [];
+
+  for (const chain of chains) {
+    const tokenPrice = Number(tokenPrices[chain as keyof typeof tokenPrices]);
+    if (!tokenPrice) {
+      console.error(`No token price found for ${chain}`);
```
paulbalaji, line 42: can we change uses of `console` logging to `rootLogger`, and where possible import and use `chalk` to add more colour/formatting to the output

typescript/infra/scripts/funding/print-dollar-value-key-funder-threshold.ts
```diff
@@ -0,0 +1,109 @@
...
+  balance: number;
+  dollarValue: number;
+  targetDollarValue: number;
+  percentageOfTarget: number;
+  excessFunding: number;
+}
+
+function processChains(
+  chains: readonly ChainName[],
+  balances: Record<ChainName, number>,
```
paulbalaji, line 35: nit: this is referred to as "balances" but it's just the thresholds not the balance itself

PR #6148, opened 2025-05-06
feat(submitter): Build and submit transaction for EVM chains

ameten:
### Description

Build and submit transaction for EVM chains

- Serialise data required to build a ContractCall into byte vector in EthereumMailbox
- Desialise data required to build a ContractCall into transaction precursor
- Submit transaction to chain using ContractCall

### Drive-by changes

- Reformat of workspace Cargo.toml

### Backward compatibility

Yes

### Testing

None


rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
```diff
@@ -683,10 +683,14 @@ where
...
     async fn process_calldata(
         &self,
-        _message: &HyperlaneMessage,
-        _metadata: &[u8],
+        message: &HyperlaneMessage,
+        metadata: &[u8],
     ) -> ChainResult<Vec<u8>> {
-        todo!()
+        let contract_call = self
+            .process_contract_call(message, metadata, None, false)
```
daniel-savu, line 690: I think `process_contract_call` already does all the estimations / network calls we're looking to decouple from the MessageProcessor. This should likely just get the plain ContractCall, with no estimates set in the TypedTransaction
```suggestion
        let contract_call = self
            .contract_call(message, metadata, None)
```

daniel-savu: actually even simpler:
```
        self.contract.process(
            metadata.to_vec().into(),
            RawHyperlaneMessage::from(message).to_vec().into(),
        );
```

rust/main/chains/hyperlane-ethereum/src/rpc_clients/provider.rs
```diff
@@ -83,6 +90,23 @@ where
     ) -> ChainResult<u32> {
         get_finalized_block_number(&*self.provider, reorg_period).await
     }
+
+    async fn send(&self, tx: TypedTransaction, function: Function) -> ChainResult<H256> {
```
daniel-savu, line 94: nit: could we pass `EthereumTxPrecursor` instead of `tx` and `function`?

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        info!(?tx, "submitting transaction");
+
+        let precursor = tx.precursor();
+        let hash = self
+            .provider
+            .send(precursor.tx.clone(), precursor.function.clone())
+            .await?;
+
+        tx.tx_hashes.push(hash.into());
```
daniel-savu, line 141: realizing it's easy to forget to push the tx hash in future adapter implementations of `submit`. Should we change the signature of `submit` to return the tx hash on success, and push this to `tx_hashes` as part of the inclusion step?

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
ameten: Yes, it should work.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/provider.rs
ameten: We'll need to move `EthereumTxPrecursor` from Lander to `hyperlane-ethereum`. It is possible, just creates a dependency beyond RPC calls.

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter.rs
ameten: We update more data inside transaction in `SealevelTxAdapter`. It means that the return type will be more complicated than just hash. I think we can keep it as is.

rust/main/chains/hyperlane-ethereum/src/rpc_clients/provider.rs
daniel-savu: ah makes sense

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter.rs
```diff
@@ -0,0 +1,177 @@
...
+    async fn simulate_tx(&self, _tx: &Transaction) -> Result<bool, SubmitterError> {
+        todo!()
+    }
+
+    async fn estimate_tx(&self, _tx: &mut Transaction) -> std::result::Result<(), SubmitterError> {
+        todo!()
+    }
+
+    async fn submit(&self, tx: &mut Transaction) -> Result<(), SubmitterError> {
+        use crate::chain_tx_adapter::chains::ethereum::transaction::Precursor;
```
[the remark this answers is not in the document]
ameten: It is a different trait, but I simplified the import.

rust/main/chains/hyperlane-ethereum/src/contracts/mailbox.rs
ameten: Done

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter.rs
ameten: So, the transaction as parameter still will be modifiable inside the method, but we return hash and updated it outside of the method as well?

daniel-savu (approved):

PR #6153, opened 2025-05-06
feat: fix message status not being stored correctly

kamiyaa:
### Description

Fix message status not being stored correctly.
Turns out, instead of storing the new message status, we store the old message status,
but set it to the new message status in memory.

### Related issues

 - fixes: https://linear.app/hyperlane-xyz/issue/ENG-1024/bug-relayer-messages-are-loaded-with-the-wrong-operation-status-from

### Drive-bys

 - move some dummy test code to be public so other tests can also use them

### Backward compatibility

Yes

### Testing

 - added a test with a temp dir rocksdb to make sure `PendingMessage.set_status()` works as expected


tkporter (approved):
nice catch!

PR #6154, opened 2025-05-06
chore: Add script to add etherscan keys

ltyu:
### Description
This PR adds a script to copy the etherscan keys into the local Registry. That's it.

### Testing
Manual


paulbalaji (commented):
why not just have this in the registry?

ltyu:
> why not just have this in the registry?

ok

PR #6155, opened 2025-05-07
feat(sdk,cli):  expand ism and hook configs to compare on chain and artifacts configs [eng-1088]

xeno097:
### Description

This PR adds the logic to properly compare the hook and ism configuration depending on whether they are set as an address or a partial config.

### Drive-by changes

- Exports the `DerivedWarpRouteDeployConfig` and `DerivedTokenRouterConfig` types from the SDK package
- Exports the `derivedHookAddress` and `derivedIsmAddress` from the SDK package
- Updates the `runWarpRouteCheck` function signature to have the `onChainConfig` to be of type `DerivedWarpRouteDeployConfig`
- Updates the `transformConfigToCheck` function to remove the gas property from the config to check because it is not derived from on-chain config, and if specified in the deploy config, caused false positives (it is ok because the destinationGas field covers this)
- Updates the EvmIsmReader to expand an ism config if it contains any addresses instead of the full config

### Related issues

- Fixes #[ENG-1553](https://linear.app/hyperlane-xyz/issue/ENG-1553/non-expanded-hook-and-ism-configs-make-cli-checker-fail)

### Backward compatibility

- YES

### Testing

- Manual
- e2e
- unit


ltyu (approved):
lgtm, can we just get a few unit tests if it isn't too hard?

typescript/sdk/src/token/configUtils.ts
```diff
@@ -121,69 +124,104 @@ export async function expandWarpDeployConfig(params: {
...
+
+            return ContractVerificationStatus.Verified;
+          },
+        );
+      }
 
-    chainConfig.remoteRouters = formattedRemoteRouters;
+      // Expand the hook config only if we have an explicit config in the deploy config
+      // and the current chain is an EVM one.
+      // if we have an address we leave it like that to avoid deriving
```
antigremlin: Why did we want to derive nested configs but not addresses? What was the use-case?

PR #6157, opened 2025-05-07
fix: Scraper - Use previous compact mode version in Cosmos HttpClient to fix issue with Neutron chain

ameten:
### Description

Scraper is not indexing payments and delivered messages for Neutron chain since it gets serde parsing error when it attempts to request a block by height.

The issue was reproduced using unit tests with Neutron RPC and particular block. The issue was fixed by downgrading `CompactMode` from latest version (0.38 at the time of this commit) to version 0.37.

### Backward compatibility

Yes

### Testing

- Unit test of pure parsing of JSON
- Unit test of HttpClient to request a block from Neutron
- Unit test of CosmosFallbackProvider to request a block from Neutron.

### DIFF
--- a/rust/main/chains/hyperlane-cosmos/src/providers/rpc/provider/tests.rs
+++ b/rust/main/chains/hyperlane-cosmos/src/providers/rpc/provider/tests.rs
@@ -1,0 +1,54 @@
+use std::str::FromStr;
+
+use tendermint_rpc::client::CompatMode;
+use tendermint_rpc::HttpClient;
+use url::Url;
+
+use hyperlane_core::rpc_clients::FallbackProvider;
+use hyperlane_metric::prometheus_metric::{
+    ClientConnectionType, PrometheusClientMetrics, PrometheusConfig,
+};
+
+use crate::rpc::CosmosRpcClient;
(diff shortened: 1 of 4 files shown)


yjamin (approved):

rust/main/chains/hyperlane-cosmos/src/providers/rpc/provider/tests.rs
```diff
@@ -0,0 +1,54 @@
...
+/// This test passes when HttpClient is initialised with `CompactMode::V0_37` (done in prod code).
+/// This test fails when `CompactMode::V0_38` is used with Neutron url and block height.
+/// This test passes with Osmosis url and block height and any compact mode.
+#[tokio::test]
+#[ignore]
+async fn test_fallback_provider() {
+    use tendermint_rpc::Client;
+
+    // Neutron
+    let url = "https://rpc.neutron-main-eu1.ccvalidators.com/5UTUPnLpjr0A-archive";
```
yjamin, line 24: Are these urls public rpcs? Because they look like private ones to me

ameten: Ah

PR #6158, opened 2025-05-07
fix: build-programs.sh downloading newer solana CLI releases and the old 1.14.20 version

tkporter:
### Description

Solana CLI releases were migrated from https://release.solana.com to https://release.anza.xyz, but along the way some old versions weren't ported over. This includes 1.14.20, which is the version we require to build the programs.

A couple alternatives were considered (trying to use brew to find an older version, more directly querying github, etc), but in the end I think the best approach is to just try to emulate their old installation script and make sure that it fetches from the right github release.

This PR adds the installation script adapted from https://release.anza.xyz/v1.18.18/install with the following changes:
- The version is hardcoded to 1.14.20
- The URL is hardcoded to read from the old solana-labs repo
- All agave references are replaced with solana

It also moves away from using the old release.solana.com url to release.anza.xyz when installing newer versions (like 1.18.18).

### DIFF
--- a/rust/sealevel/programs/install-solana-1.14.20.sh
+++ b/rust/sealevel/programs/install-solana-1.14.20.sh
@@ -1,7 +1,9 @@
+#!/bin/sh
+
 # Solana CLI releases were migrated from https://release.solana.com to
 # https://release.anza.xyz, but along the way some old versions weren't ported over.
 # This includes 1.14.20, which is the version we require to build the programs.
-#
+
 # This script is adapted from https://release.anza.xyz/v1.18.18/install with the following
 # changes:
 # - The version is hardcoded to 1.14.20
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
LGTM - did a manual diffcheck https://www.diffchecker.com/yrjTpqun/

rust/sealevel/programs/install-solana-1.14.20.sh
```diff
@@ -0,0 +1,202 @@
...
+# This includes 1.14.20, which is the version we require to build the programs.
+
+# This script is adapted from https://release.anza.xyz/v1.18.18/install with the following
+# changes:
+# - The version is hardcoded to 1.14.20
+# - The URL is hardcoded to read from the old solana-labs repo
+# - All agave references are replaced with solana
+
+SOLANA_RELEASE=v1.14.20
+SOLANA_INSTALL_INIT_ARGS=v1.14.20
```
paulbalaji, line 14: maybe a nit but are both needed, just reuse SOLANA_RELEASE?

tkporter: this is what they do in the official install script, I think it's probably better to minimize the diff with what they do even if it means there's redundant code

paulbalaji: yeah thought as much, thought i'd ask anyway

aroralanuk (commented):
Can we make an issue for unifying the build and deploy version for this in the future?

PR #6159, opened 2025-05-07
feat(submitter): evm GasLimitEstimator

daniel-savu:
### Description

Implement `GasLimitEstimator` component, by refactoring logic out of `rust/main/chains/hyperlane-ethereum/src/tx.rs`

### Drive-by changes

- The logic for the `TxStatusChecker` and `GasLimitEstimator` is moved into their own modules. The motivation is to use dependency injection for ease of testing in the future
- The logic for building a `ContractCall` from its subcomponent fields is refactored and reused

### Related issues

- Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/6134

### Backward compatibility

Yes

### Testing

Not yet

### DIFF
--- a/rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter/tx_status_checker.rs
+++ b/rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter/tx_status_checker.rs
@@ -1,0 +1,52 @@
+// the evm provider-building logic returns a box. `EvmProviderForSubmitter` is only implemented for the underlying type rather than the boxed type.
+// implementing the trait for the boxed type would require a lot of boilerplate code.
+#![allow(clippy::borrowed_box)]
+
+use ethers::types::U64;
+use hyperlane_ethereum::{EthereumReorgPeriod, EvmProviderForSubmitter};
+use tracing::warn;
+
+use crate::{SubmitterError, TransactionStatus};
+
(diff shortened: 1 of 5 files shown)


rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter/tx_status_checker.rs
```diff
@@ -0,0 +1,52 @@
...
+// implementing the trait for the boxed type would require a lot of boilerplate code.
+#![allow(clippy::borrowed_box)]
+
+use ethers::types::U64;
+use hyperlane_ethereum::{EthereumReorgPeriod, EvmProviderForSubmitter};
+use tracing::warn;
+
+use crate::{SubmitterError, TransactionStatus};
+
+async fn block_number_result_to_tx_hash(
```
ameten, line 11: It looks like this function maps block number of a transaction to transaction status. The name of the function is a little bit confusing.

ameten (approved):

rust/main/submitter/src/chain_tx_adapter/chains/ethereum/adapter/tx_status_checker.rs
daniel-savu: meant that to be `block_number_result_to_tx_status`, good catch

PR #6161, opened 2025-05-07
feat!: Add monitor only mode

fzavalia:
Allows the user to provide a monitorOnly variable on config that will prevent rebalancing from happening, it will only monitor

BREAKING CHANGE: Config schema is different


fernandomg:
@fzavalia why wouldn't you just short-circuit the instantiations in the warp command?

fzavalia:
> @fzavalia why wouldn't you just short-circuit the instantiations in the warp command?

I don't think it is responsibility of the command to do that, you have a system composed by various components, those components should exist regardles (including metrics) and be the factory the one in charge of instantiating the correct one. Even a Stub one that doesn't do anything

fernandomg:
> > @fzavalia why wouldn't you just short-circuit the instantiations in the warp command?
> 
> I don't think it is responsibility of the command to do that, you have a system composed by various components, those components should exist regardles (including metrics) and be the factory the one in charge of instantiating the correct one. Even a Stub one that doesn't do anything

Agree on the vision. Not sure about the need for the approach. Given the other commnds; and that we're already _creating_ the different classes here, we're kind of splitting responsibilities. Right?

typescript/cli/src/commands/warp.ts
```diff
@@ -415,68 +416,84 @@ export const check: CommandModuleWithContext<{
...
+    withMetrics,
+    monitorOnly,
   }) => {
     try {
+      const { registry, key: rebalancerKey } = context;
+
+      // This might happen because despite the key being typed as required, it is not validated beforehand.
+      if (!rebalancerKey) {
+        throw new Error('No key found in context');
+      }
```
fernandomg, line 470: shouldn't we attempt to prompt for the key if the command is not running in a CI context?

fzavalia: Fixed at fbf17e8b784fd8be69d3f9515b2fad625697b575

fzavalia:
> > > @fzavalia why wouldn't you just short-circuit the instantiations in the warp command?
> > 
> > 
> > I don't think it is responsibility of the command to do that, you have a system composed by various components, those components should exist regardles (including metrics) and be the factory the one in charge of instantiating the correct one. Even a Stub one that doesn't do anything
> 
> Agree on the vision. Not sure about the need for the approach. Given the other commnds; and that we're already _creating_ the different classes here, we're kind of splitting responsibilities. Right?

Removed the stub at 69f63cc32b70772f1a2d866bb0f8225b9e6f36da

typescript/cli/src/tests/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -715,8 +734,8 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
     await sleep(3000);
 
     // Check that metrics endpoint is not responding
-    return fetch(DEFAULT_METRICS_SERVER).should.be.rejected.then(() =>
-      process.kill(),
-    );
+    await fetch(DEFAULT_METRICS_SERVER).should.be.rejected;
+
+    await process.kill();
   });
```
fernandomg, line 740: wrong rebase

fernandomg (approved):
a tiny fix in the tests, the rest all great

Issue #5244, opened 2025-01-22
hyperlane warp init writes into the expected path


nambrot:
- file format ${warp-id}-deploy-config.json

mshojaei-txfusion:
So the file would still have `yaml` format right?
[you mentioned "${warp-id}-deploy-config.json" so just wondering if there is any need to change file config or just matter of naming]

mshojaei-txfusion:
@nambrot Please take a look at [hyperlane-xyz/hyperlane-registry#666](https://github.com/hyperlane-xyz/hyperlane-registry/pull/666)
If all good I can use it on monorepo

PR #6171, opened 2025-05-08
feat: update registry to 14.0.0

mshojaei-txfusion:
### Description

Upgraded `@hyperlane-xyz/registry` from v11.1.0 to v14.0.0 across multiple packages and updated warp route config API usage to match the new version requirements.

### Drive-by changes

None

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/5244

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/.changeset/loose-numbers-tease.md
+++ b/.changeset/loose-numbers-tease.md
@@ -1,0 +1,8 @@
+---
+'@hyperlane-xyz/helloworld': minor
+'@hyperlane-xyz/widgets': minor
+'@hyperlane-xyz/infra': minor
+'@hyperlane-xyz/cli': minor
+---
+
+Upgraded @hyperlane-xyz/registry to v14.0.0 and updated warp route config API usage.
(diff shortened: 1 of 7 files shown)


.changeset/loose-numbers-tease.md
```diff
@@ -0,0 +1,8 @@
+---
+'@hyperlane-xyz/helloworld': minor
+'@hyperlane-xyz/widgets': minor
+'@hyperlane-xyz/infra': minor
```
paulbalaji, line 4: nit don't need to add infra but it's ok

paulbalaji (approved):

.changeset/loose-numbers-tease.md
mshojaei-txfusion: Oh got it, would keep in mind for next times

ltyu (approved):

PR #6175, opened 2025-05-08
fix: token metadata handling

christopherbrumm:
### Description

This PR fixes the token metadata handling by changing the return type of `deriveTokenMetadata` to TokenMetadataMap`, enabling support for multiple different token metadata.

### Drive-by changes

`assertWarpRouteConfig` expects the mailbox address to verify the correct deployed mailbox.

### Backward compatibility

Yes

### Testing

Added E2E test to verify that for `collateralFiat` and `collateral` with two different metadata, token is deployed with two different metadata.


typescript/cli/src/deploy/warp.ts
```diff
@@ -397,28 +401,31 @@ async function getWarpCoreConfig(
...
+      params.context.multiProvider,
+      params.warpDeployConfig,
+    );
+
+  for (const [, config] of Object.entries(tokenMetadata)) {
+    assert(
+      tokenMetadata && isTokenMetadata(config),
+      'Missing required token metadata',
+    );
+    assert(config.decimals, 'Missing decimals on token metadata');
```
xeno097, line 415: can you include the chain name in the error message? imagine having multiple errros like this , it would be helpful to know to which chain each one belongs to

typescript/sdk/src/token/deploy.ts
```diff
@@ -123,22 +128,25 @@ abstract class TokenDeployer<
...
+        metadataMap[chain] = TokenMetadataSchema.parse(config);
       } else if (multiProvider.getProtocol(chain) !== ProtocolType.Ethereum) {
         // If the config didn't specify the token metadata, we can only now
         // derive it for Ethereum chains. So here we skip non-Ethereum chains.
+        metadataMap[chain] = undefined;
         continue;
       }
 
       if (isNativeTokenConfig(config)) {
         const nativeToken = multiProvider.getChainMetadata(chain).nativeToken;
```
xeno097, line 144: What about adding the updated `deriveTokenMetadata` method to the EvmErc20WarpReader class instead? Our long-term plan is to stop using the deployer classes, so this sounds like a good opportunity to move this code to the modules

typescript/sdk/src/token/TokenMetadataMap.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (config?.symbol) {
+      return config.symbol;
+    }
+  }
+  return undefined;
+}
+
+export function getName(
+  map: TokenMetadataMap,
+  chain: string,
```
xeno097, line 47: All this makes me think that maybe this code could be part of a class that encapsulates this logic wdyt?

ltyu (commented):
can we keep the business logic of selecting metadata in deriveTokenMetadata()

my reasoning is that deriveTokenMetadata() already does some business logic to figure out the token metadata for each chain, but then we have some additional metadata derivation via the getDecimals(), etc

typescript/sdk/src/token/deploy.ts
ltyu: as discussed, this might not make sense yet given WarpRead only handles a single chain and deriveTokenMetadata is multichain

typescript/sdk/src/token/TokenMetadataMap.ts
```diff
@@ -0,0 +1,59 @@
...
+      );
+    }
+  }
+
+  return first!;
+}
+
+export function getSymbol(
+  map: TokenMetadataMap,
+  chain: string = '',
```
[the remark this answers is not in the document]
christopherbrumm: The default allows using the function without a specific chain, e.g. to get a general symbol in `getWarpConfig`.

typescript/sdk/src/token/TokenMetadataMap.ts
```diff
@@ -0,0 +1,59 @@
+import { TokenMetadata } from './types.js';
+
+export type TokenMetadataMap = Record<string, TokenMetadata | undefined>;
```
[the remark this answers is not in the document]
christopherbrumm: The initial idea was to not change the return type of `deriveTokenMetadata` - but this isn't required anymore. Same as `metadataMap[chain] = undefined;`

typescript/sdk/src/token/deploy.ts
```diff
@@ -123,22 +128,25 @@ abstract class TokenDeployer<
...
+  ): Promise<TokenMetadataMap> {
+    const metadataMap: TokenMetadataMap = {};
     for (const [chain, config] of Object.entries(configMap)) {
       if (isTokenMetadata(config)) {
-        return TokenMetadataSchema.parse(config);
+        metadataMap[chain] = TokenMetadataSchema.parse(config);
       } else if (multiProvider.getProtocol(chain) !== ProtocolType.Ethereum) {
         // If the config didn't specify the token metadata, we can only now
         // derive it for Ethereum chains. So here we skip non-Ethereum chains.
+        metadataMap[chain] = undefined;
```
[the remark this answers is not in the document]
christopherbrumm: Yes you're correct. I assumed that the chain needs to be existent as a key for further usage, but checked and it's not the case. So we can also remove `undefined` from the map.

typescript/cli/src/deploy/warp.ts
```diff
@@ -397,28 +401,31 @@ async function getWarpCoreConfig(
...
-  assert(decimals, 'Missing decimals on token metadata');
+  const tokenMetadata: TokenMetadataMap =
+    await HypERC20Deployer.deriveTokenMetadata(
+      params.context.multiProvider,
+      params.warpDeployConfig,
+    );
+
+  for (const [, config] of Object.entries(tokenMetadata)) {
+    assert(
+      tokenMetadata && isTokenMetadata(config),
```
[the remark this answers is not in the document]
christopherbrumm: Good catch!

typescript/cli/src/config/warp.ts
```diff
@@ -274,12 +274,15 @@ export async function createWarpRouteDeployConfig({
         context.multiProvider,
         warpRouteDeployConfig,
       );
+
+      const symbol = tokenMetadata.getSymbol('');
```
xeno097, line 278: Is this ok, wouldn't it be better to have a method called `getDefaultSymbol` that throws if no default symbol is available in the current metadata map?

typescript/cli/src/deploy/warp.ts
```diff
@@ -372,28 +372,23 @@ async function getWarpCoreConfig(
...
     contracts,
-    symbol,
-    name,
-    decimals,
+    tokenMetadataMap,
   );
 
   fullyConnectTokens(warpCoreConfig);
 
+  const symbol = tokenMetadataMap.getSymbol('');
```
xeno097, line 390: ditto

typescript/cli/src/config/warp.ts
```diff
@@ -274,12 +274,15 @@ export async function createWarpRouteDeployConfig({
...
       assert(
-        tokenMetadata?.symbol,
-        'Error deriving token metadata, please check the provided token addresses',
+        symbol === undefined,
```
xeno097, line 280: i think the assertion here should be `!symbol` because if the assertion is true then no error will be thrown but you want to throw an error if symbol is undefined that is probably why you alsos neded to add the `!` operator below

typescript/sdk/src/token/TokenMetadataMap.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      Object.entries(this.tokenMetadataMap).forEach(([chain, config]) => {
+        if (config.decimals) {
+          const scale = 10 ** (maxDecimals - config.decimals);
+
+          assert(
+            this.tokenMetadataMap[chain].scale &&
+              scale !== this.tokenMetadataMap[chain].scale,
+            `Scale is not correct for ${chain}`,
+          );
```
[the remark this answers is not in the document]
xeno097: this suggestion seems to be correct as assert will throw if the assertion is false

typescript/sdk/src/token/TokenMetadataMap.ts
christopherbrumm: Went with a class now, using the constructor to verify that an instance of `TokenMetadataMap` is always valid.

typescript/cli/src/config/warp.ts
christopherbrumm: Will do that!

xeno097 (approved):
approved but pending fixing the tests

typescript/sdk/src/token/TokenMetadataMap.ts
```diff
@@ -0,0 +1,92 @@
...
+    const [first, ...rest] = values;
+    for (const d of rest) {
+      if (d.decimals !== first.decimals) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  finalize(): void {
```
ltyu, line 67: should this be called `validate()`?

ltyu (approved):

PR #6178, opened 2025-05-08
feat: fixes to SVM warp deploy tooling

tkporter:
### Description

Two things:
1. Updates the commit for our fork of the spl lib that's used to include https://github.com/hyperlane-xyz/solana-program-library/pull/4 which fixes compilation
2. Our tooling generally requires Solana 1.18.18 to be used when running our SVM tooling to perform the program deploy. Eclipse used to allow you to deploy programs just fine with this version but Christopher and I experienced this regression recently where the command just hangs indefinitely. The “workaround” seems to be to downgrade to 1.17.31, but this comes at the cost of not being able to use a CLI flag to set the compute unit price when deploying a program
As a temporary fix, bc the compute unit price is originally set to 0 on Eclipse anyways, I'm changing the tooling to omit this compute unit price flag if it’s set to 0. But it’s still possible for us to hit issues here because we bid up compute unit prices if there are issues deploying the program (not a rare occurrence). If you do a deploy to Eclipse and hit issues, I’d recommend just commenting out the compute unit price flag at least for the eclipse leg

### Testing

- `cargo +1.76.0 install spl-token-cli --git https://github.com/hyperlane-xyz/solana-program-library --branch dan/create-token-for-mint --rev e101cca --force` works
- confirmed not providing the flag works


rust/sealevel/client/src/cmd_utils.rs
```diff
@@ -150,7 +150,9 @@ fn attempt_program_deploy(
     ];
 
     let compute_unit_price_str = compute_unit_price.to_string();
-    command.extend(vec!["--with-compute-unit-price", &compute_unit_price_str]);
+    if compute_unit_price > 0 {
```
daniel-savu, line 153: LGTM, a slightly more reliable fix would be to pass `local_domain` as an arg to `attempt_program_deploy` and not set the flag if the local domain is eclipse

daniel-savu (approved):

kamiyaa (approved):

rust/sealevel/client/src/cmd_utils.rs
tkporter: ah good idea, i'll just prevent bidding up on eclipse for now

PR #6182, opened 2025-05-08
fix: async test

fernandomg:
### Description

fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6161#discussion_r2080253589


fernandomg:
same result, no need to change

PR #6181, opened 2025-05-08
refactor: remove all references to gcp

fernandomg:
### Description

- Removes the monitor references to GCP
- Allows to set Coingecko API KEY via:
	- command param `coingecko-api-key` (or its aliases, `g`, and `coingeckoApiKey`) 
	- env var `COINGECKO_API_KEY`

### Related issues

Fixes https://github.com/BootNodeDev/hyperlane-monorepo/issues/49

### Backward compatibility

Yes

### Testing

Manual

PR #6186, opened 2025-05-09
fix: Fix config unit tests

fzavalia:


fernandomg:
@fzavalia I think this is already solved here: [`2586f6d` (#6181)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6181/commits/2586f6d9b542a70e608ff1fc990f5b2048195473)

fzavalia:
> @fzavalia I think this is already solved here: [`2586f6d` (#6181)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6181/commits/2586f6d9b542a70e608ff1fc990f5b2048195473)

Right, I had an older version.

Closing 🚪

PR #6201, opened 2025-05-12
fix: Retry on getting Relayer balance in E2E since Relayer may not be started yet

ameten:
### Description

Retry on getting Relayer balance in E2E since Relayer may not be started yet

### Backward compatibility

Yes

### Testing

Ethereum E2E locally

### DIFF
--- a/rust/main/hyperlane-base/src/settings/loader/mod.rs
+++ b/rust/main/hyperlane-base/src/settings/loader/mod.rs
@@ -2,15 +2,15 @@
 
 use std::{env, error::Error, fmt::Debug, path::PathBuf};
 
+use crate::settings::loader::{
+    arguments::CommandLineArguments, case_adapter::CaseAdapter, environment::Environment,
+};
 use config::{Config, File};
 use convert_case::Case;
 use eyre::{eyre, Context, Result};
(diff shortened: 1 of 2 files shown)


rust/main/hyperlane-base/src/settings/loader/mod.rs
```diff
@@ -22,6 +22,9 @@ where
     T: DeserializeOwned + Debug,
     R: FromRawConf<T>,
 {
+    let now = chrono::Utc::now();
+    println!("Loading settings: {:?}", now);
```
ameten, line 26: Tracing is not initialised yet.

rust/main/hyperlane-base/src/settings/loader/mod.rs
```diff
@@ -128,5 +131,9 @@ where
     if res.is_err() {
         eprintln!("Loaded config for debugging: {formatted_config}");
     }
+
+    let now = chrono::Utc::now();
+    println!("Loaded settings: {:?}", now);
```
daniel-savu, line 136: could use `Instant` + `.elapsed()` instead here but it's ok

daniel-savu (approved):

PR #6202, opened 2025-05-12
feat(rework): cli warp check should assert that contracts are verified

ltyu:
### Description
- Reworked ContractVerifer.getContractVerificationStatus() to output a status instead of boolean. This allows more granular statuses for statuses such as `Skipped`.
- Update derive contract status to no longer include the actual token contract name and to instead just say `Implementation`. This reduces code complexity and can be featured in later, if needed.
- Add `HypTokenRouterVirtualConfigSchema` to `HypTokenRouterConfigMailboxOptionalSchema`

### Testing
Manual

### DIFF
--- a/typescript/sdk/src/deploy/verify/ContractVerifier.ts
+++ b/typescript/sdk/src/deploy/verify/ContractVerifier.ts
@@ -365,7 +365,9 @@ export class ContractVerifier {
         ? ContractVerificationStatus.Verified
         : ContractVerificationStatus.Unverified;
     } catch (e) {
-      this.logger.info(`Error fetching contract verification status: ${e}`);
+      this.logger.info(
+        `Error fetching contract verification status for ${address} on chain ${chain}: ${e}`,
+      );
       return ContractVerificationStatus.Error;
     }
   }
--- a/typescript/sdk/src/deploy/verify/ContractVerifier.ts
+++ b/typescript/sdk/src/deploy/verify/ContractVerifier.ts
@@ -340,6 +340,12 @@ export class ContractVerifier {
     address: Address,
     verificationLogger: Logger,
   ): Promise<{ isVerified: false } | { isVerified: true; name: string }> {
+    const metadata = this.multiProvider.tryGetChainMetadata(chain);
+    const rpcUrl = metadata?.rpcUrls[0].http ?? '';
+    if (rpcUrl.includes('localhost') || rpcUrl.includes('127.0.0.1')) {
+      verificationLogger.debug('Skipping verification for local endpoints');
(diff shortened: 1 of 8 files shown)


typescript/sdk/src/deploy/verify/ContractVerifier.ts
```diff
@@ -338,29 +339,35 @@ export class ContractVerifier {
...
+          { address },
+        )
+      )[0]; // This specific query only returns 1 result
+
+      // Explorer won't return ContractName if unverified
+      return sourceCodeResults.ContractName
+        ? ContractVerificationStatus.Verified
+        : ContractVerificationStatus.Unverified;
+    } catch (e) {
+      this.logger.info(`Error fetching contract verification status: ${e}`);
```
xeno097, line 368: Can you add also the chain nameand the contract address in the log? it might be useful for debugging

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -140,43 +140,34 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
...
+    const virtualConfig: HypTokenRouterVirtualConfig = {
+      contractVerificationStatus: {},
+    };
 
-      if (verificationStatus.isVerified)
-        virtualConfig.contractVerificationStatus[verificationStatus.name] =
-          true;
+    const contractType = (await isProxy(this.provider, address))
+      ? 'Proxy'
+      : 'Implementation';
```
xeno097, line 150: Do we have an enum/ constant variable with the possible values for each contract type? if not i think it would be a good addition to avoid magic strings

typescript/sdk/src/token/configUtils.ts
```diff
@@ -148,17 +145,19 @@ export async function expandWarpDeployConfig(params: {
...
-      };
+    // Expand warpDeployConfig virtual to the control states
+    if (expandedOnChainWarpConfig?.[chain].contractVerificationStatus) {
+      // For most cases, we set to Verified
+      chainConfig.contractVerificationStatus = objMap(
+        expandedOnChainWarpConfig[chain].contractVerificationStatus ?? {},
+        (_, status) => {
+          // Skipped for local e2e testing
+          if (status === ContractVerificationStatus.Skipped)
+            return ContractVerificationStatus.Skipped;
```
xeno097, line 156: Can we add a condition here to skip if the protocol type is not a supported one like SVM etc?

ltyu: yup

PR #6205, opened 2025-05-12
fix: Only rebalance EvmHypCollateralAdapter tokens

fzavalia:
Previous implementation was allowing tokens that could not be rebalanced

### DIFF
--- a/typescript/cli/src/rebalancer/transformers/MonitorToStrategyTransformer.ts
+++ b/typescript/cli/src/rebalancer/transformers/MonitorToStrategyTransformer.ts
@@ -1,0 +1,30 @@
+import { EvmHypCollateralAdapter, WarpCore } from '@hyperlane-xyz/sdk';
+
+import { MonitorEvent } from '../interfaces/IMonitor.js';
+import { RawBalances } from '../interfaces/IStrategy.js';
+import { ITransformer } from '../interfaces/ITransformer.js';
+
+export class MonitorToStrategyTransformer
+  implements ITransformer<MonitorEvent, RawBalances>
+{
+  constructor(private readonly warpCore: WarpCore) {}
+
+  transform(event: MonitorEvent): RawBalances {
+    return event.tokensInfo.reduce((acc, tokenInfo) => {
+      const provider = this.warpCore.multiProvider;
+      const adapter = tokenInfo.token.getHypAdapter(provider);
+
+      if (adapter instanceof EvmHypCollateralAdapter) {
+        if (tokenInfo.bridgedSupply === undefined) {
+          throw new Error(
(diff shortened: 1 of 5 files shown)


typescript/cli/src/rebalancer/transformers/MonitorToStrategyTransformer.ts
```diff
@@ -0,0 +1,30 @@
...
+    return event.tokensInfo.reduce((acc, tokenInfo) => {
+      const provider = this.warpCore.multiProvider;
+      const adapter = tokenInfo.token.getHypAdapter(provider);
+
+      if (adapter instanceof EvmHypCollateralAdapter) {
+        if (tokenInfo.bridgedSupply === undefined) {
+          throw new Error(
+            `bridgedSupply should not be undefined for collateralized token ${tokenInfo.token.addressOrDenom}`,
+          );
+        }
```
fernandomg, line 22: I wouldn't throw here but log and return the `acc` as we were doing until now, because you'll be breaking the backwards compatiblity with the metrics module.

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6205/files#diff-98d1915a658b2f2e88c0354a1426498882fc624e7e67ea561e20bcb582c8e6dcL525

It should be able to monitor all the `hyperlane tokens`.

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/fix/monitor-strategy-transformer/typescript/infra/scripts/warp-routes/monitor/monitor-warp-route-balances.ts#L258-L261

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/fix/monitor-strategy-transformer/typescript/cli/src/rebalancer/monitor/Monitor.ts#L100-L106

fernandomg (changes requested):

typescript/cli/src/rebalancer/transformers/MonitorToStrategyTransformer.ts
fzavalia:
```ts
await monitor
  // Observe balances events and process rebalancing routes
  .on('tokeninfo', (event) => {
    if (metrics) {
      for (const tokenInfo of event.tokensInfo) {
        metrics.processToken(tokenInfo).catch((e) => {
          errorRed(
            `Error building metrics for ${tokenInfo.token.addressOrDenom}: ${e.message}`,
          );
        });
      }
    }

    const rawBalances = monitorToStrategy.transform(event);

    const rebalancingRoutes = strategy.getRebalancingRoutes(rawBalances);

    ....
```

The transform is called after the metrics, so it is still logging everything.

It is dangerous to allow a partial rebalance in a warp route if a collateral is not considered that is why it is important to fail.

fernandomg (approved):

PR #6206, opened 2025-05-12
feat(strategy): MinAmount strategy

fernandomg:
### Description

WIP:
Implement strategy for a minimum amount

also, created a BaseStrategy to extend from, for future strategies
finally, created a Factory to create each type of strategy based on config

### Related issues

Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/47


typescript/cli/src/rebalancer/strategy/MinAmountStrategy.ts
```diff
@@ -0,0 +1,68 @@
...
+        const minAmount = this.config[chain].minAmount;
+        const balance = rawBalances[chain];
+
+        // If balance is less than minAmount, it has a deficit
+        if (balance < minAmount) {
+          acc.deficits.push({ chain, amount: minAmount - balance });
+        } else {
+          // Any chain with more than minAmount potentially has surplus
+          // But only mark as surplus if there's extra beyond minAmount
+          const surplus = balance - minAmount;
```
fzavalia, line 54: Would be nice to have a buffer beyond the min amount to prevent consistent rebalances. For example, if the min amount is 10, and the buffer is 10%, fill the chain up to 11, this leaves a margin that prevents having to rebalance if there is a transfer

fzavalia (commented):
🎸

typescript/cli/src/rebalancer/config/Config.test.ts
```diff
@@ -20,11 +20,13 @@ describe('Config', () => {
       checkFrequency: 1000,
       coingeckoApiKey: COINGECKO_API_KEY,
       chain1: {
+        strategyType: 'weighted',
```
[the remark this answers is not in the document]
fernandomg: yeah... 🤦

typescript/cli/src/rebalancer/strategy/MinAmountStrategy.ts
```diff
@@ -0,0 +1,76 @@
...
+  } {
+    return this.chains.reduce(
+      (acc, chain) => {
+        const minAmount = this.config[chain].minAmount;
+        const effectiveMin =
+          (minAmount * (10_000n + this.config[chain].buffer)) / 10_000n;
+        const balance = rawBalances[chain];
+
+        // If balance is less than effectiveMin, it has a deficit
+        if (balance < effectiveMin) {
```
fzavalia, line 57: Should this be `balance < minAmount`?

fzavalia: The effectiveMin already has the buffer, minAmount indicates the actual minimum

fernandomg: Depends, I treated the min with the buffer applied as the "new min", so if you're below it, you should consider increasing.

What you're proposing is "if I go below a min amount, I should go up to min amount + buffer"?

fzavalia: Right! Because if the effectiveMin is the threshold, doesn't it defeat the purpose of the minValue?

fzavalia: Going "up to min amount + buffer" is what prevents the collateral being unbalanced just from losing 1 wei

fernandomg: I think it makes more sense now that I re-read this

fzavalia: Imagine the min is 100, I'm at 90, go back to 100 from rebalancing, any trade, no matter how small will just take me back to red

fzavalia (approved):
👨‍🎤

PR #6209, opened 2025-05-13
feat: may 12 mainnets

paulbalaji:
### Description

feat: may 12 mainnets
- kyve
- miracle
- ontology

### Drive-by changes

- fix bouncebit index.chunk after moving to private RPC

### Related issues

https://github.com/hyperlane-xyz/hyperlane-registry/pull/848


rust/main/config/mainnet_config.json
```diff
@@ -8262,7 +8262,7 @@
       "timelockController": "0x0000000000000000000000000000000000000000",
       "validatorAnnounce": "0xB2b0A80b2fa3fC9aB1564A4FaF013d4D6084B325",
       "index": {
-        "chunk": 999,
+        "chunk": 100,
```
Mo-Hussain, line 8265: The private RPC is performs worse at indexing?

Mo-Hussain (approved):

rust/main/config/mainnet_config.json
paulbalaji: "performs worse" isn't really the right term here, just the RPC doesn't allow the larger block ranges

and at this point i'd take a private endpoint with smaller ranges than a public endpoint that is unstable/doesn't work

PR #6211, opened 2025-05-13
fix(ICA): Remove ica from commitment 

larryob:
### Description

The commitment is simply `hash(salt, calls)`. 

### Drive by
- Remove redundant checks on revealed data
- Move check on existing commitment into ICA.

### DIFF
--- a/solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
+++ b/solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
@@ -62,6 +62,14 @@ contract CommitmentReadIsm is AbstractCcipReadIsm, Ownable {
         bytes calldata _metadata,
         bytes calldata _message
     ) external returns (bool) {
+        // This is hash(salt, calls). The ica address is excluded
+        bytes32 revealedHash = keccak256(_metadata[20:]);
+        bytes32 msgCommitment = _message.body().commitment();
+        require(
+            revealedHash == msgCommitment,
+            "Commitment ISM: Revealed Hash Invalid"
+        );
+
         // Fetch encoded ica, salt, and calls
         address _ica = address(bytes20(_metadata[:20]));
         OwnableMulticall ica = OwnableMulticall(payable(_ica));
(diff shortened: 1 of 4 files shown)


solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
```diff
@@ -54,30 +54,34 @@ contract CommitmentReadIsm is AbstractCcipReadIsm, Ownable {
...
+        address _ica = address(bytes20(_metadata[:20]));
+        OwnableMulticall ica = OwnableMulticall(payable(_ica));
 
-        bytes32 revealedHash = keccak256(_metadata);
+        bytes32 salt = bytes32(_metadata[20:52]);
+
+        CallLib.Call[] memory calls = abi.decode(
+            _metadata[52:],
+            (CallLib.Call[])
+        );
```
yorhodes, line 74: you can move the decoding to below the commitment check

yorhodes: should we use a library similar to InterchainAccountMessage for this?

yorhodes: dont love these "magic constants"

yorhodes: maybe we can add a lint rule for this

solidity/contracts/middleware/libs/OwnableMulticall.sol
```diff
@@ -42,11 +46,11 @@ contract OwnableMulticall {
...
     ) external payable returns (bytes32 executedCommitment) {
-        // If there is no active commitment, do nothing.
-        if (commitment == bytes32(0)) return bytes32(0);
-
-        bytes32 revealedHash = keccak256(abi.encode(calls, salt, this));
-        require(commitment == revealedHash, "Invalid Reveal");
+        // Check if metadata matches stored commitment (checks)
+        bytes32 revealedHash = keccak256(
+            abi.encodePacked(salt, abi.encode(calls))
+        );
```
yorhodes, line 52: this encoding/hash function could also live on the suggested library

yorhodes (approved):

solidity/contracts/isms/ccip-read/CommitmentReadIsm.sol
larryob: A library sounds good, I'll follow up

PR #6212, opened 2025-05-13
chore: add logs to the rebalancer flow

fernandomg:
### Description

Add logs in different parts of the rebalancing flow
Also prevents from logging in the middle of the `test:ci` tests if the user sets `NODE_ENV=test`

### Related issues

Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/51

### DIFF
--- a/typescript/cli/src/tests/test-setup.ts
+++ b/typescript/cli/src/tests/test-setup.ts
@@ -1,0 +1,5 @@
+import { LogFormat, LogLevel, configureRootLogger } from '@hyperlane-xyz/utils';
+
+// mute logging in tests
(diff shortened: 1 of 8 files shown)


typescript/cli/src/tests/test-setup.ts
```diff
@@ -0,0 +1,5 @@
+import { LogFormat, LogLevel, configureRootLogger } from '@hyperlane-xyz/utils';
+
+// mute logging in tests
+process.env.NODE_ENV === 'test' &&
```
fzavalia, line 4: ❤️

fzavalia (approved):
👍

PR #6213, opened 2025-05-13
fix: temporarily log error instead of refusing to build meta

daniel-savu:

PR #6216, opened 2025-05-14
fix: Ignore reorg flag temporarily

nambrot:
### Description

@daniel-savu's previous PR at https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6213 still seems to result in lack of delivery. I can't quite explain why, but somehow both multisig metadata builders get the error which makes the aggregation builder unable to build metadata. This PR disables reorg checking on the s3 checkpoint syncer completely so that the relayer effectively ignores the flag


yorhodes (dismissed):

daniel-savu:
issue got resolved in prod, closing this

PR #6217, opened 2025-05-14
DRAFT: Just needed for the Github Action

mbreithecker:


typescript/ccip-server/src/server.ts
```diff
@@ -31,10 +31,15 @@ for (const name of getEnabledModules()) {
   console.log(`✅  Mounted '${name}' at '/${name}'`);
 }
 
+app.get('/health', (req, res) => {
```
Mo-Hussain, line 34: I'd suggest we define a class that sets up the server and runs it with a start method, we can then import this into a script in infra and make that the entry point. It would give us the ability to inject dependencies into the server and leverage our infra tooling

Mo-Hussain: We could also add another script in ccip-server to use this new class so that we can run the server in a self contained manner

Mo-Hussain:
Great works so far!

Can we define a helm manager class that extends `HelmManager` and an infra script that we can use to deploy this workload? 

Here is the `HelmManager` https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/00e3d096a30957a423425e7d8a156be2ffaff5da/typescript/infra/src/utils/helm.ts#L120

Here is an example https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/00e3d096a30957a423425e7d8a156be2ffaff5da/typescript/infra/src/warp/helm.ts#L33, https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/00e3d096a30957a423425e7d8a156be2ffaff5da/typescript/infra/scripts/warp-routes/deploy-warp-monitor.ts#L43

Notice how we can you the same pipeline and templates for testnet and mainnet instances in these examples and we can configure the deployment by passing arguments to the deployment script, we could configure things like the ENABLED_MODULES etc

PR #6218, opened 2025-05-14
feat: update svm gas oracles

paulbalaji:
### Description

feat: update svm gas oracles

### Drive-by changes

interconnect svmbnb<>soon<>solana<>bsc

### DIFF
--- a/typescript/infra/scripts/sealevel-helpers/print-gas-oracles.ts
+++ b/typescript/infra/scripts/sealevel-helpers/print-gas-oracles.ts
@@ -103,10 +103,13 @@ function getChainConnections(
       ['solanamainnet', 'everclear'],
       ['solanamainnet', 'infinityvmmainnet'],
       ['solanamainnet', 'sophon'],
-      // for svmBNB
+      // for svmBNB routes solana<>bsc<>svmbnb<>soon
+      ['solanamainnet', 'bsc'],
       ['svmbnb', 'solanamainnet'],
       ['svmbnb', 'bsc'],
       ['svmbnb', 'soon'],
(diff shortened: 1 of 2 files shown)


typescript/infra/scripts/sealevel-helpers/print-gas-oracles.ts
```diff
@@ -103,10 +103,13 @@ function getChainConnections(
       ['solanamainnet', 'everclear'],
       ['solanamainnet', 'infinityvmmainnet'],
       ['solanamainnet', 'sophon'],
-      // for svmBNB
+      // for svmBNB routes solana<>bsc<>svmbnb<>soon
```
daniel-savu, line 106: the comments above suggest these entries are unidirectional links (`Returns a record of origin chain -> set of chains that are connected via warp routes.`) rather than bidirectional. Which makes me wonder if we should also add `['solanamainnet', 'svmbnb'],`

paulbalaji: iiuc these are intended to be defined once and we'll add the bidirectional connectivity as part of a reduction afterwards https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/e82b98b583c0d6985ad850d021adf03059864847/typescript/infra/scripts/sealevel-helpers/print-gas-oracles.ts#L129-L147

daniel-savu (approved):
can confirm running `yarn tsx ./scripts/sealevel-helpers/print-gas-oracles.ts -e mainnet3` locally have me the same output as in `rust/sealevel/environments/mainnet3/gas-oracle-configs.json`

PR #6221, opened 2025-05-14
feat(ICA): Allow multiple commitments for a given ica

larryob:
### Description
You can have multiple pending commitments for a given ica. The commitments are completely independent of each other.


yorhodes (approved):
that was nice and simple

PR #6223, opened 2025-05-14
fix: warp route id lookup source

mshojaei-txfusion:
### Description

Fixed warp route ID lookup to use deployment configs instead of warp routes when prompting users for selection on warp deployment.

### Drive-by changes

None

### Related issues

N/A

### Backward compatibility

Yes - This change only affects the user selection prompt behavior and doesn't impact infrastructure compatibility.

### Testing

Manual testing of the warp route ID selection workflow


Xaroz (approved):
LGTM

PR #6224, opened 2025-05-14
feat: Decode internal calls in forge tests

larryob:
### Description

By default we decode internal calls in our foundry tests. This is great for debugging! See these [Foundry v1.0 release notes](https://www.paradigm.xyz/2025/02/announcing-foundry-v1-0) for more.


yorhodes (approved):
nice

PR #6240, opened 2025-05-15
feat: offchain lookup ISM deployment tooling

yorhodes:
### Description

Support configuration and deployment of new offchain lookup token bridges

### Drive-by changes

Merge ICA router config

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1574/create-deployment-tooling-for-ica-router-ism-with-service-url-as-input
- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1575/create-deployment-tooling-for-tokenbridgecctp-with-service-url-as

### Backward compatibility

No

### Testing

Unit tests


typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -573,6 +593,11 @@ export class EvmERC20WarpModule extends HyperlaneModule<
       contractVerifier,
     );
 
+    // skip router table for CCTP
+    if (isCctpTokenConfig(config)) {
+      return warpModule;
+    }
```
yorhodes, line 599: this is wrong, revert

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -21,10 +21,50 @@ import {PackageVersioned} from "../../PackageVersioned.sol";
  *    message. This functions selector should be provided as the `callbackFunction` payload
  *    for the OffchainLookup error
  */
-abstract contract AbstractCcipReadIsm is ICcipReadIsm, PackageVersioned {
+abstract contract AbstractCcipReadIsm is ICcipReadIsm, MailboxClient {
```
nambrot, line 24: Why is it a MailboxClient? Just for the Mailbox/process part? I dont think we are actually using it, so imo i'd rather just remove it

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -21,10 +21,50 @@ import {PackageVersioned} from "../../PackageVersioned.sol";
...
         uint8(IInterchainSecurityModule.Types.CCIP_READ);
+
+    string[] internal _urls;
+
+    /**
+     * @notice Emitted when new CCIP-read urls are being set
+     */
+    event UrlsChanged(string[] newUrls);
+
+    function setUrls(string[] memory __urls) public onlyOwner {
```
nambrot, line 38: If we do generalize this, I wonder if we should make urls append-only. Imo that is a better story to avoid owners being able to rug by removing urls?

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -21,10 +21,50 @@ import {PackageVersioned} from "../../PackageVersioned.sol";
...
+        mailbox.process(_metadata, _message);
+    }
+
+    function getOffchainVerifyInfo(
+        bytes calldata _message
+    ) external view override {
+        revert OffchainLookup({
+            sender: address(this),
+            urls: _urls,
+            callData: _offchainLookupCalldata(_message),
```
nambrot, line 57: I like this

solidity/contracts/test/TestCcipReadIsm.sol
```diff
@@ -8,14 +8,10 @@ import {ICcipReadIsm} from "../interfaces/isms/ICcipReadIsm.sol";
...
 import {IOptimismPortal} from "../interfaces/optimism/IOptimismPortal.sol";
 import {IInterchainSecurityModule, ISpecifiesInterchainSecurityModule} from "../interfaces/IInterchainSecurityModule.sol";
+import {MailboxClient} from "../client/MailboxClient.sol";
 
-contract TestCcipReadIsm is
-    AbstractCcipReadIsm,
-    IMessageRecipient,
-    ISpecifiesInterchainSecurityModule
-{
+contract TestCcipReadIsm is AbstractCcipReadIsm, IMessageRecipient {
```
nambrot, line 13: Looks like we have two TestCcipReadIsms can we consolidate? Didn't realize it

solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -63,6 +65,20 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
...
+        bytes32 _recipient,
+        uint256 _amount
+    ) external view virtual override returns (Quote[] memory quotes) {
+        quotes = new Quote[](2);
+        quotes[0] = _quoteTransferRemote(
+            _destinationDomain,
+            _recipient,
+            _amount
+        )[0];
+        quotes[1] = Quote({token: address(wrappedToken), amount: _amount});
```
nambrot, line 79: Is this intentional? I guess this is another "nam didn't specify requirements", but imo the quote should only be for any other amounts other than the actual tokens that are being transferred

solidity/contracts/token/extensions/TokenBridgeCctpV2.sol
```diff
@@ -45,15 +46,15 @@ contract TokenBridgeCctpV2 is TokenBridgeCctp {
 
     function _cctpDepositForBurn(
         uint32 _destination,
+        bytes32 _recipient,
```
nambrot, line 49: Seems like there is a few different changes in this PR like this one that would be nice to separate PRs if possible

typescript/sdk/src/core/EvmIcaModule.ts
```diff
@@ -67,137 +62,12 @@ export class EvmIcaModule extends HyperlaneModule<
   ): Promise<AnnotatedEV5Transaction[]> {
     const actualConfig = await this.read();
 
-    const transactions: AnnotatedEV5Transaction[] = [
```
nambrot, line 70: Is this intentional to remove all the router enrollment?

typescript/sdk/src/gas/token-prices.test.ts
```diff
@@ -1,8 +1,7 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
 
-import { ethereum, solanamainnet } from '@hyperlane-xyz/registry';
-
+// import { ethereum, solanamainnet } from '@hyperlane-xyz/registry';
```
nambrot, line 4: Also intentional?

typescript/sdk/src/middleware/account/InterchainAccountDeployer.ts
```diff
@@ -47,7 +45,13 @@ export class InterchainAccountDeployer extends HyperlaneRouterDeployer<
     const interchainAccountRouter = await this.deployContract(
       chain,
       'interchainAccountRouter',
-      [config.mailbox, ethers.constants.AddressZero, owner, 50_000],
+      [
```
nambrot, line 48: I'm a bit worried about our rollout of these ICAs. Imo we are changing the bytecode and deployment tooling to the new ICAs but we would need to plan the migration to them, otherwise we are "bricking" accessing the current ICAs, no?

PR #6241, opened 2025-05-16
chore: add cosmos core module & reader

troykessler:
### Description

This PR implements and adds the Cosmos Native Core Reader and Module.

Note that this is the fourth PR of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6050 which is being split up in order to make it easier to review

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Tests

### DIFF
--- a/.changeset/better-carrots-tell.md
+++ b/.changeset/better-carrots-tell.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': minor
+---
+
+add cosmos native core module & reader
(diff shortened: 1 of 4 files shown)


xeno097 (approved):
There will be a follow up PR with e2e/unit tests?

troykessler:
> There will be a follow up PR with e2e/unit tests?

yes this was the plan, because including them here would be too big for one PR

PR #6244, opened 2025-05-16
chore: add routing ism to cosmos-types and sdk

troykessler:
### Description

This PR adds the newly added Routing ISM of the Hyperlane Cosmos Module released in [v1.0.0](https://github.com/bcp-innovations/hyperlane-cosmos/releases/tag/v1.0.0) to the cosmos-types and cosmos-sdk package. Furthermore, e2e tests have been added to test the newly added transactions.

Note that the majority of code changes are from the autogenerated code from the cosmos-types package. Only the changes in the cosmos-sdk package should be reviewed.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

E2E tests


typescript/sdk/src/hook/CosmosNativeHookModule.ts
```diff
@@ -188,6 +188,7 @@ export class CosmosNativeHookModule extends HyperlaneModule<
       await this.signer.setIgpOwner({
         igp_id: igp.id,
         new_owner: config.owner,
+        renounce_ownership: !config.owner, // if owner is empty we renounce the ownership
```
xeno097, line 191: I feel this is incorrect here because we are going to execute this code block only if the `config.owner` field is not a falsy value `[ if(config.owner && this.signer.account.address !== config.owner) ]`  but here we are checking if it is empty, but empty in javascript is falsy so this block will never execute in the first place.

troykessler: you're right here, removed the check in the if statement

xeno097 (approved):

paulbalaji (approved):

typescript/sdk/src/hook/CosmosNativeHookModule.ts
```diff
@@ -184,10 +184,11 @@ export class CosmosNativeHookModule extends HyperlaneModule<
       });
     }
 
-    if (config.owner && this.signer.account.address !== config.owner) {
+    if (this.signer.account.address !== config.owner) {
```
paulbalaji, line 187: should this be !eqAddress instead of just a !== comparison? ditto generally when we're comparing addresses

typescript/sdk/src/hook/CosmosNativeHookModule.ts
```diff
@@ -184,7 +185,7 @@ export class CosmosNativeHookModule extends HyperlaneModule<
       });
     }
 
-    if (this.signer.account.address !== config.owner) {
+    if (eqAddress(this.signer.account.address, config.owner)) {
```
paulbalaji, line 188: !eqAddress

troykessler: good catch!

PR #6247, opened 2025-05-16
feat: bridge route-based overrides

fernandomg:
### Description

Adds an `override` key to the chain config where bridge-specific props can be overriden based on the chain.

So, given the following scenario:
```ts
chain1: {
  bridge: '0x123',
  overrides: {
	chain2: {
      bridge: '0x222', 
    }
  }
}
````

When trying to bridge from `chain1` to `chain2`, brdige `0x222` will be used, while in the other cases will be `0x123`

### Related issues

Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/61

### Backward compatibility

Yes

### Testing

Manual/Unit Tests

### DIFF
--- a/typescript/cli/src/rebalancer/factories/RebalancerContextFactory.ts
+++ b/typescript/cli/src/rebalancer/factories/RebalancerContextFactory.ts
@@ -96,8 +96,8 @@ export class RebalancerContextFactory {
     const executor = new Executor(
       objMap(this.config.chains, (_, v) => ({
         bridge: v.bridge,
-        minAcceptedAmount: v.bridgeMinAcceptedAmount ?? 0n,
-        isWarp: v.bridgeIsWarp ?? false,
+        bridgeMinAcceptedAmount: v.bridgeMinAcceptedAmount ?? 0n,
+        bridgeIsWarp: v.bridgeIsWarp ?? false,
       })),
       this.config.rebalancerKey,
       this.warpCore,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: typescript/cli/src/rebalancer/utils/bridgeConfig.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+  toChain: ChainName,
+): BridgeConfig {
+  const fromConfig = bridges[fromChain];
+  const routeSpecificOverrides = fromConfig.overrides?.[toChain];
+
+  // Create a new object with the properties from bridgeConfig, excluding the overrides property
+  const { overrides: _, ...baseConfig } = fromConfig;
+
+  // Return a new object with the base config and any overrides
+  return { ...baseConfig, ...routeSpecificOverrides };
```
fzavalia, line 29: What happens if routeSpecificOverrides is undefined?

typescript/cli/src/rebalancer/factories/RebalancerContextFactory.ts
```diff
@@ -96,8 +96,8 @@ export class RebalancerContextFactory {
     const executor = new Executor(
       objMap(this.config.chains, (_, v) => ({
```
fzavalia, line 97: This doesn't seem to be receiving the overrides.

typescript/cli/src/rebalancer/utils/bridgeConfig.ts
fernandomg: nothing bad: https://stackoverflow.com/a/47156824

fzavalia (approved):
LGM

PR #6253, opened 2025-05-16
Cctp deploy

yorhodes:


typescript/sdk/src/deploy/HyperlaneDeployer.ts
```diff
@@ -603,6 +611,7 @@ export abstract class HyperlaneDeployer<
       implementation,
       proxyAdmin,
       initializeArgs,
+      this.initializeFnSignature(contractName ?? ''),
```
nambrot, line 614: If there is no contractName, does this even make sense?

nambrot: Shouldn't we warn or possibly throw an error?

typescript/sdk/src/token/deploy.ts
```diff
@@ -124,6 +133,8 @@ abstract class TokenDeployer<
       isNativeTokenConfig(config)
     ) {
       return defaultArgs;
+    } else if (isCctpTokenConfig(config)) {
```
nambrot, line 136: This is not you and probably also not for this PR, but I wonder if we should have a different abstraction that encapsulates these instead of these if/else clauses

nambrot (approved):

PR #6254, opened 2025-05-16
chore: update print-latest-checkpoints to support all chains

paulbalaji:
### Description

chore: update print-latest-checkpoints to support all chains

### Testing

tested with forma, which is not in the core supported chains list but we can read checkpoints for now!

image


ameten (approved):

daniel-savu (approved):
🙏

PR #6256, opened 2025-05-17
chore: default to only amd64 docker builds

paulbalaji:
### Description

chore: default to only amd64 docker builds
- default agent/monorepo images are only for amd64
- this saves time, cost of each image build
- if required for local runs, devs can manually trigger the workflow and say they want an arm64 build
- allowing manual runs also means that you don't need to have a PR open just to trigger image builds anymore


nambrot (approved):
Can you maybe indicate in the readme how to exactly trigger the arm64 build

PR #6262, opened 2025-05-19
Store commitments in DB

nambrot:
### Description

This PR stores the commitments in a SQL database instead of just a local map

### Backward compatibility

Yes

### Testing

Manual


typescript/ccip-server/src/services/CallCommitmentsService.ts
```diff
@@ -70,49 +61,109 @@ export class CallCommitmentsService extends BaseService {
...
+   */
+  private async insertCommitmentToDB(
+    commitment: string,
+    data: z.infer<typeof postCallsSchema>,
+  ) {
+    const { calls, relayers, salt, ica, commitmentDispatchTx, originDomain } =
+      data;
+    await prisma.commitment.create({
+      data: {
+        id: commitment,
```
mbreithecker, line 124: Is it guaranteed that nobody can overwrite the entry from somebody else?

nambrot: You would need to know the calls and salt, but we should probably add something to the key

mbreithecker: We already have a ToDo for this in line 37, so we can deal with it later 👍

PR #6263, opened 2025-05-19
chore: add cosmos native core cli commands

troykessler:
### Description

This PR finally adds support for Cosmos Native Core Init, Read, Deploy & Apply.

Note that this is the fifth and last PR of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6050 which is being split up in order to make it easier to review

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Tests

### DIFF
--- a/typescript/cli/src/config/hooks.ts
+++ b/typescript/cli/src/config/hooks.ts
@@ -309,7 +309,7 @@ async function getIgpTokenPrices(
       chainMetadata: filteredMetadata,
     });
     const results = await tokenPriceGetter.getAllTokenPrices();
-    fetchedPrices = objMap(results, (v) => v.toString());
+    fetchedPrices = objMap(results, (v) => results[v].toString());
   }
 
   logBlue(
--- a/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
+++ b/typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
@@ -148,7 +148,10 @@ export class MultiChainResolver implements ChainResolver {
 
     // If no destination is specified, return all EVM chains
     if (!argv.destination) {
-      return Array.from(this.getEvmChains(multiProvider));
+      return [
+        ...this.getEvmChains(multiProvider),
+        ...this.getCosmosNativeChains(multiProvider),
+      ];
     }
 
     chains.add(argv.destination);
@@ -195,6 +198,14 @@ export class MultiChainResolver implements ChainResolver {
 
     return chains.filter(
       (chain) => multiProvider.getProtocol(chain) === ProtocolType.Ethereum,
+    );
+  }
+
+  private getCosmosNativeChains(multiProvider: MultiProvider): ChainName[] {
+    const chains = multiProvider.getKnownChainNames();
+
+    return chains.filter(
+      (chain) => multiProvider.getProtocol(chain) === ProtocolType.CosmosNative,
     );
   }
 
--- a/typescript/cli/src/deploy/core.ts
+++ b/typescript/cli/src/deploy/core.ts
@@ -1,18 +1,22 @@
 import { stringify as yamlStringify } from 'yaml';
 
 import { buildArtifact as coreBuildArtifact } from '@hyperlane-xyz/core/buildArtifact.js';
+import { ChainAddresses } from '@hyperlane-xyz/registry';
 import {
   ChainMap,
   ChainName,
   ContractVerifier,
   CoreConfig,
+  CosmosNativeCoreModule,
   DeployedCoreAddresses,
   EvmCoreModule,
   ExplorerLicenseType,
 } from '@hyperlane-xyz/sdk';
+import { ProtocolType, assert } from '@hyperlane-xyz/utils';
 
 import { MINIMUM_CORE_DEPLOY_GAS } from '../consts.js';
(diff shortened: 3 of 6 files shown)


typescript/cli/src/deploy/core.ts
```diff
@@ -115,29 +153,72 @@ export async function runCoreDeploy(params: DeployParams) {
...
-      `Core config on ${chain} is the same as target. No updates needed.`,
-    );
+      const transactions = await cosmosNativeCoreModule.update(config);
+
+      if (transactions.length) {
+        logGray('Updating deployed core contracts');
+        const response = await signer.signAndBroadcast(
+          signer.account.address,
+          transactions,
+          2,
```
xeno097, line 207: What does 2 mean in the context of the fee?

xeno097 (commented):
are planning to add e2e tests in a follow up pr?

typescript/cli/src/deploy/core.ts
troykessler: it means that it will multiply the estimated gas from the simulation by 2 to be sure it is enough to pass

troykessler:
> are planning to add e2e tests in a follow up pr?

yes, hooking up the CosmosCoreModule logic with the commands is the last PR needed to enable e2e testing of the cosmos cli, this will follow in the next pr

xeno097 (approved):

typescript/cli/src/read/core.ts
```diff
@@ -23,17 +28,46 @@ export async function executeCoreRead({
...
-    process.exit(1);
+  const protocolType = context.multiProvider.getProtocol(chain);
+
+  switch (protocolType) {
+    case ProtocolType.Ethereum: {
+      const evmCoreReader = new EvmCoreReader(context.multiProvider, chain);
+      try {
+        return evmCoreReader.deriveCoreConfig({
+          mailbox,
+          interchainAccountRouter: addresses?.interchainAccountRouter,
```
paulbalaji, line 39: ahhh just because of the ICA router we can't dedupe more logic

@xeno097 how critical is that in the core read? like who actually needs to check ICA router on EVMs as part of the core deployment?

typescript/cli/src/deploy/core.ts
```diff
@@ -115,29 +153,72 @@ export async function runCoreDeploy(params: DeployParams) {
...
-    logGray('Updating deployed core contracts');
-    for (const transaction of transactions) {
-      await multiProvider.sendTransaction(
-        // Using the provided chain id because there might be remote chain transactions included in the batch
-        transaction.chainId ?? chain,
-        transaction,
-      );
+  const { multiProvider, multiProtocolSigner } = context;
+
+  switch (multiProvider.getProtocol(chain)) {
```
paulbalaji, line 158: is it possible to dedupe more of this? part of the idea with having the abstract module is that you can somewhat initialise them and then it should just be the same function call to get update txs for each VM, and then afterwards we can iterate through and apply the txs

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -148,7 +148,10 @@ export class MultiChainResolver implements ChainResolver {
 
     // If no destination is specified, return all EVM chains
```
paulbalaji, line 149: update comment

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -198,6 +201,14 @@ export class MultiChainResolver implements ChainResolver {
...
 
+  private getCosmosNativeChains(multiProvider: MultiProvider): ChainName[] {
+    const chains = multiProvider.getKnownChainNames();
+
+    return chains.filter(
+      (chain) => multiProvider.getProtocol(chain) === ProtocolType.CosmosNative,
+    );
+  }
+
   static forAgentKurtosis(): MultiChainResolver {
```
paulbalaji, line 212: sidenote: should probably create a ticket to nuke kurtosis references from cli/sdk

typescript/cli/src/deploy/core.ts
troykessler: I think it should be possible but it would be some more changes. I would suggest creating a single additional PR just for refactoring and trying to get rid of those (kinda) duplicated lines.

typescript/cli/src/read/core.ts
troykessler: I can also create the same method signature for `deriveCoreConfig` in `CosmosNativeCoreReader` and simply ignore the ICA address there. But like in the reply below I would suggest creating an additional PR just for refactoring

typescript/cli/src/config/hooks.ts
```diff
@@ -309,7 +309,7 @@ async function getIgpTokenPrices(
       chainMetadata: filteredMetadata,
     });
     const results = await tokenPriceGetter.getAllTokenPrices();
-    fetchedPrices = objMap(results, (v) => v.toString());
+    fetchedPrices = objMap(results, (v) => results[v].toString());
```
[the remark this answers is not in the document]
troykessler: This was not the correct value. `v` was the key (chain name) and not the value (price) of the object

paulbalaji (approved):

PR #6264, opened 2025-05-19
chore: add routing ism to cosmos native ism module

troykessler:
### Description

This PR add the Cosmos Native Routing ISM to the CosmosNativeIsmModule.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Tests


typescript/sdk/src/ism/CosmosNativeIsmModule.ts
```diff
@@ -134,27 +172,157 @@ export class CosmosNativeIsmModule extends HyperlaneModule<
...
+
+  protected async deployMessageIdMultisigIsm(
+    config: MultisigIsmConfig,
+  ): Promise<Address> {
+    const { response } = await this.signer.createMessageIdMultisigIsm({
+      validators: config.validators,
+      threshold: config.threshold,
+    });
+    return response.id;
+  }
```
xeno097, line 211: shouldn't we check that the threshold makes sense compared to the number of validators? like we don't have threshold set to 10 but only have 1 validator

troykessler: the cosmos-sdk module would throw if you would send a transaction with threshold being greater than the number of validators, added an assert check anyway

typescript/sdk/src/ism/CosmosNativeIsmModule.ts
```diff
@@ -85,16 +99,40 @@ export class CosmosNativeIsmModule extends HyperlaneModule<
...
+    // if the ISM is a static ISM we can not update it, instead
+    // it needs to be recreated with the expected config
+    if (STATIC_ISM_TYPES.includes(expectedConfig.type)) {
+      this.args.addresses.deployedIsm = await this.deploy({
+        config: expectedConfig,
+      });
 
-    return [];
+      return [];
+    }
```
paulbalaji, line 118: does the "static ism types" thing still apply to cosmos? asking because if we can update things like aggregation/multisig isms in-place then might be useful to have that functionality

typescript/sdk/src/ism/CosmosNativeIsmModule.ts
```diff
@@ -134,27 +172,165 @@ export class CosmosNativeIsmModule extends HyperlaneModule<
...
+        value: R.MsgRemoveRoutingIsmDomain.proto.converter.create({
+          owner: actual.owner,
+          ism_id: this.args.addresses.deployedIsm,
+          domain,
+        }),
+      });
+    }
+
+    // Update ownership
+    if (actual.owner !== expected.owner) {
```
paulbalaji, line 313: should use eqAddress for address comparisons

typescript/sdk/src/ism/CosmosNativeIsmModule.ts
troykessler: Yes it does, only the RoutingIsm can be updated, the other ones have to be newly deployed if something changes there

xeno097 (approved):

PR #6265, opened 2025-05-19
feat: allow for absolute or relative min amount strategy

fernandomg:
### Description

This is a refactor to the `MinAmount` strategy.

Allows to specify `minAmount` and `target` int _relative_ or _absolute_ terms.
- Relative: values goes from `0` to `1`, and represent a percentage of the sum of collaterals of rebalanceable amounts
- Absolute: values are greater than `1` and represent the token units for that particular chain

### Backward compatibility

No. The MinAmount `buffer` no longer exist. It was replaced with `target` and its meaning changed as well.

### Testing

Manual/Unit Tests


typescript/cli/src/rebalancer/config/Config.ts
```diff
@@ -35,18 +35,23 @@ const WeightedChainConfigSchema = BaseChainConfigSchema.extend({
...
 });
 
-// Min amount strategy config schema
-const MinAmountChainConfigSchema = BaseChainConfigSchema.extend({
+// Absolute values, represents in token units the min and the target
+const MinAmountAbsoluteConfigSchema = BaseChainConfigSchema.extend({
   minAmount: z
     .string()
-    .or(z.number())
+    .or(z.number().gt(1))
```
fzavalia, line 42: How will this work with strings? For example, If I configure "0.1".

typescript/cli/src/rebalancer/config/Config.ts
```diff
@@ -35,18 +35,23 @@ const WeightedChainConfigSchema = BaseChainConfigSchema.extend({
...
-    .optional(),
+    .or(z.number().gt(1))
+    .transform((val) => BigInt(val)),
+});
+
+// Relative values, with 0 = 0%, and 1 = 100%
+// It's relative to the total sum of collateral across al rebalanceable chains
+const MinAmountRelativeConfigSchema = BaseChainConfigSchema.extend({
+  minAmount: z.number().min(0).max(1),
+  target: z.number().min(0).max(1),
```
fzavalia, line 54: Wont this have some sort of type collision with the absolute which uses BigInt instead?

typescript/cli/src/rebalancer/strategy/MinAmountStrategy.ts
```diff
@@ -5,57 +5,105 @@ import type { RawBalances } from '../interfaces/IStrategy.js';
...
+        const config = this.config[chain];
         const balance = rawBalances[chain];
+        let minAmount: bigint;
+        let targetAmount: bigint;
+
+        if (!config.isRelative) {
+          minAmount = config.minAmount;
+          targetAmount = config.target;
+        } else {
+          minAmount = BigInt(Math.floor(Number(total) * config.minAmount));
```
fzavalia, line 100: Could there be an issue when transforming bigints to normal numbers?

fzavalia: 9007199254740991 is the largest safe integer before losing precision. Total, specially with tokens with 18 decimals might have larger values.

My recommendation would be to min amount convert it to 0 - 100 in this step and then divide by 100 (or even use 0 to 1 million as denominator for more precision, similar to how it is done on smart contracts)

fzavalia: for example

total * BigInt(config.minAmount * 1_000_000n) / 1_000_000n

fernandomg: I went with a BigNumber approach, works. Let me know if you feel it too expensive, I can change it to the proposedapproach

typescript/cli/src/rebalancer/config/Config.test.ts
```diff
@@ -169,6 +169,42 @@ describe('Config', () => {
     });
   });
 
+  it('should load if minAmount and target if strategy is relative', () => {
```
fzavalia, line 172: I can't quite understand this phrasing

typescript/cli/src/rebalancer/config/Config.test.ts
```diff
@@ -169,6 +169,42 @@ describe('Config', () => {
...
+    expect(
+      Config.load(REBALANCER_CONFIG_PATH, ANVIL_KEY, {}).chains.chain1,
+    ).to.deep.equal({
+      ...data.chain1,
+      minAmount: '0.2',
+      target: 0.3,
+    });
+  });
+
+  it('should load if minAmount and target if strategy is absolute', () => {
```
fzavalia, line 190: Same here

fzavalia (approved):
💯

PR #6267, opened 2025-05-19
refactor: rename `bridgeTolerance` to `bridgeLockTime`

fernandomg:
### Description

A semantic change. That is not backwards compatible.

### Backward compatibility

No


fzavalia (approved):
🚀

PR #6269, opened 2025-05-19
feat: allow to explicitly rebalance

fernandomg:
### Description

WIP: explicitly rebalance using CLI args


Mo-Hussain:
I have a preference for user to have to explicitly signaling their intent for doing a manual run as opposed to inferring it by the existence of --fromChain, --toChain, and --amount. What do you think about defining a 'manual' strategy where the user must pass --fromChain, --toChain, and --amount?

fzavalia:
> I have a preference for user to have to explicitly signaling their intent for doing a manual run as opposed to inferring it by the existence of --fromChain, --toChain, and --amount. What do you think about defining a 'manual' strategy where the user must pass --fromChain, --toChain, and --amount?

We haven't seen this message 👀 seems like a good idea, specially given that it reuses the strategy architecture

PR #6272, opened 2025-05-19
refactor: simplify ccip read ISM

yorhodes:
### Description

- Enables overriding `MailboxClient.interchainSecurityModule` to avoid unnecessary SLOAD
- Pulls URL config into shared contract
- Shares event revert and mailbox process logic

### Backward compatibility

No

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/client/MailboxClient.sol
+++ b/solidity/contracts/client/MailboxClient.sol
@@ -36,7 +36,7 @@ abstract contract MailboxClient is OwnableUpgradeable, PackageVersioned {
 
     IPostDispatchHook public hook;
 
-    IInterchainSecurityModule public interchainSecurityModule;
+    IInterchainSecurityModule internal _interchainSecurityModule;
 
     uint256[48] private __GAP; // gap for upgrade safety
 
@@ -74,6 +74,15 @@ abstract contract MailboxClient is OwnableUpgradeable, PackageVersioned {
         _transferOwnership(msg.sender);
     }
 
+    function interchainSecurityModule()
+        external
+        view
+        virtual
+        returns (IInterchainSecurityModule)
+    {
+        return _interchainSecurityModule;
+    }
+
     /**
      * @notice Sets the address of the application's custom hook.
      * @param _hook The address of the hook contract.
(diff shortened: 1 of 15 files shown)


solidity/contracts/client/MailboxClient.sol
```diff
@@ -36,7 +36,7 @@ abstract contract MailboxClient is OwnableUpgradeable, PackageVersioned {
 
     IPostDispatchHook public hook;
 
-    IInterchainSecurityModule public interchainSecurityModule;
+    IInterchainSecurityModule internal _interchainSecurityModule;
```
larryob, line 39: I think this actually might cause confusion because of the way underscores are used for parameters

solidity/contracts/isms/ccip-read/TestCccipReadIsm.sol
```diff
@@ -1,14 +1,6 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity >=0.8.0;
```
larryob, line 2: Would be nice if this whole file was moved under `contracts/test`

larryob: There's a typo in the filename too, there should only be two `C`s.

solidity/contracts/isms/ccip-read/AbstractCcipReadIsm.sol
```diff
@@ -21,10 +21,50 @@ import {PackageVersioned} from "../../PackageVersioned.sol";
...
         uint8(IInterchainSecurityModule.Types.CCIP_READ);
+
+    string[] internal _urls;
+
+    /**
+     * @notice Emitted when new CCIP-read urls are being set
+     */
+    event UrlsChanged(string[] newUrls);
+
+    function setUrls(string[] memory __urls) public onlyOwner {
```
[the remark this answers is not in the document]
yorhodes: its for liveness only so imo unnecessary

solidity/contracts/client/MailboxClient.sol
yorhodes: we do the same thing in `Router`

solidity/contracts/isms/ccip-read/TestCccipReadIsm.sol
yorhodes: gonna leave this as out of scope

solidity/contracts/client/MailboxClient.sol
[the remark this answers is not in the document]
yorhodes: gonna leave this as out of scope

PR #6273, opened 2025-05-19
feat: implement token bridge for all warp routes

yorhodes:
### Description

- TokenRouter implements ITokenBridge
- HypERC20Collateral returns token approval in quote transfer remote

### Drive-by changes

- Fixes Op token bridge inheritance tree
- Fixes broken fuzz test

### Backward compatibility

Yes

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/token/HypERC20Collateral.sol
+++ b/solidity/contracts/token/HypERC20Collateral.sol
@@ -14,6 +14,7 @@ pragma solidity >=0.8.0;
 @@@@@@@@@       @@@@@@@@*/
 
 // ============ Internal Imports ============
+import {TokenMessage} from "./libs/TokenMessage.sol";
 import {TokenRouter} from "./libs/TokenRouter.sol";
 import {FungibleTokenRouter} from "./libs/FungibleTokenRouter.sol";
 import {MovableCollateralRouter} from "./libs/MovableCollateralRouter.sol";
@@ -25,6 +26,7 @@ import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {Context} from "@openzeppelin/contracts/utils/Context.sol";
 import {ContextUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
+import {Quote} from "../interfaces/ITokenBridge.sol";
 
 /**
  * @title Hyperlane ERC20 Token Collateral that wraps an existing ERC20 with remote transfer functionality.
@@ -61,6 +63,20 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
         address _account
     ) external view override returns (uint256) {
         return wrappedToken.balanceOf(_account);
+    }
+
+    function quoteTransferRemote(
+        uint32 _destinationDomain,
+        bytes32 _recipient,
+        uint256 _amount
+    ) external view virtual override returns (Quote[] memory quotes) {
+        quotes = new Quote[](2);
+        quotes[0] = _quoteTransferRemote(
+            _destinationDomain,
+            _recipient,
+            _amount
+        )[0];
+        quotes[1] = Quote({token: address(wrappedToken), amount: _amount});
     }
 
     /**
--- a/solidity/contracts/token/HypERC20Collateral.sol
+++ b/solidity/contracts/token/HypERC20Collateral.sol
@@ -71,11 +71,10 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
         uint256 _amount
     ) external view virtual override returns (Quote[] memory quotes) {
         quotes = new Quote[](2);
-        quotes[0] = _quoteTransferRemote(
-            _destinationDomain,
-            _recipient,
-            _amount
-        )[0];
+        quotes[0] = Quote({
+            token: address(0),
+            amount: _quoteGasPayment(_destinationDomain, _recipient, _amount)
+        });
         quotes[1] = Quote({token: address(wrappedToken), amount: _amount});
     }
 
--- a/solidity/contracts/token/HypNative.sol
+++ b/solidity/contracts/token/HypNative.sol
@@ -46,15 +46,17 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
         _MovableCollateralRouter_initialize(_owner);
     }
 
-    function _quoteTransferRemote(
+    function quoteTransferRemote(
         uint32 _destination,
         bytes32 _recipient,
         uint256 _amount
-    ) internal view override returns (Quote[] memory quotes) {
-        quotes = super._quoteTransferRemote(_destination, _recipient, _amount);
-        assert(quotes[0].token == address(0));
-        // add the amount to the quote
-        quotes[0].amount += _amount;
+    ) external view override returns (Quote[] memory quotes) {
+        quotes = new Quote[](1);
+        quotes[0] = Quote({
+            token: address(0),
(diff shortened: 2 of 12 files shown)


solidity/contracts/token/extensions/OPL2ToL1TokenBridgeNative.sol
```diff
@@ -31,37 +31,27 @@ contract OPL2ToL1TokenBridgeNative is ITokenBridge, HypNative {
...
-        address remoteRouter = _mustHaveRemoteRouter(l1Domain)
-            .bytes32ToAddress();
+        uint256 _amount,
+        uint256 _value,
+        bytes memory _hookMetadata,
+        address _hook
+    ) internal virtual override returns (bytes32 messageId) {
+        bytes32 remoteRouter = _mustHaveRemoteRouter(l1Domain);
+
+        messageId = super._transferRemote(
```
larryob, line 44: I think the explicit parent class name is better than `super`.

solidity/contracts/token/libs/TokenRouter.sol
```diff
@@ -175,17 +176,28 @@ abstract contract TokenRouter is GasRouter {
...
+        bytes32 _recipient,
+        uint256 _amount
+    ) internal view virtual returns (Quote[] memory quotes) {
+        uint256 nativeFee = _GasRouter_quoteDispatch(
+            _destinationDomain,
+            TokenMessage.format(_recipient, _amount),
+            address(hook)
+        );
+        quotes = new Quote[](1);
+        quotes[0] = Quote(address(0), nativeFee);
```
larryob, line 200: Nit: Use the field names

solidity/contracts/token/HypNative.sol
```diff
@@ -45,39 +46,34 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
...
+    function _quoteTransferRemote(
         uint32 _destination,
         bytes32 _recipient,
         uint256 _amount
-    ) external payable virtual override returns (bytes32 messageId) {
-        require(msg.value >= _amount, "Native: amount exceeds msg.value");
-        uint256 _hookPayment = msg.value - _amount;
-        return _transferRemote(_destination, _recipient, _amount, _hookPayment);
+    ) internal view override returns (Quote[] memory quotes) {
+        quotes = super._quoteTransferRemote(_destination, _recipient, _amount);
```
larryob, line 54: I prefere the explicit class name.

solidity/contracts/token/HypNative.sol
```diff
@@ -45,39 +46,34 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
...
+        assert(quotes[0].token == address(0));
```
larryob, line 55: This check should be moved to the appropriate forge test.

solidity/contracts/token/HypNative.sol
```diff
@@ -91,12 +87,11 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
...
     /**
      * @inheritdoc TokenRouter
-     * @dev No-op because native amount is transferred in `msg.value`
-     * @dev Compiler will not include this in the bytecode.
      */
     function _transferFromSender(
-        uint256
+        uint256 _amount
     ) internal virtual override returns (bytes memory) {
+        require(msg.value >= _amount, "Native: amount exceeds msg.value");
```
larryob, line 94: Good cleanup

solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -63,6 +65,20 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
...
+        bytes32 _recipient,
+        uint256 _amount
+    ) external view virtual override returns (Quote[] memory quotes) {
+        quotes = new Quote[](2);
+        quotes[0] = _quoteTransferRemote(
+            _destinationDomain,
+            _recipient,
+            _amount
+        )[0];
+        quotes[1] = Quote({token: address(wrappedToken), amount: _amount});
```
larryob, line 79: This seems to imply that the second fee is the whole amount amount of the token being transferred.

solidity/contracts/token/HypNative.sol
```diff
@@ -45,39 +46,34 @@ contract HypNative is FungibleTokenRouter, MovableCollateralRouter {
...
         uint256 _amount
-    ) external payable virtual override returns (bytes32 messageId) {
-        require(msg.value >= _amount, "Native: amount exceeds msg.value");
-        uint256 _hookPayment = msg.value - _amount;
-        return _transferRemote(_destination, _recipient, _amount, _hookPayment);
+    ) internal view override returns (Quote[] memory quotes) {
+        quotes = super._quoteTransferRemote(_destination, _recipient, _amount);
+        assert(quotes[0].token == address(0));
+        // add the amount to the quote
+        quotes[0].amount += _amount;
```
larryob, line 57: Why would we do this?

larryob (commented):
Left a couple comments

solidity/contracts/token/HypNative.sol
yorhodes: so that we can have a very simple client implementation that just sends whatever is quoted

solidity/contracts/token/HypERC20Collateral.sol
nambrot: I think we should document this behavior because this probably wasn't most folks' interpretation of the original design doc. Probably would specify this at least on this implementation as well as the `ITokenBridge` interface

solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -63,6 +65,20 @@ contract HypERC20Collateral is FungibleTokenRouter, MovableCollateralRouter {
         return wrappedToken.balanceOf(_account);
     }
 
+    function quoteTransferRemote(
+        uint32 _destinationDomain,
+        bytes32 _recipient,
+        uint256 _amount
+    ) external view virtual override returns (Quote[] memory quotes) {
+        quotes = new Quote[](2);
+        quotes[0] = _quoteTransferRemote(
```
nambrot, line 74: What if a child class returns multiple quotes? IMO this cannot make assumptions here that it is only one quote

solidity/contracts/token/HypNative.sol
nambrot: Ditto, I don't think you can be asserting this like the above comment

solidity/contracts/token/HypERC20Collateral.sol
nambrot: I also think this is all the more reason to document this better. I would even go a step further and say that we should name this internal function differently. It would be too easy for consumers to believe that they have to be quoting for the wrappedTokenAmount in here.

I would maybe even question whether we should have an internal function that is being overridden at all. Wdyt about simply having a default implementaiton here and in HypNative that uses an internal function to quote from the postDispatch hook, but if a child overrides quoting, they have to explicitly override the external function, and not the internal function (which imo makes it too hard to reason about how that interplays with subclasses)

solidity/contracts/token/libs/TokenRouter.sol
```diff
@@ -175,17 +176,28 @@ abstract contract TokenRouter is GasRouter {
...
-    function quoteGasPayment(
-        uint32 _destinationDomain
-    ) public view override returns (uint256) {
-        return
-            _GasRouter_quoteDispatch(
-                _destinationDomain,
-                TokenMessage.format(bytes32(0), type(uint256).max, bytes("")),
-                address(hook)
-            );
+    function quoteTransferRemote(
```
nambrot, line 181: Is this basically for non-collateral tokenrouters? if so, probably useful to indicate in a comment

nambrot (approved):

PR #6274, opened 2025-05-19
refactor: cctp token bridge

yorhodes:
depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6272 (review that first)

### Description

- Adds check for ambiguous circle domain 0
- Send to transfer remote recipient instead of router to remove trust assumption on hyperlane message
- Proxy in tests to ensure infinite approval is exercised

### Backward compatibility

Yes

### Testing

Unit Tests


solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -134,23 +126,164 @@ abstract contract TokenBridgeCctp is
         );
 
         // Receive only if the nonce hasn't been used before
-        if (!_isMessageReceived(cctpMessage)) {
+        if (messageTransmitter.usedNonces(_messageNonce(cctpMessage)) == 0) {
```
nambrot, line 129: It did seem nicer to have a named function for this check?

solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -134,23 +126,164 @@ abstract contract TokenBridgeCctp is
...
+        )
+    {}
+
+    // @dev the minimum to consider it a Standard CCTP transfer (it applies to every network)
+    // see https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/v2/MessageTransmitterV2.sol#L224-L244
+    // and https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/v2/FinalityThresholds.sol#L21
+    uint32 internal constant CCTP_V2_DEFAULT_MIN_FINALITY_THRESHOLD = 2000;
+
+    // @dev required for CCTP v2
+    uint256 internal constant CCTP_V2_DEFAULT_MAX_FEE = 0;
```
nambrot, line 261: Should we make this an immutable that can be set in the constructor?

nambrot (approved):

solidity/contracts/token/TokenBridgeCctp.sol
yorhodes: this matches the comment better

PR #6278, opened 2025-05-19
feat: fix crashes and set metrics instead when rpcs fail

kamiyaa:
### Description

 - prevent `from_settings` from crashing
 - set chain critical error when contract fails to sync

### Related issues

 - fixes https://linear.app/hyperlane-xyz/issue/ENG-1650/either-remove-or-catch-unwraps-when-initializing-agents

### Backward compatibility

Yes


rust/main/hyperlane-base/src/settings/base.rs
```diff
@@ -210,49 +210,6 @@ impl Settings {
...
-                .await?;
-            syncs.push(sync);
-        }
-
-        syncs
-            .into_iter()
-            .map(|i| Ok((i.domain().clone(), i)))
-            .collect()
-    }
-
```
ameten, line 255: I am not sure I understand why we need to move this method to `relayer.rs` and make it non-generic?

Presumably, we can remove the early exit at `self.contract_sync`, return the map which won't contain syncs for some domains and analyse the map in relayer.

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -216,63 +216,53 @@ impl BaseAgent for Relayer {
...
-                    .map(|(d, db)| (d.clone(), Arc::new(db.clone())))
-                    .collect(),
-                false,
-                settings.tx_id_indexing_enabled,
-            )
-            .await?
-            .into_iter()
-            .map(|(k, v)| (k, v as _))
+
+        let advanced_log_meta = false;
```
ameten, line 220: It is good idea to extract advance log meta flag into a variable! It will be better if it is just a constant in Relayer.

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -1034,6 +1040,141 @@ impl Relayer {
...
+                    advanced_log_meta,
+                    settings.tx_id_indexing_enabled,
+                )
+                .await
+            {
+                Ok(sync) => sync,
+                Err(err) => {
+                    let error_msg = "contract_sync() failed";
+                    tracing::error!(?domain, ?err, "{error_msg}");
+                    Self::record_critical_error(domain, &err, &error_msg, chain_metrics.clone());
```
ameten, line 1124: It would be better if error message is customised for the sync we are creating.

rust/main/hyperlane-base/src/settings/base.rs
kamiyaa: It was easier to add in the logic to set critical error when one of the contract syncs fail.

I guess I could change `contract_syncs()` to return a list of contracts syncs that fail and then set it after this function is done running. What do you think?

kamiyaa: Updated!

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -360,8 +350,16 @@ impl BaseAgent for Relayer {
...
             .into_iter()
             .collect::<HashMap<_, _>>();
+
+        let mut destination_chains = HashMap::new();
         for (destination, dest_mailbox) in mailboxes.iter() {
-            let destination_chain_setup = core.settings.chain_setup(destination).unwrap().clone();
+            let destination_chain_setup = match core.settings.chain_setup(destination) {
+                Ok(setup) => setup.clone(),
+                Err(err) => {
+                    tracing::error!(?destination, ?err, "chain_setup() failed");
```
tkporter, line 359:
```suggestion
                    tracing::error!(?destination, ?err, "Destination chain setup failed");
```
nit: just a bit nicer / more readable than a reference to a function

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -374,9 +372,21 @@ impl BaseAgent for Relayer {
...
+                        tracing::error!(origin=?origin.name(), "DB missing");
+                        continue;
+                    }
+                };
                 let default_ism_getter = DefaultIsmCache::new(dest_mailbox.clone());
-                let origin_chain_setup = core.settings.chain_setup(origin).unwrap().clone();
+                let origin_chain_setup = match core.settings.chain_setup(origin) {
+                    Ok(chain_setup) => chain_setup.clone(),
+                    Err(err) => {
+                        tracing::error!(origin=?origin.name(), ?err, "Chain setup error");
```
tkporter, line 386:
```suggestion
                        tracing::error!(origin=?origin.name(), ?err, "Origin chain setup failed");
```
and for consistency

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -1068,11 +1099,113 @@ impl Relayer {
...
+                            "Critical error when building application operation verifier",
+                        );
+                        None
+                    }
+                },
+            )
+            .collect()
+    }
+
+    pub async fn build_message_contract_syncs(
```
tkporter, line 1119: these 3 functions are the exact same but just with different return types and a different critical error message right?

can we just make this 1 function with a generic return type and make the message be `"Critical error when building {data_type} contract sync"`?

rust/main/hyperlane-base/src/settings/base.rs
```diff
@@ -215,13 +215,12 @@ impl Settings {
...
     pub async fn contract_syncs<T, S>(
         &self,
-        domains: impl Iterator<Item = &HyperlaneDomain>,
         metrics: &CoreMetrics,
         sync_metrics: &ContractSyncMetrics,
         stores: HashMap<HyperlaneDomain, Arc<S>>,
         advanced_log_meta: bool,
         broadcast_sender_enabled: bool,
-    ) -> Result<HashMap<HyperlaneDomain, Arc<dyn ContractSyncer<T>>>>
+    ) -> HashMap<HyperlaneDomain, eyre::Result<Arc<dyn ContractSyncer<T>>>>
```
tkporter, line 223: so I understand this PR correctly: this is the most significant thing changed, where instead of propagating an Err we will handle it more gracefully?

kamiyaa: yeah, we would handle it in the top level in the relayer code instead of here

rust/main/agents/relayer/src/relayer.rs
kamiyaa: Done

tkporter (approved):
lgtm!

PR #6284, opened 2025-05-20
feat: add rebalancer roles, support creating and fetching key from GCP

Mo-Hussain:
### Description

- add a new rebalancer role
- allow us to create and fetch a rebalancer key in GCP secret manager

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/typescript/infra/src/agents/key-utils.ts
+++ b/typescript/infra/src/agents/key-utils.ts
@@ -149,6 +149,19 @@ function getRoleKeyMapPerChain(
     }
   };
 
+  const setRebalancerKeys = () => {
+    const rebalancerKey = getRebalancerKey(agentConfig);
+    // Assign the single rebalancer key to all chains that have a relayer configured
+    for (const chainName of agentConfig.contextChainNames.relayer) {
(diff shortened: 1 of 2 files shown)


typescript/infra/src/agents/key-utils.ts
```diff
@@ -149,6 +149,19 @@ function getRoleKeyMapPerChain(
     }
   };
 
+  const setRebalancerKeys = () => {
+    const rebalancerKey = getRebalancerKey(agentConfig);
+    // Assign the single rebalancer key to all chains that have a relayer configured
+    for (const chainName of agentConfig.contextChainNames.relayer) {
```
paulbalaji, line 155: intentional for all relayed chains, or do we only want chains rebalancer is running on?

paulbalaji (approved):

typescript/infra/src/agents/key-utils.ts
Mo-Hussain: This is saying that we want to use the same rebalancer for all chains, it does mean we are creating a key per chain

PR #6286, opened 2025-05-20
chore(svm client: parse domains directly as u32

daniel-savu:


daniel-savu:
realizing this won't work, we need to map the chain name to the chain id instead

PR #6287, opened 2025-05-20
chore: add `SOL/apechain-solanamainnet`

christopherbrumm:
### Description

Add `SOL/apechain-solanamainnet` token config and program IDs.

### Backward compatibility

Yes

### Testing

N/A


christopherbrumm:
Included in #6249

PR #6288, opened 2025-05-20
chore: ica server deployment

mbreithecker:


typescript/ccip-server/src/server.ts
```diff
@@ -31,14 +32,29 @@ async function startServer() {
...
 
-  const port = process.env.PORT || 3000;
+  const port = parseInt(process.env.SERVER_PORT ?? '3000');
   app.listen(port, () => console.log(`Listening on port ${port}`));
 }
 
 startServer().then(console.log).catch(console.error);
+
+/*
+ * TODO: if PRISMA throws an error the entire express application crashes.
```
nambrot, line 52: Do you know why the whole application crashes? Shouldn't this be caught in the try/catch of `createAbiHandler`?

typescript/ccip-server/src/services/CCTPService.ts
```diff
@@ -23,17 +27,29 @@ class CCTPService extends BaseService {
...
     this.hyperlaneService = new HyperlaneService(
-      process.env.HYPERLANE_EXPLORER_URL!,
+      process.env.HYPERLANE_EXPLORER_API!,
     );
     this.cctpAttestationService = new CCTPAttestationService(
       process.env.CCTP_ATTESTATION_API!,
     );
-    // TODO: fetch this from a configured MultiProvider from IRegistry
-    this.rpcService = new RPCService(process.env.RPC_URL!);
+    const registry = getRegistry({
```
nambrot, line 35: What do you think about using the same approach of allowing configuration via `REGISTRY_URI` as in `CallCommitmentServer`?

typescript/ccip-server/src/server.ts
mbreithecker: Not sure, maybe prisma throws an async error which is not awaited. Found a Stackoverflow-Thread for this. Best solution is probably currently this.

typescript/ccip-server/src/services/CCTPService.ts
mbreithecker: Added support for REGISTRY_URI

typescript/ccip-server/src/server.ts
mbreithecker: okay, express v5 breaks the build, so would leave it to v4 in this PR and we upgrade in a separate PR

PR #6290, opened 2025-05-20
fix: storage diff-check

paulbalaji:
### Description

fix: storage diff-check


larryob (approved):
LGTM. `--pretty` is not an option.

PR #6299, opened 2025-05-21
feat(relayer): optimize ccip-read initialization

daniel-savu:
### Description

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6144 made relayer initialization very slow. It creates a signing provider within nested for loops, spending quadratic time (based on chain count) to create message contexts. At 110 chains and 200ms per network call that's ~40 mins (`110*110*200/(1000*60)`).

since the origin and destination chains are identical in the relayer, the first commit optimizes this by only initializing the provider in the upper for-loop, making it linear. This got us down to ~30s of initialization overhead added by the ccip read initialization.

the second commit moves out the provider building out of the for loops and just awaits the futures in parallel. AWS KMS rate limits shouldn't be a concern since it's just `chain_count` calls. After releasing to RC, the initialization overhead dropped to near zero

see the three logs [here](https://cloudlogging.app.goo.gl/H7LmsVr5hgzJSSdk7) for performance comparison:
- first log is without the CCIP-read feature (30 sec baseline)
- second log is with linear provider building (60 sec, so 30s overhead)
- third log is with the parallelized provider building (30 sec, back to the baseline)

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -32,8 +32,8 @@ use hyperlane_base::{
 };
 use hyperlane_core::{
     rpc_clients::call_and_retry_n_times, ChainCommunicationError, ChainResult, ContractSyncCursor,
-    HyperlaneDomain, HyperlaneMessage, InterchainGasPayment, Mailbox, MerkleTreeInsertion,
-    QueueOperation, SubmitterType, ValidatorAnnounce, H512, U256,
+    HyperlaneDomain, HyperlaneDomainProtocol, HyperlaneMessage, InterchainGasPayment, Mailbox,
+    MerkleTreeInsertion, QueueOperation, SubmitterType, ValidatorAnnounce, H512, U256,
 };
 use hyperlane_operation_verifier::ApplicationOperationVerifier;
 use submitter::{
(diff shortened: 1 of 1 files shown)


rust/main/agents/relayer/src/relayer.rs
```diff
@@ -329,6 +330,33 @@ impl BaseAgent for Relayer {
...
+                    };
+                    (destination, signer)
+                }
+            })
+            .collect::<Vec<_>>();
+        let ccip_signers = join_all(ccip_signer_futures)
+            .await
+            .into_iter()
+            .collect::<HashMap<_, _>>();
         for (destination, dest_mailbox) in mailboxes.iter() {
```
yjamin, line 360: nit: Maybe we should limit the number of concurrent requests here that we perform?

rust/main/agents/relayer/src/relayer.rs
```diff
@@ -329,6 +330,33 @@ impl BaseAgent for Relayer {
...
+        let ccip_signer_futures = mailboxes
+            .iter()
+            .map(|(destination, _)| {
+                let destination_chain_setup =
+                    core.settings.chain_setup(destination).unwrap().clone();
+                // return a future for `builder.build().await`
+                let signer = destination_chain_setup.signer.clone();
+                async move {
+                    let signer = if let Some(builder) = signer {
+                        match builder.build::<hyperlane_ethereum::Signers>().await {
```
yjamin, line 342: this can be a single match statement

yjamin: Maybe something like this?

```rust
let signer = match signer {
    Some(builder) => builder.build::<hyperlane_ethereum::Signers>().await.ok(),
    None => None,
};
```

rust/main/agents/relayer/src/relayer.rs
yjamin: The futures are lazy, meaning it will only be executed once we await them here with the join_all. But yes, I wouldn't expect that to be in issue soon as well :+1:

paulbalaji (approved):

yjamin (approved):

PR #6304, opened 2025-05-21
chore: new relayer image

daniel-savu:


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -895,7 +895,7 @@ const releaseCandidate: RootAgentConfig = {
     rpcConsensusType: RpcConsensusType.Fallback,
     docker: {
       repo,
-      tag: '0ad4122-20250514-154514',
+      tag: '4c678c5-20250521-125729',
```
daniel-savu, line 898: this image ignores reorg flag since hyperevm is still not fully recovered. Uses the last PR from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6216

PR #6124, opened 2025-05-05
chore: add cosmos native signer strategy to cli

troykessler:
### Description

This PR implements and adds the cosmos native signer strategy to the cli context so users can also sign and deploy to cosmos native chains. Following PRs will use this to support the goal of having full cosmos native support in the hyperlane CLI.

Note that this is the second PR of https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6050 which is being split up in order to make it easier to review

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual testing with local cosmos chain

PR #6309, opened 2025-05-22
fix: disable cosmos signers for now

troykessler:
### Description

This PR only inits Ethereum signers and temporary reverts the changes made in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6124. This is necessary, since Cosmos Signers are being initialized with an Eth private key in some circumstances which throw an error. The cosmos signers should therefore be disabled until a solution has been found to manage multi-vm signer intializations.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Tests

### DIFF
--- a/typescript/cli/src/context/strategies/signer/MultiProtocolSignerManager.ts
+++ b/typescript/cli/src/context/strategies/signer/MultiProtocolSignerManager.ts
@@ -76,7 +76,12 @@ export class MultiProtocolSignerManager {
    * @dev Configures signers for EVM chains in MultiProvider
    */
   async getMultiProvider(): Promise<MultiProvider> {
-    for (const chain of this.compatibleChains) {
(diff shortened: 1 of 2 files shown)


paulbalaji (approved):
lgtm, just needs a patch changeset

typescript/cli/src/context/strategies/signer/MultiProtocolSignerManager.ts
```diff
@@ -50,11 +50,11 @@ export class MultiProtocolSignerManager {
     this.initializeStrategies();
   }
 
+  // TODO: readd Cosmos Native
```
paulbalaji, line 53: make codespell happy
```suggestion
  // TODO: re-add Cosmos Native
```

PR #6314, opened 2025-05-22
Share tsconfig.json from a package

antigremlin:
### Description

This PR moves the shared TypeScript configuration into a NPM private package inside the monorepo, as recommended by Turborepo. The package isn't going to be published, just reused internally. This removes the need to reference parent paths in all our TypeScript config files.

### Related issues

Part of: ENG-1668

### Backward compatibility

Yes, as the TypeScript configuration stays unchanged.

### Testing

All TypeScript builds continue to work.


Dockerfile
```diff
@@ -10,24 +10,24 @@ COPY package.json yarn.lock .yarnrc.yml ./
 COPY .yarn/plugins ./.yarn/plugins
 COPY .yarn/releases ./.yarn/releases
 COPY .yarn/patches ./.yarn/patches
-COPY typescript/utils/package.json ./typescript/utils/
-COPY typescript/sdk/package.json ./typescript/sdk/
-COPY typescript/helloworld/package.json ./typescript/helloworld/
+COPY typescript/ccip-server/package.json ./typescript/ccip-server/
```
antigremlin, line 13: This is just sorting the typescript/* lines

paulbalaji (approved):

PR #6316, opened 2025-05-22
feat: agent-specific rpc consensus defaults

daniel-savu:
### Description

We recently discovered that validators use fallback provider as the default consensus type, which leaves them susceptible to reorgs regardless of how big their RPC set is.

This PR adds agent-specific rpc consensus default: fallback for the relayer and scraper, quorum for the validator. This default can still be overriden via the regular config.

### Drive-by changes

The `defaultRpcConsensusType` is removed from the agent config schema, since it applied to all chains and agents, which is much too generic at this point. It was always set to `fallback` in our infra TS code. It now still defaults to `fallback` in all cases but when the validator is run.

### Backward compatibility

The agent schema no longer contains `defaultRpcConsensusType`. See above. This won't break any builds with new agents that use old configs.

### Testing

Manual, by running a local validator and printing the `origin_chain_conf` to make sure it's using `HttpQuorum`

### DIFF
--- a/rust/main/agents/relayer/src/settings/mod.rs
+++ b/rust/main/agents/relayer/src/settings/mod.rs
@@ -137,7 +137,7 @@ impl FromRawConf<RawRelayerSettings> for RelayerSettings {
             .parse_from_raw_config::<Settings, RawAgentConf, Option<&HashSet<&str>>>(
                 relay_chain_names.as_ref(),
                 "Parsing base config",
-                agent_name,
+                agent_name.to_string(),
             )
             .take_config_err(&mut err);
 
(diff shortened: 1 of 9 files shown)


typescript/sdk/src/metadata/agentConfig.ts
```diff
@@ -295,12 +295,6 @@ export const AgentConfigSchema = z.object({
   defaultSigner: AgentSignerSchema.optional().describe(
     'Default signer to use for any chains that have not defined their own.',
   ),
-  defaultRpcConsensusType: z
-    .nativeEnum(RpcConsensusType)
-    .describe(
-      'The default consensus type to use for any chains that have not defined their own.',
-    )
-    .optional(),
```
paulbalaji, line 303: this will need a changeset `yarn changeset add`

ameten (approved):

PR #6320, opened 2025-05-22
chore: turn off preserveSymlinks

antigremlin:
### Description

This PR turns off the `preserveSymlinks` setting in tsconfig.json globally, with the benefit of fixing code navigation in VSCode. With this set to false (default and recommended), VSCode correctly navigates to the source code across package boundaries inside the monorepo, instead of landing in node_modules.

### Drive-by changes

Removing this setting uncovered a circular dependency between the SDK and the Registry, which led to compilation errors. To fix the circular dependency, imports from `@hyperlane-xyz/registry` were removed with small code fixes.

### Related issues

Fixes ENG-1668

### Backward compatibility

Yes, as the circular dependency wasn't declared in `package.json`, and wasn't imported anywhere else.

### Testing

The unit-tests are passing.

### DIFF
--- a/typescript/sdk/src/gas/token-prices.test.ts
+++ b/typescript/sdk/src/gas/token-prices.test.ts
@@ -1,13 +1,14 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
 
-import { ethereum, solanamainnet } from '@hyperlane-xyz/registry';
-
 import { TestChainName, testChainMetadata } from '../consts/testChains.js';
 
 import { CoinGeckoTokenPriceGetter } from './token-prices.js';
 
 const MOCK_FETCH_CALLS = true;
+
+const ethereum: { name?: string } = {};
+const solanamainnet: { name?: string } = {};
 
 describe('TokenPriceGetter', () => {
   let tokenPriceGetter: CoinGeckoTokenPriceGetter;
@@ -45,8 +46,8 @@ describe('TokenPriceGetter', () => {
       // stubbed results
       expect(
         await tokenPriceGetter.getTokenPriceByIds([
-          ethereum.name,
-          solanamainnet.name,
+          ethereum.name || 'ethereum',
+          solanamainnet.name || 'solanamainnet',
         ]),
       ).to.eql([priceA, priceB]);
     });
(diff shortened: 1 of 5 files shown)


typescript/sdk/src/utils/ism.ts
```diff
@@ -11,7 +12,7 @@ import { multisigIsmVerifyCosts } from '../consts/multisigIsmVerifyCosts.js';
 export function extractIsmAndHookFactoryAddresses(
   registryAddresses: ChainAddresses,
 ) {
-  return pick(registryAddresses, [
+  return pick(registryAddresses as Record<string, string>, [
```
paulbalaji, line 15:
```suggestion
  return pick(registryAddresses as ChainAddresses, [
```

typescript/sdk/src/gas/token-prices.test.ts
```diff
@@ -1,14 +1,15 @@
...
-import { ethereum, solanamainnet } from '@hyperlane-xyz/registry';
-
 import { TestChainName, testChainMetadata } from '../consts/testChains.js';
 
 import { CoinGeckoTokenPriceGetter } from './token-prices.js';
 
 const MOCK_FETCH_CALLS = true;
 
+const ethereum: { name?: string } = {};
+const solanamainnet: { name?: string } = {};
```
paulbalaji, line 11: surprised that the only thing needed from here is the name

antigremlin: Actually, I'll remove that. It was needed because a few lines of code in this file used `ethereum.name`, etc. But I changed that to constants and the remaining test code doesn't exercise `CoinGeckoTokenPriceGetter` enough to read any of the fields.

Once we get to improving these tests, we can add fuller chain configs to `testChains.ts` and use them.

typescript/sdk/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "extends": "@hyperlane-xyz/tsconfig/tsconfig.json",
   "compilerOptions": {
-    "outDir": "./dist/",
-    "rootDir": "./src/"
+    "outDir": "dist",
```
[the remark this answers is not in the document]
antigremlin: Just cleaner looking, so we can preserve the same syntax we use in most places

typescript/sdk/src/gas/token-prices.test.ts
antigremlin: Actually, can't remove these because `CoinGeckoTokenPriceGetter` expects the chain objects to be present in the chainmap, even if the objects themselves are empty.

PR #5216, opened 2025-01-20
feat: integrate zksync on cli

mshojaei-txfusion:
### Description

This PR adds ZKSync compatibility validation for core contract deployments:
- Adds pre-deployment check to validate ISM compatibility with chain's technical stack
- Integrates `isIsmCompatible` utility from SDK
- Adds assertion handling for incompatible ISM configurations

[PR to feat/integrate-zksync-on-sdk](https://github.com/txfusion/hyperlane-monorepo/pull/25)

### Related issues

https://linear.app/hyperlane-xyz/issue/ENG-1155

### Backward compatibility

Yes - This change adds additional validation but does not modify existing deployment behavior for supported chains.

### Testing
Passed build

PR #6331, opened 2025-05-23
feat: integrate zksync on CLI

mshojaei-txfusion:
### Description
This PR adds ZKSync compatibility validation for core contract deployments:

Adds pre-deployment check to validate ISM compatibility with chain's technical stack
Integrates isIsmCompatible utility from SDK
Adds assertion handling for incompatible ISM configurations
https://github.com/txfusion/hyperlane-monorepo/pull/25

### Related issues
https://linear.app/hyperlane-xyz/issue/ENG-1155

Previously was: https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5216

### Backward compatibility
Yes - This change adds additional validation but does not modify existing deployment behavior for supported chains.

Testing
Passed build


typescript/cli/src/deploy/warp.ts
```diff
@@ -143,6 +156,34 @@ async function runDeployPlanStep({ context, warpDeployConfig }: DeployParams) {
...
+          chainTechnicalStack,
+          ismType: config.interchainSecurityModule.type,
+        }),
+        `Selected ISM of type ${config.interchainSecurityModule.type} is not compatible with the selected Chain Technical Stack of ${chainTechnicalStack} for chain ${chain}!`,
+      );
+    }
+  }
+}
+
 async function executeDeploy(
```
paulbalaji, line 187: i think this is all before the warp module existed - needs to be migrated to warp module nomenclature

typescript/cli/src/deploy/warp.ts
```diff
@@ -143,6 +156,34 @@ async function runDeployPlanStep({ context, warpDeployConfig }: DeployParams) {
...
+    ) {
+      assert(
+        isIsmCompatible({
+          chainTechnicalStack,
+          ismType: config.interchainSecurityModule.type,
+        }),
+        `Selected ISM of type ${config.interchainSecurityModule.type} is not compatible with the selected Chain Technical Stack of ${chainTechnicalStack} for chain ${chain}!`,
+      );
+    }
+  }
```
xeno097, line 184: Isn't this quite similar to [this](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/b6d03cc18301843dddb6d4d5561a23f57fe8c582/typescript/cli/src/deploy/core.ts#L163-L180) with the only difference that here we iterate over each chain?

typescript/cli/src/deploy/warp.ts
xeno097: yes it is a relic from the past

typescript/cli/src/deploy/warp.ts
paulbalaji: yeah exactly - will be refactoring as part of the move to modules once the SDK pr is merged

paulbalaji: so theyre similar but not exact - still refactored so it reuses more logic

PR #6337, opened 2025-05-23
feat: `export-svm-config`

christopherbrumm:
### Description

This PR adds the `export-svm-config` script to write the registry config based SVM deploy artifacts. Besides this, the `install-solana.sh` script is made dynamically to use it for different versions.

### Drive-by changes

N/A

### Related issues

N/A

### Backward compatibility

Yes

### Testing

Reproduces the same config for already deployed warp routes.


rust/sealevel/programs/install-solana.sh
paulbalaji, line 1: does this require runbook changes? or changes anywhere else that reference the old script path?

christopherbrumm: This seems to be the only place where we'd need to update the path. However, I can keep the script name and change the flag to `--overwrite-version`.

PR #6339, opened 2025-05-23
feat(relayer): config to ignore reported reorgs

daniel-savu:
### Description

There are cases when validators have an unreliable RPC setup and end up reporting a reorg when no other validator in the set did. We want to halt processing as a failsafe but also have the option to disable the failsafe to speed up recovery - hence the config added in this PR, which applies to a chain when it is the origin of a message.

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/base_builder.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/base_builder.rs
@@ -7,6 +7,7 @@ use derive_new::new;
 use eyre::Context;
 use futures::{stream, StreamExt};
 use hyperlane_ethereum::Signers;
+use prometheus::IntGauge;
 use tokio::sync::RwLock;
 use tracing::{debug, info, warn};
 
@@ -44,6 +45,7 @@ pub struct BaseMetadataBuilder {
     app_context_classifier: IsmAwareAppContextClassifier,
     ism_cache_policy_classifier: IsmCachePolicyClassifier,
     signer: Option<Signers>,
+    ignore_reorg_reports: bool,
 }
 
 impl Debug for BaseMetadataBuilder {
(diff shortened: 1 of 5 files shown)


rust/main/agents/relayer/src/msg/metadata/base_builder.rs
```diff
@@ -287,4 +273,39 @@ impl BaseMetadataBuilder {
...
+            Ok(checkpoint_syncer) => {
+                return Ok(Some(checkpoint_syncer));
+            }
+            Err(CheckpointSyncerBuildError::ReorgEvent(reorg_event)) => {
+                if self.ignore_reorg_reports {
+                    warn!(
+                        ?reorg_event,
+                        "Ignoring reorg event for checkpoint syncer build"
+                    );
+                    return Ok(None);
```
daniel-savu, line 293: all of the code in this function except for this snippet is pre-existing code that was moved out of `build_checkpoint_syncer` above

rust/main/agents/relayer/src/msg/metadata/base_builder.rs
```diff
@@ -287,4 +273,41 @@ impl BaseMetadataBuilder {
         fetch_storage_locations_helper(validators, &self.cache, &*self.origin_validator_announce)
             .await
     }
+
+    async fn build_and_validate(
+        &self,
+        config: &CheckpointSyncerConf,
+        validator: &H256,
+        latest_index_gauge: Option<IntGauge>,
```
ameten, line 281: It looks like this parameter is always `None`. I wonder why we need it?

ameten (approved):

rust/main/agents/relayer/src/msg/metadata/base_builder.rs
daniel-savu: kept it just in case this would ever need to get called with a different value for `latest_index_gauge`. But I'll remove for now to keep things simple

PR #6343, opened 2025-05-26
feat: cosmos block events

yjamin:
### Description
Added block event indexing for cosmos. Block events are special in the sense that they do not have any TX Hash. This will result in the scraper not finding the given events, however, the relayer & validator do.

### Testing
Local E2E Testing

### DIFF
--- a/rust/main/chains/hyperlane-cosmos-native/src/indexers/indexer.rs
+++ b/rust/main/chains/hyperlane-cosmos-native/src/indexers/indexer.rs
@@ -130,21 +130,18 @@ where
     async fn get_logs_in_block(&self, block_height: u32) -> ChainResult<Vec<(T, LogMeta)>> {
         let block = self.provider().get_block(block_height).await?;
         let block_results = self.provider().get_block_results(block_height).await?;
-        let result = self.handle_txs(block, block_results);
+        let result = self.handle_block(block, block_results);
         Ok(result)
     }
 
     /// Iterate through all txs, filter out failed txs, find target events
-    /// in successful txs, and parse them.
-    fn handle_txs(
+    /// in successful txs, and parse them. Also iterates through all events in the block and tries to parse them.
+    fn handle_block(
         &self,
         block: BlockResponse,
         block_results: BlockResultsResponse,
     ) -> Vec<(T, LogMeta)> {
-        let Some(tx_results) = block_results.txs_results else {
-            return vec![];
-        };
-
+        let tx_results = block_results.txs_results.unwrap_or_default();
         // Cosmos can also emit events in the block itself, those events do not originate from a tx, but rather from the block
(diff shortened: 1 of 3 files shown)


rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
```diff
@@ -393,6 +393,9 @@ impl HyperlaneProvider for CosmosNativeProvider {
     }
 
     async fn get_txn_by_hash(&self, hash: &H512) -> ChainResult<TxnInfo> {
+        if hash.is_zero() {
+            return Err(HyperlaneProviderError::CouldNotFindTransactionByHash(*hash).into());
+        }
```
ameten, line 398: Our assumption for this method is that we can rarely fail to get a transaction with a hash. If we fail to do so, it means that there is a type of transaction which we cannot parse and need to analyse the case, implement the fix and let scraper to reindex the range.

This change assumes that we can expect zero hash it is a usual case, as far as I understand.

I wonder if this error impacts other transaction in the chucks which are stored by scraper?

yjamin: Based on your message I'd assume this is more of a scraper issue? I would expect the scraper to be fail safe on this method to some extend. My assumption was that if the scraper fetches a batch of txn, it would not fail for all the entire batch if one single tx is wrong.

rust/main/chains/hyperlane-cosmos-native/src/indexers/indexer.rs
```diff
@@ -142,6 +145,19 @@ where
             return vec![];
         };
 
+        // Cosmos can also emit events in the block itself, those events do not originate from a tx, but rather from the block
```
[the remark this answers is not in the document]
yjamin: Good catch, yes this should not happen. I'll fix it :+1:

rust/main/chains/hyperlane-cosmos-native/src/providers/cosmos.rs
ameten: You are correct. Scraper will store other transactions from the chunk. I think it is ok. My comment is incorrect.

The issue is that Scraper will report a warning for the case which we expect. We expect now to get zero hashes. It is not an error anymore. This case should not be lumped together with error of fetching and parsing a transaction, I think.

yjamin: I think a simple filter here should be sufficient: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/4fec5d33f047e39bc5ccd8d3618942b55f2a4942/rust/main/agents/scraper/src/store/storage.rs#L138-L143

ameten (approved):

PR #6344, opened 2025-05-26
feat: multi vm signers

troykessler:
### Description

This PR adds additional features to the CLIs `--key` flag by allowing protocol specific key inputs. The --key flag will stay backwards compatible and will only work for Ethereum chains. To input keys for Cosmos or Starknet chains in the future users have to specify the protocol they want to provide their key for, example: `core deploy --key.cosmosnative=COSMOS_PK --key.ethereum=ETH_PK`. This allows the CLI to handle multiple private keys for multiple VMs which was not possible before and needed with the first non-EVM protocols joining the monorepo.

### Drive-by changes

For whatever reason during a core deploy **all** signers in the registry where initiated. This is now limited that only the one chain where the core gets deployed to gets initiated. See changes in `MultiChainResolver.ts` `resolveCoreDeployChains`

### Related issues

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6309

### Backward compatibility

Yes these changes are backwards compatbile, the --key flag can be used as before

### Testing

e2e tests


paulbalaji (commented):
i wonder if it works out better to do `--key.<protocolType>` instead, where HYP_KEY or --key is just an alias for `--key.ethereum` protocoltype 

just food for thought

xeno097 (approved):
maybe as a follow up later in the future we might add something like `key.{protocolType}` too to avoid having to define private keys chain by chain (just saw the convo above)

typescript/cli/src/context/strategies/signer/MultiProtocolSignerManager.ts
```diff
@@ -123,30 +124,57 @@ export class MultiProtocolSignerManager {
...
   private async extractPrivateKey(chain: ChainName): Promise<SignerConfig> {
-    if (this.options.key) {
+    if (this.options.key && typeof this.options.key === 'object') {
+      assert(
+        this.options.key[chain],
+        `Key flag --key.${chain} for chain ${chain} not provided`,
+      );
       this.logger.debug(
-        `Using private key passed via CLI --key flag for chain ${chain}`,
+        `Using private key passed via CLI --key.${chain} flag for chain ${chain}`,
```
ltyu, line 155: also concur that this should be --key.protocol instead of chain mostly because currently --key assumes all chains are ethereum.

troykessler: Alright will rework this PR so that key inputs are per protocol now

PR #6359, opened 2025-05-27
feat: add batching support for starknet

yjamin:
### Description
Adds batching to Starknet, due to poor tx inclusion performance on Starknet and the native support for batched calls

### Drive-by changes

Added more explicit handling for the TX polling logic. Send 10 test tx in E2E instead of 2.

### Related issues
[Closes this issue.](https://linear.app/hyperlane-xyz/issue/ENG-1686/add-batching-support-for-relayer)

### Testing
Batching is now implemented in the E2E tests as well.

Tested on Neutron Testnet Relayer.

## Summary by CodeRabbit

- **New Features**
  - Added support for batch processing of messages in Starknet mailbox operations, enabling multiple messages to be processed in a single transaction.
  - Introduced configuration options for operation batching in Starknet connection settings and agent configuration, including a new max batch size parameter.
  - Added batch size overrides for Starknet and Paradex chains in mainnet and testnet agent configurations.

- **Improvements**
  - Increased the number of messages dispatched per target from 2 to 10 in local run utilities, removing delays between dispatches.
  - Enhanced transaction receipt retrieval with more flexible retry logic and configurable timeout.

- **Configuration**
  - Extended connection and agent configuration to support operation batching parameters for Starknet.


aroralanuk (commented):
looks good, just a couple of comments. If you can add a bit of detail around the testnet test, that'll be helpful!

rust/main/chains/hyperlane-starknet/src/mailbox.rs
```diff
@@ -221,4 +220,45 @@ impl Mailbox for StarknetMailbox {
...
+        // simulate the batch execution
+        let fee_estimate = tx
+            .estimate_fee()
+            .await
+            .map_err(HyperlaneStarknetError::from)?;
+
+        let tx = tx.max_fee(fee_estimate.overall_fee * FieldElement::TWO);
+        let outcome = send_and_confirm(&self.provider.rpc_client(), tx).await?;
+
+        // Either all operations are executed successfully, or none of them are
```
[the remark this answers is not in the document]
yjamin: Yeah there is some fallback logic in place, where the relayer will fallback to serial submission if the batch failed.

rust/main/chains/hyperlane-starknet/src/mailbox.rs
```diff
@@ -221,4 +220,45 @@ impl Mailbox for StarknetMailbox {
...
+            .collect();
+
+        let tx = self.contract.account.execute(calls);
+        // simulate the batch execution
+        let fee_estimate = tx
+            .estimate_fee()
+            .await
+            .map_err(HyperlaneStarknetError::from)?;
+
+        let tx = tx.max_fee(fee_estimate.overall_fee * FieldElement::TWO);
```
tkporter, line 255: any reason to just two? is this what we've been doing so far? curious what the downsides of this are

rust/main/utils/run-locally/src/starknet/types.rs
```diff
@@ -102,6 +103,7 @@ impl AgentConfig {
             domain_id: network.domain,
             metrics_port: network.metrics_port,
             mailbox: network.deployments.mailbox.clone(),
+            max_batch_size: 10,
```
tkporter, line 106: is this the batch size you'd recommend in prod? right now we set 32 for every chain by default - you can update this by using the `batchSizeOverrides` (example here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/546c6fd460003ae7f474b3d53eb3130e8a8e1a28/typescript/infra/config/environments/mainnet3/agent.ts#L955-L958) in the testnet4 & mainnet3 agent.ts files for all relayer contexts. We should do this before merging if we're uncomfortable with a 32 batch size

rust/main/chains/hyperlane-starknet/src/mailbox.rs
yjamin: I don't know where this TWO came from, was already present when I started working on it. It works fine for now, also there are fee refunds, so we don't end up paying more with a generous estimate.

Also, this can be more dynamic with the new version of the library. We could make this a config parameter like in cosmos, where we have `gasMultiplier` to account for this.

PR #6363, opened 2025-05-27
fix: protocol fee type & read routing ism

troykessler:
### Description

This PR fixes starknets core deploy by using the correct uint256 type for protocol fee values and fixing the routing ism read

### Drive-by changes

-

### Related issues

-

### Backward compatibility

Yes

### Testing

Manual Tests


yjamin (approved):
Looks good to me ╰( ͡° ͜ʖ ͡° )つ──☆*:・ﾟ

PR #6369, opened 2025-05-27
feat: send certain SVM warp route transactions from the onchain owner always

tkporter:
### Description

- Always attempts to send certain owner-only txs from the onchain owner. In practice these will only be signed & sent if the owner matches the payer keypair passed in. If it doesn't, the unsigned tx will be logged and prompted to the user.
- This means that an SVM warp route extension should just need to specify `-k ~/path/to/payer/keypair.json` as usual, and all the txs to deploy the new warp route will be done, and all the txs to enroll the remote router etc will be logged / prompted

### DIFF
--- a/rust/sealevel/client/src/context.rs
+++ b/rust/sealevel/client/src/context.rs
@@ -96,6 +96,17 @@ impl Context {
         }
     }
 
+    /// If the pubkey matches the payer's pubkey, return the payer's signer.
+    /// Otherwise, return a NullSigner for the given pubkey.
+    pub(crate) fn signer_for_pubkey(&self, pubkey: &Pubkey) -> Box<dyn Signer> {
+        if let Some(PayerKeypair { keypair, .. }) = &self.payer_keypair {
+            if &keypair.pubkey() == pubkey {
+                return self.payer_signer();
+            }
+        }
+        Box::new(NullSigner::new(pubkey))
+    }
+
     pub(crate) fn payer_keypair_path(&self) -> &String {
         &self
             .payer_keypair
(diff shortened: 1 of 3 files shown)


rust/sealevel/client/src/context.rs
```diff
@@ -183,30 +188,46 @@ impl<'ctx, 'rpc> TxnBuilder<'ctx, 'rpc> {
...
-        self.send(&[&*payer_signer])
+        let payer_pubkey = self.ctx.payer_pubkey;
+        self.send(&[payer_signer.as_deref()], &payer_pubkey)
     }
 
-    pub(crate) fn send<T: Signers>(
+    /// Sends the transaction with a signer for the given pubkey.
+    /// Note that a pubkey may not have an associated keypair, in which case
+    /// this function will return None and transactions will be printed to stdout
+    /// for user confirmation & manual submission.
```
daniel-savu, line 198: it looks like `NullSigner` is no longer being created in `payer_signer` after these changes

rust/sealevel/client/src/router.rs
```diff
@@ -590,10 +610,28 @@ fn configure_upgrade_authority(
...
+                        expected_upgrade_authority,
+                    ),
+                )
+                .with_client(&client)
+                .send_with_pubkey_signer(&actual_upgrade_authority);
+        } else {
+            // Flag if we can't change the upgrade authority
             println!(
-                "WARNING: Upgrade authority transfer cannot be completed for chain: {} ({}) from {:?} to {:?}, the existing upgrade authority is not the payer account",
+                "WARNING: Upgrade authority transfer cannot be completed for chain: {} ({}) from {:?} to {:?}, the existing upgrade authority is None",
```
daniel-savu, line 634: under what circumstances would the owner be None?

rust/sealevel/client/src/context.rs
tkporter: yeah, I believe the current version removed that comment if that's what this was calling out

PR #6370, opened 2025-05-27
refactor: Use manual rebalance strategy as flag to determine a manual rebalance

fzavalia:
Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/63

Manual rebalance is provisioned by the 'manual' strategy mode

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -550,15 +550,15 @@ export const rebalancer: CommandModuleWithWriteContext<{
 
       if (rebalanceStrategy === StrategyOptions.Manual) {
         if (!origin) {
-          throw new Error('Origin is required for manual rebalance');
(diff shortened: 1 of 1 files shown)


typescript/cli/src/commands/warp.ts
```diff
@@ -465,25 +462,6 @@ export const rebalancer: CommandModuleWithWriteContext<{
       demandOption: true,
       alias: ['rebalancerConfigFile', 'rebalancerConfig', 'configFile'],
     },
-    origin: {
```
Mo-Hussain, line 468: Apologies, I should have said this before explicitly, I anticipate that we will be running the rebalancer in manual mode as a one off cli command likely from a local machine. In which case, it would be preferable to pass cli arguments like we have here. We don't need support for this in the config

Mo-Hussain (approved):

PR #6377, opened 2025-05-28
chore: add new token standard

troykessler:
### Description

This PR adds a new TokenStandard `StarknetHypDexCollateral ` and a new TokenType `collateralDex` for the speical Paradex Token: https://github.com/hyperlane-xyz/hyperlane-starknet/pull/13

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

-

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -97,6 +97,7 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
       [TokenType.nativeScaled]: null,
       [TokenType.collateralUri]: null,
       [TokenType.syntheticUri]: null,
+      [TokenType.collateralDex]: null,
     };
 
     this.contractVerifier =
--- a/typescript/sdk/src/token/StarknetERC20WarpModule.ts
+++ b/typescript/sdk/src/token/StarknetERC20WarpModule.ts
@@ -102,6 +102,22 @@ export class StarknetERC20WarpModule {
           addresses[chain] = tokenAddress;
           break;
         }
+        case TokenType.collateral: {
+          const tokenAddress = await deployer.deployContract(
+            StarknetContractName.HYP_ERC20_COLLATERAL,
+            {
+              mailbox: mailbox!,
+              // @ts-ignore
+              erc20: rest.token,
+              owner: deployerAccountAddress, //TODO: use config.owner, and in warp init ask for starknet owner
+              hook: getChecksumAddress(0),
+              interchain_security_module: ismAddress,
+            },
+            ContractType.TOKEN,
+          );
+          addresses[chain] = tokenAddress;
+          break;
+        }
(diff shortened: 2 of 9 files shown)


typescript/sdk/src/token/StarknetERC20WarpModule.ts
```diff
@@ -102,10 +102,9 @@ export class StarknetERC20WarpModule {
           addresses[chain] = tokenAddress;
           break;
         }
-
-        case TokenType.collateral: {
+        case TokenType.collateralDex: {
           const tokenAddress = await deployer.deployContract(
-            StarknetContractName.HYP_ERC20_COLLATERAL,
+            StarknetContractName.HYP_ERC20_DEX_COLLATERAL,
```
xeno097, line 107: Shouldn't this deploy the contract based on which token type we are supplying here? wouldn't this cause a dex collateral to be deployed on starknet even if we only want a collateral? are they equivalent ? sorry don't have lots of context on this

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -97,6 +97,7 @@ export class EvmERC20WarpRouteReader extends HyperlaneReader {
       [TokenType.nativeScaled]: null,
       [TokenType.collateralUri]: null,
       [TokenType.syntheticUri]: null,
+      [TokenType.collateralDex]: null,
```
xeno097, line 100: Considering that we are mapping dex collateral to be collateral on evm, shouldn't we use the same reading function here instead of just setting to null?

typescript/sdk/src/token/StarknetERC20WarpModule.ts
aroralanuk: TLDR on dexCollateral: it's a variant of the collateral token in cairo which deposits the tokens to paradex's DEX. So, it's only paradex specific.

xeno097: so yeah forcing to deploy dex collateral in any case would be wrong right?

troykessler: made the error of replacing `collateral` with `collateralDex`, just readded collateral and appended dexCollateral

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
troykessler: it does not exist on evm, thats why I added null there

xeno097: yeah i get just found wierd that in https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0d4fa77350cc3468ce632f3663dea9215003d1a1/typescript/sdk/src/token/contracts.ts#L50 you are mapping the type to the collateral contract and then we have the reader set to null here but not big deal

xeno097 (approved):

PR #6379, opened 2025-05-28
feat(starknet): use legacy from secret value

yjamin:


typescript/infra/src/config/agent/agent.ts
```diff
@@ -246,9 +245,9 @@ export function defaultChainSignerKeyConfig(chainName: ChainName): KeyConfig {
     // Use starknet key for starknet & paradexsepolia
     case ProtocolType.Starknet: {
       if (chainName === 'paradexsepolia') {
-        return { type: AgentSignerKeyType.Starknet, legacy: true };
+        return { type: AgentSignerKeyType.Starknet };
       }
-      return { type: AgentSignerKeyType.Starknet, legacy: false };
+      return { type: AgentSignerKeyType.Starknet };
     }
```
paulbalaji, line 251: no if required, just a `return { type: AgentSignerKeyType.Starknet };`

yjamin: Silly me, that was just he refactoring. Thanks 🙌

PR #6384, opened 2025-05-28
refactor(lander): simplify module and type names

daniel-savu:
### Description

A little overdue, quite painful, but better to do it now than keep waiting. Renames:
- PayloadDispatcher to Dispatcher (as well as its Entrypoint, module name, etc)
- ChainTxAdapter to Adapter (as well as its Factory, module name, etc)
- the `submitter` crate to `lander`
- `SubmitterError` to `LanderError`
- we already had submodules called `dispatcher` and `adapter` inside the `payload_dispatcher` and `chain_tx_adapter` respectively. Those are renamed to `core`


rust/main/lander/src/adapter/chains/sealevel/precursor.rs
```diff
@@ -4,8 +4,8 @@ use solana_sdk::instruction::Instruction as SealevelInstruction;
 
 use hyperlane_sealevel::SealevelTxCostEstimate;
 
-use crate::chain_tx_adapter::chains::sealevel::payload;
-use crate::chain_tx_adapter::chains::sealevel::payload::Instruction;
+use crate::adapter::chains::sealevel::payload;
+use crate::adapter::chains::sealevel::payload::Instruction;
 use crate::payload::FullPayload;
```
ameten, line 9:
```suggestion
use crate::payload::FullPayload;

use super::payload::{self, Instruction};
```

ameten: It will be great to fix it in other places as well.

rust/main/lander/src/dispatcher.rs
```diff
@@ -11,8 +11,8 @@ pub mod test_utils;
 #[cfg(test)]
 mod tests;
 
+pub use core::*;
```
ameten, line 14: I wonder if we want to expose all the types from every module?

For example, [Sealevel adapter](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/45d0ec5d5f68c98f10d9a48a0f7d5a613c0f1aa7/rust/main/lander/src/adapter/chains/sealevel.rs
) module exposes only two types: `SealevelTxAdapter` and `SealevelTxPrecursor`.

ameten:
It will be great to rename `SealevelTxAdapter` to `SealevelAdapter` since `ChainTxAdapter` became `Adapter` and rename adapters for other chains as well.

rust/main/lander/src/error.rs
```diff
@@ -5,7 +5,7 @@ use hyperlane_base::db::DbError;
 use crate::transaction::Transaction;
 
 #[derive(Debug, thiserror::Error)]
-pub enum SubmitterError {
+pub enum LanderError {
```
ameten, line 8: It will be great to find if we use `SubmitterError` in string literals or comments and change it there as well.

ameten (approved):

kamiyaa (approved):

PR #6390, opened 2025-05-29
refactor(signer): use handleTx instead of ethers directly

fernandomg:
### Description

This PR changes the Executor to use `handleTx` instead of ethers' `wait` directly.
Also,
 - removed `rebalancerKey` from Config and related components
 - removed `registry` and extracted it from context

### Related issues

Fixes https://github.com/BootNodeDev/hyperlane-monorepo/issues/65

### Backward compatibility

Yes

### Testing

Unit/e2e

### DIFF
--- a/typescript/cli/src/rebalancer/executor/Executor.ts
+++ b/typescript/cli/src/rebalancer/executor/Executor.ts
@@ -5,13 +5,13 @@ import {
   type ChainMetadata,
   EvmHypCollateralAdapter,
   type InterchainGasQuote,
+  type MultiProvider,
   type Token,
   type TokenAmount,
   type WarpCore,
 } from '@hyperlane-xyz/sdk';
 import { stringifyObject, toWei } from '@hyperlane-xyz/utils';
 
-import type { WriteCommandContext } from '../../context/types.js';
 import { errorRed, log } from '../../logger.js';
 import type { IExecutor } from '../interfaces/IExecutor.js';
 import type { RebalancingRoute } from '../interfaces/IStrategy.js';
@@ -23,7 +23,7 @@ export class Executor implements IExecutor {
     private readonly warpCore: WarpCore,
     private readonly chainMetadata: ChainMap<ChainMetadata>,
     private readonly tokensByChainName: ChainMap<Token>,
-    private readonly context: WriteCommandContext,
+    private readonly multiProvider: MultiProvider,
   ) {}
 
   async rebalance(routes: RebalancingRoute[]) {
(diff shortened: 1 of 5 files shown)


typescript/cli/src/rebalancer/executor/Executor.ts
```diff
@@ -19,10 +20,10 @@ import { type BridgeConfig, getBridgeConfig } from '../utils/bridgeConfig.js';
 export class Executor implements IExecutor {
   constructor(
     private readonly bridges: ChainMap<BridgeConfig>,
-    private readonly rebalancerKey: string,
     private readonly warpCore: WarpCore,
     private readonly chainMetadata: ChainMap<ChainMetadata>,
     private readonly tokensByChainName: ChainMap<Token>,
+    private readonly context: WriteCommandContext,
```
fzavalia, line 26: Instead of the context, just receive the multiprovider given that it is the only thing used from it.

fzavalia (changes requested):

typescript/cli/src/rebalancer/executor/Executor.ts
```diff
@@ -174,9 +177,9 @@ export class Executor implements IExecutor {
     // This is mainly to check that the transaction will not fail before sending them.
     const estimateGasResults = await Promise.allSettled(
       transactions.map(
-        async ({ signer, populatedTx, route, originTokenAmount }) => {
+        async ({ connectedSigner, populatedTx, route, originTokenAmount }) => {
           try {
-            await signer.estimateGas(populatedTx);
+            await connectedSigner.estimateGas(populatedTx);
```
Mo-Hussain, line 182: We can leverage the multipleProvider further here, using https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/767ca002bf1c80802073a8737830ad0a91116aec/typescript/sdk/src/providers/MultiProvider.ts#L385

typescript/cli/src/rebalancer/executor/Executor.ts
```diff
@@ -212,20 +215,23 @@ export class Executor implements IExecutor {
...
-        async ({ signer, populatedTx, route, originTokenAmount }) => {
+        async ({ connectedSigner, populatedTx, route, originTokenAmount }) => {
           log(
             `Sending transaction for route from ${route.origin} to ${
               route.destination
             } for ${originTokenAmount.getDecimalFormattedAmount()} ${originTokenAmount.token.name}`,
           );
           try {
-            const tx = await signer.sendTransaction(populatedTx);
+            const tx = await connectedSigner.sendTransaction(populatedTx);
```
Mo-Hussain, line 225: And here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/767ca002bf1c80802073a8737830ad0a91116aec/typescript/sdk/src/providers/MultiProvider.ts#L406

I think this would mean we don't need to directly use the signer

Mo-Hussain (commented):
A couple comments that probably should have been in the task description

Mo-Hussain (approved):

PR #6399, opened 2025-05-29
feat(rebalancer): improve error handling

fernandomg:
### Description

Modifies the `WrappedError` class, as suggested, to include the `cause`
Also, does use the `util.format()` function to better trace errors.

### Related issues

Closes https://github.com/BootNodeDev/hyperlane-monorepo/issues/73

### Backward compatibility

Yes

### Testing

Manual/Unit/e2e

### DIFF
--- a/typescript/cli/src/tests/warp/warp-rebalancer.e2e-test.ts
+++ b/typescript/cli/src/tests/warp/warp-rebalancer.e2e-test.ts
@@ -381,7 +381,7 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
     });
 
     await startRebalancerAndExpectLog(
-      `Rebalancer error: Validation error: All chains must use the same minAmount type. at "minAmount.type"`,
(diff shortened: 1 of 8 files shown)


typescript/cli/src/tests/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -733,9 +733,7 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
     );
 
     await startRebalancerAndExpectLog(
-      `Could not get rebalance quotes from ${chain3Metadata.name} to ${
-        chain2Metadata.name
-      }, for 5 token: All providers failed on chain unknown for method call and params`,
+      `Could not get rebalance quotes from ${chain3Metadata.name} to ${chain2Metadata.name}, for 5 token`,
```
fernandomg, line 736: Needed to change this expected error, as the `All providers failed on chain unknown for method call and params` is now part of the error and stack trace and not captured in the chunks/lines.

PR #6409, opened 2025-05-30
feat(sdk): switch starknet hooks config

aroralanuk:
### Description

- switched protocol fee to be default and merkle to be required for legal reasons. (we should avoid charging non-zero protocol fee payment as advised by the lawyers)
- changed native token from ETH to STRK bc ETH will not be supported after June

### Drive-by changes

- v0.3.8 -> v0.3.9 of starknet contracts (now on main :) )

### Related issues

None

### Backward compatibility

Yes

### Testing

Live deployment

### DIFF
--- a/typescript/sdk/src/core/StarknetCoreModule.ts
+++ b/typescript/sdk/src/core/StarknetCoreModule.ts
@@ -100,13 +100,7 @@ export class StarknetCoreModule {
       [],
     );
 
-    // 2. Default Hook - A basic hook implementation for message processing
-    const requiredHook = await this.deployer.deployContract(
-      StarknetContractName.HOOK,
-      [],
-    );
-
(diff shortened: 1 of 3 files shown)


typescript/sdk/src/core/StarknetCoreModule.ts
```diff
@@ -101,7 +101,7 @@ export class StarknetCoreModule {
     );
 
     // 2. Default Hook - A basic hook implementation for message processing
```
troykessler, line 103: If protocol fee hook is now the default hook it should be 2. in order, for 3. then the required hook (also comment needs to be updated here (2. Default Hook -> 3. Required Hook)

troykessler (changes requested):

troykessler (approved):

PR #6415, opened 2025-05-30
fix(infra): filter out testnets for check-warp-deploy

paulbalaji:
### Description

fix(infra): filter out testnets for check-warp-deploy

### DIFF
--- a/typescript/infra/scripts/check/check-warp-deploy.ts
+++ b/typescript/infra/scripts/check/check-warp-deploy.ts
@@ -56,18 +56,18 @@ async function main() {
 
   const warpConfigChains = new Set<ChainName>();
   const warpIdsToCheck = await Promise.all(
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/check/check-warp-deploy.ts
```diff
@@ -36,23 +35,39 @@ async function main() {
...
+      const isTestnet = await isTestnetRoute(warpRouteConfig);
+      if (!routesToSkip.includes(warpRouteId) && !isTestnet) {
+        Object.keys(warpRouteConfig).forEach((chain) =>
+          warpConfigChains.add(chain),
+        );
+        return warpRouteId;
+      }
+      return null;
+    }),
+  ).then((results) => results.filter((id) => id !== null));
```
paulbalaji, line 70: maybe this should instead look at the `environment` - and ensure we filter out testnets on mainnet3 but _only_ do testnets on testnet4

PR #6417, opened 2025-05-30
fix: use router domains for movable collateral router

yorhodes:

### DIFF
--- a/solidity/contracts/token/libs/MovableCollateralRouter.sol
+++ b/solidity/contracts/token/libs/MovableCollateralRouter.sol
@@ -2,53 +2,23 @@
 pragma solidity >=0.8.0;
 
 import {Router} from "contracts/client/Router.sol";
+import {FungibleTokenRouter} from "./FungibleTokenRouter.sol";
 import {ValueTransferBridge} from "./ValueTransferBridge.sol";
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 
-import {EnumerableMapExtended, EnumerableMap} from "contracts/libs/EnumerableMapExtended.sol";
-
-import {Context} from "@openzeppelin/contracts/utils/Context.sol";
-import {ContextUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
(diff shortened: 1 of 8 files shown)


solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -67,44 +35,79 @@ abstract contract MovableCollateralRouter is Router {
...
-        _allowedRebalancers.set(rebalancer, 1);
+    modifier onlyAllowedBridge(uint32 domain, ValueTransferBridge bridge) {
+        EnumerableSet.AddressSet storage bridges = _allowedBridges[domain];
+        require(bridges.contains(address(bridge)), "MCR: Not allowed bridge");
+        _;
     }
 
-    function removeRebalancer(address rebalancer) external onlyOwner {
-        _allowedRebalancers.remove(rebalancer);
+    function rebalancers() external view returns (address[] memory) {
```
larryob, line 44: Nit: Change this to `allowedRebalancers` or change the storage variable to `_rebalancers`

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -67,44 +35,79 @@ abstract contract MovableCollateralRouter is Router {
...
-            _allowedRebalancers.set(rebalancers[i], 1);
-        }
+    function allowedBridges(
+        uint32 domain
+    ) external view returns (address[] memory) {
+        return _allowedBridges[domain].values();
+    }
+
+    function setRecipient(uint32 domain, bytes32 recipient) external onlyOwner {
+        _mustHaveRemoteRouter(domain);
```
larryob, line 55: To clarify, the benefit here is just that there is only one set of domains to enumerate, correct? A comment would be nice.

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -67,44 +35,79 @@ abstract contract MovableCollateralRouter is Router {
...
+    }
+
+    function addBridge(
+        uint32 domain,
+        ValueTransferBridge bridge
     ) external onlyOwner {
-        for (uint256 i = 0; i < rebalancers.length; i++) {
-            _allowedRebalancers.remove(rebalancers[i]);
-        }
+        _mustHaveRemoteRouter(domain);
```
larryob, line 67: This check looks redundant. We can only set a recipient on router domains, and when no recipient is set, we default to the remote router. Might prevent setting unusable bridges though.

larryob (approved):
LGTM, left a couple nits.

PR #6425, opened 2025-06-02
fix: Skip SMOL warp route since it fails the check

ameten:
### Description

Skip SMOL warp route since it fails the check

### Backward compatibility

Yes

### Testing

Manual


xeno097 (approved):

Mo-Hussain:
Why is it failing the checks?

PR #6433, opened 2025-06-02
feat(sdk): support MultiSend in Safe TX Submitter

paulbalaji:
### Description

feat(sdk): support MultiSend in Safe TX Submitter

we do this by default in infra all the time and was caught off guard when I saw individual txs being proposed by the cli

### Testing

- used this to propose ousdt txs


typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5GnosisSafeTxSubmitter.ts
```diff
@@ -63,39 +66,43 @@ export class EV5GnosisSafeTxSubmitter implements EV5TxSubmitterInterface {
...
 
-  private async proposeIndividualTransactions(txs: AnnotatedEV5Transaction[]) {
-    const safeTransactions: SafeTransaction[] = [];
-    for (const tx of txs) {
-      const safeTransaction = await this.createSafeTransaction(tx);
-      await this.proposeSafeTransaction(safeTransaction);
-      safeTransactions.push(safeTransaction);
-    }
-    return safeTransactions;
+  public async submit(...txs: AnnotatedEV5Transaction[]): Promise<any> {
```
xeno097, line 103: why not promise void considering that `propose..` returns void?

paulbalaji: actually no it doesn't return void

xeno097 (approved):

ltyu (approved):

PR #6449, opened 2025-06-04
feat: starknet cli v3 txs

troykessler:
### Description

This PR adds support for starknet v3 transactions by exporting `STARKNET_TRANSACTION_VERSION=v3`


yjamin (commented):
I wish this was so easy on the backend as well.

PR #6451, opened 2025-06-04
feat: Override the lowest block height with value from settings

ameten:
### Description

Currently watermark-aware cursor (aka RateLimitedContractSyncCursor) will use watermark retrieved from database. If watermark is not available in the database, it will use value from `from` field in index `field` of chain configuration. It means that the default from configuration is used only when watermark-aware cursor starts indexing from empty database.

We can implement the override functionality so that watermark-aware cursor can skip some block heights if they are not provided by RPC provider.

If value in index.from is higher than the one stored in the database, agent should log a warning and use the value from `index.from` is the lowest block height to index.

### Drive-by changes

- Remove word `agent` from log `Scraper agent starting up...` to make it similar to other agents.

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1735/configurable-lowest-block-height-for-watermark-aware-cursor

### Backward compatibility

Yes

### Testing

Manual

### DIFF
--- a/rust/main/agents/scraper/src/main.rs
+++ b/rust/main/agents/scraper/src/main.rs
@@ -27,7 +27,7 @@ mod store;
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<()> {
     // Logging is not initialised at this point, so, using `println!`
-    println!("Scraper agent starting up...");
+    println!("Scraper starting up...");
 
     agent_main::<Scraper>().await
 }


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -397,9 +397,17 @@ where
         &self,
         index_settings: IndexSettings,
     ) -> Result<Box<dyn ContractSyncCursor<T>>> {
-        let watermark = self.store.retrieve_high_watermark().await.unwrap();
+        let watermark: Option<u32> = self.store.retrieve_high_watermark().await.unwrap();
```
tkporter, line 400: any reason for us to unwrap here instead of `?` ?

tkporter (approved):

rust/main/hyperlane-base/src/contract_sync/mod.rs
ameten: It is a good question. I actually don't want to change this line. `unwrap()` is there already.

I'll check if we can just return with `?`.

ameten: We are handling error at the place where we call this method. So, exiting with `?` is okay. I'll fix it.

PR #6457, opened 2025-06-04
fix: Add changesets for ICA Router and TokenBridge changes

larryob:


.changeset/slow-trees-enjoy.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hyperlane-xyz/core": major
```
yorhodes, line 2: this would be minor but nbd

yorhodes (approved):

.changeset/slow-trees-enjoy.md
larryob: Fixed

PR #6461, opened 2025-06-05
feat(sdk, cli): allow chain names in router configs

xeno097:
### Description

This PR adds the option to configure remote routers/gas and rebalancing bridges using the chain name instead of the domain id in the config

### Drive-by changes

- Fixes all the failing e2e tests in the CLI
- Adds support for the ICA router ISM to fix the tests

### Related issues

- 

### Backward compatibility

- Yes

### Testing

- E2E
- Manual

### DIFF
--- a/solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
+++ b/solidity/contracts/isms/hook/OPL2ToL1CcipReadIsm.sol
@@ -40,6 +40,7 @@ interface OpL2toL1Service {
 abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
     using Message for bytes;
     using TypeCasts for address;
+    using TypeCasts for bytes32;
     using OPL2ToL1Withdrawal for IOptimismPortal.WithdrawalTransaction;
 
     // the OP Portal contract on L1
@@ -70,9 +71,9 @@ abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
         bytes calldata _message
     ) external override returns (bool) {
         if (_isProve(_message)) {
-            _proveWithdrawal(_message.id(), _metadata);
+            _proveWithdrawal(_message, _metadata);
         } else {
-            _finalizeWithdrawal(_message.id(), _metadata);
+            _finalizeWithdrawal(_message, _metadata);
         }
 
         return true;
@@ -83,7 +84,7 @@ abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
     ) internal view virtual returns (bool);
 
     function _proveWithdrawal(
-        bytes32 messageId,
+        bytes calldata _message,
         bytes calldata _metadata
     ) internal {
         (
@@ -102,7 +103,12 @@ abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
             );
 
         require(
-            _tx.proveMessageId() == messageId,
+            _tx.sender == _message.sender().bytes32ToAddress(),
+            "OPL2ToL1CcipReadIsm: sender mismatch"
+        );
+
+        require(
+            _tx.proveMessageId() == _message.id(),
             "OPL2ToL1CcipReadIsm: prove message id mismatch"
         );
 
@@ -125,7 +131,7 @@ abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
     ) internal view virtual returns (bool);
 
     function _finalizeWithdrawal(
-        bytes32 messageId,
+        bytes calldata _message,
         bytes calldata _metadata
     ) internal {
         IOptimismPortal.WithdrawalTransaction memory _tx = abi.decode(
@@ -134,7 +140,12 @@ abstract contract OPL2ToL1CcipReadIsm is AbstractCcipReadIsm {
         );
 
         require(
-            _tx.finalizeMessageId() == messageId,
+            _tx.sender == _message.sender().bytes32ToAddress(),
+            "OPL2ToL1CcipReadIsm: sender mismatch"
+        );
+
+        require(
+            _tx.finalizeMessageId() == _message.id(),
             "OPL2ToL1CcipReadIsm: finalize message id mismatch"
         );
 
--- a/solidity/contracts/libs/CctpMessage.sol
+++ b/solidity/contracts/libs/CctpMessage.sol
@@ -19,13 +19,40 @@ import {TypedMemView} from "./TypedMemView.sol";
 
 // @dev CCTP Message version 1
 // @dev copied from https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/messages/Message.sol
+
+/**
+ * @title Message Library
+ * @notice Library for formatted messages used by Relayer and Receiver.
+ *
+ * @dev The message body is dynamically-sized to support custom message body
+ * formats. Other fields must be fixed-size to avoid hash collisions.
+ * Each other input value has an explicit type to guarantee fixed-size.
+ * Padding: uintNN fields are left-padded, and bytesNN fields are right-padded.
+ *
+ * Field                 Bytes      Type       Index
+ * version               4          uint32     0
+ * sourceDomain          4          uint32     4
+ * destinationDomain     4          uint32     8
+ * nonce                 8          uint64     12
+ * sender                32         bytes32    20
+ * recipient             32         bytes32    52
+ * destinationCaller     32         bytes32    84
+ * messageBody           dynamic    bytes      116
+ *
+ **/
 library CctpMessage {
     using TypedMemView for bytes;
     using TypedMemView for bytes29;
 
     // Indices of each field in message
+    uint8 private constant VERSION_INDEX = 0;
     uint8 private constant SOURCE_DOMAIN_INDEX = 4;
+    uint8 private constant DESTINATION_DOMAIN_INDEX = 8;
     uint8 private constant NONCE_INDEX = 12;
+    uint8 private constant SENDER_INDEX = 20;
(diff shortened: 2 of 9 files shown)


typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -244,7 +273,7 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
     } catch {
       ismType = IsmType.ROUTING;
       this.logger.debug(
-        'Error accessing mailbox property, implying this is not a fallback routing ISM.',
+        'Error accessing mailbox property, implying this is not a fallback routing ISM ism.',
```
paulbalaji, line 276:
```suggestion
        'Error accessing mailbox property, implying this is not a fallback routing ISM.',
```

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -257,6 +286,55 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
...
+    contractInstance: AbstractRoutingIsm,
+    addressDeriveFunc: (domain: ethers.BigNumberish) => Promise<string>,
+    deriveConfig: false,
+  ): Promise<ChainMap<string>>;
+  private async deriveRemoteIsmConfigs(
+    domainIds: ethers.BigNumber[],
+    contractInstance: AbstractRoutingIsm,
+    addressDeriveFunc: (domain: ethers.BigNumberish) => Promise<string>,
+    deriveConfig: boolean,
+  ): Promise<ChainMap<IsmConfig>> {
```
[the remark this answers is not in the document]
xeno097: IsmConfig allows string values so no need to make it more explicit than it is

typescript/sdk/src/router/types.ts
```diff
@@ -97,12 +98,32 @@ export const ForeignDeploymentConfigSchema = z.object({
...
+  routerMap: Record<RemoteRouterDomainOrChainName, T>,
+): Record<number, T> {
+  return Object.fromEntries(
+    Object.entries(routerMap).map(([domainIdOrChainName, value]) => {
+      if (isNumeric(domainIdOrChainName)) {
+        return [parseInt(domainIdOrChainName), value];
+      }
+
+      const meta = multiProvider.getChainMetadata(domainIdOrChainName);
+      return [meta.domainId, value];
```
yorhodes, line 117:
```suggestion
    Object.entries(routerMap).map(([chain, value]) => 
      [multiProvider.getChainMetadata(chain).domainId, value]
```
I think this is equivalent

typescript/sdk/src/token/EvmERC20WarpModule.hardhat-test.ts
```diff
@@ -876,9 +877,7 @@ describe('EvmERC20WarpHyperlaneModule', async () => {
 
         await assertAllowedRebalancers(evmERC20WarpModule, expectedRebalancers);
       });
-    }
 
-    for (const tokenType of movableCollateralTypes) {
```
yorhodes, line 881: why remove this?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -156,6 +157,7 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
           this.provider,
         ).allowedRebalancers();
 
+        // Set to undefined if the object is empty because might trigger false positives in the checker
```
yorhodes, line 160: this makes sense but it might be more clear if we treat undefined and empty as the same in the diff checker?

yorhodes: rather than pruning from the derived config

typescript/sdk/src/router/types.ts
```diff
@@ -97,12 +98,32 @@ export const ForeignDeploymentConfigSchema = z.object({
   foreignDeployment: z.string().optional(),
 });
 
-export const RemoteRouterDomain = z.string();
+export const RemoteRouterDomainOrChainNameSchema = z.string().or(z.number());
+export type RemoteRouterDomainOrChainName = z.infer<
+  typeof RemoteRouterDomainOrChainNameSchema
+>;
```
yorhodes, line 104: this is a bit verbose
do we. not already have this on the ChainMetadata schema?

yorhodes (approved):

typescript/sdk/src/ism/utils.ts
```diff
@@ -476,7 +477,7 @@ export async function routingModuleDelta(
 async function domainRoutingModuleDelta(
   destination: ChainName,
   moduleAddress: Address,
-  config: DomainRoutingIsmConfig,
+  config: DomainRoutingIsmConfig | InterchainAccountRouterIsm,
```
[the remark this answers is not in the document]
xeno097: Because it includes also the Amount routing ism type that is used separately before this func is called

typescript/sdk/src/ism/types.ts
```diff
@@ -312,6 +329,10 @@ export const RoutingIsmConfigSchema: z.ZodSchema<RoutingIsmConfig> = z.lazy(
         type: z.literal(IsmType.FALLBACK_ROUTING),
         domains: z.record(IsmConfigSchema),
       }),
+      OwnableSchema.extend({
+        type: z.literal(IsmType.INTERCHAIN_ACCOUNT_ROUTING),
+        isms: z.record(ZHash),
```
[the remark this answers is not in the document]
xeno097: Yes but i need to define it here so that the `DerivedIsmConfig` type automatically loads this definition too as it is defined as:

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/14c325a19d53359f927aae3a0d3e00f07b419f37/typescript/sdk/src/ism/types.ts#L205

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -257,6 +286,55 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
...
+    addressDeriveFunc: (domain: ethers.BigNumberish) => Promise<string>,
+    deriveConfig: true,
+  ): Promise<ChainMap<IsmConfig>>;
+  private async deriveRemoteIsmConfigs(
+    domainIds: ethers.BigNumber[],
+    contractInstance: AbstractRoutingIsm,
+    addressDeriveFunc: (domain: ethers.BigNumberish) => Promise<string>,
+    deriveConfig: false,
+  ): Promise<ChainMap<string>>;
+  private async deriveRemoteIsmConfigs(
```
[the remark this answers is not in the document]
xeno097: I added overloads so that if the `deriveConfig` var is passed explicitly as true or false we will have correct type inference on the return type instead of just `ChainMap<IsmConfig>>` because in the case of the ica router ism we can't derive the remote isms directly as they are on a different chain compared to for example the domain or fallback routing isms

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -218,24 +226,45 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
...
-      if (!chainName) {
-        this.logger.warn(
-          `Unknown domain ID ${domainId}, skipping domain configuration`,
-        );
-        return;
-      }
-      const module = this.messageContext
-        ? await defaultFallbackIsmInstance.route(this.messageContext.message)
-        : await defaultFallbackIsmInstance.module(domainId);
-      domains[chainName] = await this.deriveIsmConfig(module);
```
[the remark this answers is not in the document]
xeno097: Yeah i haven't completely removed it just moved into the `deriveRemoteIsmConfigs` function to reuse the logic to derive the related isms to the current one, see:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/14c325a19d53359f927aae3a0d3e00f07b419f37/typescript/sdk/src/ism/EvmIsmReader.ts#L320-L322

typescript/cli/src/tests/warp/warp-init.e2e-test.ts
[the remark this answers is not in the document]
xeno097: the problem is not in the enum themselves but the order how these are shown to the user on the prompts, so as soon as a new token type is added, as that order might change, this might break. I'll check in a follow up if there is a way to make this a bit more clever

typescript/sdk/src/ism/utils.ts
```diff
@@ -504,8 +505,13 @@ async function domainRoutingModuleDelta(
     if (mailbox && !eqAddress(mailboxAddress, mailbox)) delta.mailbox = mailbox;
   }
 
+  const ismByDomainName =
+    config.type === IsmType.INTERCHAIN_ACCOUNT_ROUTING
+      ? config.isms
+      : config.domains;
```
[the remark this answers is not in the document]
xeno097: The field that stores the remote isms on the ica router is called isms instead of domains so to get the differences in ism configurations the code needs to read the correct field, on the ts code side i could have still called it domains too but i wanted to keep consistency between the contract and the type in the sdk.

typescript/sdk/src/token/EvmERC20WarpModule.hardhat-test.ts
xeno097: i have just grouped the tests that had the same loop and put them all together in a single loop

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/14c325a19d53359f927aae3a0d3e00f07b419f37/typescript/sdk/src/token/EvmERC20WarpModule.hardhat-test.ts#L856-L1123

typescript/sdk/src/router/types.ts
xeno097: unfortunately noup, there we just allow wither for an int on the domain Id field
image

PR #6464, opened 2025-06-05
fix: add sanity range check for starknet indexer

yjamin:
### Description
This PR adds an extra sanity check for the starknet indexer methods. It first checks whether or not the current block height is included in the given index range.

The method we use to retrieve Starknet events doesn't check for a valid block range and will simply return an empty array for unkown / future block ranges. Sometimes it happens, that we hit a node that hasn't actually mined the latest block yet and therefore only returns an empty array instead of the actual events that might be in the range. 

This mainly impacts scraper deliveries and IGP indexing, as `merkle_tree_hook` & `dispatch` are sequence aware.

### Related issues
[Linear Issue](https://linear.app/hyperlane-xyz/issue/ENG-1702/scraper-missing-events)

### Testing
Local testing & E2E

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for block range queries to ensure requested ranges do not exceed the current blockchain height, preventing silent failures and providing clearer error messages for out-of-range requests.

### DIFF
--- a/rust/main/chains/hyperlane-starknet/src/indexer.rs
+++ b/rust/main/chains/hyperlane-starknet/src/indexer.rs
@@ -262,6 +262,17 @@ async fn fetch_logs_in_range<T>(
 where
     T: std::fmt::Debug,
 {
+    // sanity check the range, because the provider doesn't return an error if the range is invalid
+    // and only returns an empty list
+    let current_block = get_block_height_u32(provider, &ReorgPeriod::None).await?;
+    if !range.contains(&current_block) {
+        return Err(HyperlaneStarknetError::Other(format!(
+            "range {:?} is not valid for current block {}",
+            range, current_block
+        ))
+        .into());
+    }
+
     let key = get_selector_from_name(key)
         .map_err(|_| HyperlaneStarknetError::from_other("get selector cannot fail"))?;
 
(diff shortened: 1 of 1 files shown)


rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -262,6 +262,17 @@ async fn fetch_logs_in_range<T>(
 where
     T: std::fmt::Debug,
 {
+    // sanity check the range, because the provider doesn't return an error if the range is invalid
```
tkporter, line 265: So I understand correctly, this won't fully protect us but adds another layer of defense? as we already do a "get the latest block with one RPC, then try to index up to that latest block in the next RPC" and the issue is that the provider will load balance to different underlying nodes for the two RPCs, and we can end up in the same situation here where this get block height RPC is still load balanced to the node that's ahead but the getLogs RPC is still sent to a node that's behind? But the idea is this can be less likely to happen by doing the block height RPCs twice

rust/main/chains/hyperlane-starknet/src/indexer.rs
```diff
@@ -262,6 +262,17 @@ async fn fetch_logs_in_range<T>(
...
+    // and only returns an empty list
+    let current_block = get_block_height_u32(provider, &ReorgPeriod::None).await?;
+    if *range.start() > current_block || *range.end() > current_block {
```
tkporter, line 268: I'd expect that start is always <= end but I guess doesn't hurt to check both

rust/main/chains/hyperlane-starknet/src/indexer.rs
tkporter: sg, happy to merge, it will increase our RPCs made but that's probably fine

tkporter (approved):

PR #6468, opened 2025-06-05
refactor(lander): Rename TransactionId, PayloadId and PendingInclusion

ameten:
### Description

* Rename PendingInclusion into Pending
* Rename TransactionId into TransactionUuid
* Rename PayloadId into PayloadUuid

### Backward compatibility

Yes

### Testing

Current tests pass

### DIFF
--- a/rust/main/lander/src/adapter/chains/ethereum/nonce/state/tests.rs
+++ b/rust/main/lander/src/adapter/chains/ethereum/nonce/state/tests.rs
@@ -201,7 +201,7 @@ fn dummy_tx() -> Transaction {
     };
 
     Transaction {
-        id: Default::default(),
+        uuid: Default::default(),
         tx_hashes: vec![],
         vm_specific_data: VmSpecificTxData::Evm(precursor),
         payload_details: vec![],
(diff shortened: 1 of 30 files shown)


rust/main/lander/src/adapter/chains/ethereum/nonce/state/tests.rs
```diff
@@ -36,7 +36,7 @@ async fn test_update_nonce_status_pending_inclusion() {
     precursor.tx.set_nonce(nonce);
 
     state
-        .update_nonce_status(&tx, &TransactionStatus::PendingInclusion)
```
daniel-savu, line 39: IMO having `Inclusion` in the name is quite crucial, since it makes the difference between paying enough gas and not doing so; it also closely relates to the "Inclusion Stage" name. `Pending` sounds very ambiguous

ameten: The main reason to rename is that `PendingInclusion` state transitions into `Mempool` and after that into `Included`. So, there are two stages in "including" transaction: `Pending` and `Mempool`. I am happy to rename in back though.

daniel-savu: I'd rather remove the Mempool stage but keep PendingInclusion, agree it doesn't add much

daniel-savu (approved):

PR #6471, opened 2025-06-05
fix: check OP withdrawal tx matches hyperlane messages

yorhodes:
### Description

- Enforces that the prove message is processed IFF the corresponding withdrawal is proven
- Enforces that the finalize message is processed IFF the corresponding withdrawal is finalized

Note that it is still possible for the withdrawal to be proven and finalized by another actor. In this case the messages will verify this status and then be marked as delivered.

### Backward compatibility

Yes

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/mock/MockOptimism.sol
+++ b/solidity/contracts/mock/MockOptimism.sol
@@ -67,12 +67,6 @@ contract MockOptimismPortal is IOptimismPortal {
     function finalizeWithdrawalTransaction(
         WithdrawalTransaction memory _tx
     ) external {
-        CallLib.Call memory call = CallLib.Call(
-            TypeCasts.addressToBytes32(_tx.target),
-            _tx.value,
-            _tx.data
-        );
-        CallLib.call(call);
         bytes32 withdrawalHash = OPL2ToL1Withdrawal.hashWithdrawal(_tx);
         _finalizedWithdrawals[withdrawalHash] = true;
     }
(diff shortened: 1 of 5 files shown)


solidity/contracts/mock/MockOptimism.sol
```diff
@@ -67,12 +67,6 @@ contract MockOptimismPortal is IOptimismPortal {
     function finalizeWithdrawalTransaction(
         WithdrawalTransaction memory _tx
     ) external {
-        CallLib.Call memory call = CallLib.Call(
-            TypeCasts.addressToBytes32(_tx.target),
-            _tx.value,
-            _tx.data
-        );
-        CallLib.call(call);
```
yorhodes, line 75: see https://github.com/ethereum-optimism/optimism/blob/2e647210882d961f04055e656590d90ad98c9934/packages/contracts-bedrock/src/universal/StandardBridge.sol#L152-L166

yorhodes: > Note that the recipient will not be triggered with this data, but it will be emitted and can be used to identify the transaction.

solidity/test/isms/OPL2ToL1CcipReadIsm.t.sol
```diff
@@ -169,6 +177,60 @@ contract OpL1V1NativeTokenBridgeTest is Test {
...
+        );
+
+        bytes memory data = OPL2ToL1Withdrawal.encodeData(
+            bytes32(0),
+            bytes32(0)
+        );
+        IOptimismPortal.WithdrawalTransaction
+            memory withdrawalTx = _getDummyWithdrawalTx(0, amount, data);
+
+        vm.deal(address(portal), amount);
```
larryob, line 225: You can just `deal` instead of `vm.deal` thanks to forge-std

larryob (approved):
LGTM

PR #6487, opened 2025-06-09
fix: enforce withdrawal tx sender matches message sender

yorhodes:
### Description

> In PR #6471, the metadata of the OptimismPortal.WithdrawalTransaction was modified to include two Hyperlane message IDs (proveMessageId and finalizeMessageId). However, this patch allows an attacker to create a separate OP to L1 withdrawal transaction with the same pair of message IDs in the metadata, and execute OPL2ToL1CcipReadIsm.verify() using that forged metadata. In other words, it is possible to use the metadata of a different withdrawal transaction instead of the legitimate one. To address this issue, OPL2ToL1CcipReadIsm.verify() should verify that the _tx.sender() from IOptimismPortal.WithdrawalTransaction matches the sender() of the _message

Fixes this issue by enforcing message sender matches tx sender.

### Backward compatibility

No, requires new deployments

### Testing

Unit Tests

### DIFF
--- a/solidity/test/isms/OPL2ToL1CcipReadIsm.t.sol
+++ b/solidity/test/isms/OPL2ToL1CcipReadIsm.t.sol
@@ -15,7 +15,7 @@ import {LightTestRecipient} from "../../contracts/test/LightTestRecipient.sol";
 import {TokenRouter} from "../../contracts/token/libs/TokenRouter.sol";
 import {OPL2ToL1Withdrawal} from "../../contracts/libs/OPL2ToL1Withdrawal.sol";
 
-contract OpL1V1NativeTokenBridgeTest is Test {
+contract OPL1V1NativeTokenBridgeTest is Test {
     using TypeCasts for address;
     using TokenMessage for bytes;
     using OPL2ToL1Withdrawal for IOptimismPortal.WithdrawalTransaction;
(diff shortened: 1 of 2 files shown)


solidity/test/isms/OPL2ToL1CcipReadIsm.t.sol
```diff
@@ -231,6 +252,83 @@ contract OpL1V1NativeTokenBridgeTest is Test {
...
+
+        vm.expectRevert("OPL2ToL1CcipReadIsm: sender mismatch");
+        ism.verify(metadata, message);
+    }
+
+    function test_verify_revertsWhen_finalizeSenderMismatched(
+        bytes32 _recipient,
+        uint256 amount
+    ) public {
+        vm.assume(amount > 0);
```
larryob, line 296: Do we need to assume this?

solidity/test/isms/OPL2ToL1CcipReadIsm.t.sol
```diff
@@ -231,6 +252,83 @@ contract OpL1V1NativeTokenBridgeTest is Test {
...
+
+        IOptimismPortal.WithdrawalTransaction
+            memory withdrawalTx = _getDummyWithdrawalTx(
+                nonce,
+                amount,
+                badSender,
+                data
+            );
+
+        vm.deal(address(portal), amount);
```
larryob, line 323: `deal` is a bit more concise.

larryob (approved):
LGTM

PR #6386, opened 2025-05-28
fix: CCTP token bridge should enforce corresponding messages

yorhodes:
### Description

- Enforce CCTP source and nonce match hyperlane message

### Drive-by changes

- Removes CCTP V2 support 
- Removes forge script (in favor of TS deploy tooling)

### Backward compatibility

Yes

### Testing

Unit Tests

PR #6488, opened 2025-06-09
fix: enforce CCTP message matches hyperlane message

yorhodes:
### Description

> In #6386 PR, the patch still allows a 1:N mapping between CCTP messages and Hyperlane messages. For example, verify() does not check whether the length of _hyperlaneMessage.body().metadata() is exactly 8 bytes. If it is longer than 8 bytes, it will automatically be truncated to bytes8, discarding the trailing bytes. This allows malformed metadata values to pass through verify() as if they were valid. Additionally, other parts of _hyperlaneMessage.body() such as amount, recipient, and even _hyperlaneMessage.sender() can be manipulated.

Fixes this issue by checking all fields match.

### Backward compatibility

Requires new CCTP deploy

### Testing

Unit Tests

### DIFF
--- a/solidity/contracts/libs/CctpMessage.sol
+++ b/solidity/contracts/libs/CctpMessage.sol
@@ -19,13 +19,40 @@ import {TypedMemView} from "./TypedMemView.sol";
 
 // @dev CCTP Message version 1
 // @dev copied from https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/messages/Message.sol
+
+/**
+ * @title Message Library
+ * @notice Library for formatted messages used by Relayer and Receiver.
+ *
+ * @dev The message body is dynamically-sized to support custom message body
+ * formats. Other fields must be fixed-size to avoid hash collisions.
+ * Each other input value has an explicit type to guarantee fixed-size.
+ * Padding: uintNN fields are left-padded, and bytesNN fields are right-padded.
+ *
+ * Field                 Bytes      Type       Index
+ * version               4          uint32     0
+ * sourceDomain          4          uint32     4
+ * destinationDomain     4          uint32     8
+ * nonce                 8          uint64     12
+ * sender                32         bytes32    20
+ * recipient             32         bytes32    52
+ * destinationCaller     32         bytes32    84
+ * messageBody           dynamic    bytes      116
+ *
+ **/
 library CctpMessage {
     using TypedMemView for bytes;
     using TypedMemView for bytes29;
 
     // Indices of each field in message
+    uint8 private constant VERSION_INDEX = 0;
     uint8 private constant SOURCE_DOMAIN_INDEX = 4;
+    uint8 private constant DESTINATION_DOMAIN_INDEX = 8;
     uint8 private constant NONCE_INDEX = 12;
+    uint8 private constant SENDER_INDEX = 20;
+    uint8 private constant RECIPIENT_INDEX = 52;
+    uint8 private constant DESTINATION_CALLER_INDEX = 84;
+    uint8 private constant MESSAGE_BODY_INDEX = 116;
 
     /**
      * @notice Returns formatted (packed) message with provided fields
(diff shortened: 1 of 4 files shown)


solidity/contracts/libs/CctpMessage.sol
```diff
@@ -19,13 +19,40 @@ import {TypedMemView} from "./TypedMemView.sol";
 
 // @dev CCTP Message version 1
 // @dev copied from https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/messages/Message.sol
+
+/**
```
larryob, line 23: Was this taken from a different commit than the one in the comment above (https://github.com/circlefin/evm-cctp-contracts/blob/release-2025-03-11T143015/src/messages/Message.sol)?

solidity/test/token/TokenBridgeCctp.t.sol
```diff
@@ -312,6 +349,143 @@ contract TokenBridgeCctpTest is Test {
...
+        tbDestination.verify(metadata, message);
+    }
+
+    function test_verify_revertsWhen_invalidRecipient() public {
+        bytes32 recipient = user.addressToBytes32();
+        Quote[] memory quote = tbOrigin.quoteTransferRemote(
+            destination,
+            recipient,
+            amount
+        );
```
larryob, line 392: Nit: Throw some of these configuration steps into some internal helper functions. We can follow up and do this later.

solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -21,11 +21,14 @@ interface CctpService {
         returns (bytes memory cctpMessage, bytes memory attestation);
 }
 
+uint256 constant CCTP_TOKEN_BRIDGE_MESSAGE_LEN = 32 + 32 + 8; // 72 bytes
```
larryob, line 24: A comment would be great here. This is recipient, amount, and nonce.

solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -137,8 +140,28 @@ contract TokenBridgeCctp is HypERC20Collateral, AbstractCcipReadIsm {
             (bytes, bytes)
         );
 
+        bytes calldata tokenMessage = _hyperlaneMessage.body();
+        require(
+            tokenMessage.length == CCTP_TOKEN_BRIDGE_MESSAGE_LEN,
```
larryob, line 145: I think we can add one internal  `_checkLen(tokenMessage)` function that reverts if we don't have the right length. This would replace the assert on line 214.

solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -187,6 +210,8 @@ contract TokenBridgeCctp is HypERC20Collateral, AbstractCcipReadIsm {
             outboundAmount,
             abi.encodePacked(nonce)
         );
+        // sanity check: should only happen if the implementation changes
+        assert(_tokenMessage.length == CCTP_TOKEN_BRIDGE_MESSAGE_LEN);
```
larryob, line 214: See comment on line 145

solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -137,8 +140,28 @@ contract TokenBridgeCctp is HypERC20Collateral, AbstractCcipReadIsm {
...
+        );
+
         bytes29 originalMsg = TypedMemView.ref(cctpMessage, 0);
 
+        bytes32 sourceSender = originalMsg._sender();
+        require(sourceSender == _hyperlaneMessage.sender(), "Invalid sender");
+
+        bytes29 burnMessage = originalMsg._messageBody();
+        require(
+            TokenMessage.amount(tokenMessage) == burnMessage._getAmount(),
```
larryob, line 156: Can we add `TokenMessage` functions to `bytes` with `using TokenMessage for bytes`?

larryob (approved):
LGTM

solidity/contracts/token/TokenBridgeCctp.sol
yorhodes: this wasnt possible because `recipient` is ambiguous for `Message` and `TokenMessage`

solidity/contracts/libs/CctpMessage.sol
yorhodes: nope, I just copied the whole implementation

solidity/contracts/token/TokenBridgeCctp.sol
yorhodes: one is calldata and one is memory so unfortunately solidity doesnt let us share logic easily unless we memcopy

solidity/test/token/TokenBridgeCctp.t.sol
yorhodes: done

PR #6490, opened 2025-06-09
fix: Revert in TokenBridgeCctp if parent initialize is called

larryob:
### Description

For safety, we want to make sure that each contract only has one initializer.

### DIFF
--- a/solidity/contracts/token/TokenBridgeCctp.sol
+++ b/solidity/contracts/token/TokenBridgeCctp.sol
@@ -90,7 +90,7 @@ contract TokenBridgeCctp is HypERC20Collateral, AbstractCcipReadIsm {
         address _interchainSecurityModule,
         address _owner
     ) public override {
-        revert("Only one initialize() function is allowed");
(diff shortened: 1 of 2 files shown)


solidity/contracts/token/TokenBridgeCctp.sol
```diff
@@ -85,6 +85,14 @@ contract TokenBridgeCctp is HypERC20Collateral, AbstractCcipReadIsm {
         wrappedToken.approve(address(tokenMessenger), type(uint256).max);
     }
 
+    function initialize(
+        address _hook,
+        address _interchainSecurityModule,
+        address _owner
+    ) public override {
+        revert("Only one initialize() function is allowed");
```
yorhodes, line 93: nit: make it clear which initialize should be called

yorhodes (approved):

solidity/contracts/token/TokenBridgeCctp.sol
larryob: Done

PR #6494, opened 2025-06-09
feat: reduce logs

kamiyaa:
### Description

Reduce logs and log levels

### Related issues

- fixes https://linear.app/hyperlane-xyz/issue/ENG-1748/reduce-agent-logs

### Testing
- update e2e to look for different logs

### DIFF
--- a/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/aggregation.rs
@@ -4,7 +4,7 @@ use futures_util::future::join_all;
 
 use derive_new::new;
 use itertools::{Either, Itertools};
-use tracing::{debug, info, instrument};
+use tracing::{debug, info};
 use {hyperlane_base::cache::FunctionCallCache, tracing::warn};
 
 use hyperlane_core::{
@@ -288,7 +288,6 @@ impl AggregationIsmMetadataBuilder {
 
 #[async_trait]
 impl MetadataBuilder for AggregationIsmMetadataBuilder {
-    #[instrument(err, skip(self, message), ret)]
     #[allow(clippy::blocks_in_conditions)] // TODO: `rustc` 1.80.1 clippy issue
     async fn build(
         &self,
--- a/rust/main/agents/relayer/src/msg/metadata/ccip_read/mod.rs
+++ b/rust/main/agents/relayer/src/msg/metadata/ccip_read/mod.rs
@@ -13,7 +13,7 @@ use reqwest::{header::CONTENT_TYPE, Client};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use sha3::{digest::Update, Digest, Keccak256};
-use tracing::{info, instrument, warn};
+use tracing::{info, warn};
 
 use hyperlane_core::{
     utils::bytes_to_hex, CcipReadIsm, HyperlaneMessage, HyperlaneSignerExt, RawHyperlaneMessage,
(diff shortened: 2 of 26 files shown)


rust/main/agents/relayer/src/msg/metadata/aggregation.rs
```diff
@@ -288,7 +288,6 @@ impl AggregationIsmMetadataBuilder {
 
 #[async_trait]
 impl MetadataBuilder for AggregationIsmMetadataBuilder {
-    #[instrument(err, skip(self, message), ret)]
```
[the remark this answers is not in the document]
kamiyaa: I'm cool with keeping it :+1:

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -43,7 +43,6 @@ impl OpQueue {
     }
 
     /// Pop multiple elements at once from the queue and update metrics
-    #[instrument(skip(self), fields(queue_label=%self.queue_metrics_label), level = "debug")]
```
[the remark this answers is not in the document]
kamiyaa: > 2. omit the bodies, which are pretty large, e.g. https://cloudlogging.app.goo.gl/Dh31SK1EiXWQnBuD7

Yep, I've updated `HyperlaneMessage` to just Debug print its ID

rust/main/agents/relayer/src/msg/metadata/ccip_read/mod.rs
```diff
@@ -167,7 +167,6 @@ impl CcipReadIsmMetadataBuilder {
 
 #[async_trait]
 impl MetadataBuilder for CcipReadIsmMetadataBuilder {
-    #[instrument(err, skip(self, message, _params))]
```
tkporter, line 170: could we add all the `build` ones back too? as these are just for `err`s and not `ret`s, they likely don't contribute much too

rust/main/hyperlane-core/src/types/message.rs
```diff
@@ -58,18 +57,7 @@ impl Default for HyperlaneMessage {
...
-            self.id(),
-            self.version,
-            self.nonce,
-            fmt_domain(self.origin),
-            fmt_address_for_domain(self.origin, self.sender),
-            fmt_domain(self.destination),
-            fmt_address_for_domain(self.destination, self.recipient),
-            hex::encode(&self.body)
-        )
+        write!(f, "HyperlaneMessage {{ id: {:?} }}", self.id(),)
```
tkporter, line 60: can we keep everything but body and version actually? the others are decently useful and we'll probably cut down logging by at least 50%

tkporter: Also I'm a little wary of removing the body for all Debug logs we do - I was hoping we could certainly remove it for the "Popped opqueue" logs but that elsewhere we'd still have it, but happy to give it a try and see what happens. My main concern is just that if we don't have scraper support everywhere it's useful to have the body, but maybe this isn't such a big concern.

If there's an easy way to do what I'm describing where we just omit the body for the popped opqueue log, then that's probably preferable though

kamiyaa: Updated back to original Debug message minus version and body.
Changed "Popped OpQueue operations" to only show the `operation.id()` instead

rust/main/agents/relayer/src/msg/op_queue.rs
```diff
@@ -59,7 +59,7 @@ impl OpQueue {
         if !popped.is_empty() {
             debug!(
                 queue_label = %self.queue_metrics_label,
-                operations = ?popped,
+                operations = ?popped.iter().map(|p| p.id()).collect::<Vec<_>>(),
```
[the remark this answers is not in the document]
kamiyaa: yes :+1:

PR #6497, opened 2025-06-10
fix: make starknet signer key optional

yjamin:
### Description
Make signer optional for starknet protocol chains

## Summary by CodeRabbit

- **New Features**
  - Added a default signer that is automatically used when no signer is provided.

- **Refactor**
  - Updated various components to accept an optional signer instead of requiring one.
  - Simplified account creation and initialization by handling the absence of a signer gracefully.

- **Bug Fixes**
  - Improved handling of cases where signer information is missing, reducing errors and improving robustness.


aroralanuk (changes requested):
We need a key for submitting messages regardless for the relayer so we cannot make it optional and we have a dummy for the validator.

yjamin:
> We need a key for submitting messages regardless for the relayer so we cannot make it optional and we have a dummy for the validator.

I know. This just prevents the agents from crashing. We have the same logic in place for cosmos & ethereum.

PR #6500, opened 2025-06-10
feat: validator overwrite and log any checkpoints that don't match its current in-memory version

kamiyaa:
### Description

- warn if checkpoints don't match

### Related issues

 - fixes https://github.com/hyperlane-xyz/issues/issues/1377
 - fixes https://linear.app/hyperlane-xyz/issue/ENG-1613/validators-should-overwrite-and-log-any-checkpoints-that-dont-match

### Tests

 - added unittests

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced checkpoint submission logic to avoid overwriting when signatures match and to warn when they differ.  
- **Tests**
  - Added tests validating checkpoint submission behavior for matching and differing signatures.

### DIFF
--- a/rust/main/agents/validator/src/submit.rs
+++ b/rust/main/agents/validator/src/submit.rs
@@ -363,9 +363,8 @@ impl ValidatorSubmitter {
             }
         }
 
-        let signed_checkpoint = self.sign_checkpoint(checkpoint).await?;
-
         let start = Instant::now();
+        let signed_checkpoint = self.sign_checkpoint(checkpoint).await?;
         tracing::trace!(
             elapsed=?start.elapsed(),
             "Signed checkpoint",


rust/main/agents/validator/src/submit.rs
```diff
@@ -343,13 +345,27 @@ impl ValidatorSubmitter {
...
+                    new_signer = ?signer,
+                    "Checkpoint already submitted, but with different values, overwriting"
+                );
+            }
         }
 
-        let start = Instant::now();
         let signed_checkpoint = self.sign_checkpoint(checkpoint).await?;
+
+        let start = Instant::now();
```
tkporter, line 368: can you put this back before the `sign_checkpoint` call? the intent was to get the duration it takes to sign

tkporter (approved):

PR #6503, opened 2025-06-11
chore: add custom hook to remote transfer

troykessler:
### Description

This PR upgrades the cosmos-sdk and cosmos-types to the latest Cosmos Hyperlane Module v1.0.1. Furthermore, the igp hook address is added in the populateRemoteTransfer method.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

-

## Summary by CodeRabbit

- **New Features**
  - Added support for custom hook parameters in remote transfer requests, allowing for greater flexibility when quoting and executing remote transfers.

- **Bug Fixes**
  - Updated service and Docker image versions to ensure compatibility with the latest release.

- **Tests**
  - Enhanced remote transfer tests to include custom hook parameters.

- **Chores**
  - Updated configuration files to use the latest image and branch versions.


typescript/sdk/src/token/adapters/CosmosModuleTokenAdapter.ts
```diff
@@ -224,6 +227,8 @@ export class CosmNativeHypCollateralAdapter
     if (!params.interchainGas) {
       params.interchainGas = await this.quoteTransferRemoteGas(
         params.destination,
+        '',
+        params.customHook,
       );
```
yjamin, line 232: Correct me if I'm wrong, but I don't think this will end up being used in the warp up. The custom hook is only every passed in the populate remote transfer function. This function calls following to get the fee:

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/915157fc089a2635f0933c1dda0ab106dfca2706/typescript/sdk/src/warp/WarpCore.ts#L128

I believe the custom hook should be passed into that as well to account for the different quotes.

Also, right now the Quotes with zero value will fail - which should not be the case.

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/915157fc089a2635f0933c1dda0ab106dfca2706/typescript/sdk/src/token/adapters/CosmosModuleTokenAdapter.ts#L215-L219

This should return the correct zero amount, instead of assuming there is always a gas payment

typescript/sdk/src/token/adapters/CosmosModuleTokenAdapter.ts
```diff
@@ -216,8 +216,8 @@ export class CosmNativeHypCollateralAdapter
     });
 
     return {
-      addressOrDenom: gas_payment[0].denom,
-      amount: BigInt(gas_payment[0].amount),
+      addressOrDenom: gas_payment[0]?.denom ?? '',
+      amount: BigInt(gas_payment[0]?.amount ?? '0'),
```
xeno097, line 220: I don't think it is ok to have an empty string here, what happens if the code calling this expects a proper address but none is returned? shouldn't we assert that the first element of the `gas_payment` array exists and then use it instead? I don't think this would work fine on the front end for example

troykessler: @yjamin what do you think?

yjamin: Well coins is nullable on the cosmos side - I don't have context on the typescript interface and their expectations on this query, I just know that it resulted in issues already that SDK didn't allow nullable here.

typescript/cosmos-sdk/src/tests/4_warp.e2e-test.ts
```diff
@@ -185,6 +185,8 @@ describe('4. cosmos sdk warp e2e tests', async function () {
     const interchainGas = await signer.query.warp.QuoteRemoteTransfer({
       id: token.id,
       destination_domain: remoteRouter.receiver_domain.toString(),
+      custom_hook_id: '',
+      custom_hook_metadata: '',
```
paulbalaji, line 189: reuse the COSMOS_EMPTY_VALUE

PR #6506, opened 2025-06-11
fix(lander): escalation when error is immediately reported

daniel-savu:
### Description

Depends on https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6484

- Adds a unit test for escalating when a known retryable error is immediately returned when calling `sendRawTransaction` (`replacement transaction underpriced`)
- This revealed a bug where retrying the `submit` call wouldn't use the gas price set by the previous attempt of the loop inside `call_until_success_or_nonretryable_error`. The fix was to wrap the `Transaction` in an ArcMutex before calling `call_until_success_or_nonretryable_error(submit(tx))`

### Drive-by changes

Cleans up the inclusion step in various places

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1711/fully-fledged-evm-escalator

### Backward compatibility

Yes

### Testing

Unit testing


rust/main/lander/src/tests/ethereum/inclusion_stage.rs
```diff
@@ -124,32 +97,109 @@ async fn test_inclusion_gas_spike() {
...
+            elapsed,
+            base_processing_delay,
+            inclusion_stage_processing_delay,
+            block_time_clone,
+            &tx,
+            // First submission, price is 200000 - the default fee used by the ethers estimation logic
+            // Second submission, price is 10% higher, even though the spike was smaller
+            // Third submission, price is 10% higher again, even though the spike was smaller
+            // Fourth submission, price matches the spike, because it was greater than 10%
+            vec![200000, 220000],
```
ameten, line 176: I suppose the comment is incorrect since we expect only 2 submission attempts here.

daniel-savu: ah good catch, for this test we only need a single escalation. This is outdated copy pasta

ameten (approved):

PR #6517, opened 2025-06-11
feat: warp init with config hydration

ltyu:
### Description
Currently, we oread the onchain config many times, which adds significant overhead to deployment time (EZETH takes ~45 mins, _if it doesnt fail_). Most of the time cost is from deriving the default hook/ism.

This PR attempts to remedy this by allowing `warp apply` to use a cached warp config via `--actualConfigPath`. The idea is to use `warp read` and then use the filepath in the parameter.

### Backward compatibility
Yes

### Testing
Manual

### DIFF
--- a/typescript/cli/src/commands/warp.ts
+++ b/typescript/cli/src/commands/warp.ts
@@ -77,6 +77,7 @@ export const warpCommand: CommandModule = {
 
 export const apply: CommandModuleWithWarpApplyContext<{
   config?: string;
+  actualConfigPath?: string;
   warp?: string;
   symbol?: string;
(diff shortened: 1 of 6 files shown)


ltyu:
> is this the best solution to this? what if we had a global address->config lookup table in the multiprovider so the first read is always slow but then subsequent ones are quicker

played around with this idea and the challenge will be stale data. for example, i can imagine reading from an address, then updating it, and reading it again. 

we'd probably need to read the storageRoot to ensure the onchain store hasnt changed but not all rpc support `eth_getProof` for free.

PR #6525, opened 2025-06-12
test: cosmos core cli e2e tests

troykessler:
### Description

This PR adds e2e testing for the cosmos native core cli.

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

e2e tests

## Summary by CodeRabbit

* **New Features**
  * Introduced Cosmos-native end-to-end tests, configuration files, and automation scripts for streamlined testing.
  * Added utility to generate random Cosmos addresses.
  * Added new Docker Compose and chain metadata configurations for Cosmos-native environments.

* **Bug Fixes**
  * None.

* **Refactor**
  * Refactored core CLI command helpers into a unified class for improved maintainability.
  * Updated import paths throughout the test suite for consistency and clarity.
  * Simplified and trimmed Ethereum test helper utilities to focus on core CLI prompt handling.

* **Chores**
  * Separated Ethereum and Cosmos-native E2E test scripts and configurations.
  * Enhanced CI to run protocol-specific E2E tests.

* **Tests**
  * Added comprehensive Cosmos-native E2E test suites for deploy, read, check, and apply commands.
  * Added multiple Ethereum E2E test suites for core and warp command scenarios with improved class-based command handling.


typescript/cli/scripts/run-e2e-test.sh
```diff
@@ -10,20 +26,38 @@ function cleanup() {
...
-if [ -n "${CLI_E2E_TEST}" ]; then
-  echo "Running only ${CLI_E2E_TEST} test"
-  yarn mocha --config .mocharc-e2e.json "src/**/${CLI_E2E_TEST}.e2e-test.ts"
-else
-  yarn mocha --config .mocharc-e2e.json "src/**/*.e2e-test.ts"
-fi
+# cleanup_evm
+# setup_evm
+# run_evm
+# cleanup_evm
```
paulbalaji, line 56: probably should uncomment

typescript/cli/src/tests/cosmosnative/core/core-apply.e2e-test.ts
paulbalaji, line 1: i think because of the reuse of core-apply/core-check/etc as the test names - it maybe overriding the current e2e matrix to use the cosmos ones when they're there?

imo we should either split into 2 jobs (cli-e2e-evm/cli-e2e-cosmosnative) OR change the matrix here to include the protocol https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/542e7b1d986fe8926da252734bbca44243714cde/.github/workflows/test.yml#L292-L313

so it would all be ethereum-X-Y.e2e-test.ts or cosmosnative-X-Y.e2e-test.ts, and then depending on the protocol type we then adjust and pass the env var to the right test


i generally feel more in favour of splitting into 2 jobs though and keeping the run-e2e sh entrypoint in these new protocol-specific folders

.github/workflows/test.yml
```diff
@@ -332,7 +332,46 @@ jobs:
...
+        uses: ./.github/actions/install-cli
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          cache-provider: github
+
+      - name: Checkout registry
+        uses: ./.github/actions/checkout-registry
+
+      - name: CLI e2e tests (${{ matrix.test }})
+        run: yarn --cwd typescript/cli test:cosmosnative:e2e
```
xeno097, line 374: Considering that this is mostly the same as the ethereum one, can't we have another matrix variable in the original one called protocol adn the values are cosmosnative and ethereum so we don't have duplicated ci code?

typescript/cli/src/tests/cosmosnative/commands/core.ts
```diff
@@ -0,0 +1,160 @@
+import { $, ProcessPromise } from 'zx';
```
xeno097, line 1: Condiering that this file is mostly the same as the evm counterpart, can't we just update the already existing functions to accepts cosmos related values instead? so we avoid duplication?

typescript/cli/src/tests/cosmosnative/commands/helpers.ts
```diff
@@ -0,0 +1,97 @@
+import { ProcessOutput, ProcessPromise } from 'zx';
```
xeno097, line 1: ditto here, like you could have all the common stuff in a single place and then only reimplement things that are difficult to generalize for both vm types

typescript/cli/src/tests/cosmosnative/core/core-deploy.e2e-test.ts
```diff
@@ -0,0 +1,343 @@
...
+
+      expect(updatedConfig.owner.toLowerCase()).to.equal(newOwner);
+      expect(updatedConfig.proxyAdmin?.owner).to.be.undefined;
+      // Assuming that the ProtocolFeeHook is used for deployment
+      expect((updatedConfig.defaultHook as IgpConfig).owner).to.equal(
+        initialOwnerAddress,
+      );
+    });
+
+    it('should create a core deployment with ProxyAdmin owner of the mailbox set to the address in the config', async () => {
```
xeno097, line 310: Does the existence of proxy owner make sense in the cosmos world? i thought we didn't have one

typescript/cli/src/tests/commands/core.ts
```diff
@@ -1,117 +1,163 @@
 import { $, ProcessPromise } from 'zx';
 
 import { DerivedCoreConfig } from '@hyperlane-xyz/sdk';
-import { Address } from '@hyperlane-xyz/utils';
+import { ProtocolType } from '@hyperlane-xyz/utils';
```
xeno097, line 4: I'd keep the individual functions and have the enw class call them instead of completely remove them as they might be useful outside of the class context too, for example in warp e2e tests we sometimes save or read files from other paths. Having to create a class for using the methods seems a bit overkill in that case

typescript/cli/src/tests/commands/core.ts
```diff
@@ -1,117 +1,163 @@
...
-) {
-  return $`${localTestRunCmdPrefix()} hyperlane core deploy \
-        --registry ${REGISTRY_PATH} \
-        --config ${coreInputPath} \
-        --chain ${chain} \
-        --key ${ANVIL_KEY} \
-        --verbosity debug \
-        --yes`;
-}
+export class HyperlaneCore {
```
xeno097, line 10: I would change this name to `HyperlaneE2ECoreTestCommads` as it might be easy to think that this class allows to handle core deployments

typescript/cli/src/tests/cosmosnative/core/core-apply.e2e-test.ts
```diff
@@ -0,0 +1,79 @@
...
+    writeYamlOrJson(CORE_READ_CONFIG_PATH_1, coreConfig);
+    await hyperlaneCore.apply(CHAIN_NAME_1, HYP_KEY);
+    // Verify that the owner has been set correctly without modifying any other owner values
+    const updatedConfig: CoreConfig =
+      await hyperlaneCore.readConfig(CHAIN_NAME_1);
+    expect(updatedConfig.owner).to.equal(newOwner);
+    // Assuming that the IGP is used for deployment
+    expect((updatedConfig.defaultHook as IgpConfig).owner).to.equal(
+      initialOwnerAddress,
+    );
```
xeno097, line 77: Can we have a bit of spacing here? it is a bit hard to read this test

typescript/cli/src/tests/cosmosnative/commands/core.ts
troykessler: the remaining evm helpers like deploy, read, apply will be removed once the warp tests have been refactored in the next PR

xeno097 (approved):

PR #6537, opened 2025-06-13
feat: tidy up starknet contract code & added fallback to starknet pro…

yjamin:
Fallback providers for Starknet chains. Also tidy up the contract code to use `Reader` - meaning we don't have to initialize a signer/key for state queries anymore.

### Related issues
[Linear Issue](https://linear.app/hyperlane-xyz/issue/ENG-1730/add-fallback-provider-to-the-agents)

### Testing
Local end to end testing


rust/main/hyperlane-base/src/settings/parser/connection_parser.rs
```diff
@@ -325,16 +325,8 @@ fn build_starknet_connection_conf(
...
-        err.push(
-            &chain.cwp + "urls",
-            eyre!("No URLs provided for Starknet connection"),
-        );
-        return None;
-    };
-
     Some(ChainConnectionConf::Starknet(h_starknet::ConnectionConf {
-        url: url.clone(),
+        urls: urls.to_vec(),
```
yjamin, line 329: Can we safely assume this array will always contain at least one RPC URL?

rust/main/chains/hyperlane-starknet/src/provider/fallback.rs
```diff
@@ -0,0 +1,174 @@
...
+            for (idx, priority) in priorities_snapshot.iter().enumerate() {
+                // Create log span
+                let span = warn_span!("fallback_request", fallback_count = %idx, provider_index = %priority.index);
+                let _enter = span.enter();
+                let provider = &self.inner.providers[priority.index];
+
+                // first handle the stalled providers to not span the result of `send_request` in this future
+                // we have to do this, because the result does not implement the `Send` trait
+                let _ = self.handle_stalled_provider(priority, provider).await;
+
```
yjamin, line 130: I had to mimic a lot of the fallback behavior from the core implementation. The main reason why we need to do this is because the trait constraints are pretty tidy on the starknet provider traits and that ended up conflicting with the present implementation from the core package. 

Main issue was, that the result does not implement `Send` - which is needed if we perform a second await in the same span closure. 

I somewhat tested the implementation with the local e2e testing - might still be worth to do a separate test just for this implementation? 

Alternative is to re-write some of the already existing fallback logic in the fallback struct, but I don't really want to touch that tbh. 

What do you think is the best approach here? cc @tkporter

rust/main/chains/hyperlane-starknet/src/provider/fallback.rs
```diff
@@ -0,0 +1,174 @@
...
+        // If we get here, all providers failed
+        Err(FallbackHttpTransportError::Errors(errors))
+    }
+
+    async fn send_requests<R>(
+        &self,
+        requests: R,
+    ) -> Result<Vec<JsonRpcResponse<serde_json::Value>>, Self::Error>
+    where
+        R: AsRef<[ProviderRequestData]> + Send + Sync,
```
yjamin, line 157: Funny enough, the trait constraints are not at all as strict on this method. I wonder why that is the case, probably they just forgot to add the `Send` & `Clone` trait to the Result on `send_request`

rust/main/utils/run-locally/src/starknet/source.rs
```diff
@@ -80,7 +81,8 @@ impl CodeSource {
                 (entry.file_name().into_string().unwrap(), entry.path())
             })
             .filter(|(filename, _)| filename.ends_with(".contract_class.json"))
-            .filter(|(filename, _)| !filename.contains("Mock"))
+            .filter(|(filename, _)| !filename.to_lowercase().contains("test"))
+            .filter(|(filename, _)| !filename.to_lowercase().contains("mock"))
```
ameten, line 85: nit: filter in the same order as above

ameten (approved):

rust/main/chains/hyperlane-starknet/src/provider/fallback.rs
yjamin: Yeah will add a more explicit comment :+1:

PR #6546, opened 2025-06-16
feat: cosmosnative warp deploy

troykessler:
### Description

This PR implements warp deploy and read functionality for Cosmos Native

### Testing

Manual Testing, e2e tests will follow


typescript/cli/src/context/strategies/signer/MultiProtocolSignerManager.ts
```diff
@@ -221,4 +227,51 @@ export class MultiProtocolSignerManager {
...
+      }
+    }
+  }
+
+  async getBalance(
+    isDryRun: boolean,
+    address: Address,
+    chain: ChainName,
+    denom?: string,
+  ): Promise<BigNumber> {
```
xeno097, line 254: can we have all these params grouped in an object instead? i think after 2+ params it gets quite messy calling a function

typescript/cli/src/consts.ts
```diff
@@ -1,6 +1,21 @@
...
+  [ProtocolType.Cosmos]: (1e7).toString(),
+};
+export const MINIMUM_TEST_SEND_GAS = {
+  [ProtocolType.Ethereum]: (1e5).toString(),
+  [ProtocolType.Cosmos]: (1e5).toString(),
+};
+export const MINIMUM_AVS_GAS = {
+  [ProtocolType.Ethereum]: (1e6).toString(),
+  [ProtocolType.Cosmos]: (1e6).toString(),
+};
```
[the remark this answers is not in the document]
troykessler: These are the current values on `main`.

`export const MINIMUM_CORE_DEPLOY_GAS = (1e8).toString();
export const MINIMUM_WARP_DEPLOY_GAS = (3e7).toString();
export const MINIMUM_TEST_SEND_GAS = (3e5).toString();
export const MINIMUM_AVS_GAS = (3e6).toString();`

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -0,0 +1,75 @@
...
+      gas: gasOverhead(config.type),
+      ...config,
+    }));
+
+    const result: ChainMap<Address> = {};
+
+    const configMapToDeploy = objFilter(
+      resolvedConfigMap,
+      (_, config: any): config is any => !config.foreignDeployment,
+    );
```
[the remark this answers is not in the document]
troykessler: we filter the protocol specific config before and the cosmos deployer class is only used there: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/feat/cosmos-warp-deploy/typescript/cli/src/deploy/warp.ts#L268

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+
+    return transactions;
+  }
+
+  /**
+   * Create a transaction to update the remote routers for the Warp Route contract.
+   *
+   * @param actualConfig - The on-chain router configuration, including the remoteRouters array.
+   * @param expectedConfig - The expected token router configuration.
+   * @returns A array with a single Ethereum transaction that need to be executed to enroll the routers
```
xeno097, line 122: Aren't we in the cosmos context?

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+      })
+      .map(([domain]) => domain);
+
+    if (routesToEnroll.length === 0) {
+      return updateTransactions;
+    }
+
+    // in cosmos the gas is attached to the remote router. we set
+    // it to zero for now and set the real value later during the
+    // createSetDestinationGasUpdateTxs step
```
xeno097, line 156: I think it would be ok to diverge from the evm "flow" for updating the remote routers config if this helps also in reducing the gas needed for updates considering that the underlying config is different on chain.

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+      const gasRouterConfigs: { domain: string; gas: string }[] = [];
+      objMap(expectedDestinationGas, (domain: string, gas: string) => {
+        gasRouterConfigs.push({
+          domain,
+          gas,
+        });
+      });
+
+      // in cosmos updating the gas config is done by unenrolling the router and then
+      // enrolling it with the updating value again
```
xeno097, line 268: what happens when trying to set a value that is already set?

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * Create transactions to update an existing ISM config, or deploy a new ISM and return a tx to setInterchainSecurityModule
+   *
+   * @param actualConfig - The on-chain router configuration, including the ISM configuration, and address.
+   * @param expectedConfig - The expected token router configuration, including the ISM configuration.
+   * @returns Cosmos transaction that need to be executed to update the ISM configuration.
+   */
+  async createIsmUpdateTxs(
+    actualConfig: HypTokenRouterConfig,
```
xeno097, line 311: why don't you use the `DerivedTokenRouterConfig` type here so you don't have to type cast? (applies also for other places)

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+    expectedConfig: HypTokenRouterConfig,
+  ): Promise<AnnotatedCosmJsNativeTransaction[]> {
+    const updateTransactions: AnnotatedCosmJsNativeTransaction[] = [];
+    if (
+      !expectedConfig.interchainSecurityModule ||
+      expectedConfig.interchainSecurityModule === zeroAddress
```
xeno097, line 317: Instead of using `=== zeroAddress` can you use the `isZeroishAddress` func?

typescript/sdk/src/token/CosmosNativeWarpRouteReader.ts
```diff
@@ -0,0 +1,189 @@
...
+  async deriveWarpRouteConfig(
+    warpRouteAddress: Address,
+  ): Promise<DerivedTokenRouterConfig> {
+    // Derive the config type
+    const type = await this.deriveTokenType(warpRouteAddress);
+    const baseMetadata = await this.fetchMailboxClientConfig(warpRouteAddress);
+    const tokenConfig = await this.fetchTokenConfig(type, warpRouteAddress);
+    const remoteRouters = await this.fetchRemoteRouters(warpRouteAddress);
+    const proxyAdmin = await this.fetchProxyAdminConfig(warpRouteAddress);
+    const destinationGas = await this.fetchDestinationGas(warpRouteAddress);
```
xeno097, line 62: Would it be possible to `Promise.all` or there would be concerns with rpc rate limiting?

typescript/sdk/src/token/CosmosNativeWarpRouteReader.ts
```diff
@@ -0,0 +1,189 @@
...
+  ): Promise<DeployedOwnableConfig> {
+    const { token } = await this.cosmosProviderOrSigner.query.warp.Token({
+      id: tokenAddress,
+    });
+
+    return {
+      address: tokenAddress,
+      owner: token?.owner ?? '',
+    };
+  }
```
xeno097, line 173: considering that cosmos does not have a proxy admin i think you can delete this and leave the `proxyAdmin` as undefined to aovid giving the idea that there is a proxy admin when there is none

typescript/sdk/src/token/CosmosNativeWarpRouteReader.ts
```diff
@@ -0,0 +1,189 @@
...
+    const { remote_routers } =
+      await this.cosmosProviderOrSigner.query.warp.RemoteRouters({
+        id: warpRouteAddress,
+      });
+
+    const destinationGas: DestinationGas = {};
+    for (const router of remote_routers) {
+      destinationGas[router.receiver_domain] = router.gas;
+    }
+    return destinationGas;
```
xeno097, line 187: just a nit here but i would prefer a more functional approach here, assuming remote_routers is an array, i'd do:
```typescript
return Object.fromEntries(
    remote_routers.map(
        routerConfig => [routerConfig.receiver_domain, routerConfig.gas]
    )
)
```

typescript/sdk/src/token/cosmosnativeDeploy.ts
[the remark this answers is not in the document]
xeno097: mmm i don't quite like it reliance on filtering before calling this, if for some reason this is used outside of the current code and one forgets to filter it might fail, i'd prefer filtering here too just to be safe

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        case TokenType.synthetic: {
+          result[chain] = await this.deploySyntheticToken(
+            chain,
+            config.mailbox,
+          );
+          break;
+        }
+        default: {
+          throw new Error(`Token type ${config.type} not supported`);
```
xeno097, line 70: Maybe also add the chain or protocol type to be more explicit

typescript/cli/src/deploy/warp.ts
```diff
@@ -198,27 +218,114 @@ async function runDeployPlanStep({ context, warpDeployConfig }: DeployParams) {
...
+      continue;
+    }
+
+    switch (protocol) {
+      case ProtocolType.Ethereum: {
+        const deployer = warpDeployConfig.isNft
+          ? new HypERC721Deployer(multiProvider)
+          : new HypERC20Deployer(multiProvider); // TODO: replace with EvmERC20WarpModule
+
+        // TODO: enroll non-evm routes
```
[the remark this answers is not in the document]
troykessler: not valid anymore

typescript/sdk/src/token/CosmosNativeWarpRouteReader.ts
troykessler: this could lead to issues with ratelimiting. Most rpc endpoints really don't like parallel requests

typescript/sdk/src/token/CosmosNativeWarpModule.ts
troykessler: it simply won't work (values won't update on chain)

PR #6550, opened 2025-06-16
feat: integrate GCP logging with pino for warp-monitor

antigremlin:
### Description

This PR applies a GCP-specific configuration to pino logging when used from the warp-monitor. We only use the GCP-specific config when running in Kubernetes and the import is dynamic.

### Related issues

Fixes ENG-1771

### Backward compatibility

Yes, but the new behaviour is triggered by the `KUBERNETES_SERVICE_HOST` environment variable rather than a dedicated setting.

Caveat: the `@google-cloud/pino-logging-gcp-config` package is actually compatible with `pino@9`, even though we are currently using `pino@8`. We type-cast to add it to the older version (the config is a plain object), and manual testing shows that it works correctly.

### Testing

Visula testing of the logging output.

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging to support environment-specific configurations, including integration with Google Cloud Platform logging when running in Kubernetes environments.
	- Added the ability to set custom logger bindings for improved log context.

- **Chores**
	- Updated dependencies to include Google Cloud logging configuration support.

### DIFF
--- a/.changeset/lazy-candles-help.md
+++ b/.changeset/lazy-candles-help.md
@@ -1,0 +1,6 @@
+---
+'@hyperlane-xyz/cli': minor
+'@hyperlane-xyz/sdk': minor
+---
+
+add warp fork and fork commands
--- a/.changeset/lemon-apples-jam.md
+++ b/.changeset/lemon-apples-jam.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': minor
+---
+
+updates the warp core output file logic to use the input file name as seed for the filename when a warp id is not provided
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -304,6 +304,7 @@ jobs:
           - warp-bridge-1
           - warp-bridge-2
           - warp-check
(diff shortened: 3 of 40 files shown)


typescript/infra/scripts/warp-routes/monitor/utils.ts
```diff
@@ -1,6 +1,44 @@
-import { rootLogger } from '@hyperlane-xyz/utils';
+import pino, { LoggerOptions } from 'pino';
 
-export const logger = rootLogger.child({ module: 'warp-balance-monitor' });
+import { rootLogger, setRootLogger } from '@hyperlane-xyz/utils';
+
+async function createLogger() {
```
Mo-Hussain, line 5: I think we should have defined a method in utils and passed in a service name etc

antigremlin: You're right, this would be more useful. However, this file is inside the monitor subdirectory so to use it from other infra tools we need to move it. Makes sense for a separate PR

PR #6551, opened 2025-06-16
feat: Update SmartProvider to re-throw with cause

ltyu:
### Description
This PR fixes a bug where `fetchPackageVersion()` expects the raw RPC error code. However, since we use a SmartProvider and re-throw a clean string message, we need to add and use the `cause`

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause

### Related issues
- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1778/warp-read-fails-for-ezeth

### Backward compatibility
Yes

### Testing
Manual/Unit Tests

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling by preserving original error context, making errors easier to trace and understand.
	- Enhanced detection of specific contract version errors for more accurate fallback behavior.

- **Tests**
	- Added assertions to verify error causes and codes alongside messages, strengthening error validation.


typescript/sdk/src/providers/SmartProvider/SmartProvider.foundry-test.ts
```diff
@@ -156,6 +158,7 @@ describe('SmartProvider', async () => {
     try {
       await token.transfer(constants.AddressZero, 1000000);
     } catch (e: any) {
+      expect(e.error.cause.code).to.equal('UNPREDICTABLE_GAS_LIMIT');
```
xeno097, line 161: Maybe to avoid magic strings, we could have an enum to encode the errors from the smart provider that we are inteerested into

xeno097 (approved):

typescript/sdk/src/providers/SmartProvider/SmartProvider.foundry-test.ts
ltyu: good point, can import from ethers

PR #6566, opened 2025-06-18
fix: use consistent warp route selector for all warp commands

yorhodes:
### Description

Make flags to `warp XXX` commands consistent for selecting a warp route from the registry

## Summary by CodeRabbit

- **Refactor**
  - Unified and simplified command options across multiple Warp CLI commands for a more consistent user experience.
  - Made certain command options optional to enhance flexibility when running commands.

- **Bug Fixes**
  - Improved reliability of the verify command by updating its configuration handling.


typescript/cli/src/commands/warp.ts
```diff
@@ -89,27 +86,36 @@ export const warpCommand: CommandModule = {
...
 
-export const apply: CommandModuleWithWarpApplyContext<{
-  config?: string;
-  warp?: string;
-  symbol?: string;
-  warpRouteId?: string;
-  strategy?: string;
-  receiptsDir: string;
-}> = {
+const SELECT_WARP_ROUTE_BUILDER = {
```
paulbalaji, line 89: wonder if this should be in https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/cli/src/commands/options.ts ?

but if there's no other commands that might want to reuse this, happy to leave this here

paulbalaji (approved):

typescript/cli/src/commands/warp.ts
yorhodes: will leave in here for now

PR #6568, opened 2025-06-18
fix: only derive rebalancing config when bytecode version matches

yorhodes:
### Description

Only derive rebalancing config in WarpModule.read when contractVersion implies the interface

## Summary by CodeRabbit

- **New Features**
  - Improved support for rebalancing features, now only available for tokens with contract versions 8.0.0 and above.
- **Bug Fixes**
  - Ensured rebalancing-related options are shown only when supported by the token’s contract version.

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -1,3 +1,4 @@
+import { compareVersions } from 'compare-versions';
 import { BigNumber, Contract } from 'ethers';
 
 import {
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -142,9 +145,18 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
       : undefined;
     const destinationGas = await this.fetchDestinationGas(warpRouteAddress);
 
+    const hasRebalancingInterface =
+      compareVersions(
+        tokenConfig.contractVersion!,
+        REBALANCING_CONTRACT_VERSION,
+      ) > 0;
```
xeno097, line 152: Shouldn't this be `>= 0` because if the version matches we also want to fetch these values no? Checking the docs 0 is returned if the versions are equal which means we can fetch the new fields (probably the reason why tests are failing too) 
ref: https://www.npmjs.com/package/compare-versions

xeno097 (commented):
seems the tests don't like the change (probably because now the new)

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
yorhodes, line 152:
```suggestion
      ) >= 0;
```

PR #6571, opened 2025-06-19
Verify factory implementation

yorhodes:

### DIFF
--- a/typescript/sdk/src/deploy/HyperlaneProxyFactoryDeployer.ts
+++ b/typescript/sdk/src/deploy/HyperlaneProxyFactoryDeployer.ts
@@ -48,14 +48,14 @@ export class HyperlaneProxyFactoryDeployer extends HyperlaneDeployer<
       this.factories,
     ) as (keyof ProxyFactoryFactories)[]) {
       const factory = await this.deployContract(chain, factoryName, []);
-      this.addVerificationArtifacts(chain, [
-        {
-          name: proxyFactoryImplementations[factoryName],
-          address: await factory.implementation(),
-          constructorArguments: '',
-          isProxy: true,
-        },
-      ]);
+      const artifact = {
+        name: proxyFactoryImplementations[factoryName],
+        address: await factory.implementation(),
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/deploy/HyperlaneProxyFactoryDeployer.ts
```diff
@@ -48,14 +48,14 @@ export class HyperlaneProxyFactoryDeployer extends HyperlaneDeployer<
...
-        },
-      ]);
+      const artifact = {
+        name: proxyFactoryImplementations[factoryName],
+        address: await factory.implementation(),
+        constructorArguments: '',
+        isProxy: true,
+      };
+      await this.verifyContract(chain, artifact);
+      this.addVerificationArtifacts(chain, [artifact]);
```
ltyu, line 58: how come we still need to add the artifacts?

yorhodes: this is my trying to comply with good practice of minimizing backward incompatible changes

typescript/sdk/src/deploy/HyperlaneProxyFactoryDeployer.ts
```diff
@@ -48,14 +48,19 @@ export class HyperlaneProxyFactoryDeployer extends HyperlaneDeployer<
...
+          name: proxyFactoryImplementations[factoryName],
+          address: await factory.implementation(),
+          constructorArguments: '',
+          isProxy: true,
+        };
+        await this.verifyContract(chain, artifact);
+        this.addVerificationArtifacts(chain, [artifact]);
+      } catch (error) {
+        this.logger.warn(`Failed to verify ${factoryName} on ${chain}:`, error);
+        // Continue deployment even if verification fails
```
paulbalaji, line 62: i think you wanted to only try-catch the `await this.verifyContract(chain, artifact);`

now if the verification fails we have no artifact to post-verify with

PR #6585, opened 2025-06-20
feat: add core apply hooks update support [ENG-1804]

xeno097:
### Description

This PR adds support for updating required and default hooks using the `core apply` command

### Drive-by changes

- Moves the logic for updating hooks in the EVM warp module to a helper function to allow reuse in the core module 
- Moves all the hook update logic to the HookModule class 

### Related issues

- Fixes [ENG-1804](https://linear.app/hyperlane-xyz/issue/ENG-1804/bugcli-sdk-core-apply-does-not-update-hook-configs)

### Backward compatibility

- YES
- 
### Testing

- unit
- e2e

## Summary by CodeRabbit

- **New Features**
  - Added support for updating required and default hooks in core deployments, enhancing configuration flexibility for users.

- **Bug Fixes**
  - Improved handling and updating of hook configurations to ensure accurate state changes and reduce the risk of outdated settings.

- **Refactor**
  - Streamlined the update process for hooks by consolidating update logic and delegating transaction creation to a new utility, improving maintainability without affecting user-facing functionality.

- **Tests**
  - Expanded test coverage to verify correct behavior when updating required and default hooks, ensuring reliability of the new features.


typescript/sdk/src/hook/updates.ts
```diff
@@ -0,0 +1,184 @@
...
+    isZeroishAddress(expectedHookConfig)
+  ) {
+    return [];
+  }
+
+  // Try to deploy or update Hook with the expected config
+  const {
+    deployedHook: expectedDeployedHook,
+    updateTransactions: hookUpdateTransactions,
+  } = await deployOrUpdateHook(updateHookParams);
```
paulbalaji, line 62: why not move the `deployOrUpdate` logic here into `update()` of the hook module more natively?

xeno097: that is a good point, I'll try and see what happens

typescript/sdk/src/core/EvmCoreModule.ts
```diff
@@ -113,17 +116,36 @@ export class EvmCoreModule extends HyperlaneModule<
...
+    if (expectedConfig.requiredHook) {
+      transactions.push(
+        ...(await this.createHookUpdateTxs(
+          proxyAdminAddress,
+          'requiredHook',
+          actualConfig.requiredHook,
+          expectedConfig.requiredHook,
+        )),
+      );
+    }
```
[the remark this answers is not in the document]
xeno097: the proxyAdmin for the addresses will always be defined so this is a non issue

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -321,6 +278,79 @@ export class EvmHookModule extends HyperlaneModule<
...
+      current.type === target.type,
+      `Mutable hook update requires both hook configs to be of the same type. Expected ${current.type}, got ${target.type}`,
+    );
+    assert(
+      MUTABLE_HOOK_TYPE.includes(current.type),
+      'Expected update config to be of mutable hook type',
+    );
+    // Checking both objects type fields to help typescript narrow the type down correctly
+    switch (true) {
+      case current.type === HookType.INTERCHAIN_GAS_PAYMASTER &&
```
ltyu, line 298: nit: switch (true) feels a bit strange here. 

can we use if-statements, or perhaps we can keep the switch and use `current as IgpHookConfig`?

ltyu (approved):
LGTM. just a few nits

typescript/sdk/src/hook/EvmHookModule.ts
```diff
@@ -151,27 +153,41 @@ export class EvmHookModule extends HyperlaneModule<
     targetConfig: HookConfig,
   ): Promise<AnnotatedEV5Transaction[]> {
     // Nothing to do if its the default hook
-    if (targetConfig === zeroAddress) {
-      return Promise.resolve([]);
+    if (typeof targetConfig === 'string' && isZeroishAddress(targetConfig)) {
```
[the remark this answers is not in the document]
xeno097: unfortunately we can't do this because the typeof operator works only on js primitive types, meaning we can only check if the type is string, number, object etc. In this case as the `Address` is an alias for the string type we need to assert that it is a string.
Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

typescript/sdk/src/hook/EvmHookModule.ts
xeno097: I'll change it to if-statements becuase i want to avoid type assertions as much as possible evne if they are safe in this context

typescript/sdk/src/hook/EvmHookModule.ts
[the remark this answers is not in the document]
ltyu: ah did not know that!

PR #6588, opened 2025-06-20
fix: rebalance undercollateralization

yorhodes:
### Description

Ensures the rebalancer is charged for fees on top of the collateral amount

### Drive-by changes

- infinite approves collateral token to configured bridges

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/ENG-1805/take-fees-from-the-rebalancer-instead-of-bridge-collateral

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fee handling to ensure the rebalancer covers all associated fees during the rebalancing process.
  * Added checks to prevent undercollateralization during rebalance operations.

* **New Features**
  * Enhanced documentation for fee quoting, clarifying token ordering and uniqueness in returned quotes.

* **Refactor**
  * Streamlined bridge management and fee charging logic for better modularity and flexibility.
  * Updated parameter naming for improved clarity in fee-related contracts.

* **Tests**
  * Added and updated tests to verify fee coverage and prevent undercollateralization in rebalancing scenarios.
  * Introduced new test tokens and focused test cases for improved reliability.


solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -67,6 +67,11 @@ contract HypERC20Collateral is MovableCollateralRouter {
         return address(wrappedToken);
     }
 
+    function _addBridge(uint32 domain, ITokenBridge bridge) internal override {
+        MovableCollateralRouter._addBridge(domain, bridge);
+        IERC20(wrappedToken).safeApprove(address(bridge), type(uint256).max);
```
larryob, line 72: No need for the `IERC20` cast since `wrappedToken` is already `IERC20`

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -69,6 +69,10 @@ abstract contract MovableCollateralRouter is FungibleTokenRouter {
     function addBridge(uint32 domain, ITokenBridge bridge) external onlyOwner {
         // constrain to a subset of Router.domains()
         _mustHaveRemoteRouter(domain);
```
larryob, line 71: Maybe move this call into `_addBridge()`?

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -113,20 +117,45 @@ abstract contract MovableCollateralRouter is FungibleTokenRouter {
...
+        bytes32 recipient = _recipient(domain);
+
+        Quote[] memory quotes = bridge.quoteTransferRemote(
+            domain,
+            recipient,
+            amount
+        );
+
+        uint256 collateralFee = 0;
+        for (uint256 i = 0; i < quotes.length; i++) {
```
larryob, line 129: Gas nit: Cache the length.

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -113,20 +117,45 @@ abstract contract MovableCollateralRouter is FungibleTokenRouter {
...
+            if (quotes[i].token == token()) {
+                collateralFee += quotes[i].amount;
```
larryob, line 131: I think we can break here.

solidity/contracts/token/libs/MovableCollateralRouter.sol
```diff
@@ -113,20 +117,45 @@ abstract contract MovableCollateralRouter is FungibleTokenRouter {
...
+            }
+        }
+
+        // charge the rebalancer any bridging fees denominated in the collateral
+        // token to avoid undercollateralization
+        if (collateralFee > amount) {
+            _transferFromSender(collateralFee - amount);
```
larryob, line 138: Maybe this is done elsewhere, but shouldn't we just take `collateralFee` from the rebalancer in all cases?

solidity/test/token/HypnativeMovable.t.sol
```diff
@@ -31,21 +31,27 @@ contract MockITokenBridgeEth is ITokenBridge {
 
 contract HypNativeMovableTest is Test {
```
larryob, line 32: Can we add a test to prove undercollateralization never happens?

larryob (commented):
Looks good, left some nits and one question.

solidity/contracts/token/libs/MovableCollateralRouter.sol
yorhodes: the collateral fee (and quote) will include the amount being bridged, which we want to use this contract's balance for

solidity/contracts/token/libs/MovableCollateralRouter.sol
yorhodes: if there are multiple quotes with the same token, wont that undercount?

solidity/contracts/token/libs/MovableCollateralRouter.sol
yorhodes: does this actually save gas? expanding memory by a uint256

solidity/test/token/HypnativeMovable.t.sol
yorhodes: good idea

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: It should. If you cache you're doing one `mload`, if you don't you're doing `n` `mload`s where `n` is the size of your loop.

solidity/contracts/token/libs/MovableCollateralRouter.sol
larryob: I don't think multiple quotes with the same token should be supported, maybe we can add that to the spec for `quoteTransferRemote`.

solidity/contracts/token/libs/MovableCollateralRouter.sol
yorhodes: I think this is the superior UX
otherwise the client needs to implement its own logic for deciding when to approve and how much

solidity/contracts/token/libs/MovableCollateralRouter.sol
yorhodes: isnt quotes.length already in memory though?

yorhodes: without some docs or minimal repro I'm inclined to leave this for readability

PR #6591, opened 2025-06-23
chore: add cosmos native sdk warp deploy logic

troykessler:
### Description

This PR includes the SDK implementation for cosmos native warp deploy

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Testing

### DIFF
--- a/typescript/sdk/src/token/CosmosNativeWarpModule.ts
+++ b/typescript/sdk/src/token/CosmosNativeWarpModule.ts
@@ -29,8 +29,12 @@ import { AnnotatedCosmJsNativeTransaction } from '../providers/ProviderType.js';
 import { ChainName, ChainNameOrId } from '../types.js';
 
 import { CosmosNativeWarpRouteReader } from './CosmosNativeWarpRouteReader.js';
-import { TokenType } from './config.js';
-import { HypTokenRouterConfig, HypTokenRouterConfigSchema } from './types.js';
+import { CosmosNativeDeployer } from './cosmosnativeDeploy.js';
+import {
+  DerivedTokenRouterConfig,
+  HypTokenRouterConfig,
+  HypTokenRouterConfigSchema,
+} from './types.js';
 
 type WarpRouteAddresses = {
   deployedTokenRoute: Address;
@@ -71,7 +75,7 @@ export class CosmosNativeWarpModule extends HyperlaneModule<
    * @param address - The address to derive the token router configuration from.
    * @returns A promise that resolves to the token router configuration.
    */
-  async read(): Promise<HypTokenRouterConfig> {
+  async read(): Promise<DerivedTokenRouterConfig> {
     return this.reader.deriveWarpRouteConfig(
       this.args.addresses.deployedTokenRoute,
     );
@@ -122,7 +126,7 @@ export class CosmosNativeWarpModule extends HyperlaneModule<
    * @returns An array with Cosmos Native transactions that need to be executed to enroll the routers
    */
   createEnrollRemoteRoutersUpdateTxs(
-    actualConfig: HypTokenRouterConfig,
+    actualConfig: DerivedTokenRouterConfig,
     expectedConfig: HypTokenRouterConfig,
   ): AnnotatedCosmJsNativeTransaction[] {
     const updateTransactions: AnnotatedCosmJsNativeTransaction[] = [];
@@ -176,7 +180,7 @@ export class CosmosNativeWarpModule extends HyperlaneModule<
   }
 
   createUnenrollRemoteRoutersUpdateTxs(
-    actualConfig: HypTokenRouterConfig,
+    actualConfig: DerivedTokenRouterConfig,
     expectedConfig: HypTokenRouterConfig,
   ): AnnotatedCosmJsNativeTransaction[] {
     const updateTransactions: AnnotatedCosmJsNativeTransaction[] = [];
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * Create a transaction to update the remote routers for the Warp Route contract.
+   *
+   * @param actualConfig - The on-chain router configuration, including the remoteRouters array.
+   * @param expectedConfig - The expected token router configuration.
+   * @returns An array with Cosmos Native transactions that need to be executed to enroll the routers
+   */
+  createEnrollRemoteRoutersUpdateTxs(
+    actualConfig: HypTokenRouterConfig,
```
xeno097, line 125: Why not use `DerivedTokenRouterConfig` for the actualConfig var? this is the type returned by the reader so there shouldn't be any problems here

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+          },
+        }),
+      });
+    });
+
+    return updateTransactions;
+  }
+
+  createUnenrollRemoteRoutersUpdateTxs(
+    actualConfig: HypTokenRouterConfig,
```
xeno097, line 179: ditto

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * Create a transaction to update the remote routers for the Warp Route contract.
+   *
+   * @param actualConfig - The on-chain router configuration, including the remoteRouters array.
+   * @param expectedConfig - The expected token router configuration.
+   * @returns A array with Cosmos transactions that need to be executed to update the destination gas
+   */
+  async createSetDestinationGasUpdateTxs(
+    actualConfig: HypTokenRouterConfig,
```
xeno097, line 227: ditto

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * Create transactions to update an existing ISM config, or deploy a new ISM and return a tx to setInterchainSecurityModule
+   *
+   * @param actualConfig - The on-chain router configuration, including the ISM configuration, and address.
+   * @param expectedConfig - The expected token router configuration, including the ISM configuration.
+   * @returns Cosmos transaction that need to be executed to update the ISM configuration.
+   */
+  async createIsmUpdateTxs(
+    actualConfig: HypTokenRouterConfig,
```
xeno097, line 311: ditto

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+      ...this.createEnrollRemoteRoutersUpdateTxs(actualConfig, expectedConfig),
+      ...this.createUnenrollRemoteRoutersUpdateTxs(
+        actualConfig,
+        expectedConfig,
+      ),
+      ...(await this.createSetDestinationGasUpdateTxs(
+        actualConfig,
+        expectedConfig,
+      )),
+      ...this.createOwnershipUpdateTxs(actualConfig, expectedConfig),
```
xeno097, line 111: No hook update?

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -0,0 +1,108 @@
...
+  rootLogger,
+} from '@hyperlane-xyz/utils';
+
+import { ChainMetadataManager } from '../metadata/ChainMetadataManager.js';
+import { ChainMap, ChainName } from '../types.js';
+
+import { TokenType, gasOverhead } from './config.js';
+import { WarpRouteDeployConfigMailboxRequired } from './types.js';
+
+export class CosmosNativeDeployer {
```
xeno097, line 19: Why is this class defined but not used anywhere? In the Module i see that the signer is used directly for deploying the tokens

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -0,0 +1,469 @@
...
+      const { response } = await signer.createCollateralToken({
+        origin_mailbox: config.mailbox,
+        origin_denom: config.token,
+      });
+      deployedTokenRoute = response.id;
+    } else if (config.type === TokenType.synthetic) {
+      const { response } = await signer.createSyntheticToken({
+        origin_mailbox: config.mailbox,
+      });
+      deployedTokenRoute = response.id;
```
xeno097, line 448: Should the `CosmosNativeDeployer` be used here instead?

xeno097 (commented):
It seems that changeset is also missing

typescript/sdk/src/token/cosmosnativeDeploy.ts
troykessler: it is the cosmos native equivalent class to https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/chore/cosmos-sdk-warp-deploy/typescript/sdk/src/token/deploy.ts#L107 and gets only directly used in the cli later

typescript/sdk/src/token/CosmosNativeWarpModule.ts
troykessler: currently there are no hooks in tokens yet on cosmos native, this could change with future releases but therefore for now no hooks updates are needed

PR #6594, opened 2025-06-23
feat: deploy to TAC

paulbalaji:
### Description

feat: deploy to TAC

Deployment was done from the commit just before merging pre-audit branch (33c203cb7). PR was then rebased ontop of main, since we can just reuse the old addresses with latest infra.

The intention is to migrate over all chains to new ICAs, but in the interest of speed and consistency, this method ensures we still keep all chains on the legacy ICA implementation. If the ICA migration is ready by the time we start transferring TAC ownership, then we will go straight to the new ICAs. Either way, we want parity with whichever ICA implementation the other chains use.

## Summary by CodeRabbit

- **New Features**
  - Added support for the new "tac" blockchain network across mainnet configurations, including validator setup, contract verification, ISM contracts, and middleware accounts.
  - Enabled "tac" for Validator, Relayer, and Scraper roles in agent configurations.
  - Included "tac" in supported chain lists and multisig configurations.

- **Configuration Updates**
  - Set relayer balances, gas prices, and token price for "tac".
  - Updated Docker image tags for agent and funding components.

- **Bug Fixes**
  - Corrected typographical errors in contract and ISM configuration names from "DomaingRoutingIsm" to "DomainRoutingIsm".

### DIFF
--- a/.changeset/pink-buckets-glow.md
+++ b/.changeset/pink-buckets-glow.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/sdk": patch
+---
+
+Fix typo DomaingRoutingIsm -> DomainRoutingIsm.
--- a/typescript/infra/config/environments/mainnet3/ism/verification.json
+++ b/typescript/infra/config/environments/mainnet3/ism/verification.json
@@ -58,7 +58,7 @@
       "address": "0x12Ed1BbA182CbC63692F813651BD493B7445C874",
       "constructorArguments": "",
       "isProxy": true,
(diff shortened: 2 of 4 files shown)


.registryrc
```diff
@@ -1 +1 @@
-b661a6b20053a9c6c35eb5ecf2fa18e88912e47b
+66548f9e4d05d0849ac61d536ca6433e8d6aaa76
```
paulbalaji, line 1: update after merging https://github.com/hyperlane-xyz/hyperlane-registry/pull/985

tkporter (approved):

PR #6604, opened 2025-06-24
test(lander): tx db state assertions

daniel-savu:
### Description

The original incident that motivated this PR was when the relayer restarted, with some txs stuck in the mempool that it wasn't aware of - this would result in a "tx underpriced" error because the nonce would end up reused. That incident can be broken up into two parts:
1. Make sure that after each iteration of the Inclusion Stage loop, the status of txs is correctly updated in the Transactions DB. **This is what this PR does**
2. The Transactions are loaded correctly from the DB with the expected state on startup. This is partly tested in `rust/main/lander/src/dispatcher/db/transaction/transaction_db.rs`, but there are no assertions over the actual Transaction fields. This will come in a follow up PR

### Drive-by changes

- A bug was fixed in the transaction DB persistence logic. Previously, the tx's state (including VM specific details) wouldn't be updated in the DB if the status of the tx stayed the same. This is wrong, because resubmitting with a higher gas price would not have update dthe DB until the tx were reported as included by the chain. Restarting the relayer after the escalation would've resulted in the tx getting loaded with an old gas price
- The Inclusion Stage logic is slightly refactored. Now, the logic inside the blocktime-drive loop is moved into its own function, called `process_txs_step`. This makes unit testing via dependency inversion very easy
- Inlcusion Stage unit tests are refactored to call `process_txs_step` instead of the previous flakier approach of `tokio::select!`-ing against a blocktime timeout
- Clean up of how nonces are mocked in the Ethereum Inclusion Stage tests

### Backward compatibility

Yes

### Testing

Unit tests

## Summary by CodeRabbit

* **New Features**
  * Exposed additional modules and functions to enable broader integration capabilities.
  * Added a new function for gas estimation buffer handling with explicit type conversions.

* **Refactor**
  * Modularized transaction processing logic for improved maintainability and clarity.
  * Enhanced test structure for the Ethereum inclusion stage, enabling detailed, stepwise verification of transaction state transitions.

* **Tests**
  * Introduced more granular and explicit transaction state assertions within inclusion stage tests.
  * Improved test mocks and harnesses for more accurate simulation of transaction scenarios.

* **Bug Fixes**
  * Ensured transaction status updates are always logged and stored, even when the status remains unchanged.


rust/main/lander/src/adapter/chains/ethereum/adapter.rs
```diff
@@ -36,7 +36,7 @@ use super::{
     EthereumTxPrecursor,
 };
 
-mod gas_limit_estimator;
+pub mod gas_limit_estimator;
```
ameten, line 39:
```suggestion
#[cfg(test)]
pub use gas_limit_estimator::apply_estimate_buffer_to_ethers;
```

ameten (approved):

PR #6611, opened 2025-06-25
fix(sdk): respect expectedRemoteChains when checking enrollments

paulbalaji:
### Description

fix(sdk): respect expectedRemoteChains when checking enrollments

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the enrollment verification process to correctly respect specified remote chains during router checks. If a list of expected remote chains is provided, only those chains are checked; otherwise, all enrolled chains are checked as before.


antigremlin (approved):
🤦

PR #6639, opened 2025-06-30
chore: reconstruct token configs

christopherbrumm:
### Description

chore: reconstruct token configs for dreams, Fartcoin & SOL


rust/sealevel/environments/mainnet3/warp-routes/Fartcoin-starknet/pogram-ids.json
```diff
@@ -0,0 +1,10 @@
+{
```
tkporter, line 1: seems like this file name is missing an r, `pogram` -> `program`

tkporter (approved):
lgtm, thank you!

PR #6649, opened 2025-06-30
fix: Remove absolute imports

### LINKED ISSUE #6559: `@hyperlane-xyz/core` contracts compile with an error
I am trying to upgrade to v8 of the core contracts but the compiler gives me the following compiler error.

Image

There could be a hack around with implementing some remapping, but from the developer perspective importing these contracts, this is not ideal.

The issue comes from the following import statement:

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/64092311c0e0bbaef0aa798e0115dac641fcfabe/solidity/contracts/token/libs/MovableCollateralRouter.sol#L4

This can be easily fixed by changing this line to:

```solidity
import { Router } from "../../client/Router.sol";
```

larryob:
### Description

All absolute imports are removed. Absolute imports let us easily see the full path to a file, but prevent compilation when this repo is installed as a dependency. See https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6560/files. 

### Issues
Fixes #6559

## Summary by CodeRabbit

* **Chores**
  * Updated import paths throughout the Solidity project to use consistent relative paths.
  * Updated the ignore list to exclude the broadcast directory from version control.
  * Simplified import statements in select script files for improved readability.

### DIFF
--- a/.changeset/large-flies-rescue.md
+++ b/.changeset/large-flies-rescue.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/core": patch
+---
+
+Remove absolute imports. Fixes compilation for users who import from files under `solidity/contracts`.
--- a/solidity/.gitignore
+++ b/solidity/.gitignore
@@ -15,6 +15,7 @@ docs
 flattened/
 buildArtifact.json
 fixtures/
+broadcast/
 # ZKSync
 artifacts-zk
 cache-zk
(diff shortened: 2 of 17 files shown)


yorhodes (approved):

paulbalaji:
given this was merged to main, can we close this PR?

PR #6654, opened 2025-07-01
feat: add regex to matching list

kamiyaa:
### Description

 - add a regex field for the body to matching list

### Related issues

- fixes [https://linear.app/hyperlane-xyz/issue/ENG-1828/add-a-body-regex-filter-to-matchinglists]

### Backward compatibility

Yes

### Testing

- unittests

### DIFF
--- a/rust/main/agents/relayer/src/settings/matching_list.rs
+++ b/rust/main/agents/relayer/src/settings/matching_list.rs
@@ -9,9 +9,11 @@ use std::{
 };
 
 use derive_new::new;
+use ethers::utils::hex;
 use hyperlane_core::{
     config::StrOrInt, utils::hex_or_base58_to_h256, HyperlaneMessage, QueueOperation, H256,
 };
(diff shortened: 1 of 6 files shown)


rust/main/agents/relayer/src/settings/matching_list.rs
```diff
@@ -236,6 +267,8 @@ pub struct ListElement {
     destination_domain: Filter<u32>,
     #[serde(default, rename = "recipientaddress")]
     recipient_address: Filter<H256>,
+    #[serde(default, rename = "regex")]
+    body: Option<RegexWrapper>,
```
ameten, line 271: A little bit confusing name of the field `body` with rename `regex`. Possibly, `body_regex` and corresponding `bodyregex` are clearer.

kamiyaa: ah, good call. I'll change it to `body_regex`

PR #6666, opened 2025-07-02
feat(lander): Take into account processing time when sleeping in Inclusion Stage

ameten:
### Description

Inclusion Stage should run roughly every time when a new block is confirmed by a blockchain. We use estimated block time to make it happen at the moment. If the block time of a chain is low, processing time becomes significant compared to the block time. We take it into account in this change. We are reducing the time between every run of inclusion stage by the time it took to process the previous run.

### Backward compatibility

Yes

### Testing

None

### DIFF
--- a/rust/main/lander/src/dispatcher/stages/inclusion_stage.rs
+++ b/rust/main/lander/src/dispatcher/stages/inclusion_stage.rs
@@ -109,10 +109,15 @@ impl InclusionStage {
         domain: String,
     ) -> Result<(), LanderError> {
         let estimated_block_time = state.adapter.estimated_block_time();
+        let mut sleep_duration = *estimated_block_time;
         loop {
             // evaluate the pool every block
-            sleep(*estimated_block_time).await;
+            sleep(sleep_duration).await;
+
+            let before_processing = std::time::Instant::now();
             Self::process_txs_step(&pool, &finality_stage_sender, &state, &domain).await?;
+            let elapsed = before_processing.elapsed();
+            sleep_duration = estimated_block_time.saturating_sub(elapsed);
         }
     }
 


rust/main/lander/src/dispatcher/stages/inclusion_stage.rs
```diff
@@ -109,10 +109,15 @@ impl InclusionStage {
...
+        let mut sleep_duration = *estimated_block_time;
         loop {
             // evaluate the pool every block
-            sleep(*estimated_block_time).await;
+            sleep(sleep_duration).await;
+
+            let before_processing = std::time::Instant::now();
             Self::process_txs_step(&pool, &finality_stage_sender, &state, &domain).await?;
+            let elapsed = before_processing.elapsed();
+            sleep_duration = estimated_block_time.saturating_sub(elapsed);
```
daniel-savu, line 120: nice, can we implement `tx_ready_for_resubmission` in this PR as well, since they're closely related?

rust/main/lander/src/adapter/chains/ethereum/adapter.rs
```diff
@@ -123,26 +128,92 @@ impl EthereumAdapter {
         Ok(())
     }
 
-    async fn set_gas_price(&self, tx: &mut Transaction) -> Result<(), LanderError> {
+    async fn set_gas_price(&self, tx: &Transaction) -> Result<GasPrice, LanderError> {
```
daniel-savu, line 131: rename this, since it no longer `set`s

rust/main/lander/src/adapter/chains/ethereum/adapter.rs
```diff
@@ -474,14 +545,17 @@ impl AdaptsChain for EthereumAdapter {
...
-        let nonce = self.set_nonce_if_needed(&tx.clone()).await?;
+        let tx_for_nonce = tx.clone();
+        let tx_for_gas_price = tx.clone();
+
+        let (nonce, gas_price) = try_join!(
+            self.set_nonce_if_needed(&tx_for_nonce),
+            self.set_gas_price(&tx_for_gas_price)
+        )?;
+
         self.set_gas_limit_if_needed(tx).await?;
```
daniel-savu, line 556: I think we can remove this since it does nothing now

daniel-savu (approved):

rust/main/lander/src/dispatcher/stages/inclusion_stage.rs
ameten: probably, in another PR.

rust/main/lander/src/adapter/chains/ethereum/adapter.rs
ameten: Fixed

PR #6674, opened 2025-07-02
feat: allow for relative configurable lowest block height

kamiyaa:
### Description

- change `from` to be `i64`
- add logic to fetch latest finalized block if `from < 0`, then subtract

### Related issues

- https://linear.app/hyperlane-xyz/issue/ENG-1729/relative-configurable-lowest-block-height-or-sequence


rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -405,16 +405,22 @@ where
         // RPC provider does not provide historical blocks.
         // It should be used with care since it can lead to missing events.
         let from = index_settings.from;
-        let from = watermark
-            .map(|w| if w <= from {
+
+        let from = if from < 0 {
+            let finalized_block_number = self.indexer.get_finalized_block_number().await?;
+            (finalized_block_number as i64).saturating_add(from)
```
ameten, line 411: If the resulting `from` value will be below `watermark`, we should use `watermark` as starting point.

rust/main/hyperlane-base/src/contract_sync/mod.rs
```diff
@@ -470,7 +476,7 @@ where
                 self.indexer.clone(),
                 Arc::new(self.store.clone()),
                 index_settings.chunk_size,
-                index_settings.from,
+                (index_settings.from as u64).min(u32::MAX as u64) as u32,
```
ameten, line 479: `ForwardBackwardSequenceAwareSyncCursor` should support negative `index_settings.from` values as well.

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -152,26 +163,28 @@ impl<T: Debug + Clone + Sync + Send + Indexable + 'static> BackwardSequenceAware
...
     }
 
     /// Gets the next sequence range to index.
     /// Only used in sequence mode.
-    fn get_next_sequence_range(
+    async fn get_next_sequence_range(
         &self,
         current_indexing_snapshot: &TargetSnapshot,
     ) -> Option<RangeInclusive<u32>> {
+        let lowest_block_height_or_sequence = self.get_lowest_block_height_or_sequence().await?;
```
ameten, line 176: I think that this method does not return the lowest sequence correctly.

rust/main/hyperlane-base/src/contract_sync/cursors/sequence_aware/backward.rs
```diff
@@ -200,15 +227,34 @@ impl<T: Debug + Clone + Sync + Send + Indexable + 'static> BackwardSequenceAware
...
+                    (latest_tip as i64).saturating_add(self.lowest_block_height_or_sequence);
+                if lowest_block_height < 0 {
+                    Some(0)
+                } else {
+                    Some(lowest_block_height as u32)
+                }
+            }
+            IndexMode::Sequence => sequence_count.map(|seq_count| {
+                let lowest_sequence_count =
+                    (seq_count as i64).saturating_add(self.lowest_block_height_or_sequence);
```
ameten, line 251: I think that cast is not needed here.

```suggestion
                    seq_count.saturating_add(self.lowest_block_height_or_sequence);
```

PR #6675, opened 2025-07-02
refactor: Add internal init function to HypERC20Collateral

larryob:
### Description
Add internal init function to `HypERC20Collateral`. This can be used by any contract that wants to inherit from HypERC20Collateral. 

### Drive-by changes

Removed a duplicate call to `Ownable_init` from `TokenBridgeCctp.intialize`. This function is already called in `_MailboxClient_initialize`.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal initialization logic for improved encapsulation in token contracts. No changes to public interfaces or user-facing behavior.


solidity/contracts/token/HypERC20Collateral.sol
```diff
@@ -55,6 +55,14 @@ contract HypERC20Collateral is MovableCollateralRouter {
...
     ) public virtual initializer {
+        _HypERC20_initialize(_hook, _interchainSecurityModule, _owner);
+    }
+
+    function _HypERC20_initialize(
+        address _hook,
+        address _interchainSecurityModule,
+        address _owner
+    ) internal {
         _MailboxClient_initialize(_hook, _interchainSecurityModule, _owner);
```
yorhodes, line 66: I dont totally see the value in this if there is no additional logic

yorhodes (approved):

PR #6701, opened 2025-07-07
chore: remove unused code

cmcewen:
### Description

LZ, portal, liquidity layer code is not used anywhere


solidity/storage.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 OUTPUT_PATH=${1:-storage}
-EXCLUDE="test|mock|interfaces|libs|upgrade|README|Abstract|Static"
+EXCLUDE="test|mock|interfaces|libs|upgrade|README|Abstract|Static|LayerZero|PolygonPos|Portal"
```
cmcewen, line 3: @yorhodes any suggestion for what to do other than this?

PR #6704, opened 2025-07-08
feat/bump zod to v4 [DRAFT]

xeno097:


paulbalaji:
good to share this for context as to why a move to v4 would be beneficial https://zod.dev/v4

PR #6715, opened 2025-07-09
feat: Add FileSubmitterStrategy

ltyu:

### DIFF
--- a/typescript/sdk/package.json
+++ b/typescript/sdk/package.json
@@ -61,7 +61,8 @@
   },
   "type": "module",
   "exports": {
-    ".": "./dist/index.js"
+    ".": "./dist/index.js",
+    "./fs": "./dist/fs/index.js"
   },
   "types": "./dist/index.d.ts",
   "files": [
--- a/typescript/sdk/src/fs/index.ts
+++ b/typescript/sdk/src/fs/index.ts
@@ -1,0 +1,1 @@
+export * from './submitters/EV5FileSubmitter.js';
--- a/typescript/sdk/src/fs/submitters/EV5FileSubmitter.ts
+++ b/typescript/sdk/src/fs/submitters/EV5FileSubmitter.ts
@@ -1,0 +1,40 @@
+// eslint-disable-next-line
+import fs from 'fs';
+// eslint-disable-next-line
+import path from 'path';
+import { Logger } from 'pino';
+
+import { Annotated, ProtocolType, rootLogger } from '@hyperlane-xyz/utils';
+
+import { ProtocolTypedTransaction } from '../../providers/ProviderType.js';
+import { TxSubmitterInterface } from '../../providers/transactions/submitter/TxSubmitterInterface.js';
+import { TxSubmitterType } from '../../providers/transactions/submitter/TxSubmitterTypes.js';
+import { EV5FileTxSubmitterProps } from '../../providers/transactions/submitter/ethersV5/types.js';
+
+export class EV5FileSubmitter
+  implements TxSubmitterInterface<ProtocolType.Ethereum>
+{
+  txSubmitterType: TxSubmitterType = TxSubmitterType.FILE;
+  protected readonly logger: Logger = rootLogger.child({
+    module: 'file-submitter',
+  });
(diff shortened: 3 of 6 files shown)


.changeset/thick-fishes-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperlane-xyz/sdk": minor
+---
+
+Add FileSubmitter to SDK, which exports through @hyperlane-xyz/sdk/fs
```
ltyu, line 5: todo: update this

typescript/sdk/src/providers/transactions/submitter/builder/types.ts
```diff
@@ -15,77 +15,89 @@ export const SubmissionStrategySchema = z
 
 export type SubmissionStrategy = z.infer<typeof SubmissionStrategySchema>;
 
-export const ChainSubmissionStrategySchema = z.preprocess(
+export function preprocessChainSubmissionStrategy<
+  T extends { submitter: { type: string } },
```
ltyu, line 19: This only extends 
`{ submitter: { type: string } }` 
instead of 
`{ submitter: SubmitterMetadata }`

because SubmitterMetadata is actually a union, which will get wider as the caller extends it, leading to type errors. alternatively, we can use `{ submitter: any }`

typescript/cli/src/submitters/EV5FileSubmitter.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { appendYamlOrJson } from '../utils/files.js';
+
+import { CustomTxSubmitterType, EV5FileTxSubmitterProps } from './types.js';
+
+export class EV5FileSubmitter
+  implements TxSubmitterInterface<ProtocolType.Ethereum>
+{
+  txSubmitterType: TxSubmitterType =
+    CustomTxSubmitterType.FILE as TxSubmitterType;
```
ltyu, line 18: todo: try to fix `as`

typescript/cli/examples/submit/strategy/file-strategy.yaml
```diff
@@ -0,0 +1,8 @@
+basefork:
+  submitter:
+    type: file
+    filepath: './generated/transactions/base_arbitrum_combined.json'
+arbitrumfork:
+  submitter:
+    type: file
+    filepath: './generated/transactions/base_arbitrum_combined.json'
```
xeno097, line 8: Probably some doc comments would help here

typescript/cli/src/submitters/EV5FileSubmitter.ts
xeno097: I don't know if it will work but maybe what you can do is update the `TxSubmitterInterface` defintion to:
```typescript
export interface TxSubmitterInterface<TProtocol extends ProtocolType, TSubmitterType extends TxSubmitterType = TxSubmitterType> { ...
```

and then provide the extended enum as second generic type in the `EV5FileSubmitter` def

ltyu: i dont think its possible to extend the enum in ts.

i've updated to extend string, but default to `TxSubmitterType`

xeno097 (approved):

typescript/cli/src/deploy/warp.ts
```diff
@@ -933,9 +938,14 @@ export async function getSubmitterByStrategy<T extends ProtocolType>({
...
     submitter: await getSubmitterBuilder<T>({
-      submissionStrategy: SubmissionStrategySchema.parse(submissionStrategy),
+      submissionStrategy: submissionStrategy as SubmissionStrategy, // TODO: fix this
       multiProvider,
       registry,
+      additionalSubmitterFactories: {
+        file: (_multiProvider: MultiProvider, metadata: any) => {
+          return new EV5FileSubmitter(metadata);
+        },
+      },
```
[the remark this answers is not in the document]
ltyu: neither are correct solutions. 

due to time constraints, i will address in this a separate PR. i think we will be okay given that there are runtime checks in getStrategy (throws if strategy factory doesnt exist)

PR #6718, opened 2025-07-10
fix: Measure init time correctly in Relayer

ameten:
### Description

* Fix for application context verifiers (init from destination chain)
* Measure init time for ccip signers separately from message contexts

### Backward compatibility

Yes

### Testing

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization sequence for more reliable startup.
  * Enhanced debug logging during agent initialization for better traceability.
  * Updated application operation verifiers to use destination chains instead of origin chains.

### DIFF
--- a/rust/main/agents/relayer/src/relayer.rs
+++ b/rust/main/agents/relayer/src/relayer.rs
@@ -154,10 +154,11 @@ impl BaseAgent for Relayer {
         Self::reset_critical_errors(&settings, &chain_metrics);
 
         let start = Instant::now();
-        let mut start_entity_init = Instant::now();
 
         let core = settings.build_hyperlane_core(core_metrics.clone());
-        let db = DB::from_path(&settings.db)?;
+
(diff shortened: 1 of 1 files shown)


rust/main/agents/relayer/src/relayer.rs
```diff
@@ -1255,7 +1256,7 @@ impl Relayer {
         chain_metrics: &ChainMetrics,
     ) -> HashMap<HyperlaneDomain, Arc<dyn ApplicationOperationVerifier>> {
         settings
-            .build_application_operation_verifiers(settings.origin_chains.iter(), core_metrics)
+            .build_application_operation_verifiers(settings.destination_chains.iter(), core_metrics)
```
daniel-savu, line 1259: given that in practice `destination_chains` is identical to `origin_chains`, will this change any behavior?

daniel-savu (approved):

PR #6719, opened 2025-07-10
feat: Update getSubmitter to be stateful

ltyu:
### Description
This PR updates getSubmitter() to be stateful, which reads from a submitter registry. It also exports the helper methods. 

More description in linear task.

### Related issues
https://linear.app/hyperlane-xyz/issue/ENG-1910/refactor-getsubmitter

### Backward compatibility
Yes, accessing these methods remain the same (through getSubmitte)

### Testing
Existing Unit Tests

## Summary by CodeRabbit

* **New Features**
  * Added the ability to dynamically register new submitter types.
  * Enhanced flexibility for obtaining submitters through a registry-based system.

* **Refactor**
  * Improved the internal logic for retrieving submitters, making it more extensible and maintainable.


xeno097 (commented):
wdyt of instead of having a singelton for the submitter registry, have the `getSubmitter` func optionally receive a submitter registry override object and uses that in the get submitter? and have our current submitters as  the "default ones", so that if a submitter of a certain type is not defined we use our default ones and if not the use the suer supplied. I'm not a fan of using global variables and modifying them as it could lead to wierd bugs if we are not careful

PR #6726, opened 2025-07-11
chore: starknet routing ism fix

troykessler:
### Description

Improve deploy performance by executing all routing ism transactions in one


yjamin:
These changes have were included in the following PR #6738

PR #6735, opened 2025-07-11
feat: timelock submitter [ENG-1855]

xeno097:
### Description

This PR implements the Timelock submitter strategy to allow  users to propose and execute transactions using timelock contracts

### Drive-by changes

- Moves the ica submitter types in the same file as the other evm submitter configs
- Moves the warp apply ica tests to the new test file

### Related issues

- Fixes:
- [ENG-1855](https://linear.app/hyperlane-xyz/issue/ENG-1855/sdk-cli-implement-timelock-submitter)
- [ENG-1822](https://linear.app/hyperlane-xyz/issue/ENG-1822/create-timelocksubmitter)

### Backward compatibility

- YES

### Testing

- Manual
- e2e

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -301,6 +301,7 @@ jobs:
           # Other commands
           - relay
           # Warp Commands
+          - warp-apply-submitters
           - warp-apply
           - warp-bridge-1
           - warp-bridge-2
--- a/typescript/cli/examples/submit/strategy/json-rpc-timelock-strategy.yaml
+++ b/typescript/cli/examples/submit/strategy/json-rpc-timelock-strategy.yaml
@@ -1,0 +1,9 @@
+anvil3:
+  submitter:
+    type: timelockController
+    chain: anvil3
+    # Address of the timelock contract
+    timelockAddress: '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266'
+    # Configuration for how to propose this trasnaction on chain
+    proposerSubmitter:
+      type: jsonRpc
--- a/typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5TimelockSubmitter.ts
+++ b/typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5TimelockSubmitter.ts
@@ -51,11 +51,12 @@ export class EV5TimelockSubmitter
       provider,
     );
 
-    const minDelay = await timelockInstance.getMinDelay();
-
-    const delay = config.delay ?? minDelay.toBigInt();
-
-    assert(delay >= minDelay.toBigInt(), '');
+    const minDelay = (await timelockInstance.getMinDelay()).toBigInt();
+    const delay = config.delay ?? minDelay;
+    assert(
+      delay >= minDelay,
+      `Expected user supplied delay ${delay} to be greater or equal than the configured mindDelay ${minDelay}`,
+    );
 
     const internalSubmitter = await getSubmitter<ProtocolType.Ethereum>(
       multiProvider,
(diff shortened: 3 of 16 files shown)


.github/workflows/test.yml
```diff
@@ -301,6 +301,7 @@ jobs:
           # Other commands
           - relay
           # Warp Commands
+          - warp-apply-submitters
```
paulbalaji, line 304: from a different PR?

typescript/sdk/src/metadata/customZodTypes.ts
```diff
@@ -18,3 +18,17 @@ export const ZHash = z
...
+export const ZBigNumberish = z
+  .bigint()
+  .or(ZUint)
+  .or(z.string().regex(/^[0-9]+$/))
+  .transform(BigInt);
+
+export const ZBytes32String = z
+  .string()
+  .regex(
+    /^0x[0-9a-fA-F]{64}$/,
```
ltyu, line 31: isn't this the same as ZHash?

typescript/cli/examples/submit/strategy/json-rpc-timelock-strategy.yaml
```diff
@@ -0,0 +1,17 @@
+anvil3:
+  submitter:
+    type: timelockController
+    chain: anvil3
+    # Address of the deployed timelock
+    timelockAddress: '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266'
+    # Optional, the salt to be used for computing the operation id
```
ltyu, line 7: how come these are optional? aren't they required in order to generate the id?

ltyu (commented):
very initial review. think it makes sense so far

typescript/cli/examples/submit/strategy/json-rpc-timelock-strategy.yaml
xeno097: By default they are set to the 0 hash if not provided

ltyu (approved):

typescript/sdk/src/providers/transactions/submitter/ethersV5/EV5TimelockSubmitter.ts
```diff
@@ -0,0 +1,182 @@
...
+        'TimelockController transactions should have all the same destination chain',
+      );
+    }
+
+    // If no domain id is set on the transactions
+    // assume they are to be sent on the current configured chain
+    const [domainId] = transactionChains.values();
+    const destinationChain = this.multiProvider.getDomainId(
+      domainId ?? this.config.chain,
+    );
```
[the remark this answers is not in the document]
xeno097: I'd prefer to use the domainId here because as you say our native way of identifying chains is the domain id while on chain stuff relies on the chain id, so i'd say it is the role of the code submitting the tx on chain to make sure we are using the chain id and not the domain id. Considering that this submitter delegates tx submission to something else I think it should be fine in this case, wdyt?

typescript/sdk/src/providers/transactions/submitter/ethersV5/types.ts
```diff
@@ -41,7 +49,60 @@ export type EV5ImpersonatedAccountTxSubmitterProps = z.infer<
...
+  destinationInterchainAccountRouter?: Address;
+  interchainSecurityModule?: Address;
+  internalSubmitter: EvmSubmitterMetadata;
+};
+
+// @ts-expect-error due to zod3 type inference logic even if the
+// EV5GnosisSafeTxBuilderPropsSchema defines the version field with a default value
+// it is inferred recursively as an optional field making typescript complain that
+// EvmSubmitterMetadataSchema can't be used here.
+export const EvmIcaTxSubmitterPropsSchema: z.ZodSchema<EvmIcaTxSubmitterProps> =
```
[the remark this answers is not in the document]
xeno097: resolved in dms

PR #6625, opened 2025-06-26
feat: Add Everclear TokenBridge

larryob:
### Description

Add Everclear token bridge. The signatures needed to create intents are added to storage by the bridge owner. Signatures are based off a `(inputAddress, feeInInputTokenUnits, deadline)` tuple. We are using the same fee for all transfers, so we simply store the fee data in a single struct (containing the everclear signature) in storage.

## Summary by CodeRabbit

* **New Features**
  * Introduced a token bridge contract enabling cross-chain asset transfers via the Everclear intent-based system.
  * Added support for configuring output assets and fee parameters for different destination domains.
  * Provided new interfaces for Everclear adapters and Wrapped Ether (WETH) functionality.

* **Chores**
  * Added a new Arbitrum RPC endpoint to the configuration for improved network support.

* **Tests**
  * Implemented comprehensive unit and integration tests, including fork tests on Arbitrum, to ensure robust token bridge functionality.

* **Documentation**
  * Added deployment and interaction scripts as examples for using the new token bridge.

PR #6736, opened 2025-07-11
audit q3 2025

yorhodes:
- **refactor: Remove ValueTransferBridge (#6605)**
- **fix: CCTP burn message sender patch (#6632)**
- **fix: Fix absolute imports (#6661)**
- **Disable registry warp ID check**
- **feat: emit event on protocol fee payment (#6688)**
- **feat: token bridge fees (#6562)**
- **fix: rebalance undercollateralization (#6588)**
- **feat: Add Everclear TokenBridge (#6625)**
- **feat: collateral LP interface (#6589)**
- **refactor: Add internal init function to HypERC20Collateral (#6675)**


.github/workflows/storage-analysis.yml
```diff
@@ -2,7 +2,8 @@ name: Check Storage Layout Changes
 
 on:
   pull_request:
-    branches: [main]
+    branches:
+      - '*'
```
paulbalaji: can remove the `branches:` option entirely if we're gonna do `*`

solidity/contracts/PackageVersioned.sol
```diff
@@ -7,5 +7,5 @@ pragma solidity >=0.6.11;
  **/
 abstract contract PackageVersioned {
     // GENERATED CODE - DO NOT EDIT
-    string public constant PACKAGE_VERSION = "9.0.13";
+    string public constant PACKAGE_VERSION = "10.0.0-beta.0";
```
paulbalaji: probably don't want to include this in the diff?

solidity/core-utils/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './typechain/index.js';
 export * from './zksync/index.js';
 // GENERATED CODE - DO NOT EDIT
-export const CONTRACTS_PACKAGE_VERSION = '9.0.13';
+export const CONTRACTS_PACKAGE_VERSION = '10.0.0-beta.0';
```
paulbalaji: ditto probs don't want to check this in

PR #6744, opened 2025-07-15
feat(utils): implement convertToScaledAmount

Xaroz:
### Description

- Implement `convertToScaledAmount` which calculates the amount from the origin chain scaled to the destination chain
- Use this function in WarpCore for checking scaled amount
- Implement tests for this

### Backward compatibility

Yes
### Testing

Unit tests

## Summary by CodeRabbit

* **New Features**
  * Introduced a utility function for scaling amounts between different token scales.
* **Refactor**
  * Updated internal logic to use the new scaling utility for improved consistency.
* **Tests**
  * Added comprehensive tests to ensure correct scaling behavior across various scenarios.


typescript/utils/src/amount.ts
```diff
@@ -155,3 +155,26 @@ export function addBufferToGasLimit(
...
+ * Calculates the amount from the origin chain scaled to the destination chain
+ * This calculation is in line with the FungibleTokenRouter contract _outboundAmount
+ * and _inboundAmount functions
+ * @param fromScale The origin scale number.
+ * @param toScale The destination scale number.
+ * @param amount The number to scale.
+ * @param precisionFactor  Number used to get accurate conversion for smaller numbers.
+ * Take into account the resulting amount will be have this precision factor multiplied into it.
+ */
+export function convertToScaledAmount(
```
ltyu, line 169: maybe use an object parameter so we dont have to do `convertToScaledAmount(undefined, undefined, 100, 100)`

typescript/utils/src/amount.ts
```diff
@@ -155,3 +155,26 @@ export function addBufferToGasLimit(
...
+  fromScale: number | undefined,
+  toScale: number | undefined,
+  amount: bigint,
+  precisionFactor: number,
+) {
+  if (!fromScale || !toScale || fromScale === toScale) return amount;
```
ltyu, line 175: would defauling these to some number (e.g. 1) reduce these conditionals?

Xaroz: not necessarily, we would still wanna check if one of the scales are `undefined`, for example if we defaulted the values to 1

`convertToScaledAmount(6, undefined, 100n, 100)` would not produce the result we want

typescript/utils/src/amount.ts
Xaroz: happy to do this but wondering if we are sure, because from what I can see most of (if not all) our functions do not use object parameters (unless it is from the `object.ts` file), I wanted to remain consistent

PR #6750, opened 2025-07-15
refactor: Send funds directly to recipient for ERC20 intents

larryob:
### Description

Send funds directly to recipient for ERC20 intents in `EverclearTokenBridge`. In `EverclearEthBridge` we still send funds to the remote router and use a hyperlane message to transfer funds to the true recipient.

### DIFF
--- a/solidity/contracts/token/bridge/EverclearEthBridge.sol
+++ b/solidity/contracts/token/bridge/EverclearEthBridge.sol
@@ -41,6 +41,20 @@ contract EverclearEthBridge is EverclearTokenBridge {
         )
     {}
 
+    function _getReceiver(
+        uint32 _destination,
+        bytes32 _recipient
+    ) internal view override returns (bytes32) {
+        return _mustHaveRemoteRouter(_destination);
+    }
+
+    function _getIntentCalldata(
+        bytes32 _recipient,
+        uint256 _amount
+    ) internal view override returns (bytes memory) {
+        return abi.encode(_recipient, _amount);
+    }
+
     function quoteTransferRemote(
         uint32 _destination,
         bytes32 _recipient,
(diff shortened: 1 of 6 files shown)


solidity/contracts/token/bridge/EverclearEthBridge.sol
```diff
@@ -92,4 +106,40 @@ contract EverclearEthBridge is EverclearTokenBridge {
...
+
+        /* EFFECTS */
+        intentSettled[intentId] = true;
+        emit ReceivedTransferRemote(_origin, _recipient, _amount);
+
+        /* INTERACTIONS */
+        _transferTo(_recipient.bytes32ToAddress(), _amount);
+    }
+
+    receive() external payable {}
```
Mo-Hussain, line 144: Should we restrict this to only allow receiving eth from WETH withdraw?

larryob: Hmm, the amount to transfer to the user is always taken from the intent, so any donated amounts will be ignored. Given that, I don't think checking that the msg.sender is WETH is needed.

yorhodes (commented):
is the patch coverage accurate?

solidity/contracts/token/bridge/EverclearEthBridge.sol
Mo-Hussain: maybe something like this

```
receive() external payable {
    require(msg.sender == address(wrappedToken), "EEB: Only WETH can send ETH");
}
```

solidity/contracts/token/bridge/EverclearTokenBridge.sol
```diff
@@ -212,7 +198,7 @@ contract EverclearTokenBridge is HypERC20Collateral {
         // We always send the funds to the remote router, which will then send them to the recipient in _handle
         (, IEverclear.Intent memory intent) = everclearAdapter.newIntent({
             _destinations: destinations,
-            _receiver: _mustHaveRemoteRouter(_destination),
+            _receiver: _getReceiver(_destination, _recipient),
```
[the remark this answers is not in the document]
larryob: `_getIntentCalldata` has been removed.

solidity/contracts/token/bridge/EverclearEthBridge.sol
larryob: Done

PR #6754, opened 2025-07-16
feat(HttpRegistry): Add getWarpDeployConfig route

ltyu:
### Description
Related PR: https://github.com/hyperlane-xyz/hyperlane-registry/pull/1050

This PR updates HttpServer to serve getWarpDeployConfig route

### Backward compatibility
Yes

### Testing
Manual/Unit Tests
- tested with warp apply

## Summary by CodeRabbit

* **New Features**
  * Added new API endpoints to retrieve warp route core and deployment configurations by ID.

* **Tests**
  * Updated tests to reflect endpoint and method renaming for core configurations.
  * Introduced tests for the new deploy configuration endpoint and related service logic.
  * Added new mock data for deployment configurations to support testing.


typescript/http-registry-server/src/routes/warp.ts
```diff
@@ -7,7 +7,24 @@ import { WarpService } from '../services/warpService.js';
...
+    validateRequestParam('id', z.string()),
+    async (req: Request, res: Response) => {
+      const warpRoute = await warpService.getWarpDeployConfig(req.params.id);
+      res.json(warpRoute);
+    },
+  );
+
+  // get warp core config
   router.get(
     '/*id',
```
ltyu, line 29: open to potentially updating this to `/core/*id` to be symmetric to the deploy route

Mo-Hussain: I think `/core/*id` makes sense. @xeno097 can you explain what you mean?

typescript/http-registry-server/src/services/warpService.ts
```diff
@@ -20,4 +20,14 @@ export class WarpService extends AbstractService {
       return warpRoute;
```
Mo-Hussain, line 20: We should rename this to `getWarpCoreConfig` if we make the core change

typescript/http-registry-server/src/routes/warp.ts
```diff
@@ -7,7 +7,24 @@ import { WarpService } from '../services/warpService.js';
 export function createWarpRouter(warpService: WarpService) {
   const router = Router();
 
-  // get warp route
+  // get warp deploy config
+  router.get(
+    '/deploy/*id',
+    (req, res, next) => {
```
Mo-Hussain, line 13: nit: we should probably define this as a reusable middleware method

Mo-Hussain (approved):

typescript/http-registry-server/src/routes/warp.ts
xeno097: so i think we can add the `/core/*id` in this pr and leave the `/*id` endpoint too until we make the changes where is needed, basically leaving the original endpoint as an alias and then remove it once the clients of this code have been updated if any. If there aren't any current users of this endpoint then we can just rename it here

typescript/http-registry-server/src/services/warpService.ts
xeno097: +

typescript/http-registry-server/src/routes/warp.ts
ltyu: as discussed offline with @Mo-Hussain, even though this is a breaking change and @xeno097 suggestion is good, i believe we can just update this route as we don't use this route outside of the HttpClientRegistry. 

going forward, i think it'll make sense to have some plan to update routes such as the one suggested.

typescript/http-registry-server/src/services/warpService.ts
ltyu: fixed

PR #6759, opened 2025-07-17
chore: update gas prices

Mo-Hussain:
### Description

- update gas prices


typescript/infra/config/environments/mainnet3/gasPrices.json
paulbalaji, line 1: i very rarely change this and only touch the token prices

typescript/infra/config/environments/mainnet3/gasPrices.json
```diff
@@ -48,35 +48,35 @@
     "decimals": 9
   },
   "botanix": {
-    "amount": "0.000000021",
+    "amount": "0.000152321",
     "decimals": 9
   },
   "flame": {
-    "amount": "42.393213952",
+    "amount": "11.0",
```
paulbalaji, line 55: ditto

typescript/infra/config/environments/mainnet3/gasPrices.json
```diff
@@ -48,35 +48,35 @@
...
     "decimals": 9
   },
   "base": {
-    "amount": "0.007305675",
+    "amount": "0.015851751",
     "decimals": 9
   },
   "berachain": {
-    "amount": "0.00000103",
+    "amount": "0.000000073",
```
paulbalaji, line 71: does this need to be applied?

paulbalaji (commented):
generally the process is to update tokenPrices regularly and gasPrices irregularly

gasPrices is only updated if there is a big discrepancy on average

typescript/infra/config/environments/mainnet3/gasPrices.json
Mo-Hussain: Ah right, I only applied the change to ethereum and not anything else

PR #6763, opened 2025-07-17
fix: skip `EDGEN`

christopherbrumm:
### Description

fix: skip `EDGEN`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated route checking to skip the "EDGEN" warp route, preventing it from being processed during checks.


paulbalaji (approved):

paulbalaji:
note: you don't need to pull main every time there's new changes, the merge queue's explicit purpose is to avoid pulling main on every change as that will test this PR ontop of the latest main at the time of queuing

PR #6776, opened 2025-07-18
feat: update timelock deployer

xeno097:
### Description

allows to set cancellers when deploying timelocks

### Drive-by changes

- no

### Backward compatibility

-yes

### Testing

unit

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying cancellers in timelock configurations, allowing explicit assignment of who can cancel scheduled transactions.
  * Executors are now optional in timelock configurations; if not provided, anyone can execute scheduled transactions.
* **Bug Fixes**
  * Improved validation and error reporting for missing and unexpected canceller roles in timelock configurations.
* **Tests**
  * Enhanced test coverage for role assignments, executor/canceller configurations, multi-chain deployments, and transaction execution permissions in timelock deployments.
* **Chores**
  * Introduced a new constant representing a 32-byte zero-filled value for internal use.

### DIFF
--- a/typescript/infra/src/utils/timelock.ts
+++ b/typescript/infra/src/utils/timelock.ts
@@ -82,7 +82,9 @@ export async function timelockConfigMatches({
     // Ensure the cancellers have the CANCELLER_ROLE
     // by default proposers are also cancellers
     const expectedCancellers =
-      expectedConfig.cancellers ?? expectedConfig.proposers;
+      expectedConfig.cancellers && expectedConfig.cancellers.length !== 0
+        ? expectedConfig.cancellers
+        : expectedConfig.proposers;
     const cancellerRoles = await Promise.all(
       expectedCancellers.map(async (canceller) => {
         return timelock.hasRole(CANCELLER_ROLE, canceller);
--- a/typescript/infra/src/utils/timelock.ts
+++ b/typescript/infra/src/utils/timelock.ts
@@ -98,6 +98,27 @@ export async function timelockConfigMatches({
         `Canceller missing role for ${chain} at ${address}: ${cancellerMissing.join(', ')}`,
       );
     }
+
+    // Ensure the proposers that are not in the cancellers array
+    // do not have the CANCELLER_ROLE
+    const proposersThatShouldNotBeCancellers = expectedConfig.proposers.filter(
+      (proposer) => !expectedCancellers.includes(proposer),
+    );
+
+    const proposersThatShouldNotBeCancellersRoles = await Promise.all(
+      proposersThatShouldNotBeCancellers.map(async (proposer) => {
+        return timelock.hasRole(CANCELLER_ROLE, proposer);
+      }),
+    );
+
+    const extraCancellers = proposersThatShouldNotBeCancellersRoles.filter(
+      (_, i) => cancellerRoles[i],
+    );
+    if (extraCancellers.length > 0) {
+      issues.push(
+        `Proposers that should not be cancellers for ${chain} at ${address}: ${cancellerMissing.join(', ')}`,
+      );
+    }
   }
 
   return { matches: issues.length === 0, issues };
--- a/typescript/infra/src/utils/timelock.ts
+++ b/typescript/infra/src/utils/timelock.ts
@@ -1,5 +1,8 @@
+import { ethers } from 'ethers';
+
 import { TimelockController__factory } from '@hyperlane-xyz/core';
 import {
+  CANCELLER_ROLE,
   ChainMap,
   ChainName,
   EXECUTOR_ROLE,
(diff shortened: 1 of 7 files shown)


paulbalaji (commented):
needs changeset too

typescript/infra/src/utils/timelock.ts
```diff
@@ -71,6 +78,26 @@ export async function timelockConfigMatches({
         `Proposers missing role for ${chain} at ${address}: ${proposersMissing.join(', ')}`,
       );
     }
+
+    // Ensure the cancellers have the CANCELLER_ROLE
```
paulbalaji, line 82: If we have cancellers defined, we'll want to check both that the expected cancellers have the role and also that other proposers _do not_ have the role

typescript/sdk/src/timelock/evm/EvmTimelockDeployer.hardhat-test.ts
```diff
@@ -164,4 +171,187 @@ describe('EvmTimelockDeployer', async () => {
...
+      timelockAddress,
+      0,
+      testTxData,
+      _0_BYTES_32,
+      _0_BYTES_32,
+      0,
+    );
+    await tx.wait();
+
+    const randomSignerTimelockInstance = TimelockController__factory.connect(
```
nambrot, line 220: Nit: Be consistent in the random/other prefix usage

typescript/sdk/src/timelock/evm/EvmTimelockDeployer.hardhat-test.ts
```diff
@@ -164,4 +171,187 @@ describe('EvmTimelockDeployer', async () => {
...
+    const timelock = TimelockController__factory.connect(
+      timelockAddress,
+      multiProvider.getProvider(TestChainName.test4),
+    );
+
+    expect(
+      await timelock.hasRole(EXECUTOR_ROLE, hre.ethers.constants.AddressZero),
+    ).to.be.true;
+
+    const _0_BYTES_32 =
```
nambrot, line 197: Why is this calldata construction different from everywhere else? Imo for unit testing, you want to only change the one thing that is necessary for the test (which in this case is no executor in the config).

typescript/sdk/src/timelock/evm/EvmTimelockDeployer.hardhat-test.ts
```diff
@@ -164,4 +171,187 @@ describe('EvmTimelockDeployer', async () => {
...
+    // Timelock should still be the admin of itself
+    expect(
+      await timelockInstance.hasRole(
+        timelockAdminRoleHash,
+        timelockInstance.address,
+      ),
+    ).to.be.true;
+
+    // if an admin was set it should be the admin after the changes
+    if (expectedConfig.admin) {
```
nambrot, line 278: This seems unrelated to the function name/purpose of asserting canceller role assignments

typescript/sdk/src/timelock/evm/EvmTimelockDeployer.hardhat-test.ts
```diff
@@ -164,4 +171,187 @@ describe('EvmTimelockDeployer', async () => {
...
+  });
+
+  it('should set the expected admin from the config after applying the canceller changes', async () => {
+    const proposer = randomAddress();
+    const cancellerConfig: TimelockConfig = {
+      minimumDelay: 30,
+      proposers: [proposer],
+      executors: [signer.address],
+      cancellers: [proposer, randomAddress()],
+      admin: randomAddress(),
```
nambrot, line 341: Like my other comment, making the assertion through a shared function with a conditional inside that function makes these tests harder to reason about. You in effect have multiple dimensions you want to cross-test, so i would consider having multiple test groups with separate `beforeEach`s where you set the config, and then reusable, but explicit assertions about the on-chain state

nambrot (approved):
Going to approve the PR because functionality wise it looks good, but would recommend improving the test setup/readability

typescript/sdk/src/timelock/evm/EvmTimelockDeployer.hardhat-test.ts
```diff
@@ -3,165 +3,371 @@ import chai, { expect } from 'chai';
...
-    expect(await timelock.hasRole(PROPOSER_ROLE, config.proposers[0])).to.be
-      .true;
-    expect(await timelock.hasRole(EXECUTOR_ROLE, config.executors[0])).to.be
-      .true;
-    expect(await timelock.hasRole(timelockAdminRoleHash, config.admin!)).to.be
-      .true;
+        // proposers are also cancellers by default
+        for (const proposer of config.proposers) {
+          expect(await timelock.hasRole(CANCELLER_ROLE, proposer)).to.be.true;
+        }
```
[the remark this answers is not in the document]
xeno097: check the code again that piece has been removed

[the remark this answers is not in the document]
xeno097: check the test cases and you will se that no testCases where that comment is sets the cancellers array

typescript/sdk/src/timelock/evm/constants.ts
```diff
@@ -1,5 +1,8 @@
 import { keccak256 } from 'ethers/lib/utils.js';
 
+export const EMPTY_BYTES_32 =
```
paulbalaji, line 3: Surprised we don't have a const for this already?

paulbalaji (approved):

typescript/sdk/src/timelock/evm/constants.ts
xeno097: technically we do have it but it is called `COSMOS_NATIVE_ZERO_ADDR` or `STARKNET_ZERO_ADDR` so i'd say the name doesn't actually match the current use case ahahaa

PR #6779, opened 2025-07-21
fix: solana web3js external provider compatibility

amit-starkware:
### Description

added `.compileMessage()` to transaction sent to `simulateTransaction` to fix an issue where `simulateTransaction` expected a Transaction of the same class prototype as the one used by `@hyperlane-xyz/sdk` however when provided with a `Transaction` created by an external provider that might have used a different version or package instance of `@solana/web3.js` it would cause an error.
 
### Backward compatibility

Yes

### Testing

Manual


shaharb-starkware (approved):

antigremlin (commented):
Looks great to the best of my knowledge, but I'll ask someone else to have a look as well.

PR #6780, opened 2025-07-21
fix: fix circular dependency import in sdk package relying on the registry

xeno097:
### Description

Removes `@hyperlane-xyz/registry` imports from the `sdk` package

### Drive-by changes

- None

### Backward compatibility

- No, the `getSubmitter` function signature and related code interface has changes

### Testing

- Manual
- e2e

## Summary by CodeRabbit

* **Refactor**
  * Updated various components to use a direct mapping of core addresses by chain instead of relying on a registry interface for chain-specific details.
  * Modified function and method signatures to accept the new address mapping structure.
  * Improved internal handling of chain addresses for account and transaction submitters.

* **Chores**
  * Added documentation for a major update to the SDK package.

### DIFF
--- a/.changeset/rich-pianos-watch.md
+++ b/.changeset/rich-pianos-watch.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/sdk": minor
+---
+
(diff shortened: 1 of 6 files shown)


antigremlin (approved):
This is great!

PR #6814, opened 2025-07-30
refactor: Add app operation verifier and ccip signers into destination chain init

ameten:
### Description

Add app operation verifier and ccip signers into destination chain init

### Related issues

- Contributes into https://linear.app/hyperlane-xyz/issue/ENG-1929/refactor-relayer-start-up-for-destination

### Backward compatibility

Yes

### Testing

None


kamiyaa (approved):
lgtm!

PR #6833, opened 2025-08-04
Xeno/everclear token bridge support eng 1918

xeno097:
### Description

This PR add support for deploying the new `EverclearEthBridge` and `EverclearTokenBridge` contracts

### Drive-by changes

- Updates the hardhat and anvil start config to skip the contract size limit (for now)
- Moves the `MockEverclearAdapter` from the `solidity/test/token/EverclearTokenBridge.t.sol` to its own file
- implements the `objDiff` function to filter out properties in an object when compared to one of the same type

### Related issues

- Fixes #[ENG-1918](https://linear.app/hyperlane-xyz/issue/ENG-1918/support-configuration-and-deployment-of-everclear-token-bridges)

### Backward compatibility

- YES

### Testing

- E2E
- unit

### DIFF
--- a/typescript/cli/src/tests/commands/helpers.ts
+++ b/typescript/cli/src/tests/commands/helpers.ts
@@ -6,6 +6,8 @@ import {
   ERC20Test,
   ERC20Test__factory,
   ERC4626Test__factory,
+  MockEverclearAdapter,
+  MockEverclearAdapter__factory,
   XERC20LockboxTest,
   XERC20LockboxTest__factory,
   XERC20VSTest,
@@ -475,6 +477,25 @@ export async function deployToken(
   await token.deployed();
 
   return token;
+}
+
+export async function deployEverclearBridgeAdapter(
+  privateKey: string,
+  chain: string,
+): Promise<MockEverclearAdapter> {
+  const { multiProvider } = await getContext({
+    registryUris: [REGISTRY_PATH],
+    key: privateKey,
+  });
+
+  multiProvider.setSigner(chain, new ethers.Wallet(privateKey));
+
+  const adapter = await new MockEverclearAdapter__factory(
+    multiProvider.getSigner(chain),
+  ).deploy();
+  await adapter.deployed();
+
+  return adapter;
 }
 
 export async function deploy4626Vault(
--- a/typescript/sdk/src/token/EvmERC20WarpModule.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpModule.ts
@@ -5,6 +5,7 @@ import { UINT_256_MAX } from 'starknet';
 import { zeroAddress } from 'viem';
 
 import {
+  EverclearTokenBridge__factory,
   GasRouter__factory,
   IERC20__factory,
   MailboxClient__factory,
@@ -18,6 +19,7 @@ import {
   Domain,
   EvmChainId,
   ProtocolType,
+  ZERO_ADDRESS_HEX_32,
   addressToBytes32,
   assert,
   deepEquals,
@@ -25,6 +27,8 @@ import {
   eqAddress,
   isObjEmpty,
   normalizeAddressEvm,
+  objDiff,
+  objKeys,
   objMap,
   promiseObjAll,
   rootLogger,
@@ -64,6 +68,7 @@ import {
   contractVersionMatchesDependency,
   derivedHookAddress,
   derivedIsmAddress,
+  isEverclearTokenBridgeConfig,
   isMovableCollateralTokenConfig,
 } from './types.js';
 
@@ -171,6 +176,8 @@ export class EvmERC20WarpModule extends HyperlaneModule<
         expectedConfig,
       )),
       ...this.createRemoveBridgesTxs(actualConfig, expectedConfig),
+      ...this.createAddRemoteOutputAddressTxs(actualConfig, expectedConfig),
+      ...this.createRemoveRemoteOutputAddressTxs(actualConfig, expectedConfig),
       ...this.createOwnershipUpdateTxs(actualConfig, expectedConfig),
       ...proxyAdminUpdateTxs(
         this.chainId,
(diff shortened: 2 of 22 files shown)


typescript/sdk/src/token/deploy.ts
```diff
@@ -133,6 +150,20 @@ abstract class TokenDeployer<
...
+      return [
+        config.token,
+        scale,
+        config.mailbox,
+        config.everclearBridgeAddress,
+      ];
+    } else if (isEverclearEthBridgeTokenConfig(config)) {
+      return [
+        config.wethAddress,
+        scale,
```
yorhodes: I believe this is removed from the constructor now
this highlights that we dont really have type safety on this code

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -552,6 +567,222 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+        destination: parseInt(domainId),
+        outputAsset: addressToBytes32(outputAsset),
+      }),
+    );
+
+    return [
+      {
+        chainId: this.multiProvider.getEvmChainId(this.chainId),
+        to: this.args.addresses.deployedTokenRoute,
+        annotation: `Adding "${Object.keys(assets)}" output assets for token "${this.args.addresses.deployedTokenRoute}" on chain "${this.chainName}"`,
```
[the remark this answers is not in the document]
yorhodes: ditto with the `chain` and `chainId` fields

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -552,6 +567,222 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+        data: EverclearTokenBridge__factory.createInterface().encodeFunctionData(
+          'setOutputAssetsBatch((uint32,bytes32)[])',
+          [assets],
+        ),
+      },
+    ];
+  }
+
+  createRemoveRemoteOutputAssetsTxs(
```
yorhodes: I think we have hashed this out before but im still surprised you prefer having the add sets and remove sets calculated in different places

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -552,6 +567,222 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+          )
+        );
+      },
+    );
+
+    return Object.entries(feesToSet).map(([domainId, feeConfig]) => {
+      const { deadline, fee, signature } = feeConfig;
+
+      return {
+        annotation: `Setting Everclear fee params for domain "${domainId}" on token "${this.args.addresses.deployedTokenRoute}" and chain "${this.chainName}"`,
```
yorhodes: a human readable deadline might be useful to include?

typescript/sdk/src/token/EvmERC20WarpModule.ts
```diff
@@ -1015,6 +1246,29 @@ export class EvmERC20WarpModule extends HyperlaneModule<
...
+      }
+    }
+
+    if (isEverclearTokenBridgeConfig(config)) {
+      const updateEverclearFeeParamsTxs =
+        warpModule.createUpdateEverclearFeeParamsTxs(actualConfig, config);
+
+      for (const tx of updateEverclearFeeParamsTxs) {
+        await multiProvider.sendTransaction(chain, tx);
+      }
```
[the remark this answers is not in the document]
yorhodes: and do we not wait for receipts here anymore?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -317,7 +327,45 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
...
+            const maybeEverclearTokenBridge =
+              EverclearTokenBridge__factory.connect(
+                warpRouteAddress,
+                this.provider,
+              );
+
+            await maybeEverclearTokenBridge.callStatic.everclearAdapter();
+
+            let everclearTokenType = TokenType.collateralEverclear;
+            try {
```
yorhodes: personally not a huge fan of these nested try/catch blocks as it becomes hard to read but I recognize that there is no obvious alternative

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -662,6 +710,114 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
...
+          deadline: deadline.toNumber(),
+          fee: fee.toNumber(),
+          signature,
+        };
+      }),
+    );
+
+    return {
+      everclearBridgeAddress,
+      // Remove unset domains from the output
```
yorhodes: personally would not expect this to be in the return block but rather at the variable declaration

yorhodes (approved):

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
xeno097: yeah hopefully we can revisit when we move the EVM code to the multi vm format

typescript/sdk/src/token/EvmERC20WarpModule.ts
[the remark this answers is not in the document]
xeno097: I think it is pretty useful as when we are updating multiple chains at the same time, the annotation message that gets logged might be interleaved with messages of other chains so we can explicitly see on which chan and token we are updating.

typescript/cli/src/tests/commands/helpers.ts
```diff
@@ -477,6 +479,25 @@ export async function deployToken(
...
+
+  multiProvider.setSigner(chain, new ethers.Wallet(privateKey));
+
+  const adapter = await new MockEverclearAdapter__factory(
+    multiProvider.getSigner(chain),
+  ).deploy();
+  await adapter.deployed();
+
+  return adapter;
+}
```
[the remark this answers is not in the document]
xeno097: makes it easy to get a signer connected to a provider with the test chain metadata

PR #6834, opened 2025-08-04
feat: migrate from tendermint to cometbft

kamiyaa:
### Description

- migrate from `tendermint` to `cometbft`

### DIFF
--- a/rust/main/chains/hyperlane-cosmos/src/libs/account.rs
+++ b/rust/main/chains/hyperlane-cosmos/src/libs/account.rs
@@ -1,11 +1,12 @@
+use cometbft::account::Id as TendermintAccountId;
+use cometbft::public_key::PublicKey as TendermintPublicKey;
 use cosmrs::{crypto::PublicKey, AccountId};
 use hyperlane_cosmwasm_interface::types::keccak256_hash;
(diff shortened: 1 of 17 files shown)


ameten:
General comment for many places - Let's use `cometbft` in the variable names and other places instead of `tendermint` to avoid confusion.

rust/main/chains/hyperlane-cosmos/src/libs/account.rs
```diff
@@ -1,11 +1,12 @@
...
+use cometbft::public_key::PublicKey as TendermintPublicKey;
 use cosmrs::{crypto::PublicKey, AccountId};
 use hyperlane_cosmwasm_interface::types::keccak256_hash;
-use tendermint::account::Id as TendermintAccountId;
-use tendermint::public_key::PublicKey as TendermintPublicKey;
 
 use crypto::decompress_public_key;
 use hyperlane_core::Error::Overflow;
 use hyperlane_core::{AccountAddressType, ChainCommunicationError, ChainResult, H256};
+use protobuf::reflect::ProtobufValue;
```
ameten, line 9: Unused import?

PR #6855, opened 2025-08-06
feat: update USDC/paradex rebalancer config

Mo-Hussain:


paulbalaji (approved):
would also like @ltyu to approve for his resolver knowledge

ltyu (approved):

PR #6857, opened 2025-08-06
chore: strategy resolver fix

ltyu:
### Description
This PR fixes chain resolver for `hyperlane submit` by adding the `STRATEGY` enum into the chain resolver switch.

### Drive-by changes
- Remove default and instead throw an error when a strategy is not found for a command

### Related issues
- Fixes https://linear.app/hyperlane-xyz/issue/ENG-2084/some-commands-throw-error-no-private-key-found-for-chain-celestia

### Backward compatibility
Yes - should not break existing commands

### Testing
Manual/Unit Tests


paulbalaji (approved):

typescript/cli/src/context/strategies/chain/MultiChainResolver.ts
```diff
@@ -56,7 +57,6 @@ export class MultiChainResolver implements ChainResolver {
       case ChainSelectionMode.CORE_DEPLOY:
         return this.resolveCoreDeployChains(argv);
       case ChainSelectionMode.DEFAULT:
-      default:
```
xeno097, line 59: Why remove the default fallback?

xeno097 (approved):

PR #6869, opened 2025-08-11
chore: Cosmos ISM and IGP infra command generators

mbreithecker:

### DIFF
--- a/typescript/infra/scripts/cosmos-helpers/generate-igp-commands.ts
+++ b/typescript/infra/scripts/cosmos-helpers/generate-igp-commands.ts
@@ -93,8 +93,8 @@ async function main() {
     let gasPrice = gasPriceSmallestUnit * originDestDecimalsRatio;
 
     // Split scaling factor equally between price and ratio
-    ratio = Math.round(ratio * 1000);
-    const gas_price = Math.round(gasPrice * 10000000);
+    ratio = Math.round(ratio);
+    const gas_price = Math.round(gasPrice * 1e10);
 
     console.log(`${entry.name}: ${entry.domain_id}`);
     console.log(


typescript/infra/config/environments/mainnet3/gasPrices.json
paulbalaji, line 1: do we need these gas/tokenprice changes?

typescript/infra/scripts/cosmos-helpers/generate-igp-commands.ts
paulbalaji, line 1: why create generators instead of calling the functionality directly?

typescript/infra/scripts/cosmos-helpers/generate-igp-commands.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (chains == undefined) {
+    throw new Error('Chains must be provided');
+  }
+
+  const gasPrices = JSON.parse(
+    await readFile('config/environments/mainnet3/gasPrices.json', 'utf-8'),
+  );
+  const tokenPrices = JSON.parse(
+    await readFile('config/environments/mainnet3/tokenPrices.json', 'utf-8'),
+  );
```
paulbalaji, line 48: can import at the top
```
import gasPrices from '../../config/environments/mainnet3/gasPrices.json' with { type: 'json' };
import tokenPrices from '../config/environments/mainnet3/tokenPrices.json' with { type: 'json' };
```

typescript/infra/config/environments/mainnet3/gasPrices.json
mbreithecker: reverted the changes

PR #6881, opened 2025-08-12
Improve logging when we cannot batch payloads for a chain

ameten:

### DIFF
--- a/rust/main/lander/src/adapter/chains/ethereum/adapter.rs
+++ b/rust/main/lander/src/adapter/chains/ethereum/adapter.rs
@@ -23,7 +23,10 @@ use hyperlane_base::{
     },
     CoreMetrics,
 };
-use hyperlane_core::{config::OpSubmissionConfig, ContractLocator, HyperlaneDomain, H256, U256};
+use hyperlane_core::{
+    config::OpSubmissionConfig, ChainCommunicationError, ContractLocator, HyperlaneDomain, H256,
+    U256,
+};
 use hyperlane_ethereum::multicall::BatchCache;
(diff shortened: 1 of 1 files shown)


rust/main/lander/src/adapter/chains/ethereum/adapter.rs
```diff
@@ -320,9 +323,12 @@ impl EthereumAdapter {
             .await
             .map(|(tx, f)| EthereumTxPrecursor::new(tx, f));
 
-        let Ok(multi_precursor) = multi_precursor else {
-            error!("Failed to batch payloads");
-            return vec![];
+        let multi_precursor = match multi_precursor {
```
yjamin, line 326: Why do we pattern match like this now, instead of keeping the destruction like before?

Can't we just do:

````rust
let Ok(multi_precursor) = multi_precursor else {
            error!(error = ?e, "Failed to batch payloads");
            return vec![];
}
```

rust/main/lander/src/adapter/chains/ethereum/adapter.rs
```diff
@@ -320,9 +323,12 @@ impl EthereumAdapter {
...
+            Ok(precursor) => precursor,
+            Err(e) => {
+                error!(error = ?e, "Failed to batch payloads");
+                return vec![];
```
yjamin, line 330: Would it maybe make sense to propagate errors? I don't have a lot of context on the lander, but if `build_batched_transaction` is fail-able it might make sense to return a `Result<_>` instead.

yjamin (approved):

PR #6885, opened 2025-08-12
feat(fees): Deploy and set Fees in Deployer

ltyu:
### Description
This PR adds the Fee deploy logic into token deployer. There are also changes the the Modules to support this deployment with some additional changes to the schema as mentioned in the **Drive-by changes**.

### Drive-by changes
This PR also refactors the schema such that a linear fee can be configured using `bps`.

### Related issues
- Fixes ENG-2070

### Backward compatibility
Yes, these additions should not impact existing warp deployments

### Testing
Unit Tests

### DIFF
--- a/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
+++ b/typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
@@ -214,18 +214,11 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
           error,
         );
       }
-      const hasTokenFeeInterface =
-        compareVersions(
-          tokenConfig.contractVersion!,
-          TOKEN_FEE_CONTRACT_VERSION,
-        ) >= 0;
 
-      const tokenFee = hasTokenFeeInterface
-        ? await this.fetchTokenFee(
-            warpRouteAddress,
-            await movableToken.domains(),
-          )
-        : undefined;
+      const tokenFee = await this.fetchTokenFee(
+        warpRouteAddress,
+        await movableToken.domains(),
+      );
 
       return {
         ...routerConfig,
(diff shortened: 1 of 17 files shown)


typescript/sdk/src/token/deploy.ts
```diff
@@ -594,6 +600,54 @@ export class HypERC20Deployer extends TokenDeployer<HypERC20Factories> {
...
+          );
+          tokenFeeInput.owner = config.owner;
+        }
+
+        if (isCollateralTokenConfig(config) && !tokenFeeInput.token) {
+          this.logger.debug(
+            `Token fee token not specified for ${config.type}, using router token ${config.token}`,
+          );
+          tokenFeeInput.token = config.token;
+        }
```
xeno097, line 631: Should we maybe enforce at warp deploy config level that the tokens match? or are we planning to have different token types to pay the fees?

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -206,13 +213,27 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
           error,
         );
       }
+      const hasTokenFeeInterface =
+        compareVersions(
+          tokenConfig.contractVersion!,
+          TOKEN_FEE_CONTRACT_VERSION,
+        ) >= 0;
```
xeno097, line 220: What do you think of moving this check into the `fetchTokenFee` method and here just assign the result of the method call to the `tokenFee` so you cna avoid the ternary below?

typescript/sdk/src/token/deploy.ts
xeno097: Also wondering if we should enforce at the warp deploy config level that the destinations of the token fee router are the connected chains of the warp route or a subset

ltyu: good idea. I'm going to follow this up in the next PR since this is getting big

typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
ltyu: done

typescript/sdk/src/token/deploy.ts
ltyu: actually, this is not explicitly defined in the product spec so will do this at a later time

PR #6887, opened 2025-08-13
feat: support inflight rebalance detection

Mo-Hussain:
### Description

- Supports inflight rebalance detection by implementing `WithInflightGuard` `IRebalancer` implementation
- `WithInflightGuard` uses the explorer api to query whether there are any messages that have not been delivered that initiated by the rebalancer and sent from the bridge contracts for given warp route.
- if there are any inflight rebalances, we don't initiate the rebalance
- `WithInflightGuard` is composed with `WithSemaphore` in case there are issue will scrapping in flight rebalances, this will ensure we will always wait the configured time window for the semaphore before initiating further rebalances. We may opt to remove this if we observer stability over time.

### Testing

Manual/Unit Tests/e2e

## Summary by CodeRabbit

* **New Features**
  * Adds inflight protection: rebalancing is skipped when an undelivered rebalance is detected via Explorer.
  * Supports optional Explorer endpoint configuration and now requires a rebalancer address via environment variables.
* **Refactor**
  * Consolidates test scaffolding into shared helpers and cleans up imports.
* **Tests**
  * Introduces unit tests for inflight guarding and expanded end-to-end tests covering Explorer-driven skip/proceed scenarios and existing concurrency behavior.


typescript/cli/src/tests/ethereum/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -400,6 +406,129 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
...
+    const server = http.createServer((req, res) => {
+      if (req.method === 'POST') {
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+        // Return a non-empty message_view to simulate inflight
+        res.end(JSON.stringify({ data: { message_view: [{ msg_id: '1' }] } }));
+      } else {
+        res.statusCode = 404;
+        res.end();
+      }
+    });
```
xeno097, line 421: maybe you can use this https://www.npmjs.com/package/fetch-mock  to mock the server so you don't have to spin up a real http server with dummy data

typescript/cli/src/tests/ethereum/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -400,6 +406,129 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
...
+              weighted: { weight: '25', tolerance: '0' },
+              bridge: ethers.constants.AddressZero,
+              bridgeLockTime: 1,
+            },
+            [CHAIN_NAME_3]: {
+              weighted: { weight: '75', tolerance: '0' },
+              bridge: ethers.constants.AddressZero,
+              bridgeLockTime: 1,
+            },
+          },
```
xeno097, line 443: nit: might be better to have the config defined into its own variable with the type annotation and then pass it to the write function, this way if the config changes because we add a new field or rename an already existing one the compiler will complain and it will be easier to debug why the test is failing for someone not familiar with the rebalancer code. 

```typescript
const config: RebalancerConfig = {...};


writeYamlOrJson(REBALANCER_CONFIG_PATH, config);
```

typescript/cli/src/tests/ethereum/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -400,6 +406,129 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
...
+              bridge: ethers.constants.AddressZero,
+              bridgeLockTime: 1,
+            },
+            [CHAIN_NAME_3]: {
+              weighted: { weight: '25', tolerance: '0' },
+              bridge: bridgeContract.address,
+              bridgeLockTime: 1,
+            },
+          },
+        },
```
xeno097, line 509: nit: ditto

typescript/cli/src/tests/ethereum/warp/warp-rebalancer.e2e-test.ts
```diff
@@ -400,6 +406,129 @@ describe('hyperlane warp rebalancer e2e tests', async function () {
...
+    const http = await import('http');
+    const server = http.createServer((req, res) => {
+      if (req.method === 'POST') {
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+        // Return empty message_view to simulate no inflight
+        res.end(JSON.stringify({ data: { message_view: [] } }));
+      } else {
+        res.statusCode = 404;
+        res.end();
+      }
```
xeno097, line 466: ditto

typescript/cli/src/tests/ethereum/commands/warp.ts
```diff
@@ -258,8 +258,10 @@ export function hyperlaneWarpRebalancer(
   destination?: string,
   amount?: string,
   key?: string,
+  explorerUrl?: string,
 ): ProcessPromise {
-  return $`${localTestRunCmdPrefix()} hyperlane warp rebalancer \
+  const rebalancerAddress = new Wallet(key ?? ANVIL_KEY).address;
```
xeno097, line 263: you can use the `ANVIL_DEPLOYER_ADDRESS` variable

typescript/cli/src/rebalancer/utils/ExplorerClient.ts
```diff
@@ -0,0 +1,93 @@
...
+        ) {
+          msg_id
+          origin_domain_id
+          destination_domain_id
+          sender
+          recipient
+          origin_tx_hash
+          origin_tx_sender
+          is_delivered
+        }
```
xeno097, line 65: do you have a rough estimate of how slow this query is in the worst case? do we have the necessary indexes on the explorer side to make it work fast? if not should we ask the backend team to add them?

typescript/cli/src/rebalancer/utils/ExplorerClient.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    if (!res.ok) {
+      let body: string | undefined;
+      try {
+        body = await res.text();
+      } catch (_e) {
+        // ignore
+      }
+      throw new Error(`Explorer query failed: ${res.status} ${body ?? ''}`);
+    }
```
xeno097, line 84: why not have this directly read the json body instead?

typescript/cli/src/rebalancer/utils/ExplorerClient.ts
```diff
@@ -0,0 +1,93 @@
+import type { Logger } from 'pino';
+
+export type InflightRebalanceQueryParams = {
+  bridges: string[];
+  domains: number[];
+  txSender: string;
+  limit?: number;
+};
+
+export class ExplorerClient {
```
xeno097, line 10: I'm curious to know if you considered alternatives to using our explorer. An alternative solution might be
store the message id or the transaction hash and the origin chain when sending a rebalance transaction and use the sdk to check if the message has been delivered (maybe this won't work with everclear?)

yorhodes: alternatively the the collateral balance view could be adjusted when a rebalance is initiated

xeno097: that sound like another really good alternative

typescript/cli/src/tests/ethereum/warp/warp-rebalancer.e2e-test.ts
Mo-Hussain: Will look into it, have we used this before?

typescript/cli/src/rebalancer/utils/ExplorerClient.ts
Mo-Hussain: I don't understand the collateral balance view @yorhodes, can you expand on this?

typescript/cli/src/rebalancer/utils/ExplorerClient.ts
Mo-Hussain: I've done some basic performance testing and the worst case is less than 0.5s. Happy to look into performance improvements separately

xeno097 (approved):

PR #6908, opened 2025-08-15
chore: update ousdt pruner to easily re-enable chains

paulbalaji:
### Description

chore: update ousdt pruner to easily re-enable chains
- to easily/automatically un-hide ousdt chains https://github.com/hyperlane-xyz/hyperlane-registry/pull/1094

### DIFF
--- a/typescript/infra/scripts/warp-routes/ousdt/prune-warp-config.ts
+++ b/typescript/infra/scripts/warp-routes/ousdt/prune-warp-config.ts
@@ -4,24 +4,30 @@ import yargs from 'yargs';
 import {
   TokenStandard,
   WarpCoreConfig,
+  getTokenConnectionId,
   parseTokenConnectionId,
 } from '@hyperlane-xyz/sdk';
+import { ProtocolType } from '@hyperlane-xyz/utils';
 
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/warp-routes/ousdt/prune-warp-config.ts
```diff
@@ -4,24 +4,30 @@ import yargs from 'yargs';
...
+const chainsToPrune = ['worldchain', 'linea'];
 
 function pruneProdConfig({ tokens, options }: WarpCoreConfig): WarpCoreConfig {
   const prunedTokens = tokens.map((token) => {
+    // first force-add all connections and then prune the ones we don't want
+    token.connections = tokens
+      .filter((t) => t.chainName !== token.chainName)
+      .map((t) => ({
+        token: getTokenConnectionId(
+          ProtocolType.Ethereum,
```
Xaroz, line 25: could use `getAddressProtocolType` if we ever plan to have non-EVM chains

paulbalaji: we'll cross that bridge when we get there :P (pun intended)

PR #6919, opened 2025-08-19
feat: stop RPC retries for permanent errors

antigremlin:
## Description

Optimizes SmartProvider error handling by stopping provider iteration and retries for CALL_EXCEPTION errors, which are permanent failures that indicate contract execution issues rather than transient network problems.

Introduces a new retry mechanism with `stopRetry` flag support that prevents wasted retries on permanent failures. When a CALL_EXCEPTION is encountered, the provider immediately stops trying additional providers and marks the error as non-retryable.

## Related Issues

TBD

## Backward Compatibility

Yes - this change only optimizes behavior for CALL_EXCEPTION errors while preserving existing retry and fallback logic for all other error types.

### DIFF
--- a/.changeset/twelve-humans-type.md
+++ b/.changeset/twelve-humans-type.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/sdk": patch
+---
+
+Optimized SmartProvider to stop retrying on CALL_EXCEPTION errors as they are permanent failures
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/providers/SmartProvider/SmartProvider.test.ts
```diff
@@ -0,0 +1,385 @@
...
+
+      const provider1 = MockProvider.error(serverError);
+      const provider2 = MockProvider.error(blockchainError);
+      const provider = new TestableSmartProvider([provider1, provider2]);
+
+      try {
+        await provider.simplePerform('getBlockNumber', 1);
+        expect.fail('Should have thrown an error');
+      } catch (e: any) {
+        expect(e.name).to.equal('BlockchainError'); // Should get blockchain error, not server error
```
xeno097, line 376: why not check by validating that the error is an instance of the `BlockchainError` instead?

PR #6925, opened 2025-08-20
feat: radix token adapter

yjamin:
### Description

This PR splits the Radix CLI PR #6795 and only implements the TokenAdapter on Radix.

There were some minor changes made to the code:
- more accurate gas estimation
- fixed radix address conversion
- don't request proof of account to get the account public key
  - this was used for tx simulation, however, we don't actually need the pubkey of the account and instead simulation tx with a random one and disable signature checks for simulation


yjamin:
This PR will be split even further into smaller pieces. See #6929

PR #6927, opened 2025-08-20
feat: deny unsafe arithmetic

kamiyaa:
### Description

- prevent unsafe arithmetic actions

### Related issues

- https://linear.app/hyperlane-xyz/issue/ENG-2032/deny-unchecked-arithmetic-operations

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved arithmetic safety across the codebase: counters, ranges, timings, retries/backoffs, serialization, gas/fee math and pagination now guard against overflow/underflow, reducing panics and edge-case failures.

- **Chores**
  - Enforced stricter linting (arithmetic-side-effect checks) and standardized safer arithmetic patterns across many crates.
  - A previously test-gated helper is now compiled in all builds.

### DIFF
--- a/rust/main/chains/hyperlane-cosmos/src/utils.rs
+++ b/rust/main/chains/hyperlane-cosmos/src/utils.rs
@@ -46,7 +46,8 @@ pub(crate) async fn get_block_height_for_reorg_period(
     };
 
     let tip = provider.latest_block_height().await?;
-    let block_height = tip.saturating_sub(period);
+    // Cosmos heights start at 1; avoid querying block 0.
+    let block_height = tip.saturating_sub(period).max(1);
     Ok(block_height)
 }
 


rust/main/agents/relayer/src/msg/pending_message.rs
```diff
@@ -926,23 +926,29 @@ impl PendingMessage {
             2 => 10,
             3 => 30,
             4 => 60,
-            i if (5..25).contains(&i) => 60 * 3,
+            i if (5..25).contains(&i) => 60u64.saturating_mul(3),
             // linearly increase from 5min to ~25min, adding 1.5min for each additional attempt
-            i if (25..40).contains(&i) => 60 * 5 + (i as u64 - 25) * 90,
+            i if (25..40).contains(&i) => 60u64
+                .saturating_mul(5)
+                .saturating_add((i as u64).saturating_sub(25).saturating_mul(90)),
```
yjamin, line 933: nit: I wonder if we can just type out the magic values here and add comments what they mean? But nothing I feel strong about

rust/main/agents/relayer/src/settings/mod.rs
```diff
@@ -159,18 +159,18 @@ impl FromRawConf<RawRelayerSettings> for RelayerSettings {
                 Ok(Some(parser)) => match parse_json_array(parser) {
                     Some((path, value)) => (path, value, true),
                     None => (
-                        &p.cwp + "gas_payment_enforcement",
+                        (&p.cwp).add("gas_payment_enforcement"),
```
yjamin, line 162: nit: maybe we want to use format! here as well to increase readability

yjamin: This would apply for all string operations

rust/main/agents/validator/src/submit.rs
```diff
@@ -386,7 +386,10 @@ impl ValidatorSubmitter {
     /// Signs and submits any previously unsubmitted checkpoints.
     async fn sign_and_submit_checkpoints(&self, mut checkpoints: Vec<CheckpointWithMessageId>) {
         // The checkpoints are ordered by index, so the last one is the highest index.
-        let last_checkpoint_index = checkpoints[checkpoints.len() - 1].index;
+        let last_checkpoint_index = match checkpoints.iter().last() {
```
yjamin, line 389: nit: could use `.last()` on the Vec directly.

yjamin: Looking at the usage of this method it is ensured that there is at least one element in `checkpoints`. Still think this is a good check.

yjamin (approved):

PR #6932, opened 2025-08-21
feat: use core provider to get state_version

yjamin:
### Description

I noticed a lot of warnings in the logs. We use 2 RPCs (Core & Gateway) for Radix chains and it  turns out that the Gateway usually reports a higher state_version (equal to block height) than the Core API. This results in errors, because we try to Query state from the higher, not available, state version. 

This PR now uses the Core API to query for the state version, which resolved all of the errors.

## Summary by CodeRabbit

- New Features
  - Added network status support in the Radix provider.
  - More accurate finalized block height and tip reporting across Radix indexers.

- Performance
  - Reduced overhead by passing binary arguments directly in Radix ISM and Mailbox calls.
  - Streamlined state-version lookups for faster, leaner queries.

- Bug Fixes
  - Improved stability and consistency of Radix reads during reorg scenarios.

- Refactor
  - Migrated indexing and provider logic to a state-version-based API for consistency across components.

### DIFF
--- a/rust/main/chains/hyperlane-radix/src/ism.rs
+++ b/rust/main/chains/hyperlane-radix/src/ism.rs
@@ -1,5 +1,5 @@
 use async_trait::async_trait;
-use scrypto::{data::manifest::manifest_encode, types::ComponentAddress};
+use scrypto::types::ComponentAddress;
 
 use hyperlane_core::{
     ChainResult, ContractLocator, Encode, HyperlaneChain, HyperlaneContract, HyperlaneDomain,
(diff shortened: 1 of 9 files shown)


rust/main/chains/hyperlane-radix/src/ism.rs
```diff
@@ -94,15 +93,14 @@ impl MultisigIsm for RadixIsm {
         message: &HyperlaneMessage,
     ) -> ChainResult<(Vec<H256>, u8)> {
         let message = message.to_vec();
-        let message = manifest_encode(&message).map_err(HyperlaneRadixError::from)?;
 
         let (validators, threshold): (Vec<EthAddress>, usize) = self
             .provider
-            .call_method(
+            .call_method_with_arg(
```
kamiyaa, line 99: this is just a better way to make calls?

kamiyaa (approved):

rust/main/chains/hyperlane-radix/src/ism.rs
yjamin: Yeah

PR #6940, opened 2025-08-25
chore: add radix typescript sdk

troykessler:
### Description

This PR adds a new radix-sdk package containing all the low-level logic for calling hyperlane transactions on the radix blockchain. This package exposes methods for submitting transactions and querying hyperlane contract states.

### Testing

Manual Testing on local network and Stokenet (their official testnet)


typescript/radix-sdk/src/modules/populate.ts
```diff
@@ -0,0 +1,580 @@
...
+    const details =
+      await this.gateway.state.getEntityDetailsVaultAggregated(
+        contract_address,
+      );
+
+    const ownerResource = (details.details as any).role_assignments.owner.rule
+      .access_rule.proof_rule.requirement.resource;
+
+    return new ManifestBuilder()
+      .callMethod(from_address, 'lock_fee', [decimal(this.gasAmount)])
```
xeno097, line 78: Considering that i see here multiple methods that have these hardcoded strings, what do you think of creating an enum with all the radix methods that we use so we can avoid magic strings? It would also help having some kind of doc comment or reference to the original doc to know what each one of these does

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+    amount,
+  }: {
+    to_address: string;
+    resource_address: string;
+    amount: string;
+  }) {
+    const transactionManifest = this.populate.transfer({
+      from_address: this.account.address,
+      to_address,
+      resource_address,
```
xeno097, line 56: Is there a way to determine if a resource exists on chain? we could check if it does exists and throw a human friendly error here if it doesn't? or does the transaction fail and the message is clear enough?

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+
+  private async getNewComponent(transaction: TransactionHash): Promise<string> {
+    const transactionReceipt =
+      await this.gateway.transaction.getCommittedDetails(transaction.id);
+
+    const receipt = transactionReceipt.transaction.receipt;
+    assert(receipt, `found no receipt on transaction: ${transaction.id}`);
+
+    const newGlobalGenericComponent = (
+      receipt.state_updates as any
```
xeno097, line 499: Is there any way we can avoid the any here? This type assertion could hide errors in case we upgrade the radix library to a newer version and there is a breaking change

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+  public async signAndBroadcast(
+    manifest: TransactionManifest,
+  ): Promise<TransactionHash> {
+    const constructionMetadata =
+      await this.gateway.transaction.innerClient.transactionConstruction();
+
+    const transactionHeader: TransactionHeader = {
+      networkId: this.networkId,
+      startEpochInclusive: constructionMetadata.ledger_state.epoch,
+      endEpochExclusive: constructionMetadata.ledger_state.epoch + 2,
```
xeno097, line 433: Is this similar to Solana's latest block height where if we reach this height then the transaction expires? if it is should we add some kind of retry logic that automatically updates this value and after X number of retries the transaction submission fails?

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+      nonce: generateRandomNonce(),
```
xeno097, line 434: Why is this random? would help having a comment or a link to the official docs explaining how the transaction is build

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+    };
+
+    const transaction: NotarizedTransaction =
+      await TransactionBuilder.new().then((builder) =>
+        builder
+          .header(transactionHeader)
+          .manifest(manifest)
+          .sign(this.signIntent)
+          .notarize(this.notarizeIntent),
+      );
```
xeno097, line 447: Can we not use `.then` and operate instead on the result of `TransactionBuilder.new()` please?

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+    await this.gateway.transaction.innerClient.transactionSubmit({
+      transactionSubmitRequest: {
+        notarized_transaction_hex: Buffer.from(
+          compiledNotarizedTransaction,
+        ).toString('hex'),
+      },
+    });
+    await this.query.pollForCommit(intentHashTransactionId.id);
+
+    return intentHashTransactionId;
```
xeno097, line 464: Does the radix sdk expose a way to simulate transactions before submitting them to avoid sending transactions that might fail?

typescript/radix-sdk/src/modules/tx.ts
```diff
@@ -0,0 +1,519 @@
...
+      token,
+      receiver_domain,
+      receiver_address,
+      gas,
+    });
+
+    await this.signAndBroadcast(transactionManifest);
+  }
+
+  public async unrollRemoteRouter({
```
xeno097, line 374: I think that there is a type here, shouldn't it be `unenroll...`?

typescript/radix-sdk/src/modules/tx.ts
troykessler: yes with the `query.estimateTransactionFee` method we can simulate it and get the estimated gas amount

typescript/radix-sdk/src/modules/tx.ts
troykessler: Unfortunately some properties just have the type "object" which says nothing, will add custom types instead of any

typescript/radix-sdk/src/modules/tx.ts
xeno097: I think it would be a good idea to add it then if it isn't too much work

typescript/radix-sdk/src/modules/populate.ts
```diff
@@ -0,0 +1,580 @@
...
+      await this.gateway.state.getEntityDetailsVaultAggregated(igp);
+
+    const resource = (details.details as any).role_assignments.owner.rule
+      .access_rule.proof_rule.requirement.resource;
+
+    return this.transfer({
+      from_address,
+      to_address: new_owner,
+      resource_address: resource,
+      amount: '1',
```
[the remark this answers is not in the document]
ltyu: can we add a comment here for future readers?

typescript/radix-sdk/src/modules/query.ts
```diff
@@ -0,0 +1,769 @@
...
+    );
+
+    const fee_summary: FeeSummary = (response.receipt as any).fee_summary;
+    const costing_parameters: CostingParameters = (response.receipt as any)
+      .costing_parameters;
+
+    const gasUnits =
+      BigInt(fee_summary.execution_cost_units_consumed) +
+      BigInt(fee_summary.finalization_cost_units_consumed);
+    const fee = BigInt(
```
ltyu, line 90: why not just use BigInt?

typescript/radix-sdk/src/modules/query.ts
```diff
@@ -0,0 +1,769 @@
...
+      new BigNumber(fee_summary.xrd_total_execution_cost)
+        .plus(BigNumber(fee_summary.xrd_total_finalization_cost))
+        .plus(BigNumber(fee_summary.xrd_total_storage_cost))
+        .times(new BigNumber(10).exponentiatedBy(18))
+        .toFixed(0),
+    );
+    const gasPrice =
```
ltyu, line 97: given that gasUnits and fee are both bigint, should this also be a bigint? i can imagine a caller want to do some arithmatic with these values and having to convert

typescript/radix-sdk/src/modules/query.ts
```diff
@@ -0,0 +1,769 @@
...
+    const { divisibility } = await this.getMetadata({ resource });
+
+    return BigInt(
+      new BigNumber(fungibleResource.vaults.items[0].amount)
+        .times(new BigNumber(10).exponentiatedBy(divisibility))
+        .toFixed(0),
+    );
+  }
+
+  public async getXrdBalance({
```
ltyu, line 181: this class feels way too big. can we split it up somehow? for example, the xrd methods seem to inherit from some base method (getBalance in this case)

ltyu (commented):
the classes within the modules folders feel big. would they fit into the existing Modules abstraction (Module and Reader)?

typescript/radix-sdk/src/modules/tx.ts
troykessler: don't know exactly, added a comment to the official example: https://github.com/radixdlt/typescript-radix-engine-toolkit?tab=readme-ov-file#constructing-transactions

ltyu:
> > the classes within the modules folders feel big. would they fit into the existing Modules abstraction (Module and Reader)?
> 
> They have nothing to do with our modules. The classes basically the result of already splitting up what the radix typescript sdk should be capabable of like creating mailboxes and querying tokens for example. In the modules abstractions like module & reader we would use this radix sdk package then and simply call `createMailbox(...)` and all the low level stuff that we don't want to have in the main sdk package is hidden in those "big classes"

what does "nothing to do with our modules" mean? the modules abstraction is just a way to organize the code into create, read, update. you mentioned "creating mailboxes" and "querying tokens", which is what module create and read does. For example, the `RadixPopulate` class creates manifests for isms, mailbox, hooks, enrolls routers, etc. it's doing the same-ish things with regards to different entities. seems separable

ltyu:
> > the classes within the modules folders feel big. would they fit into the existing Modules abstraction (Module and Reader)?
> 
> They have nothing to do with our modules. The classes basically the result of already splitting up what the radix typescript sdk should be capabable of like creating mailboxes and querying tokens for example. In the modules abstractions like module & reader we would use this radix sdk package then and simply call `createMailbox(...)` and all the low level stuff that we don't want to have in the main sdk package is hidden in those "big classes"

how does "creating mailboxes" have nothing to do with our modules?

troykessler:
> > > the classes within the modules folders feel big. would they fit into the existing Modules abstraction (Module and Reader)?
> > 
> > 
> > They have nothing to do with our modules. The classes basically the result of already splitting up what the radix typescript sdk should be capabable of like creating mailboxes and querying tokens for example. In the modules abstractions like module & reader we would use this radix sdk package then and simply call `createMailbox(...)` and all the low level stuff that we don't want to have in the main sdk package is hidden in those "big classes"
> 
> ~what does "nothing to do with our modules" mean? the modules abstraction is just a way to organize the code into create, read, update. you mentioned "creating mailboxes" and "querying tokens", which is what module create and read does. For example, the `RadixPopulate` class creates manifests for isms, mailbox, hooks, enrolls routers, etc. it's doing the same-ish things with regards to different entities. seems separable~ i re-read your message and its more clear to me.
> 
> fyi, im not saying you have to use the module abstraction. just saying the class could be split up in some way.

yep, split the query module class in half by creating a new subclass `RadixBase` which holds all basic functionalities

typescript/radix-sdk/src/modules/tx.ts
troykessler: added it

xeno097 (approved):
I'll approve but please wait for @ltyu to also have another look before merging

ltyu (approved):

PR #6945, opened 2025-08-26
chore: validator rotation

paulbalaji:
### Description

chore: validator rotation

### Drive-by changes

- update check-validator-announce script to display aliases
- update check-validator-announce to better check upper limits on thresholds

### Testing

check-validator-announce


typescript/sdk/src/consts/multisigIsm.ts
```diff
@@ -752,29 +784,32 @@ export const defaultMultisigConfigs: ChainMap<MultisigConfig> = {
   },
 
   ethereum: {
-    threshold: 6,
+    threshold: 5,
```
nambrot, line 787: Why set the threshold to 5 here?

PR #6946, opened 2025-08-26
feat: handle failed finalized transactions on sealevel

kamiyaa:
### Description
 - Handle the case where a sealevel transaction could be finalized/confirmed, but still failed.

### Related issue
- fixes https://linear.app/hyperlane-xyz/issue/ENG-2152/transaction-being-confirmed-without-actually-being-confirmed-on-chain

### Drive-by changes
 - Add a new reprepare reason: Manual. So manual retries are still honoured instead of just being sent back to confirm queue.

## Summary by CodeRabbit

- New Features
  - Added manual retry support for pending operations so manual re-prepares are retried.

- Bug Fixes
  - Confirm only truly submitted operations; not-yet-submitted ops are re-prepared.
  - Manually retried ops have attempts reset to avoid stuck workflows.
  - Sealevel (Solana) transaction checks now inspect transaction metadata to mark finalized/included failures as dropped with clearer logs.

- Tests
  - Updated test fixtures to include transaction status metadata for more realistic coverage.

### DIFF
--- a/rust/main/lander/src/adapter/chains/sealevel/adapter/tests/tests_common.rs
+++ b/rust/main/lander/src/adapter/chains/sealevel/adapter/tests/tests_common.rs
@@ -13,8 +13,9 @@ use solana_sdk::{
     transaction::Transaction as SealevelTransaction,
 };
 use solana_transaction_status::{
-    EncodedConfirmedTransactionWithStatusMeta, EncodedTransaction,
-    EncodedTransactionWithStatusMeta, UiConfirmedBlock,
+    option_serializer::OptionSerializer, EncodedConfirmedTransactionWithStatusMeta,
+    EncodedTransaction, EncodedTransactionWithStatusMeta, UiConfirmedBlock,
+    UiTransactionStatusMeta,
 };
 
 use hyperlane_base::settings::{ChainConf, RawChainConf};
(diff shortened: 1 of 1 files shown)


rust/main/lander/src/adapter/chains/sealevel/adapter.rs
```diff
@@ -250,25 +250,47 @@ impl SealevelAdapter {
...
+            if let Some(meta) = tx.transaction.meta {
+                // It is possible for a failed transaction to be finalized.
+                // In this case, we need to re-submit.
+                if meta.err.is_some() {
+                    warn!(?signature, ?tx_hash, "Transaction finalized, but failed");
+                    return Ok(TransactionStatus::Dropped(
+                        TransactionDropReason::DroppedByChain,
+                    ));
+                } else {
+                    info!(?tx_hash, "transaction finalized");
```
yjamin, line 268: Is meta only present if there is an error?

yjamin:
I don't have a lot of context on the submit flow for the relayer, but looks good to me. I'll have to mostly trust you on this one. 

As for the Manual state, will `reprioritize_match` only ever be called by someone manually? Or am I misreading something?

rust/main/lander/src/adapter/chains/sealevel/adapter.rs
kamiyaa: No, I think it should always be present. Unless the tx does not exist

yjamin (approved):

PR #6956, opened 2025-08-28
fix: only deploy supported chains

troykessler:
### Description

This PR fixes warp deploy so that it only deploys chains which are supported by the CLI (currently EVM and Cosmos Native). This was discovered by wanting to deploy a warp route between cosmos and solana where the cli failed to deploy to solana because it does not support it. The fix was simply filtering the chains for the supported protocols

### Testing

Manual Testing

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -122,14 +122,26 @@ export async function runWarpRouteDeploy({
 
   assert(multiProtocolSigner, `multiProtocolSigner not defined`);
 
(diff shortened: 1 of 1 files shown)


christopherbrumm (approved):

xeno097 (commented):
Probably this PR might have breaking changes in the way the config is exported after the deployment is complete, if you want to avoid the cli complaining you should add the`foreignDeployment` field in the warp deploy config. Example:  https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/deployments/warp_routes/ES/eclipse-deploy.yaml#L2

PR #6968, opened 2025-08-29
chore: scale down to only 1 AW validator per chain

paulbalaji:
### Description

chore: scale down legacy AW validators

### Related issues

https://linear.app/hyperlane-xyz/issue/ENG-2106/scale-down-to-1-aw-validator-on-all-chains


paulbalaji:
> I wonder whether or not any multisigs have to be updated?

none - they haven't been on default isms in years

PR #6976, opened 2025-09-02
chore: update agent image

kamiyaa:
### Description

Update agent images to include velodrome message blacklist


paulbalaji (commented):
You don't need to update image tags to update the blacklist - this is done at agent deploy time

kamiyaa:
> You don't need to update image tags to update the blacklist - this is done at agent deploy time

ah got it

PR #6980, opened 2025-09-03
feat: add /merkle_proofs endpoint to show proof

kamiyaa:
### Description

 - add `/merkle_proof` endpoint to generate a proof for a given leaf index and root index


yorhodes (approved):
nice

PR #6983, opened 2025-09-03
feat: gitleaks config update [ENG-2162]

xeno097:
### Description

Updates the gitleaks config to catch svm private keys in the current commit 

### Related issues

- Fixes [ENG-2162](https://linear.app/hyperlane-xyz/issue/ENG-2162/update-the-gitleaks-hook-to-detect-buffer-files-and-flag-them)

### Backward compatibility

- yes

### Testing

- Manual
- unit

### DIFF
--- a/typescript/infra/package.json
+++ b/typescript/infra/package.json
@@ -55,6 +55,7 @@
     "hardhat": "^2.22.2",
     "mocha": "^11.5.0",
     "prettier": "^3.5.3",
+    "smol-toml": "1.4.2",
     "tsx": "^4.19.1",
     "typescript": "5.3.3"
   },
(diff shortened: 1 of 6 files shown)


typescript/infra/package.json
```diff
@@ -55,6 +55,7 @@
     "hardhat": "^2.22.2",
     "mocha": "^11.5.0",
     "prettier": "^3.5.3",
+    "smol-toml": "1.4.2",
```
paulbalaji, line 58: where is this being used?

rust/main/config/mainnet_config.json
```diff
@@ -4029,7 +4029,7 @@
       "validatorAnnounce": "0xf57d954bf3ddb5f1032a0e020a99e931215cf83ceb4de987c781488065aaae0d",
       "gasPrice": {
         "denom": "ustrd",
-        "amount": "0.005"
+        "amount": "0.025"
```
paulbalaji, line 4032: where is this coming from? ditto the osmosis change

paulbalaji: i still see 0.005 for stride on the cosmos registry here https://github.com/cosmos/chain-registry/blob/master/stride/chain.json

and 0.1 for osmosis here https://github.com/cosmos/chain-registry/blob/master/osmosis/chain.json

xeno097: just ran `yarn agent-configs`  locally as the ci was complaining, and that is the result i got, i'll double check what happened there

PR #6992, opened 2025-09-04
feat(fees): enforce fee and warp route tokens to match

ltyu:
### Description
This PR enforces that each tokenFee.token matches the warp route token

### Related issues
- Fixes le/eng-2150-enforce-fee-and-warp-route-tokens-to-match

### Backward compatibility
Yes

### Testing
Manual/Unit Tests

## Summary by CodeRabbit

- Bug Fixes
  - Enforced token consistency: fee token addresses must match the main warp route token (including nested routing fee entries). Validation now yields clear errors for mismatches; native fee types are normalized to the zero address.

- Tests
  - Added tests covering linear and routing fee scenarios to ensure validation fails with descriptive errors when fee tokens differ.

- Chores
  - Added a changeset marking a minor SDK version bump.

### DIFF
--- a/typescript/sdk/src/token/types.test.ts
+++ b/typescript/sdk/src/token/types.test.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { ethers } from 'ethers';
+import { constants, ethers } from 'ethers';
 
 import { assert } from '@hyperlane-xyz/utils';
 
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/types.ts
```diff
@@ -309,10 +309,39 @@ export type HypTokenRouterConfigMailboxOptionalBase = z.infer<
...
-);
+export const HypTokenRouterConfigMailboxOptionalSchema = z
+  .preprocess(
+    preprocessWarpRouteDeployConfig,
+    HypTokenRouterConfigMailboxOptionalBaseSchema,
+  ) // Check that each tokenFee.token is the same as config.token
+  .transform((config, ctx) => {
+    if (!isCollateralTokenConfig(config) || !config.tokenFee) return config;
+
+    if (config.tokenFee.token !== config.token) {
```
xeno097, line 320: What happens if the deployment is for a native token? shouldn't we check in that case that the fee token is set to the 0 addy?

ltyu: good question.
this is the input schema so it is expected that `token` will be undefined _for both_  `config.token` and `config.tokenFee.token`. in other words, the existing native config does not have address 0, so it is not expected that config.tokenFee.token to be address 0. 

however, after parsing, we will populate fee.token to be address 0. ideally, both `config.token` and `config.tokenFee.token` are set to address 0 for consistency, but i believe it is fine for now.

PR #6998, opened 2025-09-05
fix: add decimals verification

christopherbrumm:
### Description

fix: add decimals verification

### DIFF
--- a/rust/sealevel/libraries/hyperlane-sealevel-token/src/processor.rs
+++ b/rust/sealevel/libraries/hyperlane-sealevel-token/src/processor.rs
@@ -144,6 +144,12 @@ where
     /// 3.   `[signer]` The payer and access control owner.
     /// 4..N `[??..??]` Plugin-specific accounts.
     pub fn initialize(program_id: &Pubkey, accounts: &[AccountInfo], init: Init) -> ProgramResult {
+        // Validate that decimals are not greater than 9
+        if init.decimals > 9 {
+            msg!("Error: decimals ({}) cannot be greater than 9. Use remoteDecimals for higher precision.", init.decimals);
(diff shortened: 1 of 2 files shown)


rust/sealevel/libraries/hyperlane-sealevel-token/src/processor.rs
```diff
@@ -144,6 +144,12 @@ where
     /// 3.   `[signer]` The payer and access control owner.
     /// 4..N `[??..??]` Plugin-specific accounts.
     pub fn initialize(program_id: &Pubkey, accounts: &[AccountInfo], init: Init) -> ProgramResult {
+        // Validate that decimals are not greater than 9
+        if init.decimals > 9 {
```
yjamin, line 148: This is a contract change right? I don't know about the pipeline on Sealevel, but will this contract code be used for all token deployments?

yjamin: If this will not be used directly, we might want to add an extra check in the CLI

yjamin (approved):

rust/sealevel/libraries/hyperlane-sealevel-token/src/processor.rs
paulbalaji: would prefer making the changes in the sealevel CLI rather than contract code, since technically we'd be adding unaudited code (even if it's a simple exception)

@ameten thoughts?

christopherbrumm: Added it to the CLI instead. However, I still think this should be part of the contracts, but probably in another follow up that could be part of an audit.

PR #7004, opened 2025-09-08
fix: don't use radix-engine-toolkit for remote transfer

yjamin:
### Description
We don't allow `WebAssembly.instantiate` in the `warp-ui/nexus` and `radix-engine-toolkit` uses this, as a result we were unable to sign transactions in production in `nexus`.

This PR removes the usage of `radix-engine-toolkit` for the remote transfer and instead builds the manifest directly as a string.


Xaroz (approved):
looks good to me, but would @xeno097 to take a look if possible

xeno097 (approved):

PR #7021, opened 2025-09-10
feat: validate warp owners on deploy 

ltyu:
### Description
Update `warp deploy` to validate owners for each chain before deploying. **Note that this does not verify hook and ism owners for now**

### Drive-by changes
- Refactor `WarpModule.getOwnerStatus()` and split out owner checking logic into a separate function (`validateOwnerAddress()`
- Call this new `validateOwnerAddress` in the CLI with prompt confirmation

**With inactive address:**
Screenshot 2025-09-10 at 2 38 49 PM

**With address 0:**
Screenshot 2025-09-10 at 2 39 27 PM

### Related issues
- Fixes https://linear.app/hyperlane-xyz/issue/ENG-2194/validate-top-level-owner

### Backward compatibility
Yes - Existing flow can be ignored throw CLI prompt confirmation

### Testing
Manual/Unit Tests

## Release Notes

* **New Features**
  * Warp deploy now validates route owners before deployment to ensure they are active.
  * Users are prompted to confirm or abort deployment if invalid owners are detected.

* **Tests**
  * Added comprehensive integration and unit tests for owner validation scenarios.


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -255,36 +255,61 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
...
+      return {
+        [address]: OwnerStatus.Skipped,
+      };
+    }
+
+    const provider = this.multiProvider.getProvider(chain);
+    const owner = await Ownable__factory.connect(address, provider).owner();
+
+    // Use the extracted function to validate the owner
+    ownerStatus = await this.validateOwnerAddress(chain, owner);
```
xeno097, line 311: Should we have the `getOwnerStatus` method as a separate function so that it can be called from other places without needing to create an instance of the `EvmERC20WarpRouteReader?

typescript/sdk/src/deploy/validation.ts
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Validates activity status of a single owner address on specific chain.
+ * Uses EvmERC20WarpRouteReader.validateOwnerAddress() for direct owner validation.
+ */
+export async function validateOwnerActivity(
+  chain: ChainName,
+  address: Address,
+  multiProvider: MultiProvider,
+): Promise<OwnerValidationResult> {
+  const reader = new EvmERC20WarpRouteReader(multiProvider, chain);
```
xeno097, line 132: if we extract the method we can avoid doing this here and just call the function

typescript/sdk/src/deploy/validation.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (primaryOwner) {
+      owners.push(primaryOwner);
+    }
+
+    // 2. Proxy admin owner (if proxy deployment configured)
+    if (config.proxyAdmin?.owner) {
+      owners.push(config.proxyAdmin.owner);
+    }
+
+    return [...new Set(owners)];
```
xeno097, line 119: why not use a set from the start and then convert it to an array?

```typescript
const owners = new Set<Address>();
```

typescript/sdk/src/deploy/validation.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  if (invalidOwners.length > 0) {
+    const errorMessage = `Owner validation failed: Found ${invalidOwners.length} invalid owner(s). Invalid owners:\n${invalidOwners.map(
+      (o) => `\n- Chain: ${o.chain}, Owner: ${o.address}, Status: ${o.status}`,
+    )}`;
+
+    throw new OwnerValidationError(invalidOwners, errorMessage);
+  }
+
+  logger.info('Owner validation completed successfully');
```
xeno097, line 95: why not return an object thatindicates if the validation was succesful or not and let the caller decide if throwing an error is the best option?

typescript/cli/src/deploy/warp.ts
```diff
@@ -137,6 +140,7 @@ export async function runWarpRouteDeploy({
   };
 
   await runDeployPlanStep(deploymentParams);
+  await validateAndConfirmOwners(multiProvider, warpDeployConfig);
```
xeno097, line 143: i think this does not cover cosmos chain right? should we also add support for that later?

PR #7022, opened 2025-09-10
feat(infra): hot wallet funder script

xeno097:
### Description

This PR adds a script to fund wallets using the deployer keys from GCP. it only works for native tokens but can be expanded to support other token types

### Drive-by changes

- Defined a `IMultiProtocolSigner` interface to implement abstracted signer that can use the transactions generated from the token adapters.
- Implements the `populateTransfer` method on the `CosmNativeTokenAdapter` 

### Related issues

-

### Backward compatibility

- YES

### Testing

- Manual
- Cosmos (Celestia): https://www.mintscan.io/celestia/tx/DCE22AAA3B5CE7479CFB6E36B43E4053C2DA66F2882D5DF443E80E1D282DE490?height=7425689
- Starknet: https://voyager.online/tx/0x056e234d7fe6bdd1fdd7c57d0541aea3c1e011df5da3c1e56194a3659bdcbfcb
- EVM (op): https://optimistic.etherscan.io/tx/0x5f3b1661811e8566a41bf3a65ca7aad5c4e1accbd2f720858c871f8eacbf6265
- SVM (solana):  https://solscan.io/tx/3Zeb2uvqL8CRJQ7XZKVFKqh3S4syzpVkbTunWNipXw9cVhWHURmep7o4VJMmCJGbtWR8RPjaDMUsWm8Rk2677juk


typescript/infra/src/agents/agent.ts
```diff
@@ -23,6 +23,9 @@ function identifier(
       if (!chainName) throw Error('Expected chainName for validator key');
       if (index === undefined) throw Error('Expected index for validator key');
       return `${prefix}${chainName}-${role}-${index}`;
+    // In the case of deployer keys not all the keys have the chainName included in their identifier
+    case Role.Deployer:
+      return `${context}-${environment}-${chainName ? chainName + '-' : ''}${isKey ? 'key-' : ''}${role}`;
```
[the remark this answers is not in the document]
xeno097: That sounds reasonable, adopting a standard naming convention for the keys would help in not having to handle per protocol/chain cases

typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts
```diff
@@ -0,0 +1,269 @@
...
+  const privateKeyAgent = getDeployerKey(agentConfig, chainName);
+
+  await privateKeyAgent.fetch();
+
+  let accountInfo: { privateKey: string; address: Address };
+  if (protocol === ProtocolType.Sealevel) {
+    accountInfo = {
+      address: privateKeyAgent.address,
+      // Assumes the private key is base64 encoded on secrets manager
+      privateKey: String(Buffer.from(privateKeyAgent.privateKey, 'base64')),
```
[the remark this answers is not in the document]
xeno097: I'll have to test again the implementation in `AgentGCPKey.addressForProtocol()` to see if it works but i had to do that wierd conversion because it wasn't working as expected if i recall correctly, also checking the key on  GCP it gave me base64 vibes and in the end it worked after changing the parsing logic to the current one

typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts
```diff
@@ -0,0 +1,269 @@
...
+  // Get signer
+  logger.info({ chainName, protocol }, 'Retrieving signer info');
+
+  const agentConfig = getAgentConfig(Contexts.Hyperlane, config.environment);
+  const privateKeyAgent = getDeployerKey(agentConfig, chainName);
+
+  await privateKeyAgent.fetch();
+
+  let accountInfo: { privateKey: string; address: Address };
+  if (protocol === ProtocolType.Sealevel) {
```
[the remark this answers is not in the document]
xeno097: that sounds like a really good idea, thanks!

typescript/infra/src/agents/key-utils.ts
```diff
@@ -306,9 +306,18 @@ export function getKathyKeyForChain(
 
 // Returns the deployer key. This is always a GCP key, not chain specific,
```
Mo-Hussain, line 307: I guess we need to change this comment

typescript/sdk/src/signers/cosmos/stargate.ts
```diff
@@ -0,0 +1,80 @@
+import { DirectSecp256k1Wallet } from '@cosmjs/proto-signing';
```
Mo-Hussain, line 1: Why have we named this file `stargate`?

typescript/sdk/src/signers/evm/ethersv5.ts
```diff
@@ -0,0 +1,51 @@
...
+import { EthersV5Transaction } from '../../providers/ProviderType.js';
+import { ChainName } from '../../types.js';
+import { IMultiProtocolSigner } from '../types.js';
+
+export class EvmMultiProtocolSignerAdapter
+  implements IMultiProtocolSigner<ProtocolType.Ethereum>
+{
+  private readonly multiProvider: MultiProvider;
+
+  constructor(
```
Mo-Hussain, line 18: Would is make sense to setup the signer as part of a the base protocol adapter and reuse here? Like this https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6125/files#diff-ba786a5cc05f1477b509a9f93007fda439e155c853e1267f3aed7fa0278f440a

That way other adapters can reuse the signer

typescript/sdk/src/signers/cosmos/stargate.ts
xeno097: yaeh i should call it cosmjs like the others have the library name instead, i'll change it

typescript/sdk/src/signers/evm/ethersv5.ts
xeno097: but that would work only for evm chains right? because rn we don't have a multi protocol signer (at least in the sdk), unless i'm misunderstading

Mo-Hussain: We have the `MultiProtocolApp` class and we define a `BaseAppAdapter` per protocol. It may not be clean to reuse this abstraction in this case but it would be nice if we could. We could also add signer support to the `MultiProtocolProvider` like I have done here https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6125/files#diff-eaeb30ae5ebadd7c49b753cbd99c41fb72e67eb939c444fbcdc1b635e1a962de.

I don't have a clear view on whether we can can cleanly reuse abstractions here but my hunch is that we can benefit from some reuse

antigremlin: TBH, I believe we should gradually move away from catch-all classes that provide multi-protocol APIs to the application logic. Use-case specific interfaces like TokenAdapter and MultiProtocolSigner (I'd give it a shorter name, or even a SignerAdapter to align with TokenAdapter) are a design choice that reduces coupling. Imagine more smaller interfaces like these with protocol-specific implementations but not tied together by protocol providers that know everything.

More to the point, I'd go with the smaller surface interfaces, and so with the way signers are introduced in this PR.

typescript/sdk/src/signers/signers.ts
```diff
@@ -0,0 +1,68 @@
...
+      privateKey: Uint8Array;
+    }
+  | {
+      protocol: ProtocolType.Starknet;
+      privateKey: HexString;
+      address: HexString;
+    };
+
+export async function getSignerForChain<
+  TProtocol extends ProtocolType = ProtocolType,
```
antigremlin, line 31: Do we need this default `= ProtocolType`?

typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts
```diff
@@ -0,0 +1,283 @@
...
+      privateKey: privateKeyAgent.privateKeyForProtocol(protocol),
+    };
+  } else {
+    accountInfo = {
+      protocol,
+      privateKey: privateKeyAgent.privateKeyForProtocol(protocol),
+    } as MultiProtocolSignerSignerAccountInfo;
+  }
+
+  const signer = await getSignerForChain<typeof protocol>(
```
antigremlin, line 186: Do we need this `<typeof protocol>`?

typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts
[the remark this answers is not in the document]
antigremlin: I was wrong about the format, it's base64. Sorry.

typescript/infra/src/agents/gcp.ts
```diff
@@ -149,6 +177,26 @@ export class AgentGCPKey extends CloudAgentKey {
     }
   }
 
+  override privateKeyForProtocol(
+    protocol: Exclude<ProtocolType, ProtocolType.Sealevel>,
+  ): HexString;
+  override privateKeyForProtocol(protocol: ProtocolType.Sealevel): Uint8Array;
```
antigremlin, line 183: Nice!

antigremlin (approved):

typescript/infra/src/agents/gcp.ts
```diff
@@ -149,6 +177,26 @@ export class AgentGCPKey extends CloudAgentKey {
...
+  override privateKeyForProtocol(
+    protocol: ProtocolType,
+  ): HexString | Uint8Array {
+    this.requireFetched();
+
+    if (protocol === ProtocolType.Sealevel) {
+      return Uint8Array.from(
+        JSON.parse(String(Buffer.from(this.privateKey, 'base64'))),
```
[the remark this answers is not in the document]
xeno097: So the reason here is that the private key is stored on gcp as the base64 string of the string representation of the private key in buffer format so we need to:
- convert the base64 string to a buffer
- convert the buffer to a string to have the stringified version of the private key array
- convert the string to an array using json.parse
- finally get the buffer from the parsed array

Please let me know if it makes sense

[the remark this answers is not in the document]
antigremlin: It did make sense after another chat 🤣 
Please add a comment explaining the expected format

PR #7039, opened 2025-09-15
feat: add radix hook reader

troykessler:
### Description

This PR adds the Radix Hook Reader class needed for core and warp cli commands

### Testing

Manual Tests on `feat/radix-cli` with same codebase

## Summary by CodeRabbit

* **New Features**
  * Added a Radix Hook Reader that detects a hook's type (Interchain Gas Paymaster or Merkle Tree) for an address and returns its configuration.
  * Exposed the Radix Hook Reader in the public SDK for easier integration.

* **Chores**
  * Added a changeset to publish a minor version update for the SDK.

### DIFF
--- a/.changeset/hungry-rabbits-mix.md
+++ b/.changeset/hungry-rabbits-mix.md
@@ -1,5 +1,6 @@
 ---
+"@hyperlane-xyz/radix-sdk": minor
 "@hyperlane-xyz/sdk": minor
 ---
 
-add radix hook reader
+export radix hook reader


typescript/sdk/src/hook/RadixHookReader.ts
```diff
@@ -0,0 +1,107 @@
...
+export class RadixHookReader {
+  protected readonly logger = rootLogger.child({
+    module: 'RadixHookReader',
+  });
+
+  constructor(
+    protected readonly metadataManager: ChainMetadataManager,
+    protected readonly sdk: RadixSDK | RadixSigningSDK,
+  ) {}
+
```
xeno097, line 22: nit: Can you implement a function that returns the hook type directly? 

```typecript
async deriveHookType(address: Address): Promise<HookType>{
   ....
}
```

so you can then reuse it in the `deriveHookConfig` instead of having to call multiple functions here?

typescript/sdk/src/hook/RadixHookReader.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.metadataManager.getChainMetadata(remoteDomain);
+
+      const gasConfig = igp.destination_gas_configs[remoteDomain];
+
+      overhead[name] = parseInt(gasConfig.gas_overhead);
+      oracleConfig[name] = {
+        gasPrice: gasConfig.gas_oracle?.gas_price ?? '',
+        tokenExchangeRate: gasConfig.gas_oracle?.token_exchange_rate ?? '',
+        tokenDecimals: nativeToken?.decimals,
+      };
```
xeno097, line 59: if a config is not set for a given domain i'd rather not include it in the final output instead of leaving its fields as an empty string.

xeno097 (commented):
Is there any way we can add tests fro this? does radix have something similar to anvil or hardhat for local development?

xeno097:
> > Is there any way we can add tests fro this? does radix have something similar to anvil or hardhat for local development?
> 
> radix has no way to test against a local chain, the e2e tests that will cover all these classes will execute transactions and reads against the radix testnet
> 
> more context regarding e2e tests: [#7040 (comment)](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/7040#issuecomment-3296578415)

np then! it seems that the linter is complaining, after that is fixed i think you can merge

typescript/sdk/src/hook/RadixHookReader.ts
```diff
@@ -0,0 +1,116 @@
...
+    module: 'RadixHookReader',
+  });
+
+  constructor(
+    protected readonly metadataManager: ChainMetadataManager,
+    protected readonly sdk: RadixSDK,
+  ) {}
+
+  async deriveHookConfig(address: Address): Promise<DerivedHookConfig> {
+    const hookType = await this.deriveHookType(address);
```
antigremlin, line 24: As `deriveHookType` already calls `deriveXxxConfig`, this RPC organisation doesn't seem optimal at all. Shouldn't we strive to minimize RPC calls?

Two solutions seem feasible:

1. Call `deriveXxxConfig` once, in sequence as we do here, but preserve the data collected from the ledger. Then pass this data to the remaining hook-type-specific logic.
2. Extract the type-checking logic in `RadixCoreQuery` into a function and call it here. This is more work but feels cleaner to me -- we check the type as the first step, and avoid exceptions.

If we choose option 1, we can also change `RadixCoreQuery` to reduce the number of exceptions thrown (a null/unknown hook returned is presumably better than an exception).

Here's what I mean by option 2. There's a similar function for ISMs already but I believe we can make it more generic and export an enum of Hook/ISM blueprint names to compare against.

```typescript
public async getEntityBlueprintName(address: string): Promise<string> {
  const details =
    await this.gateway.state.getEntityDetailsVaultAggregated(address);
  if (details.details?.type === 'Component') {
    return details.details.blueprint_name;
  }
  return '';
}
```

antigremlin (commented):
Please see my comments

PR #7040, opened 2025-09-15
chore: add radix ism reader

troykessler:
### Description

This PR adds the Radix ISM Reader class needed for core and warp cli commands

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Tests on `feat/radix-cli` with same codebase

### DIFF
--- a/typescript/sdk/src/ism/RadixIsmReader.ts
+++ b/typescript/sdk/src/ism/RadixIsmReader.ts
@@ -1,0 +1,106 @@
+import { RadixSDK, RadixSigningSDK } from '@hyperlane-xyz/radix-sdk';
+import { Address, WithAddress, assert, rootLogger } from '@hyperlane-xyz/utils';
+
+import { ChainMetadataManager } from '../metadata/ChainMetadataManager.js';
+
+import {
+  DerivedIsmConfig,
+  DomainRoutingIsmConfig,
+  IsmType,
+  MultisigIsmConfig,
+} from './types.js';
+
+export class RadixIsmReader {
+  protected readonly logger = rootLogger.child({
+    module: 'RadixIsmReader',
+  });
+
+  constructor(
+    protected readonly metadataManager: ChainMetadataManager,
+    protected readonly sdk: RadixSDK | RadixSigningSDK,
+  ) {}
+
+  async deriveIsmConfig(address: Address): Promise<DerivedIsmConfig> {
+    try {
+      const ismType = await this.sdk.query.core.getIsmType({ ism: address });
+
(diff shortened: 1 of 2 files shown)


typescript/sdk/src/ism/RadixIsmReader.ts
```diff
@@ -0,0 +1,107 @@
...
+} from './types.js';
+
+export class RadixIsmReader {
+  protected readonly logger = rootLogger.child({
+    module: 'RadixIsmReader',
+  });
+
+  constructor(
+    protected readonly metadataManager: ChainMetadataManager,
+    protected readonly sdk: RadixSDK | RadixSigningSDK,
```
xeno097, line 21: nit: could we maybe define an interface that describes which methods are needed for the `RadixIsmReader` class to work instead of specifying the concrete types?

typescript/sdk/src/ism/types.ts
```diff
@@ -374,3 +374,10 @@ export const IsmConfigSchema = z.union([
   OffchainLookupIsmConfigSchema,
   InterchainAccountRouterIsmSchema,
 ]);
+
+export enum RadixIsmTypes {
+  MERKLE_ROOT_MULTISIG = 'MerkleRootMultisigIsm',
+  MESSAGE_ID_MULTISIG = 'MessageIdMultisigIsm',
+  ROUTING_ISM = 'RoutingIsm',
+  NOOP_ISM = 'NoopIsm',
+}
```
xeno097, line 383: I see that the `radix-sdk` package defines a similar type to this one with the difference that that one is a union type could we maybe have this enum replace that type and be exported from the `radix-sdk` package instead? so we can avoid this "duplication"

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/radix-sdk/src/utils/types.ts#L29-L38

typescript/sdk/src/ism/RadixIsmReader.ts
troykessler: There is an IsmReader interface but it requires implemented EVM specific methods which don't exist in other VMs like Cosmos or Radix: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/chore/radix-ism-reader/typescript/sdk/src/ism/EvmIsmReader.ts#L54

There is one method which is shared by every VM, should I remove all the other methods from the IsmReader interface then to make it accessible for the other Readers?

antigremlin (approved):

antigremlin:
@coderabbitai full review

xeno097 (approved):
pending checking why tests are failing

PR #7043, opened 2025-09-15
WIP: cosmos warp apply

troykessler:
### Description

WIP - DO NOT MERGE

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Testing


typescript/cli/src/deploy/warp.ts
```diff
@@ -625,26 +642,61 @@ async function updateExistingWarpRoute(
...
-            },
-          },
-          ccipContractCache,
-          contractVerifier,
-        );
-        transactions.push(
-          ...(await evmERC20WarpModule.update(configWithMailbox)),
-        );
+        switch (multiProvider.getProtocol(chain)) {
+          case ProtocolType.Ethereum: {
```
yjamin, line 646: I kind of don't like this structure. I feel like when someone will add a new groupedTx it will need a lot of manual work to make sure it is supported on every place. Like here, the compiler won't complain when someone adds a new type, but the logic will be missing.

Usually, I would expect there to be some kind of type, like `supportedWarpProtocols` and `groupedTransactions` would derive its type based on its values. This would ensure when someone adds a new warp apply type that the project won't compile until all of the logic necessary is implemented. Maybe too big of a change right now, but definitely something that we should do.

typescript/cli/src/deploy/warp.ts
```diff
@@ -847,28 +899,20 @@ function transformIsmConfigForDisplay(ismConfig: IsmDisplayConfig): any[] {
...
-  );
-
   const { extendedChains } = getWarpRouteExtensionDetails(
     params.warpCoreConfig,
     params.warpDeployConfig,
   );
 
-  await promiseObjAll(
-    objMap(chainTransactions, async (chainId, transactions) => {
+  for (const [protocol, chainTransactions] of Object.entries(
```
yjamin, line 909: Is this change intentional? It now will await the Futures one by one, instead of creating all of them at once and waiting for them to finish in parallel.

typescript/cli/src/tests/commands/warp.ts
```diff
@@ -148,4 +148,23 @@ export class HyperlaneE2EWarpTestCommands {
       warpRouteId,
     });
   }
+
+  /**
+   * Updates a Hyperlane warp deployment on the specified chain using the provided config.
+   */
+  public apply(
```
yjamin, line 155: I don't see this being used anywhere

typescript/sdk/src/ism/CosmosNativeIsmReader.ts
```diff
@@ -54,6 +64,22 @@ export class CosmosNativeIsmReader {
...
   }
 
+  async deriveIsmConfig(config: IsmConfig): Promise<DerivedIsmConfig> {
+    if (typeof config === 'string')
+      return this.deriveIsmConfigFromAddress(config);
+
+    // Extend the inner isms
+    switch (config.type) {
+      case IsmType.ROUTING:
+        config.domains = await promiseObjAll(
```
yjamin, line 74: nit: maybe limit the number of requests that we do simultaneously, I can imagine doing 150+ requests at once might result in errors.

typescript/sdk/src/providers/transactions/submitter/cosmosnative/CosmosNativeJsonRpcTxSubmitter.ts
```diff
@@ -0,0 +1,41 @@
+import { DeliverTxResponse } from '@cosmjs/stargate';
```
yjamin, line 1: nit: maybe move this into a different folder, like `submitter/...`

typescript/sdk/src/providers/transactions/submitter/submitterBuilderGetter.ts
```diff
@@ -40,72 +48,98 @@ export async function getSubmitterBuilder<TProtocol extends ProtocolType>({
...
-  [TxSubmitterType.TIMELOCK_CONTROLLER]: (
-    multiProvider,
-    metadata,
-    coreAddressesByChain,
-  ) => {
-    assert(
-      metadata.type === TxSubmitterType.TIMELOCK_CONTROLLER,
-      `Invalid metadata type: ${metadata.type}, expected ${TxSubmitterType.TIMELOCK_CONTROLLER}`,
-    );
+const defaultSubmitterFactories: ProtocolMap<Record<string, SubmitterFactory>> =
```
yjamin, line 56: I don't feel strong about this, but it is maybe good to add an explicit error handling when someone tries to access a submitter that is not supported. Like for ProtocolType Radix or something

typescript/sdk/src/token/configUtils.ts
```diff
@@ -243,28 +245,49 @@ export async function expandWarpDeployConfig(params: {
...
       // if we have an address we leave it like that to avoid deriving
-      if (
-        isEVMChain &&
-        chainConfig.hook &&
-        typeof chainConfig.hook !== 'string'
-      ) {
-        const reader = new EvmHookReader(multiProvider, chain);
-
-        chainConfig.hook = await reader.deriveHookConfig(chainConfig.hook);
+      if (chainConfig.hook && typeof chainConfig.hook !== 'string') {
```
yjamin, line 248: I believe we should test here for compatible protocols as well. Right now it doesn't test at all. At least we should add a default case to the switch statement to log if we encountered a non-compatible chain.

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -76,6 +77,28 @@ export class CosmosNativeDeployer {
         }
       }
```
yjamin, line 79: On line 51 of this file, I believe it should state
```ts
```

typescript/sdk/src/types.ts
```diff
@@ -12,7 +12,7 @@ export type ChainName = string;
 // A map of chain names to a value type
 export type ChainMap<Value> = Record<ChainName, Value>;
 // A map of protocol types to a value type
-export type ProtocolMap<Value> = Record<ProtocolType, Value>;
+export type ProtocolMap<Value> = Partial<Record<ProtocolType, Value>>;
```
yjamin, line 15: This has huge implications no? This makes everything optional, what was the intention behind this?

typescript/sdk/src/token/CosmosNativeWarpModule.ts
```diff
@@ -402,9 +402,9 @@ export class CosmosNativeWarpModule extends HyperlaneModule<
         addresses: {
           ...this.args.addresses,
           mailbox: expectedConfig.mailbox,
-          deployedIsm: (
-            actualConfig.interchainSecurityModule as DerivedIsmConfig
-          ).address,
+          deployedIsm:
+            (actualConfig.interchainSecurityModule as DerivedIsmConfig)
+              ?.address ?? '',
```
xeno097, line 407: Isn't this a bit dangerous? can you explain why it is set to an empty string as fallback? wouldn't it be better to explicitly deploy if no ism is set on the current config?

xeno097 (commented):
left some comments but i will have another pass on monday

typescript/cli/src/deploy/warp.ts
troykessler: removed the concept of grouped transactions and simply work with a chain map of annotated txs now

typescript/cli/src/deploy/warp.ts
```diff
@@ -378,22 +378,16 @@ export async function runWarpRouteApply(
...
-        return config;
-      }
-
-      return {
-        ...config,
-        owner: await multiProvider.getSignerAddress(chain),
-      };
-    }),
+    objMap(params.warpDeployConfig, async (chain, config) => ({
+      ...config,
```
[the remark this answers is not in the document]
troykessler: No but we need to address this issue in general because adding chains in the config like solana breaks the cli currently because of other issues too. Will address this in this separate PR (still in progress): https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/6956

typescript/sdk/src/token/CosmosNativeWarpModule.ts
troykessler: Had to do it because in the edge case where there is currently no custom ism for a warp route deployed and now there should be created one in warp apply it would fail otherwise. It is handled here: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/feat/cosmos-warp-apply/typescript/sdk/src/ism/CosmosNativeIsmModule.ts#L94

typescript/cli/src/deploy/warp.ts
```diff
@@ -959,10 +996,13 @@ export async function getSubmitterByStrategy<T extends ProtocolType>({
...
       multiProvider,
+      multiProtocolSigner,
       coreAddressesByChain: await registry.getAddresses(),
       additionalSubmitterFactories: {
-        file: (_multiProvider: MultiProvider, metadata: any) => {
-          return new EV5FileSubmitter(metadata);
+        [ProtocolType.Ethereum]: {
+          file: (_multiProvider: MultiProvider, metadata: any) => {
+            return new EV5FileSubmitter(metadata);
+          },
```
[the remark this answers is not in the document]
troykessler: This PR adds warp apply for cosmos, I won't change any EVM behaviour here, can do a separate if needed

yjamin (approved):

PR #7044, opened 2025-09-15
chore: update agent config

kamiyaa:
### Description

- update zircuit reorgPeriod to 10mins


paulbalaji (commented):
need to update the registryrc too

rust/main/config/mainnet_config.json
paulbalaji, line 1: this is the diff i'd expect
image


ensure your local registry is pulled to latest before running `yarn agent-configs`

paulbalaji (approved):

PR #7052, opened 2025-09-17
feat: support rebalancer key in key funder cron job

Mo-Hussain:
### Description

- support rebalancer key in key funder cron job

### Testing

Manual

## Summary by CodeRabbit

* **New Features**
  * Added per-chain rebalancer balance configuration and mainnet3 target dataset; funding flow and Helm values now pass these per-chain rebalancer balances; CLI option added to set them.
  * Rebalancer role included in Hyperlane funding contexts.

* **Tests**
  * Added tests ensuring rebalancer balance entries exist and parse as numbers for relevant chains.

* **Chores**
  * Adjusted balance-flag handling to accept at least one per-chain balance flag.

### DIFF
--- a/typescript/infra/config/environments/mainnet3/funding.ts
+++ b/typescript/infra/config/environments/mainnet3/funding.ts
@@ -28,7 +28,7 @@ export const keyFunderConfig: KeyFunderConfig<
 > = {
   docker: {
     repo: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
-    tag: '093e0be-20250904-212218',
+    tag: 'b120e26-20250922-122003',
   },
   // We're currently using the same deployer/key funder key as mainnet2.
   // To minimize nonce clobbering we offset the key funder cron


typescript/infra/config/environments/mainnet3/balances/desiredRebalancerBalances.json
```diff
@@ -0,0 +1,8 @@
+{
+  "arbitrum": 0.6,
+  "base": 0.2,
+  "ethereum": 0.3,
+  "optimism": 0.3,
+  "polygon": 130,
+  "unichain": 0.07
+}
```
xeno097, line 8: should we maybe have a test that ensure that all the chains in the mainnet cctp warp route have a desired balance setting here?  I think this can be something that can be easily forgotten to add if we cover a new chain that is in there but not here

paulbalaji (approved):
lgtm % sebastiano's comments

PR #7062, opened 2025-09-19
fix: cosmos native warp deploy

troykessler:
### Description

This PR fixes an issue where the Token ISM and Owner is not properly set after initially deploying a token to cosmos native

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Tests

## Summary by CodeRabbit

- New Features
  - Tokens deployed on Cosmos-native chains now automatically apply the configured interchain security module, reducing manual configuration.
  - Post-deployment ownership is automatically aligned with the configured owner when different, ensuring proper control without extra steps.
  - Deployment logs are clearer, indicating when the security module and ownership are set, improving visibility during rollout.

### DIFF
--- a/typescript/sdk/src/token/cosmosnativeDeploy.ts
+++ b/typescript/sdk/src/token/cosmosnativeDeploy.ts
@@ -5,6 +5,7 @@ import {
   Address,
   ProtocolType,
   assert,
+  eqAddress,
   objFilter,
   objMap,
   rootLogger,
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -76,6 +77,28 @@ export class CosmosNativeDeployer {
         }
       }
 
+      if (config.interchainSecurityModule) {
+        this.logger.info(`Set ISM for token`);
+
+        await this.signersMap[chain].setToken({
+          token_id: result[chain],
+          new_owner: '',
```
xeno097, line 85: why an empty string?

xeno097: not for this PR but could it be maybe worth changing the interface so that fields that are not required/we don't want to update do not need to be provided by the user?

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -76,6 +77,28 @@ export class CosmosNativeDeployer {
         }
       }
```
yjamin, line 78: On line 51 of this file, I believe it should state 
```ts
```

typescript/sdk/src/token/cosmosnativeDeploy.ts
```diff
@@ -76,6 +77,28 @@ export class CosmosNativeDeployer {
...
 
+      if (config.interchainSecurityModule) {
```
yjamin, line 80: I wonder what happens when one need to enroll remote chains? Is that pat of the `WarpRouteDeployConfigMailboxRequired`? Or is this performed on a different step?

PR #7070, opened 2025-09-22
chore: export radix ism module

troykessler:
### Description

This PR implements and exports the RadixIsmModule in the sdk package. Furthermore, it fixes the normalizeAddress method for radix addresses

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Tests

### DIFF
--- a/typescript/sdk/src/ism/RadixIsmModule.ts
+++ b/typescript/sdk/src/ism/RadixIsmModule.ts
@@ -263,6 +263,11 @@ export class RadixIsmModule extends HyperlaneModule<
     expected: DomainRoutingIsmConfig;
     logger: Logger;
   }): Promise<AnnotatedRadixTransaction[]> {
+    assert(
+      eqAddress(this.signer.getAddress(), actual.owner),
+      `can not update routing ism since the owner ${actual.owner} is not the signer: ${this.signer.getAddress()}`,
+    );
+
     const updateTxs: AnnotatedRadixTransaction[] = [];
 
     const knownChains = new Set(this.metadataManager.getKnownChainNames());


typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,241 @@
...
+    return this.reader.deriveIsmConfig(this.args.addresses.deployedIsm);
+  }
+
+  // whoever calls update() needs to ensure that targetConfig has a valid owner
+  public async update(
+    expectedConfig: IsmConfig,
+  ): Promise<AnnotatedRadixTransaction[]> {
+    expectedConfig = IsmConfigSchema.parse(expectedConfig);
+
+    // Do not support updating to a custom ISM address
```
yjamin, line 80: Might be a stupid question, but why don't we support this? Is this normally not supported, or when would this even be a usecase?

typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,241 @@
...
+    // if the ISM is a static ISM we can not update it, instead
+    // it needs to be recreated with the expected config
+    if (STATIC_ISM_TYPES.includes(expectedConfig.type)) {
+      this.args.addresses.deployedIsm = await this.deploy({
+        config: expectedConfig,
+      });
+
+      return [];
+    }
+
```
yjamin, line 114: What happens when the update is a `DomainRoutingIsm` and only the domain change? I think we need to update the domains here in that case

yjamin: Also following on that, only the owner might be different. In that case we would need to transfer the onwership of that `DomainRoutingIsm`

typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,241 @@
...
+      },
+      signer,
+    );
+
+    module.args.addresses.deployedIsm = await module.deploy({ config });
+    return module;
+  }
+
+  protected async deploy({ config }: { config: IsmConfig }): Promise<Address> {
+    if (typeof config === 'string') {
```
yjamin, line 152: When would this be the case?

typescript/sdk/src/ism/RadixIsmModule.ts
troykessler: Implemented now

typescript/sdk/src/ism/RadixIsmModule.ts
yjamin: In that case just leave it, it is probably not supported anyways

typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,241 @@
...
+          ...addresses,
+          deployedIsm: '',
+        },
+        chain,
+        config,
+      },
+      signer,
+    );
+
+    module.args.addresses.deployedIsm = await module.deploy({ config });
```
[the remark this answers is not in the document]
yjamin: Okay, weird.

typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,341 @@
...
+            domain,
+          },
+        }),
+      });
+    }
+
+    const knownUnenrolls = intersection(
+      knownChains,
+      new Set(domainsToUnenroll),
+    );
```
yjamin, line 304: nit: I wonder what happens when we have domains to unenroll which are not part of the knownChains. I think it should still unenroll the unknown domains. WDYT?

yjamin (approved):

typescript/sdk/src/ism/RadixIsmModule.ts
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    // Update ownership
+    if (!eqAddress(actual.owner, expected.owner)) {
+      updateTxs.push({
+        annotation: `Transferring ownership of ISM from ${
+          actual.owner
+        } to ${expected.owner}`,
+        networkId: this.signer.getNetworkId(),
+        manifest: await this.signer.populate.core.setRoutingIsmOwner({
```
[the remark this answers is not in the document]
troykessler: If you check the implementation of the `setRoutingIsmOwner` method it does exactly that. Created a method with a clear name so that users of the radix-sdk package can use it out of the box and don't need to figure out how to transfer ownership tokens

typescript/sdk/src/ism/RadixIsmModule.ts
yjamin: Isn`t `domainsToUnenroll` straight up the domainId?

PR #7086, opened 2025-09-23
test: lint for functions that are virtual and override

yorhodes:
### Description

Adds a lint rule for erroring when functions are `virtual override` which will create multiple layers of overrides in the inheritance tree

### Testing

Manual

`$ yarn solhint contracts/token/**/*.sol`

## Summary by CodeRabbit

- Refactor
  - Reduced override-ability across many Solidity contracts to harden behavior without changing runtime logic in most cases.
- Changes
  - Service manager deregistration now updates the directory and emits an event.
  - Token bridge behavior adjusted to delegate handling and modify transfer path.
  - Simplified fee calculation for ERC-4626 collateral transfers.
- Chores
  - Added Hyperlane Solhint plugin and enforced new lint rules.
  - Introduced a new workspace package for the lint plugin and updated configurations.
- Tests
  - Updated tests to align with token type usage and refactors.


larryob (approved):
LGTM

PR #7110, opened 2025-09-26
refactor: align radix custom rpc header with evm implementation

yjamin:
### Description
This PR aligns the `customRpcHeader` for Radix with the EVM implementation. It allows users to pass RPC Headers by setting a query argument with `customRpcHeader`. Previously, you'd need to se `coreHeader` and `gatewayHeader` respectively to set your custom Header. This implementation is now aligned with the EVM implementation.

## Summary by CodeRabbit

* **New Features**
  * Support for custom RPC headers via a custom_rpc_header URL query parameter, with unit tests validating parsing and URL reconstruction.

* **Refactor**
  * Simplified connection configuration to require fewer parameters.
  * Per-URL HTTP client creation that applies parsed headers and preserves other query params.
  * Added a short (5s) pause after local remote transfers.

* **Chores**
  * Removed legacy header-parsing paths, cleaned unused imports, added a shared reqwest-utils helper, and updated dependency references.

### DIFF
--- a/rust/main/chains/hyperlane-radix/src/provider/radix.rs
+++ b/rust/main/chains/hyperlane-radix/src/provider/radix.rs
@@ -106,7 +106,7 @@ impl RadixProvider {
                 ..Default::default()
             });
             let provider =
-                RadixMetricGatewayProvider::new(provider, url, metrics.clone(), chain.clone());
+                RadixMetricGatewayProvider::new(provider, &url, metrics.clone(), chain.clone());
             gateway_provider.push(provider);
         }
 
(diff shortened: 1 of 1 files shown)


rust/main/chains/hyperlane-radix/src/provider/radix.rs
```diff
@@ -77,18 +79,50 @@ impl Deref for RadixProvider {
     }
 }
 
+/// Builds a new reqwest client with the given URL.
+fn build_new_reqwest_client(url: Url) -> ChainResult<Client> {
```
ameten, line 83: Could you please add unit tests for this function?

PR #7113, opened 2025-09-26
feat(infra): Manifest-Generator for IGP updates on Radix

mbreithecker:
### Description

Adds a TS helper file to generate the transaction manifest for updating IPG prices on Radix.

### Testing

The script was already used to update the Radix IGP.

### DIFF
--- a/typescript/infra/scripts/radix-helpers/generate-igp-manifest.ts
+++ b/typescript/infra/scripts/radix-helpers/generate-igp-manifest.ts
@@ -1,0 +1,147 @@
+import { Argv } from 'yargs';
+
+import rawGasPrices from '../../config/environments/mainnet3/gasPrices.json' with { type: 'json' };
+import rawTokenPrices from '../../config/environments/mainnet3/tokenPrices.json' with { type: 'json' };
+import { getArgs, withChains } from '../agent-utils.js';
+import { getEnvironmentConfig } from '../core-utils.js';
+
+export function withOriginChain<T>(args: Argv<T>) {
+  return args
+    .describe('originChain', 'The chain for which IGP prices are calculated')
+    .alias('o', 'origin-chain')
+    .default('originChain', '')
+    .demandOption('originChain');
+}
+
+const RADIX_ADDRESSES = {
+  igp: 'component_rdx1cznxpn5m3kutzr6jrhgnvv0x7uhcs0rf8fl2w59hkclm6m7axzlqgu',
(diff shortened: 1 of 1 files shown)


typescript/infra/scripts/radix-helpers/generate-igp-manifest.ts
```diff
@@ -0,0 +1,147 @@
...
+import { getArgs, withChains } from '../agent-utils.js';
+import { getEnvironmentConfig } from '../core-utils.js';
+
+export function withOriginChain<T>(args: Argv<T>) {
+  return args
+    .describe('originChain', 'The chain for which IGP prices are calculated')
+    .alias('o', 'origin-chain')
+    .default('originChain', '')
+    .demandOption('originChain');
+}
```
paulbalaji, line 14: why add this as an arg if it's just going to be radix?

typescript/infra/scripts/radix-helpers/generate-igp-manifest.ts
```diff
@@ -0,0 +1,147 @@
...
+      gas_price_decimals: gasPrices[chain].decimals,
+    };
+  });
+
+  if (!tokenPrices[originChain]) {
+    throw Error(`No token price found for ${originChain}`);
+  }
+
+  const tupleLines = [];
+  for (const entry of tokenPriceConfigs) {
```
paulbalaji, line 71: why not support the radix protocol type in igp.ts/gas-oracle.ts and then reuse the mapping we get with that? that way we don't have to configure this specially for radix? a la https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/scripts/sealevel-helpers/print-gas-oracles.ts

typescript/infra/scripts/radix-helpers/generate-igp-manifest.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (entry.domain_id == 1) {
+      combinedFactor = (1.5 / (268000 + 50000) / origin_token_price) * 1e10;
+    }
+
+    tupleLines.push(
+      `    Tuple(${entry.domain_id}u32, Tuple(Tuple(${Math.round(combinedFactor)}u128, 1u128), 250000u128))`,
+    );
+  }
+
+  // The output needs to be submitted as an RADIX transaction
```
paulbalaji, line 116: is it possible to also submit via the typescript providers we have?

PR #7114, opened 2025-09-29
chore: add radix warp route reader

troykessler:
### Description

This PR adds the Radix warp route reader class

### Drive-by changes

-

### Related issues

-

### Backward compatibility

-

### Testing

Manual Testing

### DIFF
--- a/typescript/sdk/src/token/RadixWarpRouteReader.ts
+++ b/typescript/sdk/src/token/RadixWarpRouteReader.ts
@@ -61,7 +61,10 @@ export class RadixWarpRouteReader {
    * @returns The derived token type, which can be one of: collateralVault, collateral, native, or synthetic.
    */
   async deriveTokenType(warpRouteAddress: Address): Promise<TokenType> {
-    const token = await this.findToken(warpRouteAddress);
+    const token = await this.signer.query.warp.getToken({
+      token: warpRouteAddress,
+    });
+    assert(token, `Failed to find token for address ${warpRouteAddress}`);
 
     switch (token.token_type) {
       case RadixTokenTypes.COLLATERAL:
@@ -84,7 +87,10 @@ export class RadixWarpRouteReader {
   async fetchMailboxClientConfig(
     routerAddress: Address,
   ): Promise<MailboxClientConfig> {
-    const token = await this.findToken(routerAddress);
+    const token = await this.signer.query.warp.getToken({
+      token: routerAddress,
+    });
+    assert(token, `Failed to find token for address ${routerAddress}`);
 
     const config: MailboxClientConfig = {
       mailbox: token.mailbox,
@@ -110,7 +116,10 @@ export class RadixWarpRouteReader {
     type: TokenType,
     warpRouteAddress: Address,
   ): Promise<HypTokenConfig> {
-    const token = await this.findToken(warpRouteAddress);
+    const token = await this.signer.query.warp.getToken({
+      token: warpRouteAddress,
+    });
+    assert(token, `Failed to find token for address ${warpRouteAddress}`);
 
     return {
       type,
@@ -149,14 +158,5 @@ export class RadixWarpRouteReader {
         routerConfig.gas,
       ]),
     );
-  }
-
-  private async findToken(warpRouteAddress: string) {
-    const token = await this.signer.query.warp.getToken({
-      token: warpRouteAddress,
-    });
-    assert(token, `Failed to find token for address ${warpRouteAddress}`);
-
-    return token;
   }
 }
--- a/typescript/radix-sdk/src/warp/query.ts
+++ b/typescript/radix-sdk/src/warp/query.ts
@@ -9,7 +9,12 @@ import { utils } from 'ethers';
 import { assert } from '@hyperlane-xyz/utils';
 
 import { RadixBase } from '../utils/base.js';
-import { EntityDetails, EntityField, Receipt } from '../utils/types.js';
+import {
+  EntityDetails,
+  EntityField,
+  RadixTokenTypes,
+  Receipt,
+} from '../utils/types.js';
 
 export class RadixWarpQuery {
   protected networkId: number;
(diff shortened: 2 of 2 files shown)


typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Derives the token type for a given Warp Route address using specific methods
+   *
+   * @param warpRouteAddress - The Warp Route address to derive the token type for.
+   * @returns The derived token type, which can be one of: collateralVault, collateral, native, or synthetic.
+   */
+  async deriveTokenType(warpRouteAddress: Address): Promise<TokenType> {
+    const token = await this.findToken(warpRouteAddress);
```
yjamin, line 64: I think the `findToken` method should return the `RadixTokenTypes` directly.

typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+   * @returns The base metadata for the Warp Route contract, including the mailbox, owner, hook, and ism.
+   */
+  async fetchMailboxClientConfig(
+    routerAddress: Address,
+  ): Promise<MailboxClientConfig> {
+    const token = await this.findToken(routerAddress);
+
+    const config: MailboxClientConfig = {
+      mailbox: token.mailbox,
+      owner: token.owner,
```
yjamin, line 91: Should this be the token owner or the mailbox owner?

typescript/sdk/src/token/RadixWarpRouteReader.ts
xeno097: I agree on this

typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+    const remoteRouters = await this.fetchRemoteRouters(warpRouteAddress);
+    const destinationGas = await this.fetchDestinationGas(warpRouteAddress);
+
+    return {
+      ...baseMetadata,
+      ...tokenConfig,
+      remoteRouters,
+      destinationGas,
+      type,
+    } as DerivedTokenRouterConfig;
```
xeno097, line 54: Considering that this is new code can we try to avoid this type assertion as it makes it more difficult to find type errors? I'd use the schema to parse the object so we can be sure the data type is in the expected shape instead

typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+        routerConfig.receiver_domain,
+        routerConfig.gas,
+      ]),
+    );
+  }
+
+  private async findToken(warpRouteAddress: string) {
+    const token = await this.signer.query.warp.getToken({
+      token: warpRouteAddress,
+    });
```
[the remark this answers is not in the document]
xeno097: +1, this could just be in the `deriveTokenType` method and others methods can call it instead

typescript/radix-sdk/src/utils/types.ts
```diff
@@ -33,6 +33,11 @@ export enum RadixIsmTypes {
   NOOP_ISM = 'NoopIsm',
 }
 
+export enum RadixTokenTypes {
+  COLLATERAL = 'Collateral',
+  SYNTHETIC = 'Synthetic',
+}
```
xeno097, line 39: I think we should have this type in the same file as the `RadixWarpRouteReader` and not export it as sdk users might think that this and the `TokenType` are interchangeable when they are not

typescript/radix-sdk/src/index.ts
```diff
@@ -29,7 +29,7 @@ const NETWORKS = {
 };
 
 export { NetworkId };
-export { RadixIsmTypes } from './utils/types.js';
+export { RadixIsmTypes, RadixTokenTypes } from './utils/types.js';
```
xeno097, line 32: I think we shouldn't export this, if we want to have an externally usable `RadixTokenTypes` it should be derived from the already existing `TokenType` so for example it should be something like this:

`export type RadixTokenTypes = TokenType.collateral | TokenType.synthetic` so that `RadixTokenTypes` can be used with `TokenType` but not viceversa

typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+      mailbox: token.mailbox,
+      owner: token.owner,
+    };
+
+    if (token.ism) {
+      const derivedIsm = await this.ismReader.deriveIsmConfig(token.ism);
+      config.interchainSecurityModule = derivedIsm;
+    }
+
+    return config;
```
xeno097, line 99: I think you are not deriving the hook if it is set?

typescript/sdk/src/token/RadixWarpRouteReader.ts
xeno097: token owner as the mailbox client "abstraction" is used for defining contracts that interact/have message processing capabilities

typescript/sdk/src/token/RadixWarpRouteReader.ts
```diff
@@ -0,0 +1,162 @@
...
+  ): Promise<HypTokenConfig> {
+    const token = await this.findToken(warpRouteAddress);
+
+    return {
+      type,
+      token: warpRouteAddress,
+      name: token.name,
+      symbol: token.symbol,
+      decimals: token.divisibility,
+    } as HypTokenConfig;
```
xeno097, line 121: once again can we try to reduce type assertions as much as possible? if there are fields from the token metadata that might be undefined but are required by the `HypTokenConfig` type we should assert that they exist instead of silencing the error with a type assertion.

typescript/sdk/src/token/RadixWarpRouteReader.ts
troykessler: will revert to original implementation

typescript/sdk/src/token/RadixWarpRouteReader.ts
troykessler: there is lots of evm specific stuff expected here which will never be used in cosmos or radix

PR #7117, opened 2025-09-29
Jeff/radix lander integration

kamiyaa:


rust/main/lander/src/adapter/chains/radix/precursor.rs
```diff
@@ -33,6 +33,41 @@ pub struct RadixTxPrecursor {
     pub tx_hash: Option<H512>,
 }
 
+impl std::fmt::Debug for RadixTxPrecursor {
```
kamiyaa, line 36: Implemented Debug manually to omit `encoded_arguments` so logs don't get huge.

rust/main/lander/src/adapter/chains/radix/conf.rs
ameten, line 1: We can add a few tests for this function, I think.

rust/main/lander/src/adapter/chains/radix/precursor.rs
ameten: Try use `#[debug(skip)]` annotation on the field you would like to skip on logging.

ameten (approved):

rust/main/lander/src/adapter/chains/radix/precursor.rs
kamiyaa: doesn't seem to work

PR #7125, opened 2025-09-29
fix: add polymer ism type enum

RnkSngh:
### Description

"Reserving" the ISM enum "slot" so that we don't have multiple different ISMs with the same type ~

### Drive-by changes

None

### Related issues

None

### Backward compatibility

Should be backwards compatible since we're only adding a new enum value.

## Summary by CodeRabbit

* **New Features**
  * Added “Polymer” as a selectable interchain security module type, expanding available options for configuring interchain security.

* **Chores**
  * Updated type listings to include the new module option.


yorhodes:
> missing a changeset :'(

I did consider this but decided that we actually dont want to version this (which would imply a bytecode change)

PR #7141, opened 2025-10-02
chore: alt-multi-vm

troykessler:
### Description

This PR introduces a multi-vm interface for non ethereum VMs to allow inclusions of new VMs like Radix or Aleo more easily in the future

### Testing

Unit Tests, e2e tests, manual testing


typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,661 @@
...
+  default_hook: string;
+  required_hook: string;
+  message_sent: number;
+  message_received: number;
+};
+
+export type ReqDelivered = { mailbox_id: string; message_id: string };
+export type ResDelivered = boolean;
+
+export enum IsmType {
```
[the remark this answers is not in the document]
troykessler: I'm good with a separate file in utils which contains the source of truth defintion for the entire monorepo

typescript/utils/src/mingas.ts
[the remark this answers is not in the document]
troykessler: I think this should be one big goal of the multi vm project or? My big priority was to speed up development and review time for new VM implementations for us

xeno097 (commented):
I asked this in slack but i'll say this here too, could we split this PR in possibly:
- new interfaces and type definitions
- cosmos sdk changes
- cli changes
cause reviewing all this together is kinda difficult, i would like to focus on the new code first as it will be the basis for the multi vm refactor

typescript/sdk/src/hook/AltVmHookReader.ts
```diff
@@ -0,0 +1,99 @@
...
+      );
+      oracleConfig[name] = {
+        gasPrice:
+          igp.destination_gas_configs[domain_id].gas_oracle?.gas_price ?? '',
+        tokenExchangeRate:
+          igp.destination_gas_configs[domain_id].gas_oracle
+            ?.token_exchange_rate ?? '',
+        tokenDecimals: nativeToken?.decimals,
+      };
+    });
```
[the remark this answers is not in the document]
troykessler: keys in json are always parsed to string automatically

typescript/widgets/src/walletIntegrations/cosmos.ts
```diff
@@ -189,22 +188,27 @@ export function useCosmosTransactionFns(
...
+          tx.transaction,
+        ]);
+
+        result = txResponse;
+        txDetails = {
+          ...txResponse,
+          hash: txResponse.transactionHash,
+          tx: new Uint8Array(),
+          rawLog: txResponse.rawLog || '',
+        };
```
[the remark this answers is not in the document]
troykessler: refactored this

typescript/sdk/src/token/AltVMWarpRouteReader.ts
```diff
@@ -19,28 +14,20 @@ import { ChainNameOrId } from '../types.js';
...
-      metadataManager,
-      cosmosProviderOrSigner,
-    );
-    this.ismReader = new CosmosNativeIsmReader(
-      metadataManager,
-      cosmosProviderOrSigner,
-    );
+    this.hookReader = new AltVMHookReader(metadataManager, provider);
+    this.ismReader = new AltVMIsmReader(metadataManager, provider);
   }
```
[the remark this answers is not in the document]
troykessler: maybe later the hookReader will be needed

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
...
+export type ResRemoteTransfer = {
+  messageId: string;
+};
+
+export type ResSignAndBroadcast = {
+  height: number;
+  transactionHash: string;
+};
+
+export interface IProvider<T = any> {
```
antigremlin, line 414: Would it make sense to split this interface into smaller, more focused interfaces. For example, we could have core-related and warp-related functions grouped together into smaller interfaces. Or we can go further and have something like Mailbox, ISM, Hook, Token, Rotuer interfaces.

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  public static async createSigners(
+    metadataManager: ChainMetadataManager,
+    chains: string[],
+    key: SignerKeyProtocolMap,
+    strategyConfig: Partial<ExtendedChainSubmissionStrategy>,
+  ) {
+    const signers: ChainMap<AltVM.ISigner> = {};
+
+    for (const chain of chains) {
```
antigremlin, line 168: Why do we create signers up-front, while creating providers on demand? I mean that `providerFactory.get()` instantiates and connects a provider but `signerFactory.get()` just returns a signer that was created at some initialisation time.

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,202 @@
...
+    const metadata = this.metadataManager.getChainMetadata(chain);
+
+    if (!this.supports(metadata.protocol)) {
+      throw new Error(
+        `Chain ${chain} with protocol type ${metadata.protocol} not supported in AltVM`,
+      );
+    }
+
+    const { provider } = ALT_VM_SUPPORTED_PROTOCOLS[metadata.protocol]!;
+    return provider.connect(metadata.rpcUrls[0].http);
```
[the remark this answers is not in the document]
troykessler: so for now all the providers also just took one rpc (always first index). Will change the interface now however so that it takes an array of rpcs

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  public getSupportedProtocols(): ProtocolType[] {
+    return Object.keys(ALT_VM_SUPPORTED_PROTOCOLS) as ProtocolType[];
+  }
+
+  public supports(protocol: ProtocolType) {
+    return !!ALT_VM_SUPPORTED_PROTOCOLS[protocol];
+  }
+
+  public getGas(protocol: ProtocolType) {
```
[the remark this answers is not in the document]
troykessler: yeah this is probably the better method name

typescript/cli/src/context/altvm.ts
troykessler: Yes that is exactly the difference between the provider and signer factory. The difference exists because for all signing commands (core/warp deploy,apply) we know exactly for which chains we need signers for. On the provider side it is expected to create providers dynamically if requested. This structure was introduced by the original `multiProvider` which gives you providers dynamically but initializes signers for signing commands when the chains are known. Else of course our AltVM factory classes would be much smaller and more generic

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
...
+
+// ### QUERY BASE ###
+export type ReqGetBalance = { address: string; denom: string };
+export type ResGetBalance = bigint;
+
+export type ReqGetTotalSupply = { denom: string };
+export type ResGetTotalSupply = bigint;
+
+export type ReqEstimateTransactionFee = {
+  transaction: any;
```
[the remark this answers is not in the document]
troykessler: Yes, added the generic type for the transactions

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
...
+  populateSetTokenIsm(req: ReqSetTokenIsm): Promise<T>;
+
+  populateEnrollRemoteRouter(req: ReqEnrollRemoteRouter): Promise<T>;
+
+  populateUnenrollRemoteRouter(req: ReqUnenrollRemoteRouter): Promise<T>;
+
+  populateRemoteTransfer(req: ReqRemoteTransfer): Promise<T>;
+}
+
+export interface ISigner<T = any, R extends ResSignAndBroadcast = any>
```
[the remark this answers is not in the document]
troykessler: Would resolve this and continue discussing in the comment above

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
+import { MINIMUM_GAS } from './mingas.js';
+import { ProtocolType } from './types.js';
+
+// ### QUERY BASE ###
+export type ReqGetBalance = { address: string; denom: string };
```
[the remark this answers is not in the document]
troykessler: You did a very good pro/con list on that. For me the most important aspect was uniformity and consistency.

Although using a hybrid solution would be a compromise, I don't like this structure. A user sometimes expects direct parameters and then sometimes complex ones, I think this is not nice to use. Furthermore, where would we draw the line for methods being to complex, 2 params, 3 params? I think we should either decide for direct or complex params for every method so it stays consistent.

Personally, I like the complex (current) structure more. You're absolutely right that I took inspiration from the cosmos protobuf message protocol and since it was designed to be an language agnostic interface so I figured that it might be a good example of how to do it. Furthermore, I really don't like direct params if the same argument type repeats itself like `method(a: number, b: number, c: number)`, it happened a lot to me that I started mixing up the params in the wrong order where even type safety does not protect you. Having "complex" object types on the other hand makes sure that you always map the args correctly. It's true that IDE's might not be able to auto suggest the direct params more nicely or that other protocols use more direct params (they also use simpler methods) but I think the benefits of the current implementation outweighs that. 

Also happy to discuss further on this of course!

typescript/utils/src/mingas.ts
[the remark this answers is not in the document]
paulbalaji: I'm happy to defer this to later, my question is more to make sure we've considered it but not essential for this PR

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+
+export type ReqGetTotalSupply = { denom: string };
+
+export type ReqEstimateTransactionFee<T> = {
+  transaction: T;
+  estimatedGasPrice: string;
+  sender: string;
+  senderPubKey?: string;
```
xeno097, line 13: what is the difference between `sender` and `senderPubKey`?

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+  memo?: string;
```
xeno097, line 14: i think memo is something cosmos specific, shouldn't this be part of the provided transaction data instead?

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+export type ResGetMailbox = {
+  address: string;
+  owner: string;
+  localDomain: number;
+  defaultIsm: string;
+  defaultHook: string;
+  requiredHook: string;
+  messageSent: number;
+  messageReceived: number;
+};
```
xeno097, line 34: What is the difference between this and the derived config type? that we already have defined in the sdk?

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
+import { MINIMUM_GAS } from './mingas.js';
+import { ProtocolType } from './types.js';
+
+// ### QUERY BASE ###
+export type ReqGetBalance = { address: string; denom: string };
```
xeno097, line 5: nit: I would prefer if instead of following the naming convention `Req<action><resource>` we use `<action><resource><context>`. Example:
- ReqGetBalance -> GetBalanceOptions
- ResGetBalance -> GetBalanceResponse

it makes it easier to read to me

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+
+export type ReqGetTotalSupply = { denom: string };
```
xeno097, line 7: i think that numeric fields should be numbers instead of strings as it make it less error prone and less confusing when having to work on math operations as the string might be either the bigint represention or the decimal one. We should stick to using bigint for internal usage and string/number for representing the numbers in a human friendly way in logs or error messages. Otherwise a comment should state explicitly what the expected format is for the field

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+export type ResGetMerkleTreeHook = {
+  address: string;
+};
+
+// ### QUERY WARP ###
+
+export enum TokenType {
+  COLLATERAL = 'COLLATERAL',
+  SYNTHETIC = 'SYNTHETIC',
+}
```
xeno097, line 135: I think there are missing token types here

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+  address: string;
+  owner: string;
+  tokenType: TokenType;
+  mailboxId: string;
+  ismId: string;
+  originDenom: string;
+  name: string;
+  symbol: string;
+  description: string;
+  divisibility: number;
```
xeno097, line 148: shouldn't this be decimals

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+
+export type ReqGetBridgedSupply = { tokenId: string };
+
+export type ReqQuoteRemoteTransfer = {
+  tokenId: string;
+  destinationDomainId: number;
+  customHookId: string;
+  customHookMetadata: string;
+};
+export type ResQuoteRemoteTransfer = { denom: string; amount: bigint };
```
xeno097, line 169: nit: in general i'd change the `denom` naming to ` tokenSymbol` instead

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+
+export type ReqCreateMailbox = {
+  signer: string;
+  domainId: number;
+  defaultIsmId: string;
+};
+export type ResCreateMailbox = { mailboxId: string };
+
+export type ReqSetDefaultIsm = {
+  signer: string;
```
xeno097, line 181: i see that this types define the signer field but it is then ignored in the external API interface, should really define it if it is a internal implementation detail?

typescript/utils/src/index.ts
```diff
@@ -202,3 +202,5 @@ export { assert } from './validation.js';
 export { BaseValidator, ValidatorConfig } from './validator.js';
 export { tryParseJsonOrYaml } from './yaml.js';
 export { createServiceLogger } from './logging.js';
+export { MINIMUM_GAS, MINIMUM_GAS_ACTION } from './mingas.js';
+export * as AltVM from './altvm.js';
```
xeno097, line 206: I'd avoid barrel exports as this might contribute to the bundling issues that we want to solve in the long term

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,196 @@
...
+  public supports(protocol: ProtocolType) {
+    return !!ALT_VM_SUPPORTED_PROTOCOLS[protocol];
+  }
+
+  public getMinGas(protocol: ProtocolType) {
+    if (!this.supports(protocol)) {
+      throw new Error(`Protocol type ${protocol} not supported in AltVM`);
+    }
+
+    const { gas } = ALT_VM_SUPPORTED_PROTOCOLS[protocol]!;
```
xeno097, line 54: why not:
```typescript
const protocolDefinition = ALT_VM_SUPPORTED_PROTOCOLS[protocol];


if (!protocolDefinition) {
      throw new Error(`Protocol type ${protocol} not supported in AltVM`);
    }

return protocolDefinition.gas
```
so you don't have to silence ts errors as it is able to infer that the type is defined after the if branch?

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,196 @@
...
+    const metadata = this.metadataManager.getChainMetadata(chain);
+
+    if (!this.supports(metadata.protocol)) {
+      throw new Error(
+        `Chain ${chain} with protocol type ${metadata.protocol} not supported in AltVM`,
+      );
+    }
+
+    const { provider } = ALT_VM_SUPPORTED_PROTOCOLS[metadata.protocol]!;
+    return provider.connect(metadata.rpcUrls.map((rpc) => rpc.http));
```
xeno097, line 81: ditto https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/7141/files#r2408727446

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      return rawConfig.privateKey;
+    }
+
+    // 3. Finally, if no key flag or strategy was provided we prompt the user
+    // for the private key
+    return password({
+      message: `Please enter the private key for chain ${chain} (will be re-used for other chains with the same protocol type)`,
+    });
+  }
```
xeno097, line 153: I think that one thing that we are loosing here compared to the existing logic is the ability reuse private keys when they are provided for a chain and that same key could be reused for another one as they share the same protocol type which would cause once again the user to be prompted multiple times for the same pk as you are not caching it by protocol here even if the message says otherwise

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
...
+export type ResGetMailbox = {
+  address: string;
+  owner: string;
+  localDomain: number;
+  defaultIsm: string;
+  defaultHook: string;
+  requiredHook: string;
+  messageSent: number;
+  messageReceived: number;
+};
```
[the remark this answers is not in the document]
troykessler: Checked all contracts again, you're right it was Cosmos specific, changed it to nonce (what `messageSent` basically is) and what all the other protocols support

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,658 @@
...
+export type ResEstimateTransactionFee = {
+  gasUnits: number;
+  gasPrice: number;
+  fee: number;
+};
+
+// ### QUERY CORE ###
+
+export type ReqGetMailbox = { mailboxId: string };
```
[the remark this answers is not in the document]
troykessler: Will change it to address

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+  populateSetTokenOwner(req: ReqSetTokenOwner): Promise<T>;
+
+  populateSetTokenIsm(req: ReqSetTokenIsm): Promise<T>;
+
+  populateEnrollRemoteRouter(req: ReqEnrollRemoteRouter): Promise<T>;
+
+  populateUnenrollRemoteRouter(req: ReqUnenrollRemoteRouter): Promise<T>;
+
+  populateRemoteTransfer(req: ReqRemoteTransfer): Promise<T>;
+}
```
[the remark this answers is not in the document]
troykessler: we have one, it is called `estimateTransactionFee`

typescript/utils/src/altvm.ts
troykessler: It is only ignored in the Signer interface since we can infer the signer address from the private key. In the provider interface however when we populate (or create) the transactions we need it

typescript/utils/src/altvm.ts
troykessler: sender is the sender address (will rename to `senderAddress`) while `senderPubKey` is the public key, on some protocols these are different and the public key can't be derived from the address

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,657 @@
...
+
+  getBridgedSupply(req: ReqGetBridgedSupply): Promise<bigint>;
+
+  quoteRemoteTransfer(
+    req: ReqQuoteRemoteTransfer,
+  ): Promise<ResQuoteRemoteTransfer>;
+
+  // ### POPULATE CORE ###
+
+  populateCreateMailbox(req: ReqCreateMailbox): Promise<T>;
```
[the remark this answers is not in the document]
troykessler: will do

typescript/cli/src/tests/cosmosnative/commands/helpers.ts
```diff
@@ -30,26 +29,27 @@ export async function deployCollateralToken(
...
+  const signer = await CosmosNativeSigner.connectWithSigner(
+    metadata.rpcUrls.map((rpc) => rpc.http),
     wallet,
     {
-      gasPrice: GasPrice.fromString(
-        `${metadata.gasPrice?.amount}${metadata.gasPrice?.denom}`,
-      ),
+      metadata,
     },
   );
```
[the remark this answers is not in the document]
troykessler: this is checked in the signer classes

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,663 @@
...
+    req: ReqSetInterchainGasPaymasterHookOwner,
+  ): Promise<T>;
+
+  populateSetDestinationGasConfig(req: ReqSetDestinationGasConfig): Promise<T>;
+
+  populateCreateValidatorAnnounce(req: ReqCreateValidatorAnnounce): Promise<T>;
+
+  // ### POPULATE WARP ###
+
+  populateCreateCollateralToken(req: ReqCreateCollateralToken): Promise<T>;
```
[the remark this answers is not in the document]
troykessler: sorry but I still don't get it, what do I need to change in the interface to make it multisig compatible?

typescript/utils/src/altvm.ts
troykessler: I think the difference in the startup time can be hardly measured with having 100 (uncompiled) lines of code more or less. If we really want to make things faster we should consider only loading code for the relevant chains (e.g. only don't load the 10,000+ lines of code and deps for evm if we only warp deploy cosmos chains).

Regardless of course splitting this up is generally considered better code practice, but for the sake of merging this PR this week I would love to move the discussion of how to split the interface into which groups to the multi-vm design doc if that is fine for you

typescript/utils/src/altvm.ts
```diff
@@ -525,7 +525,11 @@ export interface ISigner<T = any, R extends ResSignAndBroadcast = any>
   extends IProvider<T> {
   getSignerAddress(): string;
 
-  signAndBroadcast(transactions: T[]): Promise<R>;
+  supportsMultiTransactions(): boolean;
+
+  sendAndConfirmTransaction(transaction: T): Promise<R>;
+
+  sendAndConfirmMultiTransactions(transactions: T[]): Promise<R>;
```
paulbalaji, line 555: is "multitransaction" cosmos lingo? when i see "multi" I tend to think "multisig"

perhaps consider "batching":
- `supportsTransactionBatching`
- `sendAndConfirmTransaction`
- `sendAndConfirmBatch`

or something like that

paulbalaji: probably the `batching` term, as it aligns with what we call it in agents/agent configs

troykessler: renamed multi to batch now

typescript/utils/src/index.ts
troykessler: Agree, want to keep it in the short term in the AltVM namespace however since this entire file will move into the new interface package that will be created one multi-vm comes along

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,682 @@
...
+  getSetTokenIsmTransaction(req: ReqSetTokenIsm): Promise<T>;
+
+  getEnrollRemoteRouterTransaction(req: ReqEnrollRemoteRouter): Promise<T>;
+
+  getUnenrollRemoteRouterTransaction(req: ReqUnenrollRemoteRouter): Promise<T>;
+
+  getRemoteTransferTransaction(req: ReqRemoteTransfer): Promise<T>;
+}
+
+export interface ISigner<T, R extends Receipt = any> extends IProvider<T> {
```
antigremlin, line 548: This could be `R extends Receipt = Receipt` but even better would be to remove the default and propagate correct generic parameters like we did with transactions.

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,682 @@
...
+  getSignerAddress(): string;
+
+  supportsTransactionBatching(): boolean;
+
+  sendAndConfirmTransaction(transaction: T): Promise<R>;
+
+  sendAndConfirmBatchTransactions(transactions: T[]): Promise<R>;
```
antigremlin, line 555: This shouldn't return a single receipt, I don't think. Is it a Cosmos-specific convention to return one receipt in response to a batch of transactions?

antigremlin (commented):
These two also seem important.

typescript/sdk/src/providers/transactions/submitter/altvm/AltVMJsonRpcTxSubmitter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(
+    public readonly multiProvider: MultiProvider,
+    public readonly altVmSigner: AltVM.ISignerFactory,
+    public readonly config: { chain: string },
+  ) {
+    this.signer = this.altVmSigner.get(this.config.chain);
+  }
+
+  public async submit(...txs: AnnotatedAltVMTransaction[]): Promise<any[]> {
```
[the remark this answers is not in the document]
troykessler: It is used here, your comment is also marked as outdated, maybe you need to pull the latest changes

typescript/utils/src/altvm.ts
troykessler: propagating the correct generic receipt now

typescript/cli/src/context/altvm.ts
```diff
@@ -0,0 +1,210 @@
...
+  public getMinGas(protocol: ProtocolType) {
+    const protocolDefinition = ALT_VM_SUPPORTED_PROTOCOLS[protocol];
+
+    if (!protocolDefinition) {
+      throw new Error(`Protocol type ${protocol} not supported in AltVM`);
+    }
+
+    return protocolDefinition.gas;
+  }
+}
```
xeno097, line 60: Isn't this class a bit misleading? it is called `AltVMFactory` but looking at it as a whole it just looks like a wrapper on the `ALT_VM_SUPPORTED_PROTOCOLS` to get gas and supported protocols data but nothing is done about the signers and providers in here, wouldn't it be better if this class also allowed to get providers and signers so on the cli side you can use only this class instead of 3 different classes depending what you want to achieve?

xeno097: To be clear I'm not saying that you should eliminate the `AltVMProviderFactory` and `AltVMSignerFactory` classes but instead those 2 classes should be inside the `AltVMFactory` so you can get a signer or provider

typescript/cli/src/context/types.ts
```diff
@@ -56,7 +58,7 @@ export interface CommandContext
 export interface WriteCommandContext extends Omit<CommandContext, 'key'> {
   key: SignerKeyProtocolMap;
   signer: ethers.Signer;
-  multiProtocolSigner?: MultiProtocolSignerManager;
+  altVmSigner: AltVMSignerFactory;
```
xeno097, line 61: why is this required now? what happens if a command does not need a signer?

typescript/cli/src/deploy/warp.ts
```diff
@@ -388,16 +383,25 @@ export async function runWarpRouteApply(
...
+            ...config,
+            owner: await context.multiProvider.getSignerAddress(chain),
+          };
+        }
+        default: {
+          return {
+            ...config,
+            owner: context.altVmSigner.get(chain).getSignerAddress(),
+          };
+        }
```
xeno097, line 402: this will crash the deployment if we have unsupported chains with foreign deployment in the deploy config. After merging https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/7156 this should hopefully be reported in the e2e tests too

typescript/cli/src/ism/read.ts
```diff
@@ -19,19 +19,32 @@ export async function readIsmConfig({
...
+      const ismReader = new EvmIsmReader(context.multiProvider, chain);
+      const config = await ismReader.deriveIsmConfig(address);
+      const stringConfig = stringifyObject(config, resolveFileFormat(out), 2);
+      if (!out) {
+        logBlue(`ISM Config at ${address} on ${chain}:`);
+        log(stringConfig);
+      } else {
+        writeFileAtPath(out, stringConfig + '\n');
+        logBlue(`ISM Config written to ${out}.`);
+      }
```
xeno097, line 33: 1. Instead of duplicating this code block in both switch branches, can't we define a `config` variable outside of the switch and after it is assigned in one of the 2 branches we have this code block at the end of this function? 
2. Can we refrain from using switch statements if we only have 2 cases? I think using `if ... else ...` in this case is perfectly valid

typescript/cli/src/deploy/utils.ts
```diff
@@ -89,9 +94,20 @@ export async function runDeployPlanStep({
...
+
+  switch (context.multiProvider.getProtocol(chain)) {
+    case ProtocolType.Ethereum: {
+      address = await multiProvider.getSigner(chain).getAddress();
+      break;
+    }
+    default: {
+      address = altVmSigner.get(chain).getSignerAddress();
+    }
+  }
```
xeno097, line 110: ditto: on switch statements with only 2 alternatives

typescript/cli/src/tests/cosmosnative/core/core-apply.e2e-test.ts
```diff
@@ -34,30 +35,31 @@ describe('hyperlane core apply e2e tests', async function () {
...
+    ProtocolReceipt<ProtocolType.CosmosNative>
+  >;
   let initialOwnerAddress: Address;
 
   before(async () => {
     const chainMetadata: ChainMetadata = readYamlOrJson(CHAIN_1_METADATA_PATH);
 
     const wallet = await DirectSecp256k1Wallet.fromKey(
-      Buffer.from(HYP_KEY, 'hex'),
+      new Uint8Array(Buffer.from(HYP_KEY, 'hex')),
```
xeno097, line 48: why not use `Uint8Array.from` directly?

typescript/cli/src/tests/cosmosnative/core/core-deploy.e2e-test.ts
```diff
@@ -52,25 +56,23 @@ describe('hyperlane cosmosnative core deploy e2e tests', async function () {
...
       },
     );
 
-    initialOwnerAddress = signer.account.address;
+    initialOwnerAddress = signer.getSignerAddress();
   });
 
   describe('hyperlane cosmosnative core deploy', () => {
-    it('should create a core deployment with the signer as the mailbox owner', async () => {
+    it('1. should create a core deployment with the signer as the mailbox owner', async () => {
```
xeno097, line 75: i think that we can avoid adding numbers to the test description as we will need to update the indices if we add another test in between the existing ones

typescript/cli/src/tests/cosmosnative/warp/warp-deploy.e2e-test.ts
```diff
@@ -69,7 +69,8 @@ describe('hyperlane warp deploy e2e tests', async function () {
 
   before(async function () {
     const wallet = await DirectSecp256k1Wallet.fromKey(
-      Buffer.from(HYP_KEY, 'hex'),
+      new Uint8Array(Buffer.from(HYP_KEY, 'hex')),
```
xeno097, line 72: ditto: `Uint8Array.from`

typescript/cli/src/tests/cosmosnative/warp/warp-read.e2e-test.ts
```diff
@@ -63,7 +63,8 @@ describe('hyperlane warp read e2e tests', async function () {
 
   before(async function () {
     const wallet = await DirectSecp256k1Wallet.fromKey(
-      Buffer.from(HYP_KEY, 'hex'),
+      new Uint8Array(Buffer.from(HYP_KEY, 'hex')),
```
xeno097, line 66: ditto `Uint8Array.from`

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,677 @@
...
+
+  estimateTransactionFee(
+    req: ReqEstimateTransactionFee<T>,
+  ): Promise<ResEstimateTransactionFee>;
+
+  // ### QUERY CORE ###
+
+  getMailbox(req: ReqGetMailbox): Promise<ResGetMailbox>;
+
+  delivered(req: ReqDelivered): Promise<boolean>;
```
xeno097, line 438: I'd rename this to `isMessageDelivered`

typescript/cosmos-sdk/src/clients/provider.ts
```diff
@@ -0,0 +1,757 @@
...
+      address: token.id,
+      owner: token.owner,
+      tokenType: token_type,
+      mailboxAddress: token.origin_mailbox,
+      ismAddress: token.ism_id,
+      originDenom: token.origin_denom,
+      name: '',
+      symbol: '',
+      description: '',
+      decimals: 0,
```
xeno097, line 372: Isn't this bad? how can we get the token metadata on a cosmos token? Imagine if we have some code that expects these field not to be empty and it fails because this data is not populated in the first place

typescript/sdk/src/providers/transactions/submitter/submitterBuilderGetter.ts
```diff
@@ -171,6 +170,10 @@ export async function getSubmitter<TProtocol extends ProtocolType>(
     ...defaultSubmitterFactories,
   };
 
+  for (const protocol of altVmSigner.getSupportedProtocols()) {
+    mergedSubmitterRegistry[protocol] = MULTI_VM_SUBMITTERS_FACTORIES;
```
antigremlin, line 174: There is an inconsistency here between `.getSupportedProtocols()` - which is only implemented in the CLI - and `MULTI_VM_SUBMITTERS_FACTORIES` - which is defined in this file. If we just initialise the `defaultSubmitterFactories` as before, we can avoid passing the `altVmSigner` here altogether.

Later on (after this PR), we can rearrange the code so that the knowledge about which protocols are implemented doesn't reside in the CLI.

typescript/sdk/src/token/adapters/CosmosModuleTokenAdapter.ts
```diff
@@ -260,34 +255,32 @@ export class CosmNativeHypCollateralAdapter
...
-        },
+        destinationProtocol,
+      ),
+      amount: params.weiAmountOrId.toString(),
+      customHookAddress: params.customHook || '',
+      customHookMetadata: '',
+      gasLimit: router.gas,
+      maxFee: {
+        denom: params.interchainGas.addressOrDenom || '',
+        amount: params.interchainGas.amount.toString(),
```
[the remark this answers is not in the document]
troykessler: Agree, but would leave it to another PR since this is not the scope of this one

typescript/cli/src/deploy/warp.ts
troykessler: This issue was unfortunately existing before and this PR does not intend to fix it. Looking forward to your PR getting merged fixing this issue

typescript/cli/src/tests/commands/core.ts
```diff
@@ -9,6 +9,8 @@ import { readYamlOrJson } from '../../utils/files.js';
 
 import { localTestRunCmdPrefix } from './helpers.js';
 
+$.verbose = true;
```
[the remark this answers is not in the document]
troykessler: will remove it

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,677 @@
...
+export type ReqCreateCollateralToken = {
+  signer: string;
+  mailboxAddress: string;
+  originDenom: string;
+};
+export type ResCreateCollateralToken = {
+  tokenAddress: string;
+};
+
+export type ReqCreateSyntheticToken = {
```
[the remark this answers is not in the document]
troykessler: Good catch, apparently did not merge the cosmos/radix interface, of course we need this info here

typescript/utils/src/altvm.ts
```diff
@@ -0,0 +1,677 @@
...
+export type ResCreateValidatorAnnounce = {
+  validatorAnnounceId: string;
+};
+
+// ### POPULATE WARP ###
+
+export type ReqCreateCollateralToken = {
+  signer: string;
+  mailboxAddress: string;
+  originDenom: string;
```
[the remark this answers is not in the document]
troykessler: Yes because the collateral usually just stored the token address/denom and not the other metadata, will rename it to `collateralDenom`

typescript/cli/src/tests/cosmosnative/commands/helpers.ts
```diff
@@ -65,23 +65,24 @@ export async function deploySyntheticToken(
...
 
-  const { response } = await signer.createSyntheticToken({
-    origin_mailbox: mailbox,
+  const { tokenAddress } = await signer.createSyntheticToken({
+    mailboxAddress: mailbox,
+    name: '',
+    denom: '',
+    decimals: 0,
   });
```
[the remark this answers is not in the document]
troykessler: for cosmos we don't need that

.changeset/moody-hounds-hang.md
```diff
@@ -0,0 +1,9 @@
+---
+"@hyperlane-xyz/cosmos-sdk": major
+"@hyperlane-xyz/widgets": major
+"@hyperlane-xyz/utils": major
+"@hyperlane-xyz/cli": major
+"@hyperlane-xyz/sdk": major
+---
+
+feat: add AltVM interface
```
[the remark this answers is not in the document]
troykessler: will do

typescript/cosmos-sdk/src/clients/provider.ts
antigremlin: That's weird. What about the `DenomMetadata` in the bank module? It can accommodate the name, symbol, and description. And it has `DenumUnits` where the decimals are set (called "exponent") in Cosmos SDK.

Aren't we using any of this?

typescript/sdk/src/token/altVMDeploy.ts
```diff
@@ -38,9 +44,7 @@ export class CosmosNativeDeployer {
 
     const configMapToDeploy = objFilter(
```
antigremlin, line 45: This variable reverts to type `any` and so further errors are hidden. You can delcare it explicitly with
```
type ConfigMap = WarpRouteDeployConfigMailboxRequired;
const configMapToDeploy: ConfigMap = objFilter(...);
```

typescript/sdk/src/token/altVMDeploy.ts
```diff
@@ -49,9 +53,7 @@ export class CosmosNativeDeployer {
       const config = configMapToDeploy[chain];
       assert(config, `No config configured for ${chain}`);
 
-      this.logger.info(
-        `Deploying ${config.type} token to Cosmos Native chain ${chain}`,
-      );
+      this.logger.info(`Deploying ${config.type} token to chain ${chain}`);
 
       switch (config.type) {
         case TokenType.collateral: {
```
antigremlin, line 59: Here, the typing introduced above makes it easier to see that we are using a token type (`config.token`) instead of a token denom in this call.

typescript/sdk/src/token/altVMDeploy.ts
```diff
@@ -66,6 +68,10 @@ export class CosmosNativeDeployer {
           result[chain] = await this.deploySyntheticToken(
             chain,
             config.mailbox,
+            config.name,
```
antigremlin, line 71: More typing errors: `config.name` is optional, and `config.description` doesn't exist on synthetic token types.

typescript/sdk/src/token/altVMDeploy.ts
```diff
@@ -79,11 +85,9 @@ export class CosmosNativeDeployer {
...
         this.logger.info(`Set ISM for token`);
 
-        await this.signersMap[chain].setToken({
-          token_id: result[chain],
-          new_owner: '',
-          ism_id: config.interchainSecurityModule,
-          renounce_ownership: false,
+        await this.signersMap[chain].setTokenIsm({
+          tokenAddress: result[chain],
+          ismAddress: config.interchainSecurityModule,
```
antigremlin, line 90: Typing error: `config.interchainSecurityModule` is not always a string address.

antigremlin (commented):
Please see a potential issue with deploying synthetics and ISMs.

typescript/cosmos-sdk/src/clients/provider.ts
troykessler: This was a missing feature request in the Hyperlane Cosmos Module. It will be probably added in the future, so currently it is unfortunately not supported right now.

typescript/cosmos-sdk/src/tests/4_warp.e2e-test.ts
```diff
@@ -1,335 +1,370 @@
...
 import { createSigner } from './utils.js';
 
 describe('4. cosmos sdk warp e2e tests', async function () {
   this.timeout(100_000);
 
-  let signer: SigningHyperlaneModuleClient;
+  let signer: AltVM.ISigner<EncodeObject, DeliverTxResponse>;
 
   before(async () => {
     signer = await createSigner('alice');
```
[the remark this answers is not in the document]
troykessler: added the type now

typescript/sdk/src/token/altVMDeploy.ts
```diff
@@ -38,9 +44,7 @@ export class CosmosNativeDeployer {
 
     const configMapToDeploy = objFilter(
       resolvedConfigMap,
-      (chain: string, config: any): config is any =>
-        this.metadataManager.getProtocol(chain) === ProtocolType.CosmosNative &&
-        !config.foreignDeployment,
+      (_: string, config: any): config is any => !config.foreignDeployment,
```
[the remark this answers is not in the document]
antigremlin: Why would we filter it then? Overall, our logic should operate on existing and new artifacts. But in `AltVMDeployer` we are only operating on new routers (the existing ones are filtered before calling the deployer). So in the end, this code should only operate on new routers of _this_ protocol.

typescript/sdk/src/deploy/warp.ts
```diff
@@ -324,26 +319,23 @@ async function createWarpHook({
       const { deployedHook } = evmHookModule.serialize();
       return deployedHook;
     }
-    case ProtocolType.CosmosNative: {
-      rootLogger.info(
-        `No warp hooks for Cosmos Native chains, skipping deployment.`,
+    default:
+      rootLogger.warn(
+        `Skipping token hooks because they are not supported on protocol type ${protocolType}`,
```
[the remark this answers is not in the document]
troykessler: So Cosmos and Radix support for sure no token hooks. For Aleo I just asked it is still unclear. I would propose to leave it first like that and if Aleo should support this we can add a new interface getter like `doesSupportWarpHooks` or something similar, wdyt?

typescript/utils/src/altvm.ts
troykessler: For "this one" I meant that multiple renaming changes where requested which I all implemented except for "this one". I would also see this in the upcoming multi-vm design doc discussion about the final interface since this is no pressing issue or bug

antigremlin (approved):
Even GitHub UI cannot bear the weight of this PR 😓

xeno097 (approved):

PR #7172, opened 2025-10-08
chore: update Radix USDC route

mbreithecker:


paulbalaji (commented):
also missing registryrc update

PR #7209, opened 2025-10-15
fix: prevent CCTP v2 nonce manipulation

yorhodes:
### Description

> how can we prevent a malicious relayer from manipulating a Hyperlane message by reusing an already-used CCTP nonce? Since CctpISM.verify() does not check the integrity of the CCTP message when the CCTP nonce has already been used, the Hyperlane message ID embedded in the CCTP message can no longer be trusted. In this case, the Hyperlane message itself can be tampered with. Thus, while a recipient may be designed to trust CctpISM and process the message, it coul  ### Description

Fixes a critical vulnerability where malicious relayers could manipulate Hyperlane messages by reusing already-consumed CCTP nonces.

  ### Vulnerability

Previously, `CctpISM.verify()` would skip CCTP verification if a nonce was already used but still return `true`. This allowed attackers to:
  1. Wait for a CCTP nonce to be consumed
  2. Call `Mailbox.process()` with an arbitrary message + the already-used CCTP attestation
  3. Have the ISM incorrectly verify the malicious message without checking its integrity

  ### Solution

  **Always process CCTP messages**: Remove nonce checks and always call `messageTransmitter.receiveMessage()` to validate attestations.
  **Authorized message ID pattern**: Inherit from `AbstractMessageIdAuthorizedIsm` to prevent direct `verify()` calls that would waste nonces:
  - Circle's MessageTransmitter calls back into `_receiveCircleMessage()`
  - This registers the message ID via `preVerifyMessage()`
  - Only pre-verified messages can pass `verify()`

  ### Changes

  - **TokenBridgeCctpBase.sol**: Add `AbstractMessageIdAuthorizedIsm`, bidirectional domain mapping, implement
  `_receiveCircleMessage()` callback
  - **TokenBridgeCctpV1/V2.sol**: Implement Circle's message handler callbacks
  - **Tests**: Updated mocks and added validation for authorization checks

  ### Backward Compatibility

  Breaking change - requires redeployment of CCTP ISM v1 and v2.

## Summary by CodeRabbit

* **New Features**
  * ISM-based authorization added for message verification.
  * Bidirectional domain mapping between Circle and Hyperlane exposed.

* **Improvements**
  * Verification flow simplified with idempotency and short-circuit checks.
  * Relay and burn/deposit flows streamlined; message construction deferred to a single post-bridge step.
  * Mock relayers/handlers extended for V1/V2 message formats.

* **Bug Fixes**
  * Removed fragile nonce/source gating to prevent false rejects.

* **Documentation**
  * Added sequence diagrams illustrating cross-chain token flows.

* **Tests**
  * Expanded V1/V2 coverage and adjusted expectations to new flows.

### DIFF
--- a/solidity/contracts/token/TokenBridgeCctpBase.sol
+++ b/solidity/contracts/token/TokenBridgeCctpBase.sol
@@ -36,11 +36,26 @@ abstract contract TokenBridgeCctpBaseStorage is TokenRouter {
     MovableCollateralRouterStorage private __MOVABLE_COLLATERAL_GAP;
 }
 
-// see ./CCTP.md for sequence diagrams of the destination chain control flow
-abstract contract TokenBridgeCctpBase is
+struct Domain {
+    uint32 hyperlane;
+    uint32 circle;
+}
+
+// need intermediate contract to insert slots between TokenBridgeCctpBase and AbstractMessageIdAuthorizedIsm
+abstract contract TokenBridgeCctpIntermediateStorage is
     TokenBridgeCctpBaseStorage,
     AbstractCcipReadIsm,
-    AbstractPostDispatchHook,
+    AbstractPostDispatchHook
+{
+    /// @notice Hyperlane domain => Domain struct.
+    /// We use a struct to avoid ambiguity with domain 0 being unknown.
+    mapping(uint32 hypDomain => Domain circleDomain)
+        internal _hyperlaneDomainMap;
+}
+
+// see ./CCTP.md for sequence diagrams of the destination chain control flow
+abstract contract TokenBridgeCctpBase is
+    TokenBridgeCctpIntermediateStorage,
     AbstractMessageIdAuthorizedIsm
 {
     using Message for bytes;
@@ -57,18 +72,10 @@ abstract contract TokenBridgeCctpBase is
     // @notice CCTP token messenger contract
     ITokenMessenger public immutable tokenMessenger;
 
-    struct Domain {
-        uint32 hyperlane;
-        uint32 circle;
-    }
-
-    /// @notice Hyperlane domain => Domain struct.
-    /// We use a struct to avoid ambiguity with domain 0 being unknown.
-    mapping(uint32 hypDomain => Domain domain) internal _hyperlaneDomainMap;
-
     /// @notice Circle domain => Domain struct.
     // We use a struct to avoid ambiguity with domain 0 being unknown.
-    mapping(uint32 circleDomain => Domain domain) internal _circleDomainMap;
+    mapping(uint32 circleDomain => Domain hyperlaneDomain)
+        internal _circleDomainMap;
 
     /**
      * @notice Emitted when the Hyperlane domain to Circle domain mapping is updated.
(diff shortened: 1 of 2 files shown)


solidity/contracts/token/TokenBridgeCctpBase.sol
```diff
@@ -60,9 +61,13 @@ abstract contract TokenBridgeCctpBase is
         uint32 circle;
     }
 
-    /// @notice Hyperlane domain => Circle domain.
+    /// @notice Hyperlane domain => Domain struct.
     /// We use a struct to avoid ambiguity with domain 0 being unknown.
-    mapping(uint32 hypDomain => Domain circleDomain) internal _domainMap;
+    mapping(uint32 hypDomain => Domain domain) internal _hyperlaneDomainMap;
```
nambrot: can we not name the value `circleDomain`?

solidity/contracts/token/TokenBridgeCctpV1.sol
```diff
@@ -100,68 +74,48 @@ contract TokenBridgeCctpV1 is TokenBridgeCctpBase, IMessageHandler {
...
+        uint32 sourceDomain,
+        bytes32 sender,
+        bytes calldata body
+    ) external override returns (bool) {
+        _authenticateCircleSender(sourceDomain, sender);
+        preVerifyMessage(_messageId(body), 0);
         return true;
     }
 
+    function _messageId(bytes calldata body) internal pure returns (bytes32) {
```
nambrot: This seems different of how we typically access destructured info in bodies via a library. I'd say either do that, or use a more descriptive function name

solidity/contracts/token/TokenBridgeCctpV2.sol
```diff
@@ -145,37 +120,39 @@ contract TokenBridgeCctpV2 is TokenBridgeCctpBase, IMessageHandlerV2 {
...
-        bytes calldata /*messageBody*/
-    ) external pure override returns (bool) {
+        bytes calldata messageBody
+    ) external override returns (bool) {
+        _authenticateCircleSender(sourceDomain, sender);
+        preVerifyMessage(_messageId(messageBody), 0);
         return true;
     }
 
+    function _messageId(bytes calldata body) internal pure returns (bytes32) {
```
nambrot: ditto

solidity/contracts/token/TokenBridgeCctpBase.sol
```diff
@@ -259,19 +275,22 @@ abstract contract TokenBridgeCctpBase is
...
-        if (messageTransmitter.usedNonces(nonce) == 0) {
-            require(
-                messageTransmitter.receiveMessage(
-                    cctpMessageBytes,
-                    attestation
-                ),
-                "Failed to receive message"
-            );
-        }
+        return messageTransmitter.receiveMessage(cctpMessageBytes, attestation);
```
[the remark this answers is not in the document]
yorhodes: The validations in `_validateTokenMessage` and `_validateHookMessage` do ensure the CCTP message corresponds to the specific Hyperlane message being verified:

  - For token messages: We validate sender, amount, and recipient all match
  - For hook messages: We validate the message ID is embedded in the CCTP message body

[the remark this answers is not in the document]
nambrot: Gotcha. Idk that I love the implicit reliance on that logic. Every other "authrized" ism simply checks for the isVerified slot, and I would feel better if we check for that here as well

PR #7212, opened 2025-10-15
feat: report reorg if checkpoint index decreased but block height stayed the same or increased

kamiyaa:
### Description

 - report reorg if checkpoint index decreased but block height stayed the same or increased

### Related issues

- fixes https://linear.app/hyperlane-xyz/issue/ENG-2223/validator-fire-alert-if-latest-checkpoint-is-lower-or-different-than

## Summary by CodeRabbit

* **New Features**
  * Persist latest checkpoint index and block height for improved restart reliability and state recovery.
  * Reorg events now include both local and canonical checkpoint indices for clearer diagnostics.

* **Bug Fixes**
  * Centralized and strengthened reorg detection/handling with consistent state updates and clearer reporting.
  * Added checkpoint verification and warnings on index/height inconsistencies.

* **Tests**
  * Updated mocks and tests to cover storing/retrieving latest checkpoint info and new reorg fields.

### DIFF
--- a/rust/main/config/mainnet_config.json
+++ b/rust/main/config/mainnet_config.json
@@ -6007,7 +6007,12 @@
       "index": {
         "from": 860343
       },
-      "gnosisSafeTransactionServiceUrl": "https://safe-transaction-ink.safe.global/"
+      "gnosisSafeTransactionServiceUrl": "https://safe-transaction-ink.safe.global/",
+      "transactionOverrides": {
+        "minGasPrice": 1,
+        "minFeePerGas": 1,
+        "minPriorityFeePerGas": 1
+      }
     },
     "soneium": {
       "blockExplorers": [
@@ -6928,7 +6933,12 @@
       "index": {
         "from": 42656210
       },
-      "gnosisSafeTransactionServiceUrl": "https://safe-transaction-ronin.safe.onchainden.com"
+      "gnosisSafeTransactionServiceUrl": "https://safe-transaction-ronin.safe.onchainden.com",
+      "transactionOverrides": {
+        "minGasPrice": 20000000000,
+        "minFeePerGas": 20000000000,
+        "minPriorityFeePerGas": 20000000000
+      }
     },
     "sophon": {
       "blockExplorers": [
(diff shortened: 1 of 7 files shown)


rust/main/agents/validator/src/submit.rs
```diff
@@ -127,6 +138,61 @@ impl ValidatorSubmitter {
             })
             .await;
 
+            if latest_checkpoint.index < latest_seen_checkpoint_index {
+                if let Some(block_height) = latest_checkpoint.block_height {
+                    if block_height >= latest_seen_checkpoint_block_height {
```
ameten: Since we can detect reorg only if we have block height available, we can rearrange this ifs so that we can combine two of them into single if condition, but making `if let` the most outer one.

rust/main/agents/validator/src/submit.rs
```diff
@@ -127,6 +138,61 @@ impl ValidatorSubmitter {
...
+                            "Latest checkpoint index is lower than previously seen, but has a block height equal or greater.");
+
+                        let checkpoint = self.checkpoint(&tree);
+                        Self::panic_with_reorg(
+                            &self.reorg_reporter,
+                            &self.reorg_period,
+                            &self.checkpoint_syncer,
+                            tree.root(),
+                            &latest_checkpoint,
+                            &checkpoint,
```
ameten: Since source code from `panic_with_reorg` was used only after the latest checkpoint was ingested into the tree, if there won't not be any reorg, `latest_checkpoint` and `checkpoint` would be the same, and we are panicking since they are different.

Now we calculating `checkpoint = self.checkpoint(&tree)` while we have not ingested all the checkpoints up to the `latest_checkpoint`. Therefore, we cannot really report it in this way. Probably, we need to adopt `panic_with_reorg` to support this new feature.

rust/main/agents/validator/src/submit.rs
```diff
@@ -127,6 +138,61 @@ impl ValidatorSubmitter {
...
+                    }
+                    latest_seen_checkpoint_block_height = block_height;
+                    if let Err(err) = self
+                        .db
+                        .store_latest_checkpoint_block_height(latest_seen_checkpoint_block_height)
+                    {
+                        tracing::error!(?err, "Failed to store latest_checkpoint_block_height");
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
kamiyaa: moved update to the end of the loop, after reorg has been checked

PR #7219, opened 2025-10-17
feat(infra): update pulsechain usdc rebalancer config 

xeno097:
### Description

Updates the pulsechain usdc rebalancer to concentrate collateral on base and ethereum

### Drive-by changes

- 

### Related issues

- 

### Backward compatibility

- yes

### Testing

- Local testing with monitor only mode

## Summary by CodeRabbit

* **Chores**
  * Updated collateral allocation weights in the USDC rebalancing strategy to change cross-chain distribution.
  * Updated deployed image versions for the rebalancer and route-monitor components.

### DIFF
--- a/typescript/infra/src/warp/helm.ts
+++ b/typescript/infra/src/warp/helm.ts
@@ -96,7 +96,7 @@ export class WarpRouteMonitorHelmManager extends HelmManager {
     return {
       image: {
         repository: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
-        tag: '4ea3698-20251015-173216',
+        tag: '15e2d5a-20251017-141153',
       },
       warpRouteId: this.warpRouteId,
       fullnameOverride: this.helmReleaseName,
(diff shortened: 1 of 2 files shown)


typescript/infra/src/warp/helm.ts
```diff
@@ -96,7 +96,7 @@ export class WarpRouteMonitorHelmManager extends HelmManager {
     return {
       image: {
         repository: 'gcr.io/abacus-labs-dev/hyperlane-monorepo',
-        tag: '4ea3698-20251015-173216',
+        tag: '15e2d5a-20251017-141153',
```
Mo-Hussain: This change is not needed. The rebalancer itself exposes the metrics that Prometheus scrapes

Mo-Hussain (approved):

typescript/infra/src/warp/helm.ts
xeno097: yup just updated it to keep it in "sync" with the rebalancer tag but i only deployed the rebalancer

PR #7222, opened 2025-10-20
fix: change min gas fees for Ronin and Ink

ameten:
### Description

change min gas fees for Ronin and Ink

### Backward compatibility

Yes

### Testing

Manual

## Summary by CodeRabbit

* **New Features**
  * Added minimum gas price/fee overrides for Ronin, Ancient8 and Ink to enforce explicit fee parameters per chain.

* **Bug Fixes**
  * Transaction status classification updated so mempool presence now takes precedence over pending-inclusion when both are reported, changing some status outcomes.

* **Chores**
  * Added informational logging for cases where transaction resubmission is skipped due to unchanged gas pricing.


paulbalaji (commented):
how confident are you that these changes are required? is there any docs we can refer to that suggest these values?

PR #7235, opened 2025-10-21
feat: allow ism to be set to 0 address in warp apply

xeno097:
### Description

Fix a bug in warp apply that prevented an ISM to be set to the 0 address (default ISM)

### Drive-by changes

- Add logic to update pausable and offchain lokkup ISMs as it was missing

### Related issues

- 

### Backward compatibility

- YES

### Testing

- E2E
- unit

## Summary by CodeRabbit

* **Bug Fixes**
  * Warp route ISM updates now support resetting to the zero address, correctly apply pausable ISM changes, and reliably update routes when deployed ISMs differ.

* **Tests**
  * Added comprehensive E2E and unit tests covering pausable, offchain-lookup, ordering, and zero-address ISM update scenarios; includes a new test helper for offchain ISMs.

* **Validation**
  * Offchain lookup configuration now requires valid URL strings.

* **Chores**
  * Added an array-equality utility to the public utilities.

### DIFF
--- a/typescript/sdk/src/ism/EvmIsmModule.ts
+++ b/typescript/sdk/src/ism/EvmIsmModule.ts
@@ -11,12 +11,12 @@ import {
   Domain,
   EvmChainId,
   ProtocolType,
+  arrayEqual,
   assert,
   deepEquals,
   intersection,
   isZeroishAddress,
   rootLogger,
-  setEquality,
 } from '@hyperlane-xyz/utils';
 
 import { CCIPContractCache } from '../ccip/utils.js';
@@ -371,7 +371,7 @@ export class EvmIsmModule extends HyperlaneModule<
     current: OffchainLookupIsmConfig;
     target: OffchainLookupIsmConfig;
   }): AnnotatedEV5Transaction[] {
-    if (setEquality(new Set(target.urls), new Set(current.urls))) {
+    if (arrayEqual(target.urls, current.urls)) {
       return [];
     }
 
--- a/typescript/cli/src/tests/ethereum/commands/helpers.ts
+++ b/typescript/cli/src/tests/ethereum/commands/helpers.ts
@@ -3,10 +3,11 @@ import path from 'path';
 import { $ } from 'zx';
 
 import {
+  AbstractCcipReadIsm,
+  AbstractCcipReadIsm__factory,
   ERC20Test,
   ERC20Test__factory,
   ERC4626Test__factory,
-  TestCcipReadIsm,
   TestCcipReadIsm__factory,
   XERC20LockboxTest,
   XERC20LockboxTest__factory,
@@ -272,7 +273,7 @@ export async function deployTestOffchainLookupISM(
   privateKey: string,
   chain: string,
   urls: string[] = [],
-): Promise<TestCcipReadIsm> {
+): Promise<AbstractCcipReadIsm> {
   const { multiProvider } = await getContext({
     registryUris: [REGISTRY_PATH],
     key: privateKey,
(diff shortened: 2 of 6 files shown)


typescript/cli/src/tests/ethereum/warp/warp-apply-1.e2e-test.ts
```diff
@@ -206,4 +212,120 @@ describe('hyperlane warp apply owner update tests', async function () {
       normalizeConfig(warpDeployConfig[CHAIN_NAME_2].hook),
     );
   });
+
+  describe('ism updates', () => {
```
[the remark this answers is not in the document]
xeno097: ah yeah that makes sense

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -99,85 +108,129 @@ export class EvmIsmModule extends HyperlaneModule<
...
     targetConfig = IsmConfigSchema.parse(targetConfig);
 
-    // Do not support updating to a custom ISM address
-    if (typeof targetConfig === 'string') {
-      throw new Error(
-        'Invalid targetConfig: Updating to a custom ISM address is not supported. Please provide a valid ISM configuration.',
-      );
+    // Nothing to do if its the default ism
+    if (typeof targetConfig === 'string' && isZeroishAddress(targetConfig)) {
+      return [];
```
antigremlin: Does it mean we now allow changing the IGP to a plain address? This caused an error before

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -285,6 +338,57 @@ export class EvmIsmModule extends HyperlaneModule<
...
+  }
+
+  protected updateOffchainLookupIsm({
+    current,
+    target,
+  }: {
+    current: OffchainLookupIsmConfig;
+    target: OffchainLookupIsmConfig;
+  }): AnnotatedEV5Transaction[] {
+    if (setEquality(new Set(target.urls), new Set(current.urls))) {
```
antigremlin: Are we sure we don't care about the order of the URLs?

typescript/sdk/src/ism/EvmIsmModule.ts
```diff
@@ -285,6 +338,57 @@ export class EvmIsmModule extends HyperlaneModule<
...
+      return [];
+    }
+
+    return [
+      {
+        annotation: `Setting urls to ${target.type} ISM on chain "${this.chain}" and address "${this.args.addresses.deployedIsm}"`,
+        chainId: this.multiProvider.getEvmChainId(this.chain),
+        to: this.args.addresses.deployedIsm,
+        // The contract code just replaces the existing array with the new one
+        data: AbstractCcipReadIsm__factory.createInterface().encodeFunctionData(
```
antigremlin: Do we maybe want to verify the URL validity before writing to on-chain state? In the Zod schema maybe?

typescript/sdk/src/ism/EvmIsmModule.ts
xeno097: The IGP is a hook, here we are working with ISMs but yeah we are going to allow setting the ISM to a custom address as there is no real limitation to why this shouldn't be allowed.  Initially we did not allow setting to a custom address in hooks too but we changed it so this would bring feature parity to ISMs too

typescript/sdk/src/ism/EvmIsmModule.ts
xeno097: sounds like a good idea

PR #7247, opened 2025-10-23
feat: fallback to single tx submission if batching fails

kamiyaa:
### Description

 - if batching fails, we should revert to single tx submission

### Related issues

- Fixes https://linear.app/hyperlane-xyz/issue/ENG-2422/lander-fallback-to-single-payload-txs-if-multicall-is-not-deployed

### Testing

- unittest

## Summary by CodeRabbit

* **New Features**
  * Per-domain batched-transaction metrics added and exposed on the Ethereum adapter (with an accessor).

* **Bug Fixes**
  * Batched transactions now reliably fall back to single transactions on failure.
  * Clearer chain communication errors including an explicit "contract not found" case and improved failure reporting.

* **Tests**
  * Reorganized and expanded tests covering batching, resubmission logic, and VM-specific metrics.

* **Chores**
  * Test utilities made publicly accessible for reuse.


rust/main/lander/src/adapter/chains/ethereum/nonce/state/core.rs
```diff
@@ -30,4 +30,8 @@ impl NonceManagerState {
             metrics,
         }
     }
+
+    pub fn metrics(&self) -> &EthereumAdapterMetrics {
+        &self.metrics
+    }
```
ameten: Since `EthereumAdapter` will use `EthereumAdapterMetrics` directly, let's add it as a field to `EthereumAdapter` instead of exposing via `NonceManagerState`.

rust/main/lander/src/adapter/chains/ethereum/adapter.rs
ameten: We need to change this doc comment as well.

rust/main/lander/src/adapter/chains/ethereum/adapter/tests/tests_build.rs
```diff
@@ -0,0 +1,125 @@
...
+    payload, MockRadixProvider, MAILBOX_ADDRESS, TEST_PRIVATE_KEY,
+};
+use crate::adapter::{AdaptsChain, TxBuildingResult};
+use crate::payload::PayloadDetails;
+use crate::tests::test_utils::tmp_dbs;
+use crate::transaction::VmSpecificTxData;
+use crate::FullPayload;
+
+#[tokio::test]
+async fn test_build_transactions_batch_contract_missing() {
```
ameten: Let's have a test for success as well.

rust/main/chains/hyperlane-ethereum/src/contracts/multicall.rs
```diff
@@ -56,16 +58,19 @@ pub async fn build_multicall<M: Middleware + 'static>(
...
     let multicall =
         match Multicall::new(provider.clone(), Some(multicall_contract_address.into())).await {
             Ok(multicall) => multicall.version(MulticallVersion::Multicall3),
             Err(err) => {
-                return Err(eyre::eyre!(
-                    "Unable to build multicall contract: {}",
-                    err.to_string()
-                ))
+                tracing::error!(?err, "Unable to build multicall contract");
+                let err = ChainCommunicationError::CustomError(format!(
```
ameten: Could you please add test for this error?

PR #7250, opened 2025-10-23
feat: migrate provider-level alt-vm code to provider-sdk

antigremlin:
## Description

This PR migrates AltVM provider types and interfaces from `@hyperlane-xyz/utils` to a new dedicated `@hyperlane-xyz/provider-sdk` package. In preparation to further moving AltVM modules to `deploy-kit`, it also decouples AltVM modules from `HyperlaneModule` and provider-specific transaction types.

Key changes:

- `ChainLookup` interface and `altVmChainLookup()` adapter function to abstract chain metadata access from `MultiProvider`
- AltVM modules (`AltVMCoreModule`, `AltVMWarpModule`, `AltVMIsmModule`, `AltVMHookModule`) use the new `HypModule` interface instead of extending `HyperlaneModule`
- AltVM modules now use generic `AnnotatedTx` and `TxReceipt` types instead of protocol-specific transaction types
- Added subpath exports (`./altvm`, `./module`, `./protocol`) to `provider-sdk` package.json as an experiment

## Related Issues

Fixes ENG-2396

## Backward compatibility

No - this is a breaking change. Consumers of AltVM types from `@hyperlane-xyz/utils` will need to update their imports to use `@hyperlane-xyz/provider-sdk` instead. AltVM modules no longer inherit `HyperlaneModule` but that shouldn't be noticeable because we don't rely on parent class typing anywhere.

### DIFF
--- a/typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
+++ b/typescript/cosmos-sdk/src/tests/1_interchain_security.e2e-test.ts
@@ -3,11 +3,8 @@ import { DeliverTxResponse } from '@cosmjs/stargate';
 import { expect } from 'chai';
 import { step } from 'mocha-steps';
 
-import {
-  AltVM,
-  bytes32ToAddress,
-  isValidAddressEvm,
-} from '@hyperlane-xyz/utils';
+import { AltVM } from '@hyperlane-xyz/provider-sdk';
+import { bytes32ToAddress, isValidAddressEvm } from '@hyperlane-xyz/utils';
 
 import { createSigner } from './utils.js';
 
--- a/typescript/cosmos-sdk/src/tests/2_core.e2e-test.ts
+++ b/typescript/cosmos-sdk/src/tests/2_core.e2e-test.ts
@@ -3,11 +3,8 @@ import { DeliverTxResponse } from '@cosmjs/stargate';
 import { expect } from 'chai';
 import { step } from 'mocha-steps';
 
-import {
-  AltVM,
-  bytes32ToAddress,
-  isValidAddressEvm,
-} from '@hyperlane-xyz/utils';
+import { AltVM } from '@hyperlane-xyz/provider-sdk';
+import { bytes32ToAddress, isValidAddressEvm } from '@hyperlane-xyz/utils';
 
 import { createSigner } from './utils.js';
 
--- a/typescript/cosmos-sdk/src/tests/3_post_dispatch.e2e-test.ts
+++ b/typescript/cosmos-sdk/src/tests/3_post_dispatch.e2e-test.ts
@@ -3,11 +3,8 @@ import { DeliverTxResponse } from '@cosmjs/stargate';
 import { expect } from 'chai';
 import { step } from 'mocha-steps';
 
-import {
-  AltVM,
-  bytes32ToAddress,
-  isValidAddressEvm,
-} from '@hyperlane-xyz/utils';
+import { AltVM } from '@hyperlane-xyz/provider-sdk';
+import { bytes32ToAddress, isValidAddressEvm } from '@hyperlane-xyz/utils';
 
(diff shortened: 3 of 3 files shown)


troykessler:
> This is a draft migration of alt-VM code. It looks extreme but is 90% import statements

What are the other 10%, can you mark the most important files to review?

typescript/cli/src/avs/check.ts
```diff
@@ -6,13 +6,14 @@ import {
...
 } from '@hyperlane-xyz/core';
+import { ProtocolType } from '@hyperlane-xyz/provider-sdk';
 import {
   ChainMap,
   ChainName,
   MultiProvider,
   isValidValidatorStorageLocation,
 } from '@hyperlane-xyz/sdk';
-import { Address, ProtocolType, isObjEmpty } from '@hyperlane-xyz/utils';
+import { Address, isObjEmpty } from '@hyperlane-xyz/utils';
```
xeno097: i wonder if we should move also the "custom" types (like Address in this case) to the provider sdk and leave the utils package as a utility function only package

antigremlin: On one hand, yes. But I can see how moving ProtocolType is making it impossible to see other changes on the same PR. So if we do more more types, it should be a dedicated change. Do you have any other from utils you wish to move?

typescript/sdk/src/core/AltVMCoreModule.ts
```diff
@@ -39,12 +32,11 @@ export class AltVMCoreModule<PT extends ProtocolType> extends HyperlaneModule<
...
   protected coreReader: AltVMCoreReader;
 
+  // Cached chain name
   public readonly chainName: ChainName;
-  public readonly chainId: ChainId;
-  public readonly domainId: Domain;
 
   constructor(
-    protected readonly metadataManager: ChainMetadataManager,
+    protected readonly chainLookup: ChainLookup,
```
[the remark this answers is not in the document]
antigremlin: So that we can move `AltVmXxxModule`s to another package without also moving `ChainMetadataManager` -- which is difficult as it drags more dependencies and also has EVM-only functions.

typescript/sdk/src/token/AltVMWarpModule.ts
```diff
@@ -391,38 +382,30 @@ export class AltVMWarpModule<PT extends ProtocolType> extends HyperlaneModule<
...
    * @param config - The configuration for the token router.
-   * @param multiProvider - The multi-provider instance to use.
+   * @param chainLookup - Chain metadata lookup functions
    * @param signer - The AltVM signing client
    * @returns A new instance of the AltVMWarpModule.
    */
-  static async create<PT extends ProtocolType>(params: {
-    chain: ChainNameOrId;
+  static async create(params: {
+    chain: string;
```
paulbalaji: i suppose this is part of detangling the types

typescript/provider-sdk/package.json
```diff
@@ -7,7 +7,10 @@
     "dist"
   ],
   "exports": {
-    ".": "./dist/index.js"
+    ".": "./dist/index.js",
+    "./altvm": "./dist/altvm.js",
+    "./module": "./dist/module.js",
+    "./protocol": "./dist/protocol.js"
```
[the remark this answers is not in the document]
antigremlin: I wanted to test this out. Benefits to import speed or bundle size are theoretical and unmeasured. But there's a structural benefit here:
- we don't have to thread every export through `index.ts`
- it's more clear where things come from at import sites

As always, you pay for this benefit by thinking harder about code organisation. Otherwise we'll just list every single file in this section.

typescript/sdk/src/token/AltVMWarpModule.ts
paulbalaji: i think we should actually just use chainName only in places instead of conflating name/id, but again something not for this PR

PR #7256, opened 2025-10-24
feat: stricter tsconfig [ENG-2414]

xeno097:
### Description

A new `tsconfig` file which has the goal of ensuring stricter type checking when developing new code for the new restructured packages. Compared to the existing config it adds the following properties:
  - strictNullChecks: true - Treats null and undefined as distinct types
  - noUnusedParameters: true - Flags unused function parameters
  - strictBindCallApply: true - Checks bind, call, and apply methods
  - allowUnreachableCode: false - Disallows unreachable code
  - strictPropertyInitialization: true - Ensures class properties are initialized
  - strictFunctionTypes: true - Enables strict function type checking, which disallows using functions that have similar signatures
  - noUncheckedIndexedAccess: true - Adds undefined to indexed access types to avoid using potentially undefined values from arrays and objects 
  - noPropertyAccessFromIndexSignature: true - Requires bracket notation for index signature properties
  - noImplicitThis: true - Flags implicit 'any' type on 'this'
  - noImplicitOverride: true - Requires override keyword for overridden methods from parent classes
  - isolatedModules: true - Ensures each file can be safely transpiled individually
 
more info at https://www.typescriptlang.org/tsconfig

### Drive-by changes

-

### Backward compatibility

- Yes until the code starts using the new tsconfig

### Testing

- No

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript configuration with stricter type-checking rules, including enhanced validation for array and object property access and comprehensive strict mode settings.


typescript/tsconfig/new/tsconfig.json
```diff
@@ -0,0 +1,66 @@
...
+    "sourceMap": true,
+    "declaration": true,
+    "noEmitOnError": true,
+    "declarationMap": true,
+
+    // Type checking
+    "strict": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
+    "noUnusedParameters": true,
```
paulbalaji: ditto surprised we don't already conform to this

PR #7262, opened 2025-10-27
fix: do not donate on receive in `HypNative`

yorhodes:
### Description

> Recommend removing the donate() call in HypNative.receive(). When LP deposited via HypNative.deposit() is sent to another chain by the rebalancer and then returns, HypNative.receive() is invoked. Even though the LP is merely returning, donate() is executed and the share value can be incorrectly increased.

## Release Notes

* **Bug Fixes**
  * Removed automatic donation triggering when native tokens are received, enabling tokens to be accepted strictly for collateral rebalancing purposes.


yorhodes:
> Why did we have donate in the first place?

this exists to reward the LPs pro-rata

PR #7270, opened 2025-10-28
feat(cli): radix e2e tests base setup + core deploy tests

xeno097:
### Description

This PR introduces the necessary code for writing radix E2E test in the CLI

- Ports the logic for publishing packages from the Rust SDK as the radix toolkit package does not implement it
- Adds support for custom gateway urls in the radix SDK
- Adds logic for overriding the package address used by the Radix SDK to identify the hyperlane blueprints

### Drive-by changes

- Updates the `DerivedCoreConfig` type to use the correct type for the default ism (`DerivedIsmConfig`)

### Related issues

- 

### Backward compatibility

Yes

### Testing

- E2E
- Manual

## Summary by CodeRabbit

* **New Features**
  * Radix support: local node orchestration, package publishing flow, provider/signer now accept metadata (gateway URLs & package address), and manifest-building for publishing.
  * Default ISM configuration now has a concrete typed shape.

* **Bug Fixes**
  * Ownership transfer for deployed routing ISMs corrected.

* **Tests**
  * New Radix end-to-end tests, test configs, docker setup, runner script, and CI workflow integration.


typescript/cli/src/tests/radix/e2e-test.setup.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  // Write back to registry file so CLI can read the package address field injected
+  // when starting the node
+  const metadataPath =
+    TEST_CHAIN_METADATA_PATH_BY_PROTOCOL[ProtocolType.Radix].CHAIN_NAME_1;
+  const updatedMetadata =
+    TEST_CHAIN_METADATA_BY_PROTOCOL[ProtocolType.Radix].CHAIN_NAME_1;
+  writeYamlOrJson(metadataPath, updatedMetadata);
+});
```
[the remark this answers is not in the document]
xeno097: testcontainers automatically shuts down the compose containers so no clean up logic is needed

typescript/cli/src/tests/nodes.ts
```diff
@@ -45,3 +57,68 @@ export async function runCosmosNode({ rpcPort, restPort }: TestChainMetadata) {
 
   return container;
 }
+
+export async function runRadixNode(
```
paulbalaji: I wonder if this is something we eventually add into an altvm test interface

.github/workflows/test.yml
paulbalaji: you'll also want to update this:
https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/79a51debee96c8e82d1b02ee065350f37442b7f5/.github/workflows/test.yml#L430-L451

typescript/cli/src/tests/nodes.ts
xeno097: it would probably makes sense to, but we don't have a proper plan for multi vm testing yet. would be cool to have a single test package and run the same test suite against the different vm implementation which i think it is something that also Andrey was mentioning in his doc

troykessler: So with AltVM at least it is only necessary to test the AltVM modules once like the `AltVMWarpModule.test.ts`, but I have not written them for the other modules yet because with MultiVM it will change anyway. But then you only need to test the protocols like radix against their AltVM interface and nothing more so you also save a lot of e2e tests. But like you said I considered it not worth it anymore to write out everything when it will change sooner or later

typescript/cli/src/tests/constants.ts
```diff
@@ -70,6 +77,35 @@ type ProtocolChainMap<
...
+  //   new PublicKey.Ed25519(
+  //     Uint8Array.from([
+  //       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+  //       0, 0, 0, 0, 0, 0, 0, 0, 0,
+  //     ]),
+  //   ),
+  //   NetworkId.LocalNet,
+  // );
+  [ProtocolType.Radix]:
+    'account_loc1294g56ga4ckdzhksx6vnrns2jj0v47ju87flsyscxdjxu9wrkjp5vt',
```
yjamin: nit: there is also something like a null address for radix in case you want to use it here

typescript/cli/src/tests/radix/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
+# Modified from rust/main/utils/run-locally/src/radix/docker-compose.yml
```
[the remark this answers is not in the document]
xeno097: Yeah i thought the same, but i'm planing to do more changes in a follow up PR to see if we can run multiple nodes at the same time for warp tests and those changes maybe might break the rust tests so i didn't want to touch the same file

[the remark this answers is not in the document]
yjamin: Ah I see, I wonder what is the intention behind this? Because one can deploy multiple Hyperlane instances on one chain, no?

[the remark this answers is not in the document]
xeno097: jsut try to have isolated envs, but if it is not possible then i'll fall back to using one node

PR #7274, opened 2025-10-30
chore: merge Q3 audit branch

yorhodes:
### Description

See `audit-q3-2025` for contents

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8137,10 +8137,10 @@ __metadata:
   dependencies:
     "@eth-optimism/sdk": "npm:^3.3.3"
     "@google-cloud/pino-logging-gcp-config": "npm:^1.0.6"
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/adapters/EvmTokenAdapter.ts
```diff
@@ -833,6 +911,24 @@ export class EvmHypNativeAdapter
     );
   }
 
+  async populateTransferRemoteTx({
```
Xaroz: I see that the `EvmHypNativeAdapter` override for `quoteTransferRemoteGas` was removed, does this mean the contract now returns the quote as separate instead of a combined quote (igp + fee + amount) for native tokens?

paulbalaji (approved):

PR #7276, opened 2025-10-30
feat: migrate altvm modules

antigremlin:
This PR should automatically repointed to `feat/multi-vm` after merging #7250. As we normally do squash-merges, this PR should be rebased onto `feat/multi-vm` after that in order to remove duplicate commits from its history (for humans' sake as technically it'd still work).

## Description

This PR migrates AltVM Core, Hook, and ISM modules from the `sdk` package to the `deploy-sdk` package to prepare for upcoming changes to provider APIs:

- `AltVMCoreModule`, `AltVMHookModule`, and `AltVMIsmModule` along with their corresponding readers moved to `@hyperlane-xyz/deploy-sdk`
- Introduced shared type definitions (`CoreConfig`, `HookConfig`, `IsmConfig`) in `@hyperlane-xyz/provider-sdk` to decouple from Zod types in the SDK

## Drive-by Changes

- Converted `TokenType`, `HookType`, and `IsmType` enums to const object + union type pattern for easier type compatibility between packages
- Moved config normalization utilities (`normalizeConfig`, `sortArraysInConfig`) from SDK to `@hyperlane-xyz/utils` for broader reuse

## Related Issues

Fixes ENG-2396

## Backward Compatibility

No. This is a breaking change for consumers of `AltVMCoreModule`, `AltVMHookModule`, and `AltVMIsmModule`. These classes now need to be imported from `@hyperlane-xyz/deploy-sdk` instead of `@hyperlane-xyz/sdk`. The CLI has been updated accordingly.

### DIFF
--- a/typescript/deploy-sdk/src/AltVMCoreModule.ts
+++ b/typescript/deploy-sdk/src/AltVMCoreModule.ts
@@ -163,17 +163,7 @@ export class AltVMCoreModule
 
     const addresses: DeployedCoreAddresses = {
       mailbox: mailbox.mailboxAddress,
-      staticMerkleRootMultisigIsmFactory: '',
-      proxyAdmin: '',
-      staticMerkleRootWeightedMultisigIsmFactory: '',
-      staticAggregationHookFactory: '',
-      staticAggregationIsmFactory: '',
-      staticMessageIdMultisigIsmFactory: '',
-      staticMessageIdWeightedMultisigIsmFactory: '',
       validatorAnnounce: validatorAnnounce.validatorAnnounceId,
-      testRecipient: '',
-      interchainAccountRouter: '',
-      domainRoutingIsmFactory: '',
     };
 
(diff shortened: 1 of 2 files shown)


typescript/deploy-sdk/src/AltVMCoreModule.ts
```diff
@@ -195,11 +185,11 @@ export class AltVMCoreModule
 
     if (config.defaultHook && typeof config.defaultHook !== 'string') {
       switch (config.defaultHook.type) {
-        case HookType.INTERCHAIN_GAS_PAYMASTER: {
+        case 'interchainGasPaymaster': {
```
ltyu: how come we are using strings here? can't we use HookType defined below in `typescript/sdk/src/hook/types.ts`?

ltyu: i suppose we can potentially do that later

ltyu (approved):

PR #7289, opened 2025-10-31
fix: Scraper: Do nothing when block with hash is already in database

ameten:
### Description

At the moment, Scraper will fail to persist an event (like message dispatch or gas payment) if it attempts to insert a block with the same hash. It may happen due to race between message dispatch and gas payment indexing.

### Backward compatibility

Yes

### Testing

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened conflict handling for block inserts so conflicting records are skipped more generally, reducing duplicate-insert failures.

* **Migrations**
  * Block hash now enforced as unique, simplifying index usage and uniqueness guarantees.

* **Tests**
  * Added a real-Postgres integration test exercising block storage, duplicate handling, batch inserts, and cleanup.

* **Chores**
  * Added workspace test tooling to support the new integration tests.

### DIFF
--- a/rust/main/agents/scraper/src/db/block/tests.rs
+++ b/rust/main/agents/scraper/src/db/block/tests.rs
@@ -24,7 +24,6 @@ fn make_block(height: u64, hash: H256, timestamp: u64) -> BlockInfo {
 /// 2. Duplicate blocks (same hash) are handled with DO NOTHING
 /// 3. Duplicate blocks (same domain+height) are handled with DO NOTHING
 /// 4. Multiple blocks can be inserted in one call
-//#[ignore]
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


yjamin (approved):

rust/main/agents/scraper/src/db/block.rs
```diff
@@ -84,7 +84,7 @@ impl ScraperDb {
         let models = blocks
             .map(|info| block::ActiveModel {
                 id: NotSet,
-                hash: Set(address_to_bytes(&info.hash)),
+                hash: Set(h256_to_bytes(&info.hash)),
```
ameten: Fix of an old bug. It was exposed with predefined hashes with a lot of zeros at the front.

PR #7295, opened 2025-10-31
feat(ci): adjust yarn caching

paulbalaji:
### Description

feat(ci): adjust yarn caching

- install-cli action can optionally not use cache (used for cross-platform matrix in release workflow)
- add reusable yarn-cache action to have a single place where the yarn caching dirs are defined
- yarn-build-with-cache reuses yarn-cache action

### Drive-by changes

- default github cache for all yarn-build-with-cache, optional buildjet usage if you want to use yarn-cache directly

### Testing

before and after:
imageimage

## Summary by CodeRabbit

* **Chores**
  * Centralized Yarn dependency caching into a single local cache action used across CI workflows.
  * Replaced per-job external cache usage with unified caching and simplified cache logic.
  * Standardized runner environments and Node setup resolution in release and test pipelines.

* **New**
  * Added an optional cache-bypass flag to run fresh installs/builds when needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - pb/ci-yarn-testing
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


xeno097 (approved):
remember to remove this branch from the allow list for the release job

PR #7299, opened 2025-10-31
chore(ci): further split up warp-check/warp-deploy e2e tests

paulbalaji:
### Description

chore(ci): further split up warp-check/warp-deploy e2e tests


xeno097 (approved):
Only minor thing is that i don't see the old warp-deploy test suite being deleted in the changes

paulbalaji:
> Only minor thing is that i don't see the old warp-deploy test suite being deleted in the changes

heh good spot, i thought it would've picked up the deploy->deploy-1 rename

PR #7304, opened 2025-11-03
feat: allow test action to run on multi vm feature branch

xeno097:
### Description

Adds the multi vm feature branch explicitly to the allow list for running test in ci when openinf a pr as the `*` selector does not work if the target branch name has a `/`

### Drive-by changes

- Removed cli-2.0 in test ci config

### Related issues

- 

### Backward compatibility

- yes


paulbalaji:
but not worth merging main into feat/multi-vm yet if we're gonna merge it all to main in a couple weeks anyway?

PR #7305, opened 2025-11-03
chore: allow forma->celestia relaying

paulbalaji:
### Description

chore: allow forma->celestia relaying

### Drive-by changes

driveby: disable cheesechain in agents


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -602,6 +602,11 @@ const gasPaymentEnforcement: GasPaymentEnforcement[] = [
         originDomain: getDomainId('stride'),
         destinationDomain: getDomainId('forma'),
       },
+      // Not a core chain
+      {
```
yjamin: Are Celestia and Stride the only connections for forma? In case we would want to add another IGP config

PR #7315, opened 2025-11-03
fix: Add catch when calling `fetchPackageVersion()`

ltyu:
### Description
This PR removes the `.catch()` at the callsite to handle unknown errors from fetching the contract version, and moves it to `fetchPackageVersion()`. This logic breaks because not all Errors thrown conform to `err.cause.code`, and also there was handling the second throw.

### Backward compatibility
Yes

### Testing
Manual

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve resilience when fetching contract/package versions: treat pre-package contracts as version "5.3.9" for compatibility and default to "0.0.0" on unknown errors to prevent uncaught failures.

* **Documentation**
  * Added a changeset documenting a minor release and the new fallback behavior for unknown contract version errors.

### DIFF
--- a/.changeset/shiny-gifts-double.md
+++ b/.changeset/shiny-gifts-double.md
@@ -2,4 +2,4 @@
 "@hyperlane-xyz/sdk": minor
 ---
 
-Update fetchTokenConfig() to return 0.0.0 when unknown error is thrown
+Update fetchPackageVersion() to return 0.0.0 when unknown error is thrown. This error is logged out and is no longer rethrown.


typescript/sdk/src/token/EvmERC20WarpRouteReader.ts
```diff
@@ -992,13 +992,14 @@ export class EvmERC20WarpRouteReader extends EvmRouterReader {
...
-      if (err.cause.code && err.cause.code === 'CALL_EXCEPTION') {
+      if (err.cause?.code && err.cause?.code === 'CALL_EXCEPTION') {
         // PACKAGE_VERSION was introduced in @hyperlane-xyz/core@5.4.0
         // See https://github.com/hyperlane-xyz/hyperlane-monorepo/releases/tag/%40hyperlane-xyz%2Fcore%405.4.0
         // The real version of a contract without this function is below 5.4.0
         return '5.3.9';
       } else {
-        throw err;
+        this.logger.error(`Error when fetching package version ${err}`);
+        return '0.0.0';
```
xeno097: Remind me, why are we returning version 0 instead of the version before the `PACKAGE_VERSION` method was introduced?

ltyu: we can also return '5.3.9'. i was just distinguishing between an expected error that returns '5.3.9' versus a non-expected one that returns '0.0.0'

xeno097 (approved):

PR #7322, opened 2025-11-05
feat(protocol-sdk): protocol api definition [ENG-2395]

xeno097:
### Description

This PR defines the Protocol API that VM specific implementations need to define in order to be be used in a VM agnostic environment

### Drive-by changes

- none

### Related issues

- Fixes #[ENG-2395](https://linear.app/hyperlane-xyz/issue/ENG-2395/adapt-protocol-api-to-hypmoduleartifact)

### Backward compatibility

- yes, no changes to the existing code have been implemented

### Testing

-


typescript/provider-sdk/src/registry.ts
```diff
@@ -0,0 +1,176 @@
...
+  listProtocols(): ProtocolType[] {
+    return Array.from(this.protocols.keys());
+  }
+
+  registerProtocol(
+    protocol: ProtocolType,
+    factory: () => IProtocolProviderFactory,
+  ): void {
+    assert(
+      this.hasProtocol(protocol),
```
ltyu:
```suggestion
      !this.hasProtocol(protocol),
```

xeno097: good catch thanks!

troykessler (approved):

typescript/provider-sdk/src/submitter.ts
```diff
@@ -0,0 +1,27 @@
+import { AnnotatedTx, TxReceipt } from './module.js';
+
+// By default each VM implementation should at least define a JSON rpc submitter
+export const TransactionSubmitterType = {
```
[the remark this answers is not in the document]
antigremlin: I guess I used to think this way, that having named constants is better than string literals because it guards against typos. But imagine if most the string constants are TypeScript literal types. In this case, the compiler guards against typos and incomplete switch coverage. And auto-complete works as well (I wonder if renaming works too).

Benefits?
- You get one less level of indirection.
- You see the real constant being used rather than its representation dictated by code conventions (that's when you have `fooBar` in the actual API but `FOO_BAR` in code because constants).

PR #7325, opened 2025-11-05
fix: Add log on value is null

ameten:
## Summary by CodeRabbit

* **Chores**
  * Improved RPC fallback diagnostics: added conditional debug logging to surface when a provider returns an empty/null response, including fallback attempts and provider context to aid troubleshooting.


rust/main/chains/hyperlane-ethereum/src/rpc_clients/fallback.rs
```diff
@@ -220,7 +221,22 @@ where
...
+                                    provider_host = provider_host.as_str(),
+                                    method,
+                                    ?v,
+                                    "fallback_request: value is null"
+                                );
+                            }
+                            Ok(serde_json::from_value(v)?)
+                        };
+                    }
                     RetryableErr(e) | RateLimitErr(e) => errors.push(e.into()),
```
yjamin: nit: any reason why to double wrap the statements? I would imagine this code to be the same:

```rust
IsOk(v) => {
      // Add log to identify content of v when no tx receipt is found
      if v.is_null() {
          tracing::debug!(
              fallback_count = idx,
              provider_index = priority.index,
              provider_host = provider_host.as_str(),
              method,
              ?v,
              "fallback_request: value is null"
          );
      }
    return Ok(serde_json::from_value(v)?);
}
```

PR #7376, opened 2025-11-13
feat: install libcurl for rust docker

kamiyaa:
### Description

Fix
```
./relayer: error while loading shared libraries: libcurl.so.4: cannot open shared object file: No such file or directory
```

## Summary by CodeRabbit

* **Chores**
  * Added an additional runtime system library (libcurl) to improve compatibility and system library support for the runtime environment.

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -70,7 +70,7 @@ COPY --from=builder /release/* .
 # Install runtime dependencies
 # remove /var/lib/apt/lists/* to clean up the package lists
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends openssl ca-certificates tini && \
+    apt-get install -y --no-install-recommends openssl ca-certificates tini libcurl4 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
     chmod 777 /app && \


yjamin:
I don't think we need this. We should investigate how it went missing in the image.

kamiyaa:
> I don't think we need this. We should investigate how it went missing in the image.

Update on this

`rust:1.86.0` and `rust:1.88.0` does come with libcurl preinstalled.

`ubuntu:22.04` does not come with libcurl preinstalled.

Looks like some of the new Aleo dependencies require curl as a dependency:
- https://crates.io/crates/snarkvm-parameters/4.3.1/dependencies
- https://github.com/ProvableHQ/snarkVM/blob/009494f6a24cad1669c4ca5596443603481de7e2/parameters/src/macros.rs#L73
- https://github.com/ProvableHQ/snarkVM/blob/009494f6a24cad1669c4ca5596443603481de7e2/parameters/src/errors/parameter.rs#L43
- https://github.com/ProvableHQ/snarkVM/blob/009494f6a24cad1669c4ca5596443603481de7e2/parameters/Cargo.toml#L96

So looks like we need to install libcurl moving forward

PR #7412, opened 2025-11-19
fix: make DeployedCoreAddresses match Zod schema

antigremlin:
### Description

Fix the mismatch between local types and Zod schema to restore some e2e tests.


xeno097 (approved):
The comeback no one wanted

PR #7429, opened 2025-11-21
feat: radix ism modules part 1 [ENG-2467]

xeno097:
### Description

 This PR implements the artifact modules and readers for the radix ISM types
- creates the `ismTypeFromRadixIsmType` util function to convert from internal ism types to sdk ism types
- moves the `calculateDomainRoutingDelta` from the deploy-sdk to the provider-sdk
- defines the `ReaderProvider` interface to make explicit when a reader is used in a read only environment vs a write/read one

### Drive-by changes

- Moves the logic for of the `getMultisigIsm`, `getRoutingIsm` and `getKeysFromKeyValueStore` into separate functions as the provider and signer classes are still used 

### Related issues

- Partially Fixes #[ENG-2467](https://linear.app/hyperlane-xyz/issue/ENG-2467/implement-ism-artifact-modules) [deploy logic is still missing]

### Backward compatibility

- Yes 

### Testing

- Manual

### DIFF
--- a/typescript/radix-sdk/src/ism/multisig-ism.ts
+++ b/typescript/radix-sdk/src/ism/multisig-ism.ts
@@ -39,9 +39,7 @@ export class RadixMultisigIsmReader implements HypReader<MultisigIsmModule> {
   async read(address: string): Promise<WithAddress<MultisigIsmConfig>> {
     const { threshold, validators, type } = await getMultisigIsmConfig(
       this.gateway,
-      {
-        ismAddress: address,
-      },
+      address,
     );
 
     const ismType = ismTypeFromRadixIsmType(type);
--- a/typescript/radix-sdk/src/ism/routing-ism.ts
+++ b/typescript/radix-sdk/src/ism/routing-ism.ts
@@ -40,7 +40,7 @@ export class RadixRoutingIsmReader implements HypReader<RoutingIsmModule> {
   async read(address: string): Promise<WithAddress<DomainRoutingIsmConfig>> {
     const { owner, routes } = await getDomainRoutingIsmConfig(
       this.radixGateway,
-      { ism: address },
+      address,
     );
 
     const ismReader = this.readerProvider.connectReader(this.provider);
--- a/typescript/radix-sdk/src/core/query.ts
+++ b/typescript/radix-sdk/src/core/query.ts
@@ -145,7 +145,7 @@ export class RadixCoreQuery {
     threshold: number;
     validators: string[];
   }> {
-    return getMultisigIsmConfig(this.gateway, { ismAddress: ism });
+    return getMultisigIsmConfig(this.gateway, ism);
   }
 
   public async getRoutingIsm({ ism }: { ism: string }): Promise<{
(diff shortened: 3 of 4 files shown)


typescript/radix-sdk/src/ism/multisig-ism.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { ismTypeFromRadixIsmType } from '../utils/types.js';
+
+import { getMultisigIsmConfig } from './query.js';
+
+type MultisigIsmModule = {
+  config: MultisigIsmConfig;
+  addresses: IsmModuleAddresses;
+  derived: WithAddress<MultisigIsmConfig>;
+};
```
xeno097: note to self: think on how to dedup these

typescript/provider-sdk/src/ism.ts
```diff
@@ -1,11 +1,21 @@
...
 export type IsmModuleType = {
   config: IsmConfig;
   derived: DerivedIsmConfig;
   addresses: IsmModuleAddresses;
 };
 
+/**
+ * Helper type for ISM module types.
+ */
+export type ExtractIsmModuleType<TType extends IsmType> = ExtractModuleType<
```
antigremlin: Together with `ExtractModuleType` this feels too complex? I'd prefer making `IsmModuleType` generic or using the approach we take with `IsmConfigs` where you can do  `IsmConfigs['typeA' | 'typeB']`

typescript/radix-sdk/src/ism/routing-ism.ts
```diff
@@ -0,0 +1,236 @@
...
+  async read(address: string): Promise<WithAddress<DomainRoutingIsmConfig>> {
+    const { owner, routes } = await getDomainRoutingIsmConfig(
+      this.radixGateway,
+      address,
+    );
+
+    const ismReader = this.readerProvider.connectReader(this.provider);
+
+    const domains: Record<string, DerivedIsmConfig> = {};
+    for (const route of routes) {
```
antigremlin: For the record (we discussed this outside Github): let's not do any nested operations on the provider level. We should keep all compound logic (ISMs nested within ISMs in this case) in deploy-sdk.

typescript/provider-sdk/src/module.ts
```diff
@@ -30,8 +46,19 @@ export interface HypModule<M extends ModuleType> {
   update(config: Config<M>): Promise<AnnotatedTx[]>;
 }
 
-export interface ModuleProvider<M extends ModuleType> {
+/**
+ * Provides read-only access to artifacts modules on-chain.
+ */
+export interface ReaderProvider<M extends ModuleType> {
```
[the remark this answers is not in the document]
xeno097: Yup, my reason on splitting the reader only part was to make it clear by using the new interface that in the readers we are are in a readonly context, so instead of passing the whole module provider passing only the "reader" part makes it evident as we don't expose functions that only work in a read+write env

typescript/provider-sdk/src/chain.ts
```diff
@@ -19,6 +19,9 @@ export interface ChainMetadataForAltVM {
     confirmations?: number;
     estimateBlockTime?: number;
   };
+  gatewayUrls?: {
```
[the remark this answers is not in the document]
xeno097: This is a radix specific field thet is needed to retrieve some chain metadata and contract state, see:

https://github.com/hyperlane-xyz/hyperlane-registry/blob/main/chains/radix/metadata.yaml#L20

PR #7460, opened 2025-11-25
refactor: split EVM core schema

antigremlin:
## Description

Separates EVM-specific proxy factory addresses from protocol-agnostic core addresses to fix Zod validation errors when deploying to AltVM chains (Cosmos, Sealevel, etc.).

Previously, `DeployedCoreAddressesSchema` required EVM-specific factory addresses (e.g., `staticMerkleRootMultisigIsmFactory`) for all chains, causing validation failures on AltVM chains where these Solidity contracts don't exist. As a workaround, we set these fields to empty strings.

This PR is a more thorough solution:
- Introduce `BaseCoreAddressesSchema` containing only protocol-agnostic addresses (mailbox, validatorAnnounce, etc.)
- Extend with `EvmCoreAddressesSchema` for required EVM factory addresses
- Make `DeployedCoreAddressesSchema` accept optional factories for backward compatibility
- Add protocol-aware validation at CLI boundaries using appropriate schema based on chain protocol
- `EvmCoreModule` is now strictly typed with `EvmCoreAddresses`, ensuring all factories are present
- `AltVMCoreModule` returns `BaseCoreAddresses` without EVM-specific fields

## Backward Compatibility

Yes: Existing EVM chain artifacts with factory addresses continue to parse.

## Release Notes

* **Refactor**
  * Split core address schema and types into protocol-specific variants (base and EVM-specific)
  * Implemented protocol-aware validation for core addresses based on blockchain protocol
  * Enhanced type safety for deployment address handling across deployment modules

### DIFF
--- a/.changeset/famous-parrots-switch.md
+++ b/.changeset/famous-parrots-switch.md
@@ -1,0 +1,8 @@
+---
+"@hyperlane-xyz/provider-sdk": patch
+"@hyperlane-xyz/deploy-sdk": patch
+"@hyperlane-xyz/cli": patch
+"@hyperlane-xyz/sdk": patch
+---
+
+Refactored Zod schema and types for EVM Core into a base schema and an EVM-specific one


typescript/cli/src/deploy/core.ts
```diff
@@ -138,10 +140,12 @@ export async function runCoreApply(params: ApplyParams) {
 
   switch (multiProvider.getProtocol(chain)) {
     case ProtocolType.Ethereum: {
+      // Assert EVM-specific addresses for strict type checking
+      const evmAddresses = deployedCoreAddresses as EvmCoreAddresses;
```
xeno097: can't we use the zod schemas for validating and parsing here to avoid the type assertion?

typescript/deploy-sdk/src/AltVMCoreModule.ts
```diff
@@ -164,61 +164,28 @@ export class AltVMCoreModule implements HypModule<CoreModuleType> {
       mailboxAddress: mailbox.mailboxAddress,
     });
 
-    const addresses: DeployedCoreAddresses = {
+    const addresses: BaseCoreAddresses = {
       mailbox: mailbox.mailboxAddress,
-      staticMerkleRootMultisigIsmFactory: '',
       proxyAdmin: '',
```
ltyu: can we rid of these other defaults too?

paulbalaji: +1

PR #7462, opened 2025-11-25
chore: Upgrade relayers to latest

ameten:
### Description

Upgrade relayers to latest

### Backward compatibility

Yes

### Testing

Manual


ameten:
Updated mainnet to latest (and deployed)

PR #7463, opened 2025-11-25
chore: Remove MultiChainResolver for simplified chainResolver module

ltyu:
### Description
This removes the misdirection of 2 stateless classes in favor of a simplified module

### Backward compatibility
Yes

### Testing
existing e2e tests

### DIFF
--- a/.changeset/eight-guests-bake.md
+++ b/.changeset/eight-guests-bake.md
@@ -1,0 +1,5 @@
+---
+"@hyperlane-xyz/cli": minor
+---
+
+Remove MultiChainResolver for simplified chainResolver module


typescript/cli/src/context/strategies/chain/chainResolver.ts
```diff
@@ -0,0 +1,279 @@
...
+  argv: Record<string, any>,
+): Promise<ChainName[]> {
+  const commandKey = `${argv._[0]}:${argv._[1] || ''}`.trim() as CommandType;
+
+  switch (commandKey) {
+    case CommandType.WARP_DEPLOY:
+      return resolveWarpRouteConfigChains(argv);
+    case CommandType.WARP_SEND:
+    case CommandType.SEND_MESSAGE:
+    case CommandType.STATUS:
```
paulbalaji: does this mean `hyperlane status` will no longer ask me for PKs unless i add `--relay`? and specifically it will only ask me for PKs relevant to the message being relayed?

PR #7475, opened 2025-11-26
feat: implement artifact API for Radix ISMs

antigremlin:
This is a draft showcase of the would-be Artifact API and its implementation for Radix ISM. See also #7429.

While this is AI-written, it's not the first attempt and I'm satisfied with the structure.

### DIFF
--- a/typescript/radix-sdk/src/artifacts/ism-manager.ts
+++ b/typescript/radix-sdk/src/artifacts/ism-manager.ts
@@ -1,0 +1,193 @@
+import { ChainMetadataForAltVM } from '@hyperlane-xyz/provider-sdk/chain';
+import { IsmArtifact, IsmType } from '@hyperlane-xyz/provider-sdk/ism';
+import {
+  ArtifactReader,
+  ArtifactWriter,
+} from '@hyperlane-xyz/provider-sdk/module';
+import { assert } from '@hyperlane-xyz/utils';
+
+import { RadixProvider } from '../clients/provider.js';
+import { RadixSigner } from '../clients/signer.js';
+import {
+  MerkleRootMultisigIsmArtifactReader,
+  MerkleRootMultisigIsmArtifactWriter,
+  MessageIdMultisigIsmArtifactReader,
+  MessageIdMultisigIsmArtifactWriter,
+} from '../core/multisig-ism.js';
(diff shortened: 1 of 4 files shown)


typescript/provider-sdk/src/ism.ts
```diff
@@ -42,3 +42,21 @@ export type IsmModuleAddresses = {
...
+export interface RawIsmConfigs {
+  domainRoutingIsm: RawDomainRoutingIsmConfig;
+  merkleRootMultisigIsm: MultisigIsmConfig;
+  messageIdMultisigIsm: MultisigIsmConfig;
+  testIsm: TestIsmConfig;
+}
+
+export interface IsmArtifact<T extends keyof RawIsmConfigs> {
+  config: RawIsmConfigs[T];
+  addresses: { deployedIsm: string };
```
xeno097: no more `mailboxAddress` 🔥

typescript/provider-sdk/src/module.ts
```diff
@@ -30,6 +30,20 @@ export interface HypModule<M extends ModuleType> {
   update(config: Config<M>): Promise<AnnotatedTx[]>;
 }
 
+export interface ArtifactType {
+  config: unknown;
+  addresses: AddrMap;
+}
+
+export interface ArtifactReader<A extends ArtifactType> {
+  read(address: string): Promise<A['config']>;
```
xeno097: shouldn't it be `read(address: string): Promise<WithAddress<A['config']>>;` here?

typescript/provider-sdk/src/ism.ts
antigremlin: I didn't see the need for Radix ISM operations. Was it added for Aleo?

typescript/provider-sdk/src/module.ts
xeno097: mmm i'm thinking about the composite artifacts as it might be useful for the nested artifacts to have their address in the config like we do right now, i don't see how the current API might help in case we need that as reading would not populate the address field anymore 🤔

typescript/provider-sdk/src/ism.ts
xeno097: instead of having the mailbox address between the addresses, making it part of the ism config should cover those cases i think

PR #7478, opened 2025-11-26
refactor(cli, radix-sdk): move radix e2e test utils to radix sdk

xeno097:
### Description

This PR extracts the code to run a local radix node from the CLI package and puts it into the `radix-sdk` package under the `testing` sub-import path

Changes cherry picked from https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/7476

### Drive-by changes

- 

### Related issues

- 

### Backward compatibility

- YES

### Testing

- E2E
- Manual

## Summary by CodeRabbit

* **New Features**
  * Public Radix testing API added: node orchestration, contract download, and package deployment available via a testing subpath.
  * New Radix testing constants and sample local chain metadata exported.

* **Tests**
  * E2E test runner and setup enabled for local Radix integration.

* **Refactor**
  * CLI tests switched to use the public Radix testing API; local duplicate utilities removed.

* **Chores**
  * CI cache paths and dev tooling updated.


typescript/radix-sdk/src/testing/setup.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Downloads a file from a URL and returns it as a Uint8Array
+ */
+async function downloadFile(url: string): Promise<Uint8Array> {
+  const response = await fetch(url);
+  if (!response.ok || !response.body) {
+    throw new Error(`Failed to download ${url}: ${response.statusText}`);
+  }
+
+  return new Uint8Array(await response.arrayBuffer());
+}
```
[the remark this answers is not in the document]
xeno097: relax these are not use din production code, it is only for testing

.github/actions/yarn-cache/action.yml
paulbalaji: yarn-cache change lgtm

typescript/radix-sdk/.mocharc.e2e.json
```diff
@@ -0,0 +1,6 @@
+{
+  "import": ["tsx"],
+  "file": ["src/tests/e2e-test.setup.ts"],
+  "spec": ["src/tests/**/*.e2e-test.ts"],
```
antigremlin: Does this make sense, while the actual tests are still in `cli/src/tests/radix`?

typescript/radix-sdk/src/testing/index.ts
antigremlin: Having both `src/testing` and `src/tests` looks a bit confusing? Do we need two separate directories?

typescript/radix-sdk/src/testing/setup.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const packageAddress = await signer.publishPackage({
+    code: hyperlanePackageArtifacts.code,
+    packageDefinition: hyperlanePackageArtifacts.packageDefinition,
+  });
+
+  rootLogger.info(`Deployed Hyperlane package to: ${packageAddress}`);
+
+  // Update the chain metadata with the deployed package address
+  chainMetadata.packageAddress = packageAddress;
```
antigremlin: Does it make sense to modify `chainMetadata` given as input? In client code we always use constant chain metadata. Does anything later on in the tests relies on the `packageAddress` being set on the metadata by setup? That would be highly debatable.

typescript/radix-sdk/src/testing/setup.ts
```diff
@@ -0,0 +1,90 @@
...
+  // Fund the account with the internal signer
+  // Use retryAsync to handle transient errors (e.g., epoch expiry)
+  await retryAsync(
+    () => signer['signer'].getTestnetXrd(),
+    3, // attempts
+    3000, // base retry delay (3 seconds)
+  );
+
+  rootLogger.info(
+    `Funded test account on ${chainMetadata.name} before publishing the hyperlane package`,
```
antigremlin:
```suggestion
    `Funded test account on ${metadata.name} before publishing the hyperlane package`,
```

.changeset/angry-tips-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperlane-xyz/provider-sdk": minor
+---
+
+Add the gatewayUrls and packageAddress fields to the ChainMetadataForAltVM
```
antigremlin:
```suggestion
Added the gatewayUrls and packageAddress fields to the ChainMetadataForAltVM
```

typescript/radix-sdk/src/testing/constants.ts
```diff
@@ -0,0 +1,70 @@
...
+      http: 'http://127.0.0.1:5308',
+    },
+  ],
+  rpcUrls: [
+    {
+      http: 'http://127.0.0.1:3333/core',
+    },
+  ],
+  rpcPort: 3333,
+  restPort: 3333,
```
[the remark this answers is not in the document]
xeno097: Ah it because the code here is trying to keep compatibility with the `TestChainMetadata` type defined here https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/312826d10d6a768a74e37cc2c4e8bd8d54d43458/typescript/cli/src/tests/constants.ts

might be probably a good idea to move this to another package like the `provider-sdk` maybe, wdyt?

typescript/radix-sdk/src/testing/setup.ts
xeno097: yeah  i needed to set the radix hyperlane package definition address for testing as radix uses a factory pattern for creating artifacts on chain where the sdk calls a method on `packageAddress` and the code at that address instantiates the artifact based on the provided input.

As the `packageAddress` needs to be deployed before running the tests themselves and its address changes with each deployment, the way i solved this was by modifying the in memory metadata to add that property so we can use it for testing. 

while the problem can be solved here in some way the same can't be said for the cli tests as there isn't a way to pass this address to the cli for deployment hence i decided to use the metadata for reading the deployed package address

see: https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/2ed21c97d4077688bc46946ed4bb0a8a0f0f978b/typescript/cli/src/tests/radix/e2e-test.setup.ts#L83-L96

typescript/radix-sdk/src/testing/index.ts
xeno097: I think what helps in the confusion is that the `test` folder is inside the `src` folder while in other cases (other projects) i have tended to have it next to the `src` folder to make it clear that those are e2e tests but here it seems that the convention is to have it inside, we might consider changing that in the future too

PR #7483, opened 2025-11-28
fix: starknet implementation contracts, balance and addresses validation

Xaroz:
### Description

This PR includes multiple fixes to the StarknetTokenAdapter and address validation

- For proxy contracts (in the case of Paradex native token), the `contractInstance` declared did not have most of the methods, so we need to get the `implementation` contract
- Old starknet contracts do not include` balance_of` function so we need to fallback to `balanceOf`. In the case of the Fuel Native Token for Paradex, it returns an object that contains a high/low instead
- Update `isAddressStarknet` to first parse and validate with the starknet package and then test with the regex

### Testing

UI test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with StarkNet contracts using implementation-proxy patterns for contract resolution and instantiation.
  * Enhanced token balance retrieval to handle different response formats across Cairo and Ethereum-like interfaces.
  * Strengthened StarkNet address validation by parsing before applying regex, with safer error handling.

* **Chores**
  * Added changeset entries to record version bumps and release notes.

### DIFF
--- a/typescript/sdk/src/token/adapters/StarknetTokenAdapter.ts
+++ b/typescript/sdk/src/token/adapters/StarknetTokenAdapter.ts
@@ -7,6 +7,7 @@ import {
   cairo,
   num,
   shortString,
+  uint256,
 } from 'starknet';
 
 import {
@@ -17,6 +18,7 @@ import {
   addressToBytes32,
   assert,
   ensure0x,
+  isNullish,
 } from '@hyperlane-xyz/utils';
 
 import { BaseStarknetAdapter } from '../../app/MultiProtocolApp.js';
@@ -56,6 +58,9 @@ export class StarknetTokenAdapter
     super(chainName, multiProvider, addresses);
   }
 
+  // This function returns the Contract instance for the given token
+  // If the contract is a Proxy: retrieves the implementation contract ABI and returns that Contract
+  // If the contract is not a Proxy: returns the contract with the token address ABI
   async getContractInstance(): Promise<Contract> {
     if (this.tokenContract) {
       return this.tokenContract;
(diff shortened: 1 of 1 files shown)


typescript/sdk/src/token/adapters/StarknetTokenAdapter.ts
```diff
@@ -69,14 +69,27 @@ export class StarknetTokenAdapter
...
     return contractInstance;
   }
 
   async getBalance(address: Address): Promise<bigint> {
     const contract = await this.getContractInstance();
+    if (contract.balance_of) return contract.balance_of(address);
 
-    return contract.balance_of(address);
+    const response = await contract.balanceOf(address);
+    return response.balance?.low ?? response;
```
tkporter: would recommend using some uint256 utils to get this as a bigint instead of using just the .low, e.g.
```
import { uint256 } from 'starknet';

const mybigint = uint256.uint256ToBN(myuint256);
```

some links: https://starknetjs.com/docs/7.6.4/API/interfaces/types.Uint256 and https://starknetjs.com/docs/7.6.4/API/namespaces/uint256/#functions

typescript/sdk/src/token/adapters/StarknetTokenAdapter.ts
```diff
@@ -69,14 +69,27 @@ export class StarknetTokenAdapter
       provider,
     );
 
+    if (contractInstance.get_implementation) {
+      const { implementation } = await contractInstance.get_implementation();
+      const contractClass = await provider.getClassByHash(implementation);
+      const implementationContract = new Contract(
```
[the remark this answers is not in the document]
Xaroz: added a comment to clarify what is going in this branching case

typescript/sdk/src/token/adapters/StarknetTokenAdapter.ts
Xaroz: thanks for this!

PR #7509, opened 2025-12-03
chore: update validator images for testnet

kamiyaa:
### Description

update testnet4 validators to latest image


kamiyaa:
> I guess you want to roll out the new reorg parsing behavior, I think it would make sense to also update the relayer in that case, as the relayer is also affected.

Good point :+1:

PR #7525, opened 2025-12-03
chore: Add cached Providers and Signers

ltyu:


typescript/cli/src/commands/warp.ts
```diff
@@ -402,9 +403,21 @@ export const check: CommandModuleWithContext<SelectWarpRouteBuilder> = {
       deployedRoutersAddresses,
     });
 
+    // Create altVmProviders dynamically for this call
```
ltyu: This is temporary given that expandWarpDeployConfig still expects altVmProviders

PR #7531, opened 2025-12-04
chore(ci): pin Foundry version via .foundryrc

paulbalaji:
## Summary
- Add `.foundryrc` file to pin Foundry version to `v1.5.0`
- Update `setup-foundry` action to read version from `.foundryrc`
- Update Dockerfile to use pinned version from `.foundryrc`

This ensures consistent Foundry versions across CI, Docker builds, and local development.

For local development, run:
```bash
foundryup --install $(cat .foundryrc)
```

## Test plan
- [ ] CI installs pinned Foundry version
- [ ] Docker build uses pinned version


larryob (commented):
Thanks for this!

PR #7544, opened 2025-12-05
feat: separate rebalancer package

paulbalaji:
## Summary
- Extract rebalancer logic from CLI into dedicated `@hyperlane-xyz/rebalancer` package
- New package supports both manual CLI execution and continuous daemon mode for K8s deployments
- CLI now imports from new package, maintaining backward compatibility for manual rebalancing

https://www.notion.so/hyperlanexyz/Split-Rebalancer-logic-out-of-CLI-2bd6d35200d680ff8bd9f236833b5d59?source=copy_link

## Test plan
- [x] Build passes
- [x] Tests pass
- [x] Tested in infra (replaced an existing rebalancer with this new monorepo image)

## Follow-Up (TBD)

https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/7545/

## Release Notes

* **New Features**
  * Introduced a dedicated rebalancer package enabling standalone daemon mode for continuous Kubernetes deployments, while maintaining manual CLI execution support.

* **Refactoring**
  * Extracted rebalancer logic from CLI into a separate package with a unified service interface for both interactive and continuous operation modes.


typescript/rebalancer/src/utils/files.ts
```diff
@@ -0,0 +1,266 @@
+import { input } from '@inquirer/prompts';
```
xeno097: This file really makes me think that we would benefit from a `utils/fs` package/export cause the cli and infra already use different sets of filesystem based utils

typescript/rebalancer/src/service.ts
```diff
@@ -0,0 +1,144 @@
...
+  const { join } = await import('path');
+  const { fileURLToPath } = await import('url');
+  const __dirname = fileURLToPath(new URL('.', import.meta.url));
+  const packageJson = JSON.parse(
+    readFileSync(join(__dirname, '../package.json'), 'utf-8'),
+  );
+  VERSION = packageJson.version;
+} catch (error) {
+  rootLogger.warn({ error }, 'Could not read version from package.json');
+}
```
xeno097: why not have this inside the `main` func?

xeno097 (approved):

typescript/rebalancer/src/service.ts
```diff
@@ -0,0 +1,154 @@
...
+    'Starting Hyperlane Rebalancer Service',
+  );
+
+  try {
+    // Load rebalancer configuration
+    const rebalancerConfig = RebalancerConfig.load(configFile);
+    logger.info('✅ Loaded rebalancer configuration');
+
+    // Initialize registry (uses default registry URIs)
+    const registry = getRegistry({
```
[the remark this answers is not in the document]
paulbalaji: it means we're using the actual GithubRegistry instead of the hacky monorepo/registry clone thing we currently do https://github.com/hyperlane-xyz/hyperlane-registry/blob/a0eff661e93ac281810084a34e75622b1520dca3/src/registry/GithubRegistry.ts#L47

PR #7591, opened 2025-12-10
feat: Lander for Aleo: Implement payload reverted method

ameten:
### Description

Lander for Aleo: Implement payload reverted method

### Related issues

https://linear.app/hyperlane-xyz/issue/ENG-2720/lander-aleo

### Backward compatibility

Yes

### Testing

Unit tests

## Summary by CodeRabbit

* **New Features**
  * Added on-chain mapping-existence checks for Aleo and integrated them into payload delivery verification.
  * Improved payload revert detection to inspect on-chain delivery status and mark reverted payloads accordingly.
  * Introduced a DeliveryKey type, changed mapping key representation to a typed plaintext, and re-exported additional Aleo serialization utilities.

* **Tests**
  * Added comprehensive tests and updated mocks to cover mapping-existence checks and revert detection across transaction statuses.


rust/main/chains/hyperlane-aleo/src/provider/lander.rs
```diff
@@ -73,4 +92,22 @@ impl<C: AleoClient> AleoProviderForLander for AleoProvider<C> {
...
+
+        let key = StringType::from_str(mapping_key)
+            .map_err(|e| ChainCommunicationError::CustomError(e.to_string()))?;
+
+        let plain_text: Option<Plaintext<CurrentNetwork>> = self
+            .get_mapping_value(program_id, mapping_name, &key)
+            .await?;
+
+        Ok(plain_text.is_some())
+    }
```
yjamin: This code will fail. I still don't understand why you don't just pass the Plaintext object?

```rust
async fn mapping_value_exists(&self, program_id: &str, mapping_name: &str, key: &Plaintext<CurrentNetwork>) -> ChainResult<bool> {
  let plain_text: Option<Plaintext<CurrentNetwork>> = self
            .get_mapping_value(program_id, mapping_name, key)
            .await?;
  return Ok(plaintext.is_some())
}

**Review by yjamin — CHANGES_REQUESTED** [review]

**Comment by yjamin** [comment]
We can change the key type to `Plaintext<CurrentNetwork>`, I believe this would make a lot easier. 

https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/81a0ad1b920f468297d6f17d1da174c7da8c7bae/rust/main/chains/hyperlane-aleo/src/types.rs#L341

File: rust/main/lander/src/adapter/chains/aleo/adapter/payload.rs [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+                    // Query on-chain to check if the delivery record exists
+                    let delivered = self
+                        .provider
+                        .mapping_value_exists(
+                            &get_mapping_value.program_id,
+                            &get_mapping_value.mapping_name,
+                            &get_mapping_value.mapping_key,
+                        )
+                        .await
+                        .unwrap_or(false); // Treat errors as "not delivered"
```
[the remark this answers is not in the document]
ameten: Fixed

rust/main/chains/hyperlane-aleo/src/provider/lander.rs
```diff
@@ -73,4 +91,19 @@ impl<C: AleoClient> AleoProviderForLander for AleoProvider<C> {
...
         self.get_unconfirmed_transaction(transaction_id).await
     }
+
+    async fn mapping_value_exists(
+        &self,
+        program_id: &str,
+        mapping_name: &str,
+        mapping_key: &Plaintext<CurrentNetwork>,
+    ) -> ChainResult<bool> {
+        use snarkvm::prelude::Plaintext;
```
yjamin: Any reason why you import this again here?

rust/main/lander/src/adapter/chains/aleo/adapter/payload.rs
```diff
@@ -0,0 +1,81 @@
...
+                    // If provider returns error, treat as delivered (not reverted) with unwrap_or(true)
+                    let delivered = self
+                        .provider
+                        .mapping_value_exists(
+                            &get_mapping_value.program_id,
+                            &get_mapping_value.mapping_name,
+                            &key,
+                        )
+                        .await
+                        .unwrap_or(true);
```
yjamin: I'm still a bit confused by this, is this really what we want? Will the relayer still try to send a new transaction if the payload ends up not being delivered, but we interpreted it as delivered because we got an RPC error?

Should we not skip verification for this playload instead in this case?

yjamin: In that case I think the code is correct.

yjamin (approved):

PR #7602, opened 2025-12-11
chore: migrate file submitter to deploy-sdk

ltyu:
### Description
This PR migrates the altvm file submitter to deploy-sdk. This sets up for any additional Submitter abstraction changes.

### Drive-by changes
Given that the altvm json submitter exists in provider-sdk, it is also moved over to the deploy-sdk


antigremlin:
> I thought deploy/sdk is where we _could_ use fs

We should be able to. @paul has just introduced this utils/fs module. Maybe he can help with adjusting the linter rules?

xeno097 (approved):

PR #7608, opened 2025-12-12
feat: merge 7594 and 7580 new artifact state option

xeno097:
### Description

Updates the `RawArtifact` to be recursive over raw config types, enabling recursive reads in the provider packages

- defines the `ArtifactOnChain` for describing configs that might be either completely derived or just an address for an artifact that is deployed but which config has not been derived

### Related issues

- 

### Backward compatibility

- Yes

### DIFF
--- a/typescript/deploy-sdk/src/ism/routing-ism.ts
+++ b/typescript/deploy-sdk/src/ism/routing-ism.ts
@@ -9,6 +9,7 @@ import {
 } from '@hyperlane-xyz/provider-sdk/artifact';
 import {
   DeployedIsmAddresses,
+  DeployedIsmArtifact,
   IRawIsmArtifactManager,
   IsmArtifactConfig,
   RawRoutingIsmArtifactConfig,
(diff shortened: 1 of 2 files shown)


typescript/provider-sdk/src/artifact.ts
```diff
@@ -27,6 +28,21 @@ export type ArtifactDeployed<C, D> = {
   deployed: D;
 };
 
+/**
+ * Represents an artifact that has been deployed on chain
+ * but is represented only by its address
+ */
+export type ArtifactUnderived = {
+  artifactState: typeof ArtifactState.UNDERIVED;
+  artifactAddress: string;
```
antigremlin: As discussed elsewhere, there may be good reasons to use the derived type `<D>` here to accommodate situations when new contracts are represented by more than a single string address.

xeno097: Yup just updated to use the `D` param

PR #7612, opened 2025-12-12
chore: typescript version update

xeno097:
### Description

This PR explores if upgrading the typescript version would break anything in the project. Main reason for trying the upgrade are:
- The stricter TS config defined at https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/61c7ac7dc096c62786374901fef1215ca892fe6a/typescript/tsconfig/base.tsconfig.json uses `es2023` which is not supported by the current TS version (ts 5.3.3 supports up to es2022)
- Since TS 5.3.3, newer typescript versions have added type safety and inference improvements 

### Drive-by changes

- updated `@types/node` to v20 for better support with the new ts version
- updated imports of json files as TS 5.7 with `nodenext` requires to add an import attribute (see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#validated-json-imports-in---module-nodenext)
- Removed existing const enums as we are moving away from enums in general during the `multi-vm` migration
- Removed ts-node usage in tests becuase it doesn't work well with esm and replaced with tsx

### Backward compatibility

 YES, No public facing code has changed

### Testing

- Locally building and linting the packages

### DIFF
--- a/.changeset/beige-foxes-compare.md
+++ b/.changeset/beige-foxes-compare.md
@@ -1,5 +1,0 @@
----
-"@hyperlane-xyz/utils": patch
----
-
-Use stringifyObject in writeJson and writeYaml for proper ethers BigNumber serialization. Rename removeEndingSlash to removeTrailingSlash.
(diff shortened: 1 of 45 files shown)


xeno097:
> we should consider doing a big ts + node 24 update at the same time

i can try after fixing the ts errors ahaha

PR #7630, opened 2025-12-16
chore: dec16 chain deprecations

paulbalaji:
### Description

chore: dec16 chain deprecations

removing: form injective mint neutron osmosis incentivtestnet milkywaytestnet

note that we're only update registry for a subset of these chains that we know are going to stay down, but keeping option open for re-enabling others if required

### Related issues

https://github.com/hyperlane-xyz/hyperlane-registry/pull/1293


typescript/infra/config/environments/mainnet3/agent.ts
```diff
@@ -907,7 +907,7 @@ const hyperlane: RootAgentConfig = {
     rpcConsensusType: RpcConsensusType.Fallback,
     docker: {
       repo,
-      tag: '7cb519d-20251203-140154',
+      tag: '26d9ace-20251216-085234',
```
xeno097: These will need to be updated after the registyr pr is merged right?

PR #7640, opened 2025-12-17
fix: patch node-fetch to a better user-agent value

antigremlin:
## Description

Fixes Safe API rate limiting errors by patching node-fetch v2.7.0 to use a simpler User-Agent header.

The Safe Transaction Service API was returning "Too Many Requests" (429) errors when requests included node-fetch's default User-Agent: node-fetch/1.0 (+https://github.com/bitinn/node-fetch). This PR implements a dual-layer fix:

- Patches node-fetch@2.7.0 to set User-Agent to node-fetch instead of the verbose default. This automatically fixes all requests made through SafeApiKit (which depends on node-fetch internally).
- Adds explicit User-Agent headers to direct fetch() calls in safe.ts to align all calls to Safe API.

This supersedes the approach in PR #7283 which attempted to fix the issue by upgrading to node-fetch v3. Instead, we stay on v2 and set the correct User-Agent value.

## Backward Compatibility

Yes. This change only modifies the User-Agent header value in HTTP requests. It does not affect any public APIs, function signatures, or behavior of the codebase. The patch is applied automatically via pnpm's patching mechanism on every install.

## Summary by CodeRabbit

* **Chores**
  * Simplified dependency configuration by switching to a patch-based approach for the HTTP client.
  * Added a release changeset entry for a patch release.

* **Bug Fixes**
  * Ensures outgoing Safe API requests include a User-Agent header to improve compatibility and reduce 429 responses.

### DIFF
--- a/.changeset/pink-trees-smash.md
+++ b/.changeset/pink-trees-smash.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/infra': patch
+---
+
+Patched node-fetch@2.7.0 to avoid 429 errors with Safe API, reverting from premature upgrade to node-fetch@3


typescript/infra/src/utils/safe.ts
```diff
@@ -264,6 +264,7 @@ export async function deleteAllPendingSafeTxs(
     method: 'GET',
     headers: {
       'Content-Type': 'application/json',
+      'User-Agent': 'node-fetch',
```
antigremlin: This assumes this code is only run from the CLI, which seems to be true.

PR #7641, opened 2025-12-17
fix(cli): don't require keys for read-only status checks

paulbalaji:
## Summary
- Add `status` command as a conditional sign command (only requires signing when `--relay` flag is passed)
- Fix `isSignCommand` to return `false` instead of `undefined` for non-sign commands
- Add graceful handling for missing signers when relaying
- Add unit tests for `signCommands` and `message` status

## Motivation
Previously, running `hyperlane status` required providing keys for **all protocols** (EVM, CosmosNative, etc.) even when just checking message status:

```bash
# Before: required ALL protocol keys even for read-only status check
hyperlane status --id <message-id>
# Error: missing private key for CosmosNative...
```

Now keys are only required when actually needed for relaying:

```bash
# No keys required - just checking status
hyperlane status --id <message-id>

# Keys required only for destination chain protocol
hyperlane status --id <message-id> --relay --key.ethereum <key>
```

Additionally, when using `--relay`, if you only have keys for some protocols, the command will:
- Relay messages to chains you have signers for
- Skip messages to chains you don't have signers for (with a warning)
- No longer fail if you're missing keys for chains you don't need to relay to

## Test plan
- [x] Unit tests pass (`pnpm --filter @hyperlane-xyz/cli test:ci`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Message status checks no longer require private keys; keys are only needed with --relay and only for the destination chain.
  * Status now skips messages lacking a signer and emits clear warnings instead of attempting relays.

* **New Features**
  * Status command marked conditional (requires signing only when relaying).
  * Added a non-blocking "quick" option to some test helpers to speed checks.

* **Tests / CI**
  * Added unit and end-to-end tests for sign-command detection, relayability filtering, and status behaviors; CI test matrix updated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -265,6 +265,7 @@ jobs:
           - core-read
           # Other commands
           - relay
+          - status
           # Warp Apply Commands
           - warp-apply-hook-updates
           - warp-apply-ism-updates
--- a/typescript/cli/src/tests/ethereum/commands/helpers.ts
+++ b/typescript/cli/src/tests/ethereum/commands/helpers.ts
@@ -353,12 +353,41 @@ export function localTestRunCmdPrefix() {
 export async function hyperlaneSendMessage(
   origin: string,
   destination: string,
+  { quick = false }: { quick?: boolean } = {},
 ) {
   return $`${localTestRunCmdPrefix()} hyperlane send message \
         --registry ${REGISTRY_PATH} \
         --origin ${origin} \
         --destination ${destination} \
         --key ${ANVIL_KEY} \
+        --verbosity debug \
+        ${quick ? ['--quick'] : []} \
+        --yes`;
+}
+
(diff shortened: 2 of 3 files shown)


typescript/cli/src/commands/status.ts
```diff
@@ -20,8 +24,10 @@ export const statusCommand: CommandModuleWithWriteContext<
     },
   },
   handler: async ({ context, origin, id, relay, dispatchTx }) => {
+    // When --relay is passed, signers are initialized and context is a WriteCommandContext
+    // Otherwise it's just a CommandContext (read-only operations)
     await checkMessageStatus({
-      context,
+      context: context as CommandContext | WriteCommandContext,
```
xeno097: do we really need this type assertion?

xeno097 (approved):

PR #7644, opened 2025-12-17
fix: don't export file submitter from provider-sdk

antigremlin:
## Description

Removes AltVMFileSubmitter from the main public exports of @hyperlane-xyz/deploy-sdk to prevent bundling Node.js file system dependencies in browser environments. The class is still available via a subpath export: @hyperlane-xyz/deploy-sdk/AltVMFileSubmitter.

This fixes build issues when deploy-sdk is consumed by browser-based applications, as AltVMFileSubmitter uses Node.js fs module which cannot be bundled for the browser.

## Backward compatibility

No - consumers using AltVMFileSubmitter updated to use a subpath import.

## Summary by CodeRabbit

* **Chores**
  * Adjusted package exports: added a new direct public export for a submitter utility and removed its aggregated re-export from the index. No functional or runtime behavior changes for end users.

### DIFF
--- a/typescript/deploy-sdk/src/index.ts
+++ b/typescript/deploy-sdk/src/index.ts
@@ -1,6 +1,6 @@
 // AltVMFileSubmitter removed from public exports as it uses Node.js fs module
 // and should not be bundled for browser use. CLI can import directly:
-// import { AltVMFileSubmitter } from '@hyperlane-xyz/deploy-sdk/AltVMFileSubmitter.js';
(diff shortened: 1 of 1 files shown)


typescript/deploy-sdk/package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "types": "./dist/index.d.ts",
   "exports": {
```
xeno097: Should we also add the type exports here like it is done from the radix-sdk for the `testing` sub import path?

PR #7737, opened 2026-01-12
chore: add chunk size & from index for aleo indexing

yjamin:
### Description
The default chunk size of 2000 is too high for aleo indexing and results in errors. I arrived at 50 by just testing some values and this seems to be the most consistent.

### DIFF
--- a/rust/main/config/mainnet_config.json
+++ b/rust/main/config/mainnet_config.json
@@ -9396,7 +9396,9 @@
       "validatorAnnounce": "0xf3aa0d652226e21ae35cd9035c492ae41725edc9036edf0d6a48701b153b90a0",
       "technicalStack": "other",
       "availability": {
-        "reasons": ["deprecated"],
+        "reasons": [
+          "deprecated"
+        ],
         "status": "disabled"
       }
     },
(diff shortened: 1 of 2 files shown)


rust/main/config/mainnet_config.json
```diff
@@ -9255,6 +9255,10 @@
       "signer": {
         "key": "0x476e6d1e316263e22bf74e6cae6fcdf7ecc51cd4c0450e5661c35ccce3a85601"
       },
+      "index": {
+        "chunk": 50,
+        "from": 14207900
+      },
```
paulbalaji: do we need to update testnet as well?

rust/main/config/mainnet_config.json
```diff
@@ -9392,9 +9396,7 @@
       "validatorAnnounce": "0xf3aa0d652226e21ae35cd9035c492ae41725edc9036edf0d6a48701b153b90a0",
       "technicalStack": "other",
       "availability": {
-        "reasons": [
-          "deprecated"
-        ],
+        "reasons": ["deprecated"],
         "status": "disabled"
```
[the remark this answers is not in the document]
yjamin: I wonder why my formatter is different, but yeah I'll update the format.

rust/main/config/mainnet_config.json
yjamin: Probably best to update it as well, yes.

PR #7747, opened 2026-01-12
fix: use starknet regex for validating starknet addresses

Xaroz:
### Description

- Fixed an issue where EVM addresses were allowed when validating for starknet addresses
- Updated the regex to also allow 65 characters (this is happens when copy/pasting an address from Paradex)
  - 63 chars - 63 hex chars (no prefix)
  - 64 chars - 64 hex chars (no prefix)
  - 65 chars - 0x + 63 hex chars
  - 66 chars - 0x + 64 hex chars
- After verifying the regex it will continue to validate with `validateAndParse` address from the starknet library
- Add tests to `isAddressStarknet` and `isValidStarknetAddress`

### Related issues

Fixes [ENG-3071](https://linear.app/hyperlane-xyz/issue/ENG-3071/fix-issue-with-starknet-allowing-evm-addresses)
### Backward compatibility

Yes
### Testing

Tested in the UI with:
- EVM address (not allowed now)
- Starknet and Paradex addresses (was able to transfer without issues)
- Out of bound addresses

Tested with a script with an invalid address:

```
function main() {
  const address =
    '0x5ab3ac43afd012da5037f72691f9791a9fd610900c0a1d6c18d41367aee9a530';
  try {
    validateAndParseAddress(address);
    console.log('message is valid');
  } catch (err) {
    console.log(err);
  }
}

main();
-----------------------
Error: Message not signable, invalid Starknet Address length.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed StarkNet address validation to accept both 63- and 64-character hex addresses so valid addresses are not rejected.

* **New Features**
  * Added explicit utility checks to detect and validate StarkNet address formats.

* **Tests**
  * Added test coverage for varied StarkNet address formats (with/without prefix, differing lengths).

* **Chores**
  * Added a changeset entry to record the patch release.


typescript/utils/src/addresses.test.ts
```diff
@@ -101,4 +113,42 @@ describe('Address utilities', () => {
...
+
+  describe('isAddressStarknet', () => {
+    it('Validates correct Starknet addresses', () => {
+      for (const address of STARKNET_ADDRESSES) {
+        expect(isAddressStarknet(address)).to.be.true;
+      }
+    });
+
+    it('Rejects EVM addresses', () => {
+      const evmAddress = '0x0000000000000000000000000000000000000001';
```
xeno097: Probably worth using a random evm address that is not the 1 address

PR #7807, opened 2026-01-15
feat(sdk): add real-time validator signature status to MetadataBuilder

nambrot:
### Description
Adds real-time validator signature status to the SDK's MetadataBuilder, enabling applications like the Hyperlane Explorer to show which validators have signed a message before it's been delivered.

Key changes:

- MetadataBuilder.build() now returns detailed MetadataBuildResult objects containing validator signature status ('signed' | 'pending' | 'error'), checkpoint indices, actual signatures, and human-readable validator aliases
- Aggregation and routing ISMs return nested results for their sub-modules
- New helper functions: isMetadataBuildable(), getSignedValidatorCount(), isQuorumMet()
- New exports: ValidatorInfo, MetadataBuildResult, DerivedHookConfig, and all variant types
Drive-by changes

- EvmIsmReader optimization: When messageContext is provided, routing ISM derivation short-circuits to only derive the specific ISM for that message (~87% faster: ~5.7s → ~724ms)
- EvmHookReader optimization: Parallelized RPC calls across all derivation methods (protocol fee, pausable, IGP, aggregation, routing hooks)
- SmartProvider fix: Fixed retry logic to correctly identify permanent errors (JSON-RPC error code 3, empty return decode failures)

### Related issues

N/A

### Backward compatibility

No - the MetadataBuilder.build() method now returns MetadataBuildResult instead of string. Callers need to access .metadata on the result to get the encoded bytes. 

### Testing

Unit Tests - Updated existing hardhat tests for new return types, added SmartProvider tests for permanent error handling, added EvmIsmReader tests for messageContext behavior.

## Summary by CodeRabbit

* **Breaking Changes**
  * MetadataBuilder.build() now returns a result object — use the .metadata property to access encoded bytes.

* **New Features**
  * Per-validator signature status (signed/pending/error) with checkpoint info and signatures.
  * New helpers/types for metadata and quorum checks (isMetadataBuildable, getSignedValidatorCount, isQuorumMet, ValidatorInfo, MetadataBuildResult).

* **Improvements**
  * Faster ISM routing and hook derivation via parallel RPCs and routing short-circuits.
  * Improved error classification and more robust provider fallback behavior.

### DIFF
--- a/.changeset/relayer-validator-status.md
+++ b/.changeset/relayer-validator-status.md
@@ -1,7 +1,7 @@
 ---
-"@hyperlane-xyz/relayer": minor
+"@hyperlane-xyz/relayer": major
 ---
 
-**BREAKING**: `MetadataBuilder.build()` now returns `MetadataBuildResult` instead of `string`. Access `.metadata` on the result to get the encoded bytes.
+**BREAKING**: `MetadataBuilder.build()` return type changed from `string` to `MetadataBuildResult`. Access `.metadata` on the result to get the encoded bytes.
 
-Added real-time validator signature status to MetadataBuilder. The builder now returns detailed `ValidatorInfo` for each validator including signing status ('signed' | 'pending' | 'error'), checkpoint indices, and actual signatures. Aggregation and routing ISMs return nested results for sub-modules. New helper functions: `isMetadataBuildable()`, `getSignedValidatorCount()`, `isQuorumMet()`.
+Added real-time validator signature status to MetadataBuilder. The builder returns detailed `ValidatorInfo` for each validator including signing status ('signed' | 'pending' | 'error'), checkpoint indices, and actual signatures. Aggregation and routing ISMs return nested results for sub-modules. Added helper functions: `isMetadataBuildable()`, `getSignedValidatorCount()`, `isQuorumMet()`.
--- a/typescript/relayer/src/metadata/multisig.ts
+++ b/typescript/relayer/src/metadata/multisig.ts
@@ -365,7 +365,7 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
 
     // Count signed validators
     const signedValidators = validatorInfos.filter(
-      (v) => v.status === 'signed',
+      (v) => v.status === ValidatorStatus.Signed,
     );
     const signedCount = signedValidators.length;
     const quorumMet = signedCount >= context.ism.threshold;
--- a/typescript/relayer/src/metadata/types.ts
+++ b/typescript/relayer/src/metadata/types.ts
@@ -164,7 +164,8 @@ export function isMetadataBuildable(
 export function getSignedValidatorCount(
   result: MultisigMetadataBuildResult,
 ): number {
-  return result.validators.filter((v) => v.status === 'signed').length;
+  return result.validators.filter((v) => v.status === ValidatorStatus.Signed)
+    .length;
 }
 
 /** Check if quorum is met for multisig results */


typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -151,12 +174,118 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
...
+    },
+  ): Promise<{
+    validatorInfos: ValidatorInfo[];
+    checkpoint?: Checkpoint;
+  }> {
+    const originChain = this.core.multiProvider.getChainName(match.origin);
+
+    // Get S3 validators (may throw if no storage location announced)
+    let s3Validators: S3Validator[];
+    try {
```
ltyu: i wonder if this try catch should be inside of `this.s3Validators()` based on the comment i made below about partial failure.

perhaps we can use Promise.allSettled instead of Promise.all there

typescript/sdk/src/hook/EvmHookReader.ts
```diff
@@ -331,9 +331,15 @@ export class EvmHookReader extends HyperlaneReader implements HookReader {
...
-    this.assertHookType(await hook.hookType(), OnchainHookType.AGGREGATION);
 
-    const hooks = await hook.hooks(ethers.constants.AddressZero);
+    // Parallelize hookType and hooks list fetching
+    const [hookType, hooks] = await Promise.all([
+      hook.hookType(),
+      hook.hooks(ethers.constants.AddressZero),
+    ]);
+
+    this.assertHookType(hookType, OnchainHookType.AGGREGATION);
```
yorhodes: this is a little silly

typescript/sdk/src/ism/metadata/aggregation.ts
```diff
@@ -38,12 +39,23 @@ export class AggregationMetadataBuilder implements MetadataBuilder {
...
       { context, maxDepth, timeoutMs },
       'Building aggregation metadata',
     );
-    assert(maxDepth > 0, 'Max depth reached');
+
+    if (maxDepth <= 0) {
+      return {
+        type: context.ism.type,
+        ismAddress: context.ism.address,
+        threshold: context.ism.threshold,
```
yorhodes: why are we returning this when its passed in?

typescript/sdk/src/ism/metadata/arbL2ToL1.ts
```diff
@@ -45,26 +49,65 @@ export class ArbL2ToL1MetadataBuilder implements MetadataBuilder {
...
+          ...baseResult,
+          bridgeStatus: 'unconfirmed',
+          blocksRemaining,
+        };
+      }
+
+      if (errorMessage.includes('already been executed')) {
+        return {
+          ...baseResult,
+          bridgeStatus: 'executed',
```
yorhodes: was this intended?

typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -151,12 +174,118 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
...
+    match: {
+      origin: number;
+      merkleTree: Address;
+      messageId: string;
+      index: number;
+    },
+  ): Promise<{
+    validatorInfos: ValidatorInfo[];
+    checkpoint?: Checkpoint;
+  }> {
```
yorhodes: can this just refactor the currente getS3Checkpoints?

yorhodes (approved):

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
```diff
@@ -322,13 +323,28 @@ export class HyperlaneSmartProvider
...
+          const isEmptyReturnDecodeFailure =
+            errorCode === EthersError.CALL_EXCEPTION &&
+            !hasRevertData &&
+            !nestedError;
           const isCallExceptionWithoutData =
-            errorCode === EthersError.CALL_EXCEPTION && !hasRevertData;
+            errorCode === EthersError.CALL_EXCEPTION &&
+            !hasRevertData &&
+            !isJsonRpcRevert &&
+            !isEmptyReturnDecodeFailure;
```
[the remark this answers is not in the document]
nambrot-agent: Fixed! Changed `nestedError?.error?.code` to `nestedError?.code`. Good catch on the double-nesting issue.

typescript/sdk/src/ism/metadata/types.ts
```diff
@@ -27,6 +30,139 @@ export interface MetadataContext<
...
+// ============================================
+// Validator info (for multisig ISMs)
+// ============================================
+export interface ValidatorInfo {
+  /** Validator address */
+  address: Address;
+  /** Human-readable name from defaultMultisigConfigs */
+  alias?: string;
+  /** Signature status */
+  status: 'signed' | 'pending' | 'error';
```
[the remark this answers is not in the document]
nambrot-agent: Done! Added `ValidatorStatus` const object with `Signed`, `Pending`, `Error` values. This eliminates the need for `'error' as const` casts throughout the code.

typescript/sdk/src/ism/metadata/multisig.ts
```diff
@@ -151,12 +174,118 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
...
+    const originChain = this.core.multiProvider.getChainName(match.origin);
+
+    // Get S3 validators (may throw if no storage location announced)
+    let s3Validators: S3Validator[];
+    try {
+      s3Validators = await this.s3Validators(originChain, validators);
+    } catch (error) {
+      // If we can't fetch storage locations at all, return all as error
```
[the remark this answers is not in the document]
nambrot-agent: Fixed. The `s3Validators()` method now uses `mapAllSettled` to handle partial failures gracefully. Individual validators without valid storage locations are returned as `undefined` rather than failing the entire batch.

typescript/relayer/src/metadata/multisig.ts
```diff
@@ -151,12 +174,118 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
...
+      };
+    }
+
+    // Fetch checkpoints from each validator in parallel
+    const results = await Promise.allSettled(
+      s3Validators.map((v) => v?.getCheckpoint(match.index)),
+    );
+
+    let firstMatchingCheckpoint: Checkpoint | undefined;
+    const validatorInfos: ValidatorInfo[] = validators.map((address, index) => {
```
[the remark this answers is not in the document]
nambrot-agent: Done - now using `mapAllSettled` consistently throughout.

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -210,18 +210,23 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
...
 
+    // OPTIMIZATION: When we have messageContext, we only need to derive
+    // the specific ISM that will verify this message, not the full routing table.
+    // Just call route(message) and derive that single ISM directly.
+    if (this.messageContext) {
+      const routedIsmAddress = await abstractRoutingIsm.route(
+        this.messageContext.message,
+      );
+      return this.deriveIsmConfig(routedIsmAddress);
+    }
```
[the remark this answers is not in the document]
nambrot-agent: Fixed. When `messageContext` is present, we now use `${ism}:${messageContext.id}` as the cache key instead of just `ism`. This prevents message-specific configs from polluting the generic ISM cache.

typescript/sdk/src/ism/EvmIsmReader.ts
```diff
@@ -254,9 +259,7 @@ export class EvmIsmReader extends HyperlaneReader implements IsmReader {
 
     const defaultFallbackIsmInstance =
       DefaultFallbackRoutingIsm__factory.connect(address, this.provider);
-    const domainIds = this.messageContext
-      ? [BigNumber.from(this.messageContext.parsed.origin)]
-      : await defaultFallbackIsmInstance.domains();
```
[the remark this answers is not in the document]
nambrot-agent: This was intentional - the `messageContext` short-circuit at line 213-221 now returns early before reaching this code, so the `domainIds` optimization is no longer needed here.

typescript/relayer/src/metadata/arbL2ToL1.ts
```diff
@@ -45,26 +49,65 @@ export class ArbL2ToL1MetadataBuilder implements MetadataBuilder {
...
+          ...baseResult,
+          bridgeStatus: 'unconfirmed',
+          blocksRemaining,
+        };
+      }
+
+      if (errorMessage.includes('already been executed')) {
+        return {
+          ...baseResult,
+          bridgeStatus: 'executed',
```
nambrot-agent: No, good catch! Fixed to return `undefined` instead of `0` when the regex doesn't match.

nambrot-agent: To expand on my previous reply:

**Original issue:** When the regex `/Wait (\d+) blocks/` didn't match (e.g., if Arbitrum changes their error message format), the code was returning `blocksRemaining: 0`, which could be misleading - it would appear that the bridge is ready (0 blocks remaining) when we actually don't know the status.

**The fix (line 97-99):**
```typescript
const blocksRemaining = blocksMatch
  ? parseInt(blocksMatch[1], 10)
  : undefined;
```

Now returns `undefined` when the regex doesn't match, which correctly signals "we know it's unconfirmed but don't know exactly how many blocks remain."

I also added a comment (line 94-95) noting that this parsing depends on the error message format from `buildArbitrumBridgeCalldata`, so if the upstream error format changes, we'll gracefully degrade to `blocksRemaining: undefined` rather than returning an incorrect value.

nambrot: What do you mean?

typescript/relayer/src/metadata/multisig.ts
```diff
@@ -151,12 +174,118 @@ export class MultisigMetadataBuilder implements MetadataBuilder {
...
+    match: {
+      origin: number;
+      merkleTree: Address;
+      messageId: string;
+      index: number;
+    },
+  ): Promise<{
+    validatorInfos: ValidatorInfo[];
+    checkpoint?: Checkpoint;
+  }> {
```
nambrot-agent: You're right, I should reconsider this. Looking more closely, there's significant shared logic:

1. Both call `s3Validators()` to fetch validators
2. Both use `mapAllSettled` to fetch checkpoints
3. Both check if checkpoints match (merkleTree, messageId, index, origin)

**Option A: Extract shared checkpoint fetching**
```typescript
private async fetchValidatorCheckpoints(validators: Address[], match: {...}): Promise<{
  s3Validators: (S3Validator | undefined)[];
  fulfilled: Map<number, S3CheckpointWithId>;
  rejected: Map<number, Error>;
}> { ... }
```
Then both methods use this, with `getS3Checkpoints` filtering to matching checkpoints and `getValidatorInfos` building the status array.

**Option B: Make `getValidatorInfos` the primary method**
Have `getS3Checkpoints` call `getValidatorInfos` and extract just the signed validators' checkpoints:
```typescript
async getS3Checkpoints(...): Promise<S3CheckpointWithId[]> {
  const { validatorInfos } = await this.getValidatorInfos(validators, match);
  return validatorInfos
    .filter(v => v.status === 'signed')
    .map(v => /* reconstruct S3CheckpointWithId */);
}
```

The issue with Option B is that `ValidatorInfo` doesn't store the full `S3CheckpointWithId` (only the signature), so we'd need to change the type.

Happy to implement Option A if you think it's worth it. Let me know your preference.

paulbalaji:
## PR Review Summary

Nice feature! The structured `MetadataBuildResult` types are a solid improvement for Explorer integration. A few issues to address:

### P1 - JSON-RPC code 3 detection inconsistency (SmartProvider.ts)

In `getCombinedProviderError` (line ~615), you only check `e.error?.error?.code` for JSON-RPC code 3, but ethers sometimes exposes it at `e.error?.code`. This creates an inconsistency with `simplePerform` which checks both via `nestedError?.error?.code ?? nestedError?.code`.

**Fix**: Align the check:
```typescript
const jsonRpcErrorCode = e.error?.error?.code ?? e.error?.code;
const isJsonRpcRevert = jsonRpcErrorCode === 3;
```

Without this, CALL_EXCEPTION with code 3 at the shallow nesting level gets misclassified as recoverable, causing unnecessary retries of permanent reverts.

---

### P2 - Fragile error message parsing (arbL2ToL1.ts:415-422)

```typescript
if (errorMessage.includes('Wait') && errorMessage.includes('blocks')) {
  const blocksMatch = errorMessage.match(/Wait (\d+) blocks/);
```

Regex parsing error messages is brittle. If `buildArbitrumBridgeCalldata` changes wording, this silently breaks.

**Suggested fix**: Use typed errors:
```typescript
// In arbL2ToL1.ts
export class ArbL2ToL1PendingError extends Error {
  constructor(public readonly blocksRemaining: number) {
    super(`Wait ${blocksRemaining} blocks until challenge period`);
  }
}

export class ArbL2ToL1ExecutedError extends Error {
  constructor() {
    super('Arbitrum L2ToL1 message has already been executed');
  }
}

// Then in build():
} catch (error) {
  if (error instanceof ArbL2ToL1PendingError) {
    return { ...baseResult, bridgeStatus: 'unconfirmed', blocksRemaining: error.blocksRemaining };
  }
  if (error instanceof ArbL2ToL1ExecutedError) {
    return { ...baseResult, bridgeStatus: 'executed' };
  }
  throw error;
}
```

---

### P3 - Changeset package target

The changeset targets `@hyperlane-xyz/sdk` as major, but `MetadataBuilder` lives in `@hyperlane-xyz/relayer`. Should be:
```yaml
"@hyperlane-xyz/relayer": major
"@hyperlane-xyz/sdk": patch  # for EvmIsmReader/EvmHookReader/SmartProvider changes
```

---

### P4 - Use ValidatorStatus constant (multisig.ts:992)

```typescript
// Current
const signedValidators = validatorInfos.filter((v) => v.status === 'signed');

// Should be
const signedValidators = validatorInfos.filter((v) => v.status === ValidatorStatus.Signed);
```

---

LGTM once the above are addressed! The core design is solid. 🚀

.changeset/relayer-validator-status.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hyperlane-xyz/relayer": minor
```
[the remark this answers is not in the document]
nambrot: Fixed! Changed to `major` since this is a breaking change.

typescript/sdk/src/providers/SmartProvider/SmartProvider.ts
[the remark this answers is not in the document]
nambrot: I'll let you resolve this and get it merged if that is ok with you

PR #7801, opened 2026-01-15
feat(ci): add changeset validation to solidity analysis workflows

yorhodes:
## Summary

Add changeset validation to the Solidity bytecode, interface, and storage analysis CI workflows. This ensures developers include appropriate changesets when modifying contracts.

### Changeset Requirements

| Analysis | Change Type | Required Changeset |
|----------|-------------|-------------------|
| **Bytecode** | Any change | `patch` or higher |
| **Interface** | Addition (new functions, events, errors) | `minor` or higher |
| **Interface** | Removal or modification | `major` |
| **Storage** | Addition (new storage slots) | `minor` or higher |
| **Storage** | Removal | `major` |

### Changes

- Add `.github/scripts/check-solidity-changeset.sh` - validates `@hyperlane-xyz/core` changesets at required bump level
- Add `.github/scripts/check-diff-changeset.sh` - classifies diffs (additions/removals) and calls changeset validator
- Update `bytecode-analysis.yml` to use shared script
- Update `interface-analysis.yml` to check changeset based on change type
- Update `storage-analysis.yml` to use shared script and add skip condition for changeset-release PRs
- Update `solidity/interface.sh` to return exit code 2 for additions-only (distinguishing from removals)
- Update `solidity/README.md` with changeset requirements table

### Manual Testing

<details>
<summary>check-solidity-changeset.sh tests</summary>

```bash
# No changeset for @hyperlane-xyz/core
$ .github/scripts/check-solidity-changeset.sh patch
No changeset found for @hyperlane-xyz/core (required: patch or higher)

To fix this, run 'pnpm changeset' and select '@hyperlane-xyz/core' with a 'patch' (or higher) bump.
Exit code: 1

# With patch changeset
$ .github/scripts/check-solidity-changeset.sh patch
Found @hyperlane-xyz/core changeset with 'patch' bump (required: patch or higher)
Exit code: 0

$ .github/scripts/check-solidity-changeset.sh minor
Found @hyperlane-xyz/core changeset with 'patch' bump, but 'minor' or higher is required

To fix this, run 'pnpm changeset' and select '@hyperlane-xyz/core' with a 'minor' (or higher) bump.
Exit code: 1

# With minor changeset
$ .github/scripts/check-solidity-changeset.sh minor
Found @hyperlane-xyz/core changeset with 'minor' bump (required: minor or higher)
Exit code: 0

$ .github/scripts/check-solidity-changeset.sh major
[...]

PR #7808, opened 2026-01-15
fix(ci): fix solidity analysis workflow issues

yorhodes:
## Summary

Fixes issues with the solidity analysis CI workflows:

1. **Changeset CLI fails in shallow clones** - The `pnpm changeset status` command uses `git merge-base` internally, which fails in CI shallow clones (`fetch-depth: 1`) with:
   ```
   Error: Failed to find where HEAD diverged from "main". Does "main" exist?
   ```
   
   **Solution**: Parse `.changeset/*.md` files directly instead of using the CLI. This extracts the package bump levels from YAML frontmatter without requiring git history.

2. **Bytecode check fails with Foundry v1.5.0+** - The CI profile in `foundry.toml` sets `cbor_metadata = false` but was missing `bytecode_hash = "none"`. Newer Foundry versions require this combination.

## Root Cause

This is a known limitation of the changesets CLI documented in [changesets/changesets#700](https://github.com/changesets/changesets/issues/700) (open since December 2021). Multiple users have reported the same issue across GitHub Actions, Azure DevOps, and Bitbucket.

As one commenter noted:
> "Can someone explain to me why does the status actually need to work with GIT at all. Shouldn't it just be checking the existence of MD files..."

Since there's no upstream fix, parsing the changeset files directly is a reliable workaround that works in all CI environments.

## References

- Fixes issue introduced in #7801 (changeset validation)
- Fixes issue introduced in #7702 (bytecode analysis)
- Related upstream issue: https://github.com/changesets/changesets/issues/700

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal changeset validation process for build consistency.
  * Updated continuous integration build configuration settings.

### DIFF
--- a/.github/scripts/check-diff-changeset.sh
+++ b/.github/scripts/check-diff-changeset.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 # Analyzes a diff between two directories and validates changeset requirements
-# Usage: check-diff-changeset.sh <analysis-type> <base-dir> <head-dir>
+# Usage: check-diff-changeset.sh <analysis-type> <base-dir> <head-dir> <base-ref>
 # analysis-type: bytecode | storage
+# base-ref: git ref to compare changesets against (e.g., commit SHA)
 #
 # For bytecode: any change requires patch+
 # For storage: additions require minor+, removals require major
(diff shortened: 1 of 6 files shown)


yorhodes:
## Rationale for Direct File Parsing

This PR parses `.changeset/*.md` files directly instead of using `pnpm changeset status` because the changeset CLI has a known issue with shallow clones in CI environments.

**Related Issue:** https://github.com/changesets/changesets/issues/700

The CLI uses `git merge-base` internally to detect changed packages, which fails in shallow clones (like GitHub Actions with `fetch-depth: 1`). The error we were seeing:

```
Error: Failed to find where HEAD diverged from "main". Does "main" exist and it's synced with remote?
```

This issue has been open since December 2021 with no resolution. Multiple users have reported the same problem across different CI platforms (GitHub Actions, Azure DevOps, Bitbucket).

One commenter in the issue asked the same question we had:
> "Can someone explain to me why does the status actually need to work with GIT at all. Shouldn't it just be checking the existence of MD files..."

Given this is a known limitation of the changesets CLI, parsing the files directly is a reasonable workaround that works reliably in all CI environments.

PR #7820, opened 2026-01-16
fix(cli): suppress harmless startup warnings via pnpm patches

paulbalaji:
## Summary

- Suppressed the `bigint: Failed to load bindings` warning from bigint-buffer
- Suppressed the `DeprecationWarning: .data is not a valid RequestInit property` warning from node-fetch

Applied to all three bundled packages: **cli**, **rebalancer**, and **warp-monitor**.

## Approach

Uses **pnpm patches** to fix warnings at the source, rather than runtime suppression:

1. `patches/bigint-buffer@1.1.5.patch` - Silences the console.warn when native bindings fail to load
2. `patches/node-fetch@3.3.2.patch` - Removes the `.data` deprecation warning (gaxios compatibility)

This approach:
- Won't suppress future warnings that might be important
- Doesn't require `--no-warnings` shebang flag
- Doesn't need hacky `console.warn` overrides
- Uses the same patching mechanism already in use for other dependencies

## Details

Two harmless warnings were appearing during startup:

1. **bigint-buffer native bindings warning**
   - Source: `@solana/spl-token` → `@solana/buffer-layout-utils` → `bigint-buffer@1.1.5`
   - Cause: Native C++ bindings aren't compiled during install
   - Impact: None - falls back to pure JS implementation

2. **node-fetch DeprecationWarning**
   - Source: `@hyperlane-xyz/aleo-sdk` → `node-fetch@3.3.2`
   - Cause: gaxios passes `data` instead of `body` to fetch options
   - Impact: None - the code still works correctly

## Test plan

- [x] Run `hyperlane --help` and verify no warnings appear
- [x] Run rebalancer bundle and verify no warnings appear
- [x] Run warp-monitor bundle and verify no warnings appear

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Chores**
  * Applied source-level patches to silence harmless startup warnings so users no longer need runtime flags or console overrides to suppress them.
* **Bug Fixes**
  * Fixed CLI post-build error handling to fail cleanly instead of attempting on-the-fly file modifications.

### DIFF
--- a/typescript/cli/scripts/ncc.post-bundle.mjs
+++ b/typescript/cli/scripts/ncc.post-bundle.mjs
@@ -53,12 +53,7 @@ async function patchCliExecutable() {
     );
   } catch (error) {
     console.error('✖ Failed to patch CLI executable:', error);
-    if (!content.includes(dirnameDef)) {
-      const [, executable] = content.split(shebang);
-      const newContent = `${shebang}\n${dirnameDef}\n${executable}`;
-      await writeFile(outputFile, newContent, 'utf8');
-      console.log('Adding missing __dirname definition to cli executable');
-    }
+    process.exit(1);
   }
 }
 
--- a/.changeset/fix-cli-post-bundle-error-handler.md
+++ b/.changeset/fix-cli-post-bundle-error-handler.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/cli': patch
+---
+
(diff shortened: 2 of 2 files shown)


paulbalaji (commented):
nuke env.ts + extraneous comments

typescript/cli/scripts/ncc.post-bundle.mjs
paulbalaji: this was only broken in the cli, rebalancer/warpmonitor bundle scripts had the right implementation

PR #7834, opened 2026-01-20
feat(aleo-sdk): aleo e2e testing setup refactor

xeno097:
### Description

This PR adds a `/testing` export to the `aleo-sdk` so that client packages like the cli can implement their own set of e2e tests for Aleo code.

Note:
- There will be a follow up PR implementing the missing cli e2e tests for Aleo on the cli side

### Drive-by changes

- Removed docker compose file for starting the aleo test server as test containers takes care of it

### Backward compatibility

- YES

### Testing

- E2E tests from the sdk

## Summary by CodeRabbit

* **New Features**
  * Added a public testing module providing Aleo test utilities, chain metadata, node management, and signer helpers.

* **Chores**
  * Switched test infrastructure to automated container lifecycle management (testcontainers) and updated test runner/setup to use a global setup.
  * Added a testing export entrypoint and adjusted E2E environment handling.

* **Tests**
  * All 54 end-to-end tests pass with the new setup.

### DIFF
--- a/typescript/aleo-sdk/.mocharc-e2e.json
+++ b/typescript/aleo-sdk/.mocharc-e2e.json
@@ -1,6 +1,5 @@
 {
   "extensions": ["ts"],
   "file": ["src/tests/e2e-test.setup.ts"],
-  "spec": ["src/tests/*.e2e-test.ts"],
   "node-option": ["import=tsx/esm"]
 }
--- a/typescript/aleo-sdk/scripts/run-e2e-test.sh
+++ b/typescript/aleo-sdk/scripts/run-e2e-test.sh
@@ -13,9 +13,9 @@ export ALEO_WARP_SUFFIX="usdc"
 
 if [ -n "${ALEO_SDK_E2E_TEST}" ]; then
   echo "Running only ${ALEO_SDK_E2E_TEST} test"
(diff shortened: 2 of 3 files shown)


typescript/aleo-sdk/package.json
```diff
@@ -49,6 +50,7 @@
     "mocha-steps": "catalog:",
     "node-fetch": "^3.3.2",
     "prettier": "catalog:",
+    "testcontainers": "catalog:",
```
paulbalaji: codex says

> • The new exported testing entry point relies on a runtime dependency that is only declared as a dev dependency, which will break downstream consumers of the
  new module unless they add the dependency themselves.
>
>  Review comment:
>
>  - [P2] Declare testcontainers as runtime dependency — typescript/aleo-sdk/package.json:51-53
>    The new @hyperlane-xyz/aleo-sdk/testing entry point imports testcontainers at runtime (via runAleoNode), but testcontainers is only listed under
>    devDependencies. Consumers installing the package from npm won’t get dev deps, so importing the testing module will fail with “Cannot find module
>    'testcontainers'” unless they manually add it. Consider moving it to dependencies (or a peer dependency) so the new public testing module works out of the
>    box.

PR #7856, opened 2026-01-21
Solana IDL generation

antigremlin:
Here is the first step of the Solana to Multi-VM path (soon to be 2-3 staps):

1. Generate Anchor-compatible IDL form our program source, using Shank (works)
2. Generate TypeScript clients for our programs from that IDL using Codama (doens't yet work)
3. Use the generated TS clients and the existing TS clients for Loader-v3, Token-2022, etc. to port everything and build our brave new world (could maybe work next week).

Never mind the typescript/sdk part just yet, it's undercooked.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2308,9 +2308,6 @@ importers:
       '@safe-global/safe-deployments':
         specifier: 'catalog:'
         version: 1.37.47
-      '@solana/kit':
-        specifier: ^2.3.0
-        version: 2.3.0(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.8.3)(ws@8.18.3(bufferutil@4.0.9)(utf-8-validate@5.0.10))
       '@solana/spl-token':
         specifier: 'catalog:'
         version: 0.4.14(@solana/web3.js@1.98.4(bufferutil@4.0.9)(typescript@5.8.3)(utf-8-validate@5.0.10))(bufferutil@4.0.9)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.8.3)(utf-8-validate@5.0.10)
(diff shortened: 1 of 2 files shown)


rust/sealevel/programs/helloworld/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use hyperlane_core::H256;
+use shank::ShankType;
+
+/// Configuration for a remote router.
+/// This is a wrapper around hyperlane_sealevel_connection_client::router::RemoteRouterConfig
+/// to include in the IDL.
+#[derive(Debug, Clone, PartialEq, BorshDeserialize, BorshSerialize, ShankType)]
+pub struct RemoteRouterConfig {
```
antigremlin: The biggest problem with Shank is that it doesn't see it's derive macros applied to libraries (dependency crates). It only scans and uses the source of the curren crate, and so we need all types to be local

rust/sealevel/client/src/helloworld.rs
```diff
@@ -75,7 +75,10 @@ impl RouterDeployer<HelloWorldConfig> for HelloWorldDeployer {
         payer: Pubkey,
         router_configs: Vec<RemoteRouterConfig>,
     ) -> Instruction {
-        enroll_remote_routers_instruction(program_id, payer, router_configs).unwrap()
```
antigremlin: If we bring the shared types closer to contracts, we'd have to do type-conversion in the client code. This allows to avoid changes to on-chain code.

rust/sealevel/programs/helloworld/src/types.rs
antigremlin: There are potential workaround:
- bring the type into the contract code, then make conversions in the off-chain code
- extend Shank to be able to import IDL from other crates, and this way avoid all the changes we're making here.

PR #7923, opened 2026-01-28
fix(scraper): Optimize message_view query performance

ameten:
## Summary
- Added index on `message.destination` column for efficient filtering
- Replaced `LEFT JOIN total_gas_payment` with `LEFT JOIN LATERAL` to avoid aggregating all 12M gas_payment rows

## Problem
Queries on `message_view` filtered by `destination_domain_id` were timing out due to:
1. Full table scan on `message` table (no index on `destination`)
2. `total_gas_payment` view aggregating entire `gas_payment` table (12M rows) before joining

## Results
Query cost reduced from ~2.3M to ~12K (~193x improvement).

## Test plan
- [x] Verified execution plan shows index usage and LATERAL join
- [x] Confirmed query results are identical between old and new view
- [ ] Deploy to testnet scraper and verify query performance

## Summary by CodeRabbit

* **Chores**
  * Optimized message destination queries with enhanced database indexing for improved performance.
  * Refined gas payment aggregation logic for more accurate data tracking across messages.


paulbalaji:
@claude review

PR #7988, opened 2026-02-01
fix(utils): use .bind() for AbortController.abort, remove dead SDK duplicate

paulbalaji:
## Summary

- Replaced arrow wrapper `() => controller.abort()` with `controller.abort.bind(controller)` in `fetchWithTimeout`
- Deleted duplicate `fetchWithTimeout` from SDK (dead code - never imported)

## Why

An arrow function like `() => controller.abort()` captures the surrounding scope in its closure, which includes the request body and other large objects. When passing this to a long-lived `AbortSignal`, the listener prevents those objects from being GC'd for the lifetime of the signal. Using `.bind()` only retains a reference to the controller itself.

This pattern was highlighted as a 1 GB memory reduction for long Claude Code sessions:
https://x.com/jarredsumner/status/2017825694731145388

## Changes

| File | Change |
|------|--------|
| `typescript/utils/src/async.ts` | `setTimeout(controller.abort.bind(controller), timeout)` |
| `typescript/sdk/src/utils/fetch.ts` | **Deleted** (dead code, duplicate of utils version) |

## Testing

- Behavior unchanged; only the closure capture is affected
- No new tests needed (memory improvement, not functional change)
- SDK's `fetchWithTimeout` was never used (verified via grep)

## Summary by CodeRabbit

* **Refactor**
  * Enhanced timeout handling mechanism in network utilities to improve memory efficiency by preventing unnecessary retention of large objects during timeout operations.

* **Chores**
  * Removed duplicate fetch utility code from SDK utilities, reducing code redundancy and maintenance overhead across packages.


antigremlin (approved):
Nice!

PR #8002, opened 2026-02-03
refactor: artifact type utilities

antigremlin:
### Description

Test the type utilities for `Artifact<>` transforms. Support optional properties but not arrays of optional artifacts (seems unnecessary).


paulbalaji (commented):
changeset?

PR #8059, opened 2026-02-06
feat: tron hook artifact manager

troykessler:
### Description

This PR implements the hook artifact manager for Tron

## Release Notes

* **New Features**
  * Implemented full Tron hook artifact manager with complete deployment and management capabilities.
  * Added support for Interchain Gas Paymaster (IGP) hooks on Tron, enabling on-chain configuration management and updates.
  * Added support for Merkle Tree hooks on Tron with deployment and configuration reading capabilities.

### DIFF
--- a/typescript/tron-sdk/src/hook/igp-hook.ts
+++ b/typescript/tron-sdk/src/hook/igp-hook.ts
@@ -1,0 +1,322 @@
+import { AltVM } from '@hyperlane-xyz/provider-sdk';
+import {
+  type ArtifactDeployed,
+  type ArtifactNew,
+  type ArtifactReader,
+  ArtifactState,
+  type ArtifactWriter,
+} from '@hyperlane-xyz/provider-sdk/artifact';
+import {
+  type DeployedHookAddress,
+  type IgpHookConfig,
+} from '@hyperlane-xyz/provider-sdk/hook';
+import { AnnotatedTx } from '@hyperlane-xyz/provider-sdk/module';
+import { eqAddress } from '@hyperlane-xyz/utils';
+
+import { type TronSigner } from '../clients/signer.js';
+import { TronReceipt } from '../utils/types.js';
+
+import { type TronHookQueryClient, getIgpHookConfig } from './hook-query.js';
+import {
+  getCreateIgpTx,
+  getCreateOracleTx,
+  getCreateProxyTx,
+  getInitIgpTx,
+  getSetIgpDestinationGasConfigTx,
+  getSetIgpOwnerTx,
+  getSetOracleTx,
+  getSetRemoteGasTx,
+} from './hook-tx.js';
+
+/**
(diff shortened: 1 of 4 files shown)


typescript/tron-sdk/src/hook/igp-hook.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    return {
+      artifactState: ArtifactState.DEPLOYED,
+      config: {
+        type: AltVM.HookType.INTERCHAIN_GAS_PAYMASTER,
+        owner: hookConfig.owner,
+        // Tron IGP doesn't have beneficiary and oracleKey in the same way as EVM
+        // Setting them to owner as a placeholder
+        beneficiary: hookConfig.owner,
+        oracleKey: hookConfig.owner,
```
xeno097: Isn't tron using the same contracts as ethereum? We should have these available fro reading+setting in this case

typescript/tron-sdk/src/hook/igp-hook.ts
```diff
@@ -0,0 +1,322 @@
...
+        })),
+      },
+    );
+
+    const configReceipt =
+      await this.signer.sendAndConfirmTransaction(setConfigTx);
+    receipts.push(configReceipt);
+
+    // Transfer ownership if needed (deployer is initial owner)
+    if (!eqAddress(artifact.config.owner, deployerAddress)) {
```
xeno097: I think you should add a note here for later to use the specific tron function for comparing addresses as we don't have it now

typescript/tron-sdk/src/hook/igp-hook.ts
```diff
@@ -0,0 +1,322 @@
...
+          ...setConfigTx,
+        });
+      }
+    }
+
+    // Check if owner needs to be updated
+    if (!eqAddress(currentState.config.owner, config.owner)) {
+      const setOwnerTx = await getSetIgpOwnerTx(
+        this.signer.getTronweb(),
+        this.signer.getSignerAddress(),
```
xeno097: shouldn't this be the current owner addy based on how altvm generates the update txs?

typescript/tron-sdk/src/hook/igp-hook.ts
```diff
@@ -0,0 +1,322 @@
...
+    receipts.push(gasReceipt);
+
+    const setConfigTx = await getSetIgpDestinationGasConfigTx(
+      this.signer.getTronweb(),
+      deployerAddress,
+      {
+        igpAddress,
+        destinationGasConfigs: Object.keys(config.overhead).map((domainId) => ({
+          remoteDomainId: parseInt(domainId),
+          gasOverhead: config.overhead[+domainId]?.toString() ?? '0',
```
xeno097: instead of `+domainId` can't we parse int here too like it is done for the `remoteDomainId` field?

xeno097 (commented):
Changeset and e2e tests are missing here too

PR #8072, opened 2026-02-09
feat(aleo): local proving fallback

yjamin:
### Description
Allows the relayer to fallback to local proving in case the proving service is down or we encounter some issue on the way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Proving now tries a configured remote proving service first; if that fails it automatically falls back to local proving and emits a warning, improving reliability and visibility.


rust/main/chains/hyperlane-aleo/src/provider/aleo.rs
```diff
@@ -468,9 +468,30 @@ impl<C: AleoClient> AleoProvider<C> {
...
         let transaction = match self.proving_service {
-            Some(ref client) => client.proving_request(authorization, fee).await,
+            Some(ref client) => {
+                match client
+                    .proving_request(authorization.clone(), fee.clone())
+                    .await
+                {
+                    Ok(tx) => Ok::<_, ChainCommunicationError>(tx),
+                    Err(e) => {
+                        debug!(
```
tkporter: Agree with Claude, I'd probably make this a warn

PR #8099, opened 2026-02-15
chore(keyfunder): improve error logging for failed chain funding

tkporter:
## Summary

Key funder error logs were difficult to diagnose. When citrea funding failed with `UNPREDICTABLE_GAS_LIMIT`, the root cause was simply that the funder address had insufficient balance (0.0000189 CBTC) to cover the required amount (0.0000565 CBTC) — but the error gave no indication of this. Additionally, the summary log said "2/126 chains failed" without naming which chains failed, and the second failure (likely a timeout) was never logged at all.

Changes:
- **Insufficient balance pre-check**: Before sending a funding tx, compare funder balance to the required amount. If insufficient, log a clear error with both values and throw — instead of letting `eth_estimateGas` fail with an opaque `UNPREDICTABLE_GAS_LIMIT`
- **Log timeout failures**: Added catch block in `fundChainWithTimeout` so timeout errors are logged at ERROR level. Previously, timeouts were silently counted in the failure total but never logged (the catch block was only in `fundChain`, which the timeout wrapper bypasses)
- **Failed chain names in summary**: The "Some chains failed to fund" error now includes each failed chain name and its error reason as structured log data, instead of just a count

## Test plan

- [x] `pnpm -C typescript/keyfunder build` passes
- [x] `pnpm -C typescript/keyfunder test` passes (47 tests)
- [x] Tested end-to-end with deployed image

### DIFF
--- a/typescript/keyfunder/src/core/KeyFunder.ts
+++ b/typescript/keyfunder/src/core/KeyFunder.ts
@@ -39,14 +39,24 @@ export class KeyFunder {
       chains.map((chain) => this.fundChainWithTimeout(chain)),
     );
 
-    const failures = results.filter((r) => r.status === 'rejected');
+    const failures = results
+      .map((r, i) => ({ result: r, chain: chains[i] }))
+      .filter(
+        (entry): entry is { result: PromiseRejectedResult; chain: string } =>
+          entry.result.status === 'rejected',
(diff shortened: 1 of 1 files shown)


paulbalaji (approved):
Thx

PR #8102, opened 2026-02-16
chore: disable flowmmainet relaying & validator

yjamin:


typescript/infra/config/docker.ts
```diff
@@ -39,8 +39,8 @@ interface MainnetDockerTags extends BaseDockerTags {
 
 export const mainnetDockerTags: MainnetDockerTags = {
   // rust agents
-  relayer: 'f15d8c7-20260206-060304',
-  relayerRC: 'f15d8c7-20260206-060304',
+  relayer: '5302a89-20260215-172420',
```
paulbalaji: What tag is this

PR #8109, opened 2026-02-16
feat(rebalancer): [1/6] Config + IExternalBridge + LiFi bridge + gas utilities

Mo-Hussain:
## [1/6] Config + IExternalBridge + LiFi bridge + gas utilities

This PR is part of the inventory rebalancer PR stack. It adds foundational components:

- Config types for inventory rebalancing (ExecutionType, ExternalBridgeType)
- IExternalBridge interface for external bridge integrations
- LiFi bridge implementation
- Gas estimation utilities

**Base:** feat/inventory-rebalancer
**Next:** feat/ir-02-tracking (PR 2)

### Files Added
- IExternalBridge.ts, LiFiBridge.ts, bridges/index.ts
- gasEstimation.ts, lifiMocks.ts

### Testing
- All existing tests pass
- New config validation tests added


typescript/rebalancer/package.json
paulbalaji: formatting changes?

typescript/rebalancer/src/utils/index.ts
paulbalaji: i probs missed this in the previous stack of reviews but we should avoid * exports in the monorepo entirely, and we should avoid having index.ts for non-root dirs

so if we really need some of this stuff outside of the rebalancer package then we can export it otherwise we shouldn't export

PR #8243, opened 2026-02-27
feat(lander): bound EVM reorg detection and add manual reprocess APIs (ENG-3399)

ameten:
## Summary
- bound automatic EVM reorg tx reprocessing to a fixed max depth of 25 nonces
- skip automatic reprocessing entirely for oversized reorgs and require manual intervention
- persist oversized reorg nonce range for inspection and explicit manual trigger
- add relayer API endpoints:
  - `POST /evm/inspect_reorged_transactions`
  - `POST /evm/reprocess_reorged_transactions`
- add new Lander metrics for ops/alerting:
  - `hyperlane_lander_reorged_transactions`
  - `hyperlane_lander_reorg_manual_intervention_required`
- add focused tests for small vs oversized reorg behavior and manual trigger flow
- document alert query for manual intervention

## Testing
- `cd rust/main && cargo test -p lander tests_reprocess`
- `cd rust/main && cargo test -p relayer --no-run`

### DIFF
--- a/docs/ai-agents/operational-debugging.md
+++ b/docs/ai-agents/operational-debugging.md
@@ -163,6 +163,16 @@ hyperlane_lander_gas_price{destination="linea"} / 1000000000
 rate(hyperlane_lander_finalized_transactions[5m])
 ```
 
+**Alert for manual Lander reorg intervention (new):**
+
+```promql
+max by (destination, signer) (
+  hyperlane_lander_reorg_manual_intervention_required
+) > 0
+```
+
+Set `for: 5m` and page the oncall. This indicates oversized nonce rollback was detected and automatic reprocessing was intentionally skipped.
+
 ### Debugging Workflow with Grafana + GCP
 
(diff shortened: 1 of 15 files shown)


paulbalaji:
Closing as superseded after comparing the exact implementation with #9219 and #9223.

#9219 fixes the underlying correctness race by persisting and merging the signer-scoped reorg range before any finalized-boundary update can consume it. The range survives restart and is re-enqueued idempotently until the finalized nonce recovers. #9223 then removes the hot fixed polling loop while waking promptly on local activity.

I am deliberately not porting this PR's 25-nonce manual gate: its pending range is in-memory, it stops automatic recovery until an operator calls a one-shot API, and it predates the competing-boundary failure mode. If deep-range DB scanning needs a bound, it should be chunked on top of #9219's durable range rather than reintroducing a lossy/manual ownership path.

Thanks @ameten — the bounded-work concern and tests helped expose the residual, but the newer stack is the safer replacement.

PR #8347, opened 2026-03-13
feat(rebalancer): add LayerZero USDT bridge

Mo-Hussain:
### Description

Adds a LayerZero `IExternalBridge` for USDT inventory movement across the currently tested EVM, Tron, and Solana paths.

**Adapter only. No route config change or deployment.** This PR is stacked on #8459 for the shared bridge lifecycle; it does not depend on deBridge at runtime.

### Scope

- Native OFT, Legacy Mesh, and composed LayerZero paths
- Ethereum, Arbitrum, Plasma, Tron, and Solana route primitives currently covered by this implementation
- EVM, Tron, and Solana execution
- LayerZero Scan settlement polling, including destination and compose completion
- Strict signer, token, chain, spender, amount, and transaction validation
- Config schema and factory registration

Celo and TON are deliberately not enabled here. LayerZero now documents Celo in the Legacy Mesh, but enabling it should be a separate tested change.

### Verified

- Rebalancer build passed
- Full rebalancer suite: 486 passing
- Focused LayerZero/Solana suite: 74 passing
- Standard and type-aware lint passed (3 pre-existing assertion warnings)
- Commit hooks and gitleaks passed
- Independent security review found and fixed Solana Scan signature encoding and retry-unsafe terminal-state classification; no remaining findings

### Residual validation

A funded/forked send test for every VM/compose path remains a deployment-stage check. This PR does not deploy anything.


typescript/rebalancer/src/bridges/LayerZeroBridge.ts
```diff
@@ -0,0 +1,1033 @@
...
+    const oft = new ethers.Contract(route.oftContract, OFT_ABI, wallet);
+    const tx = await oft.send(
+      route.sendParam,
+      route.messagingFee,
+      wallet.address,
+      {
+        value: route.messagingFee.nativeFee,
+      },
+    );
+    await tx.wait();
```
paulbalaji, line 655: [P1] Preserve the submitted source identity before waiting for confirmation. If `oft.send()` broadcasts successfully but `tx.wait()` throws on a timeout/RPC failure, this method throws without returning the hash. `InventoryRebalancer` creates the tracked action only after `execute()` returns, so the transfer can remain untracked and a later cycle can send again. The new Solana and Tron send-and-confirm paths have the same boundary. Integrate the current parent's submission/approval observations and typed outcomes, and return/reconcile the source identity independently of confirmation. Add broadcast-success/confirmation-error tests for each VM; a receipt deadline alone does not solve the lost identity.

typescript/rebalancer/src/bridges/LayerZeroBridge.ts
```diff
@@ -0,0 +1,1033 @@
...
+        route.sendParam.amountLD === quote.fromAmount,
+      'LayerZero quote transfer parameters mismatch',
+    );
+  }
+
+  private matchesBridgeToken(
+    actualToken: string,
+    expectedToken: string,
+  ): boolean {
+    if (/^0x/i.test(actualToken) && /^0x/i.test(expectedToken)) {
```
paulbalaji, line 1027: [P2] Normalize Tron token addresses before comparison. `getUSDTAddress(TRON_CHAIN_ID)` is `0xa614...`, while the same valid USDT token can be supplied as `TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t`. The caller forwards `collateralAddressOrDenom` unchanged, and this helper falls back to literal equality, so normal base58 Tron source/destination token inputs fail `assertUsdtRoute()` before quoting. Use the existing Tron address conversion utility and test both representations on both route ends; the current tests use only the hardcoded hex form.

typescript/rebalancer/src/bridges/LayerZeroBridge.ts
```diff
@@ -0,0 +1,1033 @@
...
+  ): Promise<Response> {
+    let lastError: Error | undefined;
+    for (let attempt = 0; attempt < retries; attempt++) {
+      if (attempt > 0) {
+        await new Promise((resolve) =>
+          setTimeout(resolve, 1000 * Math.pow(2, attempt - 1)),
+        );
+      }
+      try {
+        const response = await fetch(url, options);
```
paulbalaji, line 918: [P2] Bound the Scan request and body read. This fetch has no deadline, so a request that never resolves—or returns headers and stalls the JSON body—never reaches the retry loop. Since status reconciliation awaits this call, one stalled Scan response can block subsequent tracked actions. Use an abortable request budget that stays active through `response.json()` and add stalled-header/stalled-body tests.

paulbalaji (commented):
Reviewed authored LayerZero changes at `8b79de430bd8ed155e1c0a950d39c0d3c09e2e1a`, surrounding callers, and integration with current #8459 head `db1951420ba874bffe650826a7b95fa9ce9cc854`. Three correctness/availability findings remain inline.

Restacking is required: the current deBridge head is not an ancestor, and a synthetic merge conflicts in 16 files, including shared approval/submission interfaces and deBridge code. The green August 26 checks validate the old tree, not the current combined stack. Rebase, adapt LayerZero to the new lifecycle helpers, and run fresh combined-stack CI before merging.

Doable performance gains:

- `quoteComposeFromSolana()` calls full `quoteSolanaTransfer()` twice. The first call needs only the OFT received amount but also resolves messaging accounts and simulates `quoteSend`, whose fee is discarded. Split the amount prequote from fee quoting and share resolved accounts within this operation.
- `executeTron()` unconditionally resets allowance to zero and then approves. Read allowance first; skip the zero-reset when already zero and skip approval when the intended target already matches. Reuse the shared Tron approval helper after restacking.
- Reuse read-only chain clients and batch Solana account reads; these have more impact than small serializer allocations.

Muse review completed and candidates were source-checked. Parent/deBridge focused tests passed separately; LayerZero tests were not rerun locally because the available dependency tree lacks its LayerZero SDK packages. Funded/forked sends across the VM/compose paths remain unverified. Existing address/ABI reuse discussions remain open.

Current-head CI readback: 53 successful checks, 31 skipped, but these are the old-base checks described above.

typescript/rebalancer/src/bridges/LayerZeroBridge.ts
```diff
@@ -0,0 +1,1033 @@
...
+        usdtContractTron,
+        'approve(address,uint256)',
+        {},
+        [
+          { type: 'address', value: oftContractTron },
+          { type: 'uint256', value: '0' },
+        ],
+        signerAddress,
+      );
+    await tronSigner.sendAndConfirmTransaction(resetApprovalTx);
```
paulbalaji, line 709: [P3, optional performance] Skip this reset transaction when the current allowance is already zero. The path unconditionally sends and confirms approve(0), including a first transfer and the usual case where the previous allowance was fully consumed. Reading allowance and reusing the observed exact-approval helper through TronWallet avoids one transaction, its Tron energy cost, and a confirmation roundtrip while preserving USDT’s zero-before-nonzero rule.

paulbalaji (commented):
Reviewed own diff at `8b79de430bd8ed155e1c0a950d39c0d3c09e2e1a` as part of stacks 9635 and 9636 (14 PRs).

The existing source-identity loss across EVM/Tron/Solana confirmation failures, unbounded Scan request/body, and base58 Tron USDT comparison findings remain. This branch also does not contain the current parent lifecycle hardening: a merge-tree against current #8459 reports 16 conflicting files. Rebase and reconcile those conflicts before relying on combined-stack validation. Whole-head differences alone are not evidence that a future merge will revert fixes. Added one optional inline transaction-cost improvement.

Reviewed this branch separately against its actual ancestry. The 830 combined-stack tests reported on other PRs exclude this divergent LayerZero branch; no per-head runtime test claim here.

Latest check-name view: 51 pass, 31 skipping. Independent Muse, GLM and Kimi review output was reconciled with source and prior threads; unverified model suggestions were excluded.

Issue #1030, opened 2022-09-02
Migrate to Gelato relay API v2


tkporter:
From conversations with the Gelato team, the anticipated scope of changes is small. My understanding so far is:
* Different API base url
* Some names of things are different (iiuc ForwardRequest has been renamed)
* Gas limit is no longer required (we should be sure to remove our hardcoded 5M max that we've put for now

Concretely:
* Update our code to use the new Gelato API once we can
* Test it thoroughly and ship it to all contexts on testnet2

tkporter:
done

Issue #301, opened 2022-03-28
Use interchain gas payment SDK in bridge SDK


tkporter:
Following https://github.com/abacus-network/abacus-monorepo/issues/251, the governance SDK should make use of the generic tooling to properly make interchain gas payments

tkporter:
satisfied by https://github.com/abacus-network/abacus-bridge/pull/4

PR #8433, opened 2026-03-25
perf: expose lighter sdk/widgets package surfaces

paulbalaji:
## Summary
This PR is the shared package-surface cleanup behind:
- Warp UI `#1030`
- Explorer `#301`

Goal: keep the monorepo changes to the surfaces that actually helped downstream UIs, and trim the speculative churn back out.

Main themes:
- split heavy VM package roots from runtime-only entrypoints
- separate light provider-registry behavior from the legacy `MultiProvider` bridge
- expose light SDK read helpers that apps actually use
- keep widget wallet hooks protocol-bound and avoid unnecessary runtime pulls

### 1. Runtime-only VM entrypoints
- added `./runtime` entrypoints for heavy VM packages used by UI/runtime codepaths
- switched shared SDK/provider code onto those runtime entrypoints so UI consumers stop pulling deploy/typechain baggage by default

### 2. Provider layers split by responsibility
- `MinimalProviderRegistry`: metadata + provider registry behavior only
- `MultiProviderAdapter`: legacy `MultiProvider` bridge (`fromMultiProvider`, `toMultiProvider`)
- `MultiProtocolProvider`: default-builder convenience layer on top
- provider-builder maps split from individual builder exports so narrower consumers do not pull the whole registry by default

### 3. SDK helper surfaces kept because downstream apps used them
- `@hyperlane-xyz/sdk/metadata/*`
- `@hyperlane-xyz/sdk/providers/*`
- `@hyperlane-xyz/sdk/providers/builders/*`
- `@hyperlane-xyz/sdk/token/*`
- `@hyperlane-xyz/sdk/warp/*`
- `@hyperlane-xyz/sdk/types`

Notably kept because they now have real downstream callers:
- `ChainMetadataResolver`
- `warp/read`
- `MinimalProviderRegistry`
- `MultiProviderAdapter`

### 4. Widget wallet/runtime cleanup
- cut over wallet integrations to canonical protocol-bound `*Wallet` entrypoints
- narrowed several widget surfaces onto `MinimalProviderRegistry` / `ITokenMetadata`
- `ChainAddMenu` no longer constructs `MultiProtocolProvider` just to validate duplicate chain metadata
- lazy-load the EVM watch-asset helper inside the callback so `_app` stops pulling `@hyperlane-xyz/core`

### 5. Kept trims / removed speculative surfaces
- removed the graph-first public warp surface
- removed `sdk ./consts/*`
- removed `widgets ./animations/*`
- removed dead root SDK exports and unused runtime helper scaffolding
- dropped unused `./core/*` exports again after downstream experiments showed they were not earning their keep in this PR

### 6. Follow-up review fixes now included
- tightened `ChainMetadataResolver` lookups/types (`null`-only metadata, domain-id lookup via resolver, duplicate `chainId` aliases now resolve to `null` instead of throwing)
- preserved typed-provider discriminants for zkSync and gnosis-builder paths
- fixed Radix wallet hook side effects/stability/error handling
- added required `// CAST:` comments and a few low-cost API hygiene fixes

## Why
Before this PR, downstream apps kept rebuilding local helper code because monorepo surfaces were still too coarse:
- VM package roots mixed runtime with deploy/typechain baggage
- root SDK/widgets imports made it easy to drag all-VM code into hot paths
- light provider-backed consumers still paid for the old `MultiProvider` shape
- widgets reintroduced one concrete EVM watch-asset regression that polluted Warp UI `_app`

This PR keeps the boundaries that paid off downstream and trims the ones that did not.

## Downstream impact
This PR is mainly an enabling/shared-layer change. The clearest signal is downstream:
- Warp UI `#1030` uses these lighter surfaces on the reviewable branch
- Explorer `#301` uses them on the perf branch

One concrete monorepo-side fix included here:
- lazy EVM watch-asset loading in widgets, which materially reduced Warp UI `_app` pollution on the downstream branch

## Validation
Ran on current branch head:
- [x] `pnpm build`
- [x] `pnpm --filter @hyperlane-xyz/sdk build`
- [x] `pnpm --filter @hyperlane-xyz/widgets build`
- [x] `pnpm --filter @hyperlane-xyz/sdk exec mocha --config .mocharc.json ./src/providers/MultiProtocolProvider.test.ts ./src/metadata/ChainMetadataResolver.test.ts --exit`
- [x] `pnpm exec oxlint -c oxlint.json <touched files>`

Also validated on latest `origin/main` rebase:
- targeted SDK mocha suites for token / warp / metadata paths
- root rebase + lockfile refresh
- fresh root `pnpm build` after rebase
- fresh `pnpm --filter @hyperlane-xyz/sdk build` after rebase
- fresh `pnpm --filter @hyperlane-xyz/widgets build` after rebase

## Beta
Do not publish manually from this branch.
Use `.github/workflows/release.yml` via `workflow_dispatch` on this branch with `snapshot_tag=beta`.

Latest published beta from this branch:
- `@hyperlane-xyz/sdk@30.0.0-beta.3dc12d8cfe92624f4d882a719036a95e3abf3fab`
- matching `utils` / `widgets` / VM SDK packages

Current branch head:
- `698339d1ee`

That means the latest review-fix commits after the rebase/beta cut, including the duplicate `chainId` resolver fix, are **not** in a published beta yet.

## Status
- all currently open review threads addressed
- `0` unresolved review threads at time of update

## Remaining work
- this PR does not finish app perf by itself; it gives apps cleaner package surfaces
- biggest remaining Warp UI cost is still app bootstrap / default-route runtime exposure
- further monorepo work should come from concrete downstream traces, not more speculative surface splitting

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8433" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes public package export maps and type surfaces (`MultiProtocolProvider` → `MultiProviderAdapter`/`MinimalProviderRegistry`), which can break downstream imports and bundling assumptions. Runtime-only entrypoints for non-EVM SDKs also affect module resolution and build output if misconfigured.
> 
> **Overview**
> **Reduces UI/runtime bundle weight by reshaping public package surfaces.** The PR adds new `./runtime` entrypoints for non-EVM SDK packages (Aleo/Cosmos/Radix) and updates their `package.json` `exports` plus new `src/runtime.ts` re-export files.
> 
> **Splits/leans on provider interfaces via subpath exports.** Several consumers are updated to depend on the lighter provider abstractions (`MultiProviderAdapter` and `MinimalProviderRegistry`) instead of `MultiProtocolProvider`, and the SDK README is updated to reflect the new provider-layer model.
> 
> Also includes a small robustness fix in `solidity/build-tron.sh` to only restore backed-up files when they exist, plus lockfile updates (notably removing `@tronweb3/tronwallet-adapters` and bumping `express-rate-limit`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c51f23535f7894bd74741f5d9a35760aaee33028. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


typescript/sdk/src/token/Token.ts
```diff
@@ -72,78 +37,55 @@ import type {
...
-      addressOrDenom: nativeToken.denom ?? '',
-      decimals: nativeToken.decimals,
-      symbol: nativeToken.symbol,
-      name: nativeToken.name,
-      logoURI,
-      coinGeckoId: gasCurrencyCoinGeckoId,
-    });
+  override removeConnection(token: IToken): Token {
+    const index = this.connections?.findIndex((t) => t.token.equals(token));
+    if (index && index >= 0) this.connections?.splice(index, 1);
```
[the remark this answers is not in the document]
paulbalaji: Fixed. `Token.removeConnection` now delegates to the base implementation, so the first connection is removed correctly too.

typescript/radix-sdk/src/runtime.ts
```diff
@@ -0,0 +1,12 @@
+export { RadixProtocolProvider } from './clients/protocol.js';
+
+export { NetworkId } from '@radixdlt/radix-engine-toolkit';
+
+export { RadixProvider } from './clients/provider.js';
+export { RadixSigner } from './clients/signer.js';
+
+export { RadixSDKReceipt, RadixSDKTransaction } from './utils/types.js';
```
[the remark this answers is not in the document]
paulbalaji: Changed this to a type-only export so the runtime entrypoint no longer re-exports compile-time-only symbols.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1271,11 +1271,96 @@ export class WarpCore {
...
+      assert(
+        destinationToken.chainName === destinationName,
+        `Destination token chain mismatch for ${destinationName}`,
+      );
+      const matchedToken = destinationCandidates.find(
+        (candidate) =>
+          candidate.equals(destinationToken) ||
+          candidate.addressOrDenom.toLowerCase() ===
+            destinationToken.addressOrDenom.toLowerCase(),
+      );
```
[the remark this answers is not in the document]
paulbalaji: Switched these fallback token-identifier comparisons to the same protocol-safe helper used by `TokenMetadata`, instead of unconditional lowercasing.

typescript/widgets/src/walletIntegrations/cosmosWallet.ts
```diff
@@ -0,0 +1,94 @@
...
+    const addresses: Array<ChainAddress> = [];
+    let publicKey: Promise<HexString> | undefined;
+    let isReady = false;
+
+    for (const [chainName, context] of Object.entries(chainToContext)) {
+      if (!context.address) continue;
+      addresses.push({ address: context.address, chainName });
+      publicKey = context
+        .getAccount()
+        .then((acc) => Buffer.from(acc.pubkey).toString('hex'));
```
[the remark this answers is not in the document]
paulbalaji: Fixed both issues: replaced `Buffer.from()` with a browser-safe hex conversion, and the active-chain hook now reports the first connected Cosmos chain instead of casting an empty object.

typescript/widgets/src/walletIntegrations/cosmosWallet.ts
```diff
@@ -0,0 +1,110 @@
...
+} from './types.js';
+import { getChainsForProtocol } from './utils.js';
+
+const PLACEHOLDER_COSMOS_CHAIN = cosmoshub.name;
+
+function toHexString(bytes: Uint8Array): HexString {
+  return ensure0x(
+    Array.from(bytes, (byte) => byte.toString(16).padStart(2, '0')).join(''),
+  );
+}
```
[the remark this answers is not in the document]
paulbalaji: Kept intentional. The return type here is `HexString`, so the `0x` prefix is the correct normalized form; the old unprefixed encoding was the odd one out. This patch only normalizes the existing Cosmos public key bytes into the declared type.

typescript/sdk/src/token/adapters/evmLikeHyp.ts
```diff
@@ -0,0 +1,94 @@
...
+    case standards.hypXerc20[0]:
+    case standards.hypXerc20[1]:
+      return new EvmHypXERC20Adapter(chainName, multiProvider, {
+        token: addressOrDenom,
+      });
+    case standards.hypXerc20Lockbox[0]:
+    case standards.hypXerc20Lockbox[1]:
+      return new EvmHypXERC20LockboxAdapter(chainName, multiProvider, {
+        token: addressOrDenom,
+      });
```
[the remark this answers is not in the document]
paulbalaji: Fixed. `createEvmLikeHypAdapter()` no longer hardcodes `[0]` / `[1]`; it now checks the configured standard arrays directly, so additional supported standards cannot silently fall through.

typescript/sdk/src/token/TokenMetadata.ts
```diff
@@ -0,0 +1,208 @@
...
+    return new this({
+      chainName,
+      standard: PROTOCOL_TO_NATIVE_STANDARD[knownProtocol],
+      addressOrDenom: nativeToken.denom ?? '',
+      decimals: nativeToken.decimals,
+      symbol: nativeToken.symbol,
+      name: nativeToken.name,
+      logoURI,
+      coinGeckoId: gasCurrencyCoinGeckoId,
+    }) as InstanceType<T>;
```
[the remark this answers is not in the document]
paulbalaji: Added the required `// CAST:` comment on the polymorphic static factory cast in `TokenMetadata.FromChainMetadataNativeToken`.

typescript/widgets/src/walletIntegrations/radixWallet.ts
```diff
@@ -0,0 +1,88 @@
...
+  assert(rdt, `radix dapp toolkit not defined`);
+
+  const popUp = usePopup();
+  assert(popUp, `radix wallet popup not defined`);
+
+  const { setAccounts } = useAccount();
+
+  rdt.walletApi.provideChallengeGenerator(async () => generateRolaChallenge());
+
+  return async () => {
```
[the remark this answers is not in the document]
paulbalaji: Wrapped the Radix connect handler in `useCallback` and switched popup cleanup to `try/finally` so the pending popup always clears, even on request errors.

typescript/sdk/src/providers/MultiProviderAdapter.ts
```diff
@@ -0,0 +1,149 @@
...
+export interface MultiProviderAdapterOptions extends MinimalProviderRegistryOptions {}
+
+export function wrapMultiProviderProviders<MetaExt = {}>(
+  providers: MultiProvider<MetaExt>['providers'],
+): ChainMap<TypedProvider> {
+  return objMap(providers, (_, provider) => ({
+    type: ProviderType.EthersV5,
+    provider,
+  })) as ChainMap<TypedProvider>;
+}
```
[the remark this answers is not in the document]
paulbalaji: Added the required `// CAST:` comment and also tightened `fromMultiProvider` wrapping so zkSync chains preserve the correct typed-provider discriminant.

typescript/sdk/src/token/adapters/evmLikeHyp.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+  hasChainMetadata,
+  hasOnlyHyperlaneConnections,
+  type HypTokenAdapterInput,
+} from './hypTokenAdapterUtils.js';
+import type { MultiProviderAdapter } from '../../providers/MultiProviderAdapter.js';
+import type { TokenStandard } from '../TokenStandard.js';
+
+interface EvmLikeHypAdapterStandards {
+  native: TokenStandard;
```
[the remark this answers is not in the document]
Xaroz: ah I suppose it is because of tron and we are trying to reuse

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -100,38 +101,35 @@ export class WarpCore {
...
+      (token) =>
         new Token({
-          ...t,
-          addressOrDenom: t.addressOrDenom || '',
+          ...token,
+          addressOrDenom: token.addressOrDenom || '',
           connections: undefined,
         }),
     );
-    // Connect tokens together
```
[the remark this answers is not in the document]
paulbalaji: Agreed. Restored that comment in `91f9ad9749`.

It is still doing real work here: `warpRouteId` is the disambiguator when multiple tokens share the same chain + address pair but belong to different warp routes (for example the M0 portal cases). Dropping the comment made that lookup look more arbitrary than it is.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1577,11 +1578,98 @@ export class WarpCore {
     return null;
   }
 
+  protected resolveDestinationToken({
```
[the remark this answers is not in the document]
paulbalaji: Agreed. Restored the moved method docs in `91f9ad9749` for `findToken`, `getTokenChains`, `getTokensForChain`, and `getTokensForRoute`.

Those descriptions were still useful after the refactor; they just got lost when the methods moved.

typescript/sdk/src/token/TokenAmount.ts
```diff
@@ -1,13 +1,13 @@
 import { Numberish, fromWei } from '@hyperlane-xyz/utils';
 
-import type { IToken } from './IToken.js';
+import type { ITokenMetadata } from './ITokenMetadata.js';
 
-export class TokenAmount {
+export class TokenAmount<TToken extends ITokenMetadata = ITokenMetadata> {
```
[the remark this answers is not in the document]
paulbalaji: Needed for the metadata/read split.

The point is to let read-only paths carry `TokenAmount<ITokenMetadata>` without forcing full runtime `IToken` everywhere, while write/runtime paths still preserve `TokenAmount<IToken>`. That way `plus(...)` / `minus(...)` and other returns keep the same token shape instead of widening everything back to the lowest common metadata type.

So there is no runtime effect here; it is purely to preserve type precision across the lighter read-path surfaces this PR introduces.

typescript/widgets/src/walletIntegrations/aleoWallet.ts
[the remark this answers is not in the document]
paulbalaji: Yes, but indirectly.

The split itself does not make things faster by magic. The benefit is that it gives apps a lighter protocol-bound wallet surface: account/connect/details consumers no longer have to sit in the same module as switch-network / watch-asset / tx helpers.

That only turns into a real win when apps import the narrower `*Wallet` entrypoints, or when the aggregate wrappers are rewired to those files. This PR does the latter, so the split is there to make that narrower import graph possible.

typescript/widgets/src/walletIntegrations/ethereum.ts
```diff
@@ -112,23 +70,31 @@ export function useEthereumSwitchNetwork(
...
       if (activeChainName && activeChainName !== chainName)
         await switchNetwork(chainName);
 
       let tokenAddress = '';
       if (LOCKBOX_STANDARDS.includes(token.standard)) {
-        const adapter = token.getAdapter(
-          multiProvider,
-        ) as EvmHypXERC20LockboxAdapter;
+        const { createEvmHypAdapter } =
+          await import('@hyperlane-xyz/sdk/token/adapters/evmHyp');
```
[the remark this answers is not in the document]
paulbalaji: I think so.

This branch is rare and click-driven (`watchAsset` for lockbox tokens only), and the import target pulls in hyp-adapter code we do not need on the normal account/connect/details paths. Since `ethereum.ts` is still imported by broader wallet surfaces, keeping the lockbox-specific adapter load lazy is a cheap way to keep that adapter graph off hotter paths.

So it is not required for correctness, but it is intentional for chunking/locality.

typescript/widgets/src/walletIntegrations/radixWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+      isReady: !!accounts.length,
+    }),
+    [accounts],
+  );
+}
+
+export function useRadixWalletDetails(): WalletDetails {
+  const name = 'Radix Wallet';
+  const logoUrl =
+    'https://raw.githubusercontent.com/radixdlt/radix-dapp-toolkit/refs/heads/main/docs/radix-logo.png';
```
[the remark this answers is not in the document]
paulbalaji: Done in `91f9ad9749`. Hoisted the static Radix wallet details to a top-level constant so the hook just returns the shared object.

paulbalaji:
Follow-up pass landed on this branch in `622bc297e9`.

Scope kept narrow:
- finished the remaining SDK token-adapter runtime entrypoint cleanup
- switched `AleoTokenAdapter` and `RadixTokenAdapter` to VM `./runtime` entrypoints
- switched the Cosmos module token adapter to a type-only `MsgRemoteTransferEncodeObject` import via `@hyperlane-xyz/cosmos-sdk/runtime`
- added that type export to `typescript/cosmos-sdk/src/runtime.ts`

Secret gist with the broader findings / next-step analysis:
https://gist.github.com/paulbalaji/3b02a115746386bb199187dc5a1c3ec9

I did not widen scope into the larger runtime adapter registry redesign in this pass.

typescript/sdk/src/token/adapters/sealevelHyp.ts
```diff
@@ -0,0 +1,68 @@
...
+    return undefined;
+  }
+
+  if (
+    standard !== TokenStandard.SealevelHypNative &&
+    standard !== TokenStandard.SealevelHypCollateral &&
+    standard !== TokenStandard.SealevelHypSynthetic
+  ) {
+    return undefined;
+  }
```
xeno097: do we need this early return if we are going to return undefined anyways in the default case below?

typescript/tron-sdk/src/runtime.ts
```diff
@@ -0,0 +1,11 @@
+export { TronJsonRpcProvider } from './ethers/TronJsonRpcProvider.js';
```
xeno097, line 1: I wonder if we still need a dedicated tron sdk package as for prod we are just using the evm sdk

typescript/sdk/src/providers/MinimalProviderRegistry.ts
```diff
@@ -0,0 +1,271 @@
...
+    const provider = this.getProvider(chainNameOrId, transaction.type);
+    const chainMetadata = this.getChainMetadata(chainNameOrId);
+    return estimateTransactionFee({
+      transaction,
+      provider,
+      chainMetadata,
+      sender,
+      senderPubKey,
+    });
+  }
```
xeno097: **Design (medium): `MinimalProviderRegistry` isn't minimal because it bakes in `estimateTransactionFee`.**

This method imports `estimateTransactionFee` from `transactionFeeEstimators.js`, which contains fee estimation handlers for all protocol types (EVM, Cosmos, Solana, Starknet, Radix, Aleo). Any consumer of this "minimal" class transitively loads those estimators.

If the goal is a lightweight registry for UI reads, fee estimation should be opt-in — either a standalone function that accepts a registry, or pushed down to `MultiProviderAdapter` where the heavier behavior already lives.

```ts
// Standalone alternative:
export function estimateTransactionFee(
  registry: MinimalProviderRegistry,
  opts: { chainNameOrId, transaction, sender, senderPubKey? },
): Promise<TransactionFeeEstimate> { ... }
```

typescript/widgets/src/walletIntegrations/accounts.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { useAleoAccount } from './aleoWallet.js';
+import { getAccountAddressForChain } from './accountUtils.js';
+import { useCosmosAccount } from './cosmosWallet.js';
+import { useEthereumAccount } from './ethereumWallet.js';
+import { useRadixAccount } from './radixWallet.js';
+import { useSolanaAccount } from './solanaWallet.js';
+import { useStarknetAccount } from './starknetWallet.js';
+import { type AccountInfo } from './types.js';
+import { useTronAccount } from './tronWallet.js';
```
xeno097, line 19: **Bundle size concern (high): This file unconditionally imports all 7 protocol wallet hooks.**

Every coordinator file (`accounts.tsx`, `connectFns.tsx`, `disconnectFns.tsx`, `transactionFns.tsx`, `walletDetails.tsx`, `watchAsset.tsx`) imports all 7 `*Wallet.ts` files, which each import their respective SDK (`@rainbow-me/rainbowkit`, `@cosmjs/*`, `@solana/wallet-adapter-react`, `@starknet-react/core`, etc.).

A consumer who only uses Ethereum still loads Solana, Cosmos, Starknet, Radix, Aleo, and Tron wallet SDKs. The split from `multiProtocol.tsx` into separate files improves readability but delivers approximately zero tree-shaking benefit.

To actually tree-shake per-protocol:
- Use `dynamic import()` inside the hooks, or
- Accept a protocol→hook map as a parameter so consumers inject only what they need:
```tsx
const accounts = useAccounts(multiProvider, {
  [ProtocolType.Ethereum]: useEthereumAccount,
  [ProtocolType.Sealevel]: useSolanaAccount,
});
```

typescript/sdk/package.json
```diff
@@ -17,7 +17,34 @@
   "sideEffects": false,
   "types": "./dist/index.d.ts",
   "exports": {
-    ".": "./dist/index.js"
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./metadata/*": {
+      "types": "./dist/metadata/*.d.ts",
```
xeno097, line 25: **API surface concern (medium): Wildcard exports expose internals as public API.**

`"./providers/*"` exports every file in the providers directory, including `MinimalProviderRegistry`, `MultiProviderAdapter`, `defaultProviderBuilderMaps`, `rpcHealthTest`, `SmartProvider/`, etc. Once published, these are semver commitments — any rename or refactor is a breaking change.

Consider explicit named exports instead of wildcards:
```json
"./providers/MinimalProviderRegistry": { ... },
"./providers/MultiProtocolProvider": { ... },
"./providers/ProviderType": { ... }
```

This gives you the freedom to refactor internals without breaking consumers.

typescript/sdk/src/providers/defaultProviderBuilderMaps.ts
```diff
@@ -0,0 +1,55 @@
...
+  [ProviderType.SolanaWeb3]: defaultSolProviderBuilder,
+  [ProviderType.CosmJs]: defaultCosmJsProviderBuilder,
+  [ProviderType.CosmJsWasm]: defaultCosmJsWasmProviderBuilder,
+  [ProviderType.CosmJsNative]: defaultCosmJsNativeProviderBuilder,
+  [ProviderType.Starknet]: defaultStarknetJsProviderBuilder,
+  [ProviderType.ZkSync]: defaultZKSyncProviderBuilder,
+  [ProviderType.Radix]: defaultRadixProviderBuilder,
+  [ProviderType.Aleo]: defaultAleoProviderBuilder,
+  [ProviderType.Tron]: defaultTronProviderBuilder,
+};
```
xeno097, line 41: **Bundle size concern (medium): `defaultProviderBuilderMap` eagerly imports all 12 builder modules.**

This object is a static map that imports every builder at module load time. Since `MultiProtocolProvider` imports this, and `index.ts` re-exports `MultiProtocolProvider`, any consumer importing from `@hyperlane-xyz/sdk` loads all builders.

For the subpath exports to deliver real savings, this map should not be the default. Consider making `MultiProtocolProvider`'s constructor accept an explicit builder map (no default), or providing a separate `createFullProviderBuilderMap()` factory that consumers opt into.

xeno097 (commented):
## Brutally Honest Bundle Size Review

This PR has **real wins** (ProviderType `import type` fix, barrel gutting, Token/TokenMetadata split) mixed with **structural changes that don't yet deliver the promised bundle savings**. The biggest gap: no before/after bundle size measurements for a 138-file, 46-commit PR whose stated goal is bundle reduction.

### What actually reduces bundle size

1. **`ProviderType.ts` → `import type` conversions** — Highest-value change. Converting `starknet`, `zksync-ethers`, `aleo-sdk`, `cosmos-sdk`, `radix-sdk` from value imports to type-only imports eliminates those SDKs from any module that only needs the discriminated union. This alone may justify the PR.
2. **Widgets `index.ts` barrel gutted** — Removing all `walletIntegrations/*` re-exports means `import { SomeIcon } from '@hyperlane-xyz/widgets'` no longer transitively loads every wallet SDK.
3. **AltVM `/runtime` subpath entrypoints** — Consumers referencing only types pay zero runtime cost.
4. **`ChainMetadataResolver`** — Clean zero-dep metadata lookup function. Right primitive for UI code.

### What doesn't reduce bundle size (yet)

1. **`MinimalProviderRegistry` owns `estimateTransactionFee`** — see inline comment. Any consumer of this "minimal" class loads the full fee estimator chain.
2. **Widget wallet coordinators import ALL 7 protocols unconditionally** — see inline comment on `accounts.tsx`. The split improves readability but delivers ~zero tree-shaking for single-protocol consumers.
3. **SDK `index.ts` still re-exports `MultiProtocolProvider` + `defaultProviderBuilderMap`** — Any consumer importing from `@hyperlane-xyz/sdk` (the default entrypoint) loads the full builder map. Subpath exports only help consumers who explicitly rewrite imports.

### Missing from this PR

- **Bundle size measurements.** No before/after analysis. For a PR with this blast radius, this is the most important thing missing.
- **Consumer migration guide.** Subpath exports are useless if consumers don't know to use them.

### Questions for the author

1. Do you have before/after bundle size numbers?
2. What's the expected consumer migration path — will FE packages rewrite imports to subpaths, or should this be transparent?
3. Is `MultiProviderAdapter` intended to be permanent or a deprecation bridge?
4. Was `GnosisTxBuilderProvider` intentionally part of this PR?

### Positives

- Type safety is excellent (only justified `as` casts with comments, no `any`)
- Token/TokenMetadata split is clean with no circular deps
- `warp/read.ts` is a proper lightweight surface
- `ChainMetadataResolver` is well-designed
- Per-VM hyp adapter factories create proper lazy-loading boundaries

typescript/sdk/src/providers/ProviderType.ts
```diff
@@ -245,6 +242,11 @@ export interface ZKSyncProvider extends TypedProviderBase<ZKSyncBaseProvider> {
   provider: ZKSyncBaseProvider;
 }
 
+export interface GnosisTxBuilderProvider extends TypedProviderBase<EV5Providers.Provider> {
+  type: ProviderType.GnosisTxBuilder;
+  provider: EV5Providers.Provider;
```
xeno097, line 247: **Scope creep (low): `GnosisTxBuilderProvider` added without tests or changeset mention.**

This is a new provider type variant snuck into a bundle-size PR. It's referenced in `defaultProviderBuilderMaps.ts` and `builders/ethersV5.ts` but has no dedicated test coverage and isn't mentioned in either changeset. Should this be a separate PR?

typescript/widgets/src/walletIntegrations/transactionFns.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { type KnownProtocolType, ProtocolType } from '@hyperlane-xyz/utils';
+
+import { useAleoTransactionFns } from './aleo.js';
+import { useCosmosTransactionFns } from './cosmos.js';
+import { useEthereumTransactionFns } from './ethereum.js';
+import { useRadixTransactionFns } from './radix.js';
+import { useSolanaTransactionFns } from './solana.js';
+import { useStarknetTransactionFns } from './starknet.js';
+import { type ChainTransactionFns } from './types.js';
+import { useTronTransactionFns } from './tron.js';
```
xeno097, line 13: **Bundle size concern (same as accounts.tsx): Imports all 7 protocol files unconditionally.**

Note that these imports are from the heavier protocol files (`ethereum.ts`, `cosmos.ts`, etc.) rather than the lighter `*Wallet.ts` files — these contain `sendTransaction`, `switchNetwork`, and `watchAsset` implementations which import wagmi, cosmjs signing clients, etc. This is the heaviest coordinator file in terms of transitive deps.

typescript/sdk/src/providers/MultiProtocolProvider.ts
```diff
@@ -1,302 +1,66 @@
...
-  HexString,
-  ProtocolType,
-  objFilter,
-  objMap,
-  pick,
-  rootLogger,
-} from '@hyperlane-xyz/utils';
+import { ProtocolType, pick } from '@hyperlane-xyz/utils';
 
 import { multiProtocolTestChainMetadata } from '../consts/testChains.js';
```
xeno097, line 3: **Design (low): Test fixture import at module top level.**

`multiProtocolTestChainMetadata` from `../consts/testChains.js` is imported unconditionally but only used by `createTestMultiProtocolProvider`. If `testChains.js` pulls in chain fixture data, production consumers pay for it. Consider a dynamic import or moving the test factory to a separate file.

typescript/sdk/src/token/adapters/sealevelHyp.ts
paulbalaji: Simplified this in `c51f23535f`. The pre-switch guard is gone now; we only early-return for missing metadata/standard, and the `switch` handles the Sealevel standards directly with `default -> undefined`.

typescript/tron-sdk/src/runtime.ts
paulbalaji: I think we still do for now. `./runtime` is only the narrow UI/client surface; the package still owns Tron-specific provider/signer/json-rpc classes and factory/runtime plumbing that are not just an EVM re-export. If we collapse the package entirely, that wants to be a separate package-identity cleanup rather than part of this perf PR.

typescript/sdk/src/providers/MinimalProviderRegistry.ts
paulbalaji: Agreed. I moved `estimateTransactionFee(...)` down to `MultiProviderAdapter` in `c51f23535f`, so `MinimalProviderRegistry` no longer imports `transactionFeeEstimators` at all and stays metadata/provider-registry only.

typescript/widgets/src/walletIntegrations/accounts.tsx
paulbalaji: Agreed on the coordinator wrappers themselves: they are still full-matrix convenience surfaces, not the tree-shakable path. I’m not widening this PR into hook injection / dynamic-import API changes here. The intended lean surface from this PR is the protocol-specific `*Wallet.ts` entrypoints plus app-side migration off these aggregators; keeping the aggregators for existing callers is an explicit compatibility tradeoff.

typescript/sdk/package.json
paulbalaji: Good call. I replaced the broad `./providers/*` export with explicit provider subpath exports in `c51f23535f`, while keeping the intentional patterned exports for `./providers/builders/*` and `./providers/runtime/*`.

typescript/sdk/src/providers/defaultProviderBuilderMaps.ts
paulbalaji: Agreed this remains the full, battery-included path. I’m leaving that behavior in `MultiProtocolProvider` intentionally; the leaner surfaces from this PR are `MinimalProviderRegistry`, `MultiProviderAdapter`, and the explicit runtime/builder subpaths that consumers can opt into. Making the full provider map opt-in by default is a larger API/behavior change than I want in this branch.

typescript/sdk/src/providers/ProviderType.ts
paulbalaji: This one is covering an existing builder path rather than adding a new feature branch. It does have coverage now in `MultiProtocolProvider.test.ts` (the `defaultProviderBuilderMap[ProviderType.GnosisTxBuilder]` assertion), so I’m treating it as typed-surface completion needed to keep the builder map honest rather than a separate feature PR.

typescript/widgets/src/walletIntegrations/transactionFns.tsx
paulbalaji: Same answer as `accounts.tsx`: agreed this convenience wrapper is still full-matrix and therefore not the lean import path. I don’t want to turn this branch into a new injected/dynamic coordinator API. The protocol-specific modules are the intended tree-shakable surface; `transactionFns.tsx` stays as the compatibility convenience layer for existing consumers.

typescript/sdk/src/providers/MultiProtocolProvider.ts
paulbalaji: Fair callout, but this static test helper/import is pre-existing on `main`; this PR didn’t introduce it. I’m leaving it here to avoid turning this branch into a `createTestMultiProtocolProvider` API move across the SDK test surface. If we want to move that helper into a dedicated testutils file, I’d do it as a separate cleanup.

paulbalaji:
Replying to Xeno's review summary after the inline follow-up pass.

Answers to the review questions:

1. Before/after consumer numbers do exist now. The current tracking gist has the latest Warp UI / Explorer clean-build benchmarks and emitted-output comparisons: https://gist.github.com/paulbalaji/22d11456c1f64222bade419198c21d91
2. Consumer migration is explicit, not transparent. The point of this PR is to expose lighter subpaths/surfaces so FE apps can opt into them; importing `@hyperlane-xyz/sdk` root or the full convenience wallet coordinators still keeps the heavier all-VM path.
3. `MultiProviderAdapter` is intended as the lightweight typed bridge layer, not just a temporary deprecation shim. The latest follow-up here also moved fee estimation off `MinimalProviderRegistry` and onto `MultiProviderAdapter` so the class split is cleaner.
4. `GnosisTxBuilderProvider` was intentional as typed-surface completion for an existing builder path, not a separate feature branch. It now has direct coverage in `typescript/sdk/src/providers/MultiProtocolProvider.test.ts`.

Follow-up changes from this pass:
- moved `estimateTransactionFee(...)` off `MinimalProviderRegistry`
- replaced `./providers/*` wildcard exports with explicit provider subpath exports
- cleaned up the Sealevel hyp adapter control flow

All current inline Xeno threads should now be resolved.

xeno097 (approved):

paulbalaji:
Follow-up note on scope and next steps.

Two follow-up PRs build directly on this branch and address the remaining concerns that are intentionally not solved inside `#8433` itself:

- `#8575` `feat!: require explicit sdk runtime registration`
- `#8577` `feat: add selective widget wallet surfaces`

How they help relative to `#8433`:

1. `#8433` creates the seams.
- lighter SDK subpaths
- VM `./runtime` entrypoints
- lighter widget protocol-specific entrypoints
- typed bridge layers (`MinimalProviderRegistry`, `MultiProviderAdapter`, etc.)

That makes selective consumption possible, but it does **not** by itself force consumers off the old full-matrix path.

2. `#8575` makes the SDK side explicit.
- removes the hidden root auto-registration fallback
- runtime starts empty
- consumers register only the VM runtimes they need via `@hyperlane-xyz/sdk/runtime` / `@hyperlane-xyz/sdk/register/*`
- `registerAllRuntimeAdapters()` keeps the old batteries-included behavior available, but now as an explicit opt-in

That means the seams from `#8433` stop being purely structural and start changing actual runtime selection behavior.

3. `#8577` does the same thing for widgets.
- adds injectable base wallet UI surfaces (`BaseConnectWalletButton`, `BaseAccountList`, `BaseMultiProtocolWalletModal`)
- downstream apps can keep the same visuals while supplying only the protocol-specific wallet state/hooks they actually use
- existing full-matrix widget wrappers stay available as compatibility/convenience APIs

That directly addresses the remaining bundle-size concern on the widget coordinator files raised in review here.

Practical read:
- `#8433` = expose lighter surfaces / stop obvious fat-root imports
- `#8575` = make SDK runtime selection real instead of hidden full registration
- `#8577` = make widget wallet selection real instead of always going through the full hook matrix

So yes: I think it is correct to treat `#8575` and `#8577` as the explicit follow-up chain for the bigger design/bundle-size concerns that were called out during review on this PR, rather than trying to widen `#8433` further.

Xaroz (approved):

PR #8459, opened 2026-03-26
feat(rebalancer): add deBridge inventory adapter

Mo-Hussain:
Adds the deBridge inventory adapter with direct DLN order validation before signing. EVM/Tron calls are bound to documented entrypoints and decoded tokens, chains, amounts, recipient, authorities and refund beneficiary. Solana resolves lookup tables and validates the direct DLN instruction/accounts and compute fees. Source identity is observed before receipt or status waits; cancellation requests use the documented `srcAllowedCancelBeneficiary` parameter.

The SPL-token dependency now enters this PR because the DLN Solana validator needs account/program definitions. Existing dependency versions are preserved. Release notes match supported chain mappings and in-memory identity tracking.

Validation: build/lint and 32 deBridge regressions pass. **Execution readiness remains blocked:** real BSC→Ethereum swap-forwarder and Solana→Ethereum Jupiter payloads need complete supported decoders. Unknown forms reject before signing. Parent: #9343; child: #9057.

Implementation head: `db1951420ba874bffe650826a7b95fa9ce9cc854`. Fresh CI and the amd64 validation build must pass on the rewritten heads. [Provider validation coverage and operational gates](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/dcd094411799852aef123bae24babb09740aa133/typescript/rebalancer/docs/external-bridge-validation.md). This update made no production changes. This PR must remain draft until the operator explicitly asks to mark it ready.

### DIFF
--- a/typescript/rebalancer/src/bridges/DeBridgeBridge.ts
+++ b/typescript/rebalancer/src/bridges/DeBridgeBridge.ts
@@ -1,0 +1,494 @@
+import { ethers } from 'ethers';
+
+import type { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
+import { ProtocolType, assert } from '@hyperlane-xyz/utils';
+import type { Logger } from 'pino';
+import { v4 as uuidv4 } from 'uuid';
+
+import type {
+  BridgeQuote,
+  BridgeQuoteParams,
+  BridgeTransferResult,
+  BridgeTransferStatus,
+  IExternalBridge,
+} from '../interfaces/IExternalBridge.js';
+import {
+  DEBRIDGE_API_BASE,
+  DEBRIDGE_STATUS_API,
+  DEBRIDGE_TOOL,
+  formatAddressForDebridge,
+  hyperlaneChainIdToDebridge,
+  isDebridgeTronChain,
+  type DeBridgeCreateTxResponse,
+  type DeBridgeOrderStatusResponse,
+  type DeBridgeQuoteResponse,
+} from './deBridgeUtils.js';
+
+const TIMEOUT_MS = 30_000;
+const MAX_RETRIES = 3;
+const BASE_BACKOFF_MS = 1_000;
+
+type TronWebLike = {
+  trx: {
+    getTransactionInfo: (txId: string) => Promise<{
+      receipt?: { result?: string };
+    }>;
+    sign: (tx: unknown) => Promise<{ txID: string }>;
+    sendRawTransaction: (signed: unknown) => Promise<{ result?: boolean }>;
+  };
+  address: {
+    fromPrivateKey: (privateKey: string) => string;
+    toHex: (address: string) => string;
+  };
+  transactionBuilder: {
+    triggerSmartContract: (
+      contractAddress: string,
+      functionSelector: string,
+      options: { callValue: number; feeLimit: number },
+      parameters: Array<{ type: string; value: unknown }>,
+      issuerAddress: string,
+    ) => Promise<{ transaction?: { txID: string } }>;
+  };
+};
+
+export class DeBridgeBridge implements IExternalBridge {
+  readonly externalBridgeId = DEBRIDGE_TOOL;
+  readonly logger: Logger;
+  private readonly apiUrl: string;
+  private readonly chainMetadataByChainId: Map<number, ChainMetadata>;
+
+  constructor(
+    config: { apiUrl?: string; chainMetadata?: ChainMap<ChainMetadata> },
+    logger: Logger,
+  ) {
+    this.logger = logger;
+    this.apiUrl = config.apiUrl ?? DEBRIDGE_API_BASE;
+    this.chainMetadataByChainId = new Map();
+    if (config.chainMetadata) {
+      for (const metadata of Object.values(config.chainMetadata)) {
+        if (metadata.chainId !== undefined) {
+          this.chainMetadataByChainId.set(Number(metadata.chainId), metadata);
+        }
+      }
+    }
+  }
+
+  async quote(params: BridgeQuoteParams): Promise<BridgeQuote> {
+    const { fromChain, toChain, fromToken, toToken, fromAmount, toAmount } =
+      params;
+
+    assert(
+      !(fromAmount !== undefined && toAmount !== undefined),
+      'Cannot specify both fromAmount and toAmount',
+    );
+    assert(
+      fromAmount !== undefined || toAmount !== undefined,
+      'Must specify either fromAmount or toAmount',
+    );
+
+    const srcDebridgeChainId = hyperlaneChainIdToDebridge(fromChain);
+    const dstDebridgeChainId = hyperlaneChainIdToDebridge(toChain);
+
+    const srcToken = formatAddressForDebridge(fromToken, srcDebridgeChainId);
+    const dstToken = formatAddressForDebridge(toToken, dstDebridgeChainId);
+
+    const amountStr = (fromAmount ?? toAmount!).toString();
+
+    const url =
+      `${this.apiUrl}/dln/order/quote` +
+      `?srcChainId=${srcDebridgeChainId}` +
+      `&srcChainTokenIn=${srcToken}` +
+      `&srcChainTokenInAmount=${amountStr}` +
+      `&dstChainId=${dstDebridgeChainId}` +
+      `&dstChainTokenOut=${dstToken}` +
+      `&prependOperatingExpenses=true`;
+
+    this.logger.debug(
+      { fromChain, toChain, srcDebridgeChainId, dstDebridgeChainId },
+      'Requesting deBridge quote',
+    );
+
+    const response = await this.fetchWithRetry(url);
+    const data: DeBridgeQuoteResponse = await response.json();
+
+    assert(
+      data.estimation,
+      `deBridge quote failed: ${data.errorMessage ?? 'no estimation returned'}`,
+    );
+
+    const receivedAmount = BigInt(data.estimation.dstChainTokenOut.amount);
+    const feeCosts =
+      BigInt(data.fixFee ?? '0') + BigInt(data.protocolFee ?? '0');
+
+    return {
+      id: uuidv4(),
+      tool: DEBRIDGE_TOOL,
+      fromAmount:
+        params.fromAmount ?? BigInt(data.estimation.srcChainTokenIn.amount),
+      toAmount: receivedAmount,
+      // deBridge guarantees exact output amounts for stablecoins
+      toAmountMin: receivedAmount,
+      executionDuration: 60,
+      // Gas included in operating expenses via prependOperatingExpenses=true
+      gasCosts: 0n,
+      feeCosts,
+      route: data,
+      requestParams: { ...params },
+    };
+  }
+
+  async execute(
+    quote: BridgeQuote<DeBridgeQuoteResponse>,
+    privateKeys: Partial<Record<ProtocolType, string>>,
+  ): Promise<BridgeTransferResult> {
+    const { requestParams: params } = quote;
+    const srcDebridgeChainId = hyperlaneChainIdToDebridge(params.fromChain);
+    const dstDebridgeChainId = hyperlaneChainIdToDebridge(params.toChain);
+    const isTronSource = isDebridgeTronChain(srcDebridgeChainId);
+    const isTronDest = isDebridgeTronChain(dstDebridgeChainId);
+
+    let senderAddress: string;
+    if (isTronSource) {
+      const tronKey = privateKeys['tron' as ProtocolType];
+      assert(tronKey, 'Missing private key for Tron chain');
+      const { TronWeb } = await import('tronweb');
+      const strippedKey = tronKey.replace(/^0x/, '');
+      const tw = new TronWeb({
+        fullHost: 'https://api.trongrid.io',
+        privateKey: strippedKey,
+      }) as unknown as TronWebLike;
+      senderAddress = tw.address.fromPrivateKey(strippedKey);
+    } else {
+      const evmKey = privateKeys[ProtocolType.Ethereum];
+      assert(evmKey, 'Missing private key for EVM chain');
+      senderAddress = new ethers.Wallet(evmKey).address;
+    }
+
+    let recipientAddress: string;
+    if (params.toAddress) {
+      recipientAddress = params.toAddress;
+    } else if (isTronDest && !isTronSource) {
+      assert(
+        params.toAddress,
+        'toAddress is required when bridging to Tron from a non-Tron chain',
+      );
+      recipientAddress = params.toAddress;
+    } else {
+      recipientAddress = senderAddress;
+    }
+
+    const srcSender = formatAddressForDebridge(
+      senderAddress,
+      srcDebridgeChainId,
+    );
+    const dstRecipient = formatAddressForDebridge(
+      recipientAddress,
+      dstDebridgeChainId,
+    );
+
+    const srcToken = formatAddressForDebridge(
+      params.fromToken,
+      srcDebridgeChainId,
+    );
+    const dstToken = formatAddressForDebridge(
+      params.toToken,
+      dstDebridgeChainId,
+    );
+    const amountStr = (params.fromAmount ?? quote.fromAmount).toString();
+
+    const createTxUrl =
+      `${this.apiUrl}/dln/order/create-tx` +
+      `?srcChainId=${srcDebridgeChainId}` +
+      `&srcChainTokenIn=${srcToken}` +
+      `&srcChainTokenInAmount=${amountStr}` +
+      `&dstChainId=${dstDebridgeChainId}` +
+      `&dstChainTokenOut=${dstToken}` +
+      `&dstChainTokenOutRecipient=${dstRecipient}` +
+      `&srcChainOrderAuthorityAddress=${srcSender}` +
+      `&dstChainOrderAuthorityAddress=${dstRecipient}` +
+      `&prependOperatingExpenses=true`;
+
+    this.logger.info(
+      {
+        fromChain: params.fromChain,
+        toChain: params.toChain,
+        amount: amountStr,
+        sender: srcSender,
+        recipient: dstRecipient,
+      },
+      'Creating deBridge order transaction',
+    );
+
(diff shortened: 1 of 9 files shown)


typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,700 @@
...
+    this.logger.info(
+      {
+        from: wallet.address,
+        to: tx.to,
+        fromChain: quote.requestParams.fromChain,
+        orderId,
+      },
+      `Sending deBridge ${protocol} transaction`,
+    );
+    const transaction = await wallet.sendTransaction({
```
paulbalaji, line 381: [High] Do not sign an API-selected arbitrary target and calldata with the inventory EOA. The checks above only constrain value and prevent targeting the quoted source token; a compromised response can target another token held by this signer and encode `transfer(attacker, balance)`, or invoke another privileged contract, without using the exact source-token allowance. Allowlist the documented DLN source contract and validate its create-order calldata (or use a limited-authority signer), with an unrelated-asset regression.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,700 @@
...
+
+    const connection = new Connection(
+      this.getRpcUrl(quote.requestParams.fromChain),
+    );
+    const { blockhash, lastValidBlockHeight } =
+      await connection.getLatestBlockhash();
+    transaction.message.recentBlockhash = blockhash;
+    transaction.sign([keypair]);
+
+    const signature = await connection.sendTransaction(transaction);
```
paulbalaji, line 428: [High] A single required signer and matching fee payer do not make this API-supplied Solana transaction safe. It may invoke arbitrary programs and make signer-owned/delegated/close-authority accounts writable, so a compromised response can spend assets unrelated to the quoted route. Constrain allowed DLN programs/instruction shapes and writable accounts, or enforce complete signer asset-effect bounds with a limited signer; add an unrelated-asset drain test.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,700 @@
...
+        orderId,
+      },
+      `Sending deBridge ${protocol} transaction`,
+    );
+    const transaction = await wallet.sendTransaction({
+      to: tx.to,
+      data: tx.data,
+      value: ethers.BigNumber.from(tx.value),
+    });
+    const receipt = await waitForReceiptWithTimeout(transaction.wait(), {
```
paulbalaji, line 386: [High] Preserve the broadcast identity before waiting. If `sendTransaction` succeeds but this receipt wait times out or the RPC loses the receipt, `execute` throws before `InventoryRebalancer` records the tx hash/order ID, so a later cycle or restart can submit the movement again; the Solana confirmation path below has the same ambiguity. Return/persist immediately after broadcast and reconcile confirmation asynchronously/status-first. Add a broadcast-success/confirmation-error regression.

.changeset/debridge-inventory-bridge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/rebalancer': minor
+---
+
+Added the deBridge DLN inventory adapter for Ethereum, BSC, Arbitrum, Plasma, Tron, and Solana. Quotes and execution responses were validated before exact token approvals, and persisted order IDs were used for status tracking.
```
paulbalaji, line 5: [Low] Remove Plasma from the release note: the current mapping and tests deliberately reject chain 9745, so the published note would promise unavailable behavior.

paulbalaji (commented):
Reviewed all 9 own-delta files at exact head `e302fb15b72d7a8ad7047a03806fa95a1f1aa488` against true base `578bcaab9255c5126eb094f43aee15c871e3df28`, plus callers and downstream lifecycle handling. I also reconciled Mo's prior review: all 12 unresolved threads are outdated, have no replies, and their utility/Tron-SDK/process questions are addressed in current code. Three high-severity execution/lifecycle issues and one release-note mismatch remain inline. Material source/E2E checks are green; the node-image failure was a Depot keepalive failure, not a demonstrated code failure.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+
+  private assertFeeWithinLimit(
+    response: DeBridgeQuoteResponse,
+    fromChain: number,
+    toChain: number,
+  ): void {
+    const { srcChainTokenIn, dstChainTokenOut } = response.estimation;
+    const commonDecimals = Math.max(
+      srcChainTokenIn.decimals,
+      dstChainTokenOut.decimals,
```
paulbalaji, line 602: [P1] Bind these decimals to trusted token metadata before using them to enforce `maxFeePercent`. Both currently come from the API response being checked. I exercised the actual forward `quote()` path with source USDT `amount=100000000, decimals=6`, destination USDT `amount=9900000, decimals=5`, and a 2.5% cap: it accepts a reported 1% fee. The destination token actually has 6 decimals, so this authorizes 9.9 USDT for 100 USDT (90.1% loss). `toAmountMin` inherits that same API amount, so the calldata minimum check does not restore the protection. Validate source/destination decimals against configured or on-chain token metadata and add a wrong-decimals regression. This requires a malformed/compromised API response; no live occurrence was observed.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+        const response = await fetch(url, { signal: controller.signal });
+        if (
+          response.status >= 400 &&
+          response.status < 500 &&
+          response.status !== 429
+        ) {
+          const body = await response.text();
+          throw new Error(`deBridge HTTP ${response.status}: ${body}`);
+        }
+        if (response.ok) return response;
```
paulbalaji, line 721: [P2] Consume the response body inside the timeout/retry scope. Returning the `Response` here runs `finally` and clears the 30s timer before `quote()`, `executeOrder()`, or `getStatus()` awaits `response.json()`. A 200 response that sends headers and stalls its JSON body therefore never times out or retries. Reproduced through the actual `quote()` method: still pending after 90s fake time, signal not aborted. Keep the deadline active through body consumption, and cancel/drain rejected response bodies before retrying; add a stalled-body regression.

paulbalaji (commented):
Reviewed at `db1951420ba874bffe650826a7b95fa9ce9cc854` against current parent `2892fcc6a98e6e7759cf1a289378cfb26192c439`. Two new issues remain inline: API decimals can bypass the fee limit, and the HTTP deadline stops before body consumption.

The earlier arbitrary EVM/Solana payload findings are addressed for supported direct DLN shapes; the tests reject unrelated-token calls and appended Solana transfers. The Plasma release-note mismatch is fixed. Confirmation waiting was removed from the adapter, while end-to-end ambiguous-submission recovery remains the explicitly separate #9361 activation work. Existing threads are still marked unresolved/outdated; I have not duplicated them.

Validation: all 32 deBridge tests passed using existing dependencies and a temporary source-export mapping to the exact parent helpers. The final rebase preserved the adapter delta (`range-diff` unchanged), and these tests were rerun against the final parent helpers. Additional executions of the actual `quote()` method reproduced both findings: a 2.5% cap accepted 100 USDT -> 9.9 USDT with falsified destination decimals; a stalled 200 body remained pending after 90 seconds of fake time with its abort signal unset. No live API corruption was observed. Kimi K3 review completed and its candidates were independently checked; the proposed arbitrary native-fee drain was rejected because the pinned DLN source enforces the exact native fee.

Further efficiency gains: batch Solana ALT reads (`executeSolana`, one `getMultipleAccountsInfo` instead of one request per table); reuse chain clients; use a shorter status-request budget and bounded concurrency for independent status reads in `ActionTracker` (outside this delta), since serial polls otherwise compound retries. Keep ordered state updates/nonce-sensitive submissions serialized.

The documented unsupported swap-forwarder/Jupiter payloads and funded execution validation remain readiness gates.

Current-head CI readback: 53 successful checks, 27 skipped, none failed or pending.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+    }
+  }
+
+  private assertFeeWithinLimit(
+    response: DeBridgeQuoteResponse,
+    fromChain: number,
+    toChain: number,
+  ): void {
+    const { srcChainTokenIn, dstChainTokenOut } = response.estimation;
+    const commonDecimals = Math.max(
```
paulbalaji, line 600: [P1, secondary acknowledgement] Confirming the [existing API-decimals finding](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007057126). My independent quote() probe used two 6-decimal tokens, input 1,000,000,000 and minimum output 1,000,000: reporting destination decimals=0 let 99.9% real loss pass maxFeePercent=2.5. The decoded DLN output minimum inherits the same accepted amount, so it does not repair the cap. Resolve authoritative decimals for quote/create-tx validation and test a malformed/compromised response; no live occurrence was observed.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
paulbalaji, line 721: [P2, secondary acknowledgement] Confirming the [existing body-timeout finding](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007057137). My probe returned 200 headers with a stalled stream; quote() remained pending after 31 seconds and the signal was not aborted. Returning Response runs finally before response.json(), disabling the advertised deadline and excluding body failures from retries. Consume the body within the timed attempt and cancel/drain failed responses.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+    const maxFeePercent = config.maxFeePercent ?? DEFAULT_MAX_FEE_PERCENT;
+    assert(
+      Number.isFinite(maxFeePercent) &&
+        maxFeePercent >= 0 &&
+        maxFeePercent <= MAX_PERCENT,
+      `maxFeePercent must be between 0 and ${MAX_PERCENT}`,
+    );
+    const maxFeeBps = maxFeePercent * BASIS_POINTS_PER_PERCENT;
+    assert(
+      Number.isInteger(maxFeeBps),
```
paulbalaji, line 96: [P2] Binary floating-point multiplication rejects valid two-decimal fee limits. For example `1.1 * 100` is `110.00000000000001`, so configuring 1.1% fails startup with “must have at most two decimal places”; 0.29 and 0.58 fail too. Reproduced through the constructor. Convert to integer bps using decimal-safe validation/rounding, and test these values as well as genuinely over-precise input.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+const BASIS_POINTS_DENOMINATOR = 10_000n;
+
+export interface DeBridgeBridgeConfig {
+  apiUrl?: string;
+  statusApiUrl?: string;
+  chainMetadata?: ChainMap<ChainMetadata>;
+  maxFeePercent?: number;
+}
+
+export class DeBridgeBridge implements IExternalBridge {
```
paulbalaji, line 71: [P2] Implement the native-token address contract for native inventory routes. `InventoryRebalancer.getNativeTokenAddress()` throws when this optional method is absent, and both capacity quoting and movement execution call it for HypNative. DeBridgeBridge has native-value handling below but no `getNativeTokenAddress()`, so that path is unreachable through the actual rebalancer. Return the appropriate native sentinel for supported protocols, or explicitly reject native routes in configuration and document that restriction. Add a native route test through InventoryRebalancer, not only direct adapter execution.

typescript/rebalancer/src/bridges/DeBridgeBridge.ts
```diff
@@ -0,0 +1,738 @@
...
+    this.logger.info(
+      {
+        from: wallet.address,
+        to: tx.to,
+        fromChain: quote.requestParams.fromChain,
+        orderId,
+      },
+      `Sending deBridge ${protocol} transaction`,
+    );
+    await options?.onTransferId?.(orderId);
```
paulbalaji, line 412: [P2] Bind the returned order ID to the source transaction before using it to settle the action. The decoder validates tx.data, but orderId is independently accepted from the API; getStatus then ignores txHash and queries only that ID. A valid intended order paired with an unrelated fulfilled orderId will therefore mark the wrong movement complete, while a nonexistent ID can age into a retry. Derive/validate the DLN ID where possible, or reconcile the source transaction's CreatedOrder event before accepting terminal status. Add a response with unchanged transaction bytes and a different orderId.

paulbalaji (commented):
Reviewed at `db1951420ba874bffe650826a7b95fa9ce9cc854` against current #9343. COMMENT: the prior arbitrary EVM/Solana target findings are addressed by the DLN decoders, and the Plasma release-note mismatch is removed. Submission observer activation remains the separate #9361 integration gate (its current caller now passes tracking options); not duplicated as a new adapter finding.

I reproduced the fee-decimals bypass and stalled-body deadline bug; concurrent review [already posted both](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#pullrequestreview-5199965221), and I have included explicitly linked secondary acknowledgements. Additional findings below cover valid percentage limits rejected at startup, missing native inventory support, and order-ID/source-transaction binding.

Validation: the deBridge tests passed within the 172-test bridge/receipt run; source probes reproduced a stalled body remaining un-aborted after 31 seconds and constructor rejection of 0.29, 0.58 and 1.1. The restack preserved this PR's own delta exactly. The same 99.9%-loss/2.5%-cap probe also passed deBridge by lying about destination decimals. No live provider transfer was sent.

Performance: keep body consumption inside the timed/retried request and cancel/drain failed response bodies before retrying. Reuse a provider/connection per configured chain instead of rebuilding it for each execution; preserve chain/RPC configuration identity. Existing Promise.all lookup-table reads are already concurrent.

Independent GLM/Kimi/Muse/adversarial reports were reconciled with source; speculative contract issues were excluded.

The PR already documents execution-readiness gates for real BSC→Ethereum swap-forwarder and Solana→Ethereum Jupiter payload decoders; these remain separate from unit-test/CI success.

Current-head CI: 53 successful checks, 27 skipped, none failed/pending.

paulbalaji (commented):
Reviewed own diff at `db1951420ba874bffe650826a7b95fa9ce9cc854` and combined stack at `dcd094411799852aef123bae24babb09740aa133`.

The prior arbitrary-target/program findings are addressed by the strict payload validators. Independently reproduced the existing [API-decimal loss-cap bypass](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007057126) and [response-body timeout gap](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007057137); both remain unresolved. For two actual six-decimal assets, input 1,000,000/output 10,000 with API decimals 6/4 passes a 1% limit despite 99% actual loss, through quote and execution/payload validation. The timed request also remains pending after 31 seconds if successful headers arrive but JSON stalls. Referencing existing threads rather than duplicating inline findings.

Further existing findings independently confirmed: [valid two-decimal fee limits rejected](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007087698) (1.1, 0.29, 0.58), and [missing native-token adapter hook](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007087702), which makes the actual inventory native-token resolution throw. Source inspection also confirms the [order-ID/source-transaction binding gap](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8459#discussion_r4007087707). A mismatched fulfilled ID can complete the wrong movement; on the combined stack a not-found result retains its reservation, so I do not endorse the earlier comment's automatic-retry claim.

Latest combined-tree validation: dependency build and rebalancer lint passed; 658 rebalancer, 24 SDK, 12 Tron, 5 submission and 25 Helm tests passed (724 total). One unchanged LiFi test hit its live-network timeout, then passed focused and full-suite reruns. The previous combined head also passed 14 simulation scenario/loader tests and simulator build; the restack only changed the Tron options fallback plus its regression test. These are combined-tree checks, not per-layer or live-provider execution tests.

Current-head CI rollup: 27 skipped, 53 success.

GLM, Kimi, Muse and an independent adversarial pass were reconciled against source and focused reproductions. Code approval does not establish production execution readiness: documented configured provider-path fixtures, live execution and rollout evidence remain separate gates.

paulbalaji (commented):
Reviewed own diff at `2bbdf64113b1120c46e419da567c265bac2e6dbc` as part of stacks 9635 and 9636 (14 PRs).

The existing P1 decimals/loss-cap finding remains reproducible through quote() AND prepare() after the latest validation changes. With two actual 6-decimal tokens, a mocked quote reports destination decimals=4 and minimum=990000 for input=100000000. The create-tx response then reports correct decimals and an estimate of 99000000 while encoding takeAmount=990000. prepare() accepts: 100 tokens input, only 0.99 tokens committed output, under a 2.5% cap. The decoder checks against the earlier tainted minimum. Resolve authoritative decimals during quoting and enforce the loss cap on decoded committed amounts; add inconsistent estimate/calldata coverage. This extends the evidence in the existing decimals thread, so I have not added another duplicate inline. No live transaction or exploit was performed.

Existing body-timeout, decimal fee-configuration (1.1/0.29/0.58 rejected by binary-float validation), and missing native-token integration findings remain. Source CreatedOrder plus finalized destination evidence now addresses the earlier unrelated-order false-completion concern. The payload allowlists materially improve EVM/Solana authorization.

Fresh combined-tree validation at `c7f454e83b201657f83bc733cb45f32b9983c0d9`: dependency builds and rebalancer lint passed; 736 rebalancer, 26 Helm, 24 SDK, 39 Tron and 5 submission tests passed (830 total). These are combined-tree checks, not independent tests of every intermediate head or live-provider/production execution.

Latest check-name view: 10 pass, 1 skipping. Independent Muse, GLM and Kimi review output was reconciled with source and prior threads; unverified model suggestions were excluded.

PR #8480, opened 2026-03-28
Archived

rbxict:


rbxict:
Follow-up pushed on the current head to tighten the new native-dusting path and clear the remaining low-risk review noise that was still actionable on this branch. The update adds a defensive malformed-event guard before handling SentTransferRemote payloads, removes the nullable alue! env parsing in the dusting config path, switches the new keyfunder min-delta calculation to pure igint, normalizes error handling in the service entrypoint, and fixes the RC/IGP balance math in the infra key-funder helper to respect each chain's native token decimals instead of assuming 18. I also cleaned up the new duster tests to use explicit test-only casts/comments instead of raw ny usage.\n\nLocal verification in this checkout is still limited by existing workspace setup issues: the package lint/test commands hit unresolved internal @hyperlane-xyz/* workspace artifacts and the package test scripts themselves still have broken quoted globs here, so I validated the patch by targeted diff review plus git diff --check instead of claiming green local CI where this environment cannot provide it.

PR #8517, opened 2026-04-03
fix: skip scraper lookup for txHash

troykessler:
### Description

This PR improves relaying speed for cctp fast transfers by storing the origin tx hash for each message id and submitting it to the ccip-server so it does not need to make an expensive scraper lookup. 

### Backward compatibility

The origin_tx_hash is optional and therefore fully backwards compatible

### Testing

Deployed on testnet cctp fast transfer [route](https://github.com/hyperlane-xyz/hyperlane-registry/pull/1451) and updated the testnet relayer and offchain lookup server for testing, it is confirmed that the scraper is being skipped now

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8517" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


typescript/ccip-server/src/services/CCTPService.ts
```diff
@@ -141,11 +151,20 @@ class CCTPService extends BaseService {
...
+    } else {
+      log.info(
+        { messageId },
+        'No tx hash from relayer, falling back to scraper lookup',
+      );
+      txHash = await this.hyperlaneService.getOriginTransactionHashByMessageId(
         messageId,
         log,
       );
+    }
```
paulbalaji, line 167: **No fallback to scraper when relayer-provided tx hash fails** (medium)

If `originTxHash` is present but stale (reorg) or wrong, `getTransactionReceipt` returns null and `getCCTPMessageFromReceipt` throws on `receipt.logs`. The error bubbles up as a hard request failure — the scraper path is never tried.

Pre-PR, the scraper was always the source of truth. Now a bad DB entry or caller-supplied bogus hash blocks attestation generation where pre-PR behavior would still succeed. Since this is explicitly an unsigned optimization hint, consider wrapping the receipt+event lookup in a try/catch and falling back to the scraper path:

```typescript
if (originTxHash) {
  log.info({ txHash: originTxHash, messageId }, 'Using tx hash provided by relayer');
  try {
    receipt = await provider.getTransactionReceipt(originTxHash);
    cctpMessage = await this.getCCTPMessageFromReceipt(receipt, messageId, log);
  } catch (err) {
    log.warn({ originTxHash, messageId, err }, 'Relayer hint failed, falling back to scraper');
    // fall through to scraper path
  }
}
```

This way a stale hint degrades gracefully instead of failing the request.

rust/main/hyperlane-base/src/db/rocks/hyperlane_db.rs
```diff
@@ -319,6 +320,10 @@ impl HyperlaneLogStore<HyperlaneMessage> for HyperlaneRocksDB {
             if stored_message {
                 stored = stored.saturating_add(1);
             }
+            self.store_dispatched_tx_hash_by_message_id(
```
paulbalaji, line 323: nit: This runs unconditionally, outside the `if stored_message` guard above. `store_message` skips writes for already-seen messages, but `store_dispatched_tx_hash_by_message_id` re-writes the same tx hash on every re-index pass.

If intentional (backfilling old DBs that lack this key), a comment would help. Otherwise, moving inside the `if` block avoids redundant RocksDB writes on high-reindex chains.

paulbalaji (approved):
## Review Summary

Clean PR — storing the origin tx hash to skip scraper lookups is a solid optimization. The Rust DB layer, trait additions, and TS service changes are well-structured. On-chain `verify()` in `TokenBridgeCctpBase.sol` validates CCTP message fields (sender, amount, recipient) against the Hyperlane message, so even a wrong `origin_tx_hash` just reverts on-chain. No fund safety risk.

Previous review feedback (GET handler signature, `as` cast, input validation) has been addressed. The `warn`-then-`.ok()` pattern for the DB lookup matches existing code in this file and is appropriate for a best-effort optimization.

Two items worth considering (inline comments below), neither blocking.

typescript/ccip-server/src/services/CCTPService.ts
troykessler: the scraper path would also error hard with a not found error because if we experience a reorg we are still in the reorg window, therefore the scraper wont see it anyway in that time

PR #8534, opened 2026-04-07
feat(warp-monitor): support Solana routers and protocol inventory

nambrot:
## Summary
- include Solana router nodes in warp-monitor route-node handling instead of filtering non-EVM addresses out
- preserve non-EVM router address casing while still normalizing EVM addresses
- keep explorer pending-transfer queries limited to EVM/bytea-compatible router nodes
- support inventory addresses by protocol family through `INVENTORY_ADDRESS_<PROTOCOL>` env vars, such as `INVENTORY_ADDRESS_ETHEREUM` and `INVENTORY_ADDRESS_SEALEVEL`
- keep legacy `INVENTORY_ADDRESS` fallback EVM-only; non-EVM inventory requires explicit protocol env var
- centralize inventory address trimming/fallback handling in `getConfiguredInventoryAddress`

## Why
This is the monitor-side piece needed for CROSS/ctusd Solana CCR testing:
- local monitor output now includes both Solana nodes
- projected deficit / pending destination metrics can be emitted for EVM nodes without Solana base58 routers breaking explorer bytea filters
- inventory metrics can target a shared EVM owner and a separate Solana owner without listing every chain

## Review update
- rebased/cleaned branch onto current `origin/main`
- dropped old xeno SVM adapter history now already present through later merged PRs
- branch diff now only touches `typescript/warp-monitor/src/`
- filtered `ExplorerPendingTransfersClient` input to EVM router nodes while preserving all router nodes for collateral/inventory metrics
- restored malformed EVM router/collateral skip behavior before normalization
- switched protocol inventory config from JSON env parsing to `INVENTORY_ADDRESS_<PROTOCOL>` env vars
- removed the service entry guard and restored unconditional `main()` execution
- centralized protocol-specific inventory lookup, trimming, empty-as-unset handling, and EVM-only legacy fallback in `getConfiguredInventoryAddress`

## Validation
- `pnpm turbo run build --filter=@hyperlane-xyz/warp-monitor...`
- `pnpm --dir typescript/warp-monitor lint`
- `pnpm --dir typescript/warp-monitor build`
- `pnpm --dir typescript/warp-monitor test`

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8534" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/warp-monitor/src/monitor.test.ts
+++ b/typescript/warp-monitor/src/monitor.test.ts
@@ -612,7 +612,7 @@ describe('WarpMonitor', () => {
       'CROSS/ctusd',
       pendingTransfersClient as ExplorerPendingTransfersClient,
       200,
-      '0xEA2117b24F7947647Bec60527B68f4244AE40c01',
+      '  0xEA2117b24F7947647Bec60527B68f4244AE40c01  ',
     );
 
     expect(calls).to.deep.equal([
--- a/typescript/warp-monitor/src/monitor.test.ts
+++ b/typescript/warp-monitor/src/monitor.test.ts
@@ -277,6 +277,71 @@ describe('WarpMonitor', () => {
     );
   });
 
+  it('skips Ethereum nodes with malformed router addresses', () => {
+    const monitor = new WarpMonitor(
+      {
+        warpRouteId: 'CROSS/ctusd',
+        checkFrequency: 10_000,
+      },
+      {} as IRegistry,
+    );
+
+    const nodes = invokeBuildRouterNodes(
+      monitor,
+      {
+        tokens: [
+          {
+            symbol: 'USDC',
+            chainName: 'ethereum',
+            addressOrDenom: 'not-an-evm-address',
+            collateralAddressOrDenom:
+              '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
+            decimals: 6,
+            name: 'USD Coin',
+            protocol: ProtocolType.Ethereum,
+          } as Token,
+        ],
+      } as WarpCore,
+      {
+        ethereum: { domainId: 1 },
+      },
+    );
+
+    expect(nodes).to.have.length(0);
+  });
+
+  it('skips Ethereum nodes with malformed collateral addresses', () => {
+    const monitor = new WarpMonitor(
+      {
+        warpRouteId: 'CROSS/ctusd',
+        checkFrequency: 10_000,
+      },
+      {} as IRegistry,
+    );
+
+    const nodes = invokeBuildRouterNodes(
+      monitor,
+      {
+        tokens: [
+          {
+            symbol: 'USDC',
+            chainName: 'ethereum',
+            addressOrDenom: '0xd4463cB3c90b3F49c673310BEC9bC18311134B47',
+            collateralAddressOrDenom: 'not-an-evm-address',
+            decimals: 6,
+            name: 'USD Coin',
+            protocol: ProtocolType.Ethereum,
+          } as Token,
+        ],
+      } as WarpCore,
+      {
+        ethereum: { domainId: 1 },
+      },
+    );
+
+    expect(nodes).to.have.length(0);
+  });
+
   it('excludes Solana router nodes from explorer queries', () => {
     const monitor = new WarpMonitor(
       {
--- a/typescript/warp-monitor/src/monitor.test.ts
+++ b/typescript/warp-monitor/src/monitor.test.ts
@@ -20,12 +20,15 @@ function createMockToken({
   collateralized,
   decimals,
   getBalance = async () => 0n,
+  protocol = ProtocolType.Ethereum,
 }: {
   collateralized: boolean;
   decimals: number;
-  getBalance?: () => Promise<bigint>;
+  getBalance?: (address: string) => Promise<bigint>;
+  protocol?: ProtocolType;
 }): Token {
   return {
+    protocol,
     isCollateralized: () => collateralized,
     amount: ((amount: bigint) => ({
       getDecimalFormattedAmount: () => Number(amount) / 10 ** decimals,
@@ -82,8 +85,22 @@ function invokeBuildRouterNodes(
   return buildRouterNodes.call(monitor, warpCore, chainMetadata);
 }
 
+function invokeBuildExplorerRouterNodes(
+  monitor: WarpMonitor,
+  routerNodes: RouterNodeMetadata[],
+): RouterNodeMetadata[] {
+  const buildExplorerRouterNodes = (monitor as any)
+    .buildExplorerRouterNodes as (
+    routerNodes: RouterNodeMetadata[],
+  ) => RouterNodeMetadata[];
+
+  return buildExplorerRouterNodes.call(monitor, routerNodes);
+}
+
 describe('WarpMonitor', () => {
   afterEach(() => {
+    delete process.env.INVENTORY_ADDRESS_ETHEREUM;
+    delete process.env.INVENTORY_ADDRESS_SEALEVEL;
     resetPendingDestinationMetrics();
     resetInventoryBalanceMetrics();
   });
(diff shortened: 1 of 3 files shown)


typescript/warp-monitor/src/service.ts
```diff
@@ -112,8 +115,9 @@ async function main(): Promise<void> {
   }
 }
 
-// Run the service
-main().catch((error) => {
-  rootLogger.error({ error }, 'Fatal error');
-  process.exit(1);
-});
+if (process.argv[1] && fileURLToPath(import.meta.url) === process.argv[1]) {
```
paulbalaji: This guard can silently prevent the service from starting. `fileURLToPath(import.meta.url)` resolves symlinks and returns an absolute path, but `process.argv[1]` is whatever was passed to `node`. On macOS the mismatch was reproduced locally: `import.meta.url` resolves under `/private/tmp/...` while `process.argv[1]` is `/var/tmp/...`.

Both `package.json` scripts (`start: node dist/service.js`, `start:dev: tsx src/service.ts`) invoke this file directly, so a false negative here means the service silently does nothing.

Suggestion: either revert to unconditional `main()` and export `main` for testability, or canonicalize both sides:
```ts
import { realpathSync } from 'node:fs';
if (process.argv[1] && realpathSync(fileURLToPath(import.meta.url)) === realpathSync(process.argv[1])) {
```

typescript/warp-monitor/src/monitor.ts
```diff
@@ -605,8 +631,35 @@ export class WarpMonitor {
...
+  }
+
   private buildNodeId(token: Token): string {
-    return `${token.symbol}|${token.chainName}|${token.addressOrDenom.toLowerCase()}`;
+    return `${token.symbol}|${token.chainName}|${this.normalizeNodeAddress(token)}`;
+  }
+
+  private normalizeNodeAddress(token: Token): string {
+    if (token.protocol === ProtocolType.Ethereum) {
+      return ethersUtils.getAddress(token.addressOrDenom).toLowerCase();
```
paulbalaji, line 665: Behavioral regression: the old `buildRouterNodes` had `if (!ethersUtils.isAddress(token.addressOrDenom)) continue` which gracefully skipped tokens with invalid EVM addresses. Now `ethersUtils.getAddress()` throws on malformed input, and there is no try/catch around `normalizeNodeAddress` or its caller. A single bad registry entry would crash the entire monitor before polling starts.

Consider re-adding the `isAddress` guard before calling `getAddress()`, or wrapping the token loop body in try/catch with a log-and-skip:
```ts
for (const token of warpCore.tokens) {
  try {
    // ... existing body ...
  } catch (error) {
    logger.error({ chain: token.chainName, address: token.addressOrDenom, error },
      'Skipping token with invalid address');
    continue;
  }
}
```
Same concern applies to `normalizeTokenAddress` (line 659) for malformed `collateralAddressOrDenom`.

paulbalaji (commented):
## Consolidated Review

Two medium findings, one minor. All prior review items (bytea corruption, inventory leak to non-EVM, type casts) are resolved in current head.

### Medium

1. **`fileURLToPath` entry guard is a startup regression** — `service.ts:118`. The old code unconditionally called `main()`. The new guard compares `fileURLToPath(import.meta.url)` (always absolute, resolved through symlinks) against `process.argv[1]` (whatever was passed to `node`). On macOS this mismatch was reproduced locally (`/private/tmp/...` vs `/var/tmp/...`). Both `start` and `start:dev` in `package.json` invoke this file directly, so a false negative silently prevents the service from starting. Either revert to unconditional execution and export `main` for testability, or canonicalize both sides with `realpathSync`.

2. **Malformed EVM address now crashes the monitor** — `monitor.ts:650/659`. The old `buildRouterNodes` had `if (!ethersUtils.isAddress(token.addressOrDenom)) continue` which gracefully skipped bad tokens. Now `normalizeNodeAddress`/`normalizeTokenAddress` call `ethersUtils.getAddress()` directly, which throws. One bad registry entry takes down the entire monitor instead of degrading gracefully. Consider re-adding the `isAddress` guard before normalization, or wrapping in try/catch with a log-and-skip.

### Low

3. **Trim `INVENTORY_ADDRESS` on read** — `service.ts:59`. `INVENTORY_ADDRESS_<PROTOCOL>` values get `.trim()` in `getConfiguredInventoryAddress`, but legacy `INVENTORY_ADDRESS` is used raw. A stray whitespace breaks only the EVM fallback path. (Echoing CodeRabbit's still-open suggestion: `process.env.INVENTORY_ADDRESS?.trim() || undefined`)

typescript/warp-monitor/src/service.ts
nambrot-agent: Fixed in d88c7dd6fc: the direct-entry guard now compares realpathSync(fileURLToPath(import.meta.url)) with realpathSync(process.argv[1]), so symlinked node entrypoints resolve to the same canonical path before deciding whether to start main().

typescript/warp-monitor/src/service.ts
```diff
@@ -54,6 +75,7 @@ async function main(): Promise<void> {
   const coingeckoApiKey = process.env.COINGECKO_API_KEY;
   const explorerApiUrl = process.env.EXPLORER_API_URL;
   const inventoryAddress = process.env.INVENTORY_ADDRESS;
+  const inventoryAddressesByProtocol = parseInventoryAddressesByProtocol();
```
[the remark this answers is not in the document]
nambrot-agent: Fixed in d88c7dd6fc: legacy INVENTORY_ADDRESS is now trimmed on read and empty-after-trim is treated as unset before it is logged or passed into WarpMonitor.

typescript/warp-monitor/src/monitor.ts
nambrot-agent: Fixed in d88c7dd6fc: buildRouterNodes now preserves the old skip behavior for malformed EVM router addresses and also skips malformed EVM collateral addresses before normalizeNodeAddress/normalizeTokenAddress call getAddress(). Added malformed router and collateral coverage in monitor.test.ts.

typescript/warp-monitor/src/service.ts
nambrot-agent: Follow-up in bb8076b856: removed the direct-entry guard entirely to match the sibling service entrypoint style. service.ts now calls main() unconditionally again, so the symlink mismatch concern no longer applies.

typescript/warp-monitor/src/service.ts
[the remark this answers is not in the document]
nambrot-agent: Follow-up in 6235b8c224: centralized this in getConfiguredInventoryAddress instead. service.ts now only passes the raw INVENTORY_ADDRESS through; getConfiguredInventoryAddress owns protocol-specific lookup, EVM-only fallback selection, trim, and empty-as-unset handling. Added test coverage for trimming the legacy fallback while still not applying it to Sealevel nodes.

typescript/warp-monitor/src/monitor.ts
```diff
@@ -385,6 +398,29 @@ export class WarpMonitor {
...
   }
 
+  private getConfiguredInventoryAddress(
+    node: RouterNodeMetadata,
+    inventoryAddress?: string,
+  ): string | undefined {
+    const protocolInventoryAddress =
+      process.env[
+        `INVENTORY_ADDRESS_${node.token.protocol.toUpperCase()}`
+      ]?.trim();
```
paulbalaji, line 408: Nit: This reads `process.env` on every poll cycle (once per node per cycle). It works and is useful for hot-reload, but it means protocol inventory config bypasses the `service.ts` config boundary where everything else is validated at startup. If a malformed `INVENTORY_ADDRESS_SEALEVEL` is set, the operator only finds out via repeated `getBalance` errors in logs rather than a startup failure.

Non-blocking, but consider reading these once at startup in `service.ts` and passing a `Record<ProtocolType, string>` through config in a follow-up.

typescript/warp-monitor/src/monitor.test.ts
```diff
@@ -255,6 +459,167 @@ describe('WarpMonitor', () => {
...
+
+    await invokeUpdatePendingAndInventoryMetrics(
+      monitor,
+      { multiProvider: {} } as WarpCore,
+      routerNodes,
+      new Map(),
+      'CROSS/ctusd',
+      pendingTransfersClient as ExplorerPendingTransfersClient,
+      200,
+    );
```
paulbalaji, line 537: Nit: This test sets protocol-specific env vars but doesn't pass an `inventoryAddress` fallback argument. It proves env vars work when the fallback is absent, but doesn't prove they take **precedence** over a fallback. Consider adding a variant that passes both and asserts the env var wins:
```ts
await invokeUpdatePendingAndInventoryMetrics(
  monitor, { multiProvider: {} } as WarpCore, routerNodes, new Map(),
  'CROSS/ctusd', pendingTransfersClient as ExplorerPendingTransfersClient,
  200,
  '0xShouldNotBeUsed',  // fallback should be overridden by env vars
);
```

typescript/warp-monitor/src/monitor.ts
```diff
@@ -577,12 +613,19 @@ export class WarpMonitor {
...
+      if (token.protocol === ProtocolType.Ethereum) {
+        const tokenAddress =
+          token.collateralAddressOrDenom ?? token.addressOrDenom;
+        if (
+          !ethersUtils.isAddress(token.addressOrDenom) ||
+          !ethersUtils.isAddress(tokenAddress)
+        ) {
+          continue;
+        }
+      }
```
paulbalaji, line 625: Nit: EVM tokens get `isAddress` validation before entering the node set. Non-EVM tokens pass through with no validation. A corrupt or empty `addressOrDenom` from the registry would be silently included and cause downstream adapter failures. Low probability, but a basic `token.addressOrDenom?.length > 0` guard would be consistent with the defensive pattern here.

paulbalaji (approved):
## Consolidated Review

All previously raised issues (bytea corruption, inventory leak to non-EVM, `as ProtocolType` casts, `fileURLToPath` guard, malformed EVM address crash, `INVENTORY_ADDRESS` trimming) are resolved in current head. Tests pass. Core architecture is sound: separating `routerNodes` from `explorerRouterNodes`, per-protocol inventory via env vars, and preserving non-EVM address casing.

### Suggestions (non-blocking)

See inline comments. The main theme is coverage depth rather than correctness:
- Protocol-specific env var precedence over legacy fallback isn't explicitly tested
- Only Ethereum and Sealevel protocols are exercised; other protocols (Cosmos, Starknet, Radix) follow the same non-EVM path but have zero coverage
- `INVENTORY_ADDRESS_<PROTOCOL>` naming for compound protocol types (e.g. `COSMOSNATIVE`) could use a note in the env var docs

typescript/warp-monitor/src/monitor.test.ts
```diff
@@ -68,8 +72,35 @@ async function invokeUpdatePendingAndInventoryMetrics(
...
+}
+
+function invokeBuildExplorerRouterNodes(
+  monitor: WarpMonitor,
+  routerNodes: RouterNodeMetadata[],
+): RouterNodeMetadata[] {
+  const buildExplorerRouterNodes = (monitor as any)
+    .buildExplorerRouterNodes as (
+    routerNodes: RouterNodeMetadata[],
+  ) => RouterNodeMetadata[];
```
xeno097, line 95: is there a way we can avoid all these typecasts?

PR #8555, opened 2026-04-09
fix(sdk): prevent transfers where recipient is the destination router or collateral token

### LINKED ISSUE #4964: WarpCore shouldn't allow for recipients to be the remote router or remote collateral token
## Problem

Sadly recently someone used the UI to do a Stride -> Eclipse transfer, and they set the recipient as the mint address on the Eclipse side

## Solution

In validateTransfer, we should require the recipient is not the destination router or token address

## Nice to Have

Aboudjem:
Fixes #4964.

## Problem

`WarpCore.validateTransfer` did not check whether the recipient address matches the destination router contract or its underlying collateral token. Transferring to either of those addresses locks the funds permanently — the router can't process a message it sent to itself, and the collateral token has no recovery path.

## Fix

Added `validateRecipientNotRouter` called inside `validateTransfer` after the destination token is resolved. It compares the recipient against:
- `destinationToken.addressOrDenom` (the router contract)
- `destinationToken.collateralAddressOrDenom` (the collateral ERC-20, when present)

Uses `eqAddress` from `@hyperlane-xyz/utils` for protocol-aware case-insensitive comparison.

Returns `{ recipient: '...' }` so the error surfaces in the same field as other recipient errors.

## Tests

Added a test case in `WarpCore.test.ts` that:
- Fails when recipient equals the destination router address
- Fails when recipient equals the destination collateral token address (simulated via temporary property override, consistent with existing test patterns)

### DIFF
--- a/.changeset/warm-jets-sing.md
+++ b/.changeset/warm-jets-sing.md
@@ -1,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': patch
+---
+
+Rejected WarpCore transfers when the recipient equals the destination router or collateral token address, preventing permanent fund loss.
--- a/typescript/sdk/src/warp/WarpCore.test.ts
+++ b/typescript/sdk/src/warp/WarpCore.test.ts
@@ -399,6 +399,55 @@ describe('WarpCore', () => {
     });
   });
 
+  it('Rejects transfers where recipient is the destination router or collateral contract', async () => {
+    const balanceStubs = warpCore.tokens.map((t) =>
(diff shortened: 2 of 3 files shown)


Aboudjem:
Addressed bot feedback in 1e5887c:

- @coderabbitai: wrapped the test body in try/finally so balanceStubs/quoteStubs restoration and the collateralAddressOrDenom reset always run, even if assertions fail.
- @devin-ai-integration: added missing changeset (.changeset/warm-jets-sing.md, sdk patch).

Ready for human review when someone has a moment.

PR #8567, opened 2026-04-11
fix: sealevel signer compatibility with web3 transactions

nambrot-agent:
## Description
- normalized web3-style Sealevel transactions at the `svm-sdk` signer boundary so web3-shaped `programId` / `keys[].pubkey` instructions can be serialized and signed by the kit-based signer path
- preserved the existing Sealevel adapter contract by keeping adapter-side `partialSign` behavior and bridging typed `extraSigners` through the CLI/signer boundary instead of mutating raw web3 transactions with a custom field
- added a real EVM/SVM `warp send` CLI E2E that fails on clean `main` and passes on this branch

## Scope
- `#8567` is signer-path only: web3 tx normalization, typed extra-signer handoff, and regression coverage
- Sealevel receipt/log polling hardening was moved to stacked PR `#8600`

## Related Issues
- fixes the Solana-origin `warp send` incompatibility where adapter-built web3.js transactions reached the shared Sealevel signer path unnormalized
- fixes the extra-signer regression when that signer path rebuilds normalized web3 transactions

## Backward Compatibility
- existing Sealevel adapter callers still get standard partially-signed web3 transactions
- signer compatibility is widened to also accept web3-shaped Sealevel transactions with typed `extraSigners`
- exported compute-budget helper behavior remains backward-compatible for zero-arg callers

## Testing
- `pnpm -C typescript/svm-sdk exec mocha --require tsx --timeout 10000 src/tests/signer.unit-test.ts`
- `pnpm -C typescript/cli exec mocha --config .mocharc.json src/send/transfer.test.ts`
- `CLI_E2E_TEST=warp-send-evm-svm pnpm -C typescript/cli test:cross-chain:e2e`

### DIFF
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,0 +1,6 @@
+---
+"@hyperlane-xyz/cli": patch
+"@hyperlane-xyz/sdk": patch
+---
+
+The registry dependency was bumped to 24.3.0 so CLI and SDK consumers pick up tolerant merged-registry overlay reads.
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,5 +1,0 @@
----
-"@hyperlane-xyz/cli": patch
----
-
-The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,6 +1,5 @@
 ---
 "@hyperlane-xyz/cli": patch
-"@hyperlane-xyz/sdk": patch
 ---
 
-The registry dependency was bumped to 24.3.0 so CLI and SDK consumers pick up tolerant merged-registry overlay reads.
+The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -504,7 +504,11 @@ async function executeDelivery({
           `Expected AnnotatedTx for non-EVM transfer execution, got ${typeof tx.transaction}`,
         );
       }
-      const txReceipt = await signer.sendAndConfirmTransaction(tx.transaction);
+      const txReceipt = await signer.sendAndConfirmTransaction(
+        tx.type === ProviderType.SolanaWeb3 && 'extraSigners' in tx
+          ? { ...tx.transaction, extraSigners: tx.extraSigners }
+          : tx.transaction,
+      );
       const typedReceipt =
         tx.type === ProviderType.SolanaWeb3
           ? await fetchSealevelReceiptWithLogs(
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -102,15 +102,22 @@ function toTypedAltVmReceipt(
 }
 
 export async function fetchSealevelReceiptWithLogs(
-  context: Pick<WriteCommandContext, 'multiProvider'>,
+  context: Pick<WriteCommandContext, 'multiProtocolProvider'>,
   origin: ChainName,
   signature: string,
   pollIntervalMs = SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
   maxAttempts = SEALEVEL_RECEIPT_MAX_ATTEMPTS,
 ): Promise<TypedTransactionReceipt> {
-  assert(Number.isFinite(pollIntervalMs) && pollIntervalMs >= 0);
-  assert(Number.isInteger(maxAttempts) && maxAttempts > 0);
-  const connection = context.multiProvider.getSolanaWeb3Provider(origin);
+  assert(
+    Number.isFinite(pollIntervalMs) && pollIntervalMs >= 0,
+    `Invalid Sealevel receipt poll interval: ${pollIntervalMs}`,
+  );
+  assert(
+    Number.isInteger(maxAttempts) && maxAttempts > 0,
+    `Invalid Sealevel receipt maxAttempts: ${maxAttempts}`,
+  );
+  const connection =
+    context.multiProtocolProvider.getSolanaWeb3Provider(origin);
   let receipt = null;
   let lastError: unknown;
 
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -108,14 +108,22 @@ export async function fetchSealevelReceiptWithLogs(
   pollIntervalMs = SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
   maxAttempts = SEALEVEL_RECEIPT_MAX_ATTEMPTS,
 ): Promise<TypedTransactionReceipt> {
+  assert(Number.isFinite(pollIntervalMs) && pollIntervalMs >= 0);
+  assert(Number.isInteger(maxAttempts) && maxAttempts > 0);
   const connection = context.multiProvider.getSolanaWeb3Provider(origin);
   let receipt = null;
+  let lastError: unknown;
 
   for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
-    receipt = await connection.getTransaction(signature, {
-      commitment: 'confirmed',
-      maxSupportedTransactionVersion: 0,
-    });
+    try {
+      receipt = await connection.getTransaction(signature, {
+        commitment: 'confirmed',
+        maxSupportedTransactionVersion: 0,
+      });
+      lastError = undefined;
+    } catch (error) {
+      lastError = error;
+    }
 
     if (receipt?.meta?.logMessages?.length) {
       return {
@@ -129,8 +137,12 @@ export async function fetchSealevelReceiptWithLogs(
     }
   }
 
+  const suffix =
+    lastError === undefined
+      ? ''
+      : ` (last getTransaction error: ${lastError instanceof Error ? lastError.message : String(lastError)})`;
   throw new Error(
-    `Transaction logs unavailable for Solana transaction ${signature}`,
+    `Transaction logs unavailable for Solana transaction ${signature}${suffix}`,
   );
 }
 
@@ -499,6 +511,13 @@ async function executeDelivery({
               context,
               origin,
               txReceipt.signature,
+              SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
+              Math.max(
+                SEALEVEL_RECEIPT_MAX_ATTEMPTS,
+                Math.ceil(
+                  (timeoutSec * 1000) / SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
+                ),
+              ),
             )
           : toTypedAltVmReceipt(tx.type, txReceipt);
       txReceipts.push(typedReceipt);
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -69,6 +69,8 @@ const EXPLORER_GRAPHQL_URL =
   'https://explorer4.hasura.app/v1/graphql';
 const EXPLORER_POLL_INTERVAL_MS = 5000;
 const EXPLORER_NO_RESULT_FALLBACK_COUNT = 3;
+const SEALEVEL_RECEIPT_POLL_INTERVAL_MS = 1000;
+const SEALEVEL_RECEIPT_MAX_ATTEMPTS = 10;
 
 function isAnnotatedTx(value: unknown): value is AnnotatedTx {
   return typeof value === 'object' && value !== null;
@@ -97,6 +99,39 @@ function toTypedAltVmReceipt(
         `Unsupported provider type for non-EVM transfer execution: ${providerType}`,
       );
   }
+}
+
+export async function fetchSealevelReceiptWithLogs(
+  context: Pick<WriteCommandContext, 'multiProvider'>,
+  origin: ChainName,
+  signature: string,
+  pollIntervalMs = SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
+  maxAttempts = SEALEVEL_RECEIPT_MAX_ATTEMPTS,
+): Promise<TypedTransactionReceipt> {
+  const connection = context.multiProvider.getSolanaWeb3Provider(origin);
+  let receipt = null;
+
+  for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
+    receipt = await connection.getTransaction(signature, {
+      commitment: 'confirmed',
+      maxSupportedTransactionVersion: 0,
+    });
+
+    if (receipt?.meta?.logMessages?.length) {
+      return {
+        type: ProviderType.SolanaWeb3,
+        receipt,
+      };
+    }
+
+    if (attempt < maxAttempts - 1) {
+      await sleep(pollIntervalMs);
+    }
+  }
+
+  throw new Error(
+    `Transaction logs unavailable for Solana transaction ${signature}`,
+  );
 }
 
 export async function sendTestTransfer({
@@ -458,7 +493,14 @@ async function executeDelivery({
         );
       }
       const txReceipt = await signer.sendAndConfirmTransaction(tx.transaction);
-      const typedReceipt = toTypedAltVmReceipt(tx.type, txReceipt);
+      const typedReceipt =
+        tx.type === ProviderType.SolanaWeb3
+          ? await fetchSealevelReceiptWithLogs(
+              context,
+              origin,
+              txReceipt.signature,
+            )
+          : toTypedAltVmReceipt(tx.type, txReceipt);
       txReceipts.push(typedReceipt);
       if (tx.category === WarpTxCategory.Transfer) {
         transferReceipt = typedReceipt;
--- a/typescript/cli/src/send/transfer.ts
+++ b/typescript/cli/src/send/transfer.ts
@@ -153,6 +153,55 @@ export async function fetchSealevelReceiptWithLogs(
   );
 }
 
+type AltVmTransferSigner = {
+  sendAndConfirmTransaction(transaction: unknown): Promise<TxReceipt>;
+};
+
+type ExecutableAltVmTransfer = {
+  type: ProviderType;
+  transaction: unknown;
+  extraSigners?: unknown[];
+};
+
+export async function submitAltVmTransferTx({
+  context,
+  signer,
+  origin,
+  tx,
+  timeoutSec,
+}: {
+  context: Pick<WriteCommandContext, 'multiProtocolProvider'>;
+  signer: AltVmTransferSigner;
+  origin: ChainName;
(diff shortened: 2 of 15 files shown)


typescript/sdk/src/token/adapters/SealevelCrossCollateralAdapter.ts
```diff
@@ -514,7 +514,13 @@ export class SealevelHypCrossCollateralAdapter
...
 
     if (randomWallet) {
-      tx.partialSign(randomWallet);
+      Object.assign(tx, {
+        additionalSigners: [
+          ...(((tx as Transaction & { additionalSigners?: Keypair[] })
+            .additionalSigners as Keypair[] | undefined) ?? []),
+          randomWallet,
+        ],
+      });
```
paulbalaji: **High: `additionalSigners` is not consumed by the active signing pipeline**

The previous code called `tx.partialSign(randomWallet)`, which is what `SvmMultiProtocolSignerAdapter.signAndConfirm()` expects — it re-signs the tx but preserves existing `partialSign` signatures. The new code attaches a non-standard `additionalSigners` property via `Object.assign`, but no in-repo consumer reads that field from a web3 `Transaction` object.

The active flow is: `WarpCore.transfer()` → `populateTransferRemoteToTx()` → `submitSvmTransaction()` → `SvmMultiProtocolSignerAdapter.signAndConfirm()` (in `solana-web3js.ts`), which threads extra signers through wrapper-level `extraSigners`, not `tx.additionalSigners`. So:
1. The `randomWallet` signature is silently lost for any caller going through the existing pipeline.
2. The bandaid casts + dead-code spread (the `tx` is freshly constructed above and never has `additionalSigners`) add complexity for no benefit.

Simplest fix: keep `tx.partialSign(randomWallet)` to match `SealevelTokenAdapter.ts:521` and the existing signing pipeline. If the intent is to route through `SvmSigner.send()` instead, that’s a larger change that should update both adapters and the pipeline together.

typescript/cli/src/send/transfer.ts
```diff
@@ -99,6 +101,39 @@ function toTypedAltVmReceipt(
...
+  maxAttempts = SEALEVEL_RECEIPT_MAX_ATTEMPTS,
+): Promise<TypedTransactionReceipt> {
+  const connection = context.multiProvider.getSolanaWeb3Provider(origin);
+  let receipt = null;
+
+  for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
+    receipt = await connection.getTransaction(signature, {
+      commitment: 'confirmed',
+      maxSupportedTransactionVersion: 0,
+    });
```
paulbalaji, line 119: **High: Unguarded RPC call inside poll loop**

`connection.getTransaction()` can throw on transient RPC failures (network timeout, 503, rate limit). A single throw here aborts the entire `fetchSealevelReceiptWithLogs` call and fails the transfer, even though the next poll attempt would likely succeed.

Wrap the call in try/catch and continue the loop on error:
```ts
for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
  try {
    receipt = await connection.getTransaction(signature, {
      commitment: 'confirmed',
      maxSupportedTransactionVersion: 0,
    });
  } catch {
    // transient RPC failure — retry on next attempt
    receipt = null;
  }
  // ... rest of loop
}
```

typescript/cli/src/send/transfer.ts
```diff
@@ -458,7 +493,14 @@ async function executeDelivery({
...
       const txReceipt = await signer.sendAndConfirmTransaction(tx.transaction);
-      const typedReceipt = toTypedAltVmReceipt(tx.type, txReceipt);
+      const typedReceipt =
+        tx.type === ProviderType.SolanaWeb3
+          ? await fetchSealevelReceiptWithLogs(
+              context,
+              origin,
+              txReceipt.signature,
+            )
+          : toTypedAltVmReceipt(tx.type, txReceipt);
```
paulbalaji: **High: Polling budget (10s) is much shorter than the delivery timeout**

This call uses the defaults (1s × 10 = 10s). The surrounding `executeDelivery` has `timeoutSec` (typically 120s). On congested Solana RPCs, logs can take longer than 10s to appear, so a successful send will incorrectly fail here.

Derive `maxAttempts` from the remaining time budget, e.g.:
```ts
? await fetchSealevelReceiptWithLogs(
    context,
    origin,
    txReceipt.signature,
    SEALEVEL_RECEIPT_POLL_INTERVAL_MS,
    Math.max(1, Math.ceil((timeoutSec * 1000) / SEALEVEL_RECEIPT_POLL_INTERVAL_MS)),
  )
```
(CodeRabbit raised this too — seconding it as a blocker.)

typescript/svm-sdk/src/tx.ts
```diff
@@ -19,6 +20,78 @@ import {
...
+  data?: Uint8Array;
+};
+
+export type Web3TransactionLike = {
+  instructions: readonly (SvmInstruction | Web3InstructionLike)[];
+  feePayer?: Address | Web3PublicKeyLike | null;
+  computeUnits?: number;
+  additionalSigners?: TransactionSigner[];
+  skipPreflight?: boolean;
+};
```
paulbalaji, line 50: **Medium: `Web3TransactionLike.additionalSigners` excludes web3 `Keypair`**

This type declares `additionalSigners?: TransactionSigner[]`, but the actual runtime use case (and the PR’s own tests at `signer.unit-test.ts:964`) passes web3 `Keypair` objects which are NOT `TransactionSigner`.

`signer.ts:normalizeAdditionalSigners()` handles the conversion at runtime, but the *type* doesn’t allow it. This forces tests to escape through `as unknown as SvmTransaction` double-casts (lines 894, 933, 964) and means external callers get no type-level guidance that web3 Keypairs are accepted.

Consider:
```ts
additionalSigners?: (TransactionSigner | { publicKey: { toBase58(): string }; secretKey: Uint8Array })[];
```
or a union type that includes the web3 Keypair shape. This would eliminate all three test double-casts and make the public API honest about what it accepts.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -305,7 +334,11 @@ export class SvmSigner
    * Sends a transaction and polls for confirmation. On blockhash expiry,
    * checks transaction history before resubmitting to prevent double-execution.
    */
-  async send(tx: SendableSvmTransaction): Promise<SvmReceipt> {
+  async send(tx: SendableSvmCompatTransaction): Promise<SvmReceipt> {
+    const normalizedTx = normalizeTransaction(tx);
+    normalizedTx.additionalSigners = await normalizeAdditionalSigners(
+      normalizedTx.additionalSigners,
+    );
```
paulbalaji: **Medium: `send()` accepts web3 shapes at runtime but not in the type signature**

`SendableSvmCompatTransaction` is `Parameters<typeof normalizeTransaction>[0]` which resolves to `SvmTransaction | Web3TransactionLike`. However, `Web3TransactionLike.additionalSigners` only accepts `TransactionSigner[]`, not web3 `Keypair`. So the type says "pass web3 instructions" but not "pass web3 keypairs as extra signers" — even though `normalizeAdditionalSigners` handles it.

Aligning the type with runtime behavior (see comment on `tx.ts:35-41`) would fix this and the test casts.

paulbalaji (commented):
## Review Summary

The normalization approach (type guards + idempotent web3 → @solana/kit conversion) is sound and solves a real interop problem. However there are three issues that should block merge, plus a typing gap that should be fixed.

### Findings

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **High** | `SealevelCrossCollateralAdapter.ts` | `additionalSigners` field is dead — no in-repo consumer reads it; signing pipeline still uses `partialSign` via `WarpCore` → `SvmMultiProtocolSignerAdapter` |
| 2 | **High** | `transfer.ts:115` | Single `getTransaction()` throw aborts poll loop — Solana RPCs transiently fail |
| 3 | **High** | `transfer.ts:498` | Default 10s poll budget vs surrounding `timeoutSec` (typically 120s) |
| 4 | **Medium** | `signer.ts` / `tx.ts` | Public TS types don’t expose web3 compat — tests must double-cast to exercise it |

### Validation

- **Passed:** `pnpm -C typescript/svm-sdk exec mocha --require tsx --timeout 10000 src/tests/signer.unit-test.ts`
- **Failed to run:** CLI test (`transfer.test.ts`) — missing `typescript/sdk/dist` build artifact in fresh workspace

### Existing Review State

Only bot feedback on record (7 unresolved inline threads, no author replies, no human review). We agree with CodeRabbit on polling budget and Devin on the adapter inconsistency. The test double-cast threads are symptoms of the real typing gap (finding #4).

### Observations Outside This PR

- `typescript/sdk/src/token/adapters/SealevelTokenAdapter.ts:521`: Still uses `tx.partialSign(randomWallet)` — same pattern this PR removes from the cross-collateral adapter. If the intent is to move to `additionalSigners`, both adapters should be updated together. If `partialSign` is correct for the current signing pipeline, this PR should keep it.

typescript/svm-sdk/src/tx.ts
```diff
@@ -63,6 +136,14 @@ export function getComputeBudgetInstructions(
   return instructions;
 }
 
+function hasComputeBudgetInstruction(
+  instructions: readonly SvmInstruction[],
+): boolean {
+  return instructions.some(
+    (instruction) => instruction.programAddress === COMPUTE_BUDGET_PROGRAM_ID,
+  );
+}
```
paulbalaji, line 163: **High: Compute budget dedup is too coarse — can silently drop CU limit or priority fee.**

`hasComputeBudgetInstruction` checks if *any* instruction targets the ComputeBudget program. If true, it skips *all* compute budget instructions — both `SetComputeUnitLimit` (discriminant `0x02`) and `SetComputeUnitPrice` (discriminant `0x03`).

This is not hypothetical: `SealevelIgpAdapter.populateClaimTx()` (`typescript/sdk/src/gas/adapters/SealevelIgpAdapter.ts:268-278`) builds transactions with only a `SetComputeUnitPrice` instruction and no `SetComputeUnitLimit`. When that tx flows through `buildTransactionMessage` or `transactionToInstructions`, the presence of the price ix causes the limit ix to be skipped, silently falling back to Solana's default 200k CU limit.

Conversely, a tx with only a `SetComputeUnitLimit` but no price ix would skip the priority fee.

Suggestion: check per-discriminant rather than per-program:
```ts
function hasSetComputeUnitLimit(instructions: readonly SvmInstruction[]): boolean {
  return instructions.some(
    (ix) => ix.programAddress === COMPUTE_BUDGET_PROGRAM_ID && ix.data?.[0] === 2,
  );
}
function hasSetComputeUnitPrice(instructions: readonly SvmInstruction[]): boolean {
  return instructions.some(
    (ix) => ix.programAddress === COMPUTE_BUDGET_PROGRAM_ID && ix.data?.[0] === 3,
  );
}
```
Then conditionally prepend only the missing instruction(s).

typescript/svm-sdk/src/tx.ts
```diff
@@ -80,11 +161,11 @@ export function buildTransactionMessage(params: {
...
 
-  const computeBudgetIxs = getComputeBudgetInstructions(
-    computeUnits,
-    priorityFeeMicroLamports,
-  );
-  const allInstructions = [...computeBudgetIxs, ...instructions];
+  const normalizedInstructions = instructions.map(normalizeInstruction);
+  const computeBudgetIxs = hasComputeBudgetInstruction(normalizedInstructions)
+    ? []
+    : getComputeBudgetInstructions(computeUnits, priorityFeeMicroLamports);
```
paulbalaji: Same issue as line 145 — this call site also uses `hasComputeBudgetInstruction` to skip *all* budget instructions. With per-discriminant checks, this would become:
```ts
const normalizedInstructions = instructions.map(normalizeInstruction);
const computeBudgetIxs = getComputeBudgetInstructions(
  hasSetComputeUnitLimit(normalizedInstructions) ? undefined : computeUnits,
  hasSetComputeUnitPrice(normalizedInstructions) ? undefined : priorityFeeMicroLamports,
);
```
(adjusting `getComputeBudgetInstructions` to skip each instruction when its param is `undefined`)

typescript/svm-sdk/src/tx.ts
```diff
@@ -98,9 +179,14 @@ export function buildTransactionMessage(params: {
...
 ): SvmInstruction[] {
+  const normalizedTx = normalizeTransaction(tx);
   const computeUnits = tx.computeUnits ?? DEFAULT_COMPUTE_UNITS;
-  const computeBudgetIxs = getComputeBudgetInstructions(computeUnits);
-  return [...computeBudgetIxs, ...tx.instructions];
+  const computeBudgetIxs = hasComputeBudgetInstruction(
+    normalizedTx.instructions,
+  )
+    ? []
+    : getComputeBudgetInstructions(computeUnits);
```
paulbalaji: Same coarse-grained check here in `transactionToInstructions`. Should use per-discriminant logic to match the fix in `buildTransactionMessage`.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -40,8 +41,11 @@ import type {
 import { SvmProvider } from './provider.js';
 import { DEFAULT_COMPUTE_UNITS } from '../constants.js';
 
-/** Transaction input for `send()` — feePayer is excluded since the signer provides it. */
 type SendableSvmTransaction = Omit<SvmTransaction, 'feePayer'>;
+type SendableSvmCompatTransaction = Parameters<typeof normalizeTransaction>[0];
+type SendableSvmExtraSignerTransaction = SendableSvmCompatTransaction & {
```
paulbalaji, line 48: **Medium: Web3 compat types are hidden from consumers.**

`SendableSvmExtraSignerTransaction` is private, derived via `Parameters<typeof normalizeTransaction>[0]` which resolves to `SvmTransaction | Web3TransactionLike`. But the class `implements AltVM.ISigner<SvmTransaction, SvmReceipt>`, so callers going through the interface only see `SvmTransaction`. Similarly, `transactionToPrintableJson` (line 219) still takes `AnnotatedSvmTransaction`.

Result: the web3 path works at runtime but typed callers can't use it without `as unknown as SvmTransaction` (as seen in tests at lines 895, 934, 965).

Non-blocking, but worth fixing: export `Web3TransactionLike` / `Web3InstructionLike` from the package surface and either widen the `ISigner` generic or add an explicit overload. `transactionToPrintableJson` should also accept the union.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -305,7 +338,12 @@ export class SvmSigner
    * Sends a transaction and polls for confirmation. On blockhash expiry,
    * checks transaction history before resubmitting to prevent double-execution.
    */
-  async send(tx: SendableSvmTransaction): Promise<SvmReceipt> {
+  async send(tx: SendableSvmExtraSignerTransaction): Promise<SvmReceipt> {
+    const compatAdditionalSigners = await normalizeAdditionalSigners(
+      tx.additionalSigners ?? tx.extraSigners,
```
paulbalaji: Nit: `tx.additionalSigners ?? tx.extraSigners` — if a tx has *both* fields, `extraSigners` is silently dropped. Worth a comment documenting precedence, or merging: `[...(tx.additionalSigners ?? []), ...(tx.extraSigners ?? [])]`.

paulbalaji (changes requested):
## Consolidated Review

Normalization approach is sound and solves a real interop problem. paulbalaji's prior blockers (#2 poll error handling, #3 polling budget) are fixed in the latest commits. One blocking issue remains, plus a non-blocking type surface gap.

### Existing review reconciliation

- **paulbalaji #1 (adapter `additionalSigners` dead)**: No longer applicable — this PR doesn't change the Sealevel adapters. Both adapters still use `partialSign()` on web3 `Transaction` objects. The svm-sdk normalization path is a parallel, additive mechanism.
- **paulbalaji #2 (poll loop error handling)**: **Fixed.** `getTransaction()` is now wrapped in try/catch with retry.
- **paulbalaji #3 (10s poll budget vs 120s timeout)**: **Fixed.** `maxAttempts` now scales with `timeoutSec`.
- **paulbalaji #4 (type gap)**: Still present — see finding #2 below.
- **Devin `multiProvider` complaint**: Fixed in-thread.
- **Remaining bot threads**: Cast/test-style noise, not blocking.

### Findings

| # | Severity | Finding |
|---|----------|--------|
| 1 | **High** | `hasComputeBudgetInstruction()` treats "any ComputeBudget ix exists" as "add none." If a web3 tx has only `SetComputeUnitPrice`, the CU limit is silently dropped (Solana default 200k used). This is not hypothetical — `SealevelIgpAdapter` (line 268-278) builds txs with only a price ix. See inline. |
| 2 | **Medium** | Runtime web3 compat is broader than TS surface. `send()` accepts `Web3TransactionLike` at runtime but the `ISigner<SvmTransaction>` interface and `transactionToPrintableJson(AnnotatedSvmTransaction)` don't reflect this. Tests require `as unknown as SvmTransaction`. Non-blocking but worth fixing in the same PR. See inline. |

### Validation

- `signer.unit-test.ts` passes on PR head (24 passing).
- `transfer.test.ts` could not be validated in this workspace (missing `@hyperlane-xyz/multicollateral` build artifact).

typescript/svm-sdk/src/tx.ts
nambrot-agent: Fixed in 0350fdbc4a. I replaced the coarse program-level dedupe with per-discriminant checks, so we now only skip the specific compute-budget instruction already present. Added a regression test covering the price-only case so the CU limit still gets prepended.

.changeset/registry-bump-cli-sdk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperlane-xyz/cli": patch
+---
+
+The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
```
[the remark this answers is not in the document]
nambrot-agent: Removed in 0350fdbc4a. The registry bump was unrelated to the signer-normalization fix, so I dropped the extra changeset and reverted the catalog/lock churn from this PR.

nambrot-agent:
Review cleanup:
- `transfer.test.ts` cast nits fixed in `c19260fd3c` by switching the new mocks to a narrow `Pick<WriteCommandContext, 'multiProtocolProvider'>` helper with a single documented test-only cast.
- `Web3TransactionLike.additionalSigners` now honestly includes web3 keypair shape, and the compat types are exported from the package surface in `90c8b54df2`.
- `getComputeBudgetInstructions()` zero-arg behavior was restored in `c19260fd3c`; it now preserves the old public helper semantics while still allowing explicit `undefined` at the internal dedupe call sites.
- Polling hardening comments were split out of this PR; that work now lives in stacked PR `#8600`.
- Re-ran: `pnpm -C typescript/svm-sdk exec tsc -p tsconfig.json --noEmit`, `pnpm -C typescript/svm-sdk exec mocha --require tsx --timeout 10000 src/tests/signer.unit-test.ts`, `pnpm -C typescript/cli exec mocha --config .mocharc.json src/send/transfer.test.ts`, and `CLI_E2E_TEST=warp-send-evm-svm pnpm -C typescript/cli test:cross-chain:e2e`.

PR #8568, opened 2026-04-12
fix(monitor): transfer detection for sealevel inflight messages

nambrot-agent:
Fix warp-monitor pending transfer detection for Sealevel-origin messages by dropping the sender-router filter from inflight transfer discovery. Adds a regression test for the Citrea router case where the sender is not the route router.\n\nValidation: targeted warp-monitor test execution is currently blocked in this checkout by dependency resolution issues in typescript/sdk / tron-sdk (missing @hyperlane-xyz/tron-sdk dist and then missing Node type defs during tron-sdk build).

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8568" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/warp-monitor/src/explorer.test.ts
+++ b/typescript/warp-monitor/src/explorer.test.ts
@@ -183,7 +183,7 @@ describe('Explorer Pending Transfers', () => {
     expect(thrown!.message).to.contain('GraphQL errors');
   });
 
-  it('counts sealevel-origin pending transfers even when sender is not a router', async () => {
+  it('counts sealevel-origin pending transfers without querying non-evm routers as recipients', async () => {
     const router = '0x00000000000000000000000000000000000000aa';
     const nodes: RouterNodeMetadata[] = [
       {
@@ -203,8 +203,8 @@ describe('Explorer Pending Transfers', () => {
           'USDC|solanamainnet|So11111111111111111111111111111111111111112',
         chainName: 'solanamainnet' as any,
         domainId: 1399811149,
-        routerAddress: '0x00000000000000000000000000000000000000cc',
-        tokenAddress: '0x00000000000000000000000000000000000000dd',
+        routerAddress: 'So11111111111111111111111111111111111111112',
+        tokenAddress: 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v',
         tokenName: 'USD Coin',
         tokenSymbol: 'USDC',
         tokenDecimals: 6,
@@ -260,7 +260,6 @@ describe('Explorer Pending Transfers', () => {
     expect(body.variables.originDomains).to.deep.equal([8453, 1399811149]);
     expect(body.variables.recipients).to.deep.equal([
       '\\x00000000000000000000000000000000000000aa',
-      '\\x00000000000000000000000000000000000000cc',
     ]);
   });
 });
--- a/typescript/warp-monitor/src/explorer.test.ts
+++ b/typescript/warp-monitor/src/explorer.test.ts
@@ -182,4 +182,85 @@ describe('Explorer Pending Transfers', () => {
     expect(thrown).to.not.equal(undefined);
     expect(thrown!.message).to.contain('GraphQL errors');
   });
+
+  it('counts sealevel-origin pending transfers even when sender is not a router', async () => {
+    const router = '0x00000000000000000000000000000000000000aa';
+    const nodes: RouterNodeMetadata[] = [
+      {
+        nodeId: 'USDC|base|0xrouter',
+        chainName: 'base' as any,
+        domainId: 8453,
+        routerAddress: router,
+        tokenAddress: '0x00000000000000000000000000000000000000bb',
+        tokenName: 'USD Coin',
+        tokenSymbol: 'USDC',
+        tokenDecimals: 6,
+        tokenScale: 1_000_000_000_000,
+        token: {} as any,
+      },
+      {
+        nodeId:
+          'USDC|solanamainnet|So11111111111111111111111111111111111111112',
+        chainName: 'solanamainnet' as any,
+        domainId: 1399811149,
+        routerAddress: '0x00000000000000000000000000000000000000cc',
+        tokenAddress: '0x00000000000000000000000000000000000000dd',
+        tokenName: 'USD Coin',
+        tokenSymbol: 'USDC',
+        tokenDecimals: 6,
+        tokenScale: 1_000_000_000_000,
+        token: {} as any,
+      },
+    ];
+
+    const amountCanonical18 = 2000000000000000000n;
+    const amountHex = amountCanonical18.toString(16).padStart(64, '0');
+    const recipientBytes32 =
+      '0000000000000000000000004444444444444444444444444444444444444444';
+    const messageBody = `0x${recipientBytes32}${amountHex}`;
+    const fetchStub = sinon.stub(globalThis, 'fetch' as any).resolves({
+      ok: true,
+      json: async () => ({
+        data: {
+          message_view: [
+            {
+              msg_id:
+                '\\xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee',
+              origin_domain_id: 1399811149,
+              destination_domain_id: 8453,
+              sender:
+                '\\xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
+              recipient:
+                '\\x00000000000000000000000000000000000000000000000000000000000000aa',
+              message_body: messageBody,
+              send_occurred_at: null,
+            },
+          ],
+        },
+      }),
+    } as any);
+
+    const client = new ExplorerPendingTransfersClient(
+      'https://explorer.example/v1/graphql',
+      nodes,
+      rootLogger,
+    );
+
+    const transfers = await client.getPendingDestinationTransfers();
+
+    expect(transfers).to.have.length(1);
+    expect(transfers[0].originDomainId).to.equal(1399811149);
+    expect(transfers[0].destinationNodeId).to.equal('USDC|base|0xrouter');
+    expect(transfers[0].amountBaseUnits).to.equal(2000000n);
+
+    const [, requestInit] = fetchStub.firstCall.args;
+    const body = JSON.parse(requestInit.body);
+    expect(body.query).to.not.contain('sender: { _in: $senders }');
+    expect(body.variables).to.not.have.property('senders');
+    expect(body.variables.originDomains).to.deep.equal([8453, 1399811149]);
+    expect(body.variables.recipients).to.deep.equal([
+      '\\x00000000000000000000000000000000000000aa',
+      '\\x00000000000000000000000000000000000000cc',
+    ]);
+  });
 });
--- a/typescript/warp-monitor/src/explorer.ts
+++ b/typescript/warp-monitor/src/explorer.ts
@@ -96,9 +96,11 @@ export class ExplorerPendingTransfersClient {
     this.nodeByDestinationKey = new Map<string, RouterNodeMetadata>();
 
     for (const node of nodes) {
+      domains.add(node.domainId);
+      if (!isValidAddressEvm(node.routerAddress)) continue;
+
       const routerLower = node.routerAddress.toLowerCase();
       routers.add(routerLower);
-      domains.add(node.domainId);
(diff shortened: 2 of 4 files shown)


paulbalaji:
i think you need to teach nambrot-agent to use conventional commits for PR titles :P

typescript/warp-monitor/src/explorer.ts
```diff
@@ -187,7 +187,6 @@ export class ExplorerPendingTransfersClient {
     if (this.routers.length === 0 || this.domains.length === 0) return [];
 
     const variables = {
-      senders: this.routers.map((router) => this.toBytea(router)),
       recipients: this.routers.map((router) => this.toBytea(router)),
       originDomains: this.domains,
```
paulbalaji, line 193: **This change alone doesn't fix Sealevel detection.** `this.domains` is derived from `routerNodes`, but `buildRouterNodes` (`monitor.ts:580`) drops all non-EVM tokens via `ethersUtils.isAddress()`. Sealevel domain IDs like `1399811149` never enter this list, so `origin_domain_id: { _in: $originDomains }` still excludes Sealevel-origin messages.

The sender filter removal is a necessary prerequisite, but `buildRouterNodes` also needs to emit domain IDs for non-EVM tokens for this to take effect in production.

typescript/warp-monitor/src/explorer.test.ts
```diff
@@ -173,4 +173,85 @@ describe('Explorer Pending Transfers', () => {
...
+        tokenDecimals: 6,
+        tokenScale: 1_000_000_000_000,
+        token: {} as any,
+      },
+      {
+        nodeId:
+          'USDC|solanamainnet|So11111111111111111111111111111111111111112',
+        chainName: 'solanamainnet' as any,
+        domainId: 1399811149,
+        routerAddress: '0x00000000000000000000000000000000000000cc',
```
paulbalaji: **This test exercises a setup that `buildRouterNodes` never produces.** In production, a Sealevel token like `So11111...` has a base58 `addressOrDenom` that fails `ethersUtils.isAddress()`, so `buildRouterNodes` skips it entirely. This test manually injects a Sealevel node with a fake `0x...cc` router address, which makes the `ExplorerPendingTransfersClient` include domain `1399811149` -- but that never happens in the real monitor.

Consider either:
1. Updating the test to reflect the actual production setup (where only EVM nodes are present), or
2. Landing the `buildRouterNodes` fix first so this test reflects reality.

typescript/warp-monitor/src/explorer.test.ts
```diff
@@ -173,4 +173,85 @@ describe('Explorer Pending Transfers', () => {
...
+    expect(transfers[0].originDomainId).to.equal(1399811149);
+    expect(transfers[0].destinationNodeId).to.equal('USDC|base|0xrouter');
+    expect(transfers[0].amountBaseUnits).to.equal(2000000n);
+
+    const [, requestInit] = fetchStub.firstCall.args;
+    const body = JSON.parse(requestInit.body);
+    expect(body.query).to.not.contain('sender: { _in: $senders }');
+    expect(body.variables).to.not.have.property('senders');
+    expect(body.variables.originDomains).to.deep.equal([8453, 1399811149]);
+    expect(body.variables.recipients).to.deep.equal([
```
paulbalaji, line 261: This assertion (`originDomains` includes `1399811149`) will never hold in production because the Sealevel domain is never added to the node set by `buildRouterNodes`. The test validates `ExplorerPendingTransfersClient` in isolation, which is fine as a unit test, but it doesn't demonstrate the end-to-end fix.

.changeset/registry-bump-cli-sdk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperlane-xyz/cli": patch
+---
+
+The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
```
paulbalaji: This registry bump (23.12.0 -> 24.3.0) was already merged via #8565. This changeset was consumed by the release bot in `49bdf9c6bd`. Merging this PR would re-add a stale changeset. Please rebase onto current `main` and drop the registry bump commits.

paulbalaji (commented):
## Review Summary

The core idea is right -- the `sender` filter does not work for Sealevel-origin warp transfers because the on-chain sender is not the router. However, **this PR does not fix the problem on the production path**. The query change is necessary but not sufficient.

### HIGH -- Fix is incomplete: Sealevel domains never enter the query

`buildRouterNodes` at `typescript/warp-monitor/src/monitor.ts:580` drops every token whose `addressOrDenom` is not a valid EVM address:

```typescript
if (!ethersUtils.isAddress(token.addressOrDenom)) continue;
```

Sealevel tokens have base58 addresses, so they are excluded from `routerNodes`. Since `ExplorerPendingTransfersClient` derives both `originDomains` and `destinationDomains` from the node set, Sealevel domain IDs (e.g. `1399811149`) never appear in the query variables. A Sealevel-to-EVM message with `origin_domain_id = 1399811149` still fails the `origin_domain_id: { _in: $originDomains }` filter.

The query-side sender removal is a prerequisite, but `buildRouterNodes` also needs to include non-EVM tokens (at least their domain IDs) for this fix to take effect.

### MEDIUM -- Removing sender filter globally widens the result set

After this change, the query matches any undelivered message whose recipient is a known router, regardless of sender. Client-side filtering (`parseWarpRouteMessage` + `isValidEvmWarpRecipient`) provides a second pass, but any message with a 64+ byte body starting with 24 zero bytes will pass through. On high-volume routes this can push real pending transfers past the `LIMIT 200` cap, causing the monitor to miss older stuck messages.

Consider whether a sender-side filter could be conditionally applied for EVM-origin domains only, or whether the limit should be raised/paginated.

### MEDIUM -- Registry bump is redundant

The 23.12.0 -> 24.3.0 registry bump was already merged via PR #8565 (commit `d34af7ccdd`). This branch was forked before that merge. The changeset file `.changeset/registry-bump-cli-sdk.md` was already consumed by the release bot and would be re-added on merge. Please rebase onto current `main` and drop the registry commits.

## Observations Outside This PR

- `typescript/warp-monitor/src/monitor.ts:580`: The `ethersUtils.isAddress()` gate in `buildRouterNodes` is the root cause preventing Sealevel monitoring. Fixing this (to include non-EVM token domains while still using EVM-formatted router addresses for the query) would make the query-side fix in this PR effective.

typescript/warp-monitor/src/explorer.ts
nambrot-agent: Addressed in the latest push. `buildRouterNodes` now preserves non-EVM token/router identifiers in `typescript/warp-monitor/src/monitor.ts`, so Sealevel domains reach the explorer query. `ExplorerPendingTransfersClient` was also updated to include all node domain IDs while keeping recipient matching EVM-only in `typescript/warp-monitor/src/explorer.ts`. After bootstrapping the workspace from root, focused warp-monitor validation ran against the real tests and passed for this path.

typescript/warp-monitor/src/explorer.test.ts
nambrot-agent: Updated in the latest push. The explorer test no longer invents an EVM-shaped Sealevel router. It now uses a production-shaped non-EVM router id in `typescript/warp-monitor/src/explorer.test.ts`, and the production-path regression is covered separately in `typescript/warp-monitor/src/monitor.test.ts` by verifying mixed EVM + non-EVM router-node construction.

.changeset/registry-bump-cli-sdk.md
nambrot-agent: Addressed in the latest force-push. I rebased the branch onto current `main`, dropped the old registry bump commits, and the PR diff no longer includes `.changeset/registry-bump-cli-sdk.md`, `pnpm-workspace.yaml`, or `pnpm-lock.yaml` changes from that bump.

typescript/warp-monitor/src/explorer.test.ts
nambrot-agent: Addressed. The coverage is now split so this is no longer relying on the isolated explorer assertion alone: `monitor.test.ts` verifies the production-shaped router-node set includes the Sealevel domain, and `explorer.test.ts` verifies the query behavior with non-EVM origins while keeping recipients EVM-only. I also reran focused warp-monitor tests after bootstrapping the workspace from root.

typescript/warp-monitor/src/explorer.ts
```diff
@@ -206,7 +206,6 @@ export class ExplorerPendingTransfersClient {
           where: {
             _and: [
               { is_delivered: { _eq: false } },
-              { sender: { _in: $senders } },
               { recipient: { _in: $recipients } },
```
paulbalaji, line 209: **MEDIUM: Sender filter removal enables phantom pending transfers.**

With the `sender: { _in: $senders }` filter gone, any undelivered mailbox message to a known EVM router with a warp-shaped body (64 bytes: 32-byte left-padded EVM recipient + 32-byte amount) will pass client-side validation and be counted as pending.

On-chain `Router.sol:104` enforces `_router == _sender`, so these messages will never deliver -- but the monitor reports them as real liabilities, and they can push genuine pending transfers past the `LIMIT 200`.

Consider re-adding the sender filter for EVM-origin domains only:
```graphql
_or: [
  { sender: { _in: $senders } },
  { origin_domain_id: { _nin: $evmOriginDomains } }
]
```
Or add a client-side sender check against enrolled routers post-query.

typescript/warp-monitor/src/monitor.ts
```diff
@@ -577,12 +578,10 @@ export class WarpMonitor {
         continue;
       }
       const metadata = chainMetadata[token.chainName];
-      if (!ethersUtils.isAddress(token.addressOrDenom)) continue;
-
       const domainId = metadata.domainId;
-      const routerAddress = ethersUtils
-        .getAddress(token.addressOrDenom)
-        .toLowerCase();
+      const routerAddress = this.normalizeRouterIdentifier(
```
paulbalaji, line 582: **MEDIUM: Non-EVM nodes will break inventory balance polling.**

`routerNodes` now includes Sealevel/Cosmos nodes, but the inventory loop at line 358 calls `adapter.getBalance(inventoryAddress)` on every node with a single EVM `INVENTORY_ADDRESS`. Non-EVM adapters will fail on this EVM address every poll cycle, producing repeated error logs without inventory metrics.

Suggested fix -- skip non-EVM nodes in the inventory loop:
```typescript
routerNodes.filter(n => isValidAddressEvm(n.routerAddress)).map(async (node) => {
```

paulbalaji (commented):
## Review Summary

The v2 push fixes paulbalaji's HIGH (Sealevel domains dropped by `buildRouterNodes`) and the stale registry bump. Two medium issues remain.

### MEDIUM: Sender filter removal enables phantom pending transfers

The query now matches any undelivered message to a known EVM router, regardless of sender. On-chain, `Router.sol:104` enforces `_router == _sender` so spam messages will never actually be delivered -- but the monitor doesn't know that. Any mailbox message with a warp-shaped body (32-byte recipient + 32-byte amount) to a warp router will be counted as a pending transfer.

The client-side filters (`parseWarpRouteMessage`, `isValidEvmWarpRecipient`) only check body format, not sender legitimacy. A crafted 64-byte body with 24 leading zero bytes in the recipient field passes both checks. On high-volume routes, these phantom rows can crowd real pending transfers past the LIMIT 200.

Consider re-adding the sender filter conditionally for EVM-origin domains only (where sender == router holds), or adding a client-side sender check against enrolled routers.

### MEDIUM: Non-EVM nodes in inventory balance loop

`INVENTORY_ADDRESS` (from `service.ts:56`) is a single EVM address. After this PR, `routerNodes` includes non-EVM nodes, so the inventory loop at `monitor.ts:358` will call `adapter.getBalance(inventoryAddress)` on Sealevel/Cosmos adapters with an EVM address. These will fail every poll cycle, emitting repeated error logs without producing inventory metrics.

Either skip non-EVM nodes in the inventory loop, or gate inventory tracking on `isValidAddressEvm(node.routerAddress)`.

### Other Notes

- Registry bump / stale changeset: confirmed clean, no longer in diff.
- `WarpDeployConfig` type alias: correct, tracks registry return type.
- `normalizeRouterIdentifier`: correct for EVM (checksum+lowercase) and non-EVM (passthrough).
- `warp-monitor` is `private: true` so no changeset needed.
- Tests cover the key scenarios but lack a Cosmos bech32 case (minor).

PR #8572, opened 2026-04-13
feat: added usdc igra rebalancer

troykessler:
### Description

This PR adds the rebalancer for the USDC/igra route

### Testing

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8572" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/infra/config/docker.ts
+++ b/typescript/infra/config/docker.ts
@@ -51,7 +51,7 @@ export const mainnetDockerTags: MainnetDockerTags = {
   // standalone services
   keyFunder: '3b17358-20260315-183126',
   warpMonitor: '3b17358-20260315-183126',
-  rebalancer: '3b17358-20260315-183126',
+  rebalancer: '1a19513-20260413-090011',
   feeQuoting: '12d899d-20260325-184337',
 };
 


typescript/infra/config/docker.ts
```diff
@@ -51,7 +51,7 @@ export const mainnetDockerTags: MainnetDockerTags = {
   // standalone services
   keyFunder: '3b17358-20260315-183126',
   warpMonitor: '3b17358-20260315-183126',
-  rebalancer: '3b17358-20260315-183126',
+  rebalancer: '1a19513-20260413-090011',
```
paulbalaji, line 54: fyi you only need to update the tag of these standalone services if there are _features/fixes_ i the logic you want, otherwise everything is just config

no issue updating the tag, just fyi in case you felt you had to update it

paulbalaji (approved):

typescript/infra/config/docker.ts
troykessler: ah good to know! will just update it this one then since the pr is already open but wont do it for config updates anymore

PR #8583, opened 2026-04-14
fix: batch deployer transactions and bump gas buffers

paulbalaji:
## Summary
- Added `submitBatched` utility to split large deployer transactions (hook configs, IGP configs, routing ISM enrollments) into chain-aware batch sizes to avoid exceeding block gas limits
- Refactored routing ISM deployment to deploy with an initial batch then enroll remaining domains individually, with per-chain initialization sizes
- Bumped gas buffer multiplier to 15% for ISM factory deployments
- Added configurable `minConfirmationTimeoutMs` option to `MultiProvider`
- Infra: bumped min confirmation timeout, added chain config tweaks

## Test plan
- [ ] Deploy warp routes / core on chains with lower gas limits (e.g. citrea, shibarium) to verify batching
- [ ] Verify routing ISM enrollment completes for large domain sets
- [ ] Check that existing deploy flows are unaffected on standard chains

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes on-chain deployment/config transaction shapes (batched submissions and routing ISM initialization/ownership flow), which can affect deploy correctness and partial-state retries on low-gas chains.
> 
> **Overview**
> Adds chain-aware *batched transaction submission* for large deploy-time config writes (IGP gas params/oracle configs and routing hook configs) via a new `submitBatched` utility with per-chain batch size overrides (e.g. `citrea`).
> 
> Refactors routing ISM deployment to avoid block gas limit issues by deploying with an initial domain batch (chain-specific sizing), enrolling remaining domains individually, increasing ISM factory gas buffers (15×), and skipping `transferOwnership` when the deployer is already the target owner.
> 
> Extends `MultiProvider` with configurable `minConfirmationTimeoutMs` (and sets it in infra scripts), widens `defaultEthersV5ProviderBuilder` retry override to `SmartProviderOptions` (enabling `fallbackStaggerMs`), improves HAAS checks with a smart-provider-backed `MultiProvider` and signer rebinding, adds secret-metadata override caching plus explorer API key injection, and caps HAAS chain check concurrency while updating `chainsToSkip`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ba80a197df3153abff5b139c863fc23ff812ef2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


typescript/infra/scripts/check/check-utils.ts
```diff
@@ -59,6 +61,57 @@ export function getCheckDeployArgs() {
...
+
+function reconnectSigner(signer: Signer, provider: providers.Provider): Signer {
+  try {
+    return signer.connect(provider);
+  } catch (error: unknown) {
+    const code =
+      typeof error === 'object' && error !== null && 'code' in error
+        ? String((error as Record<string, unknown>).code)
+        : undefined;
+    if (code === 'UNSUPPORTED_OPERATION') {
```
paulbalaji, line 84: Silent fallback: when `connect` throws `UNSUPPORTED_OPERATION`, the original signer is returned still bound to the OLD provider — so HAAS retry/fallback options don't apply to that chain. Low risk for a read-only check flow, but consider a `logger.warn({ chain }, 'Signer could not be reconnected to HAAS provider; smart-provider options not active')` so operators can tell when it happens.

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -59,6 +61,57 @@ export function getCheckDeployArgs() {
...
+        rpcUrls,
+        network,
+        HAAS_SMART_PROVIDER_OPTIONS,
+      ).provider,
+  });
+
+  if (baseMultiProvider.useSharedSigner) {
+    const sharedSigner = Object.values(baseMultiProvider.signers)[0];
+    if (sharedSigner) {
+      haasMultiProvider.setSharedSigner(sharedSigner);
```
paulbalaji, line 109: Shared-signer branch does no reconnect at all. `setSharedSigner` re-assigns the same signer object across every chain on the new MultiProvider, but the signer is still bound to whatever provider it was originally attached to. For signer-backed calls (`estimateGas`, `sendTransaction`) the HAAS smart-provider options won't take effect. If possible, mirror the per-chain reconnect loop for the shared-signer case (or at minimum document the gap).

typescript/infra/scripts/check/check-utils.ts
```diff
@@ -59,6 +61,57 @@ export function getCheckDeployArgs() {
 
 const ICA_ENABLED_MODULES = [Modules.INTERCHAIN_ACCOUNTS, Modules.HAAS];
 
+const HAAS_SMART_PROVIDER_OPTIONS = {
+  maxRetries: 4,
+  baseRetryDelayMs: 100,
+  fallbackStaggerMs: 2_000,
```
paulbalaji, line 67: `fallbackStaggerMs` is a `SmartProviderOptions` field, not `ProviderRetryOptions`. `defaultEthersV5ProviderBuilder`'s `retryOverride` param is typed as the latter, so this object is structurally wider than the declared type. Works at runtime (it propagates through) but the builder's signature lies about what it accepts. Widen the SDK param type or drop the field here.

paulbalaji (commented):
Consolidated review (no blockers).

## Summary

### Intentional (acknowledged, not flagged)
- Per-domain `set()` after initial factory deploy — the IncrementalDomainRoutingIsm only supports batch initialization; subsequent enrollments are one-by-one by design.
- Two per-chain batch-size tables (`CHAIN_BATCH_SIZE_OVERRIDES` vs `domainRoutingInitializationSize`) — intentional, different actions (calldata size vs init gas).

One suggestion on wording: the PR body says "batched routing ISM enrollments," which reads as if the full enrollment is batched. Since only the initial factory call is batched (by contract design) and the remainder is O(N) txs, consider clarifying the PR body (and changeset) to say "batched initial enrollment" or similar.

### Real issues (actionable)
1. **HAAS smart-provider wiring partially bypassed.** Two paths in `getHaasMultiProvider` can leave the old provider attached: `UNSUPPORTED_OPERATION` fallback in `reconnectSigner`, and the `useSharedSigner` branch which doesn't reconnect at all. Inline comments below.
2. **`submitBatched` non-atomic without docs.** Partial-failure mid-loop leaves on-chain state inconsistent. IGP/oracle pre-filter by on-chain state so retry is safe; hook routing `setHooks` re-submits every entry. Worth a JSDoc note on caller idempotency.
3. **`fetchExplorerApiKeys()` not memoized.** Hits GCP Secret Manager every `getSecretMetadataOverrides` invocation.
4. **Type signature mismatch for `fallbackStaggerMs`.** Passed through `defaultEthersV5ProviderBuilder`'s `ProviderRetryOptions` param but that type doesn't include the field. Works at runtime; widen the SDK param type.
5. **Redundant `transferOwnership` when `signerAddress === config.owner`.** One extra tx per deploy on the self-owned case.

### Test coverage gap
No new unit tests for `submitBatched` (trivially testable: empty, batch-size boundary, error propagation, chain override). Existing routing-ISM hardhat tests use 2–3 domains with `owner === signer`, so the new `initialBatchSize < domains.length` + `transferOwnership` handoff is untested. Also no test for `minConfirmationTimeoutMs`. Highest-leverage addition would be a hardhat routing-ISM test with a small `initialBatchSize` override and `owner !== signer`.

### Nits (optional)
- Gas buffer inconsistency: ISM factory paths use 15%, batched IGP/hook paths still use 10% default. If 15% is the new standard, consider updating `DEFAULT_GAS_LIMIT_BUFFER_PERCENT` instead of per-call-site.
- `submitBatched` imported into IGP deployer from `hook/utils.js` — the helper is protocol-agnostic; consider moving to a neutral module (`deploy/utils.ts`).
- `fluence`/`tangle` skip added without inline rationale (bucketed under `// downtime`).
- Dead commented `arbitrum`/`unichain` confirmation blocks in `mainnet3/chains.ts` — keep only if intentional placeholder.

typescript/infra/src/config/chain.ts
```diff
@@ -208,6 +211,26 @@ export async function getSecretMetadataOverrides(
     }
   }
 
+  // Merge explorer API keys into chain metadata so that block explorer
+  // verification checks (e.g. warp check) use authenticated requests.
+  // These overrides live only in the PartialRegistry layer and are never
+  // persisted back to the on-disk registry.
+  const explorerApiKeys = await fetchExplorerApiKeys();
```
paulbalaji, line 243: Not memoized — every `getSecretMetadataOverrides` call does a gcloud Secret Manager round-trip. If scripts construct a MultiProvider repeatedly this is a per-invocation cost. Consider a module-level cache (or lazy-initialized Promise) since explorer keys don't change across a single process run.

typescript/infra/src/config/chain.ts
```diff
@@ -65,6 +66,8 @@ export function getDisabledChains(): ChainName[] {
 export const chainsToSkip: ChainName[] = [
   // downtime
   'molten',
+  'fluence',
+  'tangle',
```
paulbalaji, line 70: nit: no inline rationale. `molten` is bucketed under `// downtime` — are `fluence`/`tangle` the same? A one-liner with the incident or ticket would help future readers figure out when to remove them.

typescript/sdk/src/hook/utils.ts
```diff
@@ -1,7 +1,46 @@
...
+
+const CHAIN_BATCH_SIZE_OVERRIDES: Partial<Record<ChainName, number>> = {
+  citrea: 16,
+};
+
+export function getHookTxBatchSize(chain: ChainName): number {
+  return CHAIN_BATCH_SIZE_OVERRIDES[chain] ?? DEFAULT_MAX_BATCH_SIZE;
+}
+
+export async function submitBatched<T>(
```
paulbalaji: Worth a JSDoc note: `submitBatched` is non-atomic. If batch N succeeds and batch N+1 fails, on-chain state is partially mutated. Callers must ensure `fn` is idempotent on retry. IGP/oracle callers pre-filter by on-chain state so retry is safe, but `HyperlaneHookDeployer.setHooks` pushes every configured destination regardless of on-chain state — re-running would re-submit the already-applied batches.

typescript/sdk/src/hook/utils.ts
```diff
@@ -1,7 +1,46 @@
...
 
 import { AggregationHookConfig, DerivedHookConfig, HookType } from './types.js';
 
+const DEFAULT_MAX_BATCH_SIZE = 64;
+
+const CHAIN_BATCH_SIZE_OVERRIDES: Partial<Record<ChainName, number>> = {
+  citrea: 16,
+};
+
+export function getHookTxBatchSize(chain: ChainName): number {
```
paulbalaji: nit: `getHookTxBatchSize` / `CHAIN_BATCH_SIZE_OVERRIDES` are hook-named but now used by `HyperlaneIgpDeployer` too. Since `submitBatched` is protocol-agnostic, consider moving both to a neutral location (e.g. `deploy/utils.ts`) or renaming to something like `getTxConfigBatchSize`.

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -654,6 +701,43 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
...
+          await this.multiProvider.handleTx(destination, enrollTx);
+        }
+
+        // Transfer ownership after all enrollments are complete
+        const transferTxEstimatedGas =
+          await routingIsm.estimateGas.transferOwnership(
+            config.owner,
+            overrides,
+          );
+        const transferTx = await routingIsm.transferOwnership(config.owner, {
```
paulbalaji: Skip when `eqAddress(signerAddress, config.owner)` — otherwise this issues a redundant `transferOwnership(signer → signer)` tx on every self-owned deploy.

typescript/sdk/src/gas/HyperlaneIgpDeployer.ts
```diff
@@ -15,6 +15,7 @@ import { TOKEN_EXCHANGE_RATE_SCALE_ETHEREUM } from '../consts/igp.js';
 import { HyperlaneContracts } from '../contracts/types.js';
 import { HyperlaneDeployer } from '../deploy/HyperlaneDeployer.js';
 import { ContractVerifier } from '../deploy/verify/ContractVerifier.js';
+import { submitBatched } from '../hook/utils.js';
```
paulbalaji: nit: cross-module import — IGP deployer depending on `hook/utils` is architecturally odd since `submitBatched` is hook-agnostic. Consider moving to `deploy/utils.ts` or similar.

typescript/infra/config/environments/mainnet3/chains.ts
```diff
@@ -140,6 +140,16 @@ export const chainMetadataOverrides: ChainMap<Partial<ChainMetadata>> = {
...
+  // arbitrum: {
+  //   blocks: {
+  //     confirmations: 3,
+  //   },
+  // },
+  // unichain: {
+  //   blocks: {
+  //     confirmations: 5,
+  //   },
+  // },
```
paulbalaji, line 152: nit: dead commented blocks. If these are intentional placeholders (to quickly re-enable during an incident), fine — otherwise drop.

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
```diff
@@ -416,22 +417,21 @@ export class HyperlaneHookDeployer extends HyperlaneDeployer<
...
-          chain,
-          routingHookAddress: routingHook.address,
-          routingConfigs,
-        },
-        'Setting routing hooks',
-      );
-      const estimatedGas =
-        await routingHook.estimateGas.setHooks(routingConfigs);
-      return this.multiProvider.handleTx(
+      await submitBatched(
```
paulbalaji, line 420: Related to the `submitBatched` non-atomicity note: unlike IGP/oracle callers which pre-filter `gasParamsToSet`/`configsToSet` by comparing against on-chain state, `routingConfigs` here is built from the desired config unconditionally. On a mid-loop failure retry, batches 1..N-1 get re-submitted even though they already applied. Not a correctness issue — just redundant gas — but worth noting.

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -654,6 +701,43 @@ export class HyperlaneIsmFactory extends HyperlaneApp<ProxyFactoryFactories> {
...
+          gasLimit: addBufferToGasLimit(transferTxEstimatedGas, 15),
+          ...overrides,
+        });
+        await this.multiProvider.handleTx(destination, transferTx);
```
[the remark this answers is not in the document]
paulbalaji: Fixed in 8118e49f9c — added `eqAddress(signerAddress, config.owner)` guard around the `transferOwnership` call so self-owned deploys skip the redundant tx.

typescript/infra/scripts/check/check-utils.ts
paulbalaji: Fixed in 8118e49f9c — `reconnectSigner` now emits `logger.warn({ chain }, 'Signer could not be reconnected to HAAS provider; smart-provider options not active')` when it hits `UNSUPPORTED_OPERATION`.

typescript/infra/scripts/check/check-utils.ts
paulbalaji: Fixed in 8118e49f9c — after `setSharedSigner`, each chain's entry in `haasMultiProvider.signers` is rebound to a signer connected to that chain's HAAS provider. Direct map mutation is required because `MultiProvider.setSigner` throws once `useSharedSigner` is true.

typescript/infra/scripts/check/check-utils.ts
paulbalaji: Fixed in 8118e49f9c — widened `defaultEthersV5ProviderBuilder`'s `retryOverride` param (and `ProviderBuilderFn`) from `ProviderRetryOptions` to `SmartProviderOptions`, so `fallbackStaggerMs` is now accepted by the type.

typescript/infra/src/config/chain.ts
paulbalaji: Fixed in 8118e49f9c — added a module-level `Map<cacheKey, Promise<...>>` keyed by `${deployEnv}:${sortedChains}`. Subsequent calls return the same in-flight / resolved promise; rejections are evicted so a retry re-fetches.

typescript/infra/src/config/chain.ts
paulbalaji: Accepting as a non-blocking nit — `fluence` and `tangle` fall under the same `// downtime` bucket as `molten`. Happy to add per-chain incident pointers in a follow-up if/when we have ticket links.

typescript/sdk/src/hook/utils.ts
paulbalaji: Done in 8118e49f9c — `submitBatched` moved to `sdk/deploy/utils.ts` with a JSDoc block calling out the non-atomic behavior and noting the pre-filter vs. redundant-resubmit tradeoff for IGP/oracle vs. hook routing callers.

typescript/sdk/src/hook/utils.ts
paulbalaji: Done in 8118e49f9c — moved to `sdk/deploy/utils.ts` and renamed to `getTxConfigBatchSize` (+ `CHAIN_BATCH_SIZE_OVERRIDES`).

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
paulbalaji: Fixed in 8118e49f9c — guarded the `transferOwnership` block with `if (!eqAddress(signerAddress, config.owner))`.

typescript/sdk/src/gas/HyperlaneIgpDeployer.ts
paulbalaji: Fixed in 8118e49f9c — `HyperlaneIgpDeployer` now imports `submitBatched` from `../deploy/utils.js` instead of `../hook/utils.js`.

typescript/infra/config/environments/mainnet3/chains.ts
paulbalaji: Intentional — keeping these as commented placeholders so we can quickly re-enable elevated confirmations during an incident without hunting for the chain name.

typescript/sdk/src/hook/HyperlaneHookDeployer.ts
paulbalaji: Covered by the JSDoc added to `submitBatched` in 8118e49f9c — the non-atomic retry semantics (and the asymmetry between IGP/oracle pre-filtered callers and the hook routing unfiltered caller) are called out explicitly there.

paulbalaji:
@claude review the changes but remember we've been using these changes for 18-24 months already

paulbalaji:
@claude thanks — fixed both import-grouping nits in 8ba80a197d:
- `HyperlaneHookDeployer.ts`: merged `submitBatched` into the existing `../` block.
- `infra/src/config/chain.ts`: moved `fetchExplorerApiKeys` into the `../` block.

Leaving the commented `arbitrum`/`unichain` confirmation blocks and the `fluence`/`tangle` downtime entries as-is — intentional placeholders / consistent with the existing `molten` pattern.

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
```diff
@@ -84,6 +84,42 @@ const ismFactories = {
...
+    destination === 'eni' ||
+    destination === 'merlin' ||
+    destination === 'megaeth' ||
+    destination === 'pulsechain'
+  ) {
+    return 120;
+  }
+
+  return 300;
+};
```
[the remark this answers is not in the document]
paulbalaji: this is fine for now until we have more generic batching configs in chain metadatas

xeno097 (approved):

typescript/sdk/src/ism/HyperlaneIsmFactory.ts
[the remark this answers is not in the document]
yorhodes: agree this is a pretty weird place for this

PR #8584, opened 2026-04-14
feat: add widget wallet address helpers

paulbalaji:
## Summary
- add reusable widget helpers for protocol-scoped wallet address lookup
- upstream the app-local wallet-address coordination pattern from Warp UI into widgets
- keep scope narrow: no consumer rewiring in this PR

## What changed
- add `@hyperlane-xyz/widgets/walletIntegrations/walletAddresses`
- export:
  - `getAddressForChain(...)`
  - `useWalletAddressesByProtocol(...)`
  - `useWalletAddressForChainAndProtocol(...)`
- duplicate Cosmos addresses onto `ProtocolType.CosmosNative` so Cosmos consumers can use the same helper path
- add a widgets patch changeset

## Why
Warp UI currently has local glue to resolve protocol-scoped wallet addresses without going through broader all-VM wrapper paths. This helper is generic enough to live in widgets instead, so downstream apps can share it.

This PR does **not** move the EVM-specific direct-balance optimization into widgets. That part remains app-specific.

## Validation
- `pnpm --filter @hyperlane-xyz/widgets build`
- `pnpm exec oxlint -c oxlint.json typescript/widgets/src/walletIntegrations/walletAddresses.tsx`
- `pnpm build`

---

> [!NOTE]
> **Low Risk**
> Adds new address-lookup utilities/hooks and lightly refactors existing address selection logic; behavior is covered by unit tests and is low-impact aside from potential edge-case differences in address fallback selection.
> 
> **Overview**
> Adds reusable wallet-address utilities to `@hyperlane-xyz/widgets` for protocol-scoped consumers, including `getAddressForChain` plus hooks to retrieve all wallet addresses by protocol or a single address for a given chain/protocol.
> 
> Refactors `getAccountAddressForChain` to delegate Cosmos vs non-Cosmos address selection to the new helper, exports the new APIs from `multiProtocol`, and adds unit tests for Cosmos/non-Cosmos fallback behavior. Updates `widgets` `tsconfig.json` to exclude `**/*.test.ts` from compilation and publishes a patch changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d44841cefc5fbacfc82c63f4d85f4efb219f90dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


typescript/widgets/src/walletIntegrations/walletAddresses.ts
```diff
@@ -0,0 +1,70 @@
...
+  ProtocolType.Sealevel,
+  ProtocolType.Cosmos,
+  ProtocolType.CosmosNative,
+  ProtocolType.Starknet,
+  ProtocolType.Radix,
+  ProtocolType.Aleo,
+  ProtocolType.Tron,
+];
+
+export function getAddressForChain(
```
Xaroz, line 21: we have a very similar function called `getAccountAddressForChain` is this utility function because we don't wanna deal with protocols or the accounts params?

paulbalaji: Yes, that is the reason.

`getAccountAddressForChain(...)` is still the right helper when the caller has the full `accounts` record and wants `multiProvider` to resolve the protocol for them.

This new helper is for the narrower case where the caller already has:
- a protocol-scoped `ChainAddress[]`
- the target `protocol`
- the target `chainName`

That lets protocol-scoped consumers avoid threading the whole `accounts` record and `multiProvider` shape through the call just to do the final address pick.

So the overlap is intentional, but the API shape is narrower for the protocol-scoped path rather than replacing `getAccountAddressForChain(...)`.

typescript/widgets/src/walletIntegrations/walletAddresses.ts
Xaroz, line 1: one thing I forgot to mention (and I see it now in other components) is that hooks files do not need to end in .tsx as you are not returning `jsx`, they can be simple .ts files

Xaroz: for example account.tsx -> account.ts, connectFns.tsx -> connectFns.ts...

typescript/widgets/src/walletIntegrations/walletAddresses.ts
Xaroz: is it really worth it to do this vs just using the other util function? you would need accounts to have the address right? and searching through the protocol and multiprovider to get the protocol seems like negligible,

Xaroz: I guess you don't need accounts to have address as in the UI you can paste the recipient but yeah the other concern still stands

paulbalaji: Fair pushback.

I cleaned up the part I agreed was not worth carrying in `d44841cefc`:
- there is now only one address-selection implementation
- `getAccountAddressForChain(...)` delegates to `getAddressForChain(...)`
- so we no longer have two separate resolution paths drifting

The remaining reason to keep `getAddressForChain(...)` is just that some protocol-scoped consumers already have:
- the `ChainAddress[]`
- the `protocol`
- the target `chainName`

In that case, routing back through `getAccountAddressForChain(...)` would mean re-threading `multiProvider + accounts` only to immediately peel them back apart again.

So the split now is:
- one shared core selector
- one broader wrapper for `multiProvider + accounts`
- one narrower wrapper/use-site for already-scoped address arrays

If we later decide the narrower wrapper still is not earning its keep, we can collapse that API too. But after `d44841cefc`, the important part is that there is only one actual selection logic path.

typescript/widgets/src/walletIntegrations/walletAddresses.ts
paulbalaji: Fixed in `d44841cefc`.

I renamed this helper file to `walletAddresses.ts` since it does not render JSX.

I did not widen scope into renaming the older `*.tsx` non-JSX hook files in this PR. That cleanup is broader and can be done separately if we want to normalize the rest of `walletIntegrations/`.

Xaroz (approved):

PR #8588, opened 2026-04-14
fix(sdk): connect mailbox with current signer in deliver()

tkporter:
## Summary
- `HyperlaneCore.deliver()` now reconnects the mailbox contract with the current signer from `MultiProvider` at call time, instead of relying on the stale connection from construction
- `status --relay` now exits non-zero when relay fails (previously silently swallowed errors and exited 0)
- Added e2e test that verifies `status --relay` actually delivers the message

## Context
`hyperlane status --relay` was failing with `"sending a transaction requires a signer"`. The destination chain signer is added lazily (after `HyperlaneCore` is constructed), but `deliver()` was using the mailbox contract as connected at construction time — which only had a read-only `Provider`.

## Test plan
- [x] `CLI_E2E_TEST=status pnpm -C typescript/cli test:ethereum:e2e` passes (6/6)
- [x] Manually verified relay of stuck messages to Story mainnet

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8588" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.changeset/fix-status-relay-signer.md
+++ b/.changeset/fix-status-relay-signer.md
@@ -1,0 +1,6 @@
+---
+'@hyperlane-xyz/sdk': patch
+'@hyperlane-xyz/cli': patch
+---
+
+Fixed `deliver()` and `sendMessage()` in HyperlaneCore to connect the mailbox with the current signer at call time, preventing "sending a transaction requires a signer" errors when signers are added after construction. The `status --relay` command now exits non-zero when relay fails.
--- a/typescript/cli/src/status/message.ts
+++ b/typescript/cli/src/status/message.ts
@@ -152,7 +152,16 @@ export async function checkMessageStatus({
         const merkleAddress = coreAddresses[origin].merkleTreeHook;
         stubMerkleTreeConfig(relayer, origin, hookAddress, merkleAddress);
       }
-      await relayer.relayAll(dispatchedReceipt, relayable);
+      // relayAll returns a receipt per message: either from the relay tx or,
+      // if the message was already delivered, from getProcessedReceipt.
+      const results = await relayer.relayAll(dispatchedReceipt, relayable);
+      const delivered = Object.values(results).flat().length;
+      const failed = relayable.length - delivered;
+      if (failed > 0) {
+        logRed(`Failed to relay ${failed}/${relayable.length} messages`);
+        process.exit(1);
+      }
+      logGreen(`Relayed ${delivered} message(s)`);
     } else if (undelivered.length > 0) {
       warnYellow(
         'No messages could be relayed - missing signers for all destination chains',
--- a/typescript/sdk/src/core/HyperlaneCore.ts
+++ b/typescript/sdk/src/core/HyperlaneCore.ts
@@ -150,7 +150,9 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     hook?: Address,
     metadata?: string,
   ): Promise<{ dispatchTx: TransactionReceipt; message: DispatchedMessage }> {
-    const mailbox = this.getContracts(origin).mailbox;
+    const mailbox = this.getContracts(origin).mailbox.connect(
+      this.multiProvider.getSigner(origin),
+    );
     const destinationDomain = this.multiProvider.getDomainId(destination);
     const recipientBytes32 = addressToBytes32(recipient);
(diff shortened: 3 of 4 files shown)


xeno097 (approved):

typescript/sdk/src/core/HyperlaneCore.ts
```diff
@@ -334,15 +334,14 @@ export class HyperlaneCore extends HyperlaneApp<CoreFactories> {
     ismMetadata: string,
   ): Promise<ethers.ContractReceipt> {
     const destinationChain = this.getDestination(message);
+    const mailbox = this.getContracts(destinationChain).mailbox.connect(
```
paulbalaji, line 339: Non-blocking: `sendMessage()` (~line 153) has the same stale-signer pattern — uses `this.getContracts(origin).mailbox` directly without reconnecting. Not a problem for the current CLI `send` path (signers attached before `HyperlaneCore` construction), but any future SDK consumer that constructs-then-attaches will hit the same bug.

Worth a follow-up one-liner to apply the same `.connect(this.multiProvider.getSigner(origin))` pattern there.

typescript/cli/src/status/message.ts
```diff
@@ -152,7 +152,16 @@ export async function checkMessageStatus({
         const merkleAddress = coreAddresses[origin].merkleTreeHook;
         stubMerkleTreeConfig(relayer, origin, hookAddress, merkleAddress);
       }
-      await relayer.relayAll(dispatchedReceipt, relayable);
+      // relayAll returns a receipt per message: either from the relay tx or,
+      // if the message was already delivered, from getProcessedReceipt.
+      const results = await relayer.relayAll(dispatchedReceipt, relayable);
+      const delivered = Object.values(results).flat().length;
```
paulbalaji, line 158: Nit: the user only sees `"Failed to relay N/M messages"` here — the actual per-message error details are swallowed inside `relayAll` (logged at debug level). Consider logging or surfacing the underlying error messages so operators can act on them without re-running with `LOG_LEVEL=debug`. Not blocking.

paulbalaji (approved):
## Approve

Fixes a real regression introduced in [`98ac676a28`](https://github.com/hyperlane-xyz/hyperlane-monorepo/commit/98ac676a28) (2026-03-06, #8259) where `status --relay` constructs `HyperlaneCore` before lazily adding destination signers. `HyperlaneApp` snapshots contract connections at construction time, so `deliver()` was hitting a provider-backed mailbox — hence `"sending a transaction requires a signer"`, not a missing-key error.

The fix is correct and minimal: `.connect()` returns a new ethers Contract instance (no mutation of the cached contract map), `getSigner()` throws loudly if no signer is set, and destination chain is derived via validated domain lookup — no cross-chain signer confusion risk.

The CLI change closing the silent-failure gap (`relayAll` swallowing errors + unconditional exit 0) is equally important. `process.exit(1)` on failure is consistent with the rest of the CLI.

Two non-blocking notes inline.

typescript/sdk/src/core/HyperlaneCore.ts
tkporter: Good catch — fixed in a413515ba

typescript/cli/src/status/message.ts
tkporter: The per-message errors are already logged by the relayer at error level (`this.logger.error` in `relayAll`), so they're visible in the output. The `logRed` here is just a summary. Leaving as-is.

typescript/sdk/src/core/HyperlaneCore.ts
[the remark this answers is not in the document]
paulbalaji: devin is right, it is missing a changeset

.changeset/fix-status-relay-signer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@hyperlane-xyz/sdk': patch
+'@hyperlane-xyz/cli': patch
+---
+
+Fixed `deliver()` and `sendMessage()` in HyperlaneCore to connect the mailbox with the current signer at call time, preventing "sending a transaction requires a signer" errors when signers are added after construction. The `status --relay` command now exits non-zero when relay fails.
```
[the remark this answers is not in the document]
tkporter: whatever

PR #8594, opened 2026-04-14
fix(sdk): bigint scale helpers and WarpCore collateral check for mixed-decimal routes

Xaroz:
## Summary
- Extracted the SDK-level scale/decimals fixes from #8563 (rebalancer PR) into a standalone PR
- Added bigint scale conversion helpers (`messageAmountFromLocal`, `localAmountFromMessage`, `minLocalAmountForMessage`, `alignLocalAmountToMessage`) to `decimals.ts`
- Updated `TokenConfigSchema` to accept `{numerator, denominator}` scale format alongside plain numbers
- Rewrote `WarpCore.isDestinationCollateralSufficient` to compare in message-space, fixing unit mismatch and precision loss bugs when origin/dest have different decimals+scales
- Added legacy decimal-conversion fallback for mixed-decimal routes where both scales are `undefined`

## Bugs fixed
These are SDK-level fixes for the bugs documented in the scale/decimals audit:

| Bug | Description | Fix |
|-----|-------------|-----|
| 1 | `convertToScaledAmount` can't handle `{num,den}` objects | New helpers accept all `ScaleInput` variants via `normalizeScale` |
| 2 | Precision loss via `Math.floor` in float arithmetic | Pure bigint math throughout, no intermediate floats |
| 3 | Unit mismatch in `isDestinationCollateralSufficient` | Both sides now compared in message-space |
| 4 | `!==` reference equality on scale objects | Legacy path only triggers when both scales are `undefined`; message-space path handles all formats |

## Note
The warp-ui display bug (`fromWei` using origin decimals instead of dest decimals) is in the warp-ui-template repo and will be addressed separately.

## Related
- Extracted from #8563 (rebalancer decimal normalization PR by @Mo-Hussain)
- Review comments from that PR (coderabbitai, paulbalaji, claude) are already incorporated: `&&` for fallback condition, `ceilDiv` assert, `alignLocalAmountToMessage` negative rejection, test covering message-space path

## Test plan
- [x] All new scale conversion helpers have unit tests (floor rounding, ceil rounding, alignment, negative rejection)
- [x] `WarpCore.test.ts` covers: plain number scales, rational `{num,den}` scales, legacy fallback for missing scale, one-sided rational scale with implicit identity
- [x] SDK builds cleanly (`pnpm -C typescript/sdk build`)
- [x] All 43 relevant tests pass

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8594" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/sdk/src/token/ITokenMetadata.ts
+++ b/typescript/sdk/src/token/ITokenMetadata.ts
@@ -50,9 +50,9 @@ export const TokenConfigSchema = z.object({
     .string()
     .optional()
     .describe('The CoinGecko id of the token, used for price lookups'),
-  scale: TokenMetadataSchema.shape.scale
-    .optional()
-    .describe('The scaling factor of the token'),
+  scale: TokenMetadataSchema.shape.scale.describe(
+    'The scaling factor of the token',
+  ),
   warpRouteId: z
     .string()
     .min(1)
--- a/typescript/sdk/src/token/ITokenMetadata.ts
+++ b/typescript/sdk/src/token/ITokenMetadata.ts
@@ -11,6 +11,7 @@ import {
   TokenConnectionConfigSchema,
 } from './TokenConnection.js';
 import { TokenStandard } from './TokenStandard.js';
+import { TokenMetadataSchema } from './types.js';
 
 export const TokenConfigSchema = z.object({
   chainName: ZChainName.describe(
@@ -49,7 +50,9 @@ export const TokenConfigSchema = z.object({
     .string()
     .optional()
     .describe('The CoinGecko id of the token, used for price lookups'),
-  scale: ZUint.lt(256).optional().describe('The scaling factor of the token'),
+  scale: TokenMetadataSchema.shape.scale
+    .optional()
+    .describe('The scaling factor of the token'),
   warpRouteId: z
     .string()
     .min(1)
--- a/typescript/sdk/src/utils/decimals.test.ts
+++ b/typescript/sdk/src/utils/decimals.test.ts
@@ -144,7 +144,7 @@ describe('scale conversion helpers', () => {
   it('rejects negative message amounts for ceil local conversion', () => {
     expect(() =>
       minLocalAmountForMessage(-1n, { numerator: 1n, denominator: 3n }),
-    ).to.throw('Numerator must be non-negative');
+    ).to.throw('Message amount must be non-negative');
   });
 
   it('aligns local amounts to exact message progress without leaking local dust', () => {
--- a/typescript/sdk/src/utils/decimals.test.ts
+++ b/typescript/sdk/src/utils/decimals.test.ts
@@ -176,6 +176,27 @@ describe('scale conversion helpers', () => {
       alignLocalAmountToMessage(-1n, { numerator: 1n, denominator: 3n }),
     ).to.throw('Local amount must be non-negative');
   });
+
+  it('returns identity for undefined scale', () => {
+    expect(messageAmountFromLocal(42n, undefined)).to.equal(42n);
+    expect(localAmountFromMessage(42n, undefined)).to.equal(42n);
+  });
+
+  it('handles zero amounts', () => {
+    expect(
+      messageAmountFromLocal(0n, { numerator: 1n, denominator: 3n }),
+    ).to.equal(0n);
+    expect(
+      localAmountFromMessage(0n, { numerator: 3n, denominator: 2n }),
+    ).to.equal(0n);
+  });
+
+  it('round-trips exact-divisible amounts', () => {
+    const scale = { numerator: 3n, denominator: 2n };
+    const local = 6n;
+    const message = messageAmountFromLocal(local, scale);
+    expect(localAmountFromMessage(message, scale)).to.equal(local);
+  });
 });
 
 describe(verifyScale.name, () => {
--- a/typescript/sdk/src/utils/decimals.test.ts
+++ b/typescript/sdk/src/utils/decimals.test.ts
@@ -7,7 +7,15 @@ import {
   WarpRouteDeployConfigMailboxRequired,
 } from '../token/types.js';
 
-import { normalizeScale, scalesEqual, verifyScale } from './decimals.js';
+import {
+  alignLocalAmountToMessage,
+  localAmountFromMessage,
+  messageAmountFromLocal,
+  minLocalAmountForMessage,
+  normalizeScale,
+  scalesEqual,
+  verifyScale,
+} from './decimals.js';
 
 describe(normalizeScale.name, () => {
   it('should normalize undefined to DEFAULT_SCALE', () => {
@@ -102,6 +110,71 @@ describe(scalesEqual.name, () => {
     const slightlyDifferent = { numerator: 10n ** 18n + 1n, denominator: 1n };
     expect(scalesEqual(large, large)).to.be.true;
     expect(scalesEqual(large, slightlyDifferent)).to.be.false;
+  });
+});
+
+describe('scale conversion helpers', () => {
+  it('converts local amount to message amount using floor rounding', () => {
+    expect(
+      messageAmountFromLocal(5n, { numerator: 1n, denominator: 3n }),
+    ).to.equal(1n);
+    expect(
+      messageAmountFromLocal(2n, { numerator: 3n, denominator: 2n }),
+    ).to.equal(3n);
+  });
+
+  it('converts message amount to local amount using inbound floor rounding', () => {
+    expect(
+      localAmountFromMessage(7n, { numerator: 3n, denominator: 2n }),
+    ).to.equal(4n);
+    expect(
+      localAmountFromMessage(1n, { numerator: 1n, denominator: 3n }),
+    ).to.equal(3n);
+  });
+
+  it('computes the minimum local amount needed to reach a message amount', () => {
+    expect(
+      minLocalAmountForMessage(7n, { numerator: 3n, denominator: 2n }),
+    ).to.equal(5n);
+    expect(
+      minLocalAmountForMessage(2n, { numerator: 1n, denominator: 3n }),
+    ).to.equal(6n);
+  });
+
+  it('rejects negative message amounts for ceil local conversion', () => {
+    expect(() =>
+      minLocalAmountForMessage(-1n, { numerator: 1n, denominator: 3n }),
+    ).to.throw('Numerator must be non-negative');
+  });
+
+  it('aligns local amounts to exact message progress without leaking local dust', () => {
+    expect(
+      alignLocalAmountToMessage(5n, { numerator: 1n, denominator: 3n }),
+    ).to.deep.equal({
+      localAmount: 3n,
+      messageAmount: 1n,
+    });
+    expect(
+      alignLocalAmountToMessage(5n, { numerator: 3n, denominator: 2n }),
+    ).to.deep.equal({
+      localAmount: 5n,
+      messageAmount: 7n,
+    });
+    expect(
+      alignLocalAmountToMessage(999_999_999_999n, {
+        numerator: 1n,
+        denominator: 1_000_000_000_000n,
+      }),
+    ).to.deep.equal({
+      localAmount: 0n,
+      messageAmount: 0n,
+    });
+  });
+
+  it('rejects negative local amounts for alignment', () => {
+    expect(() =>
+      alignLocalAmountToMessage(-1n, { numerator: 1n, denominator: 3n }),
+    ).to.throw('Local amount must be non-negative');
   });
 });
 
--- a/typescript/sdk/src/utils/decimals.ts
+++ b/typescript/sdk/src/utils/decimals.ts
@@ -25,6 +25,11 @@ export const DEFAULT_SCALE: NormalizedScale = {
   denominator: 1n,
 };
 
+export type ScaleAlignment = {
+  localAmount: bigint;
+  messageAmount: bigint;
+};
+
 /**
  * Converts any accepted scale variant to NormalizedScale (bigint).
  */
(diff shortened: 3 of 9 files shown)


xeno097 (commented):
## Review Summary

The core math is provably correct and the refactor is a strict improvement over the old `convertToScaledAmount` approach. The old code had a real bug: it silently ignored scale when only one side had it defined (due to the `&&` guard), falling back to pure decimal conversion. The new message-space projection fixes this.

Key properties verified:
- All rounding directions correct (floor outbound, floor inbound, ceil for minimum)
- `alignLocalAmountToMessage` provably never returns `localAmount > input`
- Inverse consistency holds: `localAmountFromMessage(messageAmountFromLocal(x, s), s) <= x`
- No overflow risk (BigInt arbitrary precision)
- The old `precisionFactor` workaround is no longer needed — pure bigint multiply-then-divide is both simpler and more precise

## Test Coverage Gaps

The existing test values are all mathematically correct (manually verified). However, several branch combinations in `isDestinationCollateralSufficient` lack dedicated tests:

- **Destination has scale, origin doesn't** — the reverse of the existing asymmetric test. The conditional at L1246-1248 evaluates differently (`originToken.scale === undefined` is true, `resolvedDestinationToken.scale === undefined` is false → fallback skipped, message-space path used). This is correct behavior but untested.
- **Both tokens with non-trivial scale AND different decimals** — production-relevant scenario (e.g., origin 18-dec with scale `{1, 1e12}`, dest 6-dec with scale `{2, 1}`).
- **Boundary test for the message-space path** — the exact-boundary tests (1e18 / 1e18+1) only exercise the fallback path. No equivalent boundary test exists for the message-space path where floor division could matter.

For the scale conversion helpers:
- No identity scale test (`messageAmountFromLocal(x, undefined) === x`)
- No zero-amount tests (`messageAmountFromLocal(0n, ...)`, `localAmountFromMessage(0n, ...)`)
- No round-trip inverse test (`localAmountFromMessage(messageAmountFromLocal(x, s), s) == x` for exact-divisible inputs)

typescript/sdk/src/token/ITokenMetadata.ts
```diff
@@ -49,7 +50,9 @@ export const TokenConfigSchema = z.object({
     .string()
     .optional()
     .describe('The CoinGecko id of the token, used for price lookups'),
-  scale: ZUint.lt(256).optional().describe('The scaling factor of the token'),
+  scale: TokenMetadataSchema.shape.scale
+    .optional()
+    .describe('The scaling factor of the token'),
```
xeno097: **Nit: redundant `.optional()`**

`TokenMetadataSchema.shape.scale` is already `z.union([...]).optional()` (see `types.ts:51-52`). Chaining another `.optional()` here creates a double-wrapped `ZodOptional<ZodOptional<...>>`. It works at runtime and the inferred TS type is identical, but it's unnecessary noise.

```suggestion
  scale: TokenMetadataSchema.shape.scale.describe(
    'The scaling factor of the token',
  ),
```

typescript/sdk/src/utils/decimals.ts
```diff
@@ -39,6 +44,66 @@ export function normalizeScale(scale: ScaleInput | undefined): NormalizedScale {
...
+  return (numerator + denominator - 1n) / denominator;
+}
+
+export function messageAmountFromLocal(
+  localAmount: bigint,
+  scale: ScaleInput | undefined,
+): bigint {
+  const normalized = normalizeScale(scale);
+  assertValidScale(normalized);
+  return (localAmount * normalized.numerator) / normalized.denominator;
```
xeno097, line 68: +1 to CodeRabbit's comment — `messageAmountFromLocal` and `localAmountFromMessage` (below) are the only two public conversion functions that don't guard against negative inputs. `alignLocalAmountToMessage` asserts `>= 0n`, and `ceilDiv` asserts `>= 0n`, but these two silently truncate toward zero for negative bigints (e.g., `(-5n * 1n) / 3n === -1n`, not `-2n`). Since these are now public exports, adding `assert(localAmount >= 0n)` / `assert(messageAmount >= 0n)` would make the API consistent.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1242,34 +1242,36 @@ export class WarpCore {
...
-    );
-
-    // check for scaling factor
     if (
-      originToken.scale &&
-      resolvedDestinationToken.scale &&
-      originToken.scale !== resolvedDestinationToken.scale
+      originToken.decimals !== resolvedDestinationToken.decimals &&
+      originToken.scale === undefined &&
+      resolvedDestinationToken.scale === undefined
```
xeno097, line 1255: **The fallback is necessary but deserves a comment explaining *why*.**

Without scale metadata, `messageAmountFromLocal(amount, undefined)` returns `amount` unchanged (identity scale). For mixed-decimal routes this would compare raw local units across different decimal spaces (e.g., `1_000_000` dest units vs `1_000_000_000_000_000_000` origin units), giving wrong results.

The fallback correctly handles legacy/misconfigured routes that lack scale but have different decimals. A brief comment here would save future readers from having to re-derive why the message-space path doesn't cover this case.

Also worth noting: this means routes with different decimals and no scale are implicitly treated as if the on-chain contract does decimal conversion — which may not match actual on-chain behavior. If the contract doesn't do decimal conversion either, both paths would be wrong. Might be worth a `this.logger.warn()` here for routes that hit this fallback.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1242,34 +1242,36 @@ export class WarpCore {
...
-    const isSufficient = BigInt(destinationBalanceInOriginDecimals) >= amount;
+    const requiredMessageAmount = messageAmountFromLocal(
+      amount,
+      originToken.scale,
+    );
+    const availableMessageAmount = messageAmountFromLocal(
+      destinationBalance,
+      resolvedDestinationToken.scale,
+    );
+    const isSufficient = availableMessageAmount >= requiredMessageAmount;
```
xeno097, line 1281: **The message-space comparison is correct.** Verified mathematically:

- Both sides project to a common "message unit" via their respective scales
- `verifyScale` guarantees that for well-configured routes: `scale_A.num * 10^dec_A / scale_A.den == scale_B.num * 10^dec_B / scale_B.den`
- Floor division on both sides is conservative in the right direction (never overstates available collateral)

One subtle point: when only one side has scale and the other is `undefined` (identity), this works correctly *if* the identity side's local units ARE message units (e.g., a 6-decimal token where 1 local unit = 1 message unit). This is the standard configuration for the "simple" side of an asymmetric route.

typescript/sdk/src/utils/decimals.test.ts
```diff
@@ -105,6 +113,71 @@ describe(scalesEqual.name, () => {
...
+    });
+  });
+
+  it('rejects negative local amounts for alignment', () => {
+    expect(() =>
+      alignLocalAmountToMessage(-1n, { numerator: 1n, denominator: 3n }),
+    ).to.throw('Local amount must be non-negative');
+  });
+});
+
```
xeno097, line 201: **Missing test cases worth adding:**

1. **Identity scale** — `messageAmountFromLocal(42n, undefined)` should equal `42n`. This is the most common production path (tokens with no scale) and isn't explicitly tested.
2. **Zero amounts** — `messageAmountFromLocal(0n, {1n, 3n})` should be `0n`. The `alignLocalAmountToMessage` early return for `messageAmount === 0n` is only tested indirectly via sub-dust-threshold values.
3. **Round-trip inverse** — `localAmountFromMessage(messageAmountFromLocal(6n, {3n, 2n}), {3n, 2n})` should recover `6n` (exact-divisible input). This validates both directions work together.

typescript/sdk/src/warp/WarpCore.test.ts
```diff
@@ -267,13 +267,138 @@ describe('WarpCore', () => {
...
+      expect(
+        await scaledWarpCore.isDestinationCollateralSufficient({
+          originTokenAmount: originToken.amount(750_000_000_000_000_000n),
+          destination: test2.name,
+        }),
+      ).to.be.false;
+    } finally {
+      originStub.restore();
+      destinationStub.restore();
+    }
```
xeno097, line 401: **Missing the reverse asymmetric case: destination has scale, origin doesn't.**

This test covers origin-has-scale / destination-doesn't. The reverse (origin 6-dec no scale, destination 18-dec with scale `{2, 1_000_000_000_000}`) exercises a different evaluation of the conditional at `WarpCore.ts:1246-1248` and is worth adding.

Also missing: **both tokens with non-trivial scale AND different decimals** (e.g., origin 18-dec scale `{1, 1e12}`, dest 6-dec scale `{2, 1}`). This is the production-realistic scenario for routes where both sides have been explicitly configured.

xeno097 (approved):
minor comments by claude but basically approved

typescript/sdk/src/utils/decimals.ts
```diff
@@ -39,6 +44,66 @@ export function normalizeScale(scale: ScaleInput | undefined): NormalizedScale {
...
+}
+
+export function localAmountFromMessage(
+  messageAmount: bigint,
+  scale: ScaleInput | undefined,
+): bigint {
+  const normalized = normalizeScale(scale);
+  assertValidScale(normalized);
+  return (messageAmount * normalized.denominator) / normalized.numerator;
+}
```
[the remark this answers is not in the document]
Xaroz: Handled in `6977a9d`. Added `assert(localAmount >= 0n)` and `assert(messageAmount >= 0n)` to both functions.

typescript/sdk/src/warp/WarpCore.ts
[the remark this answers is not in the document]
Xaroz: Acknowledged. This is the intended behavior — `scale: 1` on both sides with different decimals is a misconfiguration that `verifyScale()` catches at deploy time. The old code silently papered over it; the new code safely returns `false` (insufficient), which is the correct failure direction.

typescript/sdk/src/token/ITokenMetadata.ts
Xaroz: Good catch. Handled in `6977a9d` — removed the redundant `.optional()`.

typescript/sdk/src/utils/decimals.ts
Xaroz: Handled in `6977a9d` along with the CodeRabbit comment. Both functions now assert `>= 0n` consistently with the rest of the API.

typescript/sdk/src/warp/WarpCore.ts
Xaroz: Handled in `6977a9d`. Added a comment explaining why the fallback exists and that misconfigured routes safely return `false` via the message-space path. Skipped the `logger.warn` — it would fire on every collateral check for legacy routes, and `verifyScale()` already catches this at deploy time.

typescript/sdk/src/warp/WarpCore.ts
Xaroz: Thanks for the verification 👍

typescript/sdk/src/utils/decimals.test.ts
Xaroz: Handled in `6977a9d`. Added all three: identity scale (`undefined`), zero amounts, and round-trip inverse.

typescript/sdk/src/warp/WarpCore.test.ts
Xaroz: Handled in `6977a9d`. Added both: reverse asymmetric (origin 6-dec no scale, dest 18-dec with rational scale) and both-scaled mixed-decimal (origin 18-dec `{1, 1e12}`, dest 6-dec `{2, 1}`).

typescript/sdk/src/utils/decimals.ts
```diff
@@ -39,6 +44,68 @@ export function normalizeScale(scale: ScaleInput | undefined): NormalizedScale {
...
+  messageAmount: bigint,
+  scale: ScaleInput | undefined,
+): bigint {
+  assert(messageAmount >= 0n, 'Message amount must be non-negative');
+  const normalized = normalizeScale(scale);
+  assertValidScale(normalized);
+  return (messageAmount * normalized.denominator) / normalized.numerator;
+}
+
+export function minLocalAmountForMessage(
```
paulbalaji, line 81: **Nit:** Unlike `messageAmountFromLocal` (line 65) and `localAmountFromMessage` (line 75), this function does not directly assert `messageAmount >= 0n`. It relies on `ceilDiv` catching it with `"Numerator must be non-negative"`, which is a confusing error message for callers passing a negative `messageAmount`. Consider adding `assert(messageAmount >= 0n, 'Message amount must be non-negative')` for consistency with the other helpers.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1242,34 +1242,43 @@ export class WarpCore {
...
-    );
-
-    // check for scaling factor
+    // Legacy fallback: when both scales are undefined but decimals differ,
+    // we can't use message-space comparison because messageAmountFromLocal
+    // with identity scale would compare raw local units across different
+    // decimal spaces. Fall back to decimal conversion instead.
+    // Well-configured routes should have scale set — verifyScale() catches
+    // this at deploy time. Misconfigured routes that reach the message-space
+    // path below will safely return false (insufficient).
```
paulbalaji: **Nit:** "Misconfigured routes that reach the message-space path below will safely return false (insufficient)" — this isn't universally true. A route with _different_ decimals where one side has an explicit identity scale (`{1,1}`) and the other is `undefined` would skip the fallback and compare raw local units via message-space, which could return either true or false depending on values. Consider softening to "may produce incorrect results" rather than "safely return false."

typescript/sdk/src/token/ITokenMetadata.ts
```diff
@@ -49,7 +50,9 @@ export const TokenConfigSchema = z.object({
     .string()
     .optional()
     .describe('The CoinGecko id of the token, used for price lookups'),
-  scale: ZUint.lt(256).optional().describe('The scaling factor of the token'),
+  scale: TokenMetadataSchema.shape.scale.describe(
```
paulbalaji, line 53: **Info:** This widens the accepted schema from `ZUint.lt(256)` (non-negative integers < 256) to the `TokenMetadataSchema.shape.scale` union (positive integers, rational `{num,den}` objects). It also tightens by rejecting `scale: 0`. Both changes are improvements, but worth noting as a runtime compatibility change for any external consumers that were previously passing `scale: 0` or relying on the `< 256` cap.

paulbalaji (approved):
## Consolidated Review

Reviewed by four parallel analysis passes: math correctness, security, code quality, and test coverage. Also cross-referenced with Codex's independent assessment.

### Verdict: **Approve**

The core math is provably correct and the refactor is a strict improvement over the old `convertToScaledAmount` approach.

**Key properties verified:**
- All rounding directions correct (floor outbound, floor inbound, ceil for minimum)
- `alignLocalAmountToMessage` provably never returns `localAmount > input`
- Inverse consistency holds: `localAmountFromMessage(messageAmountFromLocal(x, s), s) <= x`
- No overflow risk (BigInt arbitrary precision), no division-by-zero paths
- Collateral check mirrors on-chain floor semantics — conservative in the right direction

**No blocking findings.** Two items worth release-note callouts:

| Item | Severity | Detail |
|------|----------|--------|
| Behavioral change for asymmetric scale configs | MEDIUM | Old code silently fell through to decimal comparison when only one side had scale defined. New code correctly applies message-space projection. Pre-existing deployed routes without `verifyScale()` enforcement could see different results. Intentional and correct, but worth noting for operators. |
| Schema tightening | MEDIUM | `TokenConfigSchema.scale` went from `ZUint.lt(256)` (accepted 0, capped at 255) to `TokenMetadataSchema.shape.scale` (rejects 0, no upper bound, accepts rational objects). Safer, but a runtime compatibility change for direct config consumers. |

**Test coverage** is solid — all gaps from xeno097's initial review are addressed in `6977a9d`. Minor remaining gaps (explicit `{1n,1n}` identity scale test, exact-division `ceilDiv` case) are non-blocking.

Inline comments below are nits/observations, not blockers.

typescript/sdk/src/utils/decimals.ts
Xaroz: Handled in `7d8fded`. Added `assert(messageAmount >= 0n, 'Message amount must be non-negative')` for consistency with the other helpers. Updated the test to expect the new error message.

typescript/sdk/src/warp/WarpCore.ts
Xaroz: Good catch. Handled in `7d8fded` — softened the comment to "may produce incorrect results" since the direction of failure isn't guaranteed when one side has explicit `{1,1}` and the other is `undefined` with different decimals.

typescript/sdk/src/token/ITokenMetadata.ts
Xaroz: Acknowledged. The widening (accepting rational `{num,den}` objects) and tightening (rejecting `scale: 0`) are both intentional. No existing production routes should be using `scale: 0` since that's meaningless.

typescript/sdk/src/warp/WarpCore.ts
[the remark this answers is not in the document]
Xaroz: Good find. This is a pre-existing bug — out of scope for this PR but noted as a TODO. The same `messageAmountFromLocal` pattern should be applied to `validateDestinationRateLimit` in a follow-up.

typescript/sdk/src/warp/WarpCore.ts
```diff
@@ -1242,34 +1242,43 @@ export class WarpCore {
...
+        ),
+      );
+      const isSufficient = destinationBalanceInOriginDecimals >= amount;
+      this.logger.debug(
+        `${originTokenAmount.token.symbol} to ${destination} has ${
+          isSufficient ? 'sufficient' : 'INSUFFICIENT'
+        } collateral`,
       );
+      return isSufficient;
     }
```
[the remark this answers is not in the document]
Xaroz: Acknowledged — this is the intended behavioral improvement. The old code silently ignored one-sided scale and fell through to decimal conversion, which was incorrect. The new tests explicitly verify both directions (origin-has-scale and dest-has-scale).

PR #8599, opened 2026-04-15
fix: solana routing fee

troykessler:
### Description

`warp deploy` for cross-protocol routes (EVM + SVM/Cosmos) with a `RoutingFee` token fee failed during router enrollment with:

```
✅ Warp contract deployments complete
Start enrolling cross chain routers
Failed to create enroll router transactions for chain ethereumlocal1: [
  {
    "code": "custom",
    "path": [
      "tokenFee",
      "feeContracts"
    ],
    "message": "RoutingFee must define at least one destination fee"
  }
]
Error: Failed to create router enrollment transactions for 1 chain(s): ethereumlocal1: ...
```

**Root cause**: The EVM deployer (`HypERC20Deployer`) processes only EVM chains in its batch. When `enrollRemoteRouters()` runs, `solanalocal` isn't in `deployedContractsMap` (it's deployed separately in the SVM pass) and isn't a `foreignDeployment`, so no enrollment happens. `TokenRouter.domains()` on the EVM chain is `[]` at this point.

`deployAndConfigureTokenFees()` then correctly deploys the `RoutingFee` and `LinearFee` sub-fee contracts on-chain.

Later, `enrollCrossChainRouters()` calls `evmWarpModule.read()`. The reader derives `RoutingFee.feeContracts` from `TokenRouter.domains()`, which is still `[]` — so it returns `feeContracts: {}`. This empty object is spread into `expectedConfig`, which then fails `RoutingFeeInputConfigSchema` validation (requires ≥1 destination fee) when `evmWarpModule.update()` runs.

**Fix**: In `enrollCrossChainRouters()`, override `tokenFee` in `expectedConfig` with the original value from the deploy config (non-empty `feeContracts`) instead of the one read from chain. `EvmTokenFeeModule.update()` reads the actual on-chain state via `routingDestinations` and confirms no change is needed.

**Reproduction** (deploy config):

```yaml
ethereumlocal1:
  decimals: 9
  mailbox: "0x610178dA211FEF7D417bC0e6FeD39F05609AD788"
  name: Solana
  owner: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
  symbol: SOL
  tokenFee:
    feeContracts:
      solanalocal:
        bps: 6
        owner: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
        type: LinearFee
    owner: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
    type: RoutingFee
  type: synthetic
solanalocal:
  decimals: 9
  gas: 300000
  hook: GwHaw8ewMyzZn9vvrZEnTEAAYpLdkGYs195XWcLDCN4U
  mailbox: 692KZJaoe2KRcD6uhCQDLLXnLNA5ZLnfvdqjE4aX9iu1
  owner: GhX4cjAaRAPRTb9fkgLtLq9cYTLncRBGN3pa4KSKUv4J
  type: native
```

### Drive-by changes

None

### Backward compatibility

Yes

### Testing

Manual — re-ran `pnpm hyperlane warp deploy` for `SOL/igra-solanatestnet` (EVM + Sealevel route with `RoutingFee`)

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8599" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/cli/src/tests/cross-chain/warp/warp-cc-evm-svm.e2e-test.ts
+++ b/typescript/cli/src/tests/cross-chain/warp/warp-cc-evm-svm.e2e-test.ts
@@ -21,6 +21,7 @@ import {
   runSolanaNode,
 } from '@hyperlane-xyz/sealevel-sdk/testing';
 import {
+  TokenFeeType,
   TokenType,
   type WarpCoreConfig,
   type WarpRouteDeployConfig,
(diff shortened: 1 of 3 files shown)


xeno097 (commented):
Can we also add an e2e test in the cross chain folder so we can ensure that future changes do not break this behaviour?

typescript/cli/src/tests/cross-chain/warp/warp-cc-evm-svm.e2e-test.ts
```diff
@@ -355,4 +356,75 @@ describe('hyperlane warp crossCollateral EVM+SVM e2e tests', function () {
...
+    });
+
+    const warpCorePath = getWarpCoreConfigPath(SYMBOL, [EVM_CHAIN, SVM_CHAIN]);
+    const deployedConfig = await warpCommands.readConfig(
+      EVM_CHAIN,
+      warpCorePath,
+    );
+    expect(deployedConfig[EVM_CHAIN].tokenFee?.type).to.equal(
+      TokenFeeType.RoutingFee,
+    );
```
xeno097, line 428: Should also check that the solana domain is set on the routing fee config?

xeno097 (approved):

typescript/sdk/src/deploy/warp.ts
```diff
@@ -645,6 +645,17 @@ export async function enrollCrossChainRouters(
...
+            // never enrolls non-EVM remote routers, so TokenRouter.domains()=[]
+            // at this point. The reader derives RoutingFee.feeContracts from
+            // enrolled domains, returning {} which fails
+            // RoutingFeeInputConfigSchema validation. Use the deploy config's
+            // tokenFee (non-empty feeContracts) so validation passes.
+            // EvmTokenFeeModule.update() reads actual on-chain state via
+            // routingDestinations and confirms no change is needed.
+            ...(resolvedConfigMap[currentChain].tokenFee && {
+              tokenFee: resolvedConfigMap[currentChain].tokenFee,
+            }),
```
paulbalaji, line 658: Fix looks correct. The `expectedConfig.tokenFee` override only affects the target config passed to `evmWarpModule.update()`. The actual on-chain state is re-read inside `update()` (line 170 of `EvmWarpModule.ts`), and `EvmTokenFeeModule.update()` uses the explicitly-passed `routingDestinations` to diff correctly.

Minor nit: the comment says the fee module "confirms no change is needed" -- more precisely it computes the correct diff (which happens to be empty during initial cross-protocol deploy, but wouldn't be in other contexts).

typescript/cli/src/tests/cross-chain/warp/warp-cc-evm-svm.e2e-test.ts
paulbalaji, line 428: Non-blocking, but +1 to xeno097's and CodeRabbit's earlier suggestion: asserting `tokenFee.type` alone wouldn't catch a future regression where deployment succeeds but `readConfig()` drops the SVM destination fee (the `feeContracts` mapping). Worth a follow-up:

```ts
expect(deployedConfig[EVM_CHAIN].tokenFee?.feeContracts?.[SVM_CHAIN]?.type)
  .to.equal(TokenFeeType.LinearFee);
```

The deploy-time throw is already covered by `deployRaw()` succeeding, so this is hardening, not blocking.

paulbalaji (approved):
## Review Summary

The core fix is correct. `enrollCrossChainRouters()` now seeds `expectedConfig.tokenFee` from deploy config before calling `evmWarpModule.update()`, which matches the downstream behavior: `EvmTokenFeeModule.update()` re-reads actual on-chain state with explicit `routingDestinations` and diffs against that. No path where this override masks a real deploy-time misconfiguration for the scenario this PR fixes.

For EVM-only routes with RoutingFee, the override is idempotent (routers are already enrolled, so the reader-derived and deploy-config `tokenFee` agree). For non-RoutingFee types (LinearFee, etc.), the override is harmless since those don't derive config from enrolled domains. When `tokenFee` is absent, the conditional spread correctly no-ops.

The e2e test directly covers the regression scenario. `deployRaw()` would throw if the original bug recurred. One open item from prior review discussion (xeno097 + CodeRabbit) noted inline.

`patch` semver bump is correct.

PR #8607, opened 2026-04-15
fix(svm-sdk,cli): fix Solana-origin warp send with legacy tx conversion

xeno097:
### Description

  Fixed Solana-origin `warp send` which was broken because SDK adapters (`SealevelHypCollateralAdapter`, `SealevelHypCrossCollateralAdapter`) return legacy `@solana/web3.js` `Transaction` objects, but the CLI's `SvmSigner` expects
  `@solana/kit` `SvmTransaction` format. The mismatch caused a `SolanaError` when kit tried to read `programAddress` from legacy instructions that use `programId`.

  Based on the fix in https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8567 while trying to make the minimum amount of changes and not install `@solana/web3.js` in the svm-sdk package. Instead, structural (duck-typed) interfaces
  are used to detect and convert legacy transactions without adding the dependency.

  **Changes:**

  - **`svm-sdk/src/legacy-compat.ts`** (new): Conversion layer that transforms legacy `@solana/web3.js` transactions to `@solana/kit` format. Handles instruction format translation (`programId`/`keys` → `programAddress`/`accounts`),
  extracts `SetComputeUnitLimit` into `computeUnits` field while preserving `SetComputeUnitPrice` as-is, and converts legacy `Keypair` extra signers to kit `TransactionSigner` via `createKeyPairSignerFromBytes`.
  - **`svm-sdk/src/clients/signer.ts`**: Added `fetchTransactionMeta` helper that fetches full transaction via `getTransaction` after confirmation to populate `meta.logMessages` on the receipt (needed by
  `SealevelCoreAdapter.extractMessageIds`).
  - **`svm-sdk/src/types.ts`**: Extended `SvmReceipt` with `meta?: { logMessages?: readonly string[] }`.
  - **`cli/src/send/transfer.ts`**: Detects legacy Solana transactions in the submission loop and converts them before passing to `SvmSigner`, forwarding `tx.extraSigners` for additional signer conversion.

  ### Drive-by changes

  - Registered `SealevelHypCrossCollateral` token standard in `createSealevelHypAdapter` factory (`sdk/src/token/adapters/sealevelHyp.ts`).
  - Added stale chain directory cleanup in the cross-chain e2e test global setup to prevent `unknown1`-style leftover directories from failing tests.
  - Added `@solana/spl-token` and `@solana/web3.js` as dev dependencies to the CLI (needed by the e2e test for SPL mint setup).

  ### Related issues

  - Fixes the same issue addressed in #8567

  ### Backward compatibility

  Yes. The conversion only activates when `isLegacySolanaTransaction` detects legacy format (checks for `programId` without `programAddress`). Native kit-format `SvmTransaction` objects pass through unchanged. The `SvmReceipt.meta` field
  is optional. `fetchTransactionMeta` is best-effort (catches errors, returns receipt unchanged). No existing public APIs were modified.

  ### Testing

  - E2E: Added `warp-send-evm-svm` cross-chain e2e test (spins up local EVM + Solana nodes, deploys core + warp route, sends SVM→EVM, asserts Message ID in output). Added to CI matrix.
  - Manual: Verified with mainnet cross-collateral token router (`solanamainnet` → `arbitrum`).

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8607" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


xeno097:
@claude brutally honest review pls

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,41 @@ export class SvmSigner
...
+          commitment: RPC_COMMITMENT_LEVEL,
+          maxSupportedTransactionVersion: 0,
+          encoding: 'jsonParsed',
+        })
+        .send();
+
+      return {
+        ...receipt,
+        meta: { logMessages: fullTx?.meta?.logMessages ?? [] },
+      };
```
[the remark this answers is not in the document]
xeno097: fixed

typescript/svm-sdk/src/legacy-compat.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/** Structural interface matching @solana/web3.js Keypair. */
+export interface LegacyKeypair {
+  secretKey: Uint8Array;
+}
+
+const COMPUTE_BUDGET_PROGRAM =
+  'ComputeBudget111111111111111111111111111111' as Address;
+const SET_COMPUTE_UNIT_LIMIT_DISCRIMINATOR = 2;
```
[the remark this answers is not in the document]
xeno097: should be fixed

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,48 @@ export class SvmSigner
...
+      assertIsSignature(receipt.signature);
+      const fullTx = await this.rpc
+        .getTransaction(receipt.signature, {
+          commitment: RPC_COMMITMENT_LEVEL,
+          maxSupportedTransactionVersion: 0,
+          encoding: 'jsonParsed',
+        })
+        .send();
+
+      if (!fullTx?.meta?.logMessages) {
```
nambrot-agent: This is still racy with `SealevelCoreAdapter.extractMessageIds()`, which hard-requires `receipt.meta.logMessages`. If `getTransaction()` returns `null` or lags indexing right after confirmation, we return a successful receipt without logs and the CLI can false-fail with `Transaction logs required to check message delivery` even though the tx landed on chain. I think this either needs a bounded retry/poll here for meta availability, or the downstream log consumer needs to tolerate delayed log hydration.

nambrot-agent (commented):
One blocking risk in the new receipt-meta path.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,48 @@ export class SvmSigner
...
+      }
+
+      return {
+        ...receipt,
+        meta: { logMessages: fullTx.meta.logMessages },
+      };
+    } catch (error) {
+      this.logger.warn('Failed to fetch transaction meta', { error });
+      return receipt;
+    }
```
paulbalaji, line 526: **Blocking: `fetchTransactionMeta` needs bounded retries — confirmed transactions can lag RPC indexing.**

`getTransaction()` commonly returns `null` right after confirmation on Solana RPCs (documented behavior). When that happens, this method returns the receipt without `meta.logMessages`. The downstream consumer `SealevelCoreAdapter.extractMessageIds` hard-throws:

```ts
const logs = sourceTx.receipt.meta?.logMessages;
if (!logs)
  throw new Error('Transaction logs required to check message delivery');
```

Result: a successful on-chain send surfaces as a CLI failure. The local-validator e2e test won't catch this because indexing lag is near-zero locally.

Suggested fix — bounded retry with backoff:
```ts
private async fetchTransactionMeta(receipt: SvmReceipt): Promise<SvmReceipt> {
  assertIsSignature(receipt.signature);
  const MAX_RETRIES = 5;
  const BACKOFF_MS = 1000;

  for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
    try {
      const fullTx = await this.rpc
        .getTransaction(receipt.signature, {
          commitment: RPC_COMMITMENT_LEVEL,
          maxSupportedTransactionVersion: 0,
          encoding: 'jsonParsed',
        })
        .send();

      if (fullTx?.meta?.logMessages) {
        return { ...receipt, meta: { logMessages: fullTx.meta.logMessages } };
      }
    } catch (error) {
      this.logger.debug(`Attempt ${attempt + 1} failed to fetch tx meta`, { error });
    }

    if (attempt < MAX_RETRIES - 1) await sleep(BACKOFF_MS * (attempt + 1));
  }

  this.logger.warn('Transaction meta unavailable after retries', {
    signature: receipt.signature,
  });
  return receipt;
}
```

This gives the RPC ~15s to index the transaction (1+2+3+4+5s backoff). If it still fails, the warn log makes the root cause discoverable. Agrees with nambrot-agent's open thread.

paulbalaji (commented):
## Consolidated Review

Reviewed the full diff, existing review threads (claude, coderabbit, devin, nambrot-agent), and the three fix-up commits. The conversion logic in `legacy-compat.ts` is well-designed — duck-typing to avoid the `@solana/web3.js` dep is pragmatic, instruction conversion handles all `AccountRole` variants correctly, and `SetComputeUnitLimit` extraction/recreation is sound. The e2e test is thorough and follows existing conventions. Prior review feedback has been addressed.

**One issue blocks approval** — see inline comment on `fetchTransactionMeta`. Everything else is clean or acceptable transitional tech debt with TODOs.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,48 @@ export class SvmSigner
...
+    const tx = isLegacySolanaTransaction(transaction)
+      ? await convertLegacySolanaTransaction(
+          transaction,
+          transaction.extraSigners,
+        )
+      : transaction;
+
+    const receipt = await this.send(tx);
+    return this.fetchTransactionMeta(receipt);
+  }
```
paulbalaji, line 488: **Minor: unconditional `fetchTransactionMeta` adds an RPC round-trip to every SVM transaction.**

This runs after deployments, ISM configs, hook setups — anything through `sendAndConfirmTransaction`. Only warp sends need `logMessages` for `extractMessageIds`. Not blocking, but worth a `// TODO: make meta-fetch opt-in once the legacy compat layer is removed` so it's tracked.

typescript/svm-sdk/src/types.ts
```diff
@@ -32,8 +34,20 @@ export interface SvmTransaction {
...
+  };
 }
 
+/**
+ * Mixin for legacy @solana/web3.js Keypair extra signers.
+ * Used at the signer boundary to accept extra signers from SDK adapters
+ * that still use the legacy Transaction format.
+ */
+export type WithExtraSigners<T> = T & {
+  extraSigners?: readonly LegacyKeypair[];
```
paulbalaji, line 51: **Nit: `WithExtraSigners` imports `LegacyKeypair` from `legacy-compat.ts`, creating a circular dep with core types.**

`types.ts` -> `legacy-compat.ts` (for `LegacyKeypair`) and `legacy-compat.ts` -> `types.ts` (for `SvmTransaction`). TypeScript resolves this since both are type-only, but when the legacy compat layer is removed, `types.ts` needs cleanup too. The existing TODO covers this implicitly — just flagging the coupling.

typescript/svm-sdk/src/clients/signer.ts
xeno097: should be addressed now

typescript/svm-sdk/src/types.ts
xeno097: fixed too

paulbalaji (commented):
## Consolidated Review

Reviewed the full diff, all prior review threads (claude[bot], coderabbitai, devin, nambrot-agent, paulbalaji), and all 7 commits on this branch.

**All prior blocking concerns resolved:**
- Retry blocker from @nambrot-agent / @paulbalaji → fixed in `3cd7f09` (5 retries, linear 1+2+3+4+5s backoff).
- Circular dep from @paulbalaji → fixed in `6d6cad2` (`LegacyKeypair` moved to `types.ts`).
- All claude[bot] / coderabbitai / devin nits (copy-paste assert, receipt mutation, bounds check, `as` casts, CAST comments, `readYamlOrJson` generic, missing-meta coercion) → addressed across `d48a005`, `0728e8fa`, `6971cb8`.

**No security issues.** Duck-typed detection is robust; `extraSigners` are ephemeral PDA-derivation keypairs (not user wallet keys); compute-budget extraction has proper bounds checking; retry loop is bounded.

**No build risk.** `assertIsSignature` is present in the currently-installed `@solana/kit` — no version bump required.

**Code quality is clean.** Casts have runtime validators or `// CAST:` comments; changeset is past-tense; e2e test is thorough and follows existing conventions.

One non-blocking follow-up inline — already acknowledged with a TODO and previously classified non-blocking. Ready to merge.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,61 @@ export class SvmSigner
...
-    return this.send(transaction);
+    const tx = isLegacySolanaTransaction(transaction)
+      ? await convertLegacySolanaTransaction(
+          transaction,
+          transaction.extraSigners,
+        )
+      : transaction;
+
+    const receipt = await this.send(tx);
+    return this.fetchTransactionMeta(receipt);
```
paulbalaji, line 487: Non-blocking follow-up: `fetchTransactionMeta` runs unconditionally on every `sendAndConfirmTransaction` call, so every SVM deploy / ISM config / hook setup now pays a `getTransaction` round-trip plus up to 15s of retry backoff when RPC indexing lags. Only the warp-send path actually needs `logMessages` for `extractMessageIds`. Generic submit paths (e.g. `deploy-sdk/AltVMJsonRpcSubmitter`) inherit this cost too. The existing TODO on line 493 covers this — flagging that the opt-in change is worth prioritising as the follow-up.

typescript/svm-sdk/src/clients/signer.ts
```diff
@@ -468,9 +474,61 @@ export class SvmSigner
...
+        if (fullTx?.meta?.logMessages) {
+          return {
+            ...receipt,
+            meta: { logMessages: fullTx.meta.logMessages },
+          };
+        }
+      } catch (error) {
+        this.logger.debug(`Attempt ${attempt + 1} failed to fetch tx meta`, {
+          error,
+        });
```
paulbalaji, line 520: Nit: errors inside the retry loop are logged at `debug`, but the final `logger.warn` on line 528 only says "Transaction meta unavailable" without the underlying cause. If all 5 retries fail due to a persistent issue (auth, malformed response), the root cause is effectively invisible in production logs. Consider surfacing the last error at `warn` level alongside the final "unavailable" message. Not blocking.

paulbalaji (approved):
Approving — blockers resolved, no security issues, code quality clean. See the consolidated review above for the non-blocking follow-up.

typescript/svm-sdk/src/clients/signer.ts
xeno097: will address in a follow up to also tackle the mandatory extra rpc request to get the tx receipt

PR #8615, opened 2026-04-16
fix(sdk): normalize scale in warp check for config vs on-chain comparison

Xaroz:
## Summary
Fixes false scale diff violations in `hyperlane warp check` where config values (plain numbers like `scale: 1000000000000`) were compared against on-chain values (objects like `{numerator: 1000000000000, denominator: 1}`) and reported as mismatches even though they're semantically equal.

## Root cause
The on-chain reader (`EvmWarpRouteReader.fetchScale`) always returns scale as `{numerator: bigint, denominator: bigint}` — even for legacy contracts with a single `scale()` function, it wraps it as `{numerator: scale, denominator: 1n}`. But deploy/config files use plain numbers. `diffObjMerge` does a structural comparison, so `1000000000000 !== {numerator: 1000000000000n, denominator: 1n}`.

## Fix
Normalize scale in `transformConfigToCheck` (which already normalizes other fields like `tokenFee` and `allowedRebalancers`) using `normalizeScale()` from the SDK. Both sides get converted to `{numerator: bigint, denominator: bigint}` before comparison.

Uses `!== null` (not `!= null`) so `undefined` values pass through to `normalizeScale` which returns identity `{1n, 1n}` — matching the on-chain reader which also returns `undefined` for identity scale.

## Related
- hyperlane-xyz/hyperlane-registry#1480 — adds missing scale fields to registry config files
- #8594 — SDK scale helpers
- #8602 — `ZBigNumberish.refine()` + `validateZodResult` fix

## Test

Tested locally running warp check against the modified routes in the registry 

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8615" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/sdk/src/token/configUtils.test.ts
+++ b/typescript/sdk/src/token/configUtils.test.ts
@@ -72,6 +72,7 @@ describe('configUtils', () => {
               },
             ],
           },
+          scale: { numerator: 1n, denominator: 1n },
         },
       },
       {
@@ -96,6 +97,7 @@ describe('configUtils', () => {
               address: ADDRESS,
             },
           },
+          scale: { numerator: 1n, denominator: 1n },
         },
       },
       {
@@ -118,6 +120,7 @@ describe('configUtils', () => {
             address: ADDRESS,
             owner: ADDRESS,
           },
+          scale: { numerator: 1n, denominator: 1n },
         },
       },
       {
@@ -169,6 +172,7 @@ describe('configUtils', () => {
             owner: 'milk169dcaz397j75tjfpl6ykm23dfrv39dqd58lsag',
             type: 'native',
           },
+          scale: { numerator: 1n, denominator: 1n },
         },
         input: {
           bsc: {
@@ -246,6 +250,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.LinearFee,
           owner: ADDRESS,
@@ -273,6 +278,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.OffchainQuotedLinearFee,
           owner: ADDRESS,
@@ -309,6 +315,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.RoutingFee,
           owner: ADDRESS,
@@ -360,6 +367,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.CrossCollateralRoutingFee,
           owner: ADDRESS,
@@ -406,6 +414,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.CrossCollateralRoutingFee,
           owner: ADDRESS,
@@ -445,6 +454,7 @@ describe('configUtils', () => {
       expect(transformedObj).to.eql({
         type: TokenType.collateral,
         token: ADDRESS,
+        scale: { numerator: 1n, denominator: 1n },
         tokenFee: {
           type: TokenFeeType.RoutingFee,
           owner: ADDRESS,
--- a/typescript/sdk/src/token/configUtils.test.ts
+++ b/typescript/sdk/src/token/configUtils.test.ts
@@ -233,6 +233,44 @@ describe('configUtils', () => {
       });
     }
 
+    it('normalizes plain number scale to {numerator, denominator} bigint', () => {
+      const transformedObj = transformConfigToCheck({
+        type: TokenType.collateral,
+        token: ADDRESS,
+        scale: 1000000000000,
+      } as any);
+
+      expect(transformedObj.scale).to.eql({
+        numerator: 1000000000000n,
+        denominator: 1n,
+      });
+    });
+
+    it('normalizes {number, number} scale to {bigint, bigint}', () => {
+      const transformedObj = transformConfigToCheck({
+        type: TokenType.collateral,
+        token: ADDRESS,
+        scale: { numerator: 1, denominator: 1000000000000 },
+      } as any);
+
+      expect(transformedObj.scale).to.eql({
+        numerator: 1n,
+        denominator: 1000000000000n,
+      });
+    });
+
+    it('normalizes undefined scale to identity {1n, 1n}', () => {
+      const transformedObj = transformConfigToCheck({
+        type: TokenType.collateral,
+        token: ADDRESS,
+      } as any);
+
+      expect(transformedObj.scale).to.eql({
+        numerator: 1n,
+        denominator: 1n,
+      });
+    });
+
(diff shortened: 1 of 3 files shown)


xeno097 (approved):

Xaroz:
fixing tests

typescript/sdk/src/token/configUtils.ts
```diff
@@ -687,6 +688,16 @@ export function transformConfigToCheck(
...
 
+  // Normalize scale so plain numbers and {numerator, denominator} objects compare equal.
+  // Uses !== null so undefined passes through to normalizeScale as identity {1n,1n}.
+  if (clonedTokenConfig.scale !== null) {
+    const normalized = normalizeScale(clonedTokenConfig.scale);
+    clonedTokenConfig.scale = {
+      numerator: normalized.numerator,
+      denominator: normalized.denominator,
+    };
+  }
```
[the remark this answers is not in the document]
Xaroz: correct, the issue with this is that the evm warp route token reader will return undefined when the numerator and denominator are 1, hence normalizeScale returns the opposite when its undefined
```
 if (result.numerator === 1n && result.denominator === 1n) {
      return undefined;
 }
```

paulbalaji: Three issues with this block:

**1. `!== null` guard is effectively always-true.** `scale` is `.optional()` in `TokenMetadataSchema`, so its type is `... | undefined`, not `null`. `undefined !== null` is always true -- the block runs on every config. (Same as Devin's finding. The reply about `normalizeScale` returning `undefined` for `{1n,1n}` is a conflation -- that identity-collapse is in `EvmWarpRouteReader.fetchScale:1452-1454`. `normalizeScale` itself returns `{1n,1n}` for `undefined` input; see `decimals.ts:36-45`.)

Production `warpCheck` still works only because `transformConfigToCheck` is applied to both sides in `warpCheck.ts:234-237`, so both sides get `{1n,1n}` injected. But this is fragile -- relies on `fetchScale`'s identity-collapse staying in place -- and it breaks unit tests in `configUtils.test.ts` (lines 224-229 and the `tokenFee` cases at 232/258/286/328/386/426) which assert against expected outputs with no `scale` key.

**2. Object rebuild is redundant.** `normalizeScale` already returns exactly `{ numerator: bigint; denominator: bigint }`.

**3. Comment describes *what*, not *why*,** and the described behavior ("undefined passes through to normalizeScale as identity {1n,1n}") is exactly the bug.

Suggested replacement (matches the sibling `tokenFee` pattern directly above):

```suggestion
  if (clonedTokenConfig.scale != null) {
    clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale);
  }
```

paulbalaji (commented):
## Review Summary

Fix target is right (normalizing scale before `diffObjMerge` is the correct shape of the solution), but the guard as written is buggy and the block can collapse to one line. The production warp-check path happens to still work because `transformConfigToCheck` is applied to both sides in `warpCheck.ts`, but the unit tests in `configUtils.test.ts` will break and the code relies on a fragile invariant in `EvmWarpRouteReader.fetchScale`.

### Key points

1. **`!== null` guard is effectively a no-op.** `scale` is `.optional()` in `TokenMetadataSchema` -- its type is `... | undefined`, never `null`. `undefined !== null` is always true, so the block runs on every config. Devin's finding is correct; the reply about `normalizeScale` returning `undefined` for `{1n,1n}` conflates `normalizeScale` with `EvmWarpRouteReader.fetchScale` -- the identity-collapse lives in `fetchScale` (`EvmWarpRouteReader.ts:1452-1454`), not in `normalizeScale` (`decimals.ts:36-45`, which returns `{1n,1n}` for `undefined` input).

2. **Tests will break.** `typescript/sdk/src/token/configUtils.test.ts:224-229` (plus inline `tokenFee` cases at 232/258/286/328/386/426) assert against expected outputs that have no `scale` key. With the guard always firing, every transformed output now has `scale: {numerator: 1n, denominator: 1n}` injected. The PR's "fixing tests" comment suggests this was hit locally -- the test updates should be in this PR (or, preferably, the guard should be fixed so tests don't need updating).

3. **Redundant object rebuild.** `normalizeScale` returns exactly `{ numerator: bigint; denominator: bigint }`. The 4-line rebuild is equivalent to `clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale);`.

4. **Inconsistent with siblings.** `allowedRebalancers`, `allowedRebalancingBridges`, and `tokenFee` normalizations in the same function use truthy / optional-chain checks, not `!== null`.

5. **Comment is noise / misleading.** Per CLAUDE.md, comments should explain *why*, not *what*. This one justifies a guard that doesn't do anything.

### Suggested minimal replacement

```ts
if (clonedTokenConfig.scale != null) {
  clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale);
}
```

Drops the comment, matches sibling pattern, keeps tests green, removes the dependency on `fetchScale`'s identity-collapse.

### Minor
- `.changeset/quiet-rivers-flow.md` is missing a trailing newline.

### Observations Outside This PR
- `typescript/sdk/src/token/EvmWarpRouteReader.ts:1452-1454`: the identity-collapse to `undefined` is load-bearing for the current production behavior -- worth a comment there, or a cross-reference, so a future change doesn't silently regress warp-check diffs.

typescript/sdk/src/token/configUtils.ts
Xaroz: We can't do `undefined` because `EvmWarpRouteReader` has the identity thing, and if we remove the identity thing it will cause the inverse to happen where actual is a fraction scale and the expected will be undefined.

For now I think this might require a bit of thinking on how to modify the EvmWarpRouteReader instead of this fragile way of doing things, but that would require modifications of other places too

paulbalaji: Fair point on the `fetchScale` coupling — symmetric always-normalize is more robust than `!= null` here. That still leaves the redundant object rebuild and the misleading comment, though. Minimal tightening that preserves your invariant:

```ts
// normalizeScale(undefined) -> {1n,1n}, matching EvmWarpRouteReader.fetchScale's
// identity-collapse so both sides of the diff agree symmetrically.
clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale);
```

Drops the no-op guard, removes the rebuild (normalizeScale already returns exactly that shape), and the comment now documents the actual cross-file invariant.

Xaroz:
I tested with around 100-ish routes and no scale issue in the checker, removing the null assertion or the if in `EvmWarpRouteReader` causes checker to fail, I will add it on my backlog to revise the whole thing and maybe modify the logic inside `EvmWarpRouteReader` and the utils functions

typescript/sdk/src/token/configUtils.ts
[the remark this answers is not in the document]
Xaroz: Handled in `3c272bf`. Dropped the guard entirely per paulbalaji's suggestion — just `clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale)`. Comment now documents the why (cross-file invariant with fetchScale's identity-collapse).

Xaroz: Handled in `3c272bf`. Applied your suggestion from comment 7 — dropped the guard and rebuild, just `clonedTokenConfig.scale = normalizeScale(clonedTokenConfig.scale)` with a comment documenting the fetchScale invariant.

Xaroz: Applied in `3c272bf`. Thanks for the clean suggestion.

typescript/sdk/src/token/configUtils.test.ts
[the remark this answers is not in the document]
Xaroz: Handled in `3c272bf`. Added 3 tests: plain number scale (`1000000000000` → `{1000000000000n, 1n}`), `{number, number}` object (`{1, 1e12}` → `{1n, 1000000000000n}`), and undefined → identity `{1n, 1n}`.

paulbalaji (approved):
LGTM. All three points addressed cleanly:
- Guard + rebuild collapsed to a single `normalizeScale()` call
- Comment now documents the cross-file invariant with `EvmWarpRouteReader.fetchScale`
- Added tests for plain number, `{num, den}`, and `undefined` paths

Nit (non-blocking): `.changeset/quiet-rivers-flow.md` still missing a trailing newline.

PR #8621, opened 2026-04-17
fix: cctp transfer validation

troykessler:
### Description

Fixes a false \"Insufficient ETH for interchain gas\" error when sending USDC via CCTP v2 warp routes, even when the sender has sufficient ETH.

**Root cause — two bugs compounding:**

1. **Missing `collateralAddressOrDenom` in generated warp configs** ([`generateTokenConfigs`](typescript/cli/src/deploy/warp.ts)): `isCctpTokenConfig` was not included in the condition that sets `collateralAddressOrDenom`, so CCTP collateral tokens were deployed/written to config without the underlying USDC token address. The existing V1 CCTP config has this field; V2 fast/standard configs do not.

2. **Unsafe fungibility fallback in `matchesUnderlyingAsset`** ([`TokenMetadata.ts`](typescript/sdk/src/token/TokenMetadata.ts)): The function has a branch that says "if a collateralized token has no `collateralAddressOrDenom` AND the target is native ETH → treat as fungible." This was intended for `EvmHypNative` (which wraps native ETH and has no token address by definition), but it also fires for any ERC20 collateral token whose config is missing `collateralAddressOrDenom` — including CCTP USDC.

**Effect:** In `validateTokenBalances` (Check 2), `isFungibleWith` incorrectly returned `true` for USDC vs ETH. The balance check then used the sender's USDC raw balance (~939,729 = $0.94) instead of their ETH balance to cover the IGP fee quote (~45,419,750,949,540 wei ≈ 0.000045 ETH). 939,729 < 45,419,750,949,540 → error.

**Fixes:**

- `generateTokenConfigs` in `warp.ts`: add `isCctpTokenConfig(config)` to the condition so CCTP tokens get their `collateralAddressOrDenom` set on deploy (prevents recurrence for future deployments).
- `matchesUnderlyingAsset` in `TokenMetadata.ts`: restrict the native fallback to only fire when the source token's standard is actually a native-type (i.e. present in `PROTOCOL_TO_HYP_NATIVE_STANDARD`), making the SDK defensive against incomplete configs.
- Export `isCctpTokenConfig` from the SDK index so CLI can use it.

### Drive-by changes

None.

### Backward compatibility

Yes. The `matchesUnderlyingAsset` change only tightens an incorrect fallback — tokens with a correct `collateralAddressOrDenom` are unaffected. The `generateTokenConfigs` change only affects newly generated configs.

### Testing

Manual: reproduced the failure with `hyperlane warp send` on the USDC/mainnet-cctp-v2-fast route (Ethereum → Arbitrum) and confirmed the error is gone after these changes.

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8621" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -59,6 +59,7 @@ import {
   normalizeScale,
   splitWarpCoreAndExtendedConfigs,
   tokenTypeToStandard,
+  isOftTokenConfig,
 } from '@hyperlane-xyz/sdk';
 import {
   type Address,
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -50,6 +50,7 @@ import {
   getRouterAddressesFromWarpCoreConfig,
   getSubmitterBuilder,
   getTokenConnectionId,
+  isCctpTokenConfig,
   isCollateralTokenConfig,
   isCrossCollateralTokenConfig,
   isXERC20TokenConfig,
@@ -371,6 +372,7 @@ function generateTokenConfigs(
     const config = warpDeployConfig[chainName];
     const collateralAddressOrDenom =
       isCollateralTokenConfig(config) ||
+      isCctpTokenConfig(config) ||
       isXERC20TokenConfig(config) ||
       isCrossCollateralTokenConfig(config)
         ? (config as { token: string }).token // gets set in the above deriveTokenMetadata()
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -53,6 +53,8 @@ import {
   isCctpTokenConfig,
   isCollateralTokenConfig,
   isCrossCollateralTokenConfig,
+  isDepositAddressTokenConfig,
+  isEverclearCollateralTokenConfig,
   isXERC20TokenConfig,
   normalizeScale,
   splitWarpCoreAndExtendedConfigs,
@@ -374,7 +376,9 @@ function generateTokenConfigs(
       isCollateralTokenConfig(config) ||
       isCctpTokenConfig(config) ||
       isXERC20TokenConfig(config) ||
-      isCrossCollateralTokenConfig(config)
+      isCrossCollateralTokenConfig(config) ||
+      isDepositAddressTokenConfig(config) ||
+      isEverclearCollateralTokenConfig(config)
         ? (config as { token: string }).token // gets set in the above deriveTokenMetadata()
         : undefined;
 
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -378,7 +378,8 @@ function generateTokenConfigs(
       isXERC20TokenConfig(config) ||
       isCrossCollateralTokenConfig(config) ||
       isDepositAddressTokenConfig(config) ||
-      isEverclearCollateralTokenConfig(config)
+      isEverclearCollateralTokenConfig(config) ||
+      isOftTokenConfig(config)
         ? (config as { token: string }).token // gets set in the above deriveTokenMetadata()
         : undefined;
 
--- a/typescript/cli/src/deploy/warp.ts
+++ b/typescript/cli/src/deploy/warp.ts
@@ -50,16 +50,10 @@ import {
   getRouterAddressesFromWarpCoreConfig,
   getSubmitterBuilder,
   getTokenConnectionId,
-  isCctpTokenConfig,
-  isCollateralTokenConfig,
   isCrossCollateralTokenConfig,
-  isDepositAddressTokenConfig,
-  isEverclearCollateralTokenConfig,
-  isXERC20TokenConfig,
   normalizeScale,
   splitWarpCoreAndExtendedConfigs,
   tokenTypeToStandard,
-  isOftTokenConfig,
 } from '@hyperlane-xyz/sdk';
 import {
   type Address,
@@ -374,14 +368,8 @@ function generateTokenConfigs(
   for (const chainName of Object.keys(contracts)) {
     const config = warpDeployConfig[chainName];
     const collateralAddressOrDenom =
-      isCollateralTokenConfig(config) ||
-      isCctpTokenConfig(config) ||
-      isXERC20TokenConfig(config) ||
-      isCrossCollateralTokenConfig(config) ||
-      isDepositAddressTokenConfig(config) ||
-      isEverclearCollateralTokenConfig(config) ||
-      isOftTokenConfig(config)
-        ? (config as { token: string }).token // gets set in the above deriveTokenMetadata()
+      'token' in config && typeof config.token === 'string'
+        ? config.token
         : undefined;
 
     const protocol = multiProvider.getProtocol(chainName);
--- a/typescript/sdk/src/token/TokenMetadata.ts
+++ b/typescript/sdk/src/token/TokenMetadata.ts
@@ -45,8 +45,15 @@ function matchesUnderlyingAsset(
       return true;
     }
 
+    // Only native-type collateral tokens (e.g. EvmHypNative) can be fungible
+    // with native tokens when no collateralAddressOrDenom is set — native
+    // tokens have no address by definition. ERC20 collateral tokens missing
+    // their collateralAddressOrDenom (e.g. incomplete config) must not be
+    // treated as native.
+    const hypNativeStandards = Object.values(PROTOCOL_TO_HYP_NATIVE_STANDARD);
     if (
       !source.collateralAddressOrDenom &&
+      hypNativeStandards.includes(source.standard as TokenStandard) &&
       (target.isNative() || target.isHypNative())
     ) {
       return true;
--- a/typescript/sdk/src/token/TokenMetadata.ts
+++ b/typescript/sdk/src/token/TokenMetadata.ts
@@ -45,15 +45,11 @@ function matchesUnderlyingAsset(
       return true;
     }
 
-    // Only native-type collateral tokens (e.g. EvmHypNative) can be fungible
-    // with native tokens when no collateralAddressOrDenom is set — native
-    // tokens have no address by definition. ERC20 collateral tokens missing
-    // their collateralAddressOrDenom (e.g. incomplete config) must not be
-    // treated as native.
-    const hypNativeStandards = Object.values(PROTOCOL_TO_HYP_NATIVE_STANDARD);
+    // Only HypNative collateral legitimately lacks collateralAddressOrDenom;
+    // any other collateralized token missing it is a config bug, not a native match.
     if (
       !source.collateralAddressOrDenom &&
-      hypNativeStandards.includes(source.standard as TokenStandard) &&
+      source.isHypNative() &&
       (target.isNative() || target.isHypNative())
     ) {
       return true;
--- a/.changeset/little-boxes-yawn.md
+++ b/.changeset/little-boxes-yawn.md
@@ -1,0 +1,6 @@
+---
+"@hyperlane-xyz/cli": patch
+"@hyperlane-xyz/sdk": patch
+---
+
+Fixed cctp transfer validation
--- a/typescript/sdk/src/index.ts
+++ b/typescript/sdk/src/index.ts
@@ -850,6 +850,7 @@ export {
   HypTokenRouterConfigSchema,
   HypTokenRouterVirtualConfig,
   isCollateralRebaseTokenConfig,
+  isCctpTokenConfig,
   isCollateralTokenConfig,
   isMovableCollateralTokenConfig,
   isNativeTokenConfig,


typescript/sdk/src/token/TokenMetadata.ts
```diff
@@ -45,8 +45,15 @@ function matchesUnderlyingAsset(
...
 
+    // Only native-type collateral tokens (e.g. EvmHypNative) can be fungible
+    // with native tokens when no collateralAddressOrDenom is set — native
+    // tokens have no address by definition. ERC20 collateral tokens missing
+    // their collateralAddressOrDenom (e.g. incomplete config) must not be
+    // treated as native.
+    const hypNativeStandards = Object.values(PROTOCOL_TO_HYP_NATIVE_STANDARD);
     if (
       !source.collateralAddressOrDenom &&
+      hypNativeStandards.includes(source.standard as TokenStandard) &&
```
paulbalaji: Three things can be collapsed here:

1. **Redundant cast** (also flagged by Devin, still open): `source.standard` is already `TokenStandard` via `ITokenMetadata` → `TokenArgs` → `z.nativeEnum(TokenStandard)`. CLAUDE.md's zero-tolerance-for-casts rule applies.
2. **Per-call `Object.values`**: recomputed on every invocation. The equivalent logic already exists as an instance method — `TokenMetadata.isHypNative()` at line 129.
3. **Comment length**: 5 lines for a single-line restriction; CLAUDE.md defaults to no comments, one-line max when the *why* is non-obvious.

`source` is typed as `ITokenMetadata`, which exposes `isHypNative()`. Suggest collapsing the added block to:

```ts
// Only HypNative collateral legitimately lacks collateralAddressOrDenom;
// any other collateralized token missing it is a config bug, not a native match.
if (
  !source.collateralAddressOrDenom &&
  source.isHypNative() &&
  (target.isNative() || target.isHypNative())
) {
  return true;
}
```

Drops the cast, the `hypNativeStandards` local, and the `PROTOCOL_TO_HYP_NATIVE_STANDARD` import if unused elsewhere in the file.

typescript/cli/src/deploy/warp.ts
```diff
@@ -371,6 +372,7 @@ function generateTokenConfigs(
     const config = warpDeployConfig[chainName];
     const collateralAddressOrDenom =
       isCollateralTokenConfig(config) ||
+      isCctpTokenConfig(config) ||
```
paulbalaji: Parity gap with `typescript/sdk/src/token/tokenMetadataUtils.ts:81-88`, which already treats `isDepositAddressTokenConfig` and `isEverclearCollateralTokenConfig` the same as the predicates in this branch. Both schemas have a required `token: string` (the underlying ERC20), so they'd slot into the same pattern. Deposit-address maps to `EvmHypCollateral` (in `TOKEN_COLLATERALIZED_STANDARDS`), so the same class of bug this PR fixes for CCTP is still latent there. Fine as a follow-up — flagging here so the two lists can be aligned in a separate PR.

paulbalaji (commented):
## Review Summary

Root cause analysis in the description is accurate and the two fixes compose correctly: the SDK guard removes the false-native validation path, and the CLI change prevents future deploy/apply/extend flows from regenerating bad CCTP warp-core configs. `generateTokenConfigs` in `typescript/cli/src/deploy/warp.ts` is the sole funnel for both deploy and apply/extend paths (via `getWarpCoreConfig`) — no bypass paths. `CctpTokenConfigSchema` requires `token: string`, so the existing `(config as { token: string }).token` cast holds for the new predicate branch.

### Blocker (to land this PR)

- **Missing regression test.** `typescript/sdk/src/token/Token.test.ts:515-637` already exercises `isFungibleWith`, including the positive HypNative fallback case around line 612. There is still no negative case proving that an `EvmHypCollateral` / CCTP token with missing `collateralAddressOrDenom` is **not** fungible with native ETH — which is exactly what this PR fixes. The bug reached users because this case was untested.

### Config vs. code: do both

The code fixes here are worth landing regardless of registry config state. But they should be paired with a registry patch that adds `collateralAddressOrDenom: <USDC address>` to the existing V2 CCTP warp-core configs (`USDC/mainnet-cctp-v2-fast`, `USDC/mainnet-cctp-v2-standard`). V1 already has it. Reason: several TS consumers still read the field directly and will behave incorrectly without it even after this PR lands:
- `typescript/metrics/src/balance.ts:67` and `typescript/warp-monitor/src/monitor.ts:595` use `collateralAddressOrDenom ?? addressOrDenom` to pick the ERC20 to query — today, metrics query the router address as if it were an ERC20 instead of USDC on the router.
- `typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts:143` uses it for collateral funding.

No Rust / relayer / scraper consumers of this field — it is a TS-side concept.

### Observations Outside This PR (follow-up, not blocking)

- `typescript/cli/src/deploy/warp.ts:372-377` vs. `typescript/sdk/src/token/tokenMetadataUtils.ts:81-88`: the CLI predicate set is narrower than the SDK metadata-derivation set. Even after this PR, `isDepositAddressTokenConfig` and `isEverclearCollateralTokenConfig` are still missing from the CLI branch. Deposit-address maps to `EvmHypCollateral` (in `TOKEN_COLLATERALIZED_STANDARDS`), so the same class of bug is latent there. The SDK guard in this PR prevents the false-native misdetection, but the generated warp-core configs will still silently omit `collateralAddressOrDenom`, affecting the metrics / funding consumers listed above. Worth a follow-up to align the two lists.

### Nits

- Changeset description: `Fixed cctp transfer validation` — consider spelling `CCTP` and saying what was fixed, e.g. `Fixed false insufficient-gas errors on CCTP warp transfers by including CCTP tokens in collateralAddressOrDenom generation and tightening matchesUnderlyingAsset's native fallback.`
- `typescript/sdk/src/index.ts:853`: `isCctpTokenConfig` re-export is out of alphabetical order (inserted between `isCollateralRebaseTokenConfig` and `isCollateralTokenConfig`).

Already raised elsewhere: Devin's inline flag of the redundant `as TokenStandard` cast is still open — the inline suggestion on `TokenMetadata.ts` below folds that fix in.

typescript/cli/src/deploy/warp.ts
troykessler: also fixed this in this pr now

typescript/cli/src/deploy/warp.ts
```diff
@@ -371,8 +375,12 @@ function generateTokenConfigs(
...
       isXERC20TokenConfig(config) ||
-      isCrossCollateralTokenConfig(config)
+      isCrossCollateralTokenConfig(config) ||
+      isDepositAddressTokenConfig(config) ||
+      isEverclearCollateralTokenConfig(config) ||
```
paulbalaji: **HIGH — Everclear EVM parity is incomplete downstream.** `TokenStandard.EvmHypEverclearCollateral` and `EvmHypEverclearEth` are not in `TOKEN_COLLATERALIZED_STANDARDS` at `typescript/sdk/src/token/TokenStandard.ts:211-234` — only the Tron variants are. So `isCollateralized()` returns `false` for those standards and the entire body of `matchesUnderlyingAsset` at `TokenMetadata.ts:37-57` is skipped. Setting `collateralAddressOrDenom` here has no effect on `isFungibleWith` for Everclear EVM — the user-visible fee-token resolution bug this PR fixes is still present for those routes.

Either add `EvmHypEverclearCollateral` and `EvmHypEverclearEth` to `TOKEN_COLLATERALIZED_STANDARDS` in this PR, or drop `isEverclearCollateralTokenConfig` from this branch and handle it in a follow-up.

typescript/cli/src/deploy/warp.ts
```diff
@@ -371,8 +375,12 @@ function generateTokenConfigs(
...
+      isOftTokenConfig(config)
```
paulbalaji: Two minor things on the line below (`? (config as { token: string }).token // gets set in the above deriveTokenMetadata()`):

1. The `as { token: string }` cast violates CLAUDE.md's no-cast rule. All seven predicates here narrow to schemas with a required `token: z.string()`, so a union type guard would work, e.g.

   ```ts
   type HasTokenField = CollateralTokenConfig | CctpTokenConfig | XERC20TokenConfig | CrossCollateralTokenConfig | DepositAddressTokenConfig | EverclearCollateralTokenConfig | OftTokenConfig;
   const hasTokenField = (c: HypTokenRouterConfig): c is HasTokenField =>
     isCollateralTokenConfig(c) || isCctpTokenConfig(c) || isXERC20TokenConfig(c) || isCrossCollateralTokenConfig(c) || isDepositAddressTokenConfig(c) || isEverclearCollateralTokenConfig(c) || isOftTokenConfig(c);
   const collateralAddressOrDenom = hasTokenField(config) ? config.token : undefined;
   ```

2. The trailing comment `// gets set in the above deriveTokenMetadata()` is misleading — `config.token` is a user-supplied schema field on every branch, not derived by `deriveTokenMetadata`. Worth removing or rewording.

typescript/sdk/src/token/TokenMetadata.ts
```diff
@@ -45,8 +45,11 @@ function matchesUnderlyingAsset(
       return true;
     }
 
+    // Only HypNative collateral legitimately lacks collateralAddressOrDenom;
+    // any other collateralized token missing it is a config bug, not a native match.
     if (
       !source.collateralAddressOrDenom &&
+      source.isHypNative() &&
```
paulbalaji, line 52: **MED — guard is a no-op on Radix / CosmosNative.** `source.isHypNative()` returns `true` for *any* collateral token of standard `RadixHypCollateral` / `CosmNativeHypCollateral`, because `PROTOCOL_TO_HYP_NATIVE_STANDARD` (`TokenStandard.ts:595-596`) reuses those collateral standards as the "hyp native" standards for those protocols. Verified locally: a `RadixHypCollateral` / `CosmNativeHypCollateral` token without `collateralAddressOrDenom` still matches the native asset.

The guard therefore only tightens the EVM / Sealevel / Cosmos-CW cases — which is what this PR needs, so not a blocker. But the comment on lines 48-49 overstates the invariant ("only HypNative legitimately lacks collateralAddressOrDenom"). Suggest softening the comment or adding a TODO flagging the Radix/CosmosNative ambiguity so a future reader doesn't rely on an invariant that isn't actually enforced cross-protocol.

paulbalaji (commented):
## Review Summary

Thanks for the follow-ups — cast, OFT import, and CLI predicate parity are all in. The core user-visible CCTP bug is correctly fixed for the common EVM collateral path.

**Not ready to approve.** Two blockers and one medium concern remain after a second pass (independently confirmed).

### Blockers

1. **Missing regression test (still unresolved from prior review).** The bug reached users because the negative case wasn't covered. Please add tests to the `isFungibleWith` describe block in `typescript/sdk/src/token/Token.test.ts` (around line 620) asserting that an `EvmHypCollateral` token with no `collateralAddressOrDenom` is **not** fungible with native ETH, in both directions. Suggested slot-in:

   ```ts
   const evmHypCollateralMissingCollateralAddress = new Token({
     chainName: TestChainName.test1,
     standard: TokenStandard.EvmHypCollateral,
     addressOrDenom: '0x51b5234A896FbC4b3e2F7237592D054716762131',
     decimals: 6,
     symbol: 'USDC',
     name: 'USD Coin',
   });

   it('returns false for non-native collateral without collateralAddressOrDenom vs native (CCTP regression)', () => {
     expect(evmHypCollateralMissingCollateralAddress.isFungibleWith(evmNativeToken)).to.be.false;
     expect(evmNativeToken.isFungibleWith(evmHypCollateralMissingCollateralAddress)).to.be.false;
     expect(evmHypCollateralMissingCollateralAddress.isFungibleWith(evmHypNativeToken)).to.be.false;
     expect(evmHypNativeToken.isFungibleWith(evmHypCollateralMissingCollateralAddress)).to.be.false;
   });
   ```

   CCTP collateral maps to `TokenStandard.EvmHypCollateral` (see `collateralCctp` in `PROTOCOL_AND_TOKEN_TYPE_TO_STANDARD`), so this exercises exactly the fixed branch.

2. **Everclear EVM parity is incomplete on the SDK side.** The CLI now writes `collateralAddressOrDenom` for `isEverclearCollateralTokenConfig`, but `typescript/sdk/src/token/TokenStandard.ts:211-234` (`TOKEN_COLLATERALIZED_STANDARDS`) does not include `EvmHypEverclearCollateral` or `EvmHypEverclearEth` — only the Tron variants. Verified locally: `EvmHypEverclearCollateral.isCollateralized() === false`, so the entire `matchesUnderlyingAsset` body is skipped and `isFungibleWith(underlying)` still returns `false` even with `collateralAddressOrDenom` populated. Either add those two standards to `TOKEN_COLLATERALIZED_STANDARDS` in this PR, or drop `isEverclearCollateralTokenConfig` from the CLI change and handle it in a follow-up — as-is the CLI change for Everclear is a no-op downstream. Inline comment on the exact line below.

### Medium (non-blocking, acknowledge or fix)

- **`source.isHypNative()` is a no-op on Radix / CosmosNative.** `PROTOCOL_TO_HYP_NATIVE_STANDARD` at `TokenStandard.ts:595-596` maps `Radix` → `RadixHypCollateral` and `CosmosNative` → `CosmNativeHypCollateral`. Those same standards are used for real collateral tokens on those protocols, so the tightened fallback doesn't tighten there. Verified locally. Pre-existing ambiguity, but the new comment overstates the invariant — inline suggestion below.

### Nits

- Changeset still reads `Fixed cctp transfer validation`. Consider `CCTP` casing and describing what changed, e.g. `Fixed false insufficient-gas errors on CCTP warp transfers by including CCTP/DepositAddress/Everclear/OFT tokens in collateralAddressOrDenom generation and tightening matchesUnderlyingAsset's native fallback.`
- `typescript/sdk/src/index.ts:853`: `isCctpTokenConfig` is inserted between `isCollateralRebaseTokenConfig` and `isCollateralTokenConfig` — out of alphabetical order.
- Still-open from prior review: Devin's flag that OFT is missing from `deriveTokenMetadata` / `prioritizedTypes` in `typescript/sdk/src/token/tokenMetadataUtils.ts` — pre-existing gap, now more visible; fine as a follow-up.

## Observations Outside This PR

- `typescript/sdk/src/token/deploy.ts:170,272` and `typescript/sdk/src/token/EvmWarpModule.ts` / `warpCheck.ts:332` use narrower predicate lists than `generateTokenConfigs` now does — same class of latent bug. Worth aligning in a follow-up so predicate coverage is DRY across deploy / check / warp-core-config paths.
- Registry ask from prior review stands: existing V2 CCTP warp-core configs (`USDC/mainnet-cctp-v2-fast`, `USDC/mainnet-cctp-v2-standard`) still need `collateralAddressOrDenom` added so TS consumers that read the field directly (`typescript/metrics/src/balance.ts:67`, `typescript/warp-monitor/src/monitor.ts:595`, `typescript/infra/scripts/funding/fund-wallet-from-deployer-key.ts:143`) behave correctly.

yorhodes (approved):
lgtm
I think the CCTP route was never fully compatible with warp operations and wouldnt be surprised if WarpCore/the UI also doesnt work for this

typescript/cli/src/deploy/warp.ts
troykessler: dropped everyclear for now, we can make a followup pr if needed

PR #8626, opened 2026-04-17
feat: add BSC and Solana inventory routes for USDC/eclipsemainnet rebalancer

Mo-Hussain:
## Summary
- add BSC inventory/LiFi routing to the USDC/eclipsemainnet rebalancer config
- wire protocol-specific inventory signer keys through rebalancer Helm values/templates
- fix LiFi source protocol resolution when non-EVM chains share numeric chain IDs

### DIFF
--- a/typescript/infra/config/environments/mainnet3/rebalancer/USDC/eclipsemainnet-config.yaml
+++ b/typescript/infra/config/environments/mainnet3/rebalancer/USDC/eclipsemainnet-config.yaml
@@ -1,7 +1,7 @@
 warpRouteId: USDC/eclipsemainnet
 inventorySigners:
   ethereum: '0x6056e8E8e5Db30ffa9d721e3D73b3D558011FdA9'
-  sealevel: '5seKh2p3B8Kq9nAE7X4svfX9uLt81wWAiDJuLK1XNgXf'
+  sealevel: '4ZJoMHQPEMkeExtFQLbuh8nB21dxHj741dSo6oJ5BcMo'
 externalBridges:
   lifi:
     integrator: rebalancer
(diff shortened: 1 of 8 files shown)


typescript/rebalancer/src/bridges/LiFiBridge.ts
paulbalaji, line 1: changeset?

typescript/rebalancer/src/bridges/LiFiBridge.test.ts
```diff
@@ -655,4 +685,15 @@ describe('LiFiBridge constructor chainMetadataByChainId', function () {
...
     }
   });
+
+  it('should prefer Ethereum metadata for LiFi chainId lookups when non-EVM chainIds collide', () => {
+    const bridge = new LiFiBridge(DUPLICATE_CHAIN_ID_CONFIG, testLogger);
+
+    expect((bridge as any).getProtocolTypeForChainId(1)).to.equal(
+      ProtocolType.Ethereum,
+    );
+    expect((bridge as any).getRpcUrlForChainId(1)).to.equal(
```
paulbalaji: nit: `(bridge as any).getProtocolTypeForChainId(...)` / `.getRpcUrlForChainId(...)` reach into private helpers and trip the repo's "zero tolerance for extraneous casts" policy (no `as any`, per `CLAUDE.md`). Prefer asserting observable behavior — e.g. drive a `quote()` / `execute()` with `fromChain: 1` on `DUPLICATE_CHAIN_ID_CONFIG` and check that the Ethereum protocol / RPC is selected downstream — so the test still fails the day these helpers are renamed or inlined. Not blocking.

typescript/infra/config/environments/mainnet3/rebalancer/USDC/eclipsemainnet-config.yaml
```diff
@@ -1,7 +1,7 @@
 warpRouteId: USDC/eclipsemainnet
 inventorySigners:
   ethereum: '0x6056e8E8e5Db30ffa9d721e3D73b3D558011FdA9'
-  sealevel: '5seKh2p3B8Kq9nAE7X4svfX9uLt81wWAiDJuLK1XNgXf'
+  sealevel: '4ZJoMHQPEMkeExtFQLbuh8nB21dxHj741dSo6oJ5BcMo'
```
paulbalaji, line 4: Heads-up (not a code issue): this rotates the sealevel inventory signer from `5seKh2p3…NgXf` to `4ZJoMHQP…BcMo`. `service.ts` derives the address from the configured key and asserts equality on startup, so the GCP secret `hyperlane-mainnet3-key-sealevel-inventoryrebalancer` needs to be updated in lock-step with this merge or the pod will crash on boot. Worth calling out in the PR description / deploy checklist so ops knows to stage the secret first. Not blocking.

paulbalaji (approved):
LGTM overall — approving with two non-blocking notes below.

- The LiFi chainId-collision fix is the right shape: scoping numeric-chainId indexing to EVM and introducing a separate `metadataByDomainId` map removes the read-before-write dependency the old constructor had.
- The Helm wiring keeps a clean backward-compat fallback via `HYP_INVENTORY_KEY`, and `helm.ts` derives `inventorySignerProtocols` from the validated config, so the schema (`z.record(z.nativeEnum(ProtocolType), …)`) is the single source of truth.
- Weights reallocation on `eclipsemainnet-config.yaml` is clean: old `29+24+17+10+5+10+5+10 = 110` → new `27+22+16+10+5+10+5+5+10 = 110`.

Inventory signer support is currently runtime-locked to ethereum + sealevel (`rebalancer/src/service.ts`, `factories/RebalancerContextFactory.ts`), so the sealevel-hardcoded Helm gate is fine for now — if/when a third protocol joins, the gating can generalize to a `range .Values.hyperlane.inventorySignerProtocols`.

PR #8632, opened 2026-04-19
fix(sdk): classify sealevel cross-collateral tokens

nambrot-agent:
## Summary
- fix `TokenMetadata.isCrossCollateralToken()` to include `TokenStandard.SealevelHypCrossCollateral`
- add a focused WarpCore regression test for Sealevel cross-collateral -> EVM cross-collateral routing
- ensure WarpCore takes the cross-collateral `transferRemoteTo` path instead of generic `transferRemote`

## Root cause
`WarpCore.isCrossCollateralTransfer()` relies on `originToken.isCrossCollateralToken()` and `destinationToken.isCrossCollateralToken()`.

After the `TokenMetadata` refactor, current `main` classified only:
- `EvmHypCrossCollateralRouter`
- `TronHypCrossCollateralRouter`

It omitted `SealevelHypCrossCollateral`, so a Sealevel cross-collateral source token was misclassified as non-cross-collateral. That caused WarpCore to skip the cross-collateral route and fall back to the generic transfer path.

## Repro
On current `main`, this targeted test fails:

```bash
pnpm -C typescript/sdk exec mocha --config .mocharc.json ./src/warp/WarpCore.test.ts --grep "Treats Sealevel cross-collateral to EVM cross-collateral as transferRemoteTo route" --exit
```

Failure before fix:

```text
AssertionError: expected false to equal true
```

## Validation
After this change:

```bash
pnpm turbo run build --filter @hyperlane-xyz/sdk^... --filter @hyperlane-xyz/sdk
pnpm -C typescript/sdk exec mocha --config .mocharc.json ./src/warp/WarpCore.test.ts --grep "Treats Sealevel cross-collateral to EVM cross-collateral as transferRemoteTo route" --exit
```

Result:
- `1 passing`

## Notes
- This is a source fix on current `main`, not a release-only workaround.
- I kept the fix scoped to the predicate plus regression coverage.

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8632" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


.changeset/fix-sealevel-cross-collateral-predicate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hyperlane-xyz/sdk': patch
+---
+
+Sealevel cross-collateral tokens were classified as cross-collateral again when building WarpCore routes.
```
paulbalaji, line 5: Minor wording nit (non-blocking): "were classified as cross-collateral again when building WarpCore routes" reads as if the bug is that they *were* being classified. The intent is the opposite — something like "Sealevel cross-collateral tokens are now correctly classified as cross-collateral when building WarpCore routes" or "Fix: Sealevel cross-collateral tokens were no longer classified as cross-collateral...". Cosmetic.

paulbalaji (approved):
## Review Summary

Minimal, correct fix. Replacing the hand-maintained disjunction in `TokenMetadata.isCrossCollateralToken()` with a lookup against the canonical `TOKEN_CROSS_COLLATERAL_STANDARDS` removes the drift risk that caused the regression — now the predicate cannot silently lag behind `TokenStandard.ts`. Regression test is focused and asserts both the classification and the routing outcome.

Prior bot feedback (Devin, CodeRabbit) about reusing the canonical set is already resolved in head. No unresolved threads.

### Related PRs

- #8629 adds the same regression test standalone without the production fix — redundant once this merges.
- #8628 adds the same regression test plus an unrelated registry version bump.

LGTM.

xeno097 (approved):

PR #8635, opened 2026-04-20
fix(rebalancer): align bridge execution with planned quote mode

Mo-Hussain:
## Summary
- align bridge execution with the quote mode chosen during planning
- keep source-limited inventory movements on forward quotes
- avoid reverse-quote overshoot causing retries and livelock

## Stack
- base for #8626

### DIFF
--- a/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
+++ b/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
@@ -1700,24 +1700,29 @@ describe('InventoryRebalancer E2E', () => {
       expect(results[0].success).to.be.true;
       expect(bridge.execute.callCount).to.equal(2);
 
-      const reverseQuoteRequests = bridge.quote
-        .getCalls()
-        .map((call) => call.args[0])
-        .filter((params) => params.toAmount !== undefined);
-      expect(reverseQuoteRequests).to.have.lengthOf(2);
-
-      const requestedOutputs = reverseQuoteRequests
-        .map((params) => params.toAmount as bigint)
-        .sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
       const maxPerSourceOutput = perChainInventory - reservedGas;
-      const totalAvailableCapacity = maxPerSourceOutput * 2n;
+      const executionQuoteRequests = bridge.quote
+        .getCalls()
+        .slice(-2)
+        .map((call) => call.args[0]);
+      expect(executionQuoteRequests).to.have.lengthOf(2);
 
-      expect(requestedOutputs.every((output) => output <= maxPerSourceOutput))
-        .to.be.true;
-      expect(requestedOutputs[0] < maxPerSourceOutput).to.be.true;
-      expect(requestedOutputs[1]).to.equal(maxPerSourceOutput);
-      expect(requestedOutputs[0] + requestedOutputs[1] < totalAvailableCapacity)
-        .to.be.true;
+      const forwardQuoteRequests = executionQuoteRequests.filter(
+        (params) => params.fromAmount !== undefined,
+      );
+      const reverseQuoteRequests = executionQuoteRequests.filter(
+        (params) => params.toAmount !== undefined,
+      );
+      const forwardAmounts = forwardQuoteRequests.map(
+        (params) => params.fromAmount as bigint,
+      );
+
+      expect(
+        forwardQuoteRequests.length + reverseQuoteRequests.length,
+      ).to.equal(2);
+      expect(forwardQuoteRequests.length).to.be.greaterThan(0);
+      expect(forwardAmounts.every((amount) => amount <= maxPerSourceOutput)).to
+        .be.true;
     });
 
     it('applies 5% buffer to total bridge amount', async () => {
@@ -1851,7 +1856,7 @@ describe('InventoryRebalancer E2E', () => {
       expect(quotedTargetOutput).to.equal(1_050_000_000_000_000_000n);
     });
 
-    it('fails bridge execution when reverse quote exceeds planned source capacity', async () => {
+    it('retries forward execution when reverse quote exceeds planned source capacity', async () => {
       const amount = BigInt(1e18);
       const sourceInventory = BigInt(2e18);
       const targetWithBuffer = (amount * 105n) / 100n;
@@ -1898,15 +1903,43 @@ describe('InventoryRebalancer E2E', () => {
               toAmount: targetWithBuffer,
             },
           }),
+        )
+        .onThirdCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: sourceInventory,
+            toAmount: targetWithBuffer - 1n,
+            toAmountMin: targetWithBuffer - 1n,
+            requestParams: {
+              fromChain: 42161,
+              toChain: 1399811149,
+              fromToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              toToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              fromAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              toAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              fromAmount: sourceInventory,
+            },
+          }),
         );
+
+      bridge.execute.resolves({
+        txHash: '0xBridgeTxHash',
+        fromChain: 42161,
+        toChain: 1399811149,
+      });
 
       const results = await inventoryRebalancer.rebalance([route]);
 
       expect(results).to.have.lengthOf(1);
-      expect(results[0].success).to.be.false;
-      expect(results[0].error).to.include('All inventory movements failed');
-      expect(results[0].error).to.include('exceeded planned source capacity');
-      expect(bridge.execute.called).to.be.false;
+      expect(results[0].success).to.be.true;
+      expect(bridge.execute.calledOnce).to.be.true;
+      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.equal(
+        targetWithBuffer,
+      );
+      expect(bridge.quote.getCall(2)?.args[0].fromAmount).to.equal(
+        sourceInventory,
+      );
+      expect(bridge.quote.getCall(2)?.args[0].toAmount).to.be.undefined;
     });
 
     it('continues when some bridges fail', async () => {
@@ -2177,7 +2210,10 @@ describe('InventoryRebalancer E2E', () => {
       expect(results).to.have.lengthOf(1);
       expect(results[0].success).to.be.true;
       expect(bridge.execute.calledOnce).to.be.true;
-      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.equal(tokenBalance);
+      expect(bridge.quote.getCall(1)?.args[0].fromAmount).to.equal(
+        tokenBalance,
+      );
+      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.be.undefined;
     });
 
     it('calculates max viable amount as inventory minus (gasCosts * 20) for native tokens', async () => {
@@ -2276,6 +2312,109 @@ describe('InventoryRebalancer E2E', () => {
       }
       expect(executionTargetOutput > amount).to.be.true;
       expect(executionTargetOutput <= maxViable).to.be.true;
+    });
+
+    it('uses forward quote when target output exactly matches source capacity', async () => {
+      const amount = 1000n;
+      const rawBalance = 1100n;
+      const targetWithBuffer = 1050n;
+
+      const route = createTestRoute({ amount });
+      createTestIntent({ amount });
+
+      inventoryRebalancer.setInventoryBalances({
+        [SOLANA_CHAIN]: 0n,
+        [ARBITRUM_CHAIN]: rawBalance,
+      });
+
+      bridge.quote
+        .onFirstCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: rawBalance,
+            toAmount: targetWithBuffer,
+            toAmountMin: targetWithBuffer,
+          }),
+        )
+        .onSecondCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: rawBalance,
+            toAmount: targetWithBuffer,
+            toAmountMin: targetWithBuffer,
+          }),
+        );
+
+      bridge.execute.resolves({
+        txHash: '0xForwardBoundaryBridgeTxHash',
+        fromChain: 42161,
+        toChain: 1399811149,
+      });
+
+      const results = await inventoryRebalancer.rebalance([route]);
+
+      expect(results).to.have.lengthOf(1);
+      expect(results[0].success).to.be.true;
+      expect(bridge.execute.calledOnce).to.be.true;
+      expect(bridge.quote.getCall(1)?.args[0].fromAmount).to.equal(rawBalance);
+      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.be.undefined;
+    });
+
+    it('retries reverse quote as forward when exact-output exceeds source capacity', async () => {
+      const amount = 1000n;
+      const rawBalance = 1100n;
+      const targetWithBuffer = 1050n;
+
+      const route = createTestRoute({ amount });
+      createTestIntent({ amount });
+
+      inventoryRebalancer.setInventoryBalances({
+        [SOLANA_CHAIN]: 0n,
+        [ARBITRUM_CHAIN]: rawBalance,
+      });
+
+      bridge.quote
+        .onFirstCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: rawBalance,
+            toAmount: 1051n,
+            toAmountMin: 1051n,
+          }),
+        )
+        .onSecondCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: rawBalance + 1n,
+            toAmount: targetWithBuffer,
+            toAmountMin: targetWithBuffer,
+          }),
+        )
+        .onThirdCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: rawBalance,
+            toAmount: 1049n,
+            toAmountMin: 1049n,
+          }),
+        );
+
+      bridge.execute.resolves({
+        txHash: '0xForwardFallbackBridgeTxHash',
+        fromChain: 42161,
+        toChain: 1399811149,
+      });
+
+      const results = await inventoryRebalancer.rebalance([route]);
+
+      expect(results).to.have.lengthOf(1);
+      expect(results[0].success).to.be.true;
+      expect(bridge.execute.calledOnce).to.be.true;
+      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.equal(
+        targetWithBuffer,
+      );
+      expect(bridge.quote.getCall(2)?.args[0].fromAmount).to.equal(rawBalance);
+      expect(bridge.quote.getCall(2)?.args[0].toAmount).to.be.undefined;
     });
 
     it('handles quote failures gracefully by skipping the source chain', async () => {
--- a/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
+++ b/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
@@ -13,7 +13,7 @@ import {
   WarpTxCategory,
   type WarpCore,
 } from '@hyperlane-xyz/sdk';
-import { ProtocolType } from '@hyperlane-xyz/utils';
+import { assert, ProtocolType } from '@hyperlane-xyz/utils';
 
 import { ExternalBridgeType } from '../config/types.js';
 import type { IExternalBridge } from '../interfaces/IExternalBridge.js';
@@ -2161,12 +2161,16 @@ describe('InventoryRebalancer E2E', () => {
         .getCalls()
         .find(
           (call: ReturnType<typeof logger.info.getCall>) =>
-            call.args[1] ===
-            'Parallel inventory movements completed, transferRemote will execute after bridges complete',
+            call.args[0] &&
+            typeof call.args[0] === 'object' &&
+            Object.prototype.hasOwnProperty.call(
+              call.args[0],
+              'totalQuotedOutputMin',
+            ),
         );
 
-      expect(summaryCall).to.exist;
-      expect(summaryCall!.args[0].totalQuotedOutputMin).to.equal(
+      assert(summaryCall, 'Expected summary log with totalQuotedOutputMin');
+      expect(summaryCall.args[0].totalQuotedOutputMin).to.equal(
         BigInt(1.01e18).toString(),
       );
     });
--- a/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
+++ b/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
@@ -1949,6 +1949,89 @@ describe('InventoryRebalancer E2E', () => {
       expect(bridge.quote.getCall(2).args[0].toAmount).to.be.undefined;
     });
 
+    it('fails when forward retry still exceeds planned source capacity', async () => {
+      const amount = BigInt(1e18);
+      const sourceInventory = BigInt(2e18);
+      const targetWithBuffer = (amount * 105n) / 100n;
+
+      const route = createTestRoute({ amount });
+      createTestIntent({ amount });
+
+      inventoryRebalancer.setInventoryBalances({
+        [SOLANA_CHAIN]: 0n,
+        [ARBITRUM_CHAIN]: sourceInventory,
+      });
+
+      bridge.quote
+        .onFirstCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: sourceInventory,
+            toAmount: sourceInventory,
+            toAmountMin: sourceInventory,
+            requestParams: {
+              fromChain: 42161,
+              toChain: 1399811149,
+              fromToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              toToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              fromAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              toAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              fromAmount: sourceInventory,
+            },
+          }),
+        )
+        .onSecondCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: sourceInventory + 1n,
+            toAmount: targetWithBuffer,
+            toAmountMin: targetWithBuffer,
+            requestParams: {
+              fromChain: 42161,
+              toChain: 1399811149,
+              fromToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              toToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              fromAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              toAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              toAmount: targetWithBuffer,
+            },
+          }),
+        )
+        .onThirdCall()
+        .resolves(
+          createMockBridgeQuote({
+            fromAmount: sourceInventory + 1n,
+            toAmount: targetWithBuffer - 1n,
+            toAmountMin: targetWithBuffer - 1n,
+            requestParams: {
+              fromChain: 42161,
+              toChain: 1399811149,
+              fromToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              toToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+              fromAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              toAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+              fromAmount: sourceInventory,
+            },
+          }),
+        );
+
+      const results = await inventoryRebalancer.rebalance([route]);
+
+      expect(results).to.have.lengthOf(1);
+      expect(results[0].success).to.be.false;
+      expect(results[0].error).to.include('All inventory movements failed');
+      expect(results[0].error).to.include('exceeded planned source capacity');
+      expect(bridge.quote.callCount).to.equal(3);
+      expect(bridge.execute.called).to.be.false;
+      expect(bridge.quote.getCall(1).args[0].toAmount).to.equal(
+        targetWithBuffer,
+      );
+      expect(bridge.quote.getCall(2).args[0].fromAmount).to.equal(
+        sourceInventory,
+      );
+      expect(bridge.quote.getCall(2).args[0].toAmount).to.be.undefined;
+    });
+
     it('logs actual successful bridge output floors instead of planned output totals', async () => {
       const amount = BigInt(1e18);
       const perChainInventory = BigInt(0.6e18);
--- a/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
+++ b/typescript/rebalancer/src/core/InventoryRebalancer.test.ts
@@ -2075,6 +2075,61 @@ describe('InventoryRebalancer E2E', () => {
       expect(bridge.execute.called).to.be.false;
     });
 
+    it('preserves non-Error rejection reasons from parallel bridge execution', async () => {
+      const amount = BigInt(1e18);
+      const sourceInventory = BigInt(2e18);
+
+      const route = createTestRoute({ amount });
+      createTestIntent({ amount });
+
+      inventoryRebalancer.setInventoryBalances({
+        [SOLANA_CHAIN]: 0n,
+        [ARBITRUM_CHAIN]: sourceInventory,
+      });
+
+      bridge.quote.onFirstCall().resolves(
+        createMockBridgeQuote({
+          fromAmount: sourceInventory,
+          toAmount: sourceInventory,
+          toAmountMin: sourceInventory,
+          requestParams: {
+            fromChain: 42161,
+            toChain: 1399811149,
+            fromToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+            toToken: '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC',
+            fromAddress: '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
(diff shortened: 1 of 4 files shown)


Mo-Hussain:
@claude review

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1459,17 +1470,90 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
+      sourceChain,
+      targetOutputAmount,
+      targetOutputAmount,
+      this.multiProvider,
+      this.warpCore.multiProvider,
+      this.getTokenForChain.bind(this),
+      this.getInventorySignerAddress(targetChain),
+      isNativeTokenStandard,
+      this.logger,
+    );
```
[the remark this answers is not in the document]
Mo-Hussain: Addressed in `d66c98d`. I removed the per-leg `calculateTransferCosts(...)` call from `executeInventoryMovement`, so route-level `calculateTransferCosts(...)` remains the only source of truth. That removes both the mis-targeted `minViableTransfer` estimate and the extra per-leg gas/fee RPCs.

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -921,11 +924,14 @@ export class InventoryRebalancer implements IInventoryRebalancer {
         source.maxTargetOutput >= remaining
           ? remaining
           : source.maxTargetOutput;
+      const quoteMode: BridgeQuoteMode =
+        source.maxTargetOutput > remaining ? 'reverse' : 'forward';
```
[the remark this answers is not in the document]
Mo-Hussain: Kept the forward tie-break as-is. The existing forward-boundary test still pins that behavior; I did not add the extra inline comment in this follow-up because the patch stayed focused on the correctness issues called out later in the review.

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1459,17 +1470,90 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
+          minViableTransferFormatted: this.formatLocalAmount(
+            minViableTransfer,
+            targetToken,
+          ),
+          adjustedUp: true,
+          intentId: intent.id,
+        },
+        'Over-bridging to minViableTransfer to ensure final transferRemote can complete',
+      );
+    }
```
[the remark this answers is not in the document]
Mo-Hussain: Addressed in `d66c98d`. I removed the `effectiveTargetOutput` / per-leg `minViableTransfer` adjustment entirely from bridge execution, so the misleading forward-mode log path and `adjustedForMinViable` field are gone.

typescript/rebalancer/src/core/InventoryRebalancer.test.ts
```diff
@@ -1898,15 +1903,43 @@ describe('InventoryRebalancer E2E', () => {
...
-      expect(results[0].error).to.include('exceeded planned source capacity');
-      expect(bridge.execute.called).to.be.false;
+      expect(results[0].success).to.be.true;
+      expect(bridge.execute.calledOnce).to.be.true;
+      expect(bridge.quote.getCall(1)?.args[0].toAmount).to.equal(
+        targetWithBuffer,
+      );
+      expect(bridge.quote.getCall(2)?.args[0].fromAmount).to.equal(
+        sourceInventory,
+      );
```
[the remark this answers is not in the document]
Mo-Hussain: Addressed in `d66c98d`. I replaced these optional-chained assertions with explicit `bridge.quote.callCount` checks before indexing the retry calls.

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1459,17 +1470,90 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
+            quotedInput: quote.fromAmount.toString(),
+            maxSourceInput: maxSourceInput.toString(),
+            intentId: intent.id,
+          },
+          'Reverse bridge quote exceeded source capacity, retrying with forward quote',
+        );
+
+        quoteModeUsed = 'forward';
+        quote = await quoteWithMode(quoteModeUsed);
+      }
```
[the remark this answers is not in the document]
Mo-Hussain: Addressed in `d66c98d`. I kept the full-cap forward retry, and added a short comment at the retry site documenting that this fallback is intentional and can drift relative to the planned target output.

Mo-Hussain:
Followed up on the review in `d66c98d`.

Summary:
- removed the per-leg `calculateTransferCosts(...)` / `minViableTransfer` top-up from bridge execution
- kept the aggressive reverse->forward full-cap fallback, but documented that the retry can drift relative to the planned target
- changed bridge aggregation/logging to use actual successful `quotedOutputMin` instead of assuming planned `targetOutput` was delivered
- hardened the retry-path tests with explicit `bridge.quote.callCount` assertions and added coverage for actual-output aggregation

Build + tests passed locally via `pnpm -C typescript/rebalancer build` and `pnpm -C typescript/rebalancer test`.

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1461,15 +1491,43 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
+            ? { fromAmount: maxSourceInput }
+            : { toAmount: targetOutputAmount }),
+          fromAddress,
+          toAddress,
+        });
+
+      let quoteModeUsed = quoteMode;
+      let quote = await quoteWithMode(quoteModeUsed);
+
+      if (quoteModeUsed === 'reverse' && quote.fromAmount > maxSourceInput) {
```
paulbalaji, line 1522: Medium (follow-up, not blocking): the planner derives `maxTargetOutput` from a forward-quote capacity estimate that uses `order: 'RECOMMENDED'` inside `LiFiBridge.ts:324`, but this reverse execution quote ends up at `LiFiBridge.ts:384` with `order: 'CHEAPEST'`. Different routing policies can pick different tools, so a source planned as having headroom can legitimately look over-cap at execution time and fall into the new full-cap forward fallback here — even without any real inventory change. `retriedAsForward` in the log makes this observable; worth watching frequency in production and aligning the `order` setting across capacity / forward / reverse in a follow-up.

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1461,15 +1491,43 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
+        );
+
+        // Spend the full source cap on fallback; minor output drift is acceptable
+        // and will be reconciled by later cycles rather than risking livelock.
+        quoteModeUsed = 'forward';
+        quote = await quoteWithMode(quoteModeUsed);
+      }
 
       const inputRequired = quote.fromAmount;
       if (inputRequired > maxSourceInput) {
```
paulbalaji, line 1543: Nit (test coverage): this failure branch — forward fallback still exceeds `maxSourceInput` — isn't exercised by the new tests. The happy reverse→forward retry at `InventoryRebalancer.test.ts:1862` only covers the path where the forward retry fits. Adding a case where the stubbed forward quote also returns `fromAmount > maxSourceInput` would pin the `"exceeded planned source capacity"` error shape against regressions (particularly relevant given the LiFi `order`-policy mismatch noted above).

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -1573,22 +1639,31 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
-      return { success: true, txHash: result.txHash };
+      return {
+        success: true,
+        txHash: result.txHash,
+        inputRequired,
+        quotedOutput: quote.toAmount,
+        quotedOutputMin: quote.toAmountMin,
+        quoteModeUsed,
+      };
     } catch (error) {
```
paulbalaji, line 1660: Nit (test coverage): execution-time `externalBridge.quote(...)` throws land here, but no test covers this branch — `'continues when some bridges fail'` at `InventoryRebalancer.test.ts:2048` only fails `bridge.execute`, not `bridge.quote`, and `'handles quote failures gracefully'` covers planning-time quotes. Consider a test where the first execute-time quote rejects (network error, not cap-exceeded), to confirm the leg returns `{success: false}` with the mapped error and doesn't trigger the reverse→forward retry path.

typescript/rebalancer/src/core/InventoryRebalancer.test.ts
```diff
@@ -1898,15 +1909,140 @@ describe('InventoryRebalancer E2E', () => {
...
+      const summaryCall = logger.info
+        .getCalls()
+        .find(
+          (call: ReturnType<typeof logger.info.getCall>) =>
+            call.args[1] ===
+            'Parallel inventory movements completed, transferRemote will execute after bridges complete',
+        );
+
+      expect(summaryCall).to.exist;
+      expect(summaryCall!.args[0].totalQuotedOutputMin).to.equal(
```
paulbalaji: Nit: `summaryCall!.args[0]` non-null assertion — the `expect(summaryCall).to.exist` on the line above doesn't narrow for TypeScript, and CLAUDE.md's type-safety rules prefer avoiding `!`. Could use `assert(summaryCall, ...)` from `@hyperlane-xyz/utils` (which does narrow) or an explicit `if (!summaryCall) throw new Error(...)` before the `.args[0]` access.

paulbalaji (approved):
## Review Summary

The quote-mode alignment is a clean fix for the reverse-quote livelock risk. Prior review rounds with @claude and @coderabbitai have already tightened up the executor (per-leg `calculateTransferCosts` removed, unreachable forward-mode log path gone, aggregation now uses actual `quotedOutputMin`, retry tests hardened). The only net-new behavioral risk from this PR is downstream of an existing LiFi routing quirk (see inline on the fallback block). Approving — the remaining items are non-blocking nits / follow-up tests.

### Still-open items already raised elsewhere (not re-flagging inline)
- CodeRabbit: dead ternary at `InventoryRebalancer.ts:1014-1019` (`result.value.success ? undefined : result.value.error` — the `undefined` arm is unreachable since the outer `if` excludes `success === true`).
- CodeRabbit: brittle log-message string match for `summaryCall` lookup at `InventoryRebalancer.test.ts:2034-2040` — consider matching on a stable field like `totalQuotedOutputMin`.
- CodeRabbit: multi-source forward-fallback drift — accepted trade-off per PR notes; `retriedAsForward` field in logs makes it observable in production.

### Observations outside this PR
- `typescript/rebalancer/src/bridges/LiFiBridge.ts:324` vs `:384`: the forward-quote path uses `order: 'RECOMMENDED'` while the reverse-quote path uses `order: 'CHEAPEST'`. Planner capacity is derived from forward quotes, so a source planned as `reverse` under one routing policy can come back over-cap under another and fall into the new full-cap forward fallback. Pre-existing, but this PR elevates the blast radius. Worth a follow-up to align `order` across capacity / forward / reverse, or to document the mismatch.

typescript/rebalancer/src/core/InventoryRebalancer.ts
Mo-Hussain: Addressed in `8cd2342`. Added a regression test where the execute-time `externalBridge.quote(...)` rejects after planning succeeds. It now asserts the leg fails, the mapped quote error bubbles into the route failure, there is no reverse -> forward retry, and `bridge.execute` is never called.

typescript/rebalancer/src/core/InventoryRebalancer.ts
Mo-Hussain: Addressed in `6d92986`. Reverse `toAmount` quotes now use `order: "RECOMMENDED"`, matching planner capacity and forward execution quotes. I also added a focused test in `LiFiBridge.test.ts` that inspects the reverse-quote request and pins the routing policy.

typescript/rebalancer/src/core/InventoryRebalancer.ts
Mo-Hussain: Addressed in `a30a412`. Added a regression test covering the reverse -> forward fallback where the forward retry still comes back over `maxSourceInput`. It now asserts the route fails with `All inventory movements failed`, includes the `exceeded planned source capacity` error, and never executes the bridge tx.

typescript/rebalancer/src/core/InventoryRebalancer.test.ts
Mo-Hussain: Addressed in `bb86c55`. The summary-log test now finds the logger call by the stable `totalQuotedOutputMin` field instead of matching the exact message string, and it uses `assert(summaryCall, ...)` to narrow before accessing `.args[0]`.

Mo-Hussain:
@paulbalaji followed up on the assessed review items now landed on this branch:

- `6d92986` aligns reverse LiFi `toAmount` quotes with the planner/forward `RECOMMENDED` routing policy
- `a30a412` adds the reverse -> forward fallback over-cap regression test
- `8cd2342` adds the execute-time quote rejection regression test
- `bb86c55` hardens the summary-log assertion / removes the test non-null assertion
- `ef87d38` keeps the bridge-result cleanup while restoring proper narrowing in the failure path

`pnpm -C typescript/rebalancer test` and `pnpm -C typescript/rebalancer build` both pass on the updated branch.

Mo-Hussain:
@paulbalaji followed up on the assessed review items now landed on this branch:\n\n-  aligns reverse LiFi  quotes with the planner/forward  routing policy\n-  adds the reverse -> forward fallback over-cap regression test\n-  adds the execute-time quote rejection regression test\n-  hardens the summary-log assertion / removes the test non-null assertion\n-  keeps the bridge-result cleanup while restoring proper narrowing in the failure path\n\n
> @hyperlane-xyz/rebalancer@27.2.12 test /Users/mo/dev/hyperlane/env1/hyperlane-monorepo/typescript/rebalancer
> mocha --config .mocharc.json './src/**/*.test.ts' --exit



  LiFiBridge.execute() route validation
    ✔ should pass validation when all route fields match requestParams (531ms)
    ✔ should throw when route fromChainId does not match requested
    ✔ should throw when route toChainId does not match requested
    ✔ should throw when route fromToken does not match requested
    ✔ should throw when route toToken does not match requested
    ✔ should throw when route toAddress does not match requested
    ✔ should throw when route fromAddress does not match requested
    ✔ should pass validation when fromAmount is omitted in requestParams (287ms)
    ✔ should throw when route fromAmount does not match requested
    ✔ should throw when route fromAmount is zero
    ✔ should match addresses case-insensitively (mixed vs uppercase) (287ms)
    ✔ should throw when requestParams.fromAmount is 0n and route has positive amount
    ✔ should pass validation for toAmount quote path (fromAmount undefined, toAmount present) (296ms)
    ✔ should throw when route toAmount is less than requested for reverse quote
    ✔ should pass validation when route toAmount exceeds requested (reverse quote) (289ms)

  LiFiBridge.quote() input validation
    ✔ should throw when fromAmount is 0n
    ✔ should throw when toAmount is 0n
    ✔ should throw when both fromAmount and toAmount are provided
    ✔ should throw when neither fromAmount nor toAmount is provided

  LiFiBridge.quote() routing policy
    ✔ should use RECOMMENDED order for reverse toAmount quotes

  LiFiBridge.getStatus()
    ✔ should translate Hyperlane domain IDs to LiFi chain IDs before SDK status lookup

  LiFiBridge constructor chainMetadataByChainId
    ✔ should resolve Sealevel protocol using a LiFi chain ID mapped from Hyperlane metadata

  RebalancerConfig
    ✔ should throw when the config file does not exist
    ✔ should load config from file
    ✔ should throw if chains are not configured
    ✔ should throw if no warp route id is configured
    ✔ should load relative params without modifications
    ✔ should load absolute params without modifications
    override functionality
      ✔ should parse a config with overrides
      ✔ should throw when an override references a non-existent chain
      ✔ should throw when an override references itself
      ✔ should allow multiple chain overrides
      ✔ should require externalBridge when override executionType is inventory
      ✔ should require inventorySigners when only override enables inventory execution
    composite strategy validation
      ✔ should throw if CollateralDeficitStrategy is not first in composite
      ✔ should allow CollateralDeficitStrategy first in composite

  per-chain bridge configuration
    ✔ should accept externalBridge field on chain config when using inventory execution
    ✔ should accept bridges.lifi section with integrator and optional defaultSlippage
    ✔ should require externalBridges.lifi when executionType is inventory
    ✔ should require externalBridges.lifi when externalBridge is lifi
    ✔ should require externalBridge field when executionType is inventory

  getAllBridges
    ✔ should return empty array for empty strategies
    ✔ should return bridge from single strategy
    ✔ should return all bridges from multiple strategies
    ✔ should include bridges from per-destination overrides
    ✔ should deduplicate bridges across strategies and overrides
    ✔ should handle overrides without bridge property

  InventoryRebalancer E2E
    Basic Inventory Rebalance (Sufficient Inventory)
      ✔ executes transferRemote when inventory is available on destination chain
      ✔ executes transferRemote with correct parameters (swapped direction)
      ✔ uses IMultiProtocolSigner path for solana transfer txs
      ✔ denormalizes inventory execution amounts but records canonical deposit amount
    Partial Fulfillment (Insufficient Inventory)
      ✔ executes partial transferRemote when maxTransferable >= minViableTransfer
      ✔ uses fee-aware maxTransferable for non-native token fees
      ✔ downgrades full non-native transfers to partial when fee headroom is missing
      ✔ intent remains in_progress after partial fulfillment
    Fee-Aware Probe Fallback
      ✔ bridges when non-native destination inventory is zero
      ✔ returns failure for unrelated fee-aware probe errors
    No Inventory Available
      ✔ returns failure when no inventory on destination chain and no other source available
    Single Intent Architecture
      ✔ takes only first route when multiple routes provided
      ✔ continues existing intent instead of processing new routes
      ✔ returns empty when active intent has in-flight deposit (prevents oscillation)
      ✔ returns empty results when no routes provided and no active intent
      ✔ continues existing not_started intent instead of creating new one
      ✔ continues existing intent by bridging after recoverable fee-aware probe failure
    Error Handling
      ✔ handles transaction send failure
      ✔ handles missing token for chain
      ✔ handles adapter quoteTransferRemoteGas failure
      ✔ throws when inventory signer key is missing
    Native Token IGP Reservation
      ✔ reserves IGP and gas cost when transferring native tokens
      ✔ reduces transfer amount when inventory is limited
      ✔ returns failure when inventory cannot cover costs
      ✔ does not reserve IGP for non-native tokens
    Native Token Fee Quote Reservation
      ✔ deducts tokenFeeQuote when addressOrDenom is undefined (native token)
      ✔ deducts tokenFeeQuote when addressOrDenom is zero address
      ✔ does NOT deduct tokenFeeQuote when addressOrDenom is ERC20 address
      ✔ handles undefined tokenFeeQuote (backward compatibility with v<10)
      ✔ reduces maxTransferable when large tokenFeeQuote is present
    LiFi Bridge Integration (for future inventory_movement)
      ✔ mock utilities create valid quote
      ✔ bridge mock can be configured for quote
    Smart Partial Transfer Threshold
      ✔ does partial transfer when inventory is available on destination
      ✔ does partial transfer when maxTransferable >= minViableTransfer
      ✔ does NOT do partial transfer when maxTransferable < minViableTransfer (native tokens)
    Early Exit for Small Amounts
      ✔ completes intent when amount < minViableTransfer
    Parallel Multi-Source Bridging
      ✔ bridges from multiple sources in parallel
      ✔ does not inflate each split bridge plan to minViableTransfer
      ✔ applies 5% buffer to total bridge amount
      ✔ bridges only the mixed-decimal shortfall after dust partial skip
      ✔ plans LiFi target output in destination-local units for mixed-decimal routes
      ✔ retries forward execution when reverse quote exceeds planned source capacity
      ✔ fails when forward retry still exceeds planned source capacity
      ✔ fails gracefully when execute-time bridge quote rejects
      ✔ logs actual successful bridge output floors instead of planned output totals
      ✔ continues when some bridges fail
      ✔ returns failure when all bridges fail
    Bridge Viability Check
      ✔ filters out sources where gas cost exceeds 10% of inventory
      ✔ proceeds with bridge when total cost is within available balance
      ✔ viability check only applies to native tokens (not ERC20)
      ✔ calculates max viable amount as inventory minus (gasCosts * 20) for native tokens
      ✔ uses forward quote when target output exactly matches source capacity
      ✔ retries reverse quote as forward when exact-output exceeds source capacity
      ✔ handles quote failures gracefully by skipping the source chain

  Rebalancer
    rebalance()
      ✔ should return empty array for empty routes
      ✔ should return success result for single valid route
      ✔ should return failure results for routes that fail preparation
      ✔ should handle mixed success and failure results
    validateRoute()
      ✔ should fail when origin token not found
      ✔ should fail when destination token not found
      ✔ should log scaled route amounts using origin local units
      ✔ should fail when signer is not a rebalancer
      ✔ should fail when destination is not in allowed list
      ✔ should fail when bridge is not allowed
    prepareTransactions()
      ✔ should create failure result when quote fetching throws
      ✔ should create failure result when tx population throws
      ✔ should denormalize canonical route amounts before quote and populate calls
    executeTransactions()
      ✔ should create failure result when gas estimation fails
      ✔ should continue with other routes when one fails gas estimation
      ✔ should group transactions by origin chain
    sendTransactionsForChain()
      ✔ should return error result when send fails
      ✔ should continue sending remaining transactions after one fails
      ✔ should send transactions sequentially within same origin chain
    result building
      ✔ should include messageId when dispatch message found
      ✔ should return success: false when no Dispatch event found
      ✔ should include txHash in result

  RebalancerOrchestrator
    executeCycle() - No Routes
      ✔ should complete cycle when no routes proposed
      ✔ should sync action tracker even when no routes
    executeCycle() - Movable Collateral Routes Only
      ✔ should execute movable collateral routes successfully
      ✔ should handle failed movable collateral routes
    executeCycle() - Inventory Routes Only
      ✔ should execute inventory routes successfully
    executeCycle() - Mixed Routes
      ✔ should execute both movable collateral and inventory routes
    executeCycle() - Continue Inventory Intents
      ✔ should call inventoryRebalancer.rebalance([]) when no routes proposed
      ✔ should NOT call inventoryRebalancer.rebalance([]) when routes are proposed
      ✔ should NOT call inventoryRebalancer.rebalance([]) when inventoryRebalancer is not in rebalancers
    syncActionTracker() Error Handling
      ✔ should warn but continue when syncTransfers fails
      ✔ should sync inventory movement actions when bridge is provided
    Metrics Recording
      ✔ should record success metric when all routes succeed
      ✔ should record failure metric when any route fails
      ✔ should process token metrics when metrics is provided

  RebalancerService
    executeManual()
      ✔ should execute manual rebalance successfully
      ✔ should normalize manual amount to canonical units when token has scale
      ✔ should throw when origin token not found
      ✔ should throw when amount is invalid
      ✔ should throw when amount is zero or negative
      ✔ should throw when origin chain has no bridge configured
      ✔ should throw when in monitorOnly mode
      ✔ should propagate errors from rebalancer
    start()
      ✔ should throw when not in daemon mode
      ✔ should start monitor in daemon mode
    stop()
      ✔ should stop monitor
    daemon mode metrics
      ✔ should record failure metric when rebalance has failed results
      ✔ should record success metric when all rebalance results succeed
      ✔ should record failure metric when rebalance has mixed results
    daemon mode rebalancer calls
      ✔ should call rebalancer with routes from strategy
      ✔ should call rebalancer with multiple routes
      ✔ should not call rebalancer when no routes proposed
    initialization
      ✔ should initialize only once
      ✔ should create metrics when withMetrics is enabled

  ForkIndexer
    initialize
      ✔ should set lastScannedBlock to current block for each chain
    sync - uninitialized behavior
      ✔ should be a no-op if called before initialize()
    sync - block range queries
      ✔ should query events from lastScannedBlock+1 to currentBlock
    sync - message conversion
      ✔ should convert Dispatch events to ExplorerMessage format correctly
    sync - deduplication
      ✔ should deduplicate messages by msg_id (same message not added twice)
    sync - no-op conditions
      ✔ should be a no-op when lastScannedBlock >= currentBlock
    sync - incremental indexing
      ✔ should only index new events on multiple sync() calls (incremental)
    sync - message classification
      ✔ should classify rebalancer txs as addRebalanceAction
      ✔ should classify non-rebalancer txs as addUserTransfer
      ✔ should classify correctly with case-insensitive address comparison
    sync - unknown destination chain
      ✔ should skip messages with unknown destination domain
    sync - multiple chains
      ✔ should index events from all chains

  RebalancerContextFactory
    create() — non-EVM chain handling
      ✔ should skip provider initialization for StarkNet chains
      ✔ should skip provider initialization for Sealevel chains
      ✔ should call getProvider for all chains when all are EVM
      ✔ should fail fast when bridgeMinAcceptedAmount is configured for a chain without a token
      ✔ should fail fast when bridged supply is unavailable during initial collateral calculation
      ✔ should fail early when inventory override origin protocol signer key is missing
      ✔ should fail early when an inventory-relevant chain has no token
      ✔ should fail early when inventory chain uses unsupported protocol

  CollateralDeficitStrategy
    constructor
      ✔ should throw an error when less than two chains are configured
      ✔ should create a strategy with valid config
    getCategorizedBalances
      ✔ should detect deficit when balance is negative and add buffer
      ✔ should treat zero balance as neither surplus nor deficit
      ✔ should treat positive balance as surplus
      ✔ should filter pending rebalances by configured bridges and simulate
      ✔ should filter out pending rebalances with different bridge
      ✔ should handle pending rebalance that fully covers deficit
      ✔ should handle multiple chains with mixed balances
      ✔ should handle empty pending rebalances array
    getRebalancingRoutes
      ✔ should set bridge field on output routes
      ✔ should generate routes from surplus to deficit chains
    filterByConfiguredBridges
      ✔ should filter rebalances by configured bridge for route
      ✔ should include rebalance when bridge matches configured bridge for the route
      ✔ should exclude rebalance when bridge does not match configured bridge for the route
      ✔ should return empty array for undefined pending rebalances
    inventory execution type
      ✔ should create inventory route when config is inventory type
      ✔ should include pending inventory rebalances with matching externalBridge in filter
      ✔ should exclude pending inventory rebalances with non-matching externalBridge

  CompositeStrategy
    constructor
      ✔ should throw an error when less than 2 strategies are provided
      ✔ should throw an error when no strategies are provided
      ✔ should create a strategy with 2+ sub-strategies
    getRebalancingRoutes
      ✔ should concatenate routes from all sub-strategies
      ✔ should pass routes from earlier strategies as proposedRebalances to later strategies
      ✔ should accumulate routes across multiple strategies as proposedRebalances
      ✔ should preserve original pendingRebalances and use proposedRebalances for new routes
      ✔ should preserve pendingTransfers for all strategies
      ✔ should maintain route order (first strategy routes come first)
      ✔ should handle strategies that return no routes
      ✔ should handle undefined inflightContext

  MinAmountStrategy
    constructor
      ✔ should throw an error when less than two chains are configured
      ✔ should create a strategy with minAmount and target using absolute values
      ✔ should create a strategy with minAmount and target using relative values
      ✔ should throw an error when minAmount is negative
      ✔ should throw an error when target is less than min
      ✔ should throw an error when relative target is less than relative min
      ✔ should throw an error when raw balances chains length does not match configured chains length
      ✔ should throw an error when a raw balance is missing
      ✔ should throw an error when a raw balance is negative
    getRebalancingRoutes
      ✔ should return an empty array when all chains have at least the minimum amount
      ✔ should return a single route when a chain is below minimum amount
      ✔ should normalize absolute thresholds for mixed-decimal routes
      ✔ normalizes mixed-decimal local balances before routing
      ✔ should return multiple routes for multiple chains below minimum amount
      ✔ should throw an error when the sum of `target` is greater than the sum of collaterals
      ✔ should keep minAmount validation errors human-readable for mixed-decimal routes
      ✔ should keep fractional minAmount validation errors human-readable across heterogeneous decimals
      ✔ should handle case where there is not enough surplus to meet all minimum requirements by scaling down deficits
      ✔ should not produce zero-amount routes when scaling causes amounts to round to zero
      ✔ should have no surplus or deficit when all at min
      ✔ should consider the target amount with relative configuration

  StrategyFactory
    createStrategy
      ✔ creates a WeightedStrategy when given weighted configuration
      ✔ creates a MinAmountStrategy when given minAmount configuration

  WeightedStrategy
    constructor
      ✔ should throw an error when less than two chains are configured
      ✔ should throw an error when weight is negative
      ✔ should throw an error when the total weight is 0
      ✔ should throw an error when tolerance is less than 0 or greater than 100
    getRebalancingRoutes
      ✔ should throw an error when raw balances chains length does not match configured chains length
      ✔ should throw an error when a raw balance is missing
      ✔ should throw an error when a raw balance is negative
      ✔ should return an empty array when all chains are balanced
      ✔ should return a single route when a chain is unbalanced
      ✔ should return an empty array when a chain is unbalanced but has tolerance
      ✔ should return a single route when two chains are unbalanced and can be solved with a single transfer
      ✔ should return two routes when two chains are unbalanced and cannot be solved with a single transfer
      ✔ should return routes to balance different weighted chains
    bridgeMinAcceptedAmount filtering
      ✔ should filter out routes below bridgeMinAcceptedAmount
      ✔ should keep routes at or above bridgeMinAcceptedAmount

  ActionTracker
    initialize
      ✔ should query for inflight rebalance messages and create synthetic entities
      ✔ should use send_occurred_at for createdAt when available
      ✔ should fall back to Date.now() when send_occurred_at is null
      ✔ should skip action with invalid send_occurred_at timestamp
      ✔ should skip creating action if it already exists
    syncTransfers
      ✔ should create new transfers from Explorer messages
      ✔ should not duplicate transfers that already exist
      ✔ should mark transfers as complete when delivered
    syncRebalanceIntents
      ✔ should mark intents as complete when fully fulfilled
      ✔ should not mark intents as complete if not fully fulfilled
      ✔ should mark unfulfilled intents as failed when TTL exceeded
      ✔ should not expire intents within TTL
    syncRebalanceActions
      ✔ should mark actions as complete when delivered and update parent intent
      ✔ should not mark actions as complete if not delivered
    syncInventoryMovementActions
      ✔ stores pending status on in-progress movement
      ✔ stores not_found status on in-progress movement
      ✔ clears nonPendingSince when status returns to pending
      ✔ preserves existing nonPendingSince on repeated not_found polls
    getInProgressTransfers
      ✔ should return only in_progress transfers
    getActiveRebalanceIntents
      ✔ should return only in_progress intents (origin tx confirmed)
    getPartiallyFulfilledInventoryIntents
      ✔ returns not_started inventory intents
      ✔ returns in_progress inventory intents with partial completion
      ✔ ignores inventory_movement amounts when computing remaining
      ✔ does not return non-inventory intents
      ✔ returns intent with in-flight deposit and sets hasInflightDeposit flag
      ✔ returns intent without in-flight deposit with hasInflightDeposit false
      ✔ returns intent when remaining is 0n but has in-flight deposit
      ✔ does not return fully completed intent with no in-flight deposits
      ✔ skips intent with recent in-flight inventory_movement
      ✔ fails stale movement and returns intent
      ✔ fails stale movement with undefined lastBridgeStatus (pre-deploy data)
      ✔ does not fail long-running pending movement
      ✔ does not fail old movement that recently became non-pending
      ✔ handles mix of recent and stale movements
    createRebalanceIntent
      ✔ should create a new intent with status not_started
    createRebalanceAction
      ✔ should create action and transition intent from not_started to in_progress
      ✔ should not transition intent if already in_progress
    completeRebalanceAction
      ✔ should mark action as complete and mark parent intent complete if fully fulfilled
      ✔ should throw error when action not found
    cancelRebalanceIntent
      ✔ should mark intent as cancelled
    failRebalanceAction
      ✔ should mark action as failed
    Explorer query parameters
      ✔ should pass routersByDomain to getInflightRebalanceActions for warp route filtering
      ✔ should pass routersByDomain to getInflightUserTransfers for warp route filtering
    delivery check synchronization
      ✔ should check delivery status in syncTransfers using adapter
      ✔ should check delivery status in syncRebalanceActions using adapter
      ✔ should keep transfer in_progress when not delivered
      ✔ should check delivery for multiple destinations
      ✔ should pass blockTag when checking delivery in syncTransfers
      ✔ should pass blockTag when checking delivery in syncRebalanceActions
      ✔ should pass undefined blockTag when confirmedBlockTags not provided

  InflightContextAdapter
    getInflightContext
      ✔ should return both pendingRebalances and pendingTransfers
      ✔ should handle empty arrays
      ✔ should correctly convert domain IDs to chain names
      ✔ should handle multiple intents and transfers

  InMemoryStore
    save
      ✔ should save a new entity
      ✔ should overwrite an existing entity with the same id
    get
      ✔ should return undefined for non-existent entity
      ✔ should retrieve an existing entity
    getAll
      ✔ should return empty array when no entities exist
      ✔ should return all entities
    update
      ✔ should throw error when updating non-existent entity
      ✔ should update existing entity
      ✔ should preserve non-updated fields
    delete
      ✔ should delete existing entity
      ✔ should not throw when deleting non-existent entity
    getByStatus
      ✔ should return empty array when no entities match status
      ✔ should return only entities with matching status
    getByDestination
      ✔ should return empty array when no entities match destination
      ✔ should return only entities with matching destination

  getRawBalances
    ✔ should return the bridged supply for the token (EvmHypCollateral)
    ✔ should normalize bridged supply to canonical units when token has scale
    ✔ should return the bridged supply for the token (EvmHypNative)
    ✔ should ignore non supported token standards
    ✔ should ignore tokens that are not in the chains list
    ✔ should throw if the bridged supply is undefined

  scale helpers
    ✔ normalizes local amount to canonical amount
    ✔ denormalizes canonical amount to local amount
    ✔ aligns local amount to exact canonical progress
    ✔ returns identity scale when scale is undefined
    ✔ treats equivalent ratios as identity scale
    ✔ normalizes configured amount using token decimals and scale
    ✔ normalizes sub-unit local dust to zero canonical amount

  bridgeConfig
    ✔ should return the base bridge config when no overrides exist
    ✔ should merge base config with overrides when they exist
    ✔ should handle overrides that change the bridge address
    ✔ should apply override with executionType: inventory
    ✔ should apply override with executionType: inventory and externalBridge
    ✔ should allow override to change executionType from movableCollateral to inventory
    ✔ should preserve base chain executionType when no override exists

  ExplorerClient
    getInflightUserTransfers address encoding
      ✔ encodes EVM router addresses as 20-byte hex bytea
      ✔ encodes Solana router addresses as 32-byte hex bytea
      ✔ encodes Starknet router addresses as hex bytea
      ✔ encodes mixed EVM+Solana routersByDomain correctly
      ✔ handles empty routersByDomain gracefully
    getInflightRebalanceActions address encoding
      ✔ encodes mixed EVM+Solana routersByDomain in originTxRecipients
    hasUndeliveredRebalance post-query validation
      ✔ validates non-EVM router addresses correctly in post-query filter
      ✔ returns false when origin_tx_recipient does not match expected router
      ✔ validates EVM router addresses correctly in post-query filter
    error handling
      ✔ throws on non-200 response from explorer

  parseSolanaPrivateKey
    ✔ parses valid 64-byte JSON array
    ✔ parses valid 64-byte comma-separated
    ✔ parses valid 32-byte seed JSON array and expands to 64 bytes
    ✔ parses valid 32-byte seed comma-separated and expands to 64 bytes
    ✔ throws on "1,2,abc,4" (non-numeric)
    ✔ throws on empty string
    ✔ throws on "256,1,2" (out of range)
    ✔ throws on "-1,1,2" (negative)
    ✔ throws on "1.5,2,3" (non-integer)
    ✔ throws on 10-byte array (wrong length)
    ✔ error message contains HYP_INVENTORY_KEY_SEALEVEL
    ✔ parses valid base58-encoded 64-byte key
    ✔ parses valid base58-encoded 32-byte seed and expands to 64 bytes


  356 passing (2s) and 
> @hyperlane-xyz/rebalancer@27.2.12 build /Users/mo/dev/hyperlane/env1/hyperlane-monorepo/typescript/rebalancer
> tsc both pass on the updated branch.

paulbalaji:
@claude review once more assessing author responses against any existing comments

typescript/rebalancer/src/core/InventoryRebalancer.ts
```diff
@@ -976,27 +998,32 @@ export class InventoryRebalancer implements IInventoryRebalancer {
...
-        const error =
-          result.status === 'rejected'
-            ? result.reason?.message
-            : result.value.error;
+        let error: string | undefined;
+        if (result.status === 'rejected') {
+          error = result.reason?.message;
+        } else if (!result.value.success) {
+          error = result.value.error;
+        }
```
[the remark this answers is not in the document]
Mo-Hussain: Addressed in `7d54bde`. The rejected branch now preserves non-`Error` reasons by handling `Error`, `string`, and fallback stringification for other shapes instead of assuming `.message`. I also added a regression test that forces a non-`Error` rejection through the parallel bridge path and verifies the route-level error keeps the raw rejection detail. Verified with `pnpm -C typescript/rebalancer test` and `pnpm -C typescript/rebalancer build`.

PR #8638, opened 2026-04-20
feat(validator): quorum checkpoint reads in validator

nambrot-agent:
## Description
- route validator safety-critical merkle tree hook reads through quorum-backed providers while keeping indexing/log-heavy paths on fallback providers
- preserve the existing single ordered provider list / secret shape for operators; the validator derives the safety vs fallback split internally
- extend the safety-path split to Tron validators with validator-local quorum over `wallet_solidity` safety hooks

## Drive-by changes
- add focused validator unit coverage for Ethereum safety-hook routing
- add Tron-focused validator unit coverage for quorum success and no-quorum failure

## Related issues
- none linked in this PR

## Backward compatibility
- validator-only change
- no infra, Helm, TS config, or secret-shape changes
- non-validator agents are unchanged
- indexing remains fallback-backed; only safety-path reads moved to quorum
- on Ethereum, operators with a single RPC URL still effectively have no quorum benefit

## Testing
- `cd rust/main && cargo test -p validator`
- result on current branch: `20 passed; 0 failed`

### DIFF
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,0 +1,6 @@
+---
+"@hyperlane-xyz/cli": patch
+"@hyperlane-xyz/sdk": patch
+---
+
+The registry dependency was bumped to 24.3.0 so CLI and SDK consumers pick up tolerant merged-registry overlay reads.
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,5 +1,0 @@
----
-"@hyperlane-xyz/cli": patch
----
-
-The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
--- a/.changeset/registry-bump-cli-sdk.md
+++ b/.changeset/registry-bump-cli-sdk.md
@@ -1,6 +1,5 @@
 ---
 "@hyperlane-xyz/cli": patch
-"@hyperlane-xyz/sdk": patch
 ---
 
-The registry dependency was bumped to 24.3.0 so CLI and SDK consumers pick up tolerant merged-registry overlay reads.
+The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
--- a/rust/main/agents/validator/src/validator.rs
+++ b/rust/main/agents/validator/src/validator.rs
@@ -16,17 +16,23 @@ use hyperlane_base::{
     db::{HyperlaneDb, HyperlaneRocksDB, DB},
     git_sha,
     metrics::AgentMetrics,
-    settings::{ChainConf, CheckpointSyncerBuildError},
-    BaseAgent, ChainMetrics, ChainSpecificMetricsUpdater, CheckpointSyncer, ContractSyncMetrics,
-    ContractSyncer, CoreMetrics, HyperlaneAgentCore, MetadataFromSettings, RuntimeMetrics,
-    SequencedDataContractSync,
+    settings::{
+        ChainConf, ChainConnectionConf, CheckpointSyncerBuildError, MerkleTreeHookIndexer,
+        TryFromWithMetrics,
+    },
+    BaseAgent, ChainMetrics, ChainSpecificMetricsUpdater, CheckpointSyncer, ContractSync,
+    ContractSyncMetrics, ContractSyncer, CoreMetrics, HyperlaneAgentCore, MetadataFromSettings,
+    RuntimeMetrics, SequenceAwareLogStore, SequencedDataContractSync,
 };
 use hyperlane_core::{
-    rpc_clients::RPC_RETRY_SLEEP_DURATION, Announcement, ChainResult, HyperlaneChain,
-    HyperlaneContract, HyperlaneDomain, HyperlaneSigner, HyperlaneSignerExt, Mailbox,
-    MerkleTreeHook, MerkleTreeInsertion, ReorgPeriod, TxOutcome, ValidatorAnnounce, H256, U256,
+    rpc_clients::RPC_RETRY_SLEEP_DURATION, Announcement, ChainResult, CheckpointAtBlock,
+    HyperlaneChain, HyperlaneContract, HyperlaneDomain, HyperlaneSigner, HyperlaneSignerExt,
+    IncrementalMerkleAtBlock, Mailbox, MerkleTreeHook, MerkleTreeInsertion, ReorgPeriod, TxOutcome,
+    ValidatorAnnounce, H256, U256,
 };
-use hyperlane_ethereum::{Signers, SingletonSigner, SingletonSignerHandle};
+use hyperlane_ethereum::{
+    self as h_eth, RpcConnectionConf, Signers, SingletonSigner, SingletonSignerHandle,
+};
 
 use crate::reorg_reporter::{
     LatestCheckpointReorgReporter, LatestCheckpointReorgReporterWithStorageWriter, ReorgReporter,
@@ -38,6 +44,50 @@ use crate::{
 };
 
 const CURSOR_INSTANTIATION_ATTEMPTS: usize = 10;
+
+#[derive(Debug)]
+struct ValidatorSafetyMerkleTreeHook {
+    fallback: Arc<dyn MerkleTreeHook>,
+    safety: Arc<dyn MerkleTreeHook>,
+}
+
+#[async_trait]
+impl MerkleTreeHook for ValidatorSafetyMerkleTreeHook {
+    async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock> {
+        self.fallback.tree(reorg_period).await
+    }
+
+    async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32> {
+        self.fallback.count(reorg_period).await
+    }
+
+    async fn latest_checkpoint(
+        &self,
+        reorg_period: &ReorgPeriod,
+    ) -> ChainResult<CheckpointAtBlock> {
+        self.safety.latest_checkpoint(reorg_period).await
+    }
+
+    async fn latest_checkpoint_at_block(&self, height: u64) -> ChainResult<CheckpointAtBlock> {
+        self.safety.latest_checkpoint_at_block(height).await
+    }
+}
+
+impl HyperlaneChain for ValidatorSafetyMerkleTreeHook {
+    fn domain(&self) -> &HyperlaneDomain {
+        self.fallback.domain()
+    }
+
+    fn provider(&self) -> Box<dyn hyperlane_core::HyperlaneProvider> {
+        self.fallback.provider()
+    }
+}
+
+impl HyperlaneContract for ValidatorSafetyMerkleTreeHook {
+    fn address(&self) -> H256 {
+        self.fallback.address()
+    }
+}
 
 /// A validator agent
 #[derive(Debug, AsRef)]
@@ -174,12 +224,32 @@ impl BaseAgent for Validator {
         let reorg_reporter = Arc::new(reorg_reporter_with_storage_writer) as Arc<dyn ReorgReporter>;
 
         let origin_chain_conf = core.settings.chain_setup(&settings.origin_chain)?.clone();
+        let fallback_origin_chain_conf =
+            Self::validator_chain_conf_with_fallback_rpc(&origin_chain_conf);
+        let safety_origin_chain_conf =
+            Self::validator_chain_conf_with_quorum_rpc(&origin_chain_conf);
 
         let mailbox = origin_chain_conf.build_mailbox(&metrics).await?;
 
-        let merkle_tree_hook = settings
-            .build_merkle_tree_hook(&settings.origin_chain, &metrics)
-            .await?;
+        let merkle_tree_hook = if matches!(
+            origin_chain_conf.connection,
+            ChainConnectionConf::Ethereum(_)
+        ) {
+            let fallback_hook = fallback_origin_chain_conf
+                .build_merkle_tree_hook(&metrics)
+                .await?;
+            let safety_hook = safety_origin_chain_conf
+                .build_merkle_tree_hook(&metrics)
+                .await?;
+            Box::new(ValidatorSafetyMerkleTreeHook {
+                fallback: fallback_hook.into(),
+                safety: safety_hook.into(),
+            }) as Box<dyn MerkleTreeHook>
+        } else {
+            settings
+                .build_merkle_tree_hook(&settings.origin_chain, &metrics)
+                .await?
+        };
 
         let validator_announce = settings
             .build_validator_announce(&settings.origin_chain, &metrics)
@@ -187,16 +257,36 @@ impl BaseAgent for Validator {
 
         let contract_sync_metrics = Arc::new(ContractSyncMetrics::new(&metrics));
 
-        let merkle_tree_hook_sync = settings
-            .sequenced_contract_sync::<MerkleTreeInsertion, _>(
-                &settings.origin_chain,
+        let merkle_tree_hook_sync = if matches!(
+            origin_chain_conf.connection,
+            ChainConnectionConf::Ethereum(_)
+        ) {
+            let log_store: SequenceAwareLogStore<MerkleTreeInsertion> = Arc::new(msg_db.clone());
+            let indexer: MerkleTreeHookIndexer = TryFromWithMetrics::try_from_with_metrics(
+                &fallback_origin_chain_conf,
                 &metrics,
-                &contract_sync_metrics,
-                msg_db.clone().into(),
-                false,
                 false,
             )
             .await?;
+            Arc::new(ContractSync::new(
+                settings.origin_chain.clone(),
+                log_store,
+                indexer,
+                contract_sync_metrics.as_ref().clone(),
+                false,
+            ))
+        } else {
+            settings
+                .sequenced_contract_sync::<MerkleTreeInsertion, _>(
+                    &settings.origin_chain,
+                    &metrics,
+                    &contract_sync_metrics,
+                    msg_db.clone().into(),
+                    false,
+                    false,
+                )
+                .await?
+        };
 
         Ok(Self {
             origin_chain: settings.origin_chain,
@@ -332,6 +422,31 @@ impl BaseAgent for Validator {
 }
 
 impl Validator {
+    fn validator_chain_conf_with_fallback_rpc(origin_chain_conf: &ChainConf) -> ChainConf {
+        Self::validator_chain_conf_with_rpc_connection(origin_chain_conf, |urls| {
+            RpcConnectionConf::HttpFallback { urls }
+        })
+    }
+
+    fn validator_chain_conf_with_quorum_rpc(origin_chain_conf: &ChainConf) -> ChainConf {
+        Self::validator_chain_conf_with_rpc_connection(origin_chain_conf, |urls| {
+            RpcConnectionConf::HttpQuorum { urls }
+        })
+    }
+
+    fn validator_chain_conf_with_rpc_connection(
+        origin_chain_conf: &ChainConf,
+        build_rpc_connection: impl FnOnce(Vec<url::Url>) -> RpcConnectionConf,
+    ) -> ChainConf {
+        let mut chain_conf = origin_chain_conf.clone();
+        if let ChainConnectionConf::Ethereum(conn) = &origin_chain_conf.connection {
+            let mut updated_conn: h_eth::ConnectionConf = conn.clone();
+            updated_conn.rpc_connection = build_rpc_connection(conn.rpc_urls());
+            chain_conf.connection = ChainConnectionConf::Ethereum(updated_conn);
+        }
+        chain_conf
+    }
+
     /// Try to create merkle tree hook contract sync attempts times before giving up.
     async fn try_n_times_to_run_merkle_tree_hook_sync(
         &self,
@@ -608,7 +723,36 @@ impl Validator {
 
 #[cfg(test)]
 mod tests {
+    use async_trait::async_trait;
+    use hyperlane_core::{test_utils::dummy_domain, HyperlaneProvider};
+
     use super::*;
+
+    mockall::mock! {
+        pub MerkleTreeHook {}
+
+        impl Debug for MerkleTreeHook {
+            fn fmt<'a>(&self, f: &mut std::fmt::Formatter<'a>) -> std::fmt::Result;
+        }
+
+        impl HyperlaneChain for MerkleTreeHook {
+            fn domain(&self) -> &HyperlaneDomain;
+            fn provider(&self) -> Box<dyn HyperlaneProvider>;
+        }
+
+        impl HyperlaneContract for MerkleTreeHook {
+            fn address(&self) -> H256;
+        }
+
+        #[async_trait]
+        impl MerkleTreeHook for MerkleTreeHook {
+            async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock>;
+            async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32>;
+            async fn latest_checkpoint(&self, reorg_period: &ReorgPeriod) -> ChainResult<CheckpointAtBlock>;
+            async fn latest_checkpoint_at_block(&self, height: u64) -> ChainResult<CheckpointAtBlock>;
+        }
+    }
+
     #[test]
     fn aleo_announcement_location_exactly_max_minus_null() -> Result<()> {
         // 479 bytes input should be padded to 480 with a single null
@@ -669,5 +813,78 @@ mod tests {
         assert_eq!(out_bytes.len(), 480);
         assert!(out_bytes[input_bytes.len()..].iter().all(|&b| b == 0));
         Ok(())
+    }
+
+    #[tokio::test]
+    async fn validator_safety_merkle_tree_hook_only_routes_latest_checkpoint_to_safety() {
+        let domain = dummy_domain(1337, "test-domain");
+        let checkpoint = CheckpointAtBlock {
+            checkpoint: hyperlane_core::Checkpoint {
+                merkle_tree_hook_address: H256::from_low_u64_be(11),
+                mailbox_domain: domain.id(),
+                root: H256::from_low_u64_be(22),
+                index: 7,
+            },
+            block_height: Some(99),
+        };
+
+        let mut fallback = MockMerkleTreeHook::new();
+        fallback.expect_domain().return_const(domain.clone());
+        fallback
+            .expect_address()
+            .return_const(H256::from_low_u64_be(11));
+        fallback.expect_provider().never();
+        fallback.expect_count().once().return_once(|_| Ok(3));
+        fallback.expect_latest_checkpoint_at_block().never();
+
+        let mut safety = MockMerkleTreeHook::new();
+        safety.expect_domain().return_const(domain.clone());
+        safety
+            .expect_address()
+            .return_const(H256::from_low_u64_be(11));
+        safety.expect_provider().never();
+        safety.expect_count().never();
+        safety
+            .expect_latest_checkpoint()
+            .once()
+            .return_once(|_| Ok(checkpoint));
+        safety
+            .expect_latest_checkpoint_at_block()
+            .once()
+            .with(mockall::predicate::eq(42))
+            .return_once(|height| {
+                Ok(CheckpointAtBlock {
+                    checkpoint: hyperlane_core::Checkpoint {
+                        merkle_tree_hook_address: H256::from_low_u64_be(11),
+                        mailbox_domain: 1337,
+                        root: H256::from_low_u64_be(height),
+                        index: height as u32,
+                    },
+                    block_height: Some(height),
+                })
+            });
+
+        let hook = ValidatorSafetyMerkleTreeHook {
+            fallback: Arc::new(fallback),
+            safety: Arc::new(safety),
+        };
+
+        assert_eq!(hook.count(&ReorgPeriod::None).await.unwrap(), 3);
+        assert_eq!(
+            hook.latest_checkpoint(&ReorgPeriod::None)
+                .await
+                .unwrap()
+                .checkpoint
+                .index,
+            7
+        );
+        assert_eq!(
+            hook.latest_checkpoint_at_block(42)
+                .await
+                .unwrap()
+                .checkpoint
+                .index,
+            42
+        );
     }
 }
--- a/rust/main/agents/validator/src/validator.rs
+++ b/rust/main/agents/validator/src/validator.rs
@@ -17,12 +17,11 @@ use hyperlane_base::{
     git_sha,
     metrics::AgentMetrics,
     settings::{
-        ChainConf, ChainConnectionConf, CheckpointSyncerBuildError, MerkleTreeHookIndexer,
-        TryFromWithMetrics,
+        ChainConf, ChainConnectionConf, CheckpointSyncerBuildError, ContractSyncBuildOptions,
     },
-    BaseAgent, ChainMetrics, ChainSpecificMetricsUpdater, CheckpointSyncer, ContractSync,
-    ContractSyncMetrics, ContractSyncer, CoreMetrics, HyperlaneAgentCore, MetadataFromSettings,
-    RuntimeMetrics, SequenceAwareLogStore, SequencedDataContractSync,
+    BaseAgent, ChainMetrics, ChainSpecificMetricsUpdater, CheckpointSyncer, ContractSyncMetrics,
+    ContractSyncer, CoreMetrics, HyperlaneAgentCore, MetadataFromSettings, RuntimeMetrics,
+    SequencedDataContractSync,
 };
 use hyperlane_core::{
     rpc_clients::RPC_RETRY_SLEEP_DURATION, Announcement, ChainCommunicationError, ChainResult,
@@ -46,15 +45,15 @@ use crate::{
 const CURSOR_INSTANTIATION_ATTEMPTS: usize = 10;
 
 #[derive(Debug)]
-struct ValidatorSafetyMerkleTreeHook {
+struct ValidatorQuorumMerkleTreeHook {
     fallback: Arc<dyn MerkleTreeHook>,
-    safety: Arc<dyn MerkleTreeHook>,
+    quorum: Arc<dyn MerkleTreeHook>,
 }
 
 #[async_trait]
-impl MerkleTreeHook for ValidatorSafetyMerkleTreeHook {
+impl MerkleTreeHook for ValidatorQuorumMerkleTreeHook {
     async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock> {
-        self.safety.tree(reorg_period).await
+        self.quorum.tree(reorg_period).await
     }
 
     async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32> {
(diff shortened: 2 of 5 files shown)


rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,50 @@ use crate::{
...
+struct ValidatorSafetyMerkleTreeHook {
+    fallback: Arc<dyn MerkleTreeHook>,
+    safety: Arc<dyn MerkleTreeHook>,
+}
+
+#[async_trait]
+impl MerkleTreeHook for ValidatorSafetyMerkleTreeHook {
+    async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock> {
+        self.fallback.tree(reorg_period).await
+    }
```
paulbalaji, line 57: **Blocker:** `tree()` staying on the fallback RPC undermines the safety claim. `run_checkpoint_submitters` seeds `tip_tree` and `backfill_target` from this call (`validator.rs:518-543`), and `submit.rs:86` signs up to that target. A leading/Byzantine fallback RPC can therefore still drive the validator to sign checkpoints ahead of quorum during startup and backfill — the exact path this PR is supposed to harden. Recommend routing through `self.safety`.

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,50 @@ use crate::{
...
+
+    async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32> {
+        self.fallback.count(reorg_period).await
+    }
```
paulbalaji, line 61: Same concern as `tree()` above: `count()` is used as the signing-loop bound in `submit.rs:193,472`. If a single fallback RPC reports a stale or forward count, the loop ingests leaves and the validator signs against a target the quorum hasn't confirmed. `latest_checkpoint` at `submit.rs:240` catches a root mismatch by aborting, but that's defense-in-depth, not a clean boundary. Route `count` through `safety` as well — it's a cheap `eth_call`.

rust/main/agents/validator/src/validator.rs
```diff
@@ -174,29 +224,69 @@ impl BaseAgent for Validator {
...
 
         let mailbox = origin_chain_conf.build_mailbox(&metrics).await?;
 
-        let merkle_tree_hook = settings
-            .build_merkle_tree_hook(&settings.origin_chain, &metrics)
-            .await?;
+        let merkle_tree_hook = if matches!(
+            origin_chain_conf.connection,
+            ChainConnectionConf::Ethereum(_)
+        ) {
```
paulbalaji: This gate is `Ethereum(_)` only, so `ChainConnectionConf::Tron(_)` validators silently retain the old fallback-only hook — the helper at line 442 also no-ops on non-Ethereum. If Tron is intentionally out of scope, add a comment and/or log; otherwise extend both this `matches!` and `validator_chain_conf_with_rpc_connection` to the Tron variant.

rust/main/agents/validator/src/validator.rs
```diff
@@ -174,29 +224,69 @@ impl BaseAgent for Validator {
...
+            let fallback_hook = fallback_origin_chain_conf
+                .build_merkle_tree_hook(&metrics)
+                .await?;
+            let safety_hook = safety_origin_chain_conf
+                .build_merkle_tree_hook(&metrics)
+                .await?;
+            Box::new(ValidatorSafetyMerkleTreeHook {
+                fallback: fallback_hook.into(),
+                safety: safety_hook.into(),
+            }) as Box<dyn MerkleTreeHook>
```
paulbalaji: Nit: `Box::new(...) as Box<dyn MerkleTreeHook>` is immediately re-wrapped into `Arc<dyn MerkleTreeHook>` by the caller (`Validator` stores `Arc`). Construct as `Arc::new(ValidatorSafetyMerkleTreeHook { ... })` and avoid the Box→Arc hop. Same for `fallback_hook.into()` / `safety_hook.into()` on the two lines above — `Arc::from(...)` reads more explicitly than relying on `Into`.

rust/main/agents/validator/src/validator.rs
```diff
@@ -174,29 +224,69 @@ impl BaseAgent for Validator {
...
 
         let contract_sync_metrics = Arc::new(ContractSyncMetrics::new(&metrics));
 
-        let merkle_tree_hook_sync = settings
-            .sequenced_contract_sync::<MerkleTreeInsertion, _>(
-                &settings.origin_chain,
+        let merkle_tree_hook_sync = if matches!(
+            origin_chain_conf.connection,
+            ChainConnectionConf::Ethereum(_)
+        ) {
```
paulbalaji: Duplicated `matches!(..., Ethereum(_))` gate (also at line 234). Worth unifying into a single branch — both blocks would benefit from a shared early split so the two Ethereum-only paths can't drift.

rust/main/agents/validator/src/validator.rs
```diff
@@ -174,29 +224,69 @@ impl BaseAgent for Validator {
...
                 false,
             )
             .await?;
+            Arc::new(ContractSync::new(
+                settings.origin_chain.clone(),
+                log_store,
+                indexer,
+                contract_sync_metrics.as_ref().clone(),
+                false,
+            ))
```
paulbalaji: This manually re-implements `Settings::sequenced_contract_sync` (`hyperlane-base/src/settings/base.rs:159-184`). Any future change there — telemetry, broadcast semantics, extra cursor plumbing — won't reach this branch. Prefer adding a small helper like `Settings::sequenced_contract_sync_with_chain_conf(&self, domain, chain_conf, ...)` (or an optional `&ChainConf` override) so both callers share one construction path.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +422,31 @@ impl BaseAgent for Validator {
...
+    }
+
+    fn validator_chain_conf_with_rpc_connection(
+        origin_chain_conf: &ChainConf,
+        build_rpc_connection: impl FnOnce(Vec<url::Url>) -> RpcConnectionConf,
+    ) -> ChainConf {
+        let mut chain_conf = origin_chain_conf.clone();
+        if let ChainConnectionConf::Ethereum(conn) = &origin_chain_conf.connection {
+            let mut updated_conn: h_eth::ConnectionConf = conn.clone();
+            updated_conn.rpc_connection = build_rpc_connection(conn.rpc_urls());
```
paulbalaji, line 621: Two things worth considering here:

1. `HttpQuorum` uses `Quorum::Majority` (`rust/main/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs:87`). With a single RPC URL the safety client is structurally identical to the fallback client — operators get no benefit. An `assert!(urls.len() >= 2, "HttpQuorum safety requires >=2 RPC URLs")` here (or a log + operator-docs note) would prevent silent misconfiguration.
2. `conn.rpc_urls()` panics on `RpcConnectionConf::Ws` (`chains/hyperlane-ethereum/src/config.rs:67`). Validators don't use WS for merkle reads today, but any config that ever set WS on the origin would now crash the validator at startup. A readable assert or returning a `Result` beats a panic.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +422,31 @@ impl BaseAgent for Validator {
...
+            chain_conf.connection = ChainConnectionConf::Ethereum(updated_conn);
+        }
+        chain_conf
```
paulbalaji, line 635: Nit: for non-Ethereum variants this silently returns the clone unchanged. Combined with the duplicated `matches!` at the call sites, the dead branch inside the helper is redundant. If you keep the Ethereum-only gate at the call sites, this helper can take `&h_eth::ConnectionConf` directly and return a `ChainConf`, letting the type system enforce "Ethereum only."

rust/main/agents/validator/src/validator.rs
```diff
@@ -670,4 +814,77 @@ mod tests {
...
+        );
+        assert_eq!(
+            hook.latest_checkpoint_at_block(42)
+                .await
+                .unwrap()
+                .checkpoint
+                .index,
+            42
+        );
+    }
```
paulbalaji, line 1089: Test only verifies trait-method routing on the wrapper. Worth adding:
- Unit tests for `validator_chain_conf_with_fallback_rpc` / `_with_quorum_rpc` — assert the returned `ChainConf.connection` is the expected `RpcConnectionConf` variant with URLs preserved.
- Non-Ethereum pass-through behavior (currently implicit and untested).
- At least a smoke test that the `safety` client is actually wired to `HttpQuorum` end-to-end (construct a `ChainConf`, call the helpers, round-trip through `build_merkle_tree_hook`).

paulbalaji (commented):
## Review Summary

Directionally correct — the split between quorum and fallback carves out a reasonable safety boundary. Before merge, one blocker and a couple of scope questions:

### Blocker
- **`tree()` and `count()` still route through fallback but feed the signing path.** `run_checkpoint_submitters` (`validator.rs:518-543`) seeds `tip_tree` and `backfill_target` from `self.merkle_tree_hook.tree(...)`, and `submit.rs:86,193,472` uses `tree.count()` as the loop bound for what to sign up to. A leading/Byzantine fallback RPC can therefore still cause the validator to sign checkpoints ahead of quorum consensus during startup/backfill — which is precisely the path this PR claims to harden. Root equality at `submit.rs:240` catches a mismatch by crash-looping but doesn't prevent the race from shifting the signing target. Routing `tree`/`count` through `safety` is a one-line change and preserves the framing.

### Scope questions
- **Tron is silently excluded.** `ChainConnectionConf::Tron(_)` is a separate variant; the gate (lines 234, 260) and helper (line 442) only match `Ethereum(_)`, so Tron validators keep the old behavior. Intentional? If yes, worth a comment; if not, extend both sites.
- **`HttpQuorum` degenerates at N=1.** Quorum uses `Quorum::Majority` (`rust/main/chains/hyperlane-ethereum/src/rpc_clients/trait_builder.rs:87`), so single-URL operators get zero benefit. An `assert!(urls.len() >= 2)` on the safety client or a note in operator docs would prevent silent misconfiguration.

### Architecture / style
- The manual `ContractSync::new` path (line 271) duplicates `Settings::sequenced_contract_sync` — future changes there won't flow to the EVM validator. Consider a `sequenced_contract_sync_with_chain_conf` helper on `Settings` that accepts an override `ChainConf`.
- The `matches!(..., Ethereum(_))` gate is duplicated at lines 234 and 260 — extract into a single branch.
- `Box::new(...) as Box<dyn MerkleTreeHook>` at line 244 is immediately re-wrapped into `Arc`; constructing as `Arc` directly is cleaner.
- Tests only verify routing on the wrapper. No coverage for `validator_chain_conf_with_{quorum,fallback}_rpc` (URL preservation, non-Ethereum pass-through), and no integration test for quorum disagreement.

### Unrelated
- Registry bump 23.12.0 → 24.3.0 and its changeset are unrelated to the validator change; bundling is fine but worth mentioning.

Ran `cd rust/main && cargo test -p validator` locally — passes.

nambrot-agent:
Addressed the safety-boundary issue.

Changes in the latest push:
- moved `tree()` to the safety path
- kept remote `MerkleTreeHook::count()` on fallback
- kept indexing/log sync on fallback
- added Tron support for the safety path

Reasoning:
- `tree()` is safety-relevant because validator startup/backfill seeds signing state from `merkle_tree_hook.tree(...)`
- remote `hook.count()` is only used in the startup wait loop, so I kept that on fallback
- the `submit.rs` `tree.count()` in the signing loop is local in-memory state, not the remote wrapper method

Tron note:
- Tron does not have the Ethereum `HttpQuorum` config path, so I implemented validator-local quorum for Tron safety reads by building per-`wallet_solidity` safety hooks and requiring majority agreement for `tree`, `latest_checkpoint`, and `latest_checkpoint_at_block`
- fallback indexing behavior is unchanged

Validation:
- `cd rust/main && cargo test -p validator`

nambrot-agent:
Added Tron-focused unit coverage.

New tests cover:
- majority quorum selection for Tron safety reads
- failure when Tron safety hooks do not reach quorum

Validation:
- `cd rust/main && cargo test -p validator`
- result: `20 passed; 0 failed`

Review state:
- no new actionable comments from the latest CodeRabbit pass
- the earlier CodeRabbit tree-routing nit is stale relative to current head

rust/main/agents/validator/src/validator.rs
nambrot-agent: Fixed in follow-up commits. `tree()` now routes through the safety path because it seeds `tip_tree` / `backfill_target` during validator startup and backfill.

rust/main/agents/validator/src/validator.rs
nambrot-agent: I re-checked this boundary and kept remote `MerkleTreeHook::count()` on fallback intentionally. The safety-sensitive `count()` in `submit.rs` is the local in-memory `IncrementalMerkle::count()`, not the remote hook method. The remote hook count is only used in the startup wait loop, so I left it on fallback.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Handled in follow-up commits. The validator split now supports `ChainConnectionConf::Tron(_)` as well. For Tron, I used validator-local quorum over per-`wallet_solidity` safety hooks for `tree`, `latest_checkpoint`, and `latest_checkpoint_at_block`, while leaving indexing on fallback.

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,163 @@ use crate::{
...
+            let reorg_period = reorg_period.clone();
+            async move { hook.latest_checkpoint(&reorg_period).await }
+        }))
+        .await;
+
+        self.select_quorum_result(
+            results,
+            |a, b| a.checkpoint == b.checkpoint && a.block_height == b.block_height,
+            "Failed to reach quorum for tron latest_checkpoint",
+        )
```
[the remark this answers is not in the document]
nambrot-agent: Good callout. This is intentionally strict for now: I required agreement on both value and observation context to avoid masking cases where nodes report the same checkpoint against different perceived finalized views. That said, it may be too strict for Tron in practice. I did not relax it in this PR; we should watch for spurious quorum failures and loosen to checkpoint/tree equality only if needed.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Agreed on the drift risk. The current head already collapsed the earlier duplicated Ethereum-only check into `validator_uses_split_safety_hook`, which now covers both Ethereum and Tron.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Agreed this should eventually move behind a shared helper. I kept the validator-local override path in this PR to stay scoped to the validator safety change, but I agree the `Settings::sequenced_contract_sync` duplication is real follow-up debt.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Makes sense for the original Ethereum-only version. After adding Tron support the helper now intentionally handles both Ethereum and Tron, so I left it as a chain-level helper rather than narrowing it to `h_eth::ConnectionConf`.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Both are fair follow-ups. I did not add the `>= 2` assertion in this PR because single-URL configs already exist and I did not want to widen scope into operator-facing behavior changes. Likewise, I left the existing `rpc_urls()` assumption in place rather than refactoring the helper to return a `Result` mid-PR. The updated PR body now calls out the single-URL limitation explicitly.

.changeset/registry-bump-cli-sdk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperlane-xyz/cli": patch
+---
+
+The registry dependency was bumped to 24.3.0 so CLI consumers pick up tolerant merged-registry overlay reads.
```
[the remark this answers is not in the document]
nambrot-agent: This thread is about the pre-existing registry bump bundled into the branch. I did not expand the changeset scope as part of the validator work here. I updated the PR description to call the registry bump out as unrelated drive-by context rather than part of the validator change itself.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Added direct unit coverage for the Tron safety wrapper after this comment. I have not added helper-level `ChainConf` roundtrip tests in this PR, though — current coverage is still centered on the validator-local routing boundary rather than config-helper roundtrips.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +529,110 @@ impl BaseAgent for Validator {
...
+        origin_chain_conf: &ChainConf,
+        metrics: &CoreMetrics,
+    ) -> ChainResult<Vec<Arc<dyn MerkleTreeHook>>> {
+        let ChainConnectionConf::Tron(conn) = &origin_chain_conf.connection else {
+            unreachable!("tron safety hooks only supported for tron chains");
+        };
+
+        let mut safety_hooks = Vec::new();
+        let wallet_solidity_urls = if conn.wallet_solidity_urls.is_empty() {
+            conn.wallet_urls.clone()
```
paulbalaji: **Safety regression on empty `wallet_solidity_urls`.** This silently falls back to `wallet_urls` when `wallet_solidity_urls` is empty, which routes safety reads to the non-solidified Tron endpoints. Tron's `MerkleTreeHook` assumes solidified-node reads; falling back to full-node reads weakens the exact guarantee the split is meant to establish.

Baseline wiring in `chains/hyperlane-tron/src/provider/tron.rs:70-74` uses `wallet_solidity_urls` directly with no fallback, so an empty list would already fail there — this PR is introducing new degrade-silent behavior.

Suggest failing closed instead:

```rust
if conn.wallet_solidity_urls.is_empty() {
    return Err(ChainCommunicationError::from_other_str(
        "Tron validator safety hooks require `wallet_solidity_urls`",
    ));
}
```

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,163 @@ use crate::{
...
+        context: &str,
+    ) -> ChainResult<T> {
+        let mut oks = Vec::new();
+        let mut first_err = None;
+
+        for result in results {
+            match result {
+                Ok(value) => oks.push(value),
+                Err(err) => {
+                    if first_err.is_none() {
```
paulbalaji, line 99: Nit: `select_quorum_result` returns the first hook's error. When the first hook errors but the other two return successful-but-disagreeing values, the error message points at the failing RPC and hides the disagreement — the more interesting signal operationally. Consider distinguishing "no oks" vs "oks disagreed" in the returned error.

rust/main/agents/validator/src/validator.rs
```diff
@@ -670,4 +1000,322 @@ mod tests {
...
+            block_height: Some(99),
+        };
+        let divergent_checkpoint = CheckpointAtBlock {
+            checkpoint: hyperlane_core::Checkpoint {
+                merkle_tree_hook_address: H256::from_low_u64_be(11),
+                mailbox_domain: domain.id(),
+                root: H256::from_low_u64_be(23),
+                index: 8,
+            },
+            block_height: Some(100),
```
paulbalaji, line 1118: Nit (test gap): the Tron tests cover all-success-with-quorum and all-success-without-quorum, but not the **partial-failure-with-quorum** case (e.g. 1 hook returns `Err`, 2 return the same `Ok`). That's the case where `join_all` (rather than `try_join_all`) matters most and is the core value proposition of the design. Worth a direct test.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +529,110 @@ impl BaseAgent for Validator {
...
+        })
+    }
+
+    fn validator_chain_conf_with_quorum_rpc(origin_chain_conf: &ChainConf) -> ChainConf {
+        Self::validator_chain_conf_with_rpc_connection(origin_chain_conf, |urls| {
+            RpcConnectionConf::HttpQuorum { urls }
+        })
+    }
+
+    fn validator_chain_conf_with_rpc_connection(
```
paulbalaji, line 614: Nit: `validator_chain_conf_with_rpc_connection` is now used by both EVM and Tron, but the Tron branch destructures a builder-built `RpcConnectionConf` just to extract URLs back out — the closure-based abstraction only makes sense for EVM. Plus the `Ws` arm is an `unreachable!` that's one EVM refactor away from breaking. Consider either making the helper EVM-only again, or collapsing the Tron path to rebuild directly from `conn.rpc_urls` without the closure round-trip.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +529,110 @@ impl BaseAgent for Validator {
...
+                    .await?;
+                let safety_hook = safety_origin_chain_conf
+                    .build_merkle_tree_hook(metrics)
+                    .await?;
+                Ok(Box::new(ValidatorSafetyMerkleTreeHook {
+                    fallback: fallback_hook.into(),
+                    safety: safety_hook.into(),
+                }) as Box<dyn MerkleTreeHook>)
+            }
+            ChainConnectionConf::Tron(_) => {
```
paulbalaji, line 560: Nit: `build_validator_merkle_tree_hook` takes `fallback_origin_chain_conf` and `safety_origin_chain_conf`, but the Tron branch ignores both and rebuilds from `origin_chain_conf` via `build_validator_tron_safety_hooks`. Either pass only what each branch actually needs, or drop the ignored args on the Tron side.

paulbalaji (dismissed):
Re-review at 63010ef, accounting for follow-up commits and author responses.

## Resolved since last round
- `tree()` now routes through the safety path (`validator.rs:57`, `:121-133`) and flows into the `tip_tree` seed at `:704-708` ✓
- Duplicated `matches!` gates collapsed into `validator_uses_split_safety_hook` (`:532-537`) ✓
- Tron safety path added with per-`wallet_solidity` quorum ✓
- CodeRabbit's missing `tree()` test assertion added (`:1024`, `:1035`) ✓
- `count()` on fallback: the author's `submit.rs` rationale is correct — all `tree.count()` calls in `submit.rs` operate on the local `IncrementalMerkle`, not the remote hook ✓

## Still open

### Blocker 1 (safety): empty `wallet_solidity_urls` silently falls through to `wallet_urls`
See inline on `validator.rs:583`. Routes safety reads to non-solidified Tron endpoints — the exact thing the split is meant to avoid.

### Blocker 2 (availability): startup `count()` / `tree()` split-boundary desync
The pre-existing startup gate at `validator.rs:491-505` calls `self.merkle_tree_hook.count(&self.reorg_period)`, which now routes through **fallback**. As soon as fallback reports `count > 0`, the loop exits and `run_checkpoint_submitters()` at `:704-708` calls `tree()` on the **safety** path, then asserts `tip_tree.count() > 0` at `:713`. If the safety quorum hasn't converged yet (one RPC lagging, partial agreement, etc.), the assertion fires and the validator crash-loops at startup — a new regression at the split boundary.

The author's earlier rationale for keeping `count()` on fallback was about the `submit.rs` signing loop, which I agree with. But this startup gate is a one-shot call; routing *just this one call site* through safety (or replacing the `:713` assertion with a retry) closes the boundary cleanly. Prefer routing through safety.

(These lines are pre-existing, so I couldn't attach an inline to them — flagging in the body.)

## Scoped-out items I am not re-raising
- `Settings::sequenced_contract_sync` duplication — author acknowledged as follow-up debt.
- HttpQuorum `>= 2` URL assert — author documented the single-URL limitation in the PR body.
- Helper-level `ChainConf` roundtrip tests — nice-to-have, not blocking.

Nits are inline and optional.

## Local validation
`cd rust/main && cargo test -p validator` — passes.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Addressed in the latest push. `select_quorum_result` now distinguishes between total hook failure vs successful hooks disagreeing, so disagreement returns a dedicated error instead of surfacing the first hook error.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Fair nit. I left the signature as-is in this round because the substantive behavior fixes were higher priority, but agreed the Tron branch does not need both override configs passed through.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Fixed in the latest push. The Tron safety hook builder now fails closed if `wallet_solidity_urls` is empty instead of silently falling back to `wallet_urls`. That preserves the solidified-read boundary rather than degrading it.

rust/main/agents/validator/src/validator.rs
nambrot-agent: I did not fully split the helper in this push, but I did keep the Tron path direct and left the current shape as follow-up cleanup. The more important fixes in this round were the fail-closed Tron safety boundary and startup split-boundary regression.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Added in the latest push. There is now a direct Tron unit test for the partial-failure-with-quorum case: one safety hook errors, two agree, and quorum still succeeds.

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,170 @@ use crate::{
...
+
+#[derive(Debug)]
+struct ValidatorTronSafetyMerkleTreeHook {
+    fallback: Arc<dyn MerkleTreeHook>,
+    safety_hooks: Vec<Arc<dyn MerkleTreeHook>>,
+}
+
+impl ValidatorTronSafetyMerkleTreeHook {
+    fn quorum_threshold(&self) -> usize {
+        self.safety_hooks.len() / 2 + 1
```
paulbalaji: **CI lint fix needed.** Clippy is failing here with `clippy::arithmetic_side_effects` (crate has `#![deny(...)]` at `agents/validator/src/main.rs:6`):

```
error: arithmetic operation that can potentially result in unexpected side-effects
  --> agents/validator/src/validator.rs:84:9
   |
84 |         self.safety_hooks.len() / 2 + 1
```

Overflow is not reachable in practice (`Vec::len()` is bounded), but the lint is cosmetic. Cleanest fix:

```rust
(self.safety_hooks.len() / 2).saturating_add(1)
```

Approving on substance; just blocking merge on CI.

paulbalaji (approved):
Delta re-review at `05f68c7c84`. Both blockers from my previous review are fixed; nits are addressed or acknowledged as follow-up.

## Resolved since last round ✓
- **Blocker 1 (safety)** — `build_validator_tron_safety_hooks` at `validator.rs:588-590` now fails closed when `wallet_solidity_urls` is empty, preserving the solidified-read boundary.
- **Blocker 2 (availability)** — startup gate at `:496-512` now calls `merkle_tree_hook.tree()` on the safety path instead of `.count()` on fallback, so the startup view and the `tip_tree` assertion in `run_checkpoint_submitters` share the same source of truth. Desync closed.
- **Nit 1** — `select_quorum_result` at `:87-122` now distinguishes "no oks" from "oks disagreed" with a dedicated error carrying the agreeing-ok count.
- **Nit 2** — new test at `:1330` (`validator_tron_safety_merkle_tree_hook_tolerates_partial_failures`) exercises the 1-err + 2-ok-agree → quorum-succeeds case.
- Acknowledged: helper cleanup (`validator_chain_conf_with_rpc_connection` Tron branch) and `build_validator_merkle_tree_hook` unused-arg cleanup deferred as follow-up — fine.

## One small CI fix before merge
Clippy is failing on `validator.rs:84` — see inline. Trivial, doesn't touch logic.

## Local validation
`cd rust/main && cargo test -p validator` — 21/21 passed.

Approving on substance; just needs the clippy fix for CI.

.changeset/registry-bump-cli-sdk.md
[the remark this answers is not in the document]
nambrot: This should not be in here

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +45,170 @@ use crate::{
 
 const CURSOR_INSTANTIATION_ATTEMPTS: usize = 10;
 
+#[derive(Debug)]
+struct ValidatorSafetyMerkleTreeHook {
+    fallback: Arc<dyn MerkleTreeHook>,
+    safety: Arc<dyn MerkleTreeHook>,
```
nambrot: Nit: call this quorum so that the terminology is consistent

rust/main/agents/validator/src/validator.rs
nambrot: Agree with @paulbalaji here to do this now

rust/main/agents/validator/src/validator.rs
nambrot: It doesn't look like you did?

.changeset/registry-bump-cli-sdk.md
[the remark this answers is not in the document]
nambrot-agent: Removed in `9ddfeb7b4b`. I dropped the unrelated registry bump from this PR by deleting the changeset and reverting the `pnpm-workspace.yaml` / `pnpm-lock.yaml` changes, so the branch is back to validator-only scope.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Fixed in `9ddfeb7b4b`. The drift is actually removed now: both the hook split and the sync/indexer override key off the same `validator_uses_split_quorum_hook(...)` branch, and the sync path now goes through shared `Settings::sequenced_contract_sync_with_chain_conf(...)` instead of the validator-local `ContractSync::new(...)` duplication. That also means Tron now gets the fallback override on the sync path too.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Renamed in `9ddfeb7b4b`. The wrapper types, helpers, and tests now use `quorum` terminology instead of `safety` where this PR is specifically about the quorum-backed read path.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Fixed in `9ddfeb7b4b`. `quorum_threshold()` now uses `(self.quorum_hooks.len() / 2).saturating_add(1)`, and `cargo clippy -p validator -- -D warnings` passes locally.

rust/main/agents/validator/src/validator.rs
nambrot-agent: Done in `9ddfeb7b4b`. I moved the validator override onto a shared helper in `hyperlane-base` (`sequenced_contract_sync_with_chain_conf`) and switched the validator sync construction to use it, so the fallback-indexer path no longer reimplements `Settings::sequenced_contract_sync`.

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +536,110 @@ impl BaseAgent for Validator {
...
+    ) -> ChainResult<Vec<Arc<dyn MerkleTreeHook>>> {
+        let ChainConnectionConf::Tron(conn) = &origin_chain_conf.connection else {
+            unreachable!("tron safety hooks only supported for tron chains");
+        };
+
+        if conn.wallet_solidity_urls.is_empty() {
+            return Err(ChainCommunicationError::from_other_str(
+                "Tron validator safety hooks require wallet_solidity_urls",
+            ));
+        }
```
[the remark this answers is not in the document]
nambrot-agent: Agreed. The fail-closed `wallet_solidity_urls` requirement is intentional here because the quorum path is specifically meant to stay on solidity endpoints; silently falling back to `wallet_urls` weakens that boundary. I left the behavior strict and this comment is a good callout for operators configuring Tron validators on the quorum path.

rust/main/agents/validator/src/validator.rs
```diff
@@ -174,29 +343,59 @@ impl BaseAgent for Validator {
         let reorg_reporter = Arc::new(reorg_reporter_with_storage_writer) as Arc<dyn ReorgReporter>;
 
         let origin_chain_conf = core.settings.chain_setup(&settings.origin_chain)?.clone();
+        let fallback_origin_chain_conf =
+            Self::validator_chain_conf_with_fallback_rpc(&origin_chain_conf);
+        let quorum_origin_chain_conf =
+            Self::validator_chain_conf_with_quorum_rpc(&origin_chain_conf);
```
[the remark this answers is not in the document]
nambrot: Not intended to be used with websocket

rust/main/agents/validator/src/validator.rs
```diff
@@ -332,6 +531,110 @@ impl BaseAgent for Validator {
 }
 
 impl Validator {
+    fn validator_uses_split_quorum_hook(origin_chain_conf: &ChainConf) -> bool {
+        matches!(
+            origin_chain_conf.connection,
+            ChainConnectionConf::Ethereum(_) | ChainConnectionConf::Tron(_)
+        )
```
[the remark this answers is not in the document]
nambrot: Yes that is the explicit point

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
+    async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock> {
+        let results = join_all(self.quorum_hooks.iter().cloned().map(|hook| {
+            let reorg_period = reorg_period.clone();
+            async move { hook.tree(&reorg_period).await }
+        }))
+        .await;
+
+        self.select_quorum_result(
+            results,
+            |a, b| a.tree == b.tree && a.block_height == b.block_height,
```
tkporter, line 136: could probably just derive Eq, PartialEq on `IncrementalMerkleAtBlock` directly to make this comparison a more simple `a == b`

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
+    ) -> ChainResult<CheckpointAtBlock> {
+        let results = join_all(self.quorum_hooks.iter().cloned().map(|hook| {
+            let reorg_period = reorg_period.clone();
+            async move { hook.latest_checkpoint(&reorg_period).await }
+        }))
+        .await;
+
+        self.select_quorum_result(
+            results,
+            |a, b| a.checkpoint == b.checkpoint && a.block_height == b.block_height,
```
tkporter, line 157: ditto, can derive `Eq, PartialEq` on CheckpointAtBlock to make this comparison simpler

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
+            self.quorum_hooks
+                .iter()
+                .cloned()
+                .map(|hook| async move { hook.latest_checkpoint_at_block(height).await }),
+        )
+        .await;
+
+        self.select_quorum_result(
+            results,
+            |a, b| a.checkpoint == b.checkpoint && a.block_height == b.block_height,
```
tkporter, line 173: same point for `Eq, PartialEq` on CheckpointAtBlock

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
+}
+
+#[async_trait]
+impl MerkleTreeHook for ValidatorQuorumMerkleTreeHook {
+    async fn tree(&self, reorg_period: &ReorgPeriod) -> ChainResult<IncrementalMerkleAtBlock> {
+        self.quorum.tree(reorg_period).await
+    }
+
+    async fn count(&self, reorg_period: &ReorgPeriod) -> ChainResult<u32> {
+        self.fallback.count(reorg_period).await
```
tkporter, line 60: Feels weird to me to not just use quorum here - even though we can reason about things to know that if the count is incorrect, safety isn't really violated (honestly it's not super clear to me that it's totally safe from a presence of bugs perspective tbh), it feels a bit like an abstraction violation to make that assumption here, especially for something called `ValidatorQuorumMerkleTreeHook` that seems to say that quorum is used everywhere. If things ever change I'd expect someone using `ValidatorQuorumMerkleTreeHook::count` to expect it to use a quorum to determine this

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
 const CURSOR_INSTANTIATION_ATTEMPTS: usize = 10;
 
+#[derive(Debug)]
+struct ValidatorQuorumMerkleTreeHook {
+    fallback: Arc<dyn MerkleTreeHook>,
+    quorum: Arc<dyn MerkleTreeHook>,
+}
+
+#[async_trait]
+impl MerkleTreeHook for ValidatorQuorumMerkleTreeHook {
```
tkporter, line 54: just thinking out loud - if we have a lot of messages coming in and the set of RPCs are on slightly different blocks frequently (e.g. fast block time, or some are ahead of others, etc), then it might be hard to get quorum. but we operate on quorum already anyways for EVM validators so this seems like a non-issue, just flagging

rust/main/agents/validator/src/validator.rs
```diff
@@ -39,6 +44,170 @@ use crate::{
...
+    fallback: Arc<dyn MerkleTreeHook>,
+    quorum_hooks: Vec<Arc<dyn MerkleTreeHook>>,
+}
+
+impl ValidatorTronQuorumMerkleTreeHook {
+    fn quorum_threshold(&self) -> usize {
+        (self.quorum_hooks.len() / 2).saturating_add(1)
+    }
+
+    fn select_quorum_result<T: Clone>(
```
tkporter, line 86: might be nice to pull this into a nicer helper at some point to easily build this out for alt VMs, but just thinking out loud (we have something similar for fallback provider)

tkporter (approved):
mostly nits

PR #8642, opened 2026-04-20
feat(sealevel): sealevel token program fee support [IGNORE FOR NOW]

xeno097:
### Description

  Implements the SVM fee quoting program for Hyperlane warp routes — onchain static fee curves with offchain signed quote overrides, matching EVM's fee system functionality.

  **Fee program (`hyperlane-sealevel-fee`):**

  - Three fee modes: `Leaf` (direct curve), `Routing` (per-destination domain), `CrossCollateralRouting` (per destination + target router)
  - Three curve types: Linear, Regressive, Progressive — all with `max_fee` / `half_amount` parameters
  - Offchain quoting via secp256k1 signed quotes (transient and standing), verified against per-route signer sets
  - Per-route signer isolation matching EVM's per-contract signer model: Leaf signers on `FeeData`, route signers on `RouteDomain` / `CrossCollateralRoute` PDAs, wildcard-domain signers on routed `FeeData` variants
  - Quote cascade: transient → domain standing → wildcard-domain standing → onchain fallback
  - Standing quotes stored in per-domain PDAs with BTreeMap of recipient → quote values
  - Transient quotes created and autoclosed within the same transaction (EIP-1153 pattern)
  - CC standing quotes track auth provenance (`Direct` vs `CcDefaultFallback`) — quotes authorized via DEFAULT_ROUTER fallback are invalidated once a router-specific route is created
  - `min_issued_at` global revocation threshold for emergency signer compromise
  - `ValidatableQuote` trait unifies expiry + `min_issued_at` validation across quote types
  - Simulation instructions (`GetQuoteAccountMetas`, `GetSubmitQuoteAccountMetas`) for SDK account discovery
  - `GetProgramVersion` uses a universal hash-based discriminator (`sha256("hyperlane:get-program-version")[:8]`) in the shared `package-versioned` crate, enabling consistent version queries across all Hyperlane SVM programs without
  coupling to any program's instruction enum

  **Supporting crates:**

  - `package-versioned` — shared versioning trait + universal `GetProgramVersion` discriminator
  - `quote-verifier` — secp256k1 signature verification for SVM signed quotes, compiled into the fee program

  **Signer authority matrix:**

  | Fee type | Quote destination | Auth source |
  |---|---|---|
  | Leaf | exact or wildcard | `LeafFeeConfig.signers` |
  | Routing | exact domain | `RouteDomain.signers` |
  | Routing | wildcard domain | `RoutingFeeConfig.wildcard_signers` |
  | CC | exact domain | resolved `CrossCollateralRoute.signers` |
  | CC | wildcard domain | `CrossCollateralRoutingFeeConfig.wildcard_signers` |

  ### Drive-by changes

  - Fee program added to `build-programs.sh` token paths and SVM SDK `generate-program-bytes` / `setup.ts`
  - PDA key mismatch errors use `InvalidArgument` (aligns with mailbox convention)

  ### Related issues

  

  ### Backward compatibility

  Yes — this is an entirely new program and two new library crates. No changes to existing deployed programs or their instruction sets. The `GetProgramVersion` universal discriminator in `package-versioned` is additive and can be adopted
  by other programs independently.

  ### Testing

  Unit Tests + Functional Tests — 63 unit tests + 144 functional tests (207 total), all passing.

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8642" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+    let payer_info = next_account_info(accounts_iter)?;
+    if !payer_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
+
+    // Account 2: Fee account PDA.
+    let fee_account_info = next_account_info(accounts_iter)?;
+    verify_account_uninitialized(fee_account_info)?;
+    let (fee_account_key, fee_account_bump) =
+        Pubkey::find_program_address(fee_account_pda_seeds!(data.salt), program_id);
```
paulbalaji, line 138: **High — `InitFee` has no caller authorization (owner squat).**

`data.owner`, `data.beneficiary`, `data.fee_data`, `data.salt`, and `data.domain_id` are all attacker-controlled; the only signer check above is on `payer_info`, and the PDA is derived solely from `salt`. A front-runner who sees an intended deployment can call `InitFee` first with the same salt and become owner. Once initialized, `verify_account_uninitialized` prevents the legitimate caller from ever taking the slot back.

Suggested fix: bind the salt into a deployer-authority seed (`seeds = ["fee", deployer_pubkey, salt]` with `deployer_pubkey` required to sign) or gate init behind the program's upgrade authority.

rust/sealevel/libraries/quote-verifier/src/lib.rs
```diff
@@ -0,0 +1,661 @@
...
+    ) -> H256 {
+        let context_hash = keccak::hash(&self.context);
+        let data_hash = keccak::hash(&self.data);
+
+        let hash = keccak::hashv(&[
+            DOMAIN_TAG,
+            fee_account.as_ref(),
+            &domain_id.to_le_bytes(),
+            context_hash.as_ref(),
+            data_hash.as_ref(),
```
paulbalaji, line 95: **High — signed hash does not commit to the curve variant.**

The signed message hashes `context_hash || data_hash`, where `data` is just `max_fee` (u64 LE) + `half_amount` (u64 LE). The curve variant (Linear / Regressive / Progressive) is absent from both `data` and `context`, and the stored quote (`FeeStandingQuoteValue` / `TransientQuote`) only records params.

At consumption, `processor.rs:446` / `:524` clone the *current* on-chain `FeeDataStrategy` and swap in the quoted params. If an admin rotates the curve via `UpdateFeeParams` after a quote is signed, the quote retroactively executes under the new curve — a signer has no way to express "I only authorize this for Linear."

Suggested fix: include a curve-variant discriminator byte in the signed hash (or in `data`), and at `QuoteFee` time reject quotes whose recorded variant differs from the on-chain strategy's variant. Also add a 1-byte version and a kind tag (Transient vs Standing, Leaf vs Routing vs CC) for defence-in-depth against future protocol evolution.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+
+    // Parse and validate context using the generic context type.
+    let ctx = C::try_from_bytes(&transient.context).map_err(|_| Error::TransientContextMismatch)?;
+    ctx.validate(quote_fee_data)?;
+
+    // Parse quote data and compute fee using on-chain curve with quoted params.
+    let quote_data = FeeQuoteData::try_from(transient.data.as_slice())
+        .map_err(|_| Error::InvalidTransientData)?;
+
+    let mut quoted_strategy = strategy.clone();
```
paulbalaji: **High (companion to the signed-hash finding).**

`let mut quoted_strategy = strategy.clone();` clones whatever curve variant is currently in the on-chain `fee_data`, then only overwrites the params. Combined with the fact that the signed hash does not commit to the curve variant, an already-authorized quote applies under whichever curve is active at consumption time — not the one the signer saw.

Fix together with the verifier change: when a quote is consumed, refuse if the stored/authorized variant differs from `strategy`'s variant. Same pattern needed at line ~524 for standing-quote consumption.

rust/sealevel/programs/hyperlane-sealevel-fee/src/accounts.rs
```diff
@@ -0,0 +1,1086 @@
...
+        + BORSH_LEN_PREFIX + self.data.len()             // data
+        + std::mem::size_of::<i64>() // expiry
+    }
+}
+
+impl ValidatableQuote for TransientQuote {
+    fn expiry(&self) -> i64 {
+        self.expiry
+    }
+}
```
paulbalaji, line 370: **High — `min_issued_at` emergency revocation does not apply to transient quotes.**

This `ValidatableQuote for TransientQuote` impl overrides `expiry()` but not `issued_at()`, so it inherits the trait default which returns `None` (see line 300). `validate_quote`'s `is_some_and(|ia| ia < min_issued_at)` then short-circuits to `false` and the `min_issued_at` check is silently skipped.

Attack path: a signer key leaks → admin calls `SetMinIssuedAt(now)` to revoke → the attacker can still `SubmitQuote` + `QuoteFee` a transient signed by the compromised key, because both the submit path (`:1042-1059`) and transient consumption (`:436`) allow it through.

Fix:
1. Store `issued_at` on `TransientQuote` (currently discarded — see `processor.rs:1238`).
2. Implement `issued_at()` on this trait impl to return `Some(self.issued_at)`.
3. Also gate `process_submit_quote` on `issued_at >= fee_account.min_issued_at` so stale quotes don't even get written.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+    accounts: &[AccountInfo],
+    min_issued_at: i64,
+) -> ProgramResult {
+    let accounts_iter = &mut accounts.iter();
+    let (fee_account_info, mut fee_account, _owner_info) =
+        fetch_fee_account_and_verify_owner(program_id, accounts_iter)?;
+
+    ensure_no_extraneous_accounts(accounts_iter)?;
+
+    fee_account.min_issued_at = min_issued_at;
```
paulbalaji, line 955: **High — `SetMinIssuedAt` is not monotonic.**

`fee_account.min_issued_at = min_issued_at;` unconditionally overwrites the threshold, so an owner (compromised or fat-fingering) can move it *backward* and un-revoke quotes they previously revoked. If the intent is an emergency revocation kill-switch, this should be one-way.

Suggested fix:
```rust
if min_issued_at < fee_account.min_issued_at {
    return Err(Error::MinIssuedAtMustBeMonotonic.into());
}
fee_account.min_issued_at = min_issued_at;
```

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+                return Err(Error::StaleStandingQuote.into());
+            }
+            // Equal issued_at → no-op (don't update, don't error).
+            if issued_at_ts == existing.issued_at {
+                msg!("Standing quote no-op (equal issued_at)");
+                return Ok(());
+            }
+        }
+
+        standing_pda.quotes.insert(recipient_key, new_value);
```
paulbalaji, line 1351: **High — unbounded standing-quote BTreeMap (DoS).**

`standing_pda.quotes.insert(recipient_key, new_value)` has no cap on map size, and `SubmitQuote` allows any payer holding a single valid signer signature to add entries for an unbounded set of recipient H256 values. The PDA grows via `store_with_rent_exempt_realloc` at the *payer's* expense, but it is owned by the program and reachable only by the fee-account owner via `PruneExpiredQuotes`.

Because this account is deserialized on the hot `QuoteFee` path, a griefing or compromised signer can inflate it until `QuoteFee` hits CU limits or pruning becomes too expensive to fit in one tx.

Same pattern applies to:
- `FeeAccount.standing_quote_domains` — also hot-path-deserialized; grows on each new-domain `SubmitQuote` (`:1403`).
- Signer sets: `LeafFeeConfig.signers`, `RouteDomain.signers`, CC `signers`, and the wildcard sets (`:739`, `:1011`).

Suggested fix: introduce `MAX_STANDING_RECIPIENTS_PER_PDA`, `MAX_STANDING_QUOTE_DOMAINS`, `MAX_SIGNERS_PER_SET` constants (e.g. 32 / 64) enforced at write time, and add stress tests.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+    }
+
+    // Validate quote hasn't expired.
+    let clock = Clock::get()?;
+    if clock.unix_timestamp > expiry_ts {
+        return Err(Error::QuoteExpired.into());
+    }
+
+    // Reject issued_at too far in the future (clock skew guard).
+    if issued_at_ts > clock.unix_timestamp + MAX_ISSUED_AT_SKEW {
```
paulbalaji, line 1035: **Medium — `SubmitQuote` does not check `min_issued_at`.**

This path validates expiry and future-skew only, then falls through to insert the quote into state. Quotes with `issued_at < fee_account.min_issued_at` are stored anyway and only fail later at `QuoteFee` via `ValidatableQuote::validate_quote` (line 519). Net effect for standing quotes: payer-funded rent stranded on dead-on-arrival entries, and attacker can pollute the BTreeMap with already-revoked entries (compounds with the unbounded-growth finding).

Suggested fix:
```rust
if issued_at_ts < fee_account.min_issued_at {
    return Err(Error::StaleQuote.into());
}
```
Applied after loading `fee_account`, before writing either transient or standing PDAs.

rust/sealevel/libraries/quote-verifier/src/lib.rs
```diff
@@ -0,0 +1,661 @@
...
+        &self,
+        fee_account: &Pubkey,
+        domain_id: u32,
+        scoped_salt: &H256,
+    ) -> H256 {
+        let context_hash = keccak::hash(&self.context);
+        let data_hash = keccak::hash(&self.data);
+
+        let hash = keccak::hashv(&[
+            DOMAIN_TAG,
```
paulbalaji, line 91: **Medium — signed hash has no version or kind byte.**

The input `[DOMAIN_TAG, fee_account, domain_id, context_hash, data_hash, issued_at, expiry, scoped_salt]` has no version byte and no kind discriminator (Leaf vs Routing vs CC; Transient vs Standing). Today these are indirectly separated because the fee_account PDA is mode-locked to one variant, and `expiry == issued_at` distinguishes transient from standing — but both are brittle and make schema evolution risky.

Suggested fix: prepend a `version: u8` and a `kind: u8` (fee-mode × transient/standing) into the hashv input so any future schema change or kind split cannot be cross-replayed against this encoding.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2154 @@
...
+
+    // Verify the quote signature against resolved signers.
+    quote
+        .verify_signer(
+            fee_account_info.key,
+            fee_account.domain_id,
+            payer_info.key,
+            &resolved_signers,
+        )
+        .map_err(|_| Error::InvalidQuoteSignature)?;
```
paulbalaji: **Medium — error-code collapse on signature verification.**

`.map_err(|_| Error::InvalidQuoteSignature)?` flattens `QuoteVerifyError::InvalidSignature` (malformed bytes / recovery failure) and `QuoteVerifyError::UnauthorizedSigner` (valid sig but signer not in the authorized set) into a single error code. These are very different operational events — the first is a client bug, the second is a potential key-compromise signal — and should be surfaced distinctly in logs/alerts.

Suggested fix: pattern-match the `QuoteVerifyError` variant and map to distinct `Error` codes.

paulbalaji (commented):
## Review Summary

Consolidated review of the new SVM fee-quoting program (`hyperlane-sealevel-fee` + `quote-verifier` + `package-versioned`). Accounted for existing discussion: only the Devin bot's "no issues found" comment exists; no inline threads or author replies.

Overall: architecture is coherent (five-cell signer matrix, quote cascade, CC DEFAULT_ROUTER provenance tracking, PDA-scoped salts, per-route signer isolation). The signature-verification library, curve math, PDA seed layout, owner-check plumbing, and the CC default-router invalidation flow are all in good shape. Functional coverage on happy paths and the authority matrix is strong.

The findings below concentrate on three themes: (1) signers cannot express the full scope of what they authorize (curve variant is not bound, emergency revocation is bypassable for transients), (2) state growth is uncapped on hot-path-deserialized accounts, and (3) `InitFee` has no caller authorization. None are blockers for a draft; all should be resolved before this ships.

### Findings by severity

**High**
- `InitFee` has no authorization — anyone can seize a salt + set attacker-chosen owner/beneficiary/fee_data. Inline at `processor.rs` L145.
- Signed quote does not commit to the curve variant (only `max_fee` + `half_amount`); an `UpdateFeeParams` that swaps Linear → Progressive retroactively changes the semantics of already-signed quotes. Inline at `quote-verifier/src/lib.rs` L95 and `processor.rs` L446.
- `min_issued_at` emergency revocation is a no-op for transient quotes (`TransientQuote` inherits the trait-default `issued_at() -> None`). Inline at `accounts.rs` L370.
- `SetMinIssuedAt` is not monotonic — owner can move the threshold backward and un-revoke. Inline at `processor.rs` L978.
- Unbounded `BTreeMap`/`BTreeSet` growth on accounts deserialized on hot paths (`FeeStandingQuotePda.quotes`, `FeeAccount.standing_quote_domains`, `LeafFeeConfig.signers`, `RouteDomain.signers`, CC-route `signers`). No `MAX_*` constants exist besides `MAX_ISSUED_AT_SKEW`. Inline at `processor.rs` L1374.

**Medium**
- `SubmitQuote` does not check `min_issued_at` at ingest — revoked-but-unexpired quotes get written and strand rent. Inline at `processor.rs` L1057.
- Signed message hash lacks a version + kind byte (Leaf/Routing/CC × Transient/Standing). Currently safe via fee_account PDA mode-locking, but brittle. Inline at `quote-verifier/src/lib.rs` L91.
- `.map_err(|_| Error::InvalidQuoteSignature)` collapses `InvalidSignature` and `UnauthorizedSigner` into one error. Inline at `processor.rs` L1211.
- Semantic split on unconfigured routes: Routing-mode unconfigured domain → fee = 0 (`processor.rs` ~L272); CC-mode unconfigured → `RouteNotFound` revert (`processor.rs` ~L601). Either document the divergence or unify.
- Error-code inconsistency: PDA-mismatch returns `TransientPdaMismatch` at `processor.rs` L431-434 and L1477-1479 but `InvalidArgument` everywhere else (per PR description).
- `SetWildcardQuoteSigners` replaces the whole set with no cap and no empty-set guard — a typo can silently disable wildcard quoting.

**Low**
- No low-s enforcement on secp256k1 signatures (Solana's `secp256k1_recover` does not reject high-s). Not exploitable today (scoped-salt + uninit-check dedupe) but trivial to add in `EcdsaSignature::from_bytes`.
- No validation on `FeeParams` on write: `half_amount = 0` silently returns `fee = 0`; `max_fee = 0` same. Owner-gated, but silent misconfig is a footgun. `process_init_fee` + `process_update_fee_params` + route creators.
- `RemoveQuoteSigner` silently no-ops when the target is not in the set — return `SignerNotFound` for auditability.
- Single-step `TransferOwnership` (no propose/accept); consider two-step given this governs fee revenue routing.
- Transient PDA is not auto-closed on `QuoteFee` error paths — survives the tx, must be cleaned via `CloseTransientQuote`. Document for relayers.
- `FeeParams::max_fee` docstring says "Maximum fee that can be charged" — but it is only a hard clamp for Linear; for Regressive/Progressive it is an asymptote.
- Progressive curve + integer-floor rounding: small amounts round to `fee = 0` by design (`test_progressive_small_amount`). Splitting a large transfer into many small ones pays no fee. Likely intentional; worth documenting.

### Test-coverage gaps

- No stress/cap test for the unbounded collections above — pair with the `MAX_*` fixes.
- No `proptest` / fuzz over `fee_math::compute_fee` (monotonicity, `≤ max_fee`, no overflow are ideal property targets).
- No functional-level cross-fee-account signature-replay test (only covered at the verifier unit level).
- No explicit test that transient PDAs don't leak when `QuoteFee` errors after the transient is loaded.
- No exact-boundary test for `now == expiry` on standing quotes.
- No test for `GetProgramVersion` discriminator-collision path (appears safe given borsh variant indexing, but worth a targeted test).

### Verified clear

- PDA seed design — five distinct prefixes (`fee` / `route` / `cc_route` / `transient` / `standing`), length-delimited, `DEFAULT_ROUTER = [0xFF;32]` vs `H256::zero()` sentinel; no collisions.
- Owner / discriminator / signer checks on mutating handlers are consistently present.
- CC DEFAULT_ROUTER fallback invalidation (`auth_scope: Direct` vs `CcDefaultFallback`) is correctly enforced and well-tested (`test_cc_default_authorized_standing_quote_invalidated_by_later_specific_route`).
- Quote cascade precedence (transient → domain standing → wildcard-domain standing → onchain) is well-tested.
- Signer authority matrix — all five cells (Leaf, Routing-exact, Routing-wildcard, CC-exact, CC-wildcard) tested with both correct and wrong signer.
- Fee-math overflow: U256 intermediates fit comfortably; no reachable panic; `try_into::<u64>` → `FeeComputationOverflow` is dead defensive code.
- Cross-Hyperlane-domain replay: `fee_account` pubkey + `domain_id` in the signed hash prevent this.
- Transient payer binding via `scoped_salt = keccak(payer || client_salt)` — can't be hijacked.

### Recommended fix order

1. Add authorization to `InitFee` (deployer signer or upgrade authority).
2. Bind curve variant into signed quotes (either in the signed hash or by storing+enforcing the variant at consumption).
3. Fix transient `min_issued_at` bypass + gate `SubmitQuote` on `min_issued_at`.
4. Enforce `SetMinIssuedAt` monotonicity.
5. Introduce `MAX_*` caps on signer sets, standing-quote recipients, and `standing_quote_domains` + add stress tests.
6. Add version + kind byte to the signed hash.

paulbalaji (commented):
## Review Summary

Review of the SVM fee program. Building on @paulbalaji's 9 prior findings (init owner squat, signed-hash missing curve variant + companion clone, transient `min_issued_at` bypass, non-monotonic `SetMinIssuedAt`, unbounded standing-quote map, submit-side `min_issued_at` skip, missing version/kind in signed hash, error-code collapse) — all still load-bearing. Inline comments below cover net-new findings; lower-severity items are bundled at the end.

### Quote-state lifetime gaps (High)

- Standing quotes outlive the routes that authorized them. `RemoveRoute` / `RemoveCrossCollateralRoute` close the route PDA but the standing-quote PDA at `(fee_account, domain[, target_router])` persists with its entries. For CC, fallback resolution then routes the lingering `auth_scope=Direct` quote through DEFAULT_ROUTER's curve params at QuoteFee — the cascade has no membership check against the current route's signer set.
- Same hazard from `SetRoute` / `SetCrossCollateralRoute` rotation: an old signer's `Direct` quote keeps applying after the route's signer set has been overwritten. There is no instruction to bulk-invalidate (only chain-wide `SetMinIssuedAt`).

Compounds with paulbalaji's curve-variant-not-signed finding: stale signature → different curve.

### State-bloat / DoS-adjacent (Medium)

- `PruneExpiredQuotes` retain predicate ignores `fee_account.min_issued_at`, so emergency-revoked entries persist (rent-locked, deserialized on every QuoteFee) until natural expiry.
- `FeeAccount.standing_quote_domains` is unbounded and mutated on the hot path — an authorized signer (esp. wildcard) can insert across all u32 domain IDs.
- Standing-quote rent asymmetry: submitter pays grow via `store_with_rent_exempt_realloc(_, payer_info, _)`, owner gets the close refund.

### Lower-severity / nits / questions

Non-blocking, not raised inline to keep this review focused:
- Transient quote re-creatable within issuance second (expiry check is `>` not `>=`).
- No `program_id` / cluster identifier in the signed message (defense-in-depth).
- ECDSA accepts both low-S and high-S; not exploitable here (PDA derivations don't key on signature bytes), but EVM-style hardening is cheap.
- `InitFee` initializes `min_issued_at = 0` instead of `Clock::get()?.unix_timestamp` — fresh deploys accept quotes signed against epoch.
- Inconsistent signer-init: Leaf rejects `cfg.signers == None`, Routing/CC silently auto-init.
- `process_set_route` rejects `domain == 0` without documented justification.
- `WILDCARD_RECIPIENT = H256::repeat_byte(0xFF)` collides with all-ones recipient address.
- `standing_quote_domains` is dead field on CC accounts — consider moving into the Leaf/Routing variants.
- `fetch_fee_account_and_verify_owner` skips PDA re-derivation; `FeeAccountPrefix::parse_from` skips discriminator validation.
- `is_transient_quote` returns binary yes/no but caller can't distinguish "wrong account" from "no transient".
- `fee_math.rs` U256 multiplications rely on implicit headroom; consider `checked_*` or an inline overflow argument.
- `PACKAGE_VERSION = "1.0.0"` hardcoded; prefer `env!("CARGO_PKG_VERSION")`.
- `GetQuoteAccountMetas` simulation contract for transient-PDA pre-existence undocumented (SDK consumers will hit confusing `InvalidArgument`).

### Test coverage gaps

- Several leaf-curve `quote_fee` tests assert only `Ok(_)`, never the computed fee value (inline comment below).
- No regression test for paulbalaji #1 (init owner squat) or #6 (unbounded growth).
- `SetRoute` / `SetCrossCollateralRoute` always called with `signers: None`; the `Some(...)` initial path is unexercised (including "second `SetRoute` overwrite" semantics).
- `RemoveQuoteSigner` only tested for Leaf — Routing/CC mismatched-mode rejections (`NotRoutingFeeData`, `NotCrossCollateralRoutingFeeData`) and spoofed-route-PDA paths untested.
- `AddQuoteSigner` cross-mode/spoofed-PDA rejections untested.
- `StandingQuoteAuthScope::CcDefaultFallback` lacks a Borsh discriminant round-trip test; only one direction of invalidation flow is tested in `cc.rs`.
- Curve `FeeComputationOverflow` path untested.
- Spoofed-PDA tests collapse on generic `InvalidArgument` (could pass for the wrong reason).
- Hardcoded fee values rather than computed via `compute_fee` won't catch curve-formula drift.
- `test_routed_wildcard_transient_rejected` may fire on the wrong check after a future refactor (which check fires depends on signer-resolution ordering).
- Significant helper duplication between `standing.rs` / `cc.rs` / `transient.rs` — pull common helpers into `helpers.rs`.

Do not approve, do not request changes — leaving as comment for triage.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+        return Err(Error::NotCrossCollateralRoutingFeeData.into());
+    }
+
+    let dest_le = data.destination.to_le_bytes();
+    let (expected_key, _) = Pubkey::find_program_address(
+        cc_route_pda_seeds!(fee_account_info.key, &dest_le, data.target_router),
+        program_id,
+    );
+
+    remove_route_pda(program_id, accounts_iter, expected_key, owner_info)?;
```
paulbalaji: **[High]** Closing the route PDA leaves the standing-quote PDA at `(fee_account, dest, target_router)` intact. For CC mode, after a specific route is removed, `resolve_cc_routing` falls back to DEFAULT_ROUTER returning `cc_specific_route_active = false` — so the skip at `try_resolve_standing_quote` (`auth_scope=CcDefaultFallback` && `cc_specific_route_active`) never fires, and a previously-authorized `auth_scope=Direct` quote signed by the now-removed route's signers continues to apply through the *default* router's curve params at QuoteFee.

Same hazard from `SetCrossCollateralRoute` rotation (line 1911): overwrites signers on the route, but pre-existing `Direct` standing quotes signed by the old set are still consumed.

Compounded with the existing finding that the signed hash doesn't bind the curve variant, this lets stale signatures execute under whichever curve is active at consumption time. There is no instruction to bulk-invalidate per-route (only chain-wide `SetMinIssuedAt`).

Options: (a) close the affected standing-quote PDA(s) atomically with route removal/rotation; (b) store the recovered signer address on `FeeStandingQuoteValue` and re-check membership at QuoteFee time; (c) at minimum, document the operator-side `SetMinIssuedAt` requirement after every route rotation.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+    for recipient_key in [quote_fee_data.recipient, WILDCARD_RECIPIENT] {
+        if let Some(value) = standing.quotes.get(&recipient_key) {
+            // A CC exact-domain standing quote may have been authorized earlier via the
+            // DEFAULT_ROUTER fallback because no router-specific route existed yet. Once a
+            // specific route exists, that quote must no longer apply to the specific route's
+            // trust domain, even if the standing quote PDA key still matches.
+            if cc_specific_route_active
+                && value.auth_scope == StandingQuoteAuthScope::CcDefaultFallback
+            {
+                continue;
```
paulbalaji, line 500: **[High — companion to the route-removal/rotation finding above]** This is the only post-hoc invalidation lever for stale CC standing quotes: the `cc_specific_route_active && CcDefaultFallback` skip. After a specific route is removed (or its signers rotated), `cc_specific_route_active` flips to `false` and an `auth_scope=Direct` quote signed by an obsolete signer set continues to be consumed.

If storage of the actual recovered signer address on `FeeStandingQuoteValue` is too expensive, an alternative is to bump `fee_account.min_issued_at` automatically inside `RemoveCrossCollateralRoute` / `SetCrossCollateralRoute` when signers change — but that's chain-wide. A targeted PDA-close on route teardown is the cleaner fix.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+    ensure_no_extraneous_accounts(accounts_iter)?;
+
+    let mut standing = FeeStandingQuotePdaAccount::fetch(&mut &domain_pda_info.data.borrow()[..])?
+        .into_inner()
+        .data;
+
+    let clock = Clock::get()?;
+    standing
+        .quotes
+        .retain(|_, value| clock.unix_timestamp <= value.expiry);
```
paulbalaji: **[Medium]** Retain only filters by expiry. After an emergency `SetMinIssuedAt(now)`, the BTreeMap fills with dead-on-arrival entries that won't be evicted until natural expiry — yet they're iterated/deserialized on every `QuoteFee` (compounds with the unbounded-map finding from prior review).

Suggestion: `standing.quotes.retain(|_, v| clock.unix_timestamp <= v.expiry && v.issued_at >= fee_account.min_issued_at);`

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+            // Update standing_quote_domains on fee account (non-CC only).
+            // CC accounts have multiple PDAs per domain (one per target_router),
+            // making domain tracking unreliable — offchain tooling handles CC domain discovery.
+            if !matches!(
+                fee_account.data.fee_data,
+                FeeData::CrossCollateralRouting(_)
+            ) {
+                let mut fee_account_mut = fee_account.data;
+                fee_account_mut
+                    .standing_quote_domains
```
paulbalaji: **[Medium]** `FeeAccount.standing_quote_domains` is unbounded and grown on the SubmitQuote hot path. An authorized signer (especially a wildcard signer) can submit standing quotes across all 2³² domain IDs, bloating the FeeAccount that gets deserialized on every `QuoteFee` (line 218 in this file). Distinct from prior reviewer's per-PDA recipient-map finding — this is the fee-account-level set.

Also note: rent for these grows is paid by `payer_info` (the current submitter), but `process_prune_expired_quotes` line 1534 sends the close refund to `owner_info`. Submitters fund growth, owner harvests refund.

Suggestions: cap `standing_quote_domains.len()` (e.g. 256), or move discovery off-chain entirely and drop the on-chain set.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+                FeeData::Leaf(cfg) => {
+                    if let Some(ref mut signers) = cfg.signers {
+                        signers.remove(&signer);
+                    }
+                }
+                _ => {
+                    return Err(Error::NotLeafFeeData.into());
+                }
+            }
+            FeeAccountData::new(fee_account.into()).store(fee_account_info, false)?;
```
paulbalaji, line 860: **[Medium]** Leaf-mode signer removal calls `store(_, false)` (no realloc). Combined with `store_with_rent_exempt_realloc`'s `required_size.max(account_data_len)` (which never shrinks), the FeeAccount size grows monotonically as signers are added/removed and never reclaims rent. Especially noticeable since `AddQuoteSigner` actively grows the account.

Either document monotonic-only growth as the design contract, or thread a true `realloc` here that allows shrinking.

While here: process_remove_quote_signer is also untested for the Routing and CC paths — `build_remove_quote_signer_ix` in tests hardcodes `route: None`. The `NotRoutingFeeData` / `NotCrossCollateralRoutingFeeData` rejections and spoofed-route-PDA paths around line 875+ are unexercised.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
```diff
@@ -0,0 +1,2117 @@
...
+        return Err(ProgramError::IncorrectProgramId);
+    }
+
+    let (fee_account_info, fee_account, owner_info) =
+        fetch_fee_account_and_verify_owner(program_id, accounts_iter)?;
+
+    if !matches!(fee_account.fee_data, FeeData::CrossCollateralRouting(_)) {
+        return Err(Error::NotCrossCollateralRoutingFeeData.into());
+    }
+
```
paulbalaji, line 1913: **[Medium]** Validation rejects `data.destination ∈ {0, WILDCARD_DOMAIN}` but does not reject `data.target_router == H256::zero()` (the non-CC sentinel) or `data.target_router == DEFAULT_ROUTER`. The submit-side `process_submit_quote` already rejects both values for CC standing quotes (search `ZeroTargetRouterNotAllowed`). The route can therefore be created at a sentinel target_router producing dead state — no standing quote can ever populate the corresponding PDA — and is a footgun if any future code path trusts the sentinel discrimination.

Mirror the submit-time invariant here: reject `target_router == H256::zero() || target_router == DEFAULT_ROUTER` in `process_set_cc_route`.

While here: `process_quote_fee` resolution (around line 574) on `data.target_router == DEFAULT_ROUTER` short-circuits the first iteration of `[(target_router, true), (DEFAULT_ROUTER, false)]` with `cc_specific_route_active = true` — semantically the caller asked for the *default* route. Either reject CC `QuoteFee` calls where `data.target_router == DEFAULT_ROUTER`, or skip the duplicate iteration.

rust/sealevel/programs/hyperlane-sealevel-fee/tests/functional.rs
```diff
@@ -0,0 +1,2581 @@
...
+
+        let ix = build_quote_fee_leaf_ix(
+            &fee_key,
+            &payer.pubkey(),
+            42,
+            H256::zero(),
+            500, // at half_amount → fee = 500
+        );
+        process_tx(&mut banks_client, &payer, ix, &[])
+            .await
```
paulbalaji: **[Test gap]** `test_leaf_linear`, `test_leaf_regressive`, `test_leaf_progressive` use `process_tx(...).await.unwrap()` instead of `simulate_quote_fee` — they confirm "instruction returned Ok" but never assert the curve produced the expected value. A regression that returned `0` for all amounts (or skipped the curve switch) would pass.

Replace with `simulate_quote_fee` and `assert_eq!(fee, expected)` at half-amount and at saturation (boundary cases for each curve).

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Won't fix. None of the existing SVM programs (mailbox, token, IGP) gate init behind deployer authority — this is consistent with the SVM pattern. The PDA is derived from a user-chosen salt, so collisions require knowing the salt. Front-running is a theoretical concern but in practice the deployer submits init as part of a deployment script and the salt is not publicly broadcast beforehand.

rust/sealevel/libraries/quote-verifier/src/lib.rs
xeno097: Fixed in c981650. Quote data now Borsh-encodes the full FeeDataStrategy (1-byte variant tag + params) instead of bare max_fee/half_amount. The variant tag is committed into the signature via keccak256(data). At consumption time, same_variant() rejects quotes where the signed curve doesn't match the on-chain strategy, returning CurveVariantMismatch. This also future-proofs for new curve types with different param layouts — the TryFrom<&[u8]> impl on FeeDataStrategy handles deserialization.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Fixed in the same commit as the parent finding. try_consume_transient_quote and try_resolve_standing_quote no longer clone the on-chain strategy and override params. Instead they deserialize the quoted FeeDataStrategy directly from the data bytes, verify the variant matches on-chain via same_variant(), and compute the fee using the quoted strategy as-is. The clone+params_mut() override pattern is gone.

rust/sealevel/programs/hyperlane-sealevel-fee/src/accounts.rs
xeno097: Already fixed in the current code. TransientQuote implements issued_at() returning self.expiry, which equals issued_at by construction for transient quotes (enforced by expiry == issued_at invariant at submit time). The validate_quote trait method checks issued_at() < min_issued_at and rejects. Additionally, SubmitQuote itself now gates on min_issued_at at ingest time (L1059-1063), so dead-on-arrival transient quotes are rejected before the PDA is even created.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Already fixed. Monotonicity is enforced at L949-951 with MinIssuedAtMustBeMonotonic error — the threshold can only increase.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Partially addressed. The standing_quote_domains field — the main unbounded growth vector on the hot-path-deserialized FeeAccount — was removed entirely in 24e0a1d. Offchain discovery replaces it for both routing and CC modes. Signer sets (BTreeSet<H160>) remain uncapped but are owner-controlled and practically bounded by CU/heap limits. We'll add operator guidance from benchmarking rather than hard on-chain caps, consistent with our approach across other SVM programs.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Already fixed at L1059-1063. Quotes with issued_at < fee_account.min_issued_at are now rejected with QuoteBelowMinIssuedAt before writing any PDA, preventing rent from being stranded on dead-on-arrival entries.

rust/sealevel/libraries/quote-verifier/src/lib.rs
xeno097: Won't fix for now. We verified the EVM implementation — OffchainQuotedLinearFee also does not include a version or kind byte in the signed payload. The EIP-712 domain separator has a hardcoded version "1" but that's domain-level, not quote-level. The curve variant is now committed via the data field (see parent finding fix), which addresses the most critical aspect of payload ambiguity.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Already fixed. The .map_err now pattern-matches on the QuoteVerifyError variant: InvalidSignature maps to InvalidQuoteSignature (error 7) and UnauthorizedSigner maps to UnauthorizedQuoteSigner (error 25). On-chain monitoring can now distinguish between cryptographic failures and key-compromise signals.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Fixed in 96d20c7. Route instructions were unified into SetRemoteFeeRoute / RemoveRemoteFeeRoute. SetRemoteFeeRoute atomically resets the standing quote PDA for the affected (domain, target_router) to an empty state, so stale authorized quotes can't survive route changes. RemoveRemoteFeeRoute closes the standing PDA entirely. The upsert_route_pda helper is reused for the standing PDA reset to avoid code duplication.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Addressed by the same route unification commit. Since standing quotes are now cleared atomically on any route set/remove, the window for stale CcDefaultFallback-scoped quotes to linger is eliminated. The auth_scope check remains as defense-in-depth for quotes submitted between route changes within a single slot.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Fixed in 3866b2b. The retain predicate now checks both conditions: clock.unix_timestamp <= value.expiry && value.issued_at >= fee_account.min_issued_at. After an emergency SetMinIssuedAt(now), a single prune call removes all revoked entries regardless of their natural expiry.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Fixed in 24e0a1d. The field was removed from FeeAccount entirely. Both routing and CC modes now rely on offchain PDA discovery. This eliminates the unbounded growth vector, the hot-path realloc during SubmitQuote, and the write-lock contention on the fee account during standing quote submission. Fee account is now always read-only for SubmitQuote.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Won't fix. This is consistent with other SVM programs in the repo that don't shrink accounts on data removal — the monotonic-growth pattern avoids realloc complexity and CU overhead on the remove path. Account size is practically bounded by signer count which is owner-controlled.

rust/sealevel/programs/hyperlane-sealevel-fee/src/processor.rs
xeno097: Partially fixed in 96d20c7. SetRemoteFeeRoute now rejects H256::zero() as target_router for CC mode via ZeroTargetRouterNotAllowed. DEFAULT_ROUTER remains a valid target_router for CC routes because it's the designated fallback — the resolve_cc_routing cascade depends on being able to set a route at the DEFAULT_ROUTER key.

rust/sealevel/programs/hyperlane-sealevel-fee/tests/functional.rs
xeno097: Fixed in 62f8c3e. test_leaf_linear, test_leaf_regressive, and test_leaf_progressive now use simulate_quote_fee and assert the computed fee matches the expected curve output (500, 666, 800 respectively).

xeno097:
superseded by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8698

PR #8647, opened 2026-04-21
feat(sdks): sealevel fee support svm sdk [IGNORE FOR NOW]

xeno097:
### Description

  Implements Phase 2 of the SVM fee system: full SDK support for deploying and managing fee programs on Solana/SVM chains via the `IRawFeeArtifactManager` interface.

  **Requires [#8642](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8642) (SVM fee program implementation) to be merged first.** The fee program binary is temporarily injected via a droppable commit (`0ae46733a6`) that should be
  removed once the program lands on main.

  #### What's included:

  - **Readers and writers for all 6 fee types:** linear, regressive, progressive, offchainQuotedLinear, routing, crossCollateralRouting. Each type has a dedicated reader/writer following the existing artifact pattern. Leaf types
  (linear/regressive/progressive) share an abstract base to avoid duplication.
  - **Instruction builders for all 18 fee program instructions:** deployment/management instructions (InitFee, SetRoute, SetCrossCollateralRoute, UpdateFeeParams, SetBeneficiary, TransferOwnership, AddQuoteSigner, RemoveQuoteSigner,
  SetWildcardQuoteSigners, SetMinIssuedAt, RemoveRoute, RemoveCrossCollateralRoute) and runtime instructions (QuoteFee, SubmitQuote transient/standing, CloseTransientQuote, PruneExpiredQuotes, GetQuoteAccountMetas,
  GetSubmitQuoteAccountMetas).
  - **PDA derivation functions** for FeeAccount, RouteDomain, CrossCollateralRoute, TransientQuote, and StandingQuote PDAs.
  - **Borsh codecs and account decoders** for all fee program account types (FeeAccount, RouteDomain, CrossCollateralRoute, TransientQuote, StandingQuotePda).
  - **SvmFeeArtifactManager** implementing `IRawFeeArtifactManager`, wired into `SvmProtocolProvider.createFeeArtifactManager()`.
  - **Shared utilities:** fee strategy mapping between provider-sdk and on-chain formats, wildcard signer computation, H160 signer conversion with validation, canonical BTreeSet encoding.
  - **Stress tests** that discovered on-chain limits via binary search:
    - InitFee max signers: **41** (tx size)
    - SetRoute max signers: **42** (tx size)
    - SetWildcardQuoteSigners max signers: **47** (tx size)
    - Incremental AddQuoteSigner: **409** (CU/heap)
    - Standing quotes per domain PDA: **113** (BTreeMap heap, Leaf and CC identical)

  ### Drive-by changes

  - **Breaking:** `ProtocolProvider.createFeeArtifactManager()` now requires a `FeeReadContext` parameter to ensure routed fee types receive domain/router context for non-enumerable route PDA discovery. All protocol SDK implementations
  (tron, radix, cosmos, aleo, starknet) updated to accept the new parameter.
  - Added `ByteCursor.readI64LE()` and `i64le()` encoder for signed 64-bit integer support.
  - Added `@noble/curves` as a devDependency in svm-sdk for secp256k1 quote signing in stress tests.

  ### Related issues

  - Depends on #8642 (SVM fee program)
  - Follows #8627 (multi-VM fee type system in provider-sdk/deploy-sdk)

  ### Backward compatibility

  No — `ProtocolProvider.createFeeArtifactManager()` has a new required `FeeReadContext` parameter. This is a major version bump for `@hyperlane-xyz/provider-sdk` and `@hyperlane-xyz/deploy-sdk`. All downstream protocol SDK consumers need
  to pass the context. The deploy-sdk factories (`createFeeReader`, `createFeeWriter`) already had context and now thread it through.

  ### Testing

  **E2E tests (32 tests across 6 suites, all in CI):**
  - fee-linear: 4 tests (create/read, no-op update, param update, beneficiary update)
  - fee-regressive: 4 tests (shared leaf suite)
  - fee-progressive: 4 tests (shared leaf suite)
  - fee-offchain-quoted-linear: 8 tests (shared leaf suite + signer create/add/remove + empty signer roundtrip)
  - fee-routing: 6 tests (create/read, add route, update params, change strategy type, remove route, offchainQuotedLinear route with signers)
  - fee-cross-collateral-routing: 5 tests (create/read, add pair, update params, change strategy, remove pair)

---

<a href="https://app.devin.ai/review/hyperlane-xyz/hyperlane-monorepo/pull/8647" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


typescript/svm-sdk/src/fee/leaf-fee.ts
```diff
@@ -0,0 +1,245 @@
...
+      // property narrows the discriminant. OffchainQuotedLinearFeeConfig (which
+      // has an extra quoteSigners field) has its own dedicated reader and never
+      // flows through this base class read path.
+      config: {
+        type: this.feeType,
+        owner,
+        beneficiary,
+        maxFee: maxFee.toString(),
+        halfAmount: halfAmount.toString(),
+      } as C,
```
paulbalaji, line 95: **[High — confirms Devin's flag]** `as C` cast hides a type-design bug. The `LeafFeeConfig` union (per `provider-sdk`) includes `OffchainQuotedLinearFeeConfig`, which has a required `quoteSigners: string[]` field that this base class never populates. The CAST: comment claims the offchain-quoted config "never flows through this base class read path" but nothing at the type level enforces that — `SvmOffchainQuotedLinearFeeReader` is a separate class only by convention.

If a future reader is parameterized as `SvmLeafFeeReader<OffchainQuotedLinearFeeConfig>`, the cast silently produces a config missing `quoteSigners`.

Fix: tighten the type parameter constraint to `LinearFeeConfig | RegressiveFeeConfig | ProgressiveFeeConfig` (drop offchain-quoted from the leaf union here) so the cast can be removed entirely. CLAUDE.md's zero-tolerance-for-`as`-casts rule applies.

typescript/svm-sdk/src/fee/fee-strategy-utils.ts
```diff
@@ -0,0 +1,104 @@
...
+    maxFee: maxFee.toString(),
+    halfAmount: halfAmount.toString(),
+  };
+
+  if (!isNullish(route.signers)) {
+    return {
+      type: FeeStrategyType.offchainQuotedLinear,
+      ...base,
+      quoteSigners: route.signers.map(h160ToSigner),
+    };
```
paulbalaji, line 32: **[High]** When `route.signers !== null`, this returns `FeeStrategyType.offchainQuotedLinear` unconditionally — losing the actual `route.feeData.kind` (Linear / Regressive / Progressive) on the read path. On round-trip, `feeStrategyToOnChain` then writes `FeeStrategyKind.Linear` regardless of what was originally on-chain.

A Routing or CC config with a Regressive curve plus signers will be misread as Linear. The on-chain writes from this PR can't currently produce that state (TS only writes Linear when signers are present), so this is latent — but anyone who manually authors a Regressive route with signers via the Rust client (or a future TS path) will see wrong reads.

Fix: extend `provider-sdk` to model `offchainQuoted{Regressive,Progressive}` (or carry the kind separately), or add an `assert(route.feeData.kind === FeeStrategyKind.Linear, ...)` here so silent mislabeling fails loudly until the type is broadened.

typescript/svm-sdk/src/codecs/fee.ts
```diff
@@ -0,0 +1,152 @@
...
+        u8(data.kind),
+        encodeCrossCollateralRoutingFeeConfig(data.config),
+      );
+  }
+}
+
+// ── RouteKey enum (for Add/RemoveQuoteSigner) ───────────────────────
+
+export type SvmRouteKey =
+  | { kind: 'domain'; domain: number }
```
paulbalaji, line 128: **[Medium]** `SvmRouteKey` discriminant uses string literals (`kind: 'domain'` / `kind: 'crossCollateral'`); the encoder switch (line 135) uses `case 'domain':` etc. CLAUDE.md mandates enum members for discriminants. Other parts of this PR use proper enum-style objects (`FeeDataKind`, `FeeStrategyKind`, `FeeInstructionKind`).

Fix: add `export const SvmRouteKeyKind = { Domain: 'domain', CrossCollateral: 'crossCollateral' } as const;` and use members in the union and in the encoder switch.

typescript/svm-sdk/src/clients/protocol.ts
```diff
@@ -92,9 +96,11 @@ export class SvmProtocolProvider implements ProtocolProvider {
   }
 
   createFeeArtifactManager(
-    _chainMetadata: ChainMetadataForAltVM,
+    chainMetadata: ChainMetadataForAltVM,
+    context: FeeReadContext,
   ): IRawFeeArtifactManager | null {
-    return null;
+    const rpc = createRpc(this.getRpcUrls(chainMetadata)[0]);
```
paulbalaji, line 102: **[Medium]** `this.getRpcUrls(chainMetadata)[0]` produces `string | undefined`. If a chain metadata is misconfigured with zero RPC URLs, `createRpc(undefined)` will throw deep inside `@solana/kit` rather than at the SDK boundary.

CLAUDE.md prefers `assert()` at boundaries:

```ts
const url = this.getRpcUrls(chainMetadata)[0];
assert(url, `No RPC URL configured for chain ${chainMetadata.name}`);
const rpc = createRpc(url);
```

If the same pattern exists in nearby artifact-manager factories, propagate. (Pre-existing pattern? — fix here, document above, or open a follow-up.)

typescript/svm-sdk/scripts/generate-program-bytes.mjs
```diff
@@ -24,22 +25,29 @@ const PROGRAMS = {
...
   } catch {
-    console.error(`  ❌ ${key}: required .so file not found at ${path}`);
-    process.exit(1);
+    try {
+      bytes = readFileSync(fallbackPath);
+    } catch {
+      console.error(`  ❌ ${key}: .so not found at ${primaryPath} or ${fallbackPath}`);
+      process.exit(1);
+    }
   }
```
paulbalaji, line 48: **[Low / Devin still open]** Fee-program binary fallback to `so-binaries/` is transitional per the PR description (depends on #8642 landing on main with a built `.so`). Add a `# TODO(post-PR-8642): drop fallback once cargo build produces hyperlane_sealevel_fee.so` comment near `LOCAL_SO_DIR` so the droppable commit is paired with this hunk and not silently forgotten.

paulbalaji (commented):
## Review Summary

Review of Phase 2 SVM fee SDK support. **Codec / PDA / instruction-account audit comes back clean**: every Borsh layout, PDA seed, account list (writable / signer flags, ordering), enum discriminant, and account-discriminator string was diffed against the Rust source. All 18 fee-program instructions match. A SDK→Rust hash construction parity check isn't possible yet because the TS path still consumes signed-quote bytes opaquely (no signing path in TS).

### Existing review accounting

- **Devin's `as FeeStrategyKind` finding (`accounts/fee.ts:163`) is stale.** Current code uses an `is FeeStrategyKind` user-defined type guard (`isFeeStrategyKind`) and only narrows after the runtime check; there is no raw byte cast. Safe to close.
- **Devin's `router!` non-null finding (`fee/cross-collateral-routing-fee.ts:302-305`) is stale.** Current code uses `assert(router, ...)` (which has `asserts predicate` semantics in `@hyperlane-xyz/utils`); no `!` non-null assertions remain on those lines.
- **Devin's unsafe `as C` finding (`fee/leaf-fee.ts:91`) is current.** See inline.
- **Devin's fee-program binary fallback in `generate-program-bytes.mjs` is current.** Transitional per author description; add a `# TODO(post-PR-8642)` marker and ensure the droppable commit drops both the fallback and the embedded binary in lockstep.

### High-leverage net-new findings

- **`as C` cast hides a type-design bug** — `LeafFeeConfig` union includes `OffchainQuotedLinearFeeConfig` (which has a required `quoteSigners: string[]` field that the base reader never populates). Inline below.
- **`routeDataToFeeStrategy` collapses signed Regressive / Progressive routes to Linear on read.** Latent today; bad for Rust-authored / future TS routes. Inline below.
- **Quote-cascade runtime path not exercised end-to-end in CI.** PR description claims all 18 instructions tested, but `SubmitTransientQuote`, `CloseTransientQuote`, `PruneExpiredQuotes`, `Get*QuoteAccountMetas`, `QuoteFee`, `SetMinIssuedAt`, and standing-quote submission outside the stress test are touched only in the stress test or not at all.
- **No on-chain math round-trip vs. off-chain calc.** Tests round-trip Borsh only; never invoke `QuoteFee` and compare to a TS reference impl. A bug in the curve math or `FeeStrategyKind` discriminant ordering would not be caught.
- **No negative tests** — bad signer, expired quote, mismatched mode, replay (post-`SetMinIssuedAt`), missing accounts. The exact paths the cascade exists to handle.
- **`TransferOwnership` instruction never tested e2e.** All 4 leaf writers wire it; no test exercises owner change.

### Medium

- `SvmRouteKey` discriminant uses string literals (`'domain' | 'crossCollateral'`) instead of an enum — CLAUDE.md mandates enum members. Inline below.
- `getRpcUrls(chainMetadata)[0]` may be `undefined` — assert at boundary. Inline below.
- `skipPreflight: true` masks failures in stress tests (`fee-stress.e2e-test.ts`). Confirm `SvmSigner.send` polls confirmation; otherwise drop.
- Routing tests use shared signer / 100-SOL airdrop on the edge of lamport exhaustion across multiple `writer.create()` calls per test (each ~2 SOL+).
- Stress test not in CI matrix — discovered limits documented as comment, never asserted. Silent regressions invisible.
- Hardcoded `expect(updateTxs).to.have.length(2|3)` couples tests to implementation.
- `IRawFeeArtifactManager.readFee(address, context)` accepts a `context` parameter while the manager already holds one in its constructor — pick one to reduce footgun risk.

### Low / nits / questions

- All non-SVM `createFeeArtifactManager` stubs return `null` while Tron's other unsupported managers throw — inconsistent stubbing pattern (`null` says "protocol doesn't support"; throw says "TODO").
- `i64le` (`codecs/binary.ts:162`) doesn't validate bigint range — out-of-range silently wraps via `U64_CODEC.encode`.
- `compareBytes` `?? 0` fallback silently differs from Rust `Ord` if non-uniform-length entries pass through.
- 15-min CI timeout may be tight for `fee-routing` + `fee-cross-collateral-routing` matrix entries on a cold Solana validator.
- No coverage of "context omits a known route" path for routing readers.
- Changeset bullet list includes implementation detail (Borsh codecs, account decoders) rather than user-facing impact — minor, per CLAUDE.md style guide.

Do not approve, do not request changes — leaving as comment for triage.

xeno097:
Superseded by https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/8714

PR #8655, opened 2026-04-22
feat(core): batch set/remove on DomainRoutingIsm

paulbalaji:
## Summary
Adds `setBatch(DomainModule[])` and `removeBatch(uint32[])` to `DomainRoutingIsm` so routing ISM owners can enroll or unenroll many domains in a single tx after initialization. Mirrors the `setHooks(HookConfig[])` pattern on `DomainRoutingHook`.

## Why
Today the only way to enroll many domains in a single tx is the `initialize(address,uint32[],address[])` initializer, which can only be called once. Post-init, `HyperlaneIsmFactory` loops `set()` one domain at a time (see #8583, which had to shard the initializer itself to stay under block gas limits on citrea / shibarium / etc.). A follow-on batch setter lets the SDK keep shard sizes tight without eating N txs per deploy.

## Applies to all routing ISMs (via inheritance)
- `DomainRoutingIsm.setBatch(...)` — upsert-many (standard behavior).
- `IncrementalDomainRoutingIsm.setBatch(...)` — **batch-add-many**. The `_set` override at `IncrementalDomainRoutingIsm.sol:22` rejects any domain that already exists, so batched calls on incremental ISMs are effectively batch-adds (any duplicate in the batch reverts the whole tx).
- `DefaultFallbackRoutingIsm.setBatch(...)` — inherits upsert-many unchanged.
- `removeBatch` reverts on incremental (covered by a test), matching its single-domain `remove`.

## Design (minimal)
Trimmed from the fuller [#5219](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/5219). Keeps the surface deliberately small:
- `setBatch(DomainModule[])` — upsert (or batch-add on incremental, via inheritance)
- `removeBatch(uint32[])` — remove many
- `DomainModule` struct (mirrors hook's `HookConfig`)

Dropped from #5219: the separate `add` / `addBatch` *function selectors* (motivated by role-based ACL on an AccessManager — orthogonal to the batching behavior itself; can be added later if that use case lands), and `ModuleSet` / `ModuleRemoved` events (`DomainRoutingHook` doesn't emit equivalents — can add symmetrically on both sides later if needed).

## Test plan
- [x] `forge test --match-contract "DomainRoutingIsmTest|DefaultFallbackRoutingIsmTest|IncrementalDomainRoutingIsmTest"` — 41 passed across all three suites
- [ ] SDK wiring lives in follow-up PR (stacked): #8656


solidity/contracts/isms/routing/DomainRoutingIsm.sol
```diff
@@ -74,6 +80,18 @@ contract DomainRoutingIsm is
         _set(_domain, address(_module));
     }
 
+    /**
+     * @notice Sets the ISMs to be used for the specified origin domains
+     * @param _domainModules The origin domains and ISMs to enroll
+     */
+    function setBatch(
```
troykessler: for hooks it is called `setHooks`, should we rather call it `setIsms` and instead of `DomainModule` call it `IsmConfig` to match the existing hook implementation more?

paulbalaji: Done in a3618f1f02: renamed the batch APIs to `setIsms` / `removeIsms`, changed `DomainModule` to `IsmConfig`, and matched the hook config shape with an `ism` field.

paulbalaji: hmmm actually thinking idk anymore - since there it's setHook and setHooks whereas here set and setIsms is inconsistent - hence set/setBatch - wdyt?

nambrot (approved):

solidity/contracts/isms/routing/DomainRoutingIsm.sol
troykessler: still kind of preferring setIsms, but i would leave the final decision to you, approve since the naming is non-blocking and the implementation looks solid

troykessler (approved):
Implementation is solid, would leave the naming question to you

